Issue #38, opened 2019-02-18
isort config inconsistent between local and travis


euri10:
**Describe the bug**
if you run isort locally, it sorts the fastapi package as a 3rd party package, so once you push a branch on travis, isort there will fail
I think https://github.com/timothycrosley/isort/issues/541#issuecomment-307367587 describes it well

**To Reproduce**

**Expected behavior**
to run isort locally and have travis succeed without having to change fastapi import orders

**Environment:**
 - OS: debian buster
 - FastAPI Version [e.g. 0.3.0], get it with: 0.5.0

```Python
import fastapi
print(fastapi.__version__)
```

- Python version, get it with:

```bash
python --version
```

**Additional context**

tiangolo:
I'm sorry for the inconvenience, this was actually lack of documentation.

The trick is to install FastAPI in your local environment with:

```bash
flit install --symlink
```

that way `isort` can detect FastAPI as an installed package and sort all the files accordingly.

I just added contributing guidelines here that should help setting up the local environment:

https://fastapi.tiangolo.com/contributing/

euri10:
oO thanks for the hint, today I learnt something, will ditch that "fix" :turtle:

PR #42, opened 2019-02-18
Add contributing/development docs

tiangolo:
:memo: Add contributing/development docs


tiangolo:
Related to #40 #38 .

As the problem is most probably solved with local development configuration (provided in this guide).

zhenyu-writer:
Hi @tiangolo — I ran a quick user search and found you as the matching maintainer. Could you please take a look at this issue (#42)? It reports intermittent 500 responses when using Depends() together with background tasks; reproduction steps and logs are included in the issue description. If you can confirm whether this is a FastAPI bug or a configuration/uvicorn problem and suggest next steps (or a patch), that would be very helpful. Thanks!

PR #64, opened 2019-03-04
Clarification about possible performance hit

### LINKED ISSUE #33: Clarify run_in_threadpool magic
Good day,

First of all - many thanks for creating this project. Such a relief - you really nailed the proper amount of "batteries included". And docs are great - not a given thing for such a young project.

I'm studying [SQL Database](https://fastapi.tiangolo.com/tutorial/sql-databases/) example where you note that SQLAlchemy is not compatible with async hence you configure path using `def` instead of `async def`. However this example raises and eyebrow because it should block the whole server while database is executing - only that it does not because the [magic](https://github.com/tiangolo/fastapi/blob/5c9c088a2a13e888cdb93eb92d7af72058003397/fastapi/routing.py#L88) you do by running request handlers through threadpool unless they are co-routines.

While this magic is nice convenience, don't you think I can confuse users, particularly those who are less experienced with the whole sync/async co-existence semantics in Python world?

My bottom line is that IMHO it worth adding a clear "Unless you define your request handlers as coroutines FastAPI will run them them in separate thread" note in the "SQL Databases" and "Concurrency and async / await" chapters. What do you think?

Me coming from aiohttp is used to have non-async request handlers for simple methods that don't involve any IO and this magical FastAPI behavior, while useful, would've caused a performance penalty on those methods.

haizaar:
Further technical details towards #33.
Fixes #33.

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -377,6 +377,9 @@ All that is what powers FastAPI (through Starlette) and what makes it have such 
 
 When you declare a *path operation function* with normal `def` instead of `async def`, it is run in an external threadpool that is then awaited, instead of being called directly (as it would block the server).
 
+If you are coming from another async framework that does not have the above heuristics and you are used to define trivial compute-only handlers with plain `def` for a tiny (about 100 nanoseconds) performance gain,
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks! :tada: :taco:

PR #89, opened 2019-03-21
Close the DB even if exceptions are raised

alexiri:
If an exception is raised within the request handler, the database session isn't closed correctly. This fixes that issue.


tiangolo:
Thanks! I think it's a good idea. I'll check it (mainly the docs) soon.

tiangolo:
Great job. I just added an additional note to the docs.

Thanks! :tada: :taco:

PR #91, opened 2019-03-21
Enable branch coverage

blueyed:
Mostly out of curiosity / for completeness, but it should also help with PRs like https://github.com/tiangolo/fastapi/pull/89.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -12,7 +12,7 @@ if [ -f ./test.db ]; then
 fi
 
 export PYTHONPATH=./docs/src
-pytest --cov=fastapi --cov=tests --cov=docs/src --cov-report=term-missing ${@}
+pytest --cov=fastapi --cov=tests --cov=docs/src --cov-branch --cov-report=term-missing ${@}
 mypy fastapi --disallow-untyped-defs --follow-imports=skip
 if [ "${PYTHON_VERSION}" = '3.7' ]; then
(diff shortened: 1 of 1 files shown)


blueyed:
Not too bad..

> but it should also help with PRs like #89.

Oh, that is docs only..

blueyed:
But there are tests for those, and it still covered completely (no except/just try/finally).. I've expected the diff to not be covered with branch coverage initially (https://codecov.io/gh/tiangolo/fastapi/compare/a174f0190120302492b85b8d19955bf677175117...77ab5642d14e7155c5518de788bee0a49538cf46/diff).

From the statuses it looks like Travis wasn't run there however? (but coverage was uploaded apparently)

tiangolo:
I'll close this now as it seems this was just an experiment you were doing.

But feel free to create new PRs adding tests and I'll check them.

blueyed:
I think it is good to have in general.
It changes how coverage is reported, so needs no tests itself - only to cover up detected missing branches then.

tiangolo:
Cool, I would accept PRs adding those corresponding tests.

blueyed:
..and for this it would be useful to have it reported in the first place.. :)  (that's what this was supposed to do)

tiangolo:
Hmm, I don't really want to turn a flag on to add some specific additional reports for tests that I don't see I (or someone else) would be able to commit to adding. I don't see a benefit in just adding the flag.

So, I prefer to add it as part of a PR that implements them. But it's not a priority now.

PR #92, opened 2019-03-21
ci: Travis: dist=xenial

blueyed:


.travis.yml
```diff
@@ -1,10 +1,12 @@
+dist: xenial
```
tiangolo: Why Xenial? What for?

blueyed: It is newer, i.e. upgraded packages etc.
And it provides `3.7` out of the box, instead of 3.7-dev.

tiangolo: Ah! I get it now. I was assuming Travis was using a newer version.

tiangolo:
Great. Thanks! :tada: :taco:

PR #97, opened 2019-03-22
Additional Responses

barsi:
#### Additional Responses Support

**Summary**

allows additional responses to be defined beside the default one

**Example**

```python
# -*- coding: utf-8 -*-
import uuid, datetime as dt

import fastapi
from starlette import responses
import pydantic as dantic

from fastapi.openapi import models, constants

app = fastapi.FastAPI(debug=True)


class Response400(dantic.BaseModel):
    '''HTTP 4xx responses schema'''
    summary: str
    description: str
    ref_code: str # functional error ref


class Point(dantic.BaseModel):
    id: dantic.UUID1
    latitude: float
    longitude: float
    pinned: bool
    pinned_at: dt.datetime


response_403 = models.AdditionalResponse(
    status_code=403,
    description='Forbidden',
    models=[
        Response400,
    ],
)

response_500 = models.AdditionalResponse(
    status_code=500,
    description='Server Error',
    content_type='text/plain',
)


@app.post(
    '/pin-my-location',
    summary='Pin My Location',
    description='Pin my location ...',
    response_model=Point,
    additional_responses=[
        response_403,
        response_500,
    ])
async def pin_my_location():
    if dt.datetime.utcnow().strftime('%Y-%m') != '2019-03':
        return responses.UJSONResponse({
            'summary': 'Access Denied!',
            'detail': 'No Access allowed after Mar 2019.',
            'ref_code': 'TNQ00723',
        },
        status_code=403,)
    return Point(
        id=uuid.uuid1(7, 3),
        latitude=0.0,
        longitude=0.0,
        pinned=True,
        pinned_at=dt.datetime.utcnow(),
    )
```
**Result**

Screen Shot 2019-03-22 at 10 44 42 PM

you can also define additional responses on `app.include_router` so it can save you from repeating the definition of common response types (e.g. `HTTP 500`)

**Limitations**

This pull request only implements the injection of
additional responses definitions into OpenAPI schema
to be appeared in swagger or other OpenAPI 3 compatible
clients (e.g. postman), here are some of the limitations:

- No validation on additional responses if they were returned.
- No support for multiple media types in response content definition.
- Extra fields on response not implemented yet (e.g. `headers` & `examples`).


tiangolo:
Thanks! I'll review it soon.

tiangolo:
Thanks @barsi, I see you put quite some effort into this, thanks for that! :clap: :cake: 

---

I took it further and updated/refactored it all to make it simpler for developers to use while extending the capabilities.

For example, it's possible to add more info to an existing `response_model`, like `example`. And they get the data from the additional response and the model combined.

I also included docs, tests for the code in the docs, etc.

With this you can now add a `responses` parameter with a `dict`, e.g.:

```Python
from fastapi import FastAPI


class Item(BaseModel):
    id: str
    value: str


class Message(BaseModel):
    message: str


app = FastAPI()


@app.get("/items/{item_id}", response_model=Item, responses={404: {"model": Message}})
async def read_item(item_id: str):
    if item_id == "foo":
        return {"id": "foo", "value": "there goes my hero"}
    else:
        return JSONResponse(status_code=404, content={"message": "Item not found"})
```

The new docs are here: https://fastapi.tiangolo.com/tutorial/additional-responses/

barsi:
@tiangolo  Thanks for the brilliant & carefully crafted  framework  🥇 , glad to be a contributor 👍

tiangolo:
Thanks! :blush: :taco: :cake:

PR #105, opened 2019-03-25
Renamed startup_event to shutdown_event

stratosgear:
I think this is a copy and paste error, that causes an error if you copy both startup and shutdown events in your server file.


tiangolo:
Thanks! Good catch! :tada: :taco:

PR #109, opened 2019-03-26
Update response model documentation to explain design choice

### LINKED ISSUE #101: Why not use the return type annotation for the response_model?
**Description**

This is a very minor question on chosen syntax.  Why not use the return type annotation for the response_model? For example rather than:

```python
@app.post("/items/", response_model=Item)
async def create_item(*, item: Item):
    return item
```

You could concievably go:

```python
@app.post("/items/") 
async def create_item(*, item: Item) -> Item:
    return item
```

which just seems like it uses the language facilities better and is a bit more pythonic. There is no real change in feature set, and I am sure i was thought of, but what was the reasoning behind not doing it?

JHSaunders:
Closes #101

### DIFF
--- a/docs/tutorial/response-model.md
+++ b/docs/tutorial/response-model.md
@@ -24,6 +24,8 @@ But most importantly:
 
 * Will limit the output data to that of the model. We'll see how that's important below.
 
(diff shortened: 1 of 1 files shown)


tiangolo:
Great! Thank you. I updated it a bit and merged. :taco: :tada:

PR #145, opened 2019-04-09
Minor offset fix for highlight in docs

blueyed:


tiangolo:
Thanks! I think I probably fixed it at some point, it seems to be right: https://fastapi.tiangolo.com/tutorial/sql-databases/

blueyed:
Yes, looks like it.

tiangolo:
Cool, thanks.

PR #157, opened 2019-04-13
modify doc

hayata-yamamoto:
I modified the code on the query parameter tutorial because `item_id` is required in this case. 

from 
```http://127.0.0.1:8000/items/?short=1```

to 

```http://127.0.0.1:8000/items/foo?short=1```


tiangolo:
Good catch, thanks!

PR #159, opened 2019-04-15
fixed small typo in tutorial

danielmichaels:
### Small Typo 

I have removed a redundant extra word within `docs/tutorials/extra-models.md`

Current:

* The **output model** should do not have a password

Proposed:

* The **output model** should not have a password

A very minor typo. Thanks for the great tutorial and effort that has gone into creating the documentation. :100:


tiangolo:
Oops... Thanks! Good catch.

I'm glad you're enjoying it!

Issue #166, opened 2019-04-19
[FEATURE] Support for Headers/Path/Query on websocket routes


jekirl:
**Is your feature request related to a problem? Please describe.**
I want to be able to use `Header` and other dependencies in a websocket route

**Describe the solution you'd like**
The ability to do this:
```python
@app.websocket_route("/ws")
async def websocket_endpoint(websocket: WebSocket, user_tag: str = Header(None),
    assert user_tag == "foo"
    await websocket.accept()
        await websocket.send_text(f"user_tag: {user_tag}")

async with websockets.connect('ws://0.0.0.0:8000/ws', extra_headers={"User-Tag" "foo"}) as websocket:
    user_tag = await websocket.recv()
    assert user_tag == "user_tag: foo" 


```

**Describe alternatives you've considered**
One bad solution that poorly works without modifying fastapi is to trick `get_dependant` and resolve within function scope, this is obviously not ideal...
```python
@app.websocket_route("/ws")
async def websocket_endpoint(websocket: Request, user_tag: str = Header(None),
    websocket = typing.cast(websocket, WebSocket)
    dep = get_dependant(path="/ws", call=websocket_endpoint)
    values, errors, bg_tasks = await solve_dependencies(request=websocket, dependant=dep)
    assert values["user_tag"] == "foo"
```

A better solution would be to extend `websocket_route` and to adjust the conditional on `get_dependant` to 
```python
        elif lenient_issubclass(param.annotation, Request) or lenient_issubclass(param.annotation, WebSocket):
            dependant.request_param_name = param_name
```

If there already exists a solution for this, great! otherwise I could probably do a PR in the next few days

tiangolo:
I see, it's an interesting idea, but the problem is that it's not supported by browsers: https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api

jekirl:
Looks like browsers only support sending Cookies alongside a websocket connection -- still useful for auth I would assume. Our main use case is from iOS where we use starscream which has strong header support. Would you be open to accepting a PR?

tiangolo:
> Our main use case is from iOS where we use starscream which has strong header support

I get it, great.

> Would you be open to accepting a PR?

Yep, absolutely. :rocket:

tiangolo:
As the PR is already merged and released a couple of versions ago, I'll close this issue now. But feel free to add more comments or create new issues.

PR #178, opened 2019-04-22
Create a WebSocketAPIRoute class to support #166

jekirl:
Initial work at adding dependencies to websocket routes (see #166 ). The way I added it in is probably less than ideal, but figured I would open the PR for now. Still to be done is adding in tests for other dependencies such as Security, Query, Cookie, etc...

### DIFF
--- a/docs/src/websockets/tutorial001.py
+++ b/docs/src/websockets/tutorial001.py
@@ -44,10 +44,9 @@ async def get():
     return HTMLResponse(html)
 
 
-@app.websocket_route("/ws")
+@app.websocket("/ws")
 async def websocket_endpoint(websocket: WebSocket):
     await websocket.accept()
     while True:
         data = await websocket.receive_text()
         await websocket.send_text(f"Message text was: {data}")
-    await websocket.close()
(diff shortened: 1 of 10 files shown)


tiangolo:
Nice! I'll review it soon.

jekirl:
I have a few more things I could add to the PR that we are using internally. i.e. pydantic validated message types for websocket input/output. But I am not sure if others would find that functionality useful.

tiangolo:
Sounds very interesting. I think we could have some optional mechanism to do it (not enforcing it on all cases).

How does it look like? I mean, the code a developer writes using your extensions.

jekirl:
I will try to put together a sample project soon™ ...

jekirl:
Circling back on this, been in our backlog. Semi tricky to separate out the functionality, but haven't forgotten about it!

tiangolo:
Cool! I just refactored this PR a bit, added docs, updated tests and merged.

I also started https://github.com/encode/starlette/pull/527. Once (and if) it is accepted, it will allow us to `raise` inside WebSocket endpoints, sub-dependencies, and handle errors more cleanly.

This is just released in version `0.24.0`, the new docs are here: https://fastapi.tiangolo.com/tutorial/websockets/

PR #195, opened 2019-04-30
Starlette #493 workaround

ricardomomm:
Hi @tiangolo , I'm creating this PR to workaround an issue with Starlette (https://github.com/encode/starlette/issues/493) that avoid developers from reading the Request.Body inside a Middleware or Exception Handler, in my specific case I need to get the original content from body when a JSON deserialization generates an error, with the change bellow I'm able to retrieve the original exception and thus the original content that FastApi tried to parse. It is a really small change that I see no harm. 

While it help me it also provides better exception handling for other developers.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -61,7 +61,7 @@ def get_app(
             logging.error(f"Error getting request body: {e}")
             raise HTTPException(
                 status_code=400, detail="There was an error parsing the body"
-            )
+            ) from e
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks! I'll check it soon.

tiangolo:
Thanks! Merged and released in FastAPI version `0.21.0`. :tada: :cake:

PR #209, opened 2019-05-09
Add py.typed so that this package ships its typing information

meadsteve:
As described in https://www.python.org/dev/peps/pep-0561/


tiangolo:
Cool. Thanks! :cake: :tada: :rocket:

PR #217, opened 2019-05-12
Enhance the performance and avoid the response repeat encoder

bs32g1038:
## Issue

I think the pydantic can cope with the data json serialization. so we should give the data to pydantic  directly and avoid repeatedly use ```jsonable_encoder```. The performance is enhanced obviously, especially in use the ```response_model```

```py
# old code
def serialize_response(*, field: Field = None, response: Response) -> Any:
    encoded = jsonable_encoder(response)
    if field:
        errors = []
        value, errors_ = field.validate(encoded, {}, loc=("response",))
        if isinstance(errors_, ErrorWrapper):
            errors.append(errors_)
        elif isinstance(errors_, list):
            errors.extend(errors_)
        if errors:
            raise ValidationError(errors)
        return jsonable_encoder(value)
    else:
        return encoded
```
```py
# new code
def serialize_response(*, field: Field = None, response: Response) -> Any:
        if field:
        errors = []

        # when return a pydantic model, if the data not convert to json or dict, fields cannot be ignored
        # compatible
        if isinstance(response, BaseModel):
            response = jsonable_encoder(response)

        value, errors_ = field.validate(response, {}, loc=("response",))
        if isinstance(errors_, ErrorWrapper):
            errors.append(errors_)
        elif isinstance(errors_, list):
            errors.extend(errors_)
        if errors:
            raise ValidationError(errors)
        return jsonable_encoder(value)
    else:
        return jsonable_encoder(response)
```
## performance

when you return directly a dict or json, if the data is bigger,  the response time is obviously shortened.

I'm sorry I didn't generate the performance chart

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -21,10 +21,15 @@ from starlette.status import HTTP_422_UNPROCESSABLE_ENTITY
 
 
 def serialize_response(*, field: Field = None, response: Response) -> Any:
-    encoded = jsonable_encoder(response)
     if field:
         errors = []
-        value, errors_ = field.validate(encoded, {}, loc=("response",))
+
+        # when return a pydantic model, if the data not convert to json or dict, fields cannot be ignored
+        # compatible
+        if isinstance(response, BaseModel):
+            response = jsonable_encoder(response)
+
+        value, errors_ = field.validate(response, {}, loc=("response",))
         if isinstance(errors_, ErrorWrapper):
             errors.append(errors_)
         elif isinstance(errors_, list):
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks!

The problem with this is that for it to work it would require the path operation to return a `dict`.

If it returned an ORM model (or similar) it wouldn't work.

But I'm currently refactoring the way that works, to use the Pydantic model's fields to extract the data from whatever object was returned. So, that should help with this. Let's see how it goes.

tiangolo:
I'm currently waiting on https://github.com/samuelcolvin/pydantic/pull/520.

After that, integrating it should help solve this use case.

tiangolo:
Thanks for your effort! :cake: 

This should be solved in recent versions that use the Pydantic model directly. And the docs suggest using ORM mode when decoding something different than a `dict` (like a DB model).

I think this should be covered now, so I'll close this, but feel free to add more comments, issues, etc. :smile:

PR #226, opened 2019-05-14
Add param functions, to override types, to make mypy happy

tiangolo:
:sparkles: Add param functions, to override types, to make `mypy` happy.

This would solve https://github.com/tiangolo/fastapi/issues/219 and other cases of `mypy` complaining about the default types.

## Simple solution

The simplest approach would be to make `mypy` ignore the types in those parameters, using comments with `# type: ignore` like in:

```Python
from fastapi import FastAPI, Path, Query

app = FastAPI()


@app.get("/items/{item_id}")
async def read_items(  # type: ignore
    item_id: int = Path(..., title="The ID of the item to get"),
    q: str = Query(None, alias="item-query"),  
):
    if q:
        results.update({"q": q})
    return results
```

### Drawback

But asking all the users to add comments to ignore type checks each time they use these classes would be difficult and cumbersome (for end users/developers).

## Reason

The issue is that `mypy` is right, from the typing point of view. Putting an instance of `Query` as the "default value" of a parameter with type `str` is not really correct (from the typing point of view).

But here in FastAPI we are "hacking" the typing system, and extending its use to the extreme (which has proven to work really well for many users/use-cases). The *path operation function* is only called by FastAPI (not by the user or any other code), and FastAPI makes sure to pass the correct type, those classes are only for meta-information, used by FastAPI.

## Alternatives

I tried overriding the type declaration of the parameter classes, but there's no way to achieve that.

Then I tried adding a typing stub (`parameters.pyi`), but although it removed `mypy` errors, there where other errors in places used by FastAPI internally.

And then, by relaxing the type declarations to avoid the internal errors, the type hints of those parameters were also lost for final users.

There's a lot of work (and code) there just to have completion for all those parameters in classes... One of the main objectives is for FastAPI to be as comfortable for final developers as possible, even if that is at the expense of the developers of the library itself (me and the community that submits PRs :grin: ).

## Final solution

Here's the solution. We can't override the type declaration of a class (the class itself).

But we can override the return type declaration of a function.

So, I converted all these parameter classes exposed by FastAPI (`Depends`, `Query`, `File`, etc) to functions. But then, as the function is actually called when the code is run and FastAPI needs to inspect it afterwards when the app is running, each of these functions returns the original class instances.

So, the default values of the parameters are actually those classes, and can be checked by the rest of the FastAPI code (dependencies, etc).

### DIFF
--- a/fastapi/__init__.py
+++ b/fastapi/__init__.py
@@ -7,5 +7,15 @@ from starlette.background import BackgroundTasks
 from .applications import FastAPI
 from .datastructures import UploadFile
 from .exceptions import HTTPException
-from .params import Body, Cookie, Depends, File, Form, Header, Path, Query, Security
+from .param_functions import (
+    Body,
+    Cookie,
+    Depends,
+    File,
+    Form,
+    Header,
+    Path,
+    Query,
+    Security,
+)
 from .routing import APIRouter
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -66,7 +66,7 @@ def get_sub_dependant(
         dependency_scopes = depends.scopes
         security_scopes.extend(dependency_scopes)
     if isinstance(dependency, SecurityBase):
-        use_scopes = []
+        use_scopes: List[str] = []
         if isinstance(dependency, (OAuth2, OpenIdConnect)):
             use_scopes = security_scopes
         security_requirement = SecurityRequirement(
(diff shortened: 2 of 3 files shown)


Victor-Savu:
:+1:  This is a very nice use of `Any`! It is an elegant solution. Kudos!

Thank you for placing the emphasis on user experience!

tiangolo:
Thanks @Victor-Savu ! :smile: :tada:

PR #241, opened 2019-05-20
handle swagger, redoc and openapi with coroutine function

### LINKED ISSUE #238: [FEATURE]: make swagger_ui_html, redoc_html and openapi.json handled by async function?
I add a loop on all my router of my app to check if the handler are all coroutine function to get better performance, but I fount that swagger_ui_html, redoc_html and openapi.json are handled by normal function.

https://github.com/tiangolo/fastapi/blob/56ab106bbbf8054af437821c6683491ca7952c3b/fastapi/applications.py#L80-L86

I'm wondering if its possible to handle these 3 router with async function, as its only a simple `getattr` operator or concat string.

https://github.com/tiangolo/fastapi/blob/56ab106bbbf8054af437821c6683491ca7952c3b/fastapi/applications.py#L78

```python
    def setup(self) -> None:
        if self.openapi_url:

            async def openapi_handler(req):
                return JSONResponse(self.openapi())

            self.add_route(
                self.openapi_url,
                openapi_handler,
                include_in_schema=False,
            )
```
**Describe alternatives you've considered**

Disable default handlers and I write async handlers by my self to replace inner handlers.

**Additional context**

In `Starlette`, if a handler is not coroutine function, it will be executed by `loop.run_in_executor` . So this could improve performance a litte bit.(In my test, about 2800 r/s to 3200r/s)

trim21:
close #238

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -9,7 +9,7 @@ from starlette.applications import Starlette
 from starlette.exceptions import ExceptionMiddleware, HTTPException
 from starlette.middleware.errors import ServerErrorMiddleware
 from starlette.requests import Request
-from starlette.responses import JSONResponse, Response
+from starlette.responses import HTMLResponse, JSONResponse, Response
 from starlette.routing import BaseRoute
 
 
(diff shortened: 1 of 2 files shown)


fastapi/applications.py
```diff
@@ -79,29 +79,28 @@ def openapi(self) -> Dict:
...
-            self.add_route(
-                self.docs_url,
-                lambda r: get_swagger_ui_html(
-                    openapi_url=self.openapi_prefix + self.openapi_url,
-                    title=self.title + " - Swagger UI",
-                ),
-                include_in_schema=False,
-            )
+
+            async def swagger_ui_html(req: Request) -> HTMLResponse:
```
tiangolo: Please move this function to `fastapi.openapi.docs`, and import it from there.

And please rename it to `swagger_ui_route` or something similar.

fastapi/applications.py
```diff
@@ -79,29 +79,28 @@ def openapi(self) -> Dict:
...
-            self.add_route(
-                self.redoc_url,
-                lambda r: get_redoc_html(
-                    openapi_url=self.openapi_prefix + self.openapi_url,
-                    title=self.title + " - ReDoc",
-                ),
-                include_in_schema=False,
-            )
+
+            async def redoc_html(req: Request) -> HTMLResponse:
```
tiangolo: Please move this function to `fastapi.openapi.docs`, and import it from there.

And please rename it to `redoc_route` or something similar.

tiangolo (changes requested):
Awesome! That was fast! :zap: 

Please help me refactor a couple of things to move some stuff to `fastapi.openapi.docs`.

Thanks!

fastapi/applications.py
Trim21: Is `swagger_ui_handler` OK?

Trim21:
Sorry for wrong click

tiangolo (approved):

tiangolo:
Thanks for implementing the changes, you did it correctly. And it made me realize all the complexity involved in having them as separate functions, as they depended on variables from outside. So I reverted it to your original implementation.

I just merged it. Thank you! :tada: :rocket: :cake:

Issue #242, opened 2019-05-21
[FEATURE] skip_defaults for path operations


wshayes:
**Is your feature request related to a problem? Please describe.**

Need to exercise the skip_defaults option for pydantic json serialization so that Optional elements do not get serialized when they are nulls (see [Pydantic issue](https://github.com/samuelcolvin/pydantic/issues/530). 

**Describe the solution you'd like**
Allow the ability to set skip_defaults for endpoints (path operations) so that null values are not emitted in the JSON serializations.

I can take a stab at putting together a PR for this. I didn't realize my front-end colleagues wanted this cleaned up so urgently.

tiangolo:
Thanks for adding it here! That helps us track the state of this feature.

If you can start a PR that would be great, otherwise I'll do it soon (but I'm fixing a couple of things first).

tiangolo:
As the PR is already merged and it is available in the latest versions, I'll close this issue now. But feel free to add more comments or create new issues.

PR #248, opened 2019-05-22
Add skip_defaults support for path operations (for #242)

wshayes:
Running into trouble getting skip_defaults passed correctly. I’m not understanding the full flow of the logic from the route decorators to the response serialization.

This is a Starlette class which has include_in_schema but does not have skip_defaults - how do we handle this? Do we need to?

    # From routing.py - line 331
    elif isinstance(route, routing.Route):
        self.add_route(
            prefix + route.path,
            route.endpoint,
            methods=route.methods,
            include_in_schema=route.include_in_schema,
            name=route.name,
        )

Haven't done any websocket work - not sure if websocket route needs to be handled for this (probably does)

I would appreciate pointers/help to finish this PR.

### DIFF
--- a/docs/src/response_model/tutorial001.py
+++ b/docs/src/response_model/tutorial001.py
@@ -1,4 +1,4 @@
-from typing import Set
+from typing import List
 
 from fastapi import FastAPI
 from pydantic import BaseModel
@@ -11,9 +11,9 @@ class Item(BaseModel):
     description: str = None
     price: float
     tax: float = None
-    tags: Set[str] = []
+    tags: List[str] = []
 
 
 @app.post("/items/", response_model=Item)
-async def create_item(*, item: Item):
+async def create_item(item: Item):
     return item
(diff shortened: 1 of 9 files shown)


tiangolo:
Thanks @wshayes ! I'll check it soon.

tiangolo:
Thank you @wshayes !

I updated it a bit, added docs, tests, etc.

I renamed the parameter to `response_model_skip_defaults` to be more explicit and intuitive.

Although I think it's too long, I think we can start with a "too long" completely explicit version and then update it to a shorter version, like `response_skip_defaults` or `model_skip_defaults`.

The updated docs are here: https://fastapi.tiangolo.com/tutorial/response-model/#response-model-encoding-parameters

It will be included in the next release.

Thanks! :tada: :rocket: :cake:

wshayes:
Thank you - sorry I couldn't complete it on my own. Very nice job with the documentation. I'll study the changes you made so I can understand the flow better.

tiangolo:
Thanks for starting it @wshayes !

PR #254, opened 2019-05-24
Mypy && Depends

meadsteve:
I've recently been using FastAPI with mypy. Mostly this has been working really well but I've run into problems when using `Depends` as it doesn't type correctly. For example the following:
```python
    @app.get("/")
    async def read_root(user: User = Depends(get_current_user)):
        pass
```
will raise the following error in mypy:
```
error: Incompatible default for argument "user" (default has type "Depends", argument has type "User")
```

One solution I thought of is in this PR. Providing functions `depends` and `depends_on` means the typing can match whilst still working in the same way.

What do you think? (The same approach could work for the other functions in `param_functions.py`)

### DIFF
--- a/fastapi/__init__.py
+++ b/fastapi/__init__.py
@@ -11,6 +11,8 @@ from .param_functions import (
     Body,
     Cookie,
     Depends,
+    depends,
+    depends_on,
     File,
     Form,
     Header,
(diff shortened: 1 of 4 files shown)


tiangolo:
Thanks! Indeed, it was also happening for `Query`, `Body`, etc.

It was solved in version `0.22.0`: https://fastapi.tiangolo.com/release-notes/#0220

Since that version, `Depends`, when imported from `fastapi`, is a function (very similar to your approach).

And the same for the others, `Query`, `Body`, etc.

meadsteve:
Oh I missed that 👍 thanks

tiangolo:
Cool! Thanks.

PR #256, opened 2019-05-24
Update path-params.md

wshayes:
Add path parameter converters (e.g. starlette urlconverters) documentation.


tiangolo:
Thanks! I'll check it soon.

tiangolo:
Thank you @wshayes ! Checking this I realized that paths in path operations could be improved to not require a `Path(...)`.

I fixed/improved it at: https://github.com/tiangolo/fastapi/pull/272

Then I updated the docs here a bit, to include a source file in the docs instead of having the Python code copied. That allows the source code to be tested. That way we are sure that the code in the docs runs correctly, and helps us ensure that features are properly documented.

Now it's merged. Thanks! :tada: :rocket:

PR #262, opened 2019-05-27
Update signature of Depends

meadsteve:
This enables mypy (and IDEs) to infer the type coming from Depends

### DIFF
--- a/fastapi/param_functions.py
+++ b/fastapi/param_functions.py
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Sequence
+from typing import Any, Callable, Sequence, TypeVar
 
 from fastapi import params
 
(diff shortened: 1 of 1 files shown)


fastapi/param_functions.py
```diff
@@ -238,8 +238,13 @@ def File(  # noqa: N802
...
 
-def Depends(dependency: Callable = None) -> Any:  # noqa: N802
-    return params.Depends(dependency=dependency)
+DependencyType = TypeVar("DependencyType")
+
+
+def Depends(
+    dependency: Callable[..., DependencyType] = None
+) -> DependencyType:  # noqa: N802
+    return params.Depends(dependency=dependency)  # type: ignore
```
meadsteve: This is the one downside of my suggest change. It makes the type signature wrong from the framework's point of view. However the upside is that it helps catch bugs in the consumer's code

tiangolo:
This looks very nice! I'll check it thoroughly soon.

tiangolo (commented):
I love the idea. Pretty smart typing usage!

Nevertheless, let's wait for editors to support this better: https://github.com/microsoft/python-language-server/issues/1014

After that is solved, I'll be happy to include this.

But before that, this type declaration ends up overriding the explicit type declaration in a path operation, at least in some editors.

So, for editors that don't support this feature yet, this breaks the development experience. So, let's give them some time to catch up.

meadsteve:
ah that's interesting that it overrides other type declarations in some editors. Shall we close this PR for now and re-implement in the future when the IDE's have caught up?

PR #264, opened 2019-05-27
Add other Pydantic dict params to path operations

tiangolo:
Add other Pydantic dict params to path operations.

To include all the parameters supported by Pydantic's model's `.dict()`.


bs32g1038:
To be honest, these are good features, but I think it makes ```get```, or ```other request functions``` bloated and less expansive.
* too many fields are confusing. I think the function should keep simple. if we need select fields,  that may be provide a extra higher order function to copy with the Pydantic model. what's more, directly using the the Pydantic model combination.
* It doesn't make the code clearer and effectively reduce code duplication. most of data are composed of multiple models in daily use.

----

In my opinion, if the docs can auto generate the doc data by the real data, it will be a great thing.
```

class Item(BaseModel):
    name: str = None
    description: str = None
	
items = {
    "foo": {"name": "Foo", "price": 50.2},
    "bar": {"name": "Bar"},
}
	

@app.get("/items/{item_id}", response_model=Item)
async def read_item(item_id: str):
    return items[item_id]


# result
only show the name field in the doc model

because the response result only include the name field in the model. 

```

Besides, Fastapi may be provide a utils function to reduce code repeat.
```

class Name(BaseModel):
    name: str = None

class Description(BaseModel):
    description: str = None

class Base(BaseModel):
    code: int = None
    data: combination({
        items: combination(Name, Description),
        count: int
    }) = None

items = {
    "foo": {
        "code": 0,
        "data": {
            "items": [],
            "count": 0
        }
    }
}
	

@app.get("/items/{item_id}", response_model=Item)
async def read_item(item_id: str):
    return items[item_id]
```

To sum up, I look forward to these features very much.

tiangolo:
@bs32g1038 these additional parameters are mainly for completeness.

The name of the parameters is temporal, as I think it's quite long. But I don't expect users to use them a lot. That's why they are at the end of the function.

But for completeness, it might help some use cases.

About your first use case, you can enable that behavior with one of these parameters (from a previous PR): `response_model_skip_defaults=True`: https://fastapi.tiangolo.com/tutorial/response-model/#use-the-response_model_skip_defaults-parameter

That's something that can help some use cases, but others might need to keep the default values.

bs32g1038:
@tiangolo thinks, I got it!

PR #278, opened 2019-05-30
Refactor param extraction using Pydantic Field

### LINKED ISSUE #179: Headers should be typed as Optional
Setting headers for routes requires adding invalid type information because while headers are optional, they cannot be typed as optional.

Headers that are typed as `str` should error if they are not provided. Otherwise, headers should be typed as `Optional` if they are so.

**Describe alternatives you've considered**
N/A

**Additional context**

`x_custom_header` will not be `str` if the request doesn't include it.
```python
async def example_route(
    payload: dict, *, x_custom_header: str = Header(None)
):
    assert x_custom_header is not None
    return {"msg": "ok"}

```

Changing the type hint to `typing.Optional[str]` raises an assertion error: `AssertionError('Parameters for Query and Header must be of type str, int, float, bool, list, tuple or set...`

tiangolo:
:sparkles: Refactor param extraction using Pydantic Field.

This fixes https://github.com/tiangolo/fastapi/issues/179.

Adds support for `Enum` in path and query parameters.

Docs about how to use `Enum`.

Large refactor, improvement, and simplification of how params are handled.

This should improve support for a wide range of types supported by Pydantic, used in FastAPI params.


tiangolo:
This also fixes https://github.com/tiangolo/fastapi/issues/263

PR #282, opened 2019-06-01
Fix HTTP Bearer security auto-error

tiangolo:
:bug: Fix HTTP Bearer security auto-error.

Currently, even with `auto_error=False`, it would raise an error if there's an Authorization header with an incorrect scheme (e.g. `Basic` instead of `Bearer`).

It should not raise errors automatically if `auto_error=False`, just return `None`.

### DIFF
--- a/fastapi/security/http.py
+++ b/fastapi/security/http.py
@@ -112,10 +112,13 @@ class HTTPBearer(HTTPBase):
             else:
                 return None
         if scheme.lower() != "bearer":
-            raise HTTPException(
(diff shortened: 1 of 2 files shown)


pepegc:
Hi @tiangolo, I would like to have endpoints that can accommodate both authenticated and non authenticated users. So instead of raising an error I just want a `user` equal to `None`. I was going to implement my own "lax" version of OAuth2PasswordBearer to do this when I came across  this `auto_error` option. Is this the intended use for this?

PR #288, opened 2019-06-03
Fix default response model docs

awiddersheim:
Fix a discrepancy in the `tax` parameters default value between the docs
and the code example.


tiangolo:
Nice catch! Thank you! :tada: :cake: :rocket:

PR #313, opened 2019-06-14
default the description of additional response to status reason phrase

duxiaoyao:


fastapi/openapi/utils.py
```diff
@@ -187,7 +188,7 @@ def get_openapi_path(
                         response.setdefault("content", {}).setdefault(
                             "application/json", {}
                         )["schema"] = response_schema
-                    response.setdefault("description", "Additional Response")
+                    response.setdefault("description", http.client.responses[int(additional_status_code)])
```
dmontagu: Probably worth using `http.client.responses.get(int(additional_status_code), "Additional Response")` in case someone uses a custom response code

duxiaoyao: Thanks your review. Done as what you suggest.

tiangolo:
Thanks! Good job :cake: :tada:

fastapi/openapi/utils.py
tiangolo: Thanks @dmontagu ! Good idea.

PR #314, opened 2019-06-14
Update routing.py

dmontagu:
https://github.com/encode/starlette/pull/541

The last comment in that pull request's review seems related to these lines.


tiangolo:
Yep. you're right. Good catch. :heavy_check_mark: Thanks! :tada:

PR #326, opened 2019-06-21
Schema regression fix for FastAPI

wongpat:
This is part of the fix for samuelcolvin/pydantic#613 - there is probably a better way to do this :)


wongpat:
Not needed anymore once samuelcolvin/pydantic#621 gets merged in.

PR #333, opened 2019-06-25
Fix schema generation for nested/repeated models

dmontagu:
Remove definitions update

Addresses #332

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -39,10 +39,9 @@ def get_model_definitions(
 ) -> Dict[str, Any]:
     definitions: Dict[str, Dict] = {}
     for model in flat_models:
-        m_schema, m_definitions = model_process_schema(
+        m_schema, _ = model_process_schema(
             model, model_name_map=model_name_map, ref_prefix=REF_PREFIX
         )
-        definitions.update(m_definitions)
         model_name = model_name_map[model]
         definitions[model_name] = m_schema
     return definitions
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -38,7 +38,7 @@ def get_model_definitions(
     *, flat_models: Set[Type[BaseModel]], model_name_map: Dict[Type[BaseModel], str]
 ) -> Dict[str, Any]:
     definitions: Dict[str, Dict] = {}
-    for model in flat_models:
+    for model in sorted(flat_models, key=lambda x: x.__name__):
         m_schema, _ = model_process_schema(
             model, model_name_map=model_name_map, ref_prefix=REF_PREFIX
         )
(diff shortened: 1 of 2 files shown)


fastapi/utils.py
```diff
@@ -39,10 +39,9 @@ def get_model_definitions(
 ) -> Dict[str, Any]:
     definitions: Dict[str, Dict] = {}
     for model in flat_models:
-        m_schema, m_definitions = model_process_schema(
+        m_schema, _ = model_process_schema(
             model, model_name_map=model_name_map, ref_prefix=REF_PREFIX
         )
-        definitions.update(m_definitions)
```
dmontagu: I believe the update should ultimately be redundant given the iteration over `flat_models`; either way this fixes a bug with my server.

I don't fully understand why this is the case yet, but I will dig a little deeper.

fastapi/utils.py
```diff
@@ -38,11 +38,10 @@ def get_model_definitions(
     *, flat_models: Set[Type[BaseModel]], model_name_map: Dict[Type[BaseModel], str]
 ) -> Dict[str, Any]:
     definitions: Dict[str, Dict] = {}
-    for model in flat_models:
-        m_schema, m_definitions = model_process_schema(
+    for model in sorted(flat_models, key=lambda x: x.__name__):
```
dmontagu: I added the sorting (by name) here to ensure reproducibility of the test I added. I assume the cost of this sort is negligible for practical purposes (given this is just used to generate the schema).

tests/test_get_openapi.py
```diff
@@ -0,0 +1,23 @@
...
+
+    class Model3(BaseModel):
+        c: Model
+        d: Model2
+
+    @app.get("/", response_model=Model3)
+    def f():
+        pass
+
+    openapi = app.openapi()
```
dmontagu: The test will fail at this line if the `definitions.update(m_definitions)` is not removed above.

dmontagu:
@tiangolo it looks like this is addressed by https://github.com/samuelcolvin/pydantic/pull/621 -- if that's right I'll close this pull request, but I think it could still make sense to add the schema-generation test I wrote here. (I don't care about the attribution here so feel free to just copy the test into another PR.)

Note that the schema generation is currently not deterministic due to iteration over the set of models; it might make sense to change this for the sake of testability (given the minimal performance implications).

tiangolo:
Thanks @dmontagu ! I'll come back to this once we have a new Pydantic release with https://github.com/samuelcolvin/pydantic/pull/621.

We can leave this open, that will remind me to include your test.

tiangolo:
Thanks @dmontagu ! I added your test in #385. So we can close this one. Thanks! :cake:

PR #346, opened 2019-06-28
Add Python 3.8 to CI for tests

tiangolo:
:construction_worker: Add Python 3.8 to CI for tests.


tiangolo:
This won't work until there's an official Pydantic for Python 3.8. Or the tests would have to compile Pydantic (with all the Cython requirements/steps). That would increase complexity too much for not much benefit at the moment.

PR #356, opened 2019-07-01
add test code sample for graphql(sync and async) in the tutorial doc

hackrole:
Hi, I add a code sample in the grpahql tutorial, to make the newbee wonot spend hours in writing testcases for grpahql, excepicilly async graphql testcasee.

If there any more I can do for this, just let me know, I will do my best to improve it.

### DIFF
--- a/docs/tutorial/graphql.md
+++ b/docs/tutorial/graphql.md
@@ -32,6 +32,41 @@ You will see GraphiQL web user interface:
 
 <img src="/img/tutorial/graphql/image01.png">
 
+## Write test for it
+
+```python
+# run with pytest
+import graphene
+from graphene.test import Client
+from graphql.execution.executors.asyncio import AsyncioExecutor
+
+
+class Query(graphene.ObjectType):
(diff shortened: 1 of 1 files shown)


tiangolo:
Thank you! I think this is very specific to Graphene, in fact, the test doesn't even need/use any FastAPI code.

I think this should be in Graphene docs, or maybe it's covered here: https://docs.graphene-python.org/en/latest/testing/

I would prefer not to add a lot of docs about other tools as Graphene, those docs are better covered in Graphene itself. E.g. the same way I would prefer not to cover a lot of the specifics about MySQL or MongoDB, etc. Also, those docs here could become obsolete very easily if they update them there.

---

But another thing I think you could do (and I encourage you to) is to write a blog post about it, and how to integrate Graphene with FastAPI, use Graphene tests, etc. Then we could link to your post from here: https://fastapi.tiangolo.com/external-links/

I'll close this now, but feel free to add more comments, issues, PRs.

PR #366, opened 2019-07-04
Change limit default parameter to 10

Smashman:
Rest of docs reference 10 as the default.


tiangolo:
Thanks! Good catch. :memo: :rocket:

PR #367, opened 2019-07-04
Update starlette to 0.12.2

dedsm:
better support for mypy in request.state

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -25,7 +25,7 @@ sqlalchemy = "*"
 uvicorn = "*"
 
 [packages]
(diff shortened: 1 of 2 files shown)


tiangolo:
Thank you! :cake: I just updated it to include the latest and rebased. It's merged now :rocket: :tada:

wozniakty:
Woooo 🥂

wozniakty:
This also fixes redirects.  Now any route that ends in a slash can be properly hit without the trailing slash, if redirect is on.

PR #370, opened 2019-07-05
relax the python version so 3.7 can be used for dev

cjw296:


tiangolo:
Thanks!

This makes sense, and given `Pipfile` claims to support PEP 508 specifiers, it should work.

...but it doesn't. It's not supported by Pipenv :disappointed: https://github.com/pypa/pipenv/issues/1050

---

For now, you can create a `conda` environment (or any other way you use to install multiple Pythons) with Python 3.7 version, etc.

Or you can have a modified version with Poetry, although as it uses the same `pyproject.toml` I don't know if it might interfere.

This gives me another reason to try Poetry. Or replace the development environment setup to pure `venv` and Flit`. I'll explore it later...

cjw296:
That's weird, this has been working fine for me locally with Pipenv for a few months now. I put in the PR as I was tired of having a local uncommitted change. Looks like your CI was happy enough with it too.

Keen to understand where this didn't work for you?

Would you accept/prefer a PR that fipped the project to use Poetry?

PR #371, opened 2019-07-05
documentation tweak

cjw296:
I couldn't follow these chunks very easily so had a go at wordsmithing them, please let me know what you think.

One thing from just below it: it's unclear to me how many times get_current_user is called if different dependencies have different scope requirements. These docs imply it will be called once for each distinct list of scopes, but how does that interact with dependency caching?


tiangolo:
Thanks! I also think this section feels complex.

I hope to create a diagram later to try and make it more intuitive.

PR #384, opened 2019-07-12
Upgrade pydantic to 0.30

jekirl:
This bumps the pydantic version to 0.30. It may be worth waiting on merging this until pydantic 0.31

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -26,7 +26,7 @@ uvicorn = "*"
 
 [packages]
 starlette = "==0.12.0"
-pydantic = "==0.29.0"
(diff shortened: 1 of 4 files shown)


tiangolo:
Great job! :tada: 

I just updated it to pin Pydantic to `0.30`, as `0.31` is not released yet and could include breaking changes.

It's perfect to have these changes for each release, I prefer to have frequent independent releases with each dependency version when possible. In case a future version breaks something, users can still use the features and bug fixes in this version.

Thanks! :tada: :rocket: :cake:

PR #398, opened 2019-07-22
Update Pydantic tu support up to 0.31

haizaar:


haizaar:
Ready for merging :)

MichalMazurek:
Hi Guys,

One proposition, wouldn't it be benficial to add all minor updates for pydantic 0.30.x? As the minor version updates are only for bug fixes and it should not break any features.

All the best,
Michal

dmontagu:
@haizaar there is a new pydantic release 😬 would be great to use `pydantic==0.31` if nothing breaks

haizaar:
@dmontagu No worries. Updated the PR.

tiangolo:
Thanks @haizaar for your work! :cake: 

@MichalMazurek FastAPI integrates deeply with the internals of Pydantic, things that are not exposed publicly (in the docs), so, any change could break the integration.

---

I prefer to have dependency updates separated (Starlette and Pydantic), that way I can release more granular versions, in case something is broken in one but not the other, users could still use the previous latest version.

I just tried to rebase it to release it in a version with a range including the previous versions of Pydantic up to 0.31, before merging @dmontagu's #423 (as that requires exactly `0.32` or above), but I'm hitting a JSON encoding error for enums :scream:, what I don't understand clearly is how the tests were passing before the rebase :man_shrugging:.

It seems #423 (I guess Pydantic 0.32) fixes the Enum JSON encoding error, so I will merge that.

I'll close this now, but feel free to add more comments, issues, PRs :smile:

PR #400, opened 2019-07-22
Fix union body

koxudaxi:
The PR fix union body bug.

## Detail
FastAPI can't handle a body as union type.
I found parameter detected condition is wrong, which don't care union type.

# Related issue
https://github.com/tiangolo/fastapi/issues/300


tiangolo:
Thank you @koxudaxi ! Great job debugging and finding the problem. :cake: :rocket: 

I just updated the implementation a bit to avoid the internal declaration of a sub-function and added a couple of comments in the tests to clarify why `Union`s with inheritance are only supported in Python 3.7.

PR #411, opened 2019-07-30
plural-s without apostrophe

bronsen:


tiangolo:
Good catch, thanks! :tada: :memo:

Issue #416, opened 2019-08-01
[BUG] Header duplicate on subdependency


henriklindgren:
**Describe the bug**
Getting duplicate headers on API methods with subdependencies requiring the same dependency with a Header argument.

**To Reproduce**
I've created an example repo to reproduce https://github.com/henriklindgren/fastapi-header-bug
1. Create an API-method which uses the same dependency, with a Header argument, in each of 2 paths of the dependency tree. Like this https://github.com/henriklindgren/fastapi-header-bug/blob/master/app/main.py
2. Visit generated OpenApi json and see duplicate header requirements for the method. 
```json
"parameters": [
  {
    "required": true,
    "schema": {
      "title": "Someheader",
      "type": "string"
    },
    "name": "someheader",
    "in": "header"
  },
  {
    "required": true,
    "schema": {
      "title": "Someheader",
      "type": "string"
    },
    "name": "someheader",
    "in": "header"
  }
]
```

**Expected behavior**
A single header requirement shown in the OpenApi definition.

**Environment:**
 - OS: Linux
 - FastAPI Version: 0.33.0
 - Python Version: 3.7.3

dmontagu:
Here's a failing test case based on the provided repo:

```python
from starlette.testclient import TestClient

from fastapi import FastAPI, Header, Depends

app = FastAPI(title='Access', redoc_url=None)


def get_header(someheader: str = Header(...)):
    return someheader


def get_something_else(someheader: str = Depends(get_header)):
    return f'{someheader}123'


@app.get('/')
def get_deps(dep1: str = Depends(get_header), dep2: str = Depends(get_something_else)):
    return {'dep1': dep1, 'dep2': dep2}


client = TestClient(app)


def test_double_header():
    response = client.get("/openapi.json")
    assert response.status_code == 200
    parameters = response.json()['paths']['/']['get']['parameters']
    assert len(parameters) == 1, str(parameters)
```
Looking into whether it's a simple fix now...

dmontagu:
I figured out the source of the issue -- it's because during openapi spec generation, the raw flat_dependant is used, which includes duplicate dependencies (which are ignored or not based on the caching). However, I can't imagine a scenario where this distinction is relevant externally -- whether the dependency is cached or not, you can't provide distinct inputs for each repeated (uncached) call.

[...]

PR #417, opened 2019-08-01
Use caching logic to determine openapi spec

dmontagu:
Dependencies that resolve to the same cache key are not re-used while building the openapi schema.

Addresses #416

### DIFF
--- a/tests/test_repeated_dependency_schema.py
+++ b/tests/test_repeated_dependency_schema.py
@@ -2,7 +2,7 @@ from fastapi import Depends, FastAPI, Header
 from starlette.status import HTTP_200_OK
 from starlette.testclient import TestClient
 
-app = FastAPI(title="Access", redoc_url=None)
+app = FastAPI()
 
 
 def get_header(*, someheader: str = Header(...)):
@@ -50,7 +50,7 @@ schema = {
             },
         }
     },
-    "info": {"title": "Access", "version": "0.1.0"},
+    "info": {"title": "Fast API", "version": "0.1.0"},
     "openapi": "3.0.2",
     "paths": {
         "/": {
@@ -89,7 +89,7 @@ schema = {
 
 def test_schema():
     response = client.get("/openapi.json")
-    assert response.status_code == 200
+    assert response.status_code == HTTP_200_OK
     actual_schema = response.json()
     assert actual_schema == schema
     assert (
--- a/tests/test_repeated_dependency_schema.py
+++ b/tests/test_repeated_dependency_schema.py
@@ -1,0 +1,103 @@
+from fastapi import Depends, FastAPI, Header
+from starlette.status import HTTP_200_OK
+from starlette.testclient import TestClient
+
+app = FastAPI(title="Access", redoc_url=None)
+
+
+def get_header(*, someheader: str = Header(...)):
+    return someheader
+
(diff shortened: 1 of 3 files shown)


dmontagu:
There's a nuance here related to security -- as implemented, if you were to have the same callable in multiple places with different security, you'd get repeats of the dependency in the schema.

In theory, this could be avoided if there was a way to "merge" all of the relevant security scopes. I'm not sure whether that is possible, and it seems like enough of an edge case to not handle for now.

tests/test_repeated_dependency_schema.py
```diff
@@ -0,0 +1,103 @@
+from fastapi import Depends, FastAPI, Header
+from starlette.status import HTTP_200_OK
+from starlette.testclient import TestClient
+
+app = FastAPI(title="Access", redoc_url=None)
```
henriklindgren: could remove title and redoc arguments here that slipped in from my project, sorry :)

tests/test_repeated_dependency_schema.py
```diff
@@ -0,0 +1,103 @@
...
+                "summary": "Get Deps",
+            }
+        }
+    },
+}
+
+
+def test_schema():
+    response = client.get("/openapi.json")
+    assert response.status_code == 200
```
henriklindgren: could use HTTP_200_OK here too

henriklindgren:
@dmontagu have you considered if this affects https://fastapi.tiangolo.com/tutorial/header-params/#duplicate-headers?

dmontagu:
@henriklindgren That's a good question, but I think this is still safe -- it only modifies things when calling the exact same dependency function. If you were to list a header argument multiple times in *different* dependencies, it would still get added multiple times.

For what it's worth, if I understand correctly, any schema that would be changed by this is probably invalid already anyway.

henriklindgren:
Ah true, read through it all now, agreed. The way I noticed the bug was from generating a client which broke from it, I would be surprised if people would use it this way on purpose so it should be safe.

>  if you were to have the same callable as a dependency in multiple places with different security schemes, you'd get a repeat of the dependency's arguments in the schema for each different security scheme

Let's see if I've understood this part correctly. It seems safe to ignore this case because it would be really awkward if multiple security schemes in the same context based key values on the order they arrived, dependent on other schemes. It honestly feels safer to know that you would get the single argument and be sure of it. Otherwise you'd use argument lists as mentioned (such as the duplicate header list), which could be used as a workaround if you really want this behaviour.

tiangolo:
Great! Thanks @dmontagu for the thorough investigation and explanation, here and in the issue. :cake: :rocket:

---

I think we can have this merged as is right now.

Combining the scopes of the same security schemes could make sense, but that would mean "in OpenAPI terms" that *all* the scopes are required. And the intention could be to have *any* of them required. That's actually not enforced by FastAPI but by the implementation itself. If the user/developer leaves the `auto_error=True`, then *all* the security schemes would be required. But if he has `auto_error=False` and then some additional logic to decide if the operation is allowed, then it could be a different case.

This is a pretty advanced use case of OpenAPI, having multiple security schemes combined that some can be required or not.

I think we can leave it as-is (with this PR merged) for now and see if there's more logic required for that for some use case later.

PR #422, opened 2019-08-06
Fix unskipped defaults

dmontagu:
Addresses https://github.com/tiangolo/fastapi/issues/421

Currently, if you return a `BaseModel` instance in a route with a `BaseModel` response_type, the `skip_defaults` value does not influence the result. This PR fixes this.

-----
**Details**

The source of the problem is that calling `field.validate` on a model results in a new model with the fields set. I dug in on the pydantic side and it looks like this is due to the recent change to re-parse the model with a non-subclass type when response_model is set. If you look in `BaseModel.validate` (which is ultimately called if you follow it deep enough), you can see the precise spot where defaults aren't being skipped:

```python
@classmethod
    def validate(cls: Type['Model'], value: Any) -> 'Model':
        if isinstance(value, dict):
            return cls.parse_obj(value)
        elif isinstance(value, cls):
            return value.copy()
        elif cls.__config__.orm_mode:
            return cls.from_orm(value)
        else:
            with change_exception(DictError, TypeError, ValueError):
                return cls.parse_obj(value)
```

It calls into `parse_obj` on the last line, which subsequently calls `dict` on the object and then parses the dict. Unfortunately, this means defaults don't get skipped.

**I don't think it makes sense to change this on the pydantic side** because otherwise you'd run into problems parsing models with default values into non-subclass models with required values. So, if we want to stick with this approach of creating a new class to force a reparse, we probably need to handle the 

----

That said, I would also be in favor of a larger refactor if we could come up with a way to get the desired field safely without another round of parsing/dumping 😬. But that does seem hard.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -53,7 +53,7 @@ def serialize_response(
         if errors:
             raise ValidationError(errors)
         if skip_defaults and isinstance(response, BaseModel):
-            exclude |= value.__fields_set__.difference(response.__fields_set__)
+            value = response.dict(skip_defaults=skip_defaults)
         return jsonable_encoder(
             value,
             include=include,
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -52,6 +52,8 @@ def serialize_response(
             errors.extend(errors_)
         if errors:
             raise ValidationError(errors)
+        if skip_defaults and isinstance(response, BaseModel):
+            exclude |= value.__fields_set__.difference(response.__fields_set__)
         return jsonable_encoder(
             value,
             include=include,
--- a/tests/test_skip_defaults.py
+++ b/tests/test_skip_defaults.py
@@ -4,17 +4,26 @@ from fastapi import FastAPI
 from pydantic import BaseModel
 from starlette.testclient import TestClient
 
+app = FastAPI()
+
+
+class SubModel(BaseModel):
+    a: Optional[str] = "foo"
+
+
+class Model(BaseModel):
+    x: Optional[int]
+    sub: SubModel
+
+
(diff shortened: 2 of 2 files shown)


fastapi/routing.py
```diff
@@ -52,6 +52,8 @@ def serialize_response(
             errors.extend(errors_)
         if errors:
             raise ValidationError(errors)
+        if skip_defaults and isinstance(response, BaseModel):
+            exclude |= value.__fields_set__.difference(response.__fields_set__)
         return jsonable_encoder(
```
dmontagu:
```python
        if isinstance(response, BaseModel):
            value = response.dict(skip_defaults=skip_defaults)
```
might also work, but I'm not sure if that would play differently with include/exclude/`jsonable_encoder`.

tiangolo:
Wow, you had to go quite deep into the rabbit hole to find that (`field.validate` and `BaseModel.validate` are quite deep inside the callback flow in Pydantic). Good job! Thanks! :clap: :rocket:

I updated the test to check for defaults in sub-models. And updated the implementation to use `value = response.dict(skip_defaults=skip_defaults)` to take into account those sub-models.

---

And yeah, it's unfortunate to have to do a second parse of the models and have to "clone" models. 

That came after adding Pydantic ORM mode, as now the data is serialized by Pydantic directly (including DB models). And a sub-model including something sensitive (e.g. a `hashed_password`) still passes validation as-is for the parent model. So, a model `class UserWithPassword(BaseUser)` passes validation for a model `class BaseUser(BaseModel)`, including the extra sensitive data.

To avoid that security issue, I had to "clone" the models declared in `response_model`, to make sure they are a "different" class, and sub-classes don't pass directly but have to be parsed and validated.

PR #423, opened 2019-08-06
Update pydantic to 0.32.1

dmontagu:
Among other things, pydantic 0.32 brings some substantial debugging improvements which I would really appreciate! (Namely, fixes the recursion issue in debuggers, and includes model name in validation error tracebacks.)

If merged, [#398 Update pydantic & starlette to include recent bug fixes](https://github.com/tiangolo/fastapi/pull/398) could be closed (as both pydantic and starlette would be on more recent versions).


euri10:
iirc pipfile is not needed but pyproject is

dmontagu:
Oops dumb. Thanks @euri10

fastapi/applications.py
```diff
@@ -138,8 +139,8 @@ def add_api_route(
         deprecated: bool = None,
         methods: List[str] = None,
         operation_id: str = None,
-        response_model_include: Set[str] = None,
-        response_model_exclude: Set[str] = set(),
+        response_model_include: Union[SetIntStr, DictIntStrAny] = None,
+        response_model_exclude: Union[SetIntStr, DictIntStrAny] = set(),
```
dmontagu: This is the new type for pydantic's `BaseModel.dict`'s `include` and `exclude` arguments

fastapi/exceptions.py
```diff
@@ -13,8 +16,10 @@ def __init__(
 
 
 class RequestValidationError(ValidationError):
-    pass
+    def __init__(self, errors: Sequence[ErrorList]) -> None:
+        super().__init__(errors, Request)
```
dmontagu: `pydantic` now requires ValidationError to receive a type to indicate as the type that failed validation. The value is *only* used as a way to obtain a name to put in the exception traceback. Given that `RequestValidationError` is collecting all of the errors from the various dependencies that may succeed or fail, I think just calling it out as the Request is an easy way to pick a type to use here, and it is still actually pretty useful for debugging because it will indicate the problem occurred before your endpoint-specific code had even started.

(This will also make it easier to diagnose problems in submitted issues/gitter/etc., especially if the submitter is light on details.)

fastapi/exceptions.py
```diff
@@ -13,8 +16,10 @@ def __init__(
...
 
 
 class WebSocketRequestValidationError(ValidationError):
-    pass
+    def __init__(self, errors: Sequence[ErrorList]) -> None:
+        super().__init__(errors, WebSocket)
```
dmontagu: Same thing applies here as applies to Request above.

tests/test_tutorial/test_path_params/test_tutorial005.py
```diff
@@ -106,8 +106,9 @@ def test_openapi():
             {
                 "detail": [
                     {
+                        "ctx": {"enum_values": ["alexnet", "resnet", "lenet"]},
                         "loc": ["path", "model_name"],
-                        "msg": "value is not a valid enumeration member",
+                        "msg": "value is not a valid enumeration member; permitted: 'alexnet', 'resnet', 'lenet'",
                         "type": "type_error.enum",
```
dmontagu: Enum error was recently updated to include a little more context

tests/test_tutorial/test_handling_errors/test_tutorial004.py
```diff
@@ -81,7 +81,7 @@ def test_get_validation_error():
     response = client.get("/items/foo")
     assert response.status_code == 400
     validation_error_str_lines = [
-        b"1 validation error",
+        b"1 validation error for Request",
```
dmontagu: Here's how the type submitted to ValidationError is now used

fastapi/exceptions.py
dmontagu: Another thing we could do would be to create new (empty) types called `RequestDependencies` and `WebsocketDependencies` to pass to `ValidationError` to make it even more clear what is happening. (But I'm also fine keeping it like this.)

tiangolo:
Excellent job @dmontagu ! :open_mouth: :rocket: :cake: 

Thanks for all the comments and clarifications. And thanks for taking the time to update the docs.

And thanks @euri10 for pointing out the version change.

I will try to first merge other PRs to have those bug fixes included in versions compatible with previous Pydantic versions and then I'll come back to merge this.

pyproject.toml
```diff
@@ -20,7 +20,7 @@ classifiers = [
 ]
 requires = [
     "starlette >=0.11.1,<=0.12.7",
-    "pydantic >=0.30,<=0.30.0"
+    "pydantic >=0.32.1,<=0.32.1"
```
dhirschfeld: Is *exact* pinning really required? It can create significant problems with tools such as conda which enforce version constraints.
```suggestion
    "pydantic >=0.32.1,<0.33"
```

skewty:
I believe the pydantic group is trying hard to get 1.0 released and just released 0.32.2.  Perhaps this should be updated to use 0.32.2

dmontagu:
@tiangolo Let me know if you'd like me to update this to 0.32.2; if you would be able to merge that in the near future, I will happily perform any necessary tweaks (still eager to get the improvements from pydantic 0.31 and 0.32 integrated with my fastapi codebases!).

Otherwise, given how close pydantic 1.0 seems to be, maybe it makes sense to wait until then?

skewty:
pydantic 1.0 is expected to be roughly a month to 1.0 alpha.  I personally think getting this release out is a good idea.  Get some testing in on the existing code changes :).   That'll leave less refactoring to do / troubleshoot should some API breaking changes come in 1.0.

Atheuz:
@tiangolo closed without merging? I was looking forward to getting this merged because I wanted to use the `schema_extra` functionality which is not in the currently pinned version of pydantic.

zamiramir:
Resolved in #463

tiangolo:
Thank you guys! Yes, this is included in #463 and already released. :tada: :rocket:

PR #427, opened 2019-08-12
Add type-checking for dependencies

dmontagu:
This change ensures that the following error is caught by mypy:
```python
def get_type_2() -> Type2:
    return Type2()

@app.get("/")
def endpoint(x: Type1 = Depends(get_type_2)):
    return ""
```
-----
* The type of `Depends(callable)` is the type of the result of `callable`.
* When `Depends()` is used without a callable, the result is treated as Any (since the type is guaranteed to be that of the annotation anyway).
* If you want the result of the `Depends` function to be type-hinted as an actual `Depends` class instance, you can use the `Depends` class directly, rather than the function.

@tiangolo let me know if you'd like me to create an issue for discussion.

### DIFF
--- a/fastapi/param_functions.py
+++ b/fastapi/param_functions.py
@@ -243,12 +243,12 @@ T = TypeVar("T")
 
 @overload
 def Depends(dependency: Callable[..., T], *, use_cache: bool = True) -> T:
-    ...
+    ...  # pragma: no cover
 
 
 @overload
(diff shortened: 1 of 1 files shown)


dmontagu:
I'm going to close this because it doesn't play nice with including `Depends` in the `dependencies=` keyword argument to the endpoint decorators. But I think it would be nice to support this somehow eventually...

PR #435, opened 2019-08-13
Additional responses can be status groups or "default"

### LINKED ISSUE #428: [BUG] OpenAPI: response keys are limited to be status codes
**Describe the bug**
According to OpenAPI 3 specification, under `responses` dictionary each definition starts with the keys which is either a valid status codes (200, 204, 404, etc.), a [range of codes](https://swagger.io/docs/specification/describing-responses#status-codes) (1XX, 2XX, 3XX, etc.) or the word ["default"](https://swagger.io/docs/specification/describing-responses#default).

If a path is defined with such response, the server crashes when opening the doc, redoc or openapi.json URLs.

**To Reproduce**
```python
from fastapi import FastAPI, HTTPException
from starlette import status
from . import schemas

app = FastAPI()

@app.get('/customer/{id}',
            status_code=status.HTTP_200_OK,
            response_model=schemas.CustomerDetailsResponse,
            responses={
                '4xx': {'model': schemas.ErrorResponse},
                '5xx': {'model': schemas.ErrorResponse},
            },
            )
def customer_details(id: str):
    raise HTTPException(status_code=status.HTTP_501_NOT_IMPLEMENTED)
```

**Expected behavior**
The openapi.json is created, documentation opens, and the responses show the range / default 
definitions.

```yaml
# ...
"/api/customer/{id}":
  get:
    responses:
      '200':
        description: Successful Response
        content:
          application/json:
            schema:
              "$ref": "#/components/schemas/CustomerDetailsResponse"
      4xx:
        description: Additional Response
        content:
          application/json:
            schema:
              "$ref": "#/components/schemas/ErrorResponse"
      5xx:
        description: Additional Response
        content:
          application/json:
            schema:
              "$ref": "#/components/schemas/ErrorResponse"
    summary: Customer Details
    operationId: customer_details_api_customer__id__get
    parameters:
    - required: true
      schema:
        title: Id
        type: string
      name: id
      in: path
```

**Environment:**
 - OS: masOS 10.14.6
 - Python: 3.6.5
 - FastAPI: 0.35.0

divums:
Catching exception when generating OpenAPI specification. Added status groups and "default" as possible key for the additional responses.

If merged, this issue fixes #428

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -43,12 +43,12 @@ validation_error_response_definition = {
     },
 }
 
-status_groups: Dict[str, str] = {
-    "1xx": "Information",
-    "2xx": "Success",
-    "3xx": "Redirection",
-    "4xx": "Client Error",
-    "5xx": "Server Error",
+status_code_ranges: Dict[str, str] = {
+    "1XX": "Information",
+    "2XX": "Success",
+    "3XX": "Redirection",
+    "4XX": "Client Error",
+    "5XX": "Server Error",
     "default": "Default Response",
 }
 
@@ -199,17 +199,14 @@ def get_openapi_path(
                         response.setdefault("content", {}).setdefault(
                             "application/json", {}
                         )["schema"] = response_schema
-                    status_text: Optional[str] = status_groups.get(
-                        str(additional_status_code).lower()
-                    )
-                    status_text = status_text or http.client.responses.get(
-                        int(additional_status_code)
-                    )
+                    status_text: Optional[str] = status_code_ranges.get(
+                        str(additional_status_code).upper()
+                    ) or http.client.responses.get(int(additional_status_code))
                     response.setdefault(
                         "description", status_text or "Additional Response"
                     )
                     operation.setdefault("responses", {})[
-                        str(additional_status_code).lower()
+                        str(additional_status_code).upper()
                     ] = response
             status_code = str(route.status_code)
             response_schema = {"type": "string"}
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -43,6 +43,15 @@ validation_error_response_definition = {
     },
 }
 
+status_groups: Dict[str, str] = {
+    "1xx": "Information",
+    "2xx": "Success",
+    "3xx": "Redirection",
+    "4xx": "Client Error",
+    "5xx": "Server Error",
+    "default": "Default Response",
+}
+
 
 def get_openapi_params(dependant: Dependant) -> List[Field]:
     flat_dependant = get_flat_dependant(dependant)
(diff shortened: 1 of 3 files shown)


fastapi/openapi/utils.py
```diff
@@ -190,12 +190,32 @@ def get_openapi_path(
...
                             "application/json", {}
                         )["schema"] = response_schema
-                    status_text = http.client.responses.get(int(additional_status_code))
+                    try:
+                        # check if valid HTTP status code
+                        status_text: Optional[str] = http.client.responses.get(
+                            int(additional_status_code)
+                        )
+                    except ValueError as ex:
+                        status_groups: Dict[str, str] = {
```
dmontagu: Probably better to put the status groups dict at the module level (near `validation_error_definition`)

Then this can be much more concise:
```
status_text: Optional[str] = status_groups.get(str(additional_status_code).lower())
status_text = status_text or http.client.responses.get(int(additional_status_code))
```

Pipfile
```diff
@@ -23,6 +23,9 @@ flake8 = "*"
 python-multipart = "*"
 sqlalchemy = "*"
 uvicorn = "*"
+bcrypt = "*"
+PyJWT = "*"
+passlib = "*"
```
dmontagu: Did you mean to include these?

divums: when installing requirements via pip, those were missing and had to install manually in order to run the tests. i was waiting feedback whether that's correct or just my issue.

fastapi/openapi/utils.py
divums: Indeed it reads more clearly than the try-except block

Pipfile
dmontagu: They are only dev requirements, which is why I think they weren't included.

divums: reverted this change. i suppose if nobody needed them so far, they were only issue because of the pip install i did

tests/test_additional_responses_bad.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+from fastapi import FastAPI
+from starlette.testclient import TestClient
+
+app = FastAPI()
+
+
+@app.get("/a", responses={"hello": {"description": "Not a valid additional response"}})
+async def a():
+    return "a"
```
dmontagu:
```suggestion
    pass  # pragma: no cover
```
I think this will make the test coverage pass; this is how it's done elsewhere where the goal is not to test the actual endpoint.

divums: i was in the middle of checking why the coverage dropped but haven't managed to reach any conclusion yet. thanks a lot, this comment saved a lot of time for me.

Pipfile
tiangolo: Yeah, I'm not completely happy with Pipenv. It's OK like this for now.

I might switch the environment handling to pure `python -m venv`, Poetry or something. But that will be later.

tiangolo:
Awesome, thanks @divums ! Good job! :rocket: :cake: 

I updated it a bit, to use uppercase instead of lowercase, to conform with the spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#patterned-fields-1

And I augmented the tests to include some more of these cases.

And thanks a lot @dmontagu for the code review and help here! :clap: :muscle: :cake:

tiangolo:
I realized I introduced a bug making it `"DEFAULT"` instead of `"default"`, while making the status code ranges uppercase.

But I just fixed it in #489 :bug: 

I also added extra tests there to include `"default"` and other status codes at the same time and discovered another bug in the models. But it's fixed now. :sweat_smile: :rocket:

PR #438, opened 2019-08-14
Article: Deploying a scikit-learn model with ONNX and FastAPI added

naxty:


tiangolo:
Great, thanks! :rocket: :memo:

PR #447, opened 2019-08-19
Fix typo in oauth2-jwt.md

pablogamboa:


tiangolo:
Good catch, thanks! :pencil2: :tada:

PR #451, opened 2019-08-21
Add support for strings and __future__ annotations

dmontagu:
This pull request adds support for forward references, string annotations, and `from __future__ import annotations` in decorated endpoint signatures.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -355,10 +355,10 @@ async def solve_dependencies(
     errors: List[ErrorWrapper] = []
     response = response or Response(
         content=None,
-        status_code=None,
+        status_code=None,  # type: ignore
         headers=None,
         media_type=None,
-        background=None,  # type: ignore
+        background=None,
     )
     dependency_cache = dependency_cache or {}
     sub_dependant: Dependant
@@ -433,7 +433,7 @@ async def solve_dependencies(
     values.update(cookie_values)
     errors += path_errors + query_errors + header_errors + cookie_errors
     if dependant.body_params:
-        body_values, body_errors = await request_body_to_args(  # type: ignore # body_params checked above
+        body_values, body_errors = await request_body_to_args(  # body_params checked above
             required_params=dependant.body_params, received_body=body
         )
         values.update(body_values)
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -353,8 +353,12 @@ async def solve_dependencies(
 ]:
     values: Dict[str, Any] = {}
     errors: List[ErrorWrapper] = []
-    response = response or Response(  # type: ignore
-        content=None, status_code=None, headers=None, media_type=None, background=None
+    response = response or Response(
+        content=None,
+        status_code=None,
+        headers=None,
+        media_type=None,
+        background=None,  # type: ignore
     )
     dependency_cache = dependency_cache or {}
     sub_dependant: Dependant
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -26,7 +26,7 @@ from pydantic.error_wrappers import ErrorWrapper
 from pydantic.errors import MissingError
 from pydantic.fields import Field, Required, Shape
 from pydantic.schema import get_annotation_from_schema
-from pydantic.utils import lenient_issubclass
+from pydantic.utils import ForwardRef, evaluate_forwardref, lenient_issubclass
 from starlette.background import BackgroundTasks
 from starlette.concurrency import run_in_threadpool
 from starlette.datastructures import FormData, Headers, QueryParams, UploadFile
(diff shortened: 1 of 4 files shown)


fastapi/dependencies/utils.py
```diff
@@ -158,6 +160,36 @@ def is_scalar_sequence_field(field: Field) -> bool:
...
+        # elif hasattr(type(call), "__call__"):
+        #     type_hints = get_type_hints(type(call).__call__)
+        # else:
+        #     raise ValueError("Could not determine callable call")
+    # elif hasattr(call, "__call__"):
+    else:
+        type_hints = get_type_hints(call.__call__)  # type: ignore
+    # else:
+    #     raise ValueError("Could not determine callable call")
+    typed_params = []
```
dmontagu: I commented out a variety of edge-case checks here to ensure code coverage would pass, but before merging this code should be uncommented with appropriate tests added.

tiangolo:
Nice! Good job!

> This PR would introduce a small breaking change [...]

No problem with that. We bump the version to release it.

> Before this is ready to merge [...]

Perfect. Let me know when you can check those things.

dmontagu:
@tiangolo Thanks for the feedback!

I actually discussed this with Ivan Levkivskyi on the python/typing gitter, and he suggested I make a bpo post (which I did here: https://bugs.python.org/issue37923). So far it hasn't gotten much attention 😕 but I figured I'd at least ask the question.

Having dug a little more into the various related PEPs and python stdlib source, I feel pretty good about the approach in this PR, so I'm inclined to run with it, and just add more edge-case handling and tests for them. Hopefully in the future there will be first class support for this built directly into the `inspect` or `typing` packages, and we can remove any custom/redundant logic.

I'll try to get this into a mergeable state this weekend.

tests/test_security_oauth2.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
dmontagu: This file looked like a good example of complex dependency usage, and allowed me to get 100% coverage for tests.

I also wanted to start by making a minimal set of changes to make the pull request easier to review; simply adding `from __future__ import annotations` seemed like an easier way to prove it works than creating an entirely new test file to review.

@tiangolo How would you propose we handle testing this functionality?
* Should I copy this file (or another) and just add the `from __future__ import annotations` import?
* Should I create an entirely new file with a more comprehensive set of string-annotation tests?
    * In this case, is there a good place for me to look for examples?

fastapi/dependencies/utils.py
```diff
@@ -171,6 +171,30 @@ def is_scalar_sequence_field(field: Field) -> bool:
     return False
 
 
+def get_typed_signature(call: Callable) -> inspect.Signature:
```
dmontagu: I originally wanted to use `typing.get_type_hints` directly, but ultimately this implementation seemed cleaner and follows `typing.get_type_hints` where it matters. It also uses `pydantic.utils` to handle differences between the `typing` api for python 3.6 and 3.7 (for `ForwardRef` and `evaluate_forwardref` anyway), and to deal with the fact that the relevant objects are technically private in `typing`. Since `pydantic` has made them public, it seems reasonable to let `pydantic` worry about ensuring the public API is stable.

dmontagu:
@tiangolo I reworked the implementation, and now it should be 100% backward compatible (no changes to the annotations in the case of optionals, etc.), and only change any of the behavior when string-valued annotations are present.

dmontagu:
Note: the currently-failing test is due to the `from __future__ import annotations` line I added; I'll remove it shortly.

tests/test_security_oauth2.py
```diff
@@ -21,18 +21,18 @@ class User(BaseModel):
     username: str
 
 
-def get_current_user(oauth_header: str = Security(reusable_oauth2)):
+def get_current_user(oauth_header: "str" = Security(reusable_oauth2)):
```
dmontagu: It might be better to make this change in another file; @tiangolo if you give me some direction I can implement the tests however you'd prefer.

tiangolo (changes requested):
Awesome job! :tada: :rocket: 

I think it's fine to leave the tests with your modifications and test both things at the same time. I just requested comments on top of them to remind us to *not* "refactor" them into their non-string version.

dmontagu:
@tiangolo

* I made the requested changes.
* I also updated the `# type: ignore` comments to:
    1. be compatible with the recent mypy 0.730 release (which is automatically used in the build)
    2. remove unused ignores (detected by calling `mypy fastapi --disallow-untyped-defs --warn-unused-ignore`)
* Do you think the docs should be updated anywhere? It seems like release notes is probably enough

fastapi/utils.py
```diff
@@ -58,10 +58,10 @@ def create_cloned_field(field: Field) -> Field:
     use_type = original_type
     if lenient_issubclass(original_type, BaseModel):
         original_type = cast(Type[BaseModel], original_type)
-        use_type = create_model(  # type: ignore
+        use_type = create_model(
             original_type.__name__,
             __config__=original_type.__config__,
-            __validators__=original_type.__validators__,
+            __validators__=original_type.__validators__,  # type: ignore
```
dmontagu: New mypy is smarter about the where the # type: ignore should go

fastapi/openapi/models.py
```diff
@@ -11,7 +11,7 @@
     import email_validator
 
     assert email_validator  # make autoflake ignore the unused import
-    from pydantic.types import EmailStr  # type: ignore
+    from pydantic.types import EmailStr
```
dmontagu: This `# type: ignore` is not necessary for mypy to pass; let me know if you want this in a separate PR

fastapi/dependencies/utils.py
```diff
@@ -405,7 +433,7 @@ def is_coroutine_callable(call: Callable) -> bool:
     values.update(cookie_values)
     errors += path_errors + query_errors + header_errors + cookie_errors
     if dependant.body_params:
-        body_values, body_errors = await request_body_to_args(  # type: ignore # body_params checked above
```
dmontagu: This `# type: ignore` is not necessary for mypy to pass; let me know if you want this in a separate PR

fastapi/dependencies/utils.py
```diff
@@ -329,8 +353,12 @@ def is_coroutine_callable(call: Callable) -> bool:
 ]:
     values: Dict[str, Any] = {}
     errors: List[ErrorWrapper] = []
-    response = response or Response(  # type: ignore
-        content=None, status_code=None, headers=None, media_type=None, background=None
+    response = response or Response(
+        content=None,
+        status_code=None,  # type: ignore
```
dmontagu: Needed to put the `# type: ignore` here to get mypy to pass (sadly this forces the line to become longer, but has the benefit of *only* applying to that one argument.)

tiangolo (approved):
Awesome! Thanks a lot!

LGTM :tada: :rocket:

PR #467, opened 2019-08-26
make it possible to change the default reponse_class

toppk:
now you can do app = FastAPI(default_response_class=UJSONResponse)

and all routes will use UJSONResponse by default.

You can still override response_class in routes as before.

the default is still kept as JSONResponse, just specified in much fewer places (two)

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -212,11 +212,12 @@ def get_openapi_path(
                 "description"
             ] = route.response_description
             assert (
-                route.response_class
-            ), "A response class is needed to generate OpenAPI"
+                route.response_class and route.response_class.media_type
+            ), "A response class with media_type is needed to generate OpenAPI"
+            media_type: str = route.response_class.media_type
             operation.setdefault("responses", {}).setdefault(
                 status_code, {}
-            ).setdefault("content", {}).setdefault(route.response_class.media_type, {})[
+            ).setdefault("content", {}).setdefault(media_type, {})[
                 "schema"
             ] = response_schema
 
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -211,6 +211,9 @@ def get_openapi_path(
             operation.setdefault("responses", {}).setdefault(status_code, {})[
                 "description"
             ] = route.response_description
+            assert (
+                route.response_class
+            ), "A response class is needed to generate OpenAPI"
             operation.setdefault("responses", {}).setdefault(
                 status_code, {}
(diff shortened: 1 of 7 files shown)


dmontagu:
@tiangolo I noticed your git commit message "Fix mypy error in Travis not shown locally..." -- this is due to the recent update to mypy. If you run `pip install mypy --upgrade` you'll get the same output locally. (Just had to make that change on another PR myself 😄 .)

tiangolo:
Thanks @dmontagu ! :sweat_smile:

I did, and I got the other errors you were having, in fact, I think I read your mind and implemented exactly the same fixes here at the same time as you :joy: 

But now I'm getting a really strange `mypy` error in Travis, reported in a line that doesn't even have the reported variable...

I'll rebase this from master to take your last PR and hopefully, that will fix it :rocket:

fastapi/openapi/utils.py
```diff
@@ -211,9 +211,13 @@ def get_openapi_path(
             operation.setdefault("responses", {}).setdefault(status_code, {})[
                 "description"
             ] = route.response_description
+            assert (
+                route.response_class and route.response_class.media_type
+            ), "A response class with media_type is needed to generate OpenAPI"
+            media_type: str = route.response_class.media_type
             operation.setdefault("responses", {}).setdefault(
                 status_code, {}
-            ).setdefault("content", {}).setdefault(route.response_class.media_type, {})[
```
dmontagu: I think you'll be able to fix the mypy issue by using `response_class = route.response_class or route.default_response_class`. Or by setting `self.response_class: Type[Response] = response_class or default_response_class` in the route `__init__`.

dmontagu: Oh nevermind, I see the problem, media_type itself could be None

tiangolo: Thanks! I don't know I wasn't getting the same line locally, nor the error :thinking: 

I added an assertion that requires the `route.response_class` to exist. As these utils should only be called after adding the sub-routers to the app. So, by that point, they all should be some class (and not `None`).

tiangolo:
Thanks @toppk ! :cake: 

I like the idea. In fact, I've been thinking about it for quite some time. :sweat_smile: 

I refactored it a bit to move the `default_response_class` from the `APIRouter` class to the `include_router` method, the same way we do with `dependencies`. And updated the logic with the `None` checks and assignments.

This fixes a future bug where a sub router would not behave as expected.

This also allows adding the same router more than once, with different response classes. :tada: 

I also added lots of tests, to check all the possible combinations of router trees I could imagine.

Issue #200, opened 2019-05-05
Hooks to modify request and response body


nav:
How can I modify request body before it's accessed by the api handler and response body before it's returned by the handler?

Is it possible to implement a middleware or view hooks to change the request body and response body as needed?

**Additional context**
I am working on implementing API versioning with FastAPI using [Stripe's approach](https://stripe.com/blog/api-versioning). As I understand, it requires modifying the request body and response body depending on requested version of the API. I've been experimenting with different approaches but unable to find a clean way to do it. I've tried implementing it as a middleware but it doesn't seem like the right place. So my question is, are there any hooks that I can rely on to modify the incoming request and modify the response. Thanks!

wshayes:
@tiangolo has made it really easy to work with the middleware - it's the way I'd handle this if I was doing it the way you are - of course using the mounts - it's really easy to handle /api/v1/... and /api/v2/... in the same application:

https://fastapi.tiangolo.com/tutorial/middleware/

nav:
Thank you for your input @wshayes. What I am trying to achieve is slightely different but your suggestion definitely pointed me in the right direction.

tiangolo:
Thanks for your help @wshayes !

@nav you might also want to look at `APIRouter`s, with them you can add an API prefix to a whole set of routes/path operations: https://fastapi.tiangolo.com/tutorial/bigger-applications/#include-an-apirouter-with-a-prefix-tags-and-responses

nav:
Thanks for the info @tiangolo but I am trying to avoid having to maintain multiple routes (versioning via routes i.e., /v1, /v2, etc). What I am trying to achieve is the ability to modify body of the request before it's processed by the route handler and response after it has been processed by the same handler. 

So far I haven't had luck in finding an appropriate way to do it. It may be that the answer lies in Starlette since the request and response objects are handled by it. Additionally, request object is a child class of `collections.abc.Mapping` which is immutable so changing the request body might require a layer of abstraction.

tiangolo:
Hmm, by what you describe here in this last comment, it seems that you just need a middleware, as @wshayes was suggesting: https://fastapi.tiangolo.com/tutorial/middleware/

nav:
Thanks @tiangolo, it does seem like middleware is the way to go but now the challenge is modifying `starlette.requests.Request` and `starlette.responses.StreamingResponse` which is what request and response objects are. There seem to be enough documentation on reading the body in both cases but not so much regarding modifying the body.

[...]

PR #468, opened 2019-08-27
Add support for custom route class

dmontagu:
This change would make it substantially easier to override the default route-matching behavior. In particular, this could simplify the implementation of version-based routing as discussed in https://github.com/tiangolo/fastapi/issues/200

I created a pull request for this to aid in discussion since it was only a three line change, but before merging it would probably merit some updates to docs and tests.

In the absence of interest, I'll be happy to just close the pull request unmerged.


tiangolo:
Given all the conversation in #200 this seems a simple way to add a lot of flexibility, while still being backwards compatible. Thanks! :tada: :rocket:

PR #481, opened 2019-08-29
implement toggle to prevent usage of google fonts in redoc

b1-luettje:
Since it is already possible to prevent the inclusion of third-party javascript and images, the should be possible for css/fonts for the redoc page


tiangolo:
Awesome! Thanks! :tada: :cake:

PR #485, opened 2019-08-29
Ensure `skip_defaults` doesn't cause extra fields to be serialized

dmontagu:
Currently, if `skip_defaults` is true, the secure cloned field is not used when serializing the response. This can lead to extra information leaking out if the `response_model` differs in type from the returned model.

This pull request fixes this bug, and updates the relevant unit test to check for it.

(The bug was introduced in #422 -- sorry about that!)

-----

I believe this pull request also improves performance in the case where `skip_defaults` is `True`, as the response will now only be validated once in the `serialize_response` call, instead of twice.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -45,6 +45,8 @@ def serialize_response(
 ) -> Any:
     if field:
         errors = []
+        if skip_defaults and isinstance(response, BaseModel):
+            response = response.dict(skip_defaults=skip_defaults)
         value, errors_ = field.validate(response, {}, loc=("response",))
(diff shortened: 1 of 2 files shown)


tiangolo:
Good catch! :mag: :male_detective: :rocket: 

Thanks a lot! :tada: :cake:

skaaptjop:
Just wonder if line 49 in routing.py shouldn't read:
```
response = response.dict(skip_defaults=skip_defaults, by_alias=by_alias)
```

When using a response model that uses field aliases, the validation step following that conversion always fails.

Issue #237, opened 2019-05-16
[FEATURE] Configurable Swagger UI


zamiramir:
I want to be able to set configurations of Swagger UI, but I can't do it the right way as it is hard coded right now.
[Reference to Swagger UI configuration](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/)

**Describe the solution you'd like**
I would love to be able to set configurations through FastAPI object.

The normal Swagger UI configurations are easy to implement. However there are also some methods like _initOAuth_ which requests some additional checks in code.

My suggesetion would be to replace the current parameter _openapi_url_ of het method _fastapi/openapi/docs.py->get_swagger_ui_html_ with an Pydantic object with the defaults set with the current values for _openapi_url_.

**Describe alternatives you've considered**
Alternative solution is to set docs_url to None and create a get method with a HTMLResponse as return value. The returned HTMLResponse is identical to _fastapi/openapi/docs.py->get_swagger_ui_html_ method, with some additional configurations and the _initOAuth_ method.

**Additional context**
**Why do I need this?** 
With a custom company Identity & Access Management with support for OAuth2 in place, we need some additional configuration in Swagger UI to get it to work. 
One of the items is already in an PR #198. (alternative for this solution is also a get method with HTMLResponse as a return value)

Some other settings we would like to set in Swagger UI are:

- oauth2RedirectUrl setting, this makes it possible to redirect to another page then Swaggers default _http://url/oauth2-redirect.html_.
- The method initOAuth [Reference to Swagger UI OAuth2 configuration](https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/) 

I hope it is clear, if you have any questions let me know.

zamiramir:
[My Fork with possible solution(in progress)](https://github.com/zamiramir/fastapi/tree/configure-swagger-ui)

tiangolo:
Thanks. I'll check it as soon as I get a chance.

PR #499, opened 2019-09-03
Make it possible to add swagger ui's clientId and appName

zamiramir:
Added 2 new parameters to FastAPI init:
- swagger_ui_client_id
- swagger_ui_app_name

Swagger UI has a method initOAuth. We are using to of these: 
- clientId
- appName

[Reference to Swagger UI OAuth2 configuration](https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/)

see #237


alimcmaster1:
Could you use pytest parametrize here to avoid repeating test boilerplate/setup?

tiangolo:
Thanks for your contribution! :tada: :rocket: :cake: 

I updated it a bit to support the full `initOauth` settings in https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/oauth2.md

PR #506, opened 2019-09-04
Remove reparse

dmontagu:
This is a work in progress attempt to remove the use of the secure cloned field.

The remaining issues relate to:
1. Generic field types parameterized by pydantic models (e.g., `List[MyModel]`)
1. Some outstanding issues with pydantic dataclasses

I'm opening this despite the fact that it currently fails to share the implementation idea (and maybe get some feedback).

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -1,9 +1,69 @@
 from enum import Enum
+from functools import partial
 from types import GeneratorType
-from typing import Any, Dict, List, Set, Union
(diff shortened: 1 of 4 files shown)


dmontagu:
I think this is more appropriate as a pydantic feature request combined with a smaller change to fastapi; https://github.com/samuelcolvin/pydantic/issues/811 is the corresponding pydantic feature request. I'll create a new pull request to FastAPI once that pydantic issue is addressed.

PR #517, opened 2019-09-10
Allow to set custom scheme for 4xx responses

tsouvarev:
Code from https://github.com/tiangolo/fastapi/pull/437 checks for custom schemas on response, 
but it missed the fact that all codes are uppercased in https://github.com/tiangolo/fastapi/blob/master/fastapi/openapi/utils.py#L196


tiangolo:
Thanks for your contribution @tsouvarev ! :tada: :cake: :rocket:

PR #530, opened 2019-09-14
Make adding openapi route configurable

alimcmaster1:
This is an example of a [Starlette application](https://gist.github.com/alimcmaster1/c5146abad6a0b1f60ea009b589ec309d)  - the following change makes it possible for users to use a Starlette schema endpoint as a source for the docs.

Making it simple to migrate from Starlette to FastAPI:

`app = Starlette()`

`app = FastAPI(add_openapi_route=False, openapi_url='/schema')`

cc. @tiangolo  to review.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -31,6 +31,7 @@ class FastAPI(Starlette):
         title: str = "Fast API",
         description: str = "",
         version: str = "0.1.0",
+        add_openapi_route: bool = True,
         openapi_url: Optional[str] = "/openapi.json",
         openapi_prefix: str = "",
         docs_url: Optional[str] = "/docs",
@@ -50,6 +51,7 @@ class FastAPI(Starlette):
         self.title = title
         self.description = description
         self.version = version
+        self.add_openapi_route = add_openapi_route
         self.openapi_url = openapi_url
         self.openapi_prefix = openapi_prefix.rstrip("/")
         self.docs_url = docs_url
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks! But I prefer to limit the extra parameters as I think too many can be confusing, even more, when it's specifically for edge cases.

Nevertheless, with your example, you can achieve the same with the current version, with a modification in a couple of lines.

But still, thank you for your contribution and your help here with the community! :rocket: :cake: :tada: 

Here's your modified example:

```Python
from fastapi import FastAPI
from fastapi.openapi.docs import get_swagger_ui_html
import uvicorn
from starlette.applications import Starlette
from starlette.schemas import SchemaGenerator


schemas = SchemaGenerator(
    {"openapi": "3.0.0", "info": {"title": "Example API", "version": "1.0"}}
)
#app = Starlette()
app = FastAPI(openapi_url=None, docs_url=None, redoc_url=None)

@app.route("/users", methods=["GET"])
def list_users(request):
    """
    responses:
      200:
        description: A list of users.
        examples:
          [{"username": "tom"}, {"username": "lucy"}]
    """
    raise NotImplementedError()


@app.route("/schema", methods=["GET"])
def openapi_schema(request):
    return schemas.OpenAPIResponse(request=request)


@app.route("/docs", methods=["GET"])
def get_docs(request):
    return get_swagger_ui_html(
        openapi_url="/schema",
        title="Starlette to FastAPI - Swagger UI"
    )


if __name__ == '__main__':
    uvicorn.run(app, host='0.0.0.0', port=8000)
```

alimcmaster1:
@tiangolo - thanks very much for looking at this, your example was most helpful.

Best,

Ali

Issue #521, opened 2019-09-11
msgpack response if client asks for application/x-msgpack


sandys:
hi,
we are in the process of evaling fastapi for a large app migration from flask. Quick question on a very interesting and useful piece of functionality. 
This builds upon the validations you have created for encoding json data and will be very useful for us out-of-the-box.

https://msgpack.org/index.html is an enhancement to json which is highly compressed. There are some benchmarks that show it is as fast as protocol buffers...while making it a drop-in alternative to json. 

msgpack is very important for Asia which are highly internet bandwidth constrained. We are prepared to tradeoff speed in exchange for lesser bandwidth (e.g. http://indiegamr.com/cut-your-data-exchange-traffic-by-up-to-50-with-one-line-of-code-msgpack-vs-json/ )

To be fair, it may be same size as gzipped json responses...but after having done android app deployments over a million devices in the field in asia, we realized that one has to write custom compression/decompression for gzipped json response handling. However msgpack has fairly well supported libraries. 

msgpack is fairly production ready. Lots of downstream python libraries use it in production - https://github.com/0rpc/zerorpc-python

This enhancement request is for fastapi to respond in msgpack if the requesting client ask for the mimetype `application/x-msgpack`. The default behavior continues to be the same as currently.

There is no impact to any client of a fastapi deployment unless they specifically ask for msgpack (in which case, they will get better performance).

I'm also pasting some criticisms of msgpack and comparisons with competing/derived json formats like CBOR

- https://news.ycombinator.com/item?id=14067747
- http://zderadicka.eu/comparison-of-json-like-serializations-json-vs-ubjson-vs-messagepack-vs-cbor/
- https://techtutorialsx.com/2019/06/20/esp32-arduinojson-messagepack-serialization/
- https://gist.github.com/frsyuki/2908191
- https://prataprc.github.io/msgpack-vs-cbor.html

dmontagu:
@sandys 

I've reworked it to be substantially simpler, and also I've put some thought into what first-class support could/should look like in FastAPI.

-----

Here's a msgpack-specific simplified implementation (with minimal docs) that runs as is to show how simple it is:
(note `include_both_serialization_routes` could be removed with a relatively minor feature request, described below).
<details>
    <summary>Click to expand</summary>

```python
    from typing import Any, List, Tuple

    import msgpack
    from fastapi import FastAPI, APIRouter
    from fastapi.routing import APIRoute
    from starlette.requests import Request
[...]

PR #538, opened 2019-09-17
Preserve route_class when calling include_router

### LINKED ISSUE #539: [BUG] route_class not preserved when calling APIRouter.include_router
**Describe the bug**

When calling `APIRouter.include_router`, the class of the included routes should be preserved. Right now, this is not the case -- the `route_class` of the parent router is used. This means that the `route_class` is lost on inclusion if one isn't careful. Worse, it means that even if you manually create a router with multiple route-types present, there is no way to include that router in another without changing the route type of some of the routes.

See the `include_typed_router` function from #521 for an example of how this currently requires a workaround.

I have addressed this in the short PR #538, but wanted to create an issue for discussion in case there is any reason to preserve the current behavior.

dmontagu:
Ensures that the `route_class` of the subrouter is preserved when including a subrouter in a parent router.

------

Right now, when including a router with a custom `route_class`, the `route_class` is not automatically preserved. This essentially prevents the use of custom `route_class` without special tricks. (For example, see #521, where overriding the `route_class` is used to enable msgpack parsing.)

This PR changes the behavior of `include_router` so that the route class is preserved during inclusion for any routes that are subclasses of `APIRoute`.

-----

Closes #539

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -345,8 +345,10 @@ class APIRouter(routing.Router):
         include_in_schema: bool = True,
         response_class: Type[Response] = None,
         name: str = None,
+        route_class_override: Optional[Type[APIRoute]] = None,
     ) -> None:
-        route = self.route_class(
+        route_class = route_class_override or self.route_class
+        route = route_class(
             path,
             endpoint=endpoint,
             response_model=response_model,
@@ -484,6 +486,7 @@ class APIRouter(routing.Router):
                     include_in_schema=route.include_in_schema,
                     response_class=route.response_class or default_response_class,
                     name=route.name,
+                    route_class_override=type(route),
                 )
             elif isinstance(route, routing.Route):
                 self.add_route(
--- a/tests/test_custom_route_class.py
+++ b/tests/test_custom_route_class.py
@@ -1,0 +1,114 @@
+import pytest
+from fastapi import APIRouter, FastAPI
+from fastapi.routing import APIRoute
+from starlette.testclient import TestClient
+
+app = FastAPI()
+
+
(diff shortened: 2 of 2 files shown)


dmontagu:
@tiangolo I just re-ran the checks for this PR. Note #568 may serve as an alternative, but would require a little more effort. I'm not sure which way (if either) you'd be inclined to go.

tiangolo:
Thanks @dmontagu ! Great! :rocket: :tada: 

I just added a couple of tests to make sure we avoid possible future regressions with this. :white_check_mark:

dmontagu:
Thanks, sorry the PR wasn't more complete!

tiangolo:
No worries at all, you're doing an awesome job helping with FastAPI here. With code and also helping others in issues and in Gitter.

You rock! :guitar: :rocket: :tada: :bowing_man:

ifigueroap:
Hi, I'm wondering on whether it is possible to externally override the `route_class` of an included router. A use case would be to add a logging implementation (via subclassing of `APIRoute`) to existing routers. This is currently possible only if the source is available, and only by doing it before adding the routes, but it appears not to be possible for external modules. The `route_class_override` parameter is also only available at the actual call of `add_api_route`.

The crux of the issue, I think, is that routes are eagerly constructed with the `route_class` constructor in the `add_api_route` method, and then added to the `self.routes` attribute. Hence, overriding the route_class after the fact would perhaps require to reconstruct all the routes... I'm not sure that would be desireable nor possible... 

Still, I think the use case for 'dynamic route_class overriding' is there...

Please let me know if you think this makes sense for the project, and in that case whether you would like me to submit a more refined issue.

Thank you for this great project!

PR #547, opened 2019-09-19
fix bug with 4xx check

DeanWay:
The status_code_key is set to uppercase above, leading to this check always being false.
This invalidated the expected behaviour of a specified 4xx or 4XX overriding the default 422 response

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -43,12 +43,17 @@ validation_error_response_definition = {
     },
 }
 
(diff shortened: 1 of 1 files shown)


DeanWay:
Just noticed #517. Either works, only difference here is moving the erroneous check away from using a string literal

tiangolo:
Thanks @DeanWay ! I just merged #517 , so, this should be solved in the next release.

But thanks for your contribution! :cake:

PR #554, opened 2019-09-22
Alphabetically sort schemas

dmontagu:
Modify openapi spec generation to include schemas in alphabetical order.

-----

Right now, the schemas occur in a haphazard order in the generated OpenAPI spec (and therefore also at the bottom of the swagger docs), since it is generated based on the order in which schemas are encountered during the spec generation process.

This change would cause the definitions to occur in alphabetical order.

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -283,7 +283,7 @@ def get_openapi(
                 if path_definitions:
                     definitions.update(path_definitions)
     if definitions:
-        components.setdefault("schemas", {}).update(definitions)
+        components["schemas"] = {k: definitions[k] for k in sorted(definitions)}
     if components:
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks @dmontagu ! :cake: :rocket: 

I haven't needed it, but I see how it could be useful. Thanks! :tada:

dmontagu:
I have found this useful for two things:

1. In the swagger docs, sometimes I just want to scroll down to the schema definition. It's almost impossible when they are out of order!

2. I want my schema diffs to be as small as possible (so I can put them in version control and have it be useful for tracking changes). There are workarounds (like sorting the json keys), but they mostly have tradeoffs I don't like (the resulting schema is in a very unorthodox order).

Thanks for merging this!

PR #556, opened 2019-09-22
Allow api docs to be truncated and used for other purposes

svalouch:
This is essentually the same Click does:
https://click.palletsprojects.com/en/7.x/documentation/#truncating-help-texts

It allows splitting the documenting text so that the lower part may be
used to describe the function further, e.g. extracting information using
Sphinx.


fastapi/routing.py
```diff
@@ -249,6 +249,10 @@ def __init__(
             self.dependencies = []
         self.summary = summary
         self.description = description or inspect.cleandoc(self.endpoint.__doc__ or "")
+        # if a form feed (page break) is found in the description text, truncate
+        # description text to the content preceding the first form feed
+        if self.description and "\f" in self.description:
```
dmontagu: No need to check for `"\f" in self.description`; the first result of `split` will just be the whole string in that case

dmontagu:
This seems like a nice, simple addition. Using `"\f"` as the splitting pattern wouldn't have been my first choice, but given other well-established tools use it for this purpose it seems reasonable enough to me. I can't imagine it getting in the way.

trustchk:
Okay, simplified the splitting, thanks for pointing it out.
We could use some other character to split it, but it would require checking other tools that are known to work (read: ignore) the character used. ``\f`` is working with Sphinx, at least, I haven't checked other tools.

svalouch:
I simplified the splitting, thanks for pointing it out.
If we used other characters than ``\f``, we need to test commonly used tools for their compatibility, that is, if they just ignore it. Sphinx is known to work with it just fine.

tests/test_split_docstrings.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@app.get("/split")
+def doc_with_linefeed():
+    """
+    Documentation that is broken appart
+    \f
+    by a line feed. This part should not appear in the description.
+    """
+    return {"id": "foo"}
```
dmontagu: I think the build will pass if you remove this return statement, and the one from the other endpoint above.

(FastAPI has a 100% coverage requirement for the *test code* as well.)

dmontagu:
Oh also, the formatting is wrong -- run `./scripts/format-imports.sh` -- it should automatically fix it for you.

svalouch:
Okay, sorry done. Sorry for the noise. I didn't know you could define functions without a body, but the test suite confirms it. I lean something new every day :)

dmontagu:
The reason it works is because it has a docstring - so technically it does have a body :)

tiangolo:
Thank you for your contribution @svalouch ! :tada: :rocket: :cake: 

I updated it a bit and moved the docs to the section on advanced path operation configurations.

Thanks @dmontagu for the help and comments! :bowing_man:

PR #557, opened 2019-09-22
Add documentation for static swagger (#112)

svalouch:
This adds an example for overwriting the location for the JS and CSS
files for the docs and redoc endpoints.

### DIFF
--- a/docs/src/extending_openapi/tutorial002.py
+++ b/docs/src/extending_openapi/tutorial002.py
@@ -1,0 +1,43 @@
+from fastapi import FastAPI
+from fastapi.openapi.docs import (
+    get_redoc_html,
+    get_swagger_ui_html,
+    get_swagger_ui_oauth2_redirect_html
+)
+from starlette.requests import Request
+from starlette.responses import HTMLResponse
+
+
+app = FastAPI(docs_url=None, redoc_url=None)
+
+
+async def swagger_ui_html(req: Request) -> HTMLResponse:
+    return get_swagger_ui_html(
+        openapi_url=app.openapi_url,
+        title=app.title + ' - Swagger UI',
+        oauth2_redirect_url=app.swagger_ui_oauth2_redirect_url,
+        swagger_js_url='/static/swagger-ui-bundle.js',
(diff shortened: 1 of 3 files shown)


svalouch:
Okay, coverage is good now. I'm using the documented method in my projects, but I'm not entirely sure if it's the easiest way to achieve the goal.

tiangolo:
Great! Thanks. :rocket: 

I updated it a bit with more instructions so that people can use it and test it right away.

youngquan:
I have followed the steps in [docs](https://fastapi.tiangolo.com/advanced/extending-openapi/#self-hosting-javascript-and-css-for-docs), and it really worked.

However, it's not very friendly to the novice. **I wonder whether these steps could be down with `pip install fastapi[all]` or something else?** How about an paramenter named offline like `Fastapi(offline=True)`.

There is no internet where I am working, I tried to use fastapi and have found that the docs  were not rendering. I had to import these JSs and CSSs extra. I hope fastapi could be more useful in an offline scenario.

EdgeSync:
> I have followed the steps in [docs](https://fastapi.tiangolo.com/advanced/extending-openapi/#self-hosting-javascript-and-css-for-docs), and it really worked.
> 
> However, it's not very friendly to the novice. **I wonder whether these steps could be down with `pip install fastapi[all]` or something else?** How about an paramenter named offline like `Fastapi(offline=True)`.
> 
> There is no internet where I am working, I tried to use fastapi and have found that the docs were not rendering. I had to import these JSs and CSSs extra. I hope fastapi could be more useful in an offline scenario.

Agree completely with this. Would be very useful to have the Fastapi(offline=True) option.

tiangolo:
@youngquan , @EdgeSync that would mean that FastAPI would have to bundle all the JS and static files increasing a lot its size, that would also increase the complexity maintaining it as it would have to get a recent version of those assets at build time, etc.

youngquan:
@tiangolo Thanks for you time. A third-party package is a good choice, I will try to porvide a one soon or later if I have time. ✋😂

Issue #558, opened 2019-09-22
HHTP request/response with application-content = text/xml


agorina:
How can I support a POST request with XML as a body and XML as a response.

I was able to get that far:
```
mport typing

from fastapi import APIRouter
from simplexml import dumps
from starlette.responses import Response

router = APIRouter()


class XmlResponse(Response):
    media_type = "text/xml"

    def render(self, content: typing.Any) -> bytes:
        return dumps({'response': content}).encode("utf-8")


@router.get("/items/scorer")
async def get_response():
    """An endpoint to return the global configuration
    """
    return XmlResponse({'person':{'name':'joaquim','age':15,'cars':[{'id':1},{'id':2}]}})
```

And this returns an XML. However, the `docs` and `redoc` API are not returning anything. 
I am seeing this in the logs:
```
[7697] [2019-09-22 13:37:06,432] [uvicorn] [DEBUG] [('127.0.0.1', 59158) - ASGI [4] Received {'type': 'http.response.start', 'status': 404, 'headers': '<...>'}]
[7697] [2019-09-22 13:37:06,432] [__main__] [INFO] [('127.0.0.1', 59158) - "GET /socs HTTP/1.1" 404]
````

Any help is very much appreciated!

**Comment by dmontagu** [comment]
> `GET /socs HTTP/1.1" 404`

Looks like it might be a typo in your browser? Like, you tried to go  to `apihost.com/socs` instead of `apihost.com/docs`.

**Comment by agorina** [comment]
Sorry about the typo. You are right about 404 error, however the page `http://127.0.0.1:8080/docs` comes black empty.

And I am not able to PST the XML in a body:
```
{
    "detail": "There was an error parsing the body"
}
```
I am guessing I need to configure something to be able to accept the body in XML, just fail to find any documentations. 

A sample on how to do it will be very much appreciated.

**Comment by dmontagu** [comment]
1. What do you mean it comes back empty? Are you getting a specific status code? What do the server logs say?

2. Currently, I don't think FastAPI supports XML as a body directly; you'd have to create a dependency that reads the object off from the raw XML content in the request body.

    Can you share the code that you are using for your endpoint that accepts an XML body? If you can do that, it might be draft a skeleton implementation that you might find helpful.

**Comment by agorina** [comment]
1. The logs include this (does not look helpful to me):
```
[16130] [2019-09-22 15:56:32,548] [__main__] [DEBUG] [('127.0.0.1', 61157) - Disconnected]
[...]

PR #568, opened 2019-09-25
Refactor APIRoute for easier overrides

dmontagu:
This pull request refactors the `APIRoute` class for two main purposes:

1. It makes it easier to override individual pieces in subclasses (e.g., `get_app`), and
2. It makes it so that `APIRouter.include_router` does not require re-instantiating `APIRoute`s, so that subclasses can be preserved.

This pull request would eliminate the need for #538 and would help address #521 and #558.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1,7 +1,19 @@
 import asyncio
 import inspect
 import logging
-from typing import Any, Callable, Dict, List, Optional, Sequence, Set, Type, Union
+from copy import deepcopy
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Pattern,
+    Sequence,
+    Set,
+    Type,
+    Union,
+)
 
 from fastapi import params
 from fastapi.dependencies.models import Dependant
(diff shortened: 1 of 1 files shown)


dmontagu:
(To be more explicit, this pull request also modifies the `APIRouter.include_router` so that it doesn't re-instantiate `APIRoute` instances.)

fastapi/routing.py
```diff
@@ -178,6 +191,23 @@ def __init__(
 
 
 class APIRoute(routing.Route):
+    body_field: Optional[Field]
```
dmontagu: I've added these as class annotations so that PyCharm doesn't complain about missing fields, since these fields are now defined in the setup functions, rather than the `__init__`. I would be happy to go with a different convention though (e.g. setting defaults in the `__init__` prior to calling the setup functions).

dmontagu:
Hmm.. It doesn't like the deepcopy for python 3.6.

I'm going to wait for positive feedback on this approach before I try to put in additional work to make it work with 3.6.

dmontagu:
@tiangolo I would be happy to reopen this pull request if it interests you, but I think for now I'm going to close it and move a little bit more incrementally toward making it easier to override some of this behavior. (That way it will also be easier to add tests demonstrating the value of the functionality.)

PR #570, opened 2019-09-26
Add support for contextmanager-style dependencies

dmontagu:
This pull request adds support for contextmanager style dependencies. 

In particular, dependencies of the following form would be supported:

```python
def get_dependency():
    value = do_setup()
    yield value  # the value received by the endpoint
    do_cleanup()  # handled in a background task after returning the response
```

Async contextmanager dependencies are also supported. (See the tests for examples of each, along with a demonstration of error handling.)

I've opened the pull request as a draft because I still need to add tutorial examples and/or other docs, but I am not currently aware of any remaining issues in the implementation.

(Please let me know if you see any.)

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -410,6 +410,14 @@ async def solve_dependencies(
             continue
         if sub_dependant.use_cache and sub_dependant.cache_key in dependency_cache:
             solved = dependency_cache[sub_dependant.cache_key]
+        elif inspect.isgeneratorfunction(call):
+            if background_tasks is None:
+                background_tasks = BackgroundTasks()
+            solved = await solve_generator(call, sub_values, background_tasks)
+        elif inspect.isasyncgenfunction(call):
+            if background_tasks is None:
+                background_tasks = BackgroundTasks()
+            solved = await solve_asyncgen(call, sub_values, background_tasks)
         elif is_coroutine_callable(call):
             solved = await call(**sub_values)
         else:
@@ -456,6 +464,49 @@ async def solve_dependencies(
             scopes=dependant.security_scopes
         )
     return values, errors, background_tasks, response, dependency_cache
+
+
+async def solve_generator(
+    call: Callable, sub_values: Dict[str, Any], background_tasks: BackgroundTasks
+) -> Any:
+    generator = call(**sub_values)
+
+    def solve() -> Any:
+        try:
+            return next(generator)
+        except StopIteration:
+            raise RuntimeError("generator didn't yield")
+
+    def teardown() -> None:
+        try:
+            next(generator)
+        except StopIteration:  # pragma: no cover
+            pass  # prevents BackgroundTasks from hanging
+        else:
+            raise RuntimeError("generator didn't stop")
+
+    background_tasks.add_task(teardown)
+    return await run_in_threadpool(solve)
+
+
+async def solve_asyncgen(
+    call: Callable, sub_values: Dict[str, Any], background_tasks: BackgroundTasks
+) -> Any:
+    asyncgen = call(**sub_values)
+
+    async def teardown() -> None:
+        try:
+            await asyncgen.__anext__()
+        except StopAsyncIteration:
+            return
+        else:
+            raise RuntimeError("generator didn't stop")
+
+    background_tasks.add_task(teardown)
+    try:
+        return await asyncgen.__anext__()
+    except StopAsyncIteration:
+        raise RuntimeError("generator didn't yield")
 
 
 def request_params_to_args(
--- a/docs/src/dependencies/tutorial008.py
+++ b/docs/src/dependencies/tutorial008.py
@@ -1,9 +1,4 @@
-from fastapi import Depends, FastAPI
-from sqlalchemy.orm import Session
-
 from sql_databases.sql_app.database import SessionLocal
-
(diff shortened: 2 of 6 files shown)


dmontagu:
Note this would make it possible to stop using middleware for the ubiquitous `get_db` dependency:
```python
@app.middleware("http")
async def db_session_middleware(request: Request, call_next):
    response = Response("Internal server error", status_code=500)
    try:
        request.state.db = SessionLocal()
        response = await call_next(request)
    finally:
        request.state.db.close()
    return response


# Dependency
def get_db(request: Request):
    return request.state.db
```
could become
```python
def get_db() -> Session:
    db = SessionLocal()
    try:
        yield db
    except Exception as exc:
        db.rollback()  # not in the original, but maybe still useful
        raise exc
    finally:
        db.close()
```

In particular, this would let you drop the database dependency on routes that don't need it. (In my experience, middleware like this is one of the things that makes it painful to reuse code across projects.)

euri10:
That's indeed super nice, will try to make use of it in my little celery ws
stuff not that I think about it

Le ven. 27 sept. 2019 à 8:14 AM, dmontagu <notifications@github.com> a
écrit :

fastapi/dependencies/utils.py
```diff
@@ -427,6 +435,49 @@ def is_coroutine_callable(call: Callable) -> bool:
...
+        except StopIteration:
+            raise RuntimeError("generator didn't yield")
+
+    def teardown() -> None:
+        try:
+            next(generator)
+        except StopIteration:  # pragma: no cover
+            pass  # prevents BackgroundTasks from hanging
+        else:
+            raise RuntimeError("generator didn't stop")
```
dmontagu: Note: this is precisely the same error message you'd get if `call` had been used in a `with` statement after decorating with `contextlib.contextmanager`.

dmontagu: The same is the case for the `RuntimeError` above and those in the `solve_asyncgen` call below.

dmontagu:
@tiangolo This now has a first "serious" attempt at adding docs for this feature. Let me know if you have any preferences for changes to their structure.

cjw296:
This looks interesting, although I can't follow quite how the code actually works, likely due to my shallow knowledge of async stuff. Some questions:

- How does this interact with the caching of resolved depedencies?

- How do I make sure the order of teardown is correct? For example, if I have a transaction as a dependency, and that transaction must be finished before the database connection is torn down, how do I express that requirement?

- How is error handling supported? A common pattern for me is semantically expressed like this:

   ```
   conn = connect_to_db()
   t = conn.transaction()
   try:
        ... do work ...
       t.commit()
   except:
       t.rollback()
    conn.close()
   ```

If `conn` and `t` were two context manager dependencies, how could I ensure the behaviour as a above?

tiangolo:
Thanks @dmontagu ! This is pretty smart! :clap: :rocket: 

I got inspired by this and started to check different cases. Sadly, if an exception occurs inside of a path operation or a dependency that requires the context dependency, the `except` and `finally` blocks of the context dependency are not executed. So, it ends up not behaving like a context manager with respect to handling internal exceptions.

---

But then I realized that we can achieve it in a different way, I'm currently sketching a PR for it.

I needed to have the most recent interface of Starlette (ASGI 3), so I merged that first. But with that I'm pretty sure that we'll be able to support it.

The only drawback is that I'll have to add 2 dependencies (backports) to support Python 3.6:

https://github.com/python-trio/async_generator and https://github.com/sorcio/async_exit_stack

The other option would be to only support context manager dependencies in Python 3.7.

dmontagu:
@tiangolo Good catch. The BackgroundTask approach definitely felt a little bit fragile, so I'm not surprised that there were some edge cases not being handled properly.

I think it would make sense to only support this with python 3.7. I also don't see much downside to adding optional dependencies to support it with python 3.6. But I use 3.7 everywhere now (and will hopefully be able to change that to 3.8 ASAP) so I'm fine either way.

-----

@cjw296 As implemented, it would require some care to ensure teardown was done in a guaranteed order. I think it would currently amount to constructing a consistent "teardown" dependency. But I think in most cases, you would be able to just build a single dependency that sets everything up in order, and tears it down in order, and then to build other dependencies on top of that. For example, handling your use case:

```python
def db_contextmanager():
    conn = connect_to_db()
    t = conn.transaction()
    try:
         yield conn, t
        t.commit()
    except:
        t.rollback()
     conn.close()

def get_connection(conn_t = Depends(db_contextmanager)):
    conn, t = conn_t
    return conn

def get_transaction(conn_t = Depends(db_contextmanager)):
    conn, t = conn_t
    return t

def endpoint(conn = Depends(get_connection)):
    # Do something with the connection
    pass
```

I think there is room for improvement here though.

The cleanest logic I've been able to think of for how to have contextmanager dependencies tear themselves down in a controllable order would be to merge the individual teardown tasks into a single "solved" teardown task by basically walking the reversed dependency DAG (restricted to contextmanager dependencies).

The idea being that if you have contextmanager dependencies d1 and d2, and d2 depends on d1, then the cleanup section of d2 would be guaranteed to execute prior to the cleanup section of d1. That maybe doesn't cover all theoretical use cases, but I think it would cover 99+% of practical uses and you could always work around the other cases by just writing a more complex contextmanager dependency.

dmontagu:
Closing in favor of #595

PR #574, opened 2019-09-28
add info block related to security tutorial about python-multipart

sliptonic:


tiangolo:
Thanks for your contribution! Good idea! :tada: :cake: 

I updated it a bit to include the note in several places, at the first points in the docs where a user would need to install it.

Issue #575, opened 2019-09-30
Catch exception type of Exception and process body request


Hedgehogues:
**Description**

How can I catch exception type of Exception and process body request?

**Additional context**

My example

@app.exception_handler(Exception)
async def exception_callback(request: requests.Request, exc: Exception):
    request_json = await request.json() # Here i want a body, query_params, headers and path of request
    print(request_json)

@app.post('/create_workflow')
async def create_workflow():
    raise Exception(13)

euri10:
mmm I'm not sure, but I don't think you can consume body in a middleware (see https://github.com/tiangolo/fastapi/issues/394)

getting the other parameters you mentioned is possible though (note that the `Request` below is from `starlette` and not `requests` I think you imported the wrong one), but should you need the body that won't happen

I'd ask in Starlette if I were you

```
import pytest
from async_asgi_testclient import TestClient
from fastapi import FastAPI
from starlette.requests import Request

app = FastAPI()


@app.exception_handler(Exception)
async def exception_callback(request: Request, exc: Exception):
    print(request.query_params)
    print(request.headers)
    print(request.path_params)
    print(exc.args)


@app.post("/create_workflow")
async def create_workflow():
    raise Exception(13)


@pytest.mark.asyncio
async def test_raise_exc():
    async with TestClient(app) as client:
        with pytest.raises(Exception) as pe:
            response = await client.post("/create_workflow")
```

Hedgehogues:
Yes, I saw it. But this code cannot extract body of request. Example of printing:


    ```
print(request.query_params)
    print(request.headers)
    print(request.path_params)
    print(exc.args)
```

euri10:
yes as I pointed out in the above linked issue #394, this is not something you can achieve in a middleware.
you might want to read https://github.com/encode/starlette/issues/495#issuecomment-494008175 as well and ask on Starlette

tiangolo:
A new way to handle these cases was added in #589.

The docs are here: https://fastapi.tiangolo.com/tutorial/custom-request-and-route/

Issue #394, opened 2019-07-18
Awaiting request body in middleware blocks the application


wyfo:
**Describe the bug**
Description in the title

**To Reproduce**
Minimal code:
```
from typing import Mapping

from fastapi import FastAPI
from starlette.requests import Request

app = FastAPI()

@app.middleware("http")
async def func(request: Request, call_next):
    print(await request.json())
    return await call_next(request)

@app.post("/")
def read_root(arg: Mapping[str, str]):
    return {"Hello": "World"}
```
Run the application with ```uvicorn <file>:app```
Test the bug with ```curl localhost:8000 -d '{"status":"ok"}'```

**Expected behavior**
The body of the request is printed, but the curl command stay pending for ever. If it is interrupted (Ctrl-C), the application then print ```ERROR: Error getting request body:```

**Environment:**
 - OS: macOS
 - fastapi 0.33.0
 - python 3.7.3
 - (tested on Ubuntu too with Python 3.7.0 and 3.7.4)

**Additional context**
 - When the route function has no body argument (```def read_root():```), there is no problem : the body is printed and the response send.
 - Thinking the issue was maybe coming from Starlette, I tested the following code, which works without issue. The bug seems thus to come from fastapi
```
from starlette.applications import Starlette
from starlette.requests import Request
from starlette.responses import JSONResponse

app = Starlette()


@app.middleware("http")
async def func(request: Request, call_next):
    print(await request.json())
    return await call_next(request)


@app.route('/', methods=["POST"])
def homepage(request):
    return JSONResponse({"Hello": "World"})
```

euri10:
works fine for me copy-pasting the exact same thing, only had to comment `    print(await request.json())
` which crashed the app

wyfo:
> works fine for me copy-pasting the exact same thing, only had to comment ` print(await request.json())` which crashed the app

That's exactly the point, the print statement, or more exactly the ``àwait``` inside blocks the application. The question is why ?
And the other question, more important for me is : how can I log my request payload if i'm not able to await it in middleware ?

euri10:
I think the middleware is unnecessary, should you want to get your payload just do:
```
@app.post("/bug")
async def read_root(request: Request, arg: Mapping[str, str]):
    print(await request.json())
    return {"Hello": "World"}
```

wyfo:
> I think the middleware is unnecessary, should you want to get your payload just do:
[...]

Issue #375, opened 2019-07-06
Access request body when reusing FastAPI's exception handlers


stewit:
**Description**

First let me say: FastAPI is exceptionally useful to me and my team. You are doing great work here! Thank you very much.

Now to my question: **How can I access the request body when writing my own exception handlers?**

I understand that I get the request object of starlette as an argument to my own exception handler. However to access its body the coroutine request.body() must be executed (or is there any other way?). When I try to call this coroutine like in this example:

```
@app.exception_handler(RequestValidationError)
async def validation_exception_handler(request, exc):
    request_body = await request.body()
    logger.info(
        "Request validation failed:\n" + str(exc) + "\nRequest:\n" + str(request_body)
    )
    return await request_validation_exception_handler(request, exc)
```
the application hangs indefinitely / the event loop somehow freezes.

I tried to achieve the same using FastAPI's middleware decorator but that leads to the same problem.

Is it possible to to access the request body in my own exception handler? If yes, what is the recommended way?

lvalladares:
Hey! Its a problem with starlette, check this: https://github.com/encode/starlette/issues/495#issuecomment-513138055 there is a workaround

stewit:
Thanks to the work done in https://github.com/tiangolo/fastapi/pull/589 and the new documentation at https://fastapi.tiangolo.com/tutorial/custom-request-and-route/ I close this question.

tiangolo:
Thanks for coming back to close the issue and for adding the link to the new docs @stewit :clap: :cake:

Issue #180, opened 2019-04-23
[Bug/Question] Response from the server is never received in case of using middleware.


ansrivas:
**Describe the bug**
Response from the server is never received in case of using middleware.

**To Reproduce**

`$ cat main.py`
```python
# !/usr/bin/env python
# -*- coding: utf-8 -*-
"""Entry point of the application."""


import logging

from fastapi import Depends, FastAPI, HTTPException
from pydantic import BaseModel
from starlette.middleware.base import BaseHTTPMiddleware
from starlette.requests import Request
from starlette.responses import Response

logger = logging.getLogger(__name__)


class AuthMiddleware(BaseHTTPMiddleware):
    async def dispatch(self, request, call_next):
        response = Response("Faled to authenticate", status_code=401)
        try:
            json = await request.json()
            request.state.is_authenticated = True
            logger.info(f"request.body: {json} ")
            response = await call_next(request)
        finally:
            request.state.is_authenticated = False
        return response


app = FastAPI(title="Nice Server",
              description="This is a very fancy project, with auto docs for the API and everything",
              version="0.1.0",
              docs_url="/docs", redoc_url=None)

app.add_middleware(AuthMiddleware)


class SomeTimestamp(BaseModel):
    ts: int


async def is_authenticated(request: Request):
    return request.state.is_authenticated


@app.post("/latest", response_model=SomeTimestamp, summary="Returning the latest ")
async def latest(*, some_timestamp: SomeTimestamp, is_authenticated: bool = Depends(is_authenticated)):
    logger.info(f"Let's see if this is authenticated: {is_authenticated}")
    if not is_authenticated:
        raise HTTPException(status_code=401, detail="Failed to verify the incoming token")

    logger.info(f"Received for prediction {some_timestamp.ts}")
    return some_timestamp.ts
```

Now run the server:
```
uvicorn main_server:app --reload
```

Now query this using `curl`:
```
curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"ts":"123123"}' \
  http://127.0.0.1:8000/latest
```

**Expected behavior**
The server is stuck before sending response.

**Screenshots**

**Environment:**
 - OS: [ Linux]
 - FastAPI Version [e.g. 0.18.0], get it with:

```Python
import fastapi
print(fastapi.__version__)

0.18.0
```

- Python version, get it with:

```bash
python --version
3.7
```

**Additional context**

ansrivas:
[...]

PR #589, opened 2019-10-04
Add better support for request body access/manipulation

dmontagu:
This pull request introduces a small change to `APIRoute` to make it easier to construct a custom `app` attribute in a custom subclass.

There is a lot of custom functionality that this change would simplify, but there are two especially important capabilities it adds:
* A way to get handle exceptions with access to the request body
* A way to modify the request body prior to its parsing/handling by FastAPI

I've added two sections to the "Using the Request Directly" docs showing how to use the changes in this PR to implement both of the above kinds of feature.

------

There have been many recent github issues that could be better-addressed with this capability.

**Related to accessing the body in middleware/exception handlers:**
* #180 accessing request body in middleware
* #375 accessing request body in exception handlers
* #394 accessing request body in middleware
* #575 accessing request body in exception handlers
* https://github.com/encode/starlette/issues/493 accessing request body in exception handlers (using fastapi)
* https://github.com/encode/starlette/issues/495  a variety of use cases involving accessing request body in middleware (some using fastapi)

**Related to modifying the request prior to handling by FastAPI:**
* #521 support for msgpack-encoded requests
* https://github.com/encode/starlette/issues/644 support for gzip-encoded requests (using fastapi)

-----
Quoting Tom Christie from https://github.com/encode/starlette/issues/495:

> "Consume request body in middleware is problematic"
>
> Indeed. Consuming request data in middleware is problematic.
> Not just to Starlette, but generally, everywhere.
>
> On the whole you should avoid doing so if at all possible.

I think using middleware to consume the request body is the obvious (or, perhaps, "naive") way to handle these issues. But that approach is explicitly discouraged, and not supported. So I think the approach described in / supported by this PR is especially valuable.

-----

(This pull request is in lieu of https://github.com/tiangolo/fastapi/pull/568, and includes what I think is both the smallest and most important change to `APIRoute` from that pull request.)


docs/tutorial/using-request-directly.md
```diff
@@ -53,3 +53,63 @@ Then declare a *path operation function* parameter with the type being the `Requ
...
+Next, we create a custom subclass of `fastapi.routing.APIRoute` that will make use of the `GzipRequest`:
+
+```Python hl_lines="20"
+{!./src/using_request_directly/tutorial002.py!}
+```
+
+!!! info
+    The `get_app` method uses all of the information stored on the `APIRoute` to *build* a callable that accepts a `Request`, and returns a `Response`.
+
+    The name doesn't refer to your *whole app*, it just reflects the additional pre- and post-processing logic beyond the endpoint call.
```
dmontagu: Not sure if I'm describing this correctly 😅 just wanted to flag that.

Also I'm not sure if the new `APIRoute.get_app` should be renamed. Since it just calls the `get_app` function that seemed like a good name to me though.

tiangolo: Good point!

The name `get_app` is probably confusing for others.

The original name was based on how it was used in Starlette, which in turn was based on the ASGI spec (technically, a request handler is an ASGI app itself, that could stand by itself).

---

But it could clearly be confusing. I think we should rename it.

How about something like:

*  `request_function`
* `get_request_function`
* `request_handler`
* `get_request_handler`
* `get_handler`

What do you think?

Also, I think we should name the stand-alone function one way and the class method something different, to avoid the editor linking the function from inside the method to the method itself (instead of the function that lives outside).

dmontagu: I'm in favor of `get_request_handler` for the function and `get_route_handler` for the method. (As discussed on gitter.)

tiangolo:
Awesome @dmontagu ! Thanks a lot! :tada: :rocket: :cake: 

I think this is sufficiently important to have its own section in the docs, so I moved it. I put it near the end as I consider this quite advanced and it could confuse newcomers.

I also added an extra example setting the `route_class` in a router. For that, I had to rebase from `master`, to include your recent fix :sweat_smile: 

Thanks again! :bowing_man: :taco:

PR #595, opened 2019-10-06
Dependencies with yield (used as context managers)

tiangolo:
Add support for context manager dependencies using functions with `yield`. Sync or async.

With inspiration from @dmontagu 's https://github.com/tiangolo/fastapi/pull/570.

This handles dependencies that `yield` by converting them into a sync or async context manager,  depending on the function. And runs them in a "context manager stack".

This way, the order of dependencies is preserved and cleanup/closing happens in the correct order (the first dependency called is cleaned last).

This also means that dependencies that require other dependencies will still have them on scope during cleanup. So, contextmanager-style dependencies that require other contextmanager dependencies will have access to them during cleanup.

When a dependency function is a normal function (not `async`), the context manager is called in a threadpool (`__enter__`), so, the same as with normal dependencies, it's safe to do non-async blocking IO in contextmanager style dependencies. E.g. start a DB connection, probably the main use-case for this. And the same for cleaning it up (`__exit__`).

This works by default in Python 3.7.

And works in Python 3.6 if the backports are installed:

* `async-exit-stack`
* `async-generator`

```bash
pip install async-exit-stack async-generator
```

But they are not forced/required dependencies. If an app doesn't use context managers and still runs in Python 3.6, it will still work without the backports.

...Although I suspect most of the apps will use this for DB connections.

These backport dependencies are checked on app initialization, so, if a contextmanager-style dependency function (with `yield`s) is found, the environment is not Python 3.7+, and the backports are not installed, it will raise a `RuntimeError` right away on app initialization (so, no risk for the error lurking into a prod environment).

---

TODO:

* [x] Fix `mypy` types.
* [x] Extend tests to check for correct order of contextmanager dependencies with subdependencies.
* [x] Extend tests to verify that background tasks can still access the dependencies from context managers (e.g. DB connections).
* [x] Add docs.

### DIFF
--- a/docs/tutorial/dependencies/dependencies-with-yield.md
+++ b/docs/tutorial/dependencies/dependencies-with-yield.md
@@ -1,0 +1,153 @@
+# Dependencies with `yield`
+
+FastAPI supports dependencies that do some <abbr title='sometimes also called "exit", "teardown", "close", "context managers"...'>extra steps after finishing</abbr>.
+
+To do this, use `yield` instead of `return`, and write the extra steps after.
+
+!!! tip
+    Make sure to use `yield` one single time.
+
+!!! info
+    For this to work, you need to use **Python 3.7** or above, or in **Python 3.6**, install the "backports":
+
+    ```bash
+    pip install async-exit-stack async-generator
+    ```
+
+    This installs <a href="https://github.com/sorcio/async_exit_stack" target="_blank">async-exit-stack</a> and <a href="https://github.com/python-trio/async_generator" target="_blank">async-generator</a>.
+
+!!! note "Technical Details"
+    Any function that is valid to use with:
+
+    * <a href="https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager" target="_blank">`@contextlib.contextmanager`</a> or 
+    * <a href="https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager" target="_blank">`@contextlib.asynccontextmanager`</a>
+
+    would be valid to use as a **FastAPI** dependency.
+
+    In fact, FastAPI uses those two decorators internally.
+
+## A database dependency with `yield`
+
+For example, you could use this to create a database session and close it after finishing.
+
+Only the code prior to and including the `yield` statement is executed before sending a response:
+
+```Python hl_lines="2 3 4"
+{!./src/dependencies/tutorial007.py!}
+```
+
+The yielded value is what is injected into *path operations* and other dependencies:
+
+```Python hl_lines="4"
+{!./src/dependencies/tutorial007.py!}
+```
+
+The code following the `yield` statement is executed after the response has been delivered:
+
+```Python hl_lines="5 6"
+{!./src/dependencies/tutorial007.py!}
+```
+
+!!! tip
+    You can use `async` or normal functions.
+
+    **FastAPI** will do the right thing with each, the same as with normal dependencies.
+
+## A dependency with `yield` and `try`
+
+If you use a `try` block in a dependency with `yield`, you'll receive any exception that was thrown when using the dependency.
+
+For example, if some code at some point in the middle, in another dependency or in a *path operation*, made a database transaction "rollback" or create any other error, you will receive the exception in your dependency.
+
+So, you can look for that specific dependency with `except SomeException`.
+
+The same way, you can use `finally` to make sure the exit steps are executed, no matter if there was an exception or not.
+
+```Python hl_lines="3 5"
+{!./src/dependencies/tutorial007.py!}
+```
+
+## Sub-dependencies with `yield`
+
+You can have sub-dependencies and "trees" of sub-dependencies of any size and shape, and any or all of them can use `yield`.
+
+**FastAPI** will make sure that the "exit code" in each dependency with `yield` is run in the correct order.
+
+For example, `dependency_c` can have a dependency on `dependency_b`, and `dependency_b` on `dependency_a`:
+
+```Python hl_lines="4 12 20"
+{!./src/dependencies/tutorial008.py!}
+```
+
+And all of them can use `yield`.
+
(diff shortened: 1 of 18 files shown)


dmontagu:
This is looking really great!

fastapi/concurrency.py
```diff
@@ -0,0 +1,44 @@
+from typing import Any
+
+from starlette.concurrency import iterate_in_threadpool, run_in_threadpool  # noqa
+
+asynccontextmanager_error_message = """
+FastAPI contextmanagers with require Python 3.7 or the backport installed with:
```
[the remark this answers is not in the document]
tiangolo: Good catch! Thanks!

tiangolo:
@dmontagu this is ready. :rocket: :tada:

I used some sections from your original PR :sweat_smile: 

Do you wanna give it a last check before I merge?

docs/tutorial/sql-databases.md
```diff
@@ -615,3 +605,51 @@ It will look like this:
...
+!!! info
+    We put the creation of the `SessionLocal()` and handling of the requests in a `try` block.
+
+    And then we close it in the `finally` block.
+    
+    This way we make sure the database session is always closed after the request. Even if there was an exception while processing the request.
+
+### About `request.state`
+
+<a href="https://www.starlette.io/requests/#other-state" target="_blank">`request.state` is a property of each Starlette `Request` object</a>, it is there to store arbitrary objects attached to the request itself, like the database session in this case.
```
dmontagu:
```suggestion
<a href="https://www.starlette.io/requests/#other-state" target="_blank">`request.state` is a property of each Starlette `Request` object</a>. It is there to store arbitrary objects attached to the request itself, like the database session in this case.
```

docs/tutorial/sql-databases.md
```diff
@@ -615,3 +605,51 @@ It will look like this:
...
+
+For us in this case, it helps us ensuring a single database session is used through all the request, and then closed afterwards (in the middleware).
```
dmontagu:
```suggestion
For us in this case, it helps us ensure a single database session is used through all the request, and then closed afterwards (in the middleware).
```

docs/tutorial/sql-databases.md
```diff
@@ -615,3 +605,51 @@ It will look like this:
...
+* The middleware has to be an `async` function.
+    * If there is code in it that has to "wait" for the network, it could "block" your application there and degrade performance a bit.
+    * Although it's probably not very problematic here with the way `SQLAlchemy` works.
+    * But if you added more code to the middleware that had a lot of <abbr title="input and output">I/O</abbr> waiting, it could then be problematic.
+* A middleware is run for *every* request.
+    * So, a connection will be created for every request.
+    * Even when the *path operation* that handles that request didn't need the DB.
+
+!!! tip
+    It's probably better to use dependencies with `yield` when they are enough for the use case.
```
dmontagu: 👍👍

fastapi/dependencies/utils.py
```diff
@@ -410,6 +441,15 @@ def is_coroutine_callable(call: Callable) -> bool:
             continue
         if sub_dependant.use_cache and sub_dependant.cache_key in dependency_cache:
             solved = dependency_cache[sub_dependant.cache_key]
+        elif inspect.isgeneratorfunction(call) or inspect.isasyncgenfunction(call):
```
dmontagu: If I recall correctly, I noticed that this code gets called at request time for each request. But it seems to me like it should (theoretically) be possible to perform this logic when the dependency is *created*, which could remove a substantial number of request-time checks.

I definitely don't think this is worth addressing now, and it may be hard/impossible to address in a way that is compatible with useful features like the dependency-overrides provider, but it might be worth looking into at some point.

dmontagu (commented):
Mostly just minor docs fixes. Some are typos, some are more opinion-based; feel free to ignore anything you disagree with!

docs/tutorial/sql-databases.md
```diff
@@ -615,3 +605,51 @@ It will look like this:
 <img src="/img/tutorial/sql-databases/image02.png">
 
 You can also use an online SQLite browser like <a href="https://inloop.github.io/sqlite-viewer/" target="_blank">SQLite Viewer</a> or <a href="https://extendsclass.com/sqlite-browser.html" target="_blank">ExtendsClass</a>.
+
+## Alternative DB session with middleware
+
+If you can't use dependencies with `yield`, for example, if you are not using **Python 3.7** and can't install the "backports" mentioned above for **Python 3.6**, you can set up the session in a "middleware" in a similar way.
+
+A "middleware" is a function that is always executed for each request, and have code before and after the request.
```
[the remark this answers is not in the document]
dmontagu: I noticed the `and have`, which I think should be `and has`, but it still sounded a little awkward to me which is why I rephrased.

docs/tutorial/dependencies/dependencies-with-yield.md
```diff
@@ -0,0 +1,153 @@
...
+
+    **FastAPI** will do the right thing with each, the same as with normal dependencies.
+
+## A dependency with `yield` and `try`
+
+If you use a `try` block in a dependency with `yield`, you'll receive any exception that was thrown when using the dependency.
+
+For example, if some code at some point in the middle, in another dependency or in a *path operation*, made a database transaction "rollback" or create any other error, you will receive the exception in your dependency.
+
+So, you can look for that specific dependency with `except SomeException`.
```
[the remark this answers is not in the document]
tiangolo: Oh! :see_no_evil: Good catch! :mag: :heavy_check_mark:

fastapi/dependencies/utils.py
tiangolo: Agreed! With both comments :smile:

tiangolo:
Thanks a lot for the thorough review @dmontagu ! You have an eagle eye! :eagle: :mag: :male_detective:

PR #605, opened 2019-10-08
Replace `pip` with `python3 -m pip`

nstapelbroek:
Heya,

If I understood https://github.com/tiangolo/fastapi/issues/531 correctly this PR should take care of replacing every occurence of `pip` with `python3 -m pip`. 

Let me know if you have any feedback. 

Thanks!

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -90,13 +90,13 @@ FastAPI stands on the shoulders of giants:
 ## Installation
 
 ```bash
-$ pip install fastapi
+$ python3 -m pip install fastapi
 ```
 
 You will also need an ASGI server, for production such as <a href="http://www.uvicorn.org" target="_blank">Uvicorn</a> or <a href="https://gitlab.com/pgjones/hypercorn" target="_blank">Hypercorn</a>.
 
 ```bash
-$ pip install uvicorn
+$ python3 -m pip install uvicorn
 ```
 
 ## Example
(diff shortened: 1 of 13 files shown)


docs/index.md
```diff
@@ -90,13 +90,13 @@ FastAPI stands on the shoulders of giants:
 ## Installation
 
 ```bash
-$ pip install fastapi
```
nstapelbroek: Hey 🤔  What is you thoughts on using `$`? It seems different compared to the rest of the docs.

tiangolo:
Thanks for your contribution! :rocket: 

The preference of `pip install`, `python -m pip install`, `python3 -m pip install`, `pip3 install`, is quite subjective and varies a lot depending on personal taste.

So, I prefer to leave the simplest version (`pip install`), and let everyone chose from that what to use.

That's the same reason why I don't suggest Pipenv, Poetry, `venv`, etc.

---

On the other side, the `$ <something>` are quite inconsistent with the rest of the docs.

I should remove those... If you wanna do it, that would be great :smile: :sparkles:

nstapelbroek:
Alrighty 👍  Thank you for your feedback. I'll create a seperate PR to get rid of the `$`. Feel free to close this if you prefer not to have these changes :)

sandys:
here's an argument in favor - https://snarky.ca/why-you-should-use-python-m-pip/

tiangolo:
Thanks @sandys , yes, I saw the article by Brett Cannon.

It almost convinced me, but doing `python -m something` removes any possibility for shell autocompletion, so I'm not sure I want to use it here in the docs.

For now, let's leave it as is.

Thanks @nstapelbroek !

Issue #497, opened 2019-09-03
422 response still exists when all paramas are optional


trim21:
**Describe the bug**
When all params are optional, and all params are `str`, is it possible to reture a `422` status code in this router?

**To Reproduce**
```python
from fastapi.applications import FastAPI
from fastapi import APIRouter
app = FastAPI(docs_url="/")
router = APIRouter()


@router.get('/router')
async def handler(code: str = ''):
    print(code)
    return "hello world"


app.include_router(router)

if __name__ == '__main__':
    import uvicorn
    uvicorn.run(app, port=8001, reload=True)

```
There is a 422 response code in openapi documents

**Expected behavior**
No `422 Validation Error` in this condition

**Screenshots**

**Environment:**
 - OS: Windows
 - FastAPI Version 0.38.1
- Python version 3.7.4

nstapelbroek:
I've created a potential fix in #606 . Let me know what you think 😄

Trim21:
@nstapelbroek How about optional body? 
I didn't try your PR yet, just a thought.

nstapelbroek:
Thanks for checking the PR @Trim21 😄.
I'll see if I can find a better way to check with the additional cases you have provided.

tiangolo:
Thanks @nstapelbroek for your help with this!

@Trim21 the required code to check for when *not* to add the schema to OpenAPI is non-trivial, and I don't really see how it could be a problem to have it.

But still, you can customize OpenAPI and remove or add anything you need, here are the docs for that: https://fastapi.tiangolo.com/tutorial/extending-openapi/

Trim21:
@tiangolo itis very hard to edit a single responses of a path in final json object.

I'm considering if it's possible to set a `422: ...` to disable 422 responses at rouer definition?

dmontagu:
@Trim21 It's not hard to update the openapi schema programatically. It is cached, so you just have to make changes to the result of `app.openapi()`.

See below for a quickly thrown together implementation that would allow you to remove the 422 response just by decorating the endpoint function.

If you have some logic to check whether it should be done automatically, you could just apply that function in the `remove_422s` function and drop the decorator.

But the logic for that is complex enough (and the capability infrequently useful enough) that I don't think it's worth maintaining as a built in part of FastAPI.

<details>
    <summary>Click to expand</summary>

```python
import json
from typing import Any, Callable, Set, TypeVar

from fastapi import FastAPI
from fastapi.openapi.utils import generate_operation_id
[...]

PR #606, opened 2019-10-08
Prevent adding HTTP 422 to the responses when all params are optional

nstapelbroek:
Heya 👋,

After reading https://github.com/tiangolo/fastapi/issues/497 I thought I'd give it a try. This PR will prevent HTTP 422 responses from being added to the operation when it only has optional arguments. It should fix the bug that occurs in the code of #497 : 

```
from fastapi.applications import FastAPI
from fastapi import APIRouter
app = FastAPI(docs_url="/")
router = APIRouter()


@router.get('/router')
async def handler(code: str = ''):
    print(code)
    return "hello world"


app.include_router(router)

if __name__ == '__main__':
    import uvicorn
    uvicorn.run(app, port=8001, reload=True)
```

### How to test:
1. Setup environment with the example code above
1. Acknowledge that the swagger ui shows a HTTP 422 as possible response code
1. Checkout PR
1. Acknowledge that the swagger ui no longer shows HTTP 422 as a possible response code.
1. Change your route params to have a required route
1. Acknowledge that the HTTP 422 is now back 🎉

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -222,7 +222,8 @@ def get_openapi_path(
             ] = response_schema
 
             http422 = str(HTTP_422_UNPROCESSABLE_ENTITY)
-            if (all_route_params or route.body_field) and not any(
+            required_route_params = [route_param for route_param in all_route_params if route_param.required]
+            if (required_route_params or route.body_field) and not any(
                 [
                     status in operation["responses"]
                     for status in [http422, "4XX", "default"]


nstapelbroek (commented):
Still needs some work 🤔  see https://github.com/tiangolo/fastapi/issues/497#issuecomment-539659250

tiangolo:
Thanks!

I actually think that all the extra "effort" in code to check when to remove the 422s with all the extra edge cases is not worth it.

Unless we find an issue/use case that is clearly having problems because of this and can't be handled in any other way.

nstapelbroek:
Alright, I get your point. If it's not really a problem feel free to close the PR and associated issue 😄.

Thanks for checking!

tiangolo:
Thanks for understanding, and thanks for contributing! :rocket: :cake:

PR #613, opened 2019-10-10
Remove $ sign from bash codeblocs in markdown

nstapelbroek:
👋 ,

Applied the changes requested in #605


prostomarkeloff:
Why?

tiangolo:
Thanks @nstapelbroek ! :cake: 

@prostomarkeloff just to make it consistent with the rest of the docs. It was the only place with `$` for Bash prompts.

PR #621, opened 2019-10-14
Fix #259: generate OpenAPI docs for responses with no content.

### LINKED ISSUE #259: [FEATURE] 204 No Content support for OpenAPI schema
I would like to be able to specify in the endpoint decorator that the endpoint will have no response (204 No Content). This is documented in the [swagger.io specification](https://swagger.io/docs/specification/describing-responses/) under the heading Empty Response Body.

This would allow my `openapi-generator`-generated client to detect that there will not be a response, and will automatically make use of an appropriate response (non-)decoder. 

It's not a terribly hard issue to work around manually by tweaking my generated client, but it would simplify things if I could generate an OpenAPI schema without the content field for the 204-response endpoints.

So far, my favorite version of an end-user-facing API would involve introducing a new response class (perhaps called `NoContentResponse`) to pass in the endpoint decorator. 

(But I'm not sure how that would interact with `starlette` or if detecting the response_class is even feasible when generating the OpenAPI spec.)

brotskydotcom:
This is a backwards-compatible change, as far as I know, for responses that explicitly return content.  There are two basic changes:

* FastAPI now knows which response codes do not return bodies, and suppresses the `content` element in their API docs.
* If the response_class for a route has no media type, FastAPI assumes that it will not return content.  (If this route specifies additional responses that do return content, those responses are assumed to return `application/json` content and use an appropriate response class.)

This change introduces two new test files which check the new logic and keep the code coverage at 100%.

This change also updates the tutorial documentation to reflect the new functionality.

### DIFF
--- a/docs/tutorial/additional-responses.md
+++ b/docs/tutorial/additional-responses.md
@@ -5,10 +5,9 @@
 
 You can declare additional responses, with additional status codes, media types, descriptions, etc.
 
-Those additional responses will be included in the OpenAPI schema, so they will also appear in the API docs. 
+Those additional responses will be included in the OpenAPI schema, so they will also appear in the API docs.
 
 But for those additional responses you have to make sure you return a `Response` like `JSONResponse` directly, with your status code and content.
-
 
 ## Additional Response with `model`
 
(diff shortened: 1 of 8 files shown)


brotskydotcom:
FWIW, I ran the black reformatter on all the newly contributed code, to make sure it's stylistically correct.

brotskydotcom:
@tiangolo Any chance of a review of this change?  I'd love to get it to where you're willing to merge it.

tiangolo:
Great! :rocket: :taco: 

And thanks for putting the effort into documenting it. :cake:

PR #637, opened 2019-10-21
Fixing validator-caused incorrect output key order

### LINKED ISSUE #630: JSON output reordered by pydantic @validator
This isn't a big problem, but I'm wondering if it's possible to keep the JSON output key ordering unaffected by the presence of a Pydantic `@validator(...)`? I like consistency. 🤷‍♂️

I have this pydantic model structure:
```
class IngredientBase(BaseIngredientMixin):
    name: str = None
    unit: str = None
    unit_price: decimal.Decimal = None
    flavour_warning: bool = None

    @validator('unit')
    def valid_units_only(cls, v):
        choices = set(SingleIngredientTable.unit_choices.keys())
        if v and v not in choices:
            raise ValueError(f'must be one of {choices}')
        return v


class IngredientIn(IngredientBase):
    pass


class IngredientNew(IngredientIn):
    name: str = ...
    unit: str = ...
    unit_price: decimal.Decimal = ...


class IngredientOut(IngredientBase, TestModelMixin, TimestampModelMixin, IDModelMixin):
    pass
```

This will produce:
```
{
  "unit": "IU",
  "id": "sing_qOYSyqZhfOcJKHsfVz7tnfP6M",
  "created": 1571357369,
  "updated": 1571418480,
  "is_test": false,
  "name": "Vitamin C",
  "unit_price": 10.568536363534536,
  "flavour_warning": false
}
```
Without the validator, it produces what I would expect:
```
{
  "id": "sing_qOYSyqZhfOcJKHsfVz7tnfP6M",
  "created": 1571357369,
  "updated": 1571418480,
  "is_test": false,
  "name": "Vitamin C",
  "unit": "IU",
  "unit_price": 10.568536363534536,
  "flavour_warning": false
}
```

How can I continue to use `@validator`s and maintain the expected JSON key ordering?

**Edit**: I can pull out the validator into a 'mixin' class that I apply ONLY to the `IngredientIn` and `IngredientNew` models (leaving `IngredientOut` untouched). But is that my only path here?

jaddison:
A quick PR that fixes #630. The cloned field was adding existing field validators before the fields themselves. 

This caused Python's default "ordered `dict`" behaviour to incorrectly output dictionary key/value pairs with all 'validated' fields first.

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -61,10 +61,11 @@ def create_cloned_field(field: Field) -> Field:
         use_type = create_model(
             original_type.__name__,
             __config__=original_type.__config__,
-            __validators__=original_type.__validators__,  # type: ignore
         )
         for f in original_type.__fields__.values():
             use_type.__fields__[f.name] = f
+        for k, f in original_type.__validators__.items():
+            use_type.__validators__[k] = f
     new_field = Field(
         name=field.name,
         type_=use_type,
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -61,10 +61,11 @@ def create_cloned_field(field: Field) -> Field:
         use_type = create_model(
             original_type.__name__,
             __config__=original_type.__config__,
-            __validators__=original_type.__validators__,  # type: ignore
         )
         for f in original_type.__fields__.values():
             use_type.__fields__[f.name] = f
+        for k, v in original_type.__validators__.items():
+            use_type.__validators__[k] = v
     new_field = Field(
         name=field.name,
         type_=use_type,


dmontagu:
Have you confirmed that this doesn't change the validation logic? Probably worth making sure validators still work on endpoints with a response model (maybe this is already tested in the existing tests).

Pydantic has some metaprogramming magic that could mean the validators don't get set up properly if they aren't present during class initialization; I'm not sure.

jaddison:
>Have you confirmed that this doesn't change the validation logic? Probably worth making sure validators still work on endpoints with a response model (maybe this is already tested in the existing tests).

Yes, all of the endpoints in my projects have `response_model` parameters set, and all are functioning as expected. That said, I assume that package tests would find any bugs introduced in this regard? 🤔

fastapi/utils.py
```diff
@@ -61,10 +61,11 @@ def create_cloned_field(field: Field) -> Field:
         use_type = create_model(
             original_type.__name__,
             __config__=original_type.__config__,
-            __validators__=original_type.__validators__,  # type: ignore
         )
         for f in original_type.__fields__.values():
```
dmontagu: I wonder if you can't just do `use_type.__fields__ = original_type.__fields__` and be done with it?

jaddison: I don't believe create_model has that signature. I played around with some options, but nothing stuck.

However, feel free to discard my PR if there is a better solution, by all means!

dmontagu: No I meant after the create_model call, instead of looping. Just replacing the whole dict instead; the call to create_model could stay as it was. I think that might pass mypy too.

jaddison: Updated, as suggested.

dmontagu:
Looks good to me!

tiangolo:
Great! Thank you @jaddison! :rocket: :cake: 

And thanks @dmontagu for the PR review, that's highly appreciated :muscle: :taco:

Issue #442, opened 2019-08-17
Using openapi-generator to generate clients results in redundant and repetitive function call names


christopherthompson81:
**Describe the bug**
Using openapi-generator to generate clients from FastAPI projects results in redundant and repetitive function call names

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a REST API with FastAPI
2. Install openapi-generator (https://github.com/OpenAPITools/openapi-generator)
3. Generate a client (I generated a javascript client, for example)
4. Open the documentation and look at the bizarre function calls like: "deleteClaimClaimsClaimIdDelete(claimId)" for a DELETE method against /Claims/{claim_id}

**Expected behaviour**
The function names should not be repetitive and redundant, and should only be based on the action. For example, the DELETE method against /Claims/{claim_id} should follow the Python function name "delete_claim", and therefore show up as "deleteClaim" when converted to camelCase by openapi-generator.

**Environment:**
 - FastAPI Version "==0.33.0"
- Python version "3.7"

**Recommended Fix**
Generation of the "operationId" when creating the "openapi.json" file should be strictly the function name.

dmontagu:
@christopherthompson81 
I think having this as the default behavior is reasonable because it ensures that there are *never* conflicts in the names that don't stem from conflicts in routes, and that the defaults still include the function name.

However, this is very easy to override manually. I use the following function to clean up the function names in my generated clients (and just take the responsibility on myself to avoid naming conflicts):
```python
def setup_openapi(app: FastAPI) -> None:
    """ Simplify operation IDs so that generated clients have simpler api function names """
    for route in app.routes:
        if isinstance(route, APIRoute):
            route.operation_id = route.name
```

christopherthompson81:
That does work. Thank-you! I will note that the necessary import is:

```python
from fastapi.routing import APIRoute
```

and that the function to redefine the operation_id should be called after setting up the routes.

I also think this might be important information to include in the tutorial related to extending OpenAPI.

dmontagu:
@christopherthompson81 updated in the example above. I think a PR adding a note about this to the docs would be well-received.

tiangolo:
Thanks @dmontagu for the help!

@SKalt added it to the docs in #642 

So I think this is solved, right?

dmontagu:
[...]

PR #642, opened 2019-10-22
documented operationId = route.name strategy from issue #442

SKalt:
I documented the `operationId = route.name` strategy from issue #442 .  I'd be happy to `git mv` the existing tutorials such that the directory structure order is the same as the order on the page, but I'm submitting the smallest possible patch first.


tiangolo:
Great, thank you @SKalt ! :tada: :cake: 

After confirming everything, I went away and updated the names to keep the order, and indeed, Git thought it was "a lot of changes" :sweat_smile: while it was just renaming files. So thanks for keeping the minimum patch change. It's a lot easier to review it like that. :nerd_face: :taco:

PR #646, opened 2019-10-23
Make compatible with pydantic v1

dmontagu:
Still needs docs updates, but tests pass and raise no deprecation warnings.


dmontagu:
@tiangolo I recognize this is a lot of changes in a single PR, and may be painful to work through (mostly minor, but still). Let me know if there is anything I can do to make it easier for you to review.

Let me know if you think this PR has a chance of being merged -- if so, I'll start work on updating the docs too.

fastapi/encoders.py
```diff
@@ -14,7 +14,7 @@ def jsonable_encoder(
     include: Union[SetIntStr, DictIntStrAny] = None,
     exclude: Union[SetIntStr, DictIntStrAny] = set(),
     by_alias: bool = True,
-    skip_defaults: bool = False,
+    exclude_unset: bool = False,
```
dmontagu: I renamed `skip_defaults` to `exclude_unset` everywhere, since that is the intended usage in the new pydantic version. I think this will be the best approach, but it would involve a little extra migration pain since it would require skip_defaults to be renamed. Since it has precisely the same functionality though that should be a pretty straightforward find-and-replace.

Eventually, I think it will make sense to also support the new `exclude_defaults` argument, but I figured that was better suited for a separate PR.

If you want to keep the argument and raise a deprecation warning (like what pydantic currently does), or just keep the argument named as is and use the new keyword argument when calling to pydantic, I can change it everywhere, just let me know.

fastapi/exceptions.py
```diff
@@ -15,11 +13,15 @@ def __init__(
         self.headers = headers
 
 
+RequestErrorModel = create_model("Request")
```
dmontagu: pydantic now actually checks for the provided type to be a `BaseModel` so that it can handle pydantic dataclasses. Since we were only using the `Request` and `WebSocket` types to get the right name, I figured the `create_model` function provided the easiest way to get compatible models with the same name.

I would be open to changing the name, but since this results in the same error messages fastapi was spitting out before, I think this makes sense for now, and we can change the naming in a separate PR if desired.

fastapi/openapi/models.py
```diff
@@ -24,13 +23,13 @@ class EmailStr(str):  # type: ignore
 
 class Contact(BaseModel):
     name: Optional[str] = None
-    url: Optional[UrlStr] = None
+    url: Optional[AnyUrl] = None
```
dmontagu: There are a few new url types, but `AnyUrl` is the only one that is compatible with everything that `UrlStr` was compatible with before (namely, `ftp` and `ftps` schemes). `AnyHttpUrl` or `HttpUrl` would work if we are okay enforcing the scheme being `http` or `https`. (`AnyHttpUrl` allows there to not be a tld, `HttpUrl` requires it; it seems to me like `HttpUrl` would be fine here but the downside of using `AnyUrl` seems pretty minimal, and it should prevent breaking anything.)

tests/test_application.py
```diff
@@ -68,7 +68,7 @@
                 "parameters": [
                     {
                         "required": True,
-                        "schema": {"title": "Item_Id"},
+                        "schema": {"title": "Item Id"},
```
dmontagu: At some point recently, the title generation logic changed to use spaces instead of underscores.

tests/test_path.py
```diff
@@ -173,10 +183,10 @@ def test_nonexistent():
         ("/path/float/True", 422, response_not_valid_float),
         ("/path/float/42", 200, 42),
         ("/path/float/42.5", 200, 42.5),
-        ("/path/bool/foobar", 200, False),
+        ("/path/bool/foobar", 422, response_not_valid_bool),
```
dmontagu: pydantic now raises validation errors for not-very-bool-like values

dmontagu (commented):
Most of the files changed are due to renaming `skip_defaults` to `exclude_unset`, `Field` -> `ModelField`, `Schema` -> `FieldInfo`, or the fact that the title generation in schemas replaced underscores with spaces.

I've tried to call out below all of the changes that fell into other categories or where there might be some decision to be made. (It's possible I missed places, but I did go over the entire diff line by line before submitting this review.)

dmontagu:
Also, @tiangolo, not sure when you think you'll be able to get to this, but it would be really great to quickly update the supported pydantic version once https://github.com/samuelcolvin/pydantic/pull/722 is merged (the mypy plugin for pydantic).

@samuelcolvin said he'd be able to take a closer look at that PR early this upcoming week; if the response is positive and it is likely to be merged in the short term, it would be great if we could update directly to that supporting version of pydantic, even if it meant waiting a little longer to get this merged. (Or at least get a second quick release if necessary when the update to pydantic is released.)

------

Alternatively, maybe FastAPI can move to a looser pinning of pydantic (e.g., `"^1.0"`) now that it has hit a major version? That would solve the issue completely, *and* make it easier to incorporate any additional new pydantic features!

samuelcolvin:
I'll look at the mypy plugin soon, is it blocking the fastapi upgrade? Hopefully it'll be added as pydantic v1.1.

dmontagu:
No, it doesn't block the fastapi upgrade at all, I would just really like to use it, and would be sad if the v1 upgrade happened quickly but then it took a while to get compatibility with v1.1 (or whenever it gets merged).

It's not a big deal either way (I can always just continue to copy and paste the mypy extension into each of my projects), it's just a pain because I have many projects all with slightly different versions of it as I have made minor fixes over time.

samuelcolvin:
Great. Like you say, pydantic should strictly follow semvar now, so fastapi should work with all version 1 releases.

bharling:
+1 for this PR, very much looking forward to using the new features of pydantic in my project

juhovh-aiven:
Also :+1: for this PR, I'm planning to build a new API project using FastAPI, but the upcoming breaking change results in some concerns and I might have trouble pitching the choice of FastAPI until there is some clarity.

martijn-iac:
@dmontagu can I help with pushing this PR through?

fastapi/concurrency.py
```diff
@@ -1,6 +1,6 @@
 from typing import Any, Callable
 
-from starlette.concurrency import iterate_in_threadpool, run_in_threadpool  # noqa
```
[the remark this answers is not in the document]
tiangolo: Actually, I was planning on importing everything from Starlette, and making it available from FastAPI, so that users (and even myself) don't have to guess what to import from FastAPI and what from Starlette.

And as I was creating that "context manager in threadpool" that was not provided by Starlette, I thought of including `iterate_in_threadpool` right away...

Anyway, that's the reason. But I'll finish that in a PR later. :sweat_smile:

fastapi/dependencies/utils.py
```diff
@@ -239,7 +250,9 @@ def get_dependant(
             continue
         if add_non_field_param_to_dependency(param=param, dependant=dependant):
             continue
-        param_field = get_param_field(param=param, default_schema=params.Query)
+        param_field = get_param_field(
+            param=param, default_field_info=params.Query, param_name=param_name
```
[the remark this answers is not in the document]
tiangolo: Perfect, thanks for the clarification.

fastapi/dependencies/utils.py
```diff
@@ -293,59 +307,63 @@ def add_non_field_param_to_dependency(
...
     else:
-        schema = default_schema(default_value)
+        field_info = default_field_info(default_value)
     required = default_value == Required
     annotation: Any = Any
     if not param.annotation == param.empty:
         annotation = param.annotation
-    annotation = get_annotation_from_schema(annotation, schema)
-    if not schema.alias and getattr(schema, "convert_underscores", None):
+    annotation = get_annotation_from_field_info(annotation, field_info, param_name)
```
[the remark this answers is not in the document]
tiangolo: Cool :+1: thanks.

fastapi/dependencies/utils.py
```diff
@@ -511,21 +529,23 @@ def request_params_to_args(
...
+        assert isinstance(
+            field_info, params.Param
+        ), "Params must be subclasses of Param"
         if value is None:
             if field.required:
                 errors.append(
                     ErrorWrapper(
-                        MissingError(),
-                        loc=(schema.in_.value, field.alias),
-                        config=BaseConfig,
```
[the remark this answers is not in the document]
tiangolo: Great, thanks.

fastapi/encoders.py
tiangolo: Agreed! It makes sense to keep in line with Pydantic.

I added some backward compatibility with warnings :nerd_face:

fastapi/dependencies/utils.py
```diff
@@ -293,59 +307,63 @@ def add_non_field_param_to_dependency(
...
 
-def add_param_to_fields(*, field: Field, dependant: Dependant) -> None:
-    field.schema = cast(params.Param, field.schema)
-    if field.schema.in_ == params.ParamTypes.path:
+def add_param_to_fields(*, field: ModelField, dependant: Dependant) -> None:
+    field_info = cast(params.Param, field.field_info)
+    if field_info.in_ == params.ParamTypes.path:
         dependant.path_params.append(field)
-    elif field.schema.in_ == params.ParamTypes.query:
+    elif field_info.in_ == params.ParamTypes.query:
```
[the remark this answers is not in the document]
tiangolo: Perfect. This also helped while to add the backward compatibility parts :sweat_smile: :rocket:

fastapi/exceptions.py
tiangolo: Cool! I think it's fine :+1: :rocket:

fastapi/openapi/models.py
tiangolo: Perfect, thanks for the explanation. :rocket:

tests/test_application.py
tiangolo: Nice! :sparkles:

tests/test_path.py
tiangolo: Cool, thanks!

pyproject.toml
```diff
@@ -20,7 +20,7 @@ classifiers = [
 ]
 requires = [
     "starlette >=0.12.9,<=0.12.9",
-    "pydantic >=0.32.2,<=0.32.2"
+    "pydantic >=0.32.2,<=1.0.0"
```
dmontagu: @tiangolo Any chance we could get either a weaker pinning or <=1.1.0? This would enable use with the new pydantic mypy plugin.

tiangolo:
Thanks @dmontagu ! Awesome job! :clap: :clap: :clap: :cake: :rocket:

And thanks for the comments in the PR, they help a lot! :nerd_face: :memo: :1st_place_medal: 

---

I added some code to make it backwards compatible with Pydantic 0.32.2 with a lot of deprecation warnings. That should help developers migrating code bases.

Now all the tests still pass with Pydantic 0.32.2 except the tests for the updated OpenAPI schemas, as the schema titles now have spaces instead of underscores, and the strict booleans. But all the rest still passes. That means that developer's codebases that somehow require Pydantic 0.32.2 will still work as expected in the next version of FastAPI, although with deprecation warnings. 

I didn't add extra tests for that, and all the backwards compatible checks in FastAPI's codebase look kinda ugly and cumbersome, but I bet that might help some developers doing the migration more smoothly. Later, in a future version, I'll remove all that.

I tried extending the range to Pydantic v1.1.1, but as in FastAPI we are using some of the internals of Pydantic (things that wouldn't be covered by SemVer, so, it's not Pydantic's fault), some things have changed and break stuff here, so I think I'll just have to keep it pinned and update it often when there are new releases of Pydantic.

We can debug and fix what changed in v1.1.1 later, but that'll be in another PR, this one is good as it is right now.

I also updated the docs for the examples that changed.

---

This is ready to be merged now, I think everything's fine. And I see you dobule-triple-checked everything too :nerd_face: :100:

But as I added some extra stuff, and this is a big refactor, I wanted to give you the chance to check if you see anything that doesn't seem right, although it's not necessary. Otherwise, I'll just merge it tomorrow and release :rocket:

pyproject.toml
tiangolo: Yeah, I just tried, but there are some breaking changes (or even a bug maybe?).

And as we are using some internals not covered by SemVer, it's not really Pydantic's fault to change the stuff that is not expected to be used by others.

I think we'll have to stick to keep checking and updating it after testing it when there are new releases.

dmontagu: Yeah, I thought that might be the case (regarding internals).

I'll take a look now and see what's breaking v1.1, and take responsibility for opening a pydantic issue (or issues) if necessary.

dmontagu:
> I didn't add extra tests for that, and all the backwards compatible checks in FastAPI's codebase look kinda ugly and cumbersome, but I bet that might help some developers doing the migration more smoothly. Later, in a future version, I'll remove all that.

Sounds good to me!

> I tried extending the range to Pydantic v1.1.1, but as in FastAPI we are using some of the internals of Pydantic (things that wouldn't be covered by SemVer, so, it's not Pydantic's fault), some things have changed and break stuff here, so I think I'll just have to keep it pinned and update it often when there are new releases of Pydantic.
> 
> We can debug and fix what changed in v1.1.1 later, but that'll be in another PR, this one is good as it is right now.

Sounds good.

> But as I added some extra stuff, and this is a big refactor, I wanted to give you the chance to check if you see anything that doesn't seem right, although it's not necessary. Otherwise, I'll just merge it tomorrow and release 🚀

I reviewed the remaining changes; they all look good to me! Feel free to release whenever.

🎉🚀

dmontagu:
@tiangolo for what it's worth, I found the two lines necessary to change for compatibility with pydantic v1.1.1:

* In fastapi.utils, line 148 needs to be changed from:
    ```python
    new_field._populate_validators()
    ```
    to
    ```python
    new_field.populate_validators()
    ```
    Looks like this became a public interface as of v1.1.1; maybe @samuelcolvin can confirm.

* Immediately below line 375 of fastapi/dependencies/utils.py, you need to add the following line:
    ```python
    field.required = required
    ```
    It looks like this is a due to a regression where things annotated as `Any` became non-required. I believe this is a bug and will create an issue in pydantic.

samuelcolvin:
I would strongly encourage fastapi to not constrain the version of pydantic beyond the major version. Doing so indicates that either pydantic is changing too much in minor versions or fastapi is using pydantic features which aren't supposed to be used externally - probably until both these are true.

To try and allow fast API to constrain pydantic stress strictly I would be happy to run all or most of the fastapi test suit during pydantic testing to avoid unexpected incompatibility. 

In addition fastapi should resist using private methods in pydantic, instead it would be better if you could submit a PR or issue to pydantic asking for a required feature to be made external, or implement the logic yourself.

Would you accept the approach of running fastapi's tests in return for not constraining pydantic beyond `>=1.1, <2` or similar?

---

Regarding the two points raised by @dmontagu above:
* `populate_validators` became public, probably best that you just require pydantic `>=v1.1` or maybe `>=1.2` since that will be released soon.
* I've answered the issue https://github.com/samuelcolvin/pydantic/issues/1028 but let me know if there's anything else I can do to mitigate this?

tiangolo:
Awesome job debugging that @dmontagu ! :rocket: :nerd_face: 

> To try and allow fast API to constrain pydantic stress strictly I would be happy to run all or most of the fastapi test suit during pydantic testing to avoid unexpected incompatibility.

> Would you accept the approach of running fastapi's tests in return for not constraining pydantic beyond >=1.1, <2 or similar?

Oh, would you do that? I would love it!

> In addition fastapi should resist using private methods in pydantic, instead it would be better if you could submit a PR or issue to pydantic asking for a required feature to be made external, or implement the logic yourself.

That sounds great! I'll check for the next versions if I'm using something internal and request/PR to make it public.

In short, I would love to keep loose constraints on Pydantic versions.

---

I'll hold this PR, for now, to merge it after Pydantic v1.2 and relax the constraints.

samuelcolvin:
Sounds good, let's do roughly the following:
* release pydantic v1.2
* fix and merge this PR 
* release the new version of fastapi
* get pydantic to run tests for the latest git tag of fastapi

samuelcolvin:
@tiangolo a decision on samuelcolvin/pydantic#1028 might be good before v1.2

tiangolo:
I just realized that if I'm adding all those temporary backward compatible layers (with deprecation warnings) to support Pydantic >= `0.32.2`, I might as well add the workaround / fix suggested by @dmontagu to also support Pydantic from `1.0.0` to `1.1.1`.

After https://github.com/samuelcolvin/pydantic/pull/1031/ is done, everything will just keep working.

And later, when we remove support for Pydantic < `1.2.0`, we can remove those extras here.

---

Also, having FastAPI's tests (or at least some of them) run along with Pydantic in its CI will give us the maximum confidence possible, so we can relax Pydantic's version constraints here right away. So I'm setting it to < `2.0.0` :tada: :rocket: .

Thanks @dmontagu , awesome job as always! And thanks @samuelcolvin for all the help! :cake: :tada:

prostomarkeloff:
It happened! Thank you @dmontagu and @samuelcolvin for it!

PR #647, opened 2019-10-24
Fix protocol separator typo.

forestmonster:


tiangolo:
Oops! Good catch! Thanks! :mag: :pencil2: :cake:

PR #681, opened 2019-11-05
Fix #679: allow async class methods as dependencies

### LINKED ISSUE #679: [BUG] Async class method dependency raises a ValueError
**Describe the bug**
If you use an async class method as a dependency, a `ValueError` is thrown. It doesn't happen for a non-async method.

Complete error: `ValueError: [KeyError(<class 'coroutine'>), TypeError("'coroutine' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]` (at `fastapi/encoders.py:106`)

**To Reproduce**

```py

class DependencyClass:
    async def async_dep(self, request: Request):
        return True

    def sync_dep(self, request: Request):
        return True

app = FastAPI()
dependency = DependencyClass()

# Error
@app.get('/async-dep')
def async_dep(r=Depends(dependency.async_dep)):
    return r

# Everything is fine
@app.get('/sync-dep')
def sync_dep(r=Depends(dependency.sync_dep)):
    return r
```

**Expected behavior**
The async class method dependency should be called and its return value injected.

**Environment:**
 - OS: macOS
 - FastAPI Version: 0.42.0
- Python version: 3.7.2

**Additional context**
I believe the issue comes from here:

https://github.com/tiangolo/fastapi/blob/65536cbf63318d111bf608960378d651b6c1596a/fastapi/dependencies/utils.py#L353-L359

Indeed, `inspect.isfunction(call)` will return `False` in case of a class method. Hence, it is [sent to `run_in_threadpool`](https://github.com/tiangolo/fastapi/blob/65536cbf63318d111bf608960378d651b6c1596a/fastapi/dependencies/utils.py#L453-L456), which never awaits the coroutine, and we end up trying to serialize it instead of its result (hence the `ValueError`).

Changing the check by:

```py
if inspect.isfunction(call) or inspect.ismethod(call):
```

solves the issue. I can make a PR with the fix and unit tests if it helps.

frankie567:
This is the fix for issue #679: allow to use async class methods as dependencies.

Before, such methods were not correctly detected as coroutine and thus were not awaited but rather sent to `run_in_threadpool`. We ended up then injecting the coroutine instead of its result.

I've used [`inspect.isroutine`](https://docs.python.org/3/library/inspect.html#inspect.isroutine), instead of doing `inspect.isfunction(call) or inspect.ismethod(call)`. It seems to fulfill our need while saving us a check.

I also took this opportunity to add some unit tests to verify the different use cases regarding class dependencies (callable, async callable, sync methods and async methods).

It could also be interesting to mention those possibilities in the [Classes as Dependencies](https://fastapi.tiangolo.com/tutorial/dependencies/classes-as-dependencies/) section of the documentation. What do you think?

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -351,7 +351,7 @@ def add_param_to_fields(*, field: Field, dependant: Dependant) -> None:
 
 
 def is_coroutine_callable(call: Callable) -> bool:
-    if inspect.isfunction(call):
+    if inspect.isroutine(call):
         return asyncio.iscoroutinefunction(call)
     if inspect.isclass(call):
         return False
--- a/tests/test_dependency_class.py
+++ b/tests/test_dependency_class.py
@@ -1,0 +1,70 @@
+import pytest
+from fastapi import Depends, FastAPI
+from starlette.testclient import TestClient
+
+app = FastAPI()
+
+
+class CallableDependency:
+    def __call__(self, value: str) -> str:
+        return value
+
+
+class AsyncCallableDependency:
+    async def __call__(self, value: str) -> str:
+        return value
+
+
+class MethodsDependency:
+    def synchronous(self, value: str) -> str:
+        return value
+
+    async def asynchronous(self, value: str) -> str:
+        return value
+
+
+callable_dependency = CallableDependency()
+async_callable_dependency = AsyncCallableDependency()
+methods_dependency = MethodsDependency()
+
+
+@app.get("/callable-dependency")
+async def get_callable_dependency(value: str = Depends(callable_dependency)):
+    return value
+
+
+@app.get("/async-callable-dependency")
+async def get_callable_dependency(value: str = Depends(async_callable_dependency)):
+    return value
+
+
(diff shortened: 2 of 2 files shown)


frankie567:
The CI fails because of linting issues (not related to my changes). Should I fix this or do you prefer a separate PR?

euri10:
Black is not pinned and indeed it raises issues it wasn't before

Le mar. 5 nov. 2019 à 9:11 AM, François Voron <notifications@github.com> a
écrit :

frankie567:
Yeah, they fixed something regarding function parameters/trailing commas. Should be quite straightforward to fix, but maybe it should live in another PR. I'll do that quickly.

prostomarkeloff:
@tiangolo what's about it?

frankie567:
Rebased the branch with fixed linting to have the pipeline checked 👍

tiangolo:
Great, thank you @frankie567 ! Thanks for implementing it, and adding extended tests. :nerd_face: :rocket: :cake: :tada: 

And thanks everyone for the discussion.

PR #698, opened 2019-11-10
Clarify docs for APIRouter dependencies

marier-nico:
Someone might expect that the dependencies specified on an APIRouter be passed along to their path operation functions, though the behavior is closer to dependencies specified in path operation decorators. Being explicit here will make the behavior unambiguous.

This is especially true for someone who might be jumping around in the documentation and who wouldn't necessarily have read [dependencies in path operation decorators](https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-in-path-operation-decorators/).

### DIFF
--- a/docs/tutorial/bigger-applications.md
+++ b/docs/tutorial/bigger-applications.md
@@ -246,7 +246,7 @@ We can also add a list of `tags` that will be applied to all the *path operation
 
 And we can add predefined `responses` that will be included in all the *path operations* too.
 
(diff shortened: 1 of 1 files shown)


tiangolo:
Great, thanks @marier-nico ! :memo: :cake: :tada: 

Thanks for the explanation / argument, and thanks for writing the update to the docs using the same conventions as in the rest of the docs (with italics for those keywords/names), very perceptive! :nerd_face:

marier-nico:
No problem, I'm happy I could contribute to the project! :smile:

PR #710, opened 2019-11-14
Fixed the returned XML

OcasoProtal:
The returned xml was not valid, it started with spaces and had a second Header tag instead of closing the first one.


tiangolo:
Oops! :see_no_evil: 

Great! Thanks @OcasoProtal ! :mag: :bug: :cake: :tada:

Issue #673, opened 2019-10-31
Include openapi callback spec for operation/route


booooh:
**Description**

It's my understanding that OpenAPI 3.0 added support for callback objects (https://swagger.io/docs/specification/callbacks/)

As such, it allows defining the schema/model for the operation that the callback would need to implement.

How can I Include the details of a callback specification for an operation?
I couldn't find a way to attach this information to a Route/Operation.

Ideally, I would be able to define the schema of the request that would be handled by the callback.

**Additional context**
This is the spec definition for callback objects:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#callbackObject

booooh:
If anyone has some information on how to add these, that would be great!
Also, if this isn't directly supported, I'd be happy to try to put a PR together, but would need some pointers on what I should be looking at.

Thanks!

booooh:
@tiangolo  I've started a  PR with a basic implementation that enables creating APIRoutes for callback objects. If you can take a look to see if this is along the lines of what you'd expect for this -- that would be great.

tiangolo:
Thanks for your PR @booooh ! As it's already merged, I'm gonna close this issue now.

PR #722, opened 2019-11-20
Adding support for documenting OpenAPI callbacks through APIRoutes

booooh:
Added support for automatic generation and validation (via pydantic) of OpenAPI callback objects.

This is currently done by providing a list of `APIRoute` objects to a new keyword argument in the path decorators (post, get, put, etc.).

This is convenient, as that list can easily be compiled by creating an `APIRouter` object, and defining the expected (callback) routes. This has the added benefit that all Request and Response Schemas are added to the documentation, and the router used for defining the callbacks makes a solid base for mocking/testing the callback endpoints that a client would implement.

Please let me know if you have any comments/concerns.
Right now, the tests are failing because the coverage decreased a little. I'll try to see if this can be fixed. I can also squash and rebase my changes if needed.


booooh:
Fixes issue #673

tiangolo:
Thanks! I'll review it soon.

booooh:
@tiangolo  - any update on this? Anything you'd like to discuss?

docs/tutorial/openapi-callbacks.md
```diff
@@ -0,0 +1,16 @@
+There are cases where the API that is being developed will require the clients to implement callback handlers, that will in turn be invoked
```
dmontagu:
```suggestion
There are cases where the API that is being developed will require the clients to implement callback handlers that will in turn be invoked
```

docs/tutorial/openapi-callbacks.md
```diff
@@ -0,0 +1,16 @@
...
+by the application. 
+
+Since this is a cuommon usecase, the OpenAPI spec defines how these callbacks should be documented: <a href="https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#callbackObject">OpenAPI callback object spec</a>,
```
dmontagu:
```suggestion
Since this is a common usecase, the OpenAPI spec defines how these callbacks should be documented: <a href="https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#callbackObject">OpenAPI callback object spec</a>,
```

docs/tutorial/openapi-callbacks.md
```diff
@@ -0,0 +1,16 @@
...
+and there are also some more detailed examples available <a href="https://swagger.io/docs/specification/callbacks/">Swagger callback example</a>.
```
dmontagu:
```suggestion
and there are also some more detailed examples available: <a href="https://swagger.io/docs/specification/callbacks/">Swagger callback example</a>.
```

docs/tutorial/openapi-callbacks.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Use case
+
+Often, clients of a `FastAPI` application, will need to register callback handlers that will be notified when certain events occur.
```
dmontagu:
```suggestion
Clients of a `FastAPI` application may need to register callback handlers that will be notified when certain events occur.
```

fastapi/__init__.py
```diff
@@ -1,6 +1,6 @@
 """FastAPI framework, high performance, easy to learn, fast to code, ready for production"""
 
-__version__ = "0.42.0"
+__version__ = "0.42.1"
```
dmontagu: You should rebase against the most recent release; I think some of the minor formatting changes (like the extra commas) will go away. Also, don’t worry about bumping the version, @tiangolo will take care of that before the next release.

fastapi/openapi/utils.py
```diff
@@ -178,6 +180,19 @@ def get_openapi_path(
...
+            if route.callbacks:
+                callbacks = {}
+                for callback in route.callbacks:
+                    cb_path, cb_security_schemes, cb_definitions, = get_openapi_path(
+                        route=callback, model_name_map=model_name_map
+                    )
+                    logging.error(cb_definitions)
+                    logging.error(cb_path)
+                    callbacks[callback.name] = {callback.path: cb_path}
+                    # TODO: - do we need to add these definitions here?
```
dmontagu: @tiangolo I haven’t had a chance to think about this point yet, wanted to draw your attention to it though.

dmontagu (changes requested):
We also keep 100% test coverage of the docs `.py` files, which it looks like you haven’t added. I would recommend importing the app declares in the docs file inside your test file, rather than duplicating the code. You should be able to find examples along these lines in some of the other test files.

booooh:
@dmontagu - thanks for taking the time to review the changes! I'll incorporate the changes you requested, and rebase the code against the latest release.

Before moving ahead though, I just want to confirm that you are OK with the approach of defining the callbacks in this manner, and that I'm not abusing any of the existing constructs 😉

dmontagu:
@booooh 

It looks like you are properly following all conventions, as far as I can tell. 👍

I haven't closely read the openapi spec for callbacks yet, and probably would want to before merging this, but as long as the schema you are outputting matches the spec I would be happy to support this feature, and I think the implementation here is good.

-----

At some point I would like to refactor the endpoint creation process so that we don't have to copy the long list of openapi keyword arguments into each of the different route decorator methods; it makes adding features like this somewhat daunting / painful. But approaches to that have been brought up in other issues, and for now I am happy to continue following that convention.

tiangolo:
Yep, this looks good, thanks for making this! :rocket: :cake: 

Maybe it feels weird to have a full router to declare endpoints that will never be used, but most of the logic was already there, and it will be the most familiar thing to use for FastAPI users, so I think it's fine, it's the most sensible approach.

For the rest, agreed with @dmontagu (thanks for the review!).

I also want to check some things that could be updated later, but some I don't know if they would work or not, I'll do it afterwards on top of your changes.

tiangolo:
I'm currently working on this, I'll send my updates on top of your branch once they are done. So you can just wait a bit.

fastapi/openapi/utils.py
tiangolo: Thanks! Yep, it's not needed.

tiangolo:
Thanks! :tada: :rocket: 

I updated it a bit. I also added some extras that improve how this feature works.

Some of the updates:

* Remove unnecessary code, logs.
* Update type declarations.
* Update `operationId` generation, to ensure all of them are valid Python names.
* Update tests according to the `operationId`s that changed.
* Include a `default_response_class` parameter when creating an `APIRouter`, to simplify the creation of these callbacks.
* Update/increase docs and tutorial example.

PR #761, opened 2019-12-01
Update gendered language to remove male bias

BriiMawler:
Changed 'he' to the singular 'they' in handling exceptions section of tutorial

### DIFF
--- a/docs/tutorial/handling-errors.md
+++ b/docs/tutorial/handling-errors.md
@@ -4,9 +4,9 @@ This client could be a browser with a frontend, the code from someone else, an I
 
 You could need to tell that client that:
 
-* He doesn't have enough privileges for that operation.
-* He doesn't have access to that resource.
-* The item he was trying to access doesn't exist.
+* The don't have enough privileges for that operation.
+* They don't have access to that resource.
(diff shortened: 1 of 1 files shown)


docs/tutorial/handling-errors.md
```diff
@@ -4,9 +4,9 @@ This client could be a browser with a frontend, the code from someone else, an I
 
 You could need to tell that client that:
 
-* He doesn't have enough privileges for that operation.
-* He doesn't have access to that resource.
-* The item he was trying to access doesn't exist.
+* The don't have enough privileges for that operation.
```
petervanderdoes: Typo. The should be They

limelights (approved):

tiangolo:
Thanks for your contribution. :memo: :woman: :man: :cake: 

I think readability and understandability are very important, and "they" might seem a bit ambiguous.

The word should actually be "it", but as "it" could be interpreted as referring to the "error" instead of the "client", I initially used "he", to make it unambiguous.

I updated it in #780 to remove the he/it/they and make it more explicit while keeping understandability. So I'll close this now.

PR #762, opened 2019-12-02
Make OAuth2PasswordRequestFormStrict importable from fastapi.security package

martyanov:
Unfortunately, everything from `fastapi.security.oauth2` available from `fastapi.security` package, except `OAuth2PasswordRequestFormStrict`. I've added the missing entry.

### DIFF
--- a/fastapi/security/__init__.py
+++ b/fastapi/security/__init__.py
@@ -10,6 +10,7 @@ from .oauth2 import (
     OAuth2,
     OAuth2PasswordBearer,
     OAuth2PasswordRequestForm,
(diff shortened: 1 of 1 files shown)


tiangolo:

But that's in fact intentional.

It's to prevent someone from accidentally using it just because it's easy to import, while most of the people should be using `OAuth2PasswordRequestForm` instead.

Having it not exported directly ensures that if someone imports and uses it is because that person really needs the strict version.

martyanov:
I cannot agree, but ok. Thanks for the clarifications!

PR #792, opened 2019-12-14
include missing callbacks parameter

jekirl:
Otherwise callbacks registered on a sub-router are not propagated back up to the application.


dmontagu:
Could you add a test that currently fails, but is fixed by this change?

tiangolo:
Thanks @jekirl ! :+1: 

Good point @dmontagu . I just added a test for this based on the one from the docs.

PR #795, opened 2019-12-16
Highlight all new lines in docs example upgrade

cdeil:


tiangolo:
Thanks @cdeil ! :memo: 

My original intention was to highlight just the input body, but I see how this can look strange, as the rest of the new code is highlighted, so, let's update it :heavy_check_mark: :tada:

PR #807, opened 2019-12-19
Use constant time comparison in HTTP Basic Auth example

zwass:
Using this security best-practice in the example will encourage users to do the same in their code.

### DIFF
--- a/docs/src/security/tutorial007.py
+++ b/docs/src/security/tutorial007.py
@@ -1,4 +1,4 @@
-import hmac
+import secrets
 
 from fastapi import Depends, FastAPI, HTTPException
 from fastapi.security import HTTPBasic, HTTPBasicCredentials
@@ -10,9 +10,9 @@ security = HTTPBasic()
 
 
 def get_current_username(credentials: HTTPBasicCredentials = Depends(security)):
-    # Be careful to use constant time comparison to prevent timing attacks
-    if (not hmac.compare_digest(credentials.username, "foo") or
-        not hmac.compare_digest(credentials.password, "password")):
+    correct_username = secrets.compare_digest(credentials.username, "stanleyjobson")
+    correct_password = secrets.compare_digest(credentials.password, "swordfish")
+    if not correct_username and correct_password:
         raise HTTPException(
             status_code=HTTP_401_UNAUTHORIZED,
             detail="Incorrect email or password",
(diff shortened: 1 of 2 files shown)


docs/src/security/tutorial007.py
```diff
@@ -8,7 +10,9 @@
 
 
 def get_current_username(credentials: HTTPBasicCredentials = Depends(security)):
-    if credentials.username != "foo" or credentials.password != "password":
+    # Be careful to use constant time comparison to prevent timing attacks
+    if (!hmac.compare_digest(credentials.username, "foo") or
```
dmontagu: shouldn't this be a `not hmac.compare_digest` rather than using `!`?

zwass: Yes, good point. I haven't been doing much Python lately.

tiangolo:
Thanks! Good idea :lock: 

I'll review it thoroughly soon, as I would like to make it easy to understand for newbies as well, explaining what an HMAC is to developers, what "constant time" means, what a "time attack" means, why it matters, etc.

tiangolo:
Thanks for your contribution @zwass ! I updated it a bit and merged it. :tada: :lock: :cake:

zwass:
Great explanation. Thank you!

docs/src/security/tutorial007.py
```diff
@@ -8,7 +10,9 @@
 
 
 def get_current_username(credentials: HTTPBasicCredentials = Depends(security)):
-    if credentials.username != "foo" or credentials.password != "password":
+    correct_username = secrets.compare_digest(credentials.username, "stanleyjobson")
+    correct_password = secrets.compare_digest(credentials.password, "swordfish")
+    if not correct_username and correct_password:
```
StephenBrown2: This needs parentheses around it, or change the comparison to `or not`:

```suggestion
    if not (correct_username and correct_password):
```

Otherwise, if the username is bad but the password is good, the check will pass, as well as if the username is good, the check will pass regardless of the password.
Example:
https://repl.it/repls/WigglyLowestBackups

AntonGsv:
@tiangolo Why do you use `secrets.compare_digest` instead of `hmac.compare_digest` from the standard module?
secrets has its own dependency tree and some of them do not compile under windows.

PR #810, opened 2019-12-22
update links

wxq0309:
Added threeFastapi Chinese documents


tiangolo:
Great! Thank you! :tada:

PR #818, opened 2019-12-26
document responses in dependency

### LINKED ISSUE #501: [FEATURE] document response in depends
No not related to a problem.

I'm using ` fastapi.security.api_key.APIKeyCookie` as a depends and it may raise `HTTPException(403, detail="Not authenticated")` for any router it is used. So i have to document a 403 responses in all related router.

Could there by a way when I'm using a class as Depends, document responses once as a class member or some thing else instead of document it many times in routers.

trim21:
when using a class as a dependency, it's possible to add aresponses with class attribute `__responses__` .

```python
import json

import pydantic

app = FastAPI()


class Response(pydantic.BaseModel):
    reason: str


class User(pydantic.BaseModel):
    id: int
    name: str


class GetUser:
    __responses__ = {401: {"model": Response}}

    async def __call__(self, user_id: int) -> User:
        return User.parse_obj({"id": user_id, "name": "Sum"})


get_user = GetUser()


@app.get("/", responses={200: {"model": User}})
async def get(user: User = Depends(get_user)):
    return user.dict()


@app.get("/v2", responses={200: {"model": User}})
async def get_v2(user: User = Depends(get_user)):
    return user.dict()


print(json.dumps(app.openapi()["paths"]["/"]["get"]["responses"], ensure_ascii=False))
"""{
    "401": {
        "description": "Unauthorized",
        "content": {
            "application/json": {"schema": {"$ref": "#/components/schemas/Response"}}
        },
    },
    "200": {
        "content": {"application/json": {"schema": {}}},
    },
    "422": {
        "content": {
            "application/json": {
                "schema": {"$ref": "#/components/schemas/HTTPValidationError"}
            }
        },
    },
}"""
print(json.dumps(app.openapi()["paths"]["/v2"]["get"]["responses"], ensure_ascii=False))
"""{
    "401": {
        "description": "Unauthorized",
        "content": {
            "application/json": {"schema": {"$ref": "#/components/schemas/Response"}}
        },
    },
    "200": {
        "content": {"application/json": {"schema": {}}},
    },
    "422": {
        "content": {
            "application/json": {
                "schema": {"$ref": "#/components/schemas/HTTPValidationError"}
            }
        },
    },
}"""
```

close #501

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -281,7 +281,26 @@ class APIRoute(routing.Route):
         # truncate description text to the content preceding the first "form feed"
         self.description = self.description.split("\f")[0]
         self.response_description = response_description
-        self.responses = responses or {}
+
+        assert inspect.isfunction(endpoint) or inspect.ismethod(
+            endpoint
+        ), f"An endpoint must be a function or method"
+        self.dependant = get_dependant(path=self.path_format, call=self.endpoint)
+        for depends in self.dependencies[::-1]:
+            self.dependant.dependencies.insert(
+                0,
+                get_parameterless_sub_dependant(depends=depends, path=self.path_format),
+            )
+
+        dependencies_response: Dict[Union[int, str], Dict[str, Any]] = {}
+        for depend in self.dependant.dependencies:
+            depend_response = getattr(depend.call, "__responses__", None)
+            if depend_response:
+                dependencies_response.update(depend_response)
+
+        dependencies_response.update(responses or {})
+        self.responses = dependencies_response
+
         response_fields = {}
         for additional_status_code, response in self.responses.items():
             assert isinstance(response, dict), "An additional response must be a dict"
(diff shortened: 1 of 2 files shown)


dmontagu:
In general, I would be in favor of anything that makes it easier to reusably-specify additional responses (as right now it just isn't very ergonomic), but I'm not sure I love this approach for the following reasons:

1. It seems difficult-to-discover and doesn't feel particularly consistent with the way things work elsewhere in the library
2. It does nothing to identify or warn against conflicts coming from different dependencies that could result in a response with the same error code

I think it would make more sense to have a single attribute on any endpoint/dependency/etc. that would be a place that you could add a variety of additional information to inject into the OpenAPI schema. This would make the behavior consistent for a variety of potential modifications, and would make it easier to package the functionality up into a more user-friendly decorator.

I proposed something similar as a way of integrating with the starlette 0.13 routing-table-style approach to declaring routes, and I think it could make sense to support the same functionality for dependencies as a way to more consistently address #501 and other OpenAPI-related feature requests.

Trim21:
So maybe use attribute `__fastapi__`?
```python
class GetUser:
    __fastapi__ = {
        'response': {401: {"model": Response}},
    }
```

Trim21:
or a decorator
```python
from someplace import depend_decorator


def depend_decorator(responses=None):
    if responses is None:
        responses = {}

    def wrapped(cls):
        cls.__fastapi__ = {'responses': responses}
        return cls

    return wrapped


@depend_decorator(responses={401: {"model": 'Response'}})
class GetUser:
    async def __call__(self, user_id: int):
        return User.parse_obj({"id": user_id, "name": "Sum"})


if __name__ == '__main__':
    print(GetUser.__fastapi__)
```

prostomarkeloff:
Defining own dunders looks very very bad for me

tiangolo:
Thanks for the effort and proposal @Trim21 . 

And thanks, everyone, for the discussion here.

I prefer to have dependencies as normal functions. I want to support this through a new way to "register" dependencies with their additional responses in `FastAPI` and `APIRouter` instances. Hopefully, I'll be able to add that soon.

Regzand:
Hi @tiangolo! Was there any progress with this?

I am looking for a way to document response headers in dependency, but I haven't found anything about it except this issue.

HTML-CODDER (commented):
thanks

Issue #821, opened 2019-12-28
None should be checked with is None in docs


GeorgeFischhof:
Hi All

the library and the documentation are fantastic 

I am a QA eng., and teach Python for my colleagues.
I saw in some examples that None is checked as it would be a boolean expression
for example:
https://fastapi.tiangolo.com/tutorial/body/

async def create_item(item_id: int, item: Item, q: str = None):
    result = {"item_id": item_id, **item.dict()}
    if q:

I saw that the documentation is written for beginners too.
I think checking the None as it would be a boolean expression in the docs encourages beginners to use this formula instead of checking with is operator. And it is not good, because not None will them result unexpected results.

These examples should be corrected.

BR,
George

shifr:
@GeorgeFischhof I wouldn't label it as a `bug` though. It's a valid expression and in the example `q` could be an empty string and we don't need it.

GeorgeFischhof:
PR #824 added

GeorgeFischhof:
Hi @shifr  yes, it is valid for strings, but singletons should be checked with is operator (as PEP8 says) and explicit is better than implicit ;-) And as you wrote too this check checks two thinks, but perhaps it is not the best for beginners

shifr:
You're right about checking `is None` but in this case, we need to check if `q` exists and it is as explicit as needed. Also, don't forget about `Simple is better than complex.`

tiangolo:
Thanks for the interest @GeorgeFischhof , but as I commented on the PR, I would prefer to keep it as is.

It's a bit more of a personal preference than best practices.

In fact, using `q is not None` would mean that `q` would be used even if it has an empty string `""`, which is probably not the intention. But thanks again for your time and interest! :rocket:

PR #824, opened 2019-12-29
bugfix bug ID #821

GeorgeFischhof:
[BUG] [DOC] None should be checked with is None in docs #821

Hi All

the library and the documentation are fantastic

I am a QA eng., and teach Python for my colleagues.
I saw in some examples that None is checked as it would be a boolean expression
for example:
https://fastapi.tiangolo.com/tutorial/body/

async def create_item(item_id: int, item: Item, q: str = None):
result = {"item_id": item_id, **item.dict()}
if q:

I saw that the documentation is written for beginners too.
I think checking the None as it would be a boolean expression in the docs encourages beginners to use this formula instead of checking with is operator. And it is not good, because not None will them result unexpected results.

These examples should be corrected.

BR,
George

### DIFF
--- a/docs/src/query_params/tutorial003.py
+++ b/docs/src/query_params/tutorial003.py
@@ -6,7 +6,7 @@ app = FastAPI()
 @app.get("/items/{item_id}")
 async def read_item(item_id: str, q: str = None, short: bool = False):
     item = {"item_id": item_id}
-    if q:
+    if q is not None and len(q) > 0:
         item.update({"q": q})
     if not short:
         item.update(
--- a/docs/src/query_params/tutorial003.py
+++ b/docs/src/query_params/tutorial003.py
@@ -6,7 +6,7 @@ app = FastAPI()
 @app.get("/items/{item_id}")
 async def read_item(item_id: str, q: str = None, short: bool = False):
     item = {"item_id": item_id}
-    if q is not None and len(q) > 0:
+    if q is not None:
         item.update({"q": q})
     if not short:
         item.update(
--- a/docs/src/query_params/tutorial004.py
+++ b/docs/src/query_params/tutorial004.py
@@ -8,7 +8,7 @@ async def read_user_item(
     user_id: int, item_id: str, q: str = None, short: bool = False
 ):
     item = {"item_id": item_id, "owner_id": user_id}
-    if q is not None and len(q) > 0:
+    if q is not None:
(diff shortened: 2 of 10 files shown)


dmontagu:
Admittedly, this is a question of personal preference, but for performance reasons I would advise against the use of

```python
q is not None and len(q) > 0
```

There is a relatively high variance, but it seems to run ~2-3x slower than just `if q:` for string inputs:

```python
import time

N = 1_000_000
q = "abc"

# Warm up
for _ in range(N):
    pass

t0 = time.time()
for _ in range(N):
    if q:
        pass

t1 = time.time()
print(t1 - t0)
# 0.07553

t0 = time.time()
for _ in range(N):
    if q is not None and len(q) > 0:
        pass
t1 = time.time()
print(t1 - t0)
# 0.20973
```

If you want to preserve the semantics but make the null check explicit, a more performant (and in my opinion more pythonic) alternative would be:
```python
if q is not None and q:
```
But this is still ~20% slower than just `if q:`.

In general, many linters in many languages will raise warnings for a check of the form `if len(x) > 0` instead of `if x.is_empty()` (which is typically done via just `if x` in python), due to the overhead involved with grabbing the length and then comparing it to zero. Since checks like this can frequently happen inside relatively hot loops in python, I think it is worth being in the habit of thinking about performance (as long as your code remains pythonic, which I would argue this pattern is).

------

Admittedly for an endpoint like this, the overhead of the checks is not going to be significant. But since the whole point of the PR is to help expose beginners to better practices, I'm not sure that I would say this is a strict improvement on that front.

--------

In this case though, it's not clear to me that the goal should really be to keep the semantics exactly the same -- if q="", I think it would be reasonable to perform the update.

So then it could make sense to change to `if q is not None`. I would totally agree that `if q is not None` is better than `if q` if that is a reasonable modification to the semantics.

GeorgeFischhof:
I think `if q is not None` will be good, also pydantic which will check the string type. I will rewrite the PR accordingly

tiangolo:
Thanks for your contribution and interest @GeorgeFischhof :memo: 

But I would prefer to keep it as is. As @dmontagu says, it's a bit more of a personal preference than best practices.

In fact, using `q is not None` would mean that `q` would be used even if it has an empty string `""`, which is probably not the intention. But thanks again for your time and interest! :rocket:

PR #853, opened 2020-01-13
add body to RequestValidationError for easier tracing and debugging

aviramha:
FastAPI is awesome, and it's schema validation is one of it's key elements. 
While debugging issues in production, the current errors data isn't sufficient to understand if it's a bad schema definition (when the API is used by multiple sources, in different ways) or really a client defect.
Adding the body data to the error can help trace problems and reproducing the issue.

### DIFF
--- a/docs/src/handling_errors/tutorial005.py
+++ b/docs/src/handling_errors/tutorial005.py
@@ -1,28 +1,27 @@
-from fastapi import FastAPI, HTTPException
-from fastapi.exception_handlers import (
-    http_exception_handler,
-    request_validation_exception_handler,
-)
+from fastapi import FastAPI
+from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
-from starlette.exceptions import HTTPException as StarletteHTTPException
+from pydantic import BaseModel
+from starlette import status
+from starlette.requests import Request
+from starlette.responses import JSONResponse
 
 app = FastAPI()
 
 
-@app.exception_handler(StarletteHTTPException)
-async def custom_http_exception_handler(request, exc):
-    print(f"OMG! An HTTP error!: {exc}")
-    return await http_exception_handler(request, exc)
-
-
 @app.exception_handler(RequestValidationError)
-async def validation_exception_handler(request, exc):
-    print(f"OMG! The client sent invalid data!: {exc}")
-    return await request_validation_exception_handler(request, exc)
+async def validation_exception_handler(request: Request, exc: RequestValidationError):
+    return JSONResponse(
+        status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+        content=jsonable_encoder({"detail": exc.errors(), "body": exc.body}),
+    )
 
 
-@app.get("/items/{item_id}")
-async def read_item(item_id: int):
-    if item_id == 3:
-        raise HTTPException(status_code=418, detail="Nope! I don't like 3.")
-    return {"item_id": item_id}
+class Item(BaseModel):
+    title: str
+    size: int
+
+
+@app.post("/items/")
+async def create_item(item: Item):
+    return item
--- a/docs/src/handling_errors/tutorial006.py
+++ b/docs/src/handling_errors/tutorial006.py
@@ -1,0 +1,28 @@
+from fastapi import FastAPI, HTTPException
+from fastapi.exception_handlers import (
+    http_exception_handler,
+    request_validation_exception_handler,
+)
+from fastapi.exceptions import RequestValidationError
+from starlette.exceptions import HTTPException as StarletteHTTPException
+
+app = FastAPI()
+
+
+@app.exception_handler(StarletteHTTPException)
+async def custom_http_exception_handler(request, exc):
+    print(f"OMG! An HTTP error!: {exc}")
+    return await http_exception_handler(request, exc)
+
+
+@app.exception_handler(RequestValidationError)
+async def validation_exception_handler(request, exc):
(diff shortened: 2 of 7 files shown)


victoraugustolls:
Can this be made optional? For sensitive bodies one might not want it to be in the error, depending on how they log them.

aviramha:
> Can this be made optional? For sensitive bodies one might not want it to be in the error, depending on how they log them.

Handling and logging of the error hasn't changed, so it's up to the user of the framework what to do with this data.
For example in my application I wish to have this so when I catch and handle this exception, I'd be able to get the body that caused the parsing exception.

victoraugustolls:
For example using sentry sdk that logs the whole exception. The body will be logged and will be up to the user to reconfigure his sentry to don’t log the body part. So this might change how logging should be handled in some applications.

aviramha:
> For example using sentry sdk that logs the whole exception automatically out of the box. The body will be logged and will be up to the user to reconfigure his sentry to don’t log the body part. So this might change how logging should be handled in some applications.

My opinion then is it's upto the middleware to handle this case, because I can't think of a pretty way to do this in a configurable manner...

victoraugustolls:
Me neither :D, just worried because this is something that caught me in the past (using sentry). I think it’s nice to have this information btw

aviramha:
> Me neither :D, just worried because this is something that caught me in the past (using sentry). I think it’s nice to have this information btw

Let's see what @tiangolo or other maintainer has to say then :)

tiangolo:
Thanks for your contribution @aviramha ! :rocket: 

And thanks for the discussion and points touched @victoraugustolls .

---

So, about Sentry, the current behavior is, if you have an app like:

```Python
import sentry_sdk
from sentry_sdk import capture_exception
from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
from starlette.responses import PlainTextResponse

from fastapi import FastAPI

app = FastAPI()


class User(BaseModel):
    name: str
    age: int


async def validation_exception_handler(request, exc: Exception):
    capture_exception(exc)
    return PlainTextResponse(str(exc), status_code=400)


@app.post("/users/")
def create_user(user: User):
    return user


sentry_sdk.init(dsn="https://deadbeef12345@sentry.io/6789")

app.add_middleware(SentryAsgiMiddleware)
```

And you post invalid data like:

```JSON
{
  "name": "string",
  "age": "a"
}
```

As that body is part of the traceback for the `RequestValidationError`, it will be sent to Sentry (I just tested it).

But it would only happen if the developer manually captures the exception for Sentry.

If the developer doesn't do it, nothing happens.

And this is all with the current implementation, before this PR.

---

The possibly sensitive data won't be sent automatically to Sentry, as by default, the exception is caught in FastAPI's code. Sensitive data would only be sent to Sentry if the developer explicitly sends the data or the exception. As @aviramha says, it would be up to the middleware how to handle it. If the developer wants to capture the exception and send it to Sentry, then that developer would have to make sure that any sensitive data is not a problem. And that still holds independently of this PR.

What I see this PR would do is allow to more easily handle and debug those bodies. In fact, it could help reducing sensitive data sent to a third party like Sentry, as it would be easier to log the failing data without having to resort to capture it and send it to Sentry or similar to be able to debug it.

---

I'll make a couple of tweaks to the PR first but I think this is a good idea :bulb: :+1: :tada: :cake:

aviramha:
> Thanks for your contribution @aviramha ! 🚀
> 
> And thanks for the discussion and points touched @victoraugustolls .
> 
> So, about Sentry, the current behavior is, if you have an app like:
> 
> ```python
> import sentry_sdk
> from pydantic import BaseModel
> from sentry_sdk import capture_exception
> from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
> from starlette.responses import PlainTextResponse
> 
> from fastapi import FastAPI
> from fastapi.exceptions import RequestValidationError
> 
> app = FastAPI()
> 
> 
> class User(BaseModel):
>     name: str
>     age: int
> 
> 
> @app.exception_handler(RequestValidationError)
> async def validation_exception_handler(request, exc: Exception):
>     capture_exception(exc)
>     return PlainTextResponse(str(exc), status_code=400)
> 
> 
> @app.post("/users/")
> def create_user(user: User):
>     return user
> 
> 
> sentry_sdk.init(dsn="https://deadbeef12345@sentry.io/6789")
> 
> app.add_middleware(SentryAsgiMiddleware)
> ```
> 
> And you post invalid data like:
> 
> ```json
> {
>   "name": "string",
>   "age": "a"
> }
> ```
> 
> As that body is part of the traceback for the `RequestValidationError`, it will be sent to Sentry (I just tested it).
> 
> But it would only happen if the developer manually captures the exception for Sentry.
> 
> If the developer doesn't do it, nothing happens.
> 
> And this is all with the current implementation, before this PR.
> 
> The possibly sensitive data won't be sent automatically to Sentry, as by default, the exception is caught in FastAPI's code. Sensitive data would only be sent to Sentry if the developer explicitly sends the data or the exception. As @aviramha says, it would be up to the middleware how to handle it. If the developer wants to capture the exception and send it to Sentry, then that developer would have to make sure that any sensitive data is not a problem. And that still holds independently of this PR.
> 
> What I see this PR would do is allow to more easily handle and debug those bodies. In fact, it could help reducing sensitive data sent to a third party like Sentry, as it would be easier to log the failing data without having to resort to capture it and send it to Sentry or similar to be able to debug it.
> 
> I'll make a couple of tweaks to the PR first but I think this is a good idea 💡 👍 🎉 🍰

Thanks for the elaborate comment! I'm glad you liked my idea.
Can't wait for it to get upstream.

tiangolo:

PR #856, opened 2020-01-14
Fix validating of list annotation in forms parsing

### LINKED ISSUE #842: [FEATURE] Support repeated key=value in form data

Yes.

Given some URL encoded data like this...

```
choices=parrot&choices=spider
```

...only the last key=value wins.

This does not work like I expected:
```python
choices: list = Form(...)
```

You can only validate against the last value.

Perhaps FastAPI should collect repeated keys in the 2-tuple list that `request.form()` gives and assign those values as a list to the same key before validation happens.

nsidnev:
Closes #842

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -629,9 +629,9 @@ async def request_body_to_args(
         for field in required_params:
             value: Any = None
             if received_body is not None:
-                if field.shape in sequence_shapes and isinstance(
-                    received_body, FormData
-                ):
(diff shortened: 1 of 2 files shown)


dmontagu (approved):
Looks good to me!

tiangolo:
Awesome, thanks @nsidnev for debugging and implementing the fix with extensive tests :clap: :bowing_man: :rocket: :cake: 

Thanks @dmontagu as always! :rocket: :taco:

PR #865, opened 2020-01-15
Check both username and password

isaevpd:
Right not the condition is checking that username is not correct and password is correct, this fixes is to be as it was probably intended


dmontagu (approved):
Thanks for this!

tiangolo:
Oops! Good catch, thanks for your contribution @isaevpd ! :rocket: :lock: :cake: 

And thanks @dmontagu as always! :rocket: :tada: :bowing_man:

PR #897, opened 2020-01-20
this change fixes an issue we encountered with pydantic validator dec…

deuce2367:
…orators

- create_model() is now called using a __base__ parameter vs. a __config__ parameter
- isinstance comparisons now work as expected
- unit test created for #889 still passes as before


deuce2367:
Deploys are failing -- will squash commits (patch + black formatting) and try again later.  Thanks!

PR #899, opened 2020-01-21
    this change fixes an issue we encountered with pydantic validator…

deuce2367:
… decorators

    - create_model() is now called using a __base__ parameter vs. a __config__ parameter
    - isinstance comparisons now work as expected
    - unit test created for #889 still passes as before
    - apply black formatting

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -96,7 +96,6 @@ def create_cloned_field(field: ModelField) -> ModelField:
         use_type = create_model(original_type.__name__, __base__=original_type)
         for f in original_type.__fields__.values():
             use_type.__fields__[f.name] = create_cloned_field(f)
-        use_type.__validators__ = original_type.__validators__
     if PYDANTIC_1:
         new_field = ModelField(
             name=field.name,
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -93,9 +93,7 @@ def create_cloned_field(field: ModelField) -> ModelField:
     use_type = original_type
     if lenient_issubclass(original_type, BaseModel):
         original_type = cast(Type[BaseModel], original_type)
-        use_type = create_model(
-            original_type.__name__, __config__=original_type.__config__
-        )
+        use_type = create_model(original_type.__name__, __base__=original_type)
         for f in original_type.__fields__.values():
             use_type.__fields__[f.name] = create_cloned_field(f)
         use_type.__validators__ = original_type.__validators__
--- a/tests/test_filter_pydantic_sub_model.py
+++ b/tests/test_filter_pydantic_sub_model.py
@@ -1,5 +1,5 @@
 from fastapi import Depends, FastAPI
-from pydantic import BaseModel
+from pydantic import BaseModel, validator
 from starlette.testclient import TestClient
 
 app = FastAPI()
@@ -17,6 +17,11 @@ class ModelA(BaseModel):
     name: str
     description: str = None
     model_b: ModelB
+
+    @validator("model_b")
+    def check_model_b(cls, model_b, values):
+        assert isinstance(model_b, ModelB)
+        return model_b
 
 
 async def get_model_c() -> ModelC:
--- a/tests/test_filter_pydantic_sub_model.py
+++ b/tests/test_filter_pydantic_sub_model.py
@@ -1,5 +1,6 @@
+import pytest
 from fastapi import Depends, FastAPI
-from pydantic import BaseModel, validator
+from pydantic import BaseModel, ValidationError, validator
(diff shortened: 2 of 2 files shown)


fastapi/utils.py
```diff
@@ -93,9 +93,7 @@ def create_cloned_field(field: ModelField) -> ModelField:
     use_type = original_type
     if lenient_issubclass(original_type, BaseModel):
         original_type = cast(Type[BaseModel], original_type)
-        use_type = create_model(
-            original_type.__name__, __config__=original_type.__config__
-        )
+        use_type = create_model(original_type.__name__, __base__=original_type)
```
Code0x58: This change looks correct to me, based on the [current v1.4 definition](https://github.com/samuelcolvin/pydantic/blob/v1.4/pydantic/main.py#L736), and old [v0.32 definition](https://github.com/samuelcolvin/pydantic/blob/v0.32.2/pydantic/main.py#L657).

Code0x58 (approved):

fastapi/utils.py
```diff
@@ -93,9 +93,7 @@ def create_cloned_field(field: ModelField) -> ModelField:
...
         for f in original_type.__fields__.values():
             use_type.__fields__[f.name] = create_cloned_field(f)
         use_type.__validators__ = original_type.__validators__
```
Code0x58: I think this line is redundant now

deuce2367: It does seem like there may be a simpler way to refactor this bit.  However I found that if I remove the create_model() call on line 96 it fails 2 of the unit tests so I don't know of a better solution for now:

```
======================================================== FAILURES =========================================================
__________________________________________________ test_filter_sub_model __________________________________________________

    def test_filter_sub_model():
        response = client.get("/model")
>       assert response.json() == {
            "name": "model-a-name",
            "description": "model-a-desc",
            "model_b": {"username": "test-user"},
        }
E       AssertionError: assert {'description...model-a-name'} == {'description...model-a-name'}
E         Omitting 2 identical items, use -vv to show
E         Differing items:
E         {'model_b': {'password': 'test-password', 'username': 'test-user'}} != {'model_b': {'username': 'test-user'}}
E         Use -v to get the full diff

tests/test_filter_pydantic_sub_model.py:92: AssertionError
_______________________________________________________ test_token ________________________________________________________

    def test_token():
        access_token = get_access_token(scope="me")
        response = client.get(
            "/users/me", headers={"Authorization": f"Bearer {access_token}"}
        )
>       assert response.json() == {
            "username": "johndoe",
            "full_name": "John Doe",
            "email": "johndoe@example.com",
            "disabled": False,
        }
E       AssertionError: assert {'disabled': ...Gga31lW', ...} == {'disabled': ...e': 'johndoe'}
E         Omitting 4 identical items, use -vv to show
E         Left contains 1 more item:
E         {'hashed_password': '$2b$12$EixZaYVK1fsbw1ZfbX3OXePaWxn96p36WQoeG6Lruj3vjPGga31lW'}
E         Use -v to get the full diff

tests/test_tutorial/test_security/test_tutorial005.py:231: AssertionError
============================================== 2 failed, 744 passed in 7.04s ==============================================
```

Code0x58: I found removing line 99 still had the tests pass:
```diff
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -96,7 +96,6 @@ def create_cloned_field(field: ModelField) -> ModelField:
         use_type = create_model(original_type.__name__, __base__=original_type)
         for f in original_type.__fields__.values():
             use_type.__fields__[f.name] = create_cloned_field(f)
-        use_type.__validators__ = original_type.__validators__
     if PYDANTIC_1:
         new_field = ModelField(
             name=field.name,
```

deuce2367: Nice, I found same.  Updated the branch and unit tests.

tests/test_filter_pydantic_sub_model.py
```diff
@@ -8,10 +8,20 @@
 class ModelB(BaseModel):
     username: str
 
+    @validator("username")
+    def check_username(cls, username, values):
```
Code0x58: does adding this test new scenario? If the tests here don't trigger a failure case, I don't think adding these two validators adds anything, whereas testing for failure would make sure that the validators are inherited (worth checking in other places to make sure that isn't covered elsewhere, as this module seems to have a specific scenario in mind)

tiangolo:
Thanks for your contribution @deuce2367 ! :tada: :cake: 

And thanks for the review and analysis @Code0x58 ! :nerd_face: :mag: 

I updated it a bit to test that the validator is preserved and merged it :tada: :rocket:

Issue #895, opened 2020-01-20
Support request body without pydantic


pawamoy:
Thank you very much for developing and maintaining FastAPI, it's great 🙂 

### Is your feature request related to a problem

I tried to create a POST route with `username: str` and `password: str` parameters in the function signature. I was expecting FastAPI to search for `username` and `password` in the request body (since I used `@app.post`), but instead it gets them from the URL, as query parameters. It seems it only works when the parameter is a Pydantic model, not a builtin type.

### The solution you would like

With 
```python
@app.post("/route/")
def my_route(username: str, password: str):
    ...
```

Accept `username` and `password` as fields of the request body, not as query parameters.

### Describe alternatives you've considered

Use Pydantic models. It works fine, it's just a bit less straight-forward.

```python
class LoginForm(BaseModel):
    username: str
    password: str

@app.post("/route/")
def my_route(login_data: LoginForm):
    ...
```

It could also be emphasized in the docs (https://fastapi.tiangolo.com/tutorial/body/) that it will not work with built-in types, only with Pydantic models.

### Additional context

I'm not sure my feature request makes sense: generally speaking, if POST requests can also use query parameters in the URL, then it's impossible to distinguish the ones that should be in the request body and the ones that appear in the URL as query parameters. It means it would not be possible to build the openapi.json I guess. In that case, please close this as invalid 🙂

nsidnev:
@pawamoy 
Hi! I think you are looking for [this part](https://fastapi.tiangolo.com/tutorial/body-multiple-params/#singular-values-in-body) of the documentation. Is that what you need?

So your example will look something like this:
```python3
from fastapi import FastAPI, Body

app = FastAPI()


@app.post("/route/")
def my_route(username: str = Body(...), password: str = Body(...)) -> None:
    print("username:", username)
    print("password length:", len(password))
```

pawamoy:
Indeed! Thank you for pointing me to this page. This feature request can definitely be closed. I'm really sorry for the noise! Cheers

tiangolo:
Thanks @nsidnev for your help here! :nerd_face: :cake: 

Thanks @pawamoy for reporting back and closing the issue :taco: :rocket:

PR #900, opened 2020-01-21
Add note about Body parameters

pawamoy:
Add a link to section containing the documentation for Body parameters, emphasizing that using Pydantic models is not mandatory.

It's a follow-up for issue #895. Me and another colleague had the same trouble, thinking that Pydantic models were mandatory to accept params in the request body.

Waiting for your feedback of course as I wrote this note quickly, just to initiate the PR.

### DIFF
--- a/docs/tutorial/body.md
+++ b/docs/tutorial/body.md
@@ -6,8 +6,6 @@ Your API almost always has to send a **response** body. But clients don't necess
 
 To declare a **request** body, you use <a href="https://pydantic-docs.helpmanual.io/" class="external-link" target="_blank">Pydantic</a> models with all their power and benefits.
 
-If you don't want to use Pydantic models, you can also use **Body** parameters: see the [Body documentation page](/tutorial/body-multiple-params/#singular-values-in-body).
-
 !!! info
     You cannot send a request body using a `GET` operation (HTTP method).
 
@@ -141,3 +139,7 @@ The function parameters will be recognized as follows:
 * If the parameter is also declared in the **path**, it will be used as a path parameter.
 * If the parameter is of a **singular type** (like `int`, `float`, `str`, `bool`, etc) it will be interpreted as a **query** parameter.
 * If the parameter is declared to be of the type of a **Pydantic model**, it will be interpreted as a request **body**.
+
+## Without Pydantic
+
+If you don't want to use Pydantic models, you can also use **Body** parameters. See the docs for [Body - Multiple Parameters: Singular values in body](body-multiple-params.md#singular-values-in-body){.internal-link target=_blank}.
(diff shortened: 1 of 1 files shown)


tiangolo:

mbello:
I tripped on this as well. And I really think it would be awesome to have FastAPI work with properly annotated but otherwise "unmodified" python functions and this "Body(...)" departs from that and is quite ugly I would say.

May I suggest we change the @app.post to:

**- Suggestion 1**: add a "param-lookup" argument to the post decorator so that we could specify the order in which FastAPI will look for 'non-path' and non BaseModel descendants parameters. It could be "query, body", "body, query", "query" or "body". Current default is "query", but should be changed to "query, body".

**- Suggestion 2**: add a "fallback-to-body" argument to the decorator, so if a parameter is missing after looking for them in the query string, FastAPI would then check the request body for them.

If any of these suggestions sound ok, I could try developing it and submitting a pull request. Just let me know if it would be welcomed.

mbello:
The "Body(...)" workaround also has the limitation described here: https://github.com/tiangolo/fastapi/issues/1011

pawamoy:
I think it's quite consistent to have `Path`, `Query`, `Body`, `Header` and `Cookie` parameters and to have __only one default__ when none of the above class/function is used, which is to consider it's a query parameter.

In any case, since the implicit behavior can become quite confusing with lots of parameters (default to query, pydantic models as body, params in the route as path, etc.), I think it's best to always be explicit and wrap your parameters in `Query`, `Body`, etc. (just my personal opinion!).

But I'm concerned by the issue your references, as I thought using `Body(...)` (with an ellipsis) would make the parameter mandatory, just as what is done for Query: https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#make-it-required.

mbello:
While pydantic and FastAPI will understand the parameter as being mandatory, linters and cpython will see it as optional and it is an error in python to have optional parameters preceeding non-optional parameters so it may force one to reorder parameters around.
Hence, we get into the business of "converting methods" to become FastAPI-friendly when I always thought that being able to use existing methods 'as-is' was a great appeal of FastAPI. 

Also, I really do not understand why it would be a bad idea to fallback to checking for parameters in the body, even if you choose to limit this behavior to post method and when the body content type is form or json.

May I suggest two more alternatives?

**Suggestion 3: Have the decorator accept kwargs like this:**

@app.post('/', a = Body(...), b = Path(...), c = Query(...))
async def test(a: str, b: int, c: bool):
    pass


**Suggestion 4: create type alias to indicate where the parameters will be found**

# somewhere on FastAPI codebase:
TPath = typing.Union
TBody = typing.Union
TQuery = typing.Union
THeader = typing.Union
TCookie = typing.Union

@app.post('/')
async def test(a: TBody[str], b: TPath[int], c: TQuery[bool]):
    pass

PR #906, opened 2020-01-22
OpenAPI schema generation: Don't convert all enums into str

songzhi:
Before doing this,the type of  values in `enum` could be at odds with `type` field.


tiangolo:
Thanks! Good point :mag: :rocket: I updated it to allow any type as subtype of `List`.

Issue #884, opened 2020-01-17
Allow json/BaseModels in query parameters


luckydonald:
### Is your feature request related to a problem

I have to support getting a structured json string as parameter:
<img width="891" alt="Screenshot 2020-01-17 at 22 08 55" src="https://user-images.githubusercontent.com/2737108/72646249-f86be980-3975-11ea-9e3c-3ea7816fd1e1.png">

```url
https://localhost:8080/test?moop={"name": "something %26 something else","description": "str" ,"price": 0.2,"tax": 0.2}&foo=bar
```

I would like to have a easy validation for those fields as well, as this could be automatic

### The solution you would like

```python
@routes.api_route('/test', methods=['GET', 'POST'], tags=['official', 'message', 'send'])
async def test(
    moop: Json[Item] = Query(..., description='Something something'),
    foo: str = Query(...),
):
    return {'woop': moop, 'lol': foo}
# end def
```

<details><summary>Initial version:</summary>

This was the initial thought, before I discovered the `Json` type.
```python
@routes.api_route('/test', methods=['GET', 'POST'], tags=['official', 'message', 'send'])
async def test(
    moop: Item = Query(..., description='Something something'),
    foo: str = Query(...),
):
    return {'woop': moop, 'lol': foo}
# end def
```

</details>

### Bug you face

```py
AssertionError: Param: moop can only be a request body, using Body(...)
```

### Describe alternatives you've considered

```py
    moop: str
```

luckydonald:
Funny, using `moop: Json` is working:     

```py
from pydantic.types import Json

@routes.api_route('/test', methods=['GET', 'POST'], tags=['official', 'message', 'send'])
async def test(
    moop: Json = Query(..., description='Something something'),
    foo: str = Query(...),
):
  moop = Item(moop)
```

but  `moop: Json[Item]` is not working:

```py
@routes.api_route('/test', methods=['GET', 'POST'], tags=['official', 'message', 'send'])
async def test(
    moop: Json[Item] = Query(..., description='Something something'),
    foo: str = Query(...),
):
    moop = Item(moop)
    return {'woop': moop, 'lol': foo}
# end def
```
is failing with:
```py
File "/path/to/project/main.py", line 482, in <module>
  from .views.api.v4_5.sendable import routes as sendable_routes
File "/path/to/project/views/api/v4_5/sendable.py", line 85, in <module>
  foo: str = Query(...),
[...]

PR #908, opened 2020-01-22
Allow `Json` type for `Query`, `Path` and `Header`.

luckydonald:
See #884 for details.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -32,6 +32,7 @@ from pydantic import BaseConfig, BaseModel, create_model
 from pydantic.error_wrappers import ErrorWrapper
 from pydantic.errors import MissingError
 from pydantic.utils import lenient_issubclass
+from pydantic.types import JsonWrapper
 from starlette.background import BackgroundTasks
 from starlette.concurrency import run_in_threadpool
 from starlette.datastructures import FormData, Headers, QueryParams, UploadFile
@@ -297,6 +298,11 @@ def get_dependant(
         elif isinstance(
             param.default, (params.Query, params.Header)
         ) and is_scalar_sequence_field(param_field):
+            add_param_to_fields(field=param_field, dependant=dependant)
+        elif isinstance(
+            param.default, (params.Query, params.Header, params.Path)
+        ) and issubclass(param_field.outer_type_, JsonWrapper):  # maybe `param.annotation` instead of `param_field.outer_type_`?
+            # handle (accept) `foobar: Json[Model] = Query(…)
             add_param_to_fields(field=param_field, dependant=dependant)
         else:
             field_info = get_field_info(param_field)
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -316,6 +316,12 @@ def get_dependant(
 def add_non_field_param_to_dependency(
     *, param: inspect.Parameter, dependant: Dependant
 ) -> Optional[bool]:
+    """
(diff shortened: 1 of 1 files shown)


fastapi/dependencies/utils.py
```diff
@@ -276,7 +277,7 @@ def get_dependant(
         param_field = get_param_field(
             param=param, default_field_info=params.Query, param_name=param_name
         )
-        if param_name in path_param_names:
+        if param_name in path_param_names and not issubclass(param.annotation, JsonWrapper):
```
dmontagu: I defer to @tiangolo, but I personally don't think we should add special handling for `JsonWrapper` here or below.

I don't believe this is a performance sensitive part of the codebase, but adding special handling for this type:

1. Adds an on-going maintenance burden where we have to have tests for `JsonWrapper` and non-`JsonWrapper` code paths
2. Adds an on-going maintenance burden where we need to handle this edge case any time we may want to refactor this code in the future
3. Opens the door for requests for special handling of other types in the future

Moreover, because pydantic now runs its tests against FastAPI, it could potentially make it more difficult for pydantic to refactor the `Json` class in the future.

I could be convinced otherwise if this was a common request, but so far that doesn't seem to be the case.

There are potentially alternative approaches to make this work though; I'll comment on them elsewhere.

fastapi/dependencies/utils.py
```diff
@@ -276,7 +277,7 @@ def get_dependant(
...
             assert is_scalar_field(
```
dmontagu: I think it would be better if we could rework this function and/or code path (and the corresponding one below) in such a way that `Json[X]` fields would be handled properly.

In particular, it would be better if we could test a different (closely-related) condition that didn't require specifically checking for `Json[X]`-based field types. I don't know if such a condition exists or could be easily made to exist though.

-----

It may also just make sense to loosen the restriction on what is allowed as a path/query/header-parameter type; I defer to @tiangolo about that.

------

Another preferable approach, in my opinion, would be to just change pydantic in such a way that `Json[X]` types pass the check you want to perform. I haven't thought enough about the `Json` type to know whether there is a reasonable change that could be made here.

Assuming there *is* a reasonable change, depending on what is required, it may be a breaking change requiring waiting until pydantic v2. But if such a change is possible (whether reasonable or not), there *may* be a simple way to accomplish your goal by subclassing the `Json` class and changing an attribute (you could call the subclass `QueryJson` or similar).

luckydonald: Is @tiangolo maintainer there as well which could make an change to pydantic for fastapi realistic?
In any case I'm not sure if Pydantic is the correct place needing a change.
After all Pydantic is handling the Json type very well, decoding json, and reading it into lists, dicts and if specified even into real models just fine.
Currently we are throwing away that capability by limiting it to just the Body as source type.

It's just FastAPI adding too restrictive code on top of pydantic, therefore I made an exception for the case we want to handle here, Json.

It might be possible to maybe loosen the existing code in a more general way, however I don't really understand the existing code _(a few comments here and there would be really nice!)_ to make any in depth changes.

tiangolo:
Thanks @luckydonald for your contribution and interest. :cake: 

But I think this would be better handled in a dependency that parses the incoming string as @dmontagu suggested in the issue. That should solve it without needing extra logic in FastAPI for that specific use case. Also, the default encoding for query parameters is `form`, so this could end up misguiding users as it wouldn't do all the extra logic necessary to document it as a custom encoding in OpenAPI. As it is more of a corner case, I think the best approach is to use a dependency that does the JSON decoding from the string in the query.

Thanks @dmontagu for the review and detailed analysis! :rocket:

PR #918, opened 2020-01-25
Fix body parsing

### LINKED ISSUE #914: [BUG] Sending incorrect data cause 500 error
I've got this simple example:

```
from typing import List

class User(BaseModel):
    name: str


@app.post('/test/')
async def test(users: List[User], test: str = Body(...)):
    return {'users': users, 'test': test}
```

When sending incorrect payload the server returns 500 error:

```
# curl -s -D - -o /dev/null -X POST "localhost:8080/test/" -H  "accept: application/json" -H  "Content-Type: application/json" -d "[]"
HTTP/1.1 500 Internal Server Error
date: Fri, 24 Jan 2020 10:59:47 GMT
server: uvicorn
content-length: 1264
content-type: text/plain; charset=utf-8
```
Server exception:

```
INFO:     127.0.0.1:60652 - "POST /test/ HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
  File "/test/venv38/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 385, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/test/venv38/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
  File "/test/venv38/lib/python3.8/site-packages/fastapi/applications.py", line 140, in __call__
  File "/test/venv38/lib/python3.8/site-packages/starlette/applications.py", line 134, in __call__
    await self.error_middleware(scope, receive, send)
  File "/test/venv38/lib/python3.8/site-packages/starlette/middleware/errors.py", line 178, in __call__
    raise exc from None
  File "/test/venv38/lib/python3.8/site-packages/starlette/middleware/errors.py", line 156, in __call__
    await self.app(scope, receive, _send)
  File "/test/venv38/lib/python3.8/site-packages/starlette/exceptions.py", line 73, in __call__
    raise exc from None
  File "/test/venv38/lib/python3.8/site-packages/starlette/exceptions.py", line 62, in __call__
    await self.app(scope, receive, sender)
  File "/test/venv38/lib/python3.8/site-packages/starlette/routing.py", line 590, in __call__
    await route(scope, receive, send)
  File "/test/venv38/lib/python3.8/site-packages/starlette/routing.py", line 208, in __call__
  File "/test/venv38/lib/python3.8/site-packages/starlette/routing.py", line 41, in app
    response = await func(request)
  File "/test/venv38/lib/python3.8/site-packages/fastapi/routing.py", line 115, in app
  File "/test/venv38/lib/python3.8/site-packages/fastapi/dependencies/utils.py", line 547, in solve_dependencies
    ) = await request_body_to_args(  # body_params checked above
  File "/test/venv38/lib/python3.8/site-packages/fastapi/dependencies/utils.py", line 637, in request_body_to_args
    value = received_body.get(field.alias)
AttributeError: 'list' object has no attribute 'get'
```

So the body is not validated in this case?

dmontagu:
Closes #914 

I believe this is the "correct" way to fix this issue, as the provided body could be invalid in a number of ways, and this should handle anything that json could get parsed into. (The error message also seems appropriate to me even for malformed bodies.)

There was actually already a test for this, but it only checked for a (malformed) body of `None`, not `[]`, so it was easy to add a test case as another parametric case.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -634,7 +634,11 @@ async def request_body_to_args(
                 ) and isinstance(received_body, FormData):
                     value = received_body.getlist(field.alias)
                 else:
-                    value = received_body.get(field.alias)
+                    try:
+                        value = received_body.get(field.alias)
+                    except AttributeError:
+                        errors.append(get_missing_field_error(field.alias))
+                        continue
             if (
                 value is None
                 or (isinstance(field_info, params.Form) and value == "")
@@ -645,18 +649,7 @@ async def request_body_to_args(
                 )
             ):
                 if field.required:
-                    if PYDANTIC_1:
-                        errors.append(
-                            ErrorWrapper(MissingError(), loc=("body", field.alias))
-                        )
-                    else:  # pragma: nocover
-                        errors.append(
-                            ErrorWrapper(  # type: ignore
-                                MissingError(),
-                                loc=("body", field.alias),
-                                config=BaseConfig,
-                            )
-                        )
+                    errors.append(get_missing_field_error(field.alias))
                 else:
                     values[field.name] = deepcopy(field.default)
                 continue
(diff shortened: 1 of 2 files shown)


fastapi/dependencies/utils.py
```diff
@@ -685,6 +678,16 @@ def request_params_to_args(
     return values, errors
 
 
+def get_missing_field_error(field_alias: str) -> ErrorWrapper:
```
dmontagu: Since this error is now used in two places, I added a utility function to generate it.

tiangolo:
Awesome, thanks @dmontagu ! :rocket: :cake: 

And thanks for the clear explanation :nerd_face: :taco:

PR #930, opened 2020-01-29
remove mutable default values from functions args

includeamin:


tiangolo:
Thanks for your contribution and interest @includeamin :cake: 

But as those mutable default values are actually never mutated and only read, it's not really a problem here.

PR #978, opened 2020-02-13
Fix path encoding

patrickmckenna:
This PR adds a case to `jsonable_encoder` so that `pathlib.PurePath` objects are serializable. Related to https://github.com/tiangolo/fastapi/issues/563.

I know there was some discussion in that thread about how to handle the OS-dependent `str` representations of `pathlib.Path` objects. Perhaps I'm misunderstanding something, but shouldn't users who need OS-agnostic treatment of paths be using `pathlib.Pure*Path` types to get `str` representations in their native form, no matter where their code is running?

/cc @devtud

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -1,4 +1,5 @@
 from enum import Enum
+from pathlib import PurePath
 from types import GeneratorType
 from typing import Any, Callable, Dict, List, Set, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


tiangolo:
Awesome! Thank you for your contribution! :rocket: :cake:

sevetseh28:
This doesn' work with the latest version of Pydantic. I'm getting:
`TypeError: Object of type PosixPath is not JSON serializable` for `Path` fields

PR #987, opened 2020-02-16
Move all mypy configuration to configuration file

hukkin:
- ~~Remove unused `# type: ignore`s~~
- ~~Make mypy complain about unused `# type: ignore`s~~
- Move all mypy configuration to configuration file (a command line flag was used in `scripts/lint.sh`)

### DIFF
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -18,7 +18,7 @@ def _fake_asynccontextmanager(func: Callable) -> Callable:
 
 
 try:
-    from contextlib import asynccontextmanager
+    from contextlib import asynccontextmanager  # type: ignore
 except ImportError:
     try:
         from async_generator import asynccontextmanager  # type: ignore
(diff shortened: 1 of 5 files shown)


hukkinj1:
Ah, different Python versions require different `# type: ignore`s it seems. We can either:
- Close this PR
- Make Travis only run mypy with one Python version (I suggest the lowest supported version).

tiangolo (changes requested):
The `ignore`s make it work in both Python 3.6 and 3.7+. And that also makes it better not to have `warn_unused_ignores = True`, or it would complain in one ignores for 3.6 and the others for 3.7.

But I think it's true we could have all the `mypy` configs in a single place instead of file and command line. Could you please remove the other changes and just leave the `disallow_untyped_defs = True` in the `mypy.ini`, and remove it from the command?

hukkinj1:
@tiangolo :+1: done

tiangolo:
Perfect, thank you! :rocket: :wrench:

Issue #365, opened 2019-07-04
[BUG]  Tutorial websocket doc example


Ben621:
**Describe the bug**
Hi, 
On the docs of websocket the last example doesn't work.

**To Reproduce**
1. Create a file main.py with the last example on the bottom of the file

>https://fastapi.tiangolo.com/tutorial/websockets/#create-a-websocket
```python
from fastapi import Cookie, Depends, FastAPI, Header
from starlette.responses import HTMLResponse
from starlette.status import WS_1008_POLICY_VIOLATION
from starlette.websockets import WebSocket

app = FastAPI()

html = """
<!DOCTYPE html>
<html>
    <head>
        <title>Chat</title>
    </head>
    <body>
        <h1>WebSocket Chat</h1>
        <form action="" onsubmit="sendMessage(event)">
            <label>Item ID: <input type="text" id="itemId" autocomplete="off" value="foo"/></label>
            <button onclick="connect(event)">Connect</button>
            <br>
            <label>Message: <input type="text" id="messageText" autocomplete="off"/></label>
            <button>Send</button>
        </form>
        <ul id='messages'>
        </ul>
        <script>
        var ws = null;
            function connect(event) {
                var input = document.getElementById("itemId")
                ws = new WebSocket("ws://localhost:8000/items/" + input.value + "/ws");
                ws.onmessage = function(event) {
                    var messages = document.getElementById('messages')
                    var message = document.createElement('li')
                    var content = document.createTextNode(event.data)
                    message.appendChild(content)
                    messages.appendChild(message)
                };
            }
            function sendMessage(event) {
                var input = document.getElementById("messageText")
                ws.send(input.value)
                input.value = ''
                event.preventDefault()
            }
        </script>
    </body>
</html>
"""


@app.get("/")
async def get():
    return HTMLResponse(html)


async def get_cookie_or_client(
    websocket: WebSocket, session: str = Cookie(None), x_client: str = Header(None)
):
    if session is None and x_client is None:
        await websocket.close(code=WS_1008_POLICY_VIOLATION)
    return session or x_client


@app.websocket("/items/{item_id}/ws")
async def websocket_endpoint(
    websocket: WebSocket,
    item_id: int,
    q: str = None,
    cookie_or_client: str = Depends(get_cookie_or_client),
):
    await websocket.accept()
    while True:
        data = await websocket.receive_text()
        await websocket.send_text(
            f"Session Cookie or X-Client Header value is: {cookie_or_client}"
        )
[...]

PR #994, opened 2020-02-18
Fixes #967 and #365

### LINKED ISSUE #967: [BUG] The generated OpenAPI schema duplicates parameter IDs when used with dependencies
### Describe the bug

The generated OpenAPI schema duplicates parameter IDs when used with dependencies.

### To Reproduce

Steps to reproduce the behavior with a minimum self-contained file.

* Create a file with:

```Python

app = FastAPI()


async def user_exists(user_id: int):
    return True


@app.post("/users/{user_id}", dependencies=[Depends(user_exists)])
async def bug(user_id: int):
    pass


client = TestClient(app)

openapi_schema = {
    "openapi": "3.0.2",
    "paths": {
        "/users/{user_id}": {
            "post": {
                "summary": "Bug",
                "operationId": "bug_users__user_id__post",
                "parameters": [
                    {
                        "required": True,
                        "schema": {"title": "User Id", "type": "integer"},
                        "name": "user_id",
                        "in": "path",
                    },
                    # This duplicated user_id shouldn't be here
                    # {
                    #     "required": True,
                    #     "schema": {"title": "User Id", "type": "integer"},
                    #     "name": "user_id",
                    #     "in": "path",
                    # },
                ],
                "responses": {
                    "200": {
                        "content": {"application/json": {"schema": {}}},
                    },
                    "422": {
                        "content": {
                            "application/json": {
                                "schema": {
                                }
                            }
                        },
                    },
                },
            }
        }
    },
    "components": {
        "schemas": {
            "HTTPValidationError": {
                "title": "HTTPValidationError",
                "type": "object",
                "properties": {
                    "detail": {
                        "title": "Detail",
                        "type": "array",
                        "items": {"$ref": "#/components/schemas/ValidationError"},
                    }
                },
            },
            "ValidationError": {
                "title": "ValidationError",
                "type": "object",
                "properties": {
                    "loc": {
                        "title": "Location",
                        "type": "array",
                        "items": {"type": "string"},
                    },
                    "msg": {"title": "Message", "type": "string"},
                    "type": {"title": "Error Type", "type": "string"},
                },
            },
        }
    },
}


def test_reused_param():
    data = response.json()
    assert data == openapi_schema
```

* Run it with `pytest`.

Alternatively, you can run it with Uvicorn go to `/openapi.json`, copy that and validate it at: https://editor.swagger.io/ . It should be valid.

### Expected behavior

The `user_id` used by the path operation and the dependency shouldn't be duplicated in the OpenAPI schema.

### Environment

- OS: all
- FastAPI Version [e.g. 0.3.0], get it with: 0.48.0

```bash
python -c "import fastapi; print(fastapi.__version__)"
```

- Python version, get it with: 3.7

```bash
python --version
```

### Additional context

Supersedes #395

merowinger92:
When generating the openapi.json-parameters list only unique values are used according to their 'name' attribute.

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -180,7 +180,7 @@ def get_openapi_path(
             operation_parameters = get_openapi_operation_parameters(all_route_params)
             parameters.extend(operation_parameters)
             if parameters:
-                operation["parameters"] = parameters
+                operation["parameters"] = list({param['name']:param for param in parameters}.values())
             if method in METHODS_WITH_BODY:
                 request_body_oai = get_openapi_operation_request_body(
                     body_field=route.body_field, model_name_map=model_name_map
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -180,7 +180,9 @@ def get_openapi_path(
             operation_parameters = get_openapi_operation_parameters(all_route_params)
             parameters.extend(operation_parameters)
             if parameters:
-                operation["parameters"] = list({param['name']:param for param in parameters}.values())
+                operation["parameters"] = list(
+                    {param["name"]: param for param in parameters}.values()
+                )
             if method in METHODS_WITH_BODY:
                 request_body_oai = get_openapi_operation_request_body(
                     body_field=route.body_field, model_name_map=model_name_map
--- a/tests/test_param_in_path_and_dependency.py
+++ b/tests/test_param_in_path_and_dependency.py
@@ -1,0 +1,93 @@
+from fastapi import Depends, FastAPI
+from starlette.testclient import TestClient
+
+app = FastAPI()
+
+
+async def user_exists(user_id: int):
+    return True
+
+
+@app.get("/users/{user_id}", dependencies=[Depends(user_exists)])
+async def read_users(user_id: int):
+    pass
+
+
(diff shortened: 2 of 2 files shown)


dmontagu:
@tiangolo looks like something is messed up with the netlify CI for pull requests? I think that might be related to why the codecov didn't hit 100%?

peaceiris:
@dmontagu I opened #1023 to fix that.

tiangolo:
Thanks @merowinger92 ! Good job! :rocket: :cake: 

I added a test to very the fix and make sure we avoid regressions.

@dmontagu Yeah, I think the coverage miss was related to 2 versions of Python (3.6 and >=3.7) with the back ports. Not sure why one of them could not have finished, but now they passed.

At the same time, there was the issue with the Netlify deploys :man_facepalming: : During the last months I was running out of the free build time allowance and they were already charging me. So, to remove that extra cost I added deploys with GitHub actions, but they _can't_ work with PRs from forks (in the straightforward implementation).

But it's sort of solved now. There's more context about the Netlify deploys here #1046 and here #1047 

@peaceiris Thanks! The issue was actually about these deploys not working for PRs from forks, but it's more or less handled in #1046 and #1047 . I'll check your PR later :slightly_smiling_face:

PR #1013, opened 2020-02-20
Add shared tags for routes in APIRouter

jbkoh:
The `APIRoutes` in an `APIRouter` commonly share the same tags. This PR helps in documenting different routes under the same context. For example, this now enables:
```python
data_router = APIRouter(tags=['data'])

@data_router.get('/')
def get_data():
    ...

@data_router.post('/')
def post_data(...):
    ...
```
, instead of:
```python
data_router = APIRouter()

@data_router.get('/', tags=['data'])
def get_data():
    ...

@data_router.post('/', tags=['data'])
def post_data(...):
    ...
```

I can add tests and documentation if the maintainers are fine with this approach.

Thanks!


tiangolo:
Nice! I like it. In fact, I created a couple of other related issues ( #1053 and #1054 ) to extend this further in future PRs.

I think the next step would be to add tests for this.

jbkoh:
@tiangolo Sorry for the late comeback. I am trying to add documentation and tests for this. Is there any recommendation for them? Shall I just add a section inside `/tutorial/bigger-applications?` or just modify the existing "users" example to accommodate common tags?

And, I had not realized there is a way to add tags when you `include_router` like this: https://fastapi.tiangolo.com/tutorial/bigger-applications/#include-an-apirouter-with-a-prefix-tags-responses-and-dependencies
Would you still want to implement the current PR?

Thanks!

euri10:
I remember implementing this and begun doing it at that level also before @tiangolo asked to put it in the include_router.
Maybe the tests wrote on that occasion could be a good start
https://github.com/tiangolo/fastapi/blob/40fe0a8960c5dd32d0a7d1dbbaee253f7372b0ed/tests/test_tags.py

chris-allnutt:
@jbkoh so the use case I could see for allowing the APIRouter to do it is that it makes the APIRouter responsible for telling fastapi it's tags, rather than it being configuration.  The main router should surely be responsible for prefixes and some other options.  My questions to @tiangolo would be if this is accepted, should the old method then be deprecated (initially via warning) otherwise we've got 2 ways to do the same thing which could be confusing.

tiangolo:
Thanks for the patience!

I wanted to include this feature with several other shared parameters. And it all had quite some tricky logic and testing. I just finished adding it in #2434.

So it is now supported in FastAPI version 0.62.0 :tada: 

Given that, I'm gonna close this one. But thanks for the effort! :coffee: :cake:

PR #1015, opened 2020-02-21
Fix markdown code section rendering error in SQL tutorial

vegarsti:
Currently the [Migrations](https://fastapi.tiangolo.com/tutorial/sql-databases/#migrations) section and below is rendering incorrectly because of unused `hl_lines` sections. I think this should fix it.

### DIFF
--- a/docs/advanced/sql-databases-peewee.md
+++ b/docs/advanced/sql-databases-peewee.md
@@ -429,31 +429,31 @@ Repeat the same process with the 10 tabs. This time all of them will wait and yo
 
 * `sql_app/database.py`:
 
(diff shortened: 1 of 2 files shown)


vegarsti:
@tiangolo Sorry, hope it's okay if I tag you here. Let me know if I can do anything more

tiangolo:
Good catch @vegarsti ! :heavy_check_mark:

I guess it was probably related to a recent version of the docs plugins. I added your fix to the SQL tutorial for Peewee that had the same issue.

Thanks for your contribution :rocket: :cake:

PR #1017, opened 2020-02-22
Admit containers of pydantic models as responses models

patrickmckenna:
This PR addresses https://github.com/tiangolo/fastapi/issues/628.

Fixing the immediate problem only required removing a brittle [runtime type check](https://github.com/tiangolo/fastapi/blob/9c3c9b6e78768374868d690bc05918d58481e880/fastapi/routing.py#L300-L302) of each `model` field in a `responses` arg (passed to [`APIRoute.__init__`](https://github.com/tiangolo/fastapi/blob/9c3c9b6e78768374868d690bc05918d58481e880/fastapi/routing.py#L204)). Putting in a more robust version seemed to lead inexorably to making changes in a few other places.

The same type checking effectively happens in parts of the code that accept a `type` and returns a `pydantic.ModelField`. As it turns out, that logic is repeated in a number of places:
- [Processing any `model` fields](https://github.com/tiangolo/fastapi/blob/9c3c9b6e78768374868d690bc05918d58481e880/fastapi/routing.py#L304-L323) in the `responses` arg to `APIRoute.__init__`
- Doing [the same](https://github.com/tiangolo/fastapi/blob/9c3c9b6e78768374868d690bc05918d58481e880/fastapi/routing.py#L246-L264) for the `response_model` arg to `APIRoute.__init__`
- At various places in `fastapi.dependencies.utils`: [here](https://github.com/tiangolo/fastapi/blob/9c3c9b6e78768374868d690bc05918d58481e880/fastapi/dependencies/utils.py#L365-L388), [here](https://github.com/tiangolo/fastapi/blob/9c3c9b6e78768374868d690bc05918d58481e880/fastapi/dependencies/utils.py#L697-L718), [here](https://github.com/tiangolo/fastapi/blob/9c3c9b6e78768374868d690bc05918d58481e880/fastapi/dependencies/utils.py#L757-L767)

Given that, it seemed best to factor out that logic into a standalone function, for the sake of [DRYness](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). That also makes it easier to raise `fastapi`-specific exceptions, with clearer error messages, when invalid types are passed. And it means that the `model` args in `responses` and `model_field` undergo the same type checking.


tests/test_apiroute.py
```diff
@@ -0,0 +1,39 @@
+from typing import List
```
patrickmckenna: I must confess a little confusion with the test layout. Files seem to be named after possible test scenarios, but there are often multiple files testing the same scenario (e.g. `tests/test_additional_response*`). This also means the unit tests for a given `class` are often spread over many files. Given that, I wasn't sure where to add the new tests, so settled for creating new files. I'm happy to rearrange them, please just LMK how.

tests/test_fastapi.py
```diff
@@ -0,0 +1,37 @@
...
+    ids=["int", "List[int]", "pydantic.BaseModel", "List[pydantic.BaseModel]"],
+)
+def fixture_valid_responses(request):
+    return {"500": {"model": request.param}}
+
+
+def test_additional_responses_documented_in_openapi_schema(client, valid_responses):
+    """
+    The only stated guarantee w/ additional responses is that they'll be included in the
+    auto-generated OpenAPI docs -- so test that guarantee.
```
patrickmckenna: I _think_ this is true, based on [the docs](https://github.com/tiangolo/fastapi/blob/9c3c9b6e78768374868d690bc05918d58481e880/docs/advanced/additional-responses.md#additional-response-with-model):

> Those additional responses will be included in the OpenAPI schema, so they will also appear in the API docs.

patrickmckenna (commented):
@tiangolo a few questions/requests for feedback.

tests/test_apiroute.py
tiangolo: Yeah, the main tests are for the examples in the docs. And then there are extra tests to cover what is not already covered in those. So, it's indeed quite mixed. But it's fine, I just updated them a bit :slightly_smiling_face:

tests/test_fastapi.py
tiangolo: Yep, that's correct. I just updated it a bit to keep the style of the tests consistent with the tests in other places.

tiangolo:
This is great! :clap: :clap: :tada: 

Very smart implementation and refactor. :rocket: :nerd_face: 

I just updated the tests to simplify them a bit and to keep the "style" consistent with the other tests.

Thanks a lot for your contribution! :cake: :clap: :bowing_man:

PR #1023, opened 2020-02-22
Add paths to trigger by docs changes

peaceiris:
The title says it all.


tiangolo:
Thanks! The issue with Netlify was actually related to the fact that PRs from forks don't have access to secrets. This was solved by #1046 and #1047 .

If this could be solved in a different way later I think it would be nice to be able to have docs previews for all PRs, possibly even if they don't modify the docs.

So I'm gonna pass this for now. But thanks for your contribution and interest! :cake:

Issue #820, opened 2019-12-27
[FEATURE] remove ujson dependency (it is dead)


hnykda:
FastAPI is using `ujson` which has been not actively developed for about 3 years by now. [Here is a relevant issue](https://github.com/esnme/ultrajson/issues/343)

dmontagu:
I would be in favor of recommending `orjson` instead of `ujson`.

I recall the creator of `orjson` reaching about this many months ago and there was some confusion due to slow responses, and also some issues on the pydantic side, but I think many/most of the barriers to usage have since been removed. If `orjson` seems to be taking over as the go-to high-performance json library, I think it would make sense to use that for development and recommend it.

To the extent that there is any difference in json encoding/decoding between json and orjson (or ujson and orjson), it would probably be best to document it.

I'd be willing to review a PR making this change. (I'd leave the ultimate decision about merging to @tiangolo though.)

tiangolo:
Yeah, I've been intending to do it for quite some time now... :sweat_smile: 

In fact, you can sort of do it right now (and since several versions ago). 

---

Here's my plan:

I want to export everything from Starlette, to avoid confusions of the type "do I import this from FastAPI or from Starlette?", I've seen reports of that and even I've struggled with that. For example, FastAPI has its own `HTTPException`.

With that, I want to have a response `ORJSONResponse` exported from `fastapi.responses`, among Starlette's responses.

In fact, I added tests for creating custom responses (as part of a PR), testing specifically creating orjson responses: https://github.com/tiangolo/fastapi/blob/f803c77515662cd2382674d9fc8df1e6641b3ba7/tests/test_default_response_class.py#L9-L13
And you can replicate creating that response class and using it. It's just 3 lines (you would remove the custom `media_type`), but still, it should be provided by FastAPI (among all the other responses from Starlette).

```Python
app = FastAPI(default_response_class=ORJSONResponse)
```

---

Also, as a side note, UJSON is not really used by default anywhere. It's optional, and you would have to use the `UJSONResponse` class for it to be used, the same way you would use the (new/custom/future) `ORJSONResponse`.

hnykda:
Well, that's elegant :clap: .

> Also, as a side note, UJSON is not really used by default anywhere. It's optional, and you would have to use the UJSONResponse class for it to be used, the same way you would use the (new/custom/future) ORJSONResponse.
[...]

PR #1026, opened 2020-02-24
Added missing unit test dependencies to pyproject.toml

sm-Fifteen:
The [contribution instructions from the doc](https://fastapi.tiangolo.com/contributing/) led to a number of unit tests failing on against a clean venv because of a few missing dependencies. I've read discussions regarding the possible removal of ujson (#820) and aiofiles (encode/starlette#480, encode/starlette#819) as dependencies, so I'm not entirely sure whether we're going to be keeping those in the future but, for the time being, their absence causes the test suite to fail.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,10 @@ test = [
     "databases[sqlite]",
     "orjson",
     "async_exit_stack",
-    "async_generator"
+    "async_generator",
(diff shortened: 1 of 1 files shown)


tiangolo:
Yep, thanks for your contribution @sm-Fifteen ! :rocket: :cake:

PR #1027, opened 2020-02-24
Moved endpoint execution outside of get_request_handler for profiling

sm-Fifteen:
At the reccomendation of @tiangolo in https://github.com/tiangolo/fastapi/issues/701#issuecomment-585319420, I have moved the endpoint execution section from `routing.get_request_handler.<locals>.app` to a separate function (currently named `_run_endpoint`) in order to facilitate profiling. Internal functions and callable object properties make it difficult for profilers to identify the function object being called, and since each request is only associated to one endpoint (not counting dependencies), this makes possible to simply profile "the endpoit that was used" for any given request.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -87,12 +87,14 @@ async def serialize_response(
         return jsonable_encoder(response)
 
 
-async def _run_endpoint(dependant: Dependant, values: Dict[str, Any]) -> Any:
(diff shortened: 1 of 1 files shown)


tiangolo:
Great! I updated it a bit to re-use the precomputed `is_coroutine`.

Thanks for your contribution! :rocket: :tada: :cake:

PR #1038, opened 2020-02-25
Docs/cache

alvarogf97:
PR for #1033

### DIFF
--- a/docs/src/cache/tutorial001.py
+++ b/docs/src/cache/tutorial001.py
@@ -1,0 +1,28 @@
+from aiocache import caches, cached
+from fastapi import FastAPI
+
+
+CACHE_TTL=2
+
+
+caches.set_config({
+    'default': {
+        'cache': 'aiocache.RedisCache',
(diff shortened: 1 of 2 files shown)


alvarogf97:
@euri10

tiangolo:
Hi @alvarogf97 , thanks for your contribution! :nerd_face: 

The problem with this approach is that it would not have into account dependencies, possibly handling authentication, etc.

I think it could be useful for some specific use cases that don't have dependencies and that probably return the same data for the same parameters in the request. For example, requests that don't change depending on current user logged in, other dependencies, changes in DB, etc.

Because of that, I would prefer not to have it in the docs like that as it would imply that it could "just work" for all the use cases while it's a delicate and possibly dangerous optimization. So I'm going to pass on this. But thanks for your interest and contribution! :tada: :cake: 

Now that the conversation is in GitHub, I'm pretty sure that other users that really need that type of optimization will be able to find it there. :memo:

PR #1046, opened 2020-02-27
Do not deploy netlify on external prs

mariacamilagl:
Do not deploy netlify on external prs. As requested by @tiangolo

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,6 +1,5 @@
 name: Build and Deploy to Netlify
-on:
-  push:
+on: [push]
 jobs:
   build:
     runs-on: ubuntu-18.04
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,6 +1,8 @@
 name: Build and Deploy to Netlify
 on:
   push:
+  pull_request:
+    types: [opened, synchronize]
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks! :+1:

tiangolo:
Adding more context about this, FastAPI has been using more Netlify building minutes than the free quota, and Netlify has been charging me for that.

To avoid that cost I set up Netlify deployments using GitHub actions: https://github.com/tiangolo/fastapi/pull/983

To deploy, it needs Netlify credentials that are passed as environment variables to the GitHub action. The credentials are stored in GitHub secrets in the repo, but those secrets are not passed to the actions triggered by PRs from forks: https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/td-p/30678

So, currently, there's no way to deploy a preview to Netlify using a simple action like that. Then I modified the action to avoid exiting with an error: https://github.com/nwtgck/actions-netlify/pull/33

And that's the action temporarily used here.

A better future (but more involved) alternative, at least that I can think of now, is to create an action that runs periodically, like https://github.com/tiangolo/issue-manager, and that checks for each open PR, and checks if it already wrote a comment with a Netlify deployment for the latest commit of that PR. If not, it would pull that PR, build it, deploy to Netlify, and add a comment with the deployment. This version would work because it's always running from `master`, not from a branch in a fork. But it wouldn't be able to use the default triggers (`on pull_request`, etc).

Issue #1045, opened 2020-02-27
[BUG] Broken link to GraphQL tutorial


e-ruiz:
### Describe the bug

It's a documentation bug, not a software bug :wink: 

Unfortunately the link for GraphQL tutorial is broken: https://fastapi.tiangolo.com/tutorial/graphql/ 

References: 
- https://fastapi.tiangolo.com/release-notes/#061
- PR #48 

Best regards.

tiangolo:
Thanks! This was fixed by @sattosan in #1052

e-ruiz:
Links mentioned working perfectly.

Closing issue.

PR #1052, opened 2020-02-29
Fix broken link in release notes

sattosan:
Fixes issue #1045 and other broken link in release notes.


tiangolo:
Great! Thanks for checking all that! :mag: :male_detective: :rocket: :cake:

PR #1059, opened 2020-03-01
convert json schema const to enum for openapi

jashandeep-sohi:
OpenAPI does not support `const`, but JSON schema output by pydantic
can include it. `const` is just syntactic sugar for a single element enum,
so let's convert it instead of just dropping it.

ref: https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values

### DIFF
--- a/fastapi/openapi/models.py
+++ b/fastapi/openapi/models.py
@@ -2,7 +2,7 @@ from enum import Enum
 from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 
 from fastapi.logger import logger
-from pydantic import BaseModel
+from pydantic import BaseModel, root_validator
 
 try:
     from pydantic import AnyUrl, Field
@@ -127,6 +127,19 @@ class SchemaBase(BaseModel):
     externalDocs: Optional[ExternalDocumentation] = None
     example: Optional[Any] = None
     deprecated: Optional[bool] = None
+
+    @root_validator(pre=True)
+    def convert_const_to_enum(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        """OpenAPI does not support `const`, but JSON schema output by pydantic
+        can include it. `const` is just syntactic sugar for a single element enum,
+        so let's convert it instead of just dropping it.
+
+        ref: https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values
+        """
+        if "const" in values and "enum" not in values:
+            values["enum"] = [values["const"]]
+
+        return values
 
 
 class Schema(SchemaBase):
--- a/tests/test_openapi.py
+++ b/tests/test_openapi.py
@@ -1,0 +1,19 @@
+from fastapi import FastAPI
+from pydantic import BaseModel, Field
+
+
+def test_enum_const_jsonschema_conversion():
+    app = FastAPI()
+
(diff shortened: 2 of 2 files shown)


tiangolo:
Thanks for your contribution @jashandeep-sohi ! :nerd_face: :rocket: 

The problem is that `root_validator` is only available in Pydantic `1.x`.

FastAPI is still backwards compatible with Pydantic `0.32.x` and there are still too many downloads for those versions (about 10 times more than for 1.x), so I can't drop support for Pydantic < 1.x yet.

---

On the other side, OpenAPI is based on JSON Schema Draft `00`, which is now a bit old. There are a couple of other small inconsistencies between newer JSON Schema versions (provided by Pydantic) and OpenAPI's schemas, but having schemas declared with modern JSON Schema allows some tools to use the generated JSON Schema directly, like https://github.com/rjsf-team/react-jsonschema-form . So, I have preferred to keep the generated recent JSON Schema versions even though there are a couple of inconsistencies in some specific corner cases.

And on the other side, if strict OpenAPI compliance is required, it's also easy to update the Pydantic model to use an enum of a single value instead of a `Literal`. That's also backwards compatible with other versions of Python, given `Literal` is only in Python 3.8.

So I'll pass on this one. But thanks for your contribution and interest! :cake:

cybojenix:
For anyone stumbling on this, you can implement using the `schema_extra` within your base model (as eluded to in Tiangolo's comment):

```py

class MyBaseModel(BaseModel):
    class Config:
        @staticmethod
        def schema_extra(schema, model) -> None:
            for name, prop in schema.get('properties', {}).items():
                if 'const' in prop:
                    prop['enum'] = [prop['const']]
```

also note that this will override the enum, to support the usecase of using a Literal to reference an Enum.

Tested on Pydantic==1.5.1, FastAPI==0.54.1

dvarrazzo:
I have solved the problem of including consts another way in a program of mine. I think this is compatible with older versions of Pydantic and renders well with the current OpenAPI version. Pretty much, if a `Field("value", const=True)` is defined, then `{"const": "value", "example": "value"}` is added to the openapi json, and the schema and example for the types will look correct.

If there is interest in the feature (I think it's badly missing) I can propose a MR doing this in the proper place.

Reference: https://github.com/OAI/OpenAPI-Specification/issues/1666 

```python
class FastAPISubclass(FastAPI):
    def openapi(self) -> Dict[str, Any]:
        if self.openapi_schema:
            return self.openapi_schema

        self.openapi_schema = schema = get_openapi(
            # ...
        )

        # Find the fields with a const value
        for model in self._iter_models():
            for field in model.__fields__.values():
                if not field.field_info.const:
                    continue
                try:
                    props = schema["components"]["schemas"][model.__name__]
                    props = props["properties"][field.name]
                except KeyError:
                    continue
                props["const"] = props["example"] = field.default

        return schema

    def _iter_models(self):
        seen = set()
        for route in self.router.routes:
            model = getattr(route, "response_model", None)
            if hasattr(model, "__fields__"):
                if model not in seen:
                    yield model
                    seen.add(model)

            if getattr(route, "body_field", None) is not None:
                model = route.body_field.type_
                if model not in seen:
                    yield model
                    seen.add(model)
```

PR #1074, opened 2020-03-03
Fix exclude_unset and aliases in response model validation.

ghost:
Should fix both https://github.com/tiangolo/fastapi/issues/1043 and https://github.com/tiangolo/fastapi/issues/1063, but might come with some performance penalty since unnecessary serialization is being done. Better solutions that pass the given tests are more than welcome, but I don't see a much better way other than recursion in cases where there are models inside basic data structures like lists and dicts.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -48,13 +48,15 @@ except ImportError:  # pragma: nocover
     from pydantic.fields import Field as ModelField  # type: ignore
 
 
-def _prepare_response_content(res: Any, *, exclude_unset: bool) -> Any:
+def _prepare_response_content(
+    res: Any, *, by_alias: bool = True, exclude_unset: bool
+) -> Any:
     if isinstance(res, BaseModel):
         if PYDANTIC_1:
-            return res.dict(by_alias=True, exclude_unset=exclude_unset)
+            return res.dict(by_alias=by_alias, exclude_unset=exclude_unset)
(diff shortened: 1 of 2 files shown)


fastapi/routing.py
```diff
@@ -48,6 +48,26 @@
...
+    if isinstance(res, list):
+        return [
+            _prepare_response_content(item, exclude_unset=exclude_unset) for item in res
+        ]
+    elif isinstance(res, dict):
+        return {
+            k: _prepare_response_content(v, exclude_unset=exclude_unset)
+            for k, v in res.items()
+        }
+    elif isinstance(res, BaseModel):
```
Kilo59: Is there any slight performance benefit of making this the first if condition?
Since this is likely (?) going to be the most common case.

juhovh-aiven: I actually thought about this right after pushing the PR, pushed some changes that make this the first condition (kept list as second, since I believe that is quite likely to be the second most common option). There's still the unnecessary recursive iteration for lists and dicts that do not contain any models at all, but it's hard to know if there are models without going through everything...

Kilo59 (approved):

tiangolo:
Excellent job @juhovh-aiven ! :clap: :nerd_face: 

And thanks @Kilo59 for the review :nerd_face: :male_detective: :mag:

PR #1079, opened 2020-03-04
Handle automatic embedding with Depends

Toad2186:
Two issues:
1) Whether or not to embed automatically wasn't taking into account the fact that `flat_dependant.body_params` could have repeated fields with the same name (Such as from a dependency chain). Fix: Look at unique names in `flat_dependant.body_params`, and do not embed if there's only one unique name and embed is false.
2) When processing the param body, automatically embedded bodies were sometimes processed incorrectly because we look at `Depends` parameters one-by-one. Ideally we look at ALL body parameters, but this doesn't seem to be available at the time when we process dependency bodies. The fix here is to set the `embed` parameter on field info for all params. (Not sure if there's a better fix).

TODO:
This case is undefined behavior and currently doesn't throw an error but perhaps it should? (Multiple types for the same body)
```
from fastapi import FastAPI, Depends, Body

app = FastAPI(title='test')

class Item(BaseModel):
    data: str

def dependency(item: str):
    return item

@app.post("/no-duplicates")
async def no_duplicates(item: Item, item2: Item = Depends(dependency)):
    return [item, item2]
```

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -703,9 +703,22 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
         return None
     first_param = flat_dependant.body_params[0]
     field_info = get_field_info(first_param)
-    embed = getattr(field_info, "embed", None)
-    if len(flat_dependant.body_params) == 1 and not embed:
+    should_not_embed = not getattr(field_info, "embed", None)
+    if should_not_embed:
+        if len(flat_dependant.body_params) == 1:
+            should_not_embed = True
+        elif len(flat_dependant.body_params) == 2:
+            should_not_embed = (
+                flat_dependant.body_params[0].name == flat_dependant.body_params[1].name
+            )
+        else:
+            should_not_embed = False
+
+    if should_not_embed:
         return get_schema_compatible_field(field=first_param)
+    # If we embed, then internally ensure that all fields are embedded
+    for param in flat_dependant.body_params:
+        setattr(get_field_info(param), "embed", True)
     model_name = "Body_" + name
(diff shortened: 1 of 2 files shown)


chbndrhnns:
This change fixed the issue I am experiencing, @Toad2186 .

chbndrhnns:
@tiangolo I would appreciate this fix to be merged for the next release. Due to the issue, FastAPI currently produces an incorrect api spec.

Rikeshpatel4u:
Please help me. I want to build a boat. Which posts the news paper in my channel in automatic pdf format every day. Every morning when the paper is placed on the news paper's website, its pdf file is automatically posted to my channel.

tiangolo:

I updated it a bit to include handling sub-dependencies and more than 2 body fields.

Merged! :tada:

acnebs:
@tiangolo this broke my use-case. I have a payload like this:

```json
{
  "field1": "foo",
  "field2": "bar",
  "dep_field": "baz"
}
```

And I've been processing it like this:

```python
from fastapi import FastAPI, Depends, Body

app = FastAPI(title='test')

class Form1(BaseModel):
    field1: str
    field2: str

def require_dep_field(dep_field: str = Body(..., embed=True)):
    return dep_field

@app.post("/form-test")
async def form_test(form: Form1, dep_field: str = Depends(require_dep_field)):
    print(form)
    print(dep_field)
    return "ok"
```

Before, this worked just fine. After this change, I now get an error that `form` is a required field.

AFAICT, I can no longer do this. Am I missing something?

tiangolo:
@acnebs if you are still having this problem, please create a new issue for me to be able to check it there as I would lose sight of it here.

PR #1080, opened 2020-03-04
Export startlette status as a var instead of directly from startlette

Toad2186:
I'm not sure if this is worth doing but it shouldn't cause any harm(?).

Doing the import like this tricks PyCharm into showing `fastapi.status` instead of `startlette.status`...
Screen Shot 2020-03-04 at 12 33 14 AM

### DIFF
--- a/docs/src/wsgi/tutorial001.py
+++ b/docs/src/wsgi/tutorial001.py
@@ -1,6 +1,6 @@
-from flask import Flask, escape, request
 from fastapi import FastAPI
 from fastapi.middleware.wsgi import WSGIMiddleware
+from flask import Flask, escape, request
 
 flask_app = Flask(__name__)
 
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks for your suggestion! :cake: 

I like that it improves editor support, but I'm not totally convinced by the fact that it's kind of a hack. But the main problem is that it wouldn't make sense to do the trick for this and not do it for all the other variables, like Response, Request, the middlewares, etc.

Nevertheless, in the end, it's actually correct, it's the same `status` variable, so there's no harm in using the `status` variable directly from Starlette.

The good thing is that when using things that are exclusive to FastAPI, like `ORJSONResponse`, it knows where to import them from. And when using `HTTPException`, that FastAPI has a custom modified version, it is suggested as well, so I think it's safe to leave it as is.

But thanks for checking it and finding out this behavior! It's interesting to know how that works underneath :nerd_face:

PR #1091, opened 2020-03-06
del doubled text

dvska:


tiangolo:
Thanks for the suggestion and interest.

But I'll pass on this, I think I like the emphasis there. :man_shrugging:

PR #1092, opened 2020-03-06
Allow subclasses of known types to be encoded with superclass encoder

StephenBrown2:
Implementation of https://github.com/samuelcolvin/pydantic/pull/1291 which was closed due to performance concerns, another option was attempted but found to be 50% slower in a related issue here: https://github.com/samuelcolvin/pydantic/issues/1157#issuecomment-572450669

I have not seen the same results in benchmarking pydantic, so I wanted to open this here to see if the performance impact is the same.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -126,15 +126,17 @@ def jsonable_encoder(
         return encoded_list
 
     if custom_encoder:
-        if type(obj) in custom_encoder:
-            return custom_encoder[type(obj)](obj)
+        for base in obj.__class__.__mro__:
+            if base in custom_encoder:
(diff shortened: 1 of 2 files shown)


StephenBrown2:
Closing as it turns out this already happens here https://github.com/tiangolo/fastapi/blob/025b38df40a18f94322fab445aee13024fe783da/fastapi/encoders.py#L132-L134 and here https://github.com/tiangolo/fastapi/blob/025b38df40a18f94322fab445aee13024fe783da/fastapi/encoders.py#L138-L140 with `isinstance`

tiangolo:
Thanks for reporting back and closing it! :cake:

PR #1096, opened 2020-03-08
add example of Pycharm in tutorial/debugging.md

youngquan:
I think many pepole use Pycharm to code, so may be the toturial could contain an example of Pycharm.

Especially, if you don't configure the "Working directory" correctly, you can not debug the code.


tiangolo:
Thanks for your contribution! :memo: :rocket: :cake: 

I just updated it a bit as I moved the location of the docs in the source code, and to have the same code example.

Issue #894, opened 2020-01-20
[BUG] RecursionError from response model in 0.47.1


ysmu:
### Describe the bug

FastAPI 0.47.1 will not be able to start due to a `RecursionError` when there is a circular reference among models. The issue seems to originate from https://github.com/tiangolo/fastapi/pull/889. This works fine in 0.46.0.

### Environment

- OS: Windows
- FastAPI Version: 0.47.1
- Python version: 3.7.0

### To Reproduce

```Python
from typing import Optional

from fastapi import FastAPI
from pydantic import BaseModel, Field


class Group(BaseModel):
    representative: Optional['Person'] = Field(None)


class Person(BaseModel):
    group: Optional[Group] = Field(None)


Group.update_forward_refs()


app = FastAPI()


@app.get('/group/{group_id}', response_model=Group)
def get_group(group_id):
    return []
```

### Expected behavior

No exception

### Actual output

```
Traceback (most recent call last):
  File "test.py", line 21, in <module>
    @app.get('/group/{group_id}', response_model=Group)
  File "D:\virtualenvs\test\lib\site-packages\fastapi\routing.py", line 494, in decorator
    callbacks=callbacks,
  File "D:\virtualenvs\test\lib\site-packages\fastapi\routing.py", line 438, in add_api_route
    callbacks=callbacks,
  File "D:\virtualenvs\test\lib\site-packages\fastapi\routing.py", line 275, in __init__
    ] = create_cloned_field(self.response_field)
  File "D:\virtualenvs\test\lib\site-packages\fastapi\utils.py", line 100, in create_cloned_field
    use_type.__fields__[f.name] = create_cloned_field(f)
  File "D:\virtualenvs\test\lib\site-packages\fastapi\utils.py", line 100, in create_cloned_field
    use_type.__fields__[f.name] = create_cloned_field(f)
  File "D:\virtualenvs\test\lib\site-packages\fastapi\utils.py", line 100, in create_cloned_field
    use_type.__fields__[f.name] = create_cloned_field(f)
  [Previous line repeated 981 more times]
  File "D:\virtualenvs\test\lib\site-packages\fastapi\utils.py", line 97, in create_cloned_field
    original_type.__name__, __config__=original_type.__config__
  File "D:\virtualenvs\test\lib\site-packages\pydantic\main.py", line 773, in create_model
    return type(model_name, (__base__,), namespace)
[...]

PR #1100, opened 2020-03-09
FIX: #894

l3str4nge:
My proposition for #894 --> mark model as cloned. 
In case when it's not enough perhaps it will be good base for main solution :)

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -140,12 +140,12 @@ def create_cloned_field(field: ModelField) -> ModelField:
         original_type = cast(Type[BaseModel], original_type)
         use_type = create_model(original_type.__name__, __base__=original_type)
 
-        if hasattr(use_type, 'cloned'):
+        if hasattr(use_type, "cloned"):
             if not use_type.cloned:
                 for f in original_type.__fields__.values():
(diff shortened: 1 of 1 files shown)


jmagnusson:
This PR fixed issue #894 for me. Thanks @mateuszz0000!

Perhaps a unit test would be good here so that the regression is not re-introduced again. (Could be based off of the code example in #894.

@tiangolo Any chance we could get this merged and released if the code looks good to you? I've been stuck at 0.46 for a while now and would love to take advantage of all the niceties in the newer versions!

jmagnusson:
@mateuszz0000 Looks like you need to format `fastapi/utils.py` with black to get the CI job to pass

mateuszz0000:
Looks like it's something wrong in my solution. How about #1164 created by @voegtlel.

jmagnusson:
@mateuszz0000 Gotcha. #1164 works fine as well.

tiangolo:
Thanks for your contribution @mateuszz0000 :rocket: :cake: 

I used the implementation in #1164 , so I'll close this one now.

And thanks for the discussion here everyone! :coffee:

jmagnusson:
@mateuszz0000 What was this something that was wrong with your solution? It still seems to work fine for me.

Issue #924, opened 2020-01-27
Support OpenAPI example[s] attribute


simonTurintech:
Hello everyone.
Today I tried to use fastAPI to add an exemple of values for a model, as decribed [here](https://swagger.io/docs/specification/adding-examples/).

After a tedious research in the documentation of Pydantic, I managed to go from this implementation:
```
from fastapi import FastAPI
from pydantic import BaseModel

class CustomModel(BaseModel):
    id: str
    name: str

app = FastAPI()

@app.get("/", response_model=CustomModel)
def root() -> CustomModel:
    return {"message": "Hello World"}
```
Giving the following result:
<img width="728" alt="Screenshot 2020-01-27 at 21 51 26" src="https://user-images.githubusercontent.com/22822015/73217441-d4678f80-414f-11ea-81fc-09499ecd155e.png">

To this implementation:
```
from fastapi import FastAPI
from pydantic import BaseModel

class CustomModel(BaseModel):
    id: str
    name: str

    class Config(object):
        schema_extra = {
            'example': {
                'id': 'n3xR4Ib79h',
                'name': 'John Doe'
            }
        }

app = FastAPI()

@app.get("/", response_model=CustomModel)
def root() -> CustomModel:
    return {"message": "Hello World"}
```

Giving the following result:
<img width="726" alt="Screenshot 2020-01-27 at 21 51 47" src="https://user-images.githubusercontent.com/22822015/73217560-10025980-4150-11ea-9abd-5376bec17a22.png">

The code used above is introduced in [this page](https://pydantic-docs.helpmanual.io/usage/schema/#__code_7) of pydantic's documentation.

As this is a behaviour documented in OpenAPI, and as fastAPI can actually provide that behaviour, should it be a good idea to add this to the doc ? It is quite hidden inside pydantic docs, and it could be usefull to other users.

Thank you for your opinion,
Simon

jmriebold:
There's actually a more straightforward way to do it, though it's not really mentioned in the docs. This should do the trick:

```
from fastapi import FastAPI, Body
from pydantic import BaseModel

class CustomModel(BaseModel):
    id: str = Body(None, example='n3xR4Ib79h')
    name: str = Body(None, example='John Doe')
```

The same can be done for query parameters and path parameters, using FastAPI's `Query` and `Path` classes, respectively.

In any case, this should definitely be added to the docs. It took me a fair bit of time to find this solution.

dmontagu:
[...]

Issue #723, opened 2019-11-20
[QUESTION] How can I add example responses to the swagger docs for an endpoint?


jjorissen52:
**Description**

How can I add example responses to the swagger docs for an endpoint?

This does not seem to work:
```python
from fastapi import FastAPI, APIRouter
from pydantic import BaseModel


app = FastAPI()
api = APIRouter()

class Message(BaseModel):
    message: str

    class Config:
        schema_extra = {
             'example': [{
                'message': 'Cool',
            }]
        }

@api.post("/activity", responses={200: {'model': Message}})
async def post_activity(body: dict = dict()):
    print(body)
    return Message(message="thanks")


app.include_router(
    api,
    prefix="/api",
    tags=["api"],
)

```

Result:

![image](https://user-images.githubusercontent.com/12461302/69287725-ed3a4e80-0bbc-11ea-88b7-2dd93c957046.png)

**Additional context**

This issue has been raised before, but the resolution was not shared.
https://github.com/tiangolo/fastapi/issues/549

prostomarkeloff:
However, you can pass example response model as `response_model` attribute in decorator.

```python
@app.get(..., response_model=MyModel)

```

tiangolo:
Thanks for the conversation here everyone!

@jjorissen52 check the new docs here: https://fastapi.tiangolo.com/tutorial/schema-extra-example/

ilya-kolpakov:
See below on how to examples specifically to response models (which is not described in the docs)
https://betterprogramming.pub/how-to-add-multiple-request-and-response-examples-in-fastapi-ce2117eac7ed

PR #1106, opened 2020-03-11
Add documentation of example kwarg of Field

JohnPaton:
Solves #924 

I wasn't sure "Providing Examples" warranted its entire own page in the tutorial, but since providing examples was treated treated in "Body - Fields", I just extended that discussion to include the usage of the `example` keyword argument to `Field` 

Related: #723


tiangolo:
Great! Thanks for your contribution @JohnPaton ! :rocket: :cake: 

I updated it to explain a bit more about how the `example` works.

JohnPaton:
Awesome, great context on there, thanks!

PR #1118, opened 2020-03-15
Add documentation about configuration and env vars

alexmitelman:
In this pull request I add a new documentation page that describes the way to work with environment variables.

Related to this https://github.com/tiangolo/fastapi/issues/190 and this https://github.com/tiangolo/fastapi/issues/399 issues.


tiangolo:
Thanks for starting this! :rocket: :cake: 

I have been intending to add these docs for awhile :sweat_smile:  So I updated it with what I had in mind and to include a couple of extra explanations.

leosussan:
Question to @alexmitelman : any benefit to using Pydantic vs. Starlette's baked-in [configuration stuff](https://www.starlette.io/config/#reading-or-modifying-the-environment)?

alexmitelman:
> Question to @alexmitelman : any benefit to using Pydantic vs. Starlette's baked-in [configuration stuff](https://www.starlette.io/config)?

IMHO, Starlette solution looks more low level, while using Pydantic you just declare class as usual.
Let's compare:

Starlette
```from starlette.config import Config

config = Config(".env")

DEBUG = config('DEBUG', cast=bool, default=False)
```

Pydantic
```from pydantic import BaseSettings

class Settings(BaseSettings):
    debug: bool = False

    class Config:
        env_file = '.env'
```

leosussan:
@alexmitelman that's a description of syntax, not a functional differentiation. It also does not really explain the strengths / benefits of that syntactical difference, or if there is one. 

I've now used both. IMO - given FastAPI's reliance on simplicity, I think Starlette's setup should be the preferred structure for the following three reasons:
1. Starlette's `config` does not require the installation of a secondary dependency (Pydantic requires `python-dotenv`).
2. Starlette's `config` caches the variable by default (no need for a custom `lru_cache` implementation).
2. Starlette's suggested [`config.py` file structure](https://www.starlette.io/config) - simple, flat - is consistent with the `.env` file format.

That said - I've very recently taken to using Pydantic for settings management in Reach for the following three reasons:

1. Pydantic is better for codebases with multiple runtimes (it wouldn't make sense to install Starlette on a worker just to gain access to its' config class for the sake of code sharing).
2. Pydantic lends itself well to the creation of constructed globals (e.g. `DATABASE_URI` as the combination of `DATABASE_PORT`, `DATABASE_HOST`, etc etc).
3. Pydantic has a much larger library of data types to cast / validate globals against.

Those three reasons are compelling (don't get me wrong, Pydantic is amazing; I use it in everything) but I'd imagine the first three would make Starlette's config preferable in most situations.

mths0x5f:
I was surprised by this change, @leosussan summarized what I needed to know.

PR #1121, opened 2020-03-16
Rename a function in tutorial

tmsick:
[Path Parameters - FastAPI](https://fastapi.tiangolo.com/tutorial/path-params/#path-convertor)

```python
from fastapi import FastAPI

app = FastAPI()


@app.get("/files/{file_path:path}")
async def read_user_me(file_path: str):
    return {"file_path": file_path}
```

I changed the name of the above function from `read_user_me` to `read_file`.


tiangolo:
Ah! Good catch. Thanks! :rocket: :cake:

PR #1126, opened 2020-03-17
Update doc string with correct class name

adg-mh:
The docstring inside `OAuth2PasswordRequestFormStrict` incorrectly referred to itself as the permissive dependency class.


tiangolo:
Good catch! :male_detective: :female_detective: :mag: Thanks for your contribution! :rocket: :cake:

PR #1138, opened 2020-03-19
Fix GZipMiddleware code block linking to TrustedHostMiddleware

arimbr:
In https://fastapi.tiangolo.com/advanced/middleware/#gzipmiddleware

### **BEFORE**

image

### **AFTER**

image

### DIFF
--- a/docs/en/docs/advanced/middleware.md
+++ b/docs/en/docs/advanced/middleware.md
@@ -78,8 +78,8 @@ Handles GZip responses for any request that includes `"gzip"` in the `Accept-Enc
 
 The middleware will handle both standard and streaming responses.
 
-```Python hl_lines="2  6 7 8"
-{!../../../docs_src/advanced_middleware/tutorial002.py!}
+```Python hl_lines="2  6"
+{!../../../docs_src/advanced_middleware/tutorial003.py!}
 ```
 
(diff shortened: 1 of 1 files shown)


bilencekic:
is it possible still wrong ? 
should be starlette.middleware.gzip rather than fastapi.middleware.gzip

arimbr:
@bilencekic based on the content in https://github.com/tiangolo/fastapi/blob/master/fastapi/middleware/gzip.py which imports from `starlette.middleware.gzip` it should work right?

@tiangolo Could you have a look at this. I've seen 4 other PR fixing the same issue ^^

bilencekic:
@arimbr  yes correct it is working i am using like that but documentation has changed to
`from fastapi.middleware.trustedhost import TrustedHostMiddleware`
i think GZIP is also embedded to this middleware now.

arimbr:
Right! Thanks, It is just syntactic sugar introduced in https://github.com/tiangolo/fastapi/pull/1064 where this issue was also introduced.

tiangolo:
Great, thank you @arimbr ! :mag: :male_detective: :cake: :rocket: 

And thanks for the discussion here everyone :clap: :cake: :coffee:

PR #1144, opened 2020-03-20
Add an example of setting up a test database.

duganchen:
This adds a block of example code to "Testing Dependencies", to set up and use a test database.

A documented example of how to set up a test database would have helped me immensely. And from reading the issue tracker, I know I'm not the only one.

I believe that what I have here is the correct way to do it.


tiangolo:

I just updated it to use the same SQL files from the tutorial and to put it in its own section.

PR #1159, opened 2020-03-24
Apply openapi_prefix to redirect URL

nunomen:
For my application I need to use Apache as a Reverse Proxy, like this:
```
<VirtualHost *:80>
    ServerName ${SERVER_NAME}
    DocumentRoot ${SERVER_DOCUMENT_ROOT}
    ##  Other Redirects
    
    ## FastApi App
    Redirect permanent /app    ${SERVER_NAME}:/app
</VirtualHost>

<VirtualHost *:443>
    ProxyPass         "/app" "http://localhost:57012"
    ProxyPassReverse  "/app" "http://localhost:57012"

```
In order to do this, I have to setup the fastapi app with the following settings:

```
app = FastAPI(
    title="app",
    description="Ipsum Lorem",
    openapi_prefix="/app",
    swagger_ui_init_oauth={"clientId": "app-id",},
)
```
The openapi_prefix is needed to set up the application behind a reverse proxy as carefully explained in FastApi documentation, otherwise `openapi.json` is not served in the right place. But, for some reason this openapi_prefix was not being applied on the oauth2 redirect_uri in the swagger html generator.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -40,7 +40,7 @@ class FastAPI(Starlette):
         default_response_class: Type[Response] = JSONResponse,
         docs_url: Optional[str] = "/docs",
         redoc_url: Optional[str] = "/redoc",
-        swagger_ui_oauth2_redirect_url: str = "/docs/oauth2-redirect",
+        swagger_ui_oauth2_redirect_url: Optional[str] = "/docs/oauth2-redirect",
         swagger_ui_init_oauth: Optional[dict] = None,
         middleware: Sequence[Middleware] = None,
         exception_handlers: Dict[Union[int, Type[Exception]], Callable] = None,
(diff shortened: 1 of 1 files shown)


littlebrat:
Travis build was failing because of this:
```
fastapi/applications.py:116: error: Unsupported operand types for + ("str" and "None")
fastapi/applications.py:116: note: Right operand is of type "Optional[str]"
```

I don't really understand how these type checkers work, but it doesn't seem to me that Optional[str] is the appropriate choice, given that swagger_ui_oauth2_redirect_url is never None and has a default value of type str.

littlebrat:
Build is still failing and I'm trying to find what is wrong since the 3 first builds don't explicitly say which error was raised.

littlebrat:
Duplicate of #1199

tiangolo:
Thanks for your work @littlebrat ! Clever implementation. But yeah, with #1199 we can handle the issue from the "root" :smile:

PR #1161, opened 2020-03-24
Adding orjson to all section in pyproject.toml

michaeloliverx:
Adding orjson to be in line with the "Optional Dependencies" docs as it was missing.


tiangolo:
Nice, thank you @michael0liver ! :rocket: :cake:

PR #1162, opened 2020-03-24
Add notes about using `typing.Dict` inplace of a Pydantic model

Kilo59:
There are good recommendations for Models and partial updates with PATCH.
However, I believe it's worth pointing out that `typing.Dict` can be used in place of a pydantic model.

https://fastapi.tiangolo.com/tutorial/body-updates/#partial-updates-recap
https://fastapi.tiangolo.com/tutorial/extra-models/

```python
from typing import Dict

@app.patch("/items/{id}")
def patch_item(item_patch: Dict):
```


tiangolo:
Thanks! I think this is covered here: https://fastapi.tiangolo.com/tutorial/body-nested-models/#bodies-of-arbitrary-dicts

I also prefer to not suggest it more than necessary, as APIs with declared bodies using Pydantic will be more secure (validated), better documented, less error-prone, etc. So I'll pass with this.

But thanks for your interest and effort! :rocket: :cake:

PR #1164, opened 2020-03-25
FIX: #894 Check already cloned fields in create_cloned_field to support recursive models

voegtlel:
Include recursion check for create_cloned_field.
Added test for recursive model.

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -132,23 +132,25 @@ def create_response_field(
 
 
 def create_cloned_field(
-    field: ModelField, _cloned_types: Dict[Type[BaseModel], Type[BaseModel]] = None,
+    field: ModelField, *, cloned_types: Dict[Type[BaseModel], Type[BaseModel]] = None,
 ) -> ModelField:
-    # The _cloned_types` store already cloned types such that recursive structures can be supported.
-    if _cloned_types is None:
-        _cloned_types = dict()
+    # _cloned_types has already cloned types, to support recursive models
+    if cloned_types is None:
+        cloned_types = dict()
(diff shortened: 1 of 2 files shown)


merowinger92:
might be worth to mention @dmontagu to review this since he has made already some thoughts about solving this

jmagnusson:
This PR solves the example issue in #894 as well as a more complicated private project of mine. LGTM 👍

tiangolo:
Very clever! Thanks for your contribution @voegtlel :cake: :rocket:

jmagnusson:
For a complex app that I'm building I've seen memory usage go up from ~140MB to ~305MB and app start up time increased from ~2.4s to ~9.5s when moving from version `0.46.0` to `0.54.1`.

It looks like most of the additional time is spent in `create_cloned_field` which was changed a lot in this PR.

Has anyone else had similar issues?

jmagnusson:
[PR 1100](https://github.com/tiangolo/fastapi/pull/1100) still works for me, removing the issues mentioned above.

I've cherry-picked his commit on top of the current head in this repo (4e77737a3f7bf2608132ea170e9ff013b5af6732) and that works as well.

phy25:
@jmagnusson I would recommend that you create a new issue, or this might be left unnoticed.

tiangolo:
As @phy25 says, it would probably be better in a new issue.

Anyway, if you want to profile specific functions and timings, you can try https://github.com/sm-Fifteen/asgi-server-timing-middleware

Or the ideas in this thread: https://github.com/tiangolo/fastapi/issues/701

That might help to detect where is the change exactly.

PR #1166, opened 2020-03-25
Implemented response_model_exclude_defaults and response_model_exclude_none

voegtlel:
Implemented response_model_exclude_defaults and response_model_exclude_none to support these pydantic options.


tiangolo:
Great job @voegtlel ! :bowing_man: :clap: :nerd_face: 

Thanks for keeping the changes minimal and for adding all the tests. :heavy_check_mark: 

I think it's okay to deprecate `include_none` and use instead the inverted `exclude_none` to keep in sync with Pydantic, so I updated that bit.

PR #1169, opened 2020-03-26
Change address to localhost for Windows

mariacamilagl:
We changed the address to localhost for Windows, since it doesn't work in Windows for 0.0.0.0. Now it should show 127.0.0.1


tiangolo:
Great, thanks! :rocket: :cake:

PR #1183, opened 2020-03-30
Update sql-databases alembic docs

HarshaLaxman:
Was helpful to refer to the full-stack project when integrating alembic into my own project


tiangolo:
Good idea! Thanks for your contribution. :cake: :rocket: :tada:

PR #1188, opened 2020-03-30
Fix minor grammatical mistakes in the async docs

mickeypash:
First of all, 
Thank you for this awesome framework and documentation! I only started using it last week but I'm already in love with it. You provide a lot of context for all the components and decisions and this makes it so easy to follow!

- I think the correct pronoun when referring to a computer is `it`. 💻 
- I also used the adjective `Next` but I think in a few places we could replace "and" and "then" with other conjunctive adverbials ([extensive list here](https://www.citationmachine.net/grammar-and-plagiarism/conjunction/conjunctions-list/)) 🤓 
- The verb tensing should `sit`, `sat`, `sitting` - I've made that mistake myself 🙈 

Let me know what you think 🙇
I hope to start by contributing to the docs and move into helping out with some of the code 👨‍💻

### DIFF
--- a/docs/en/docs/async.md
+++ b/docs/en/docs/async.md
@@ -63,13 +63,13 @@ Let's see that phrase by parts in the sections below, below:
 
 ## Asynchronous Code
 
-Asynchronous code just means that the language has a way to tell the computer / program that at some point in the code, he will have to wait for *something else* to finish somewhere else. Let's say that *something else* is called "slow-file". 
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks a lot for checking and fixing this! :mag: :male_detective: :tada: 

I was using the incorrect "he" for the computer to try and differentiate it from the other possible "it" things in the context. I just updated your fixes with some emojis to try and make it unambiguous.

Thanks for your contribution! :rocket: :cake:

Issue #829, opened 2020-01-06
[BUG] OpenAPI routes do not acknowledge root_path


sm-Fifteen:
Also raised on https://github.com/tiangolo/fastapi/pull/26#issuecomment-562755647. See also #544.

### Describe the bug

Write here a clear and concise description of what the bug is.

### To Reproduce

Replace each part with your own scenario:

1. Create a file with:

```Python
from fastapi import FastAPI

app = FastAPI()

@app.get("/app")
def read_root():
    return {"Hello": "World"}
```

2. Launch it using `uvicorn --root-path="bar" test_app:app`
3. Open the browser and go to `http://127.0.0.1:8000/docs`.
4. From the documentation, call the `GET /app` route.
5. The doc page calls `/app` and succeeds.

### Expected behavior

The above test should fail after having called `/bar/app`, since `root_path` is supposed to prefix all generated URLs in case the application is served behind a reverse-proxy, among ther things. FastAPI only acknowledges `openapi_prefix` for the API doc.

### Environment

- OS: Windows
- FastAPI Version: 0.45.0
- Python version: 3.8.0

### Additional context

A similar issue applies to sub-applications:

```py
from fastapi import FastAPI

app = FastAPI()


@app.get("/app")
def read_main():
    return {"message": "Hello World from main app"}


#subapi = FastAPI(openapi_prefix="/subapi")
subapi = FastAPI()

@subapi.get("/sub")
def read_sub(request: Request):
    return {
        "root_path": request.scope['root_path'],
        "raw_path": request.scope['raw_path'],
        "path": request.scope['path'],
        "app_url_for": app.url_path_for("read_sub"),
        "subapp_url_for": subapi.url_path_for("read_sub"),
    }

app.mount("/subapi", subapi)
```

```json
{
  "root_path":"bar/subapi",
  "raw_path":"/subapi/sub",
  "path":"/sub",
  "app_url_for":"/subapi/sub",
  "subapp_url_for":"/sub"
}
```

(`url_for` not being prefixed with `root_path` is fixed upstream by encode/starlette#699)

Unless `openapi_prefix="/subapi"` is passed when creating the subapplication, both `http://127.0.0.1:8000/docs` and `http://127.0.0.1:8000/subapi/docs` will point towards `http://127.0.0.1:8000/openapi.json`, which goes against the point of having isolated subapplications.

`openapi_prefix` should probably just be deprecated and assumed to match `root_path` if absent.

and-sm:
[...]

PR #1199, opened 2020-03-31
Add support for ASGI root_path for openapi docs

iksteen:
ASGI specifies the `root_path` connection scope variable. This is the path the application is mounted under (the equivalent of WSGI's `SCRIPT_NAME`).

This PR ensures that this root is used as a prefix for the openapi.json URL, as a prefix for paths contained in openapi.json and by the oauth2 redirect endpoint to redirect back to the swagger docs.

To override the path the ASGI server provides you can still explicitly set the `openapi_prefix` variable.

Do note that this is a breaking change for applications that overriden `openapi()` themselves.

This should fix/related to https://github.com/tiangolo/fastapi/issues/829


littlebrat:
I guess this PR solves the same issue that I was having (I didn't even know about issue #829). I solved my problem by just adding the prefix to the redirect url. You can see what I did on my proposed PR https://github.com/tiangolo/fastapi/pull/1159. But it seems to me that your solution involves more steps that I may have been missing. Could you explain? 

_I will probably delete https://github.com/tiangolo/fastapi/pull/1159 since it is redundant, now that you've done this 😄_

tiangolo:
Great! :rocket: 

Thanks for the patience, it took me a bit of time to review this as I wanted to first re-read the WSGI and ASGI specs to make sure I wasn't missing anything.

And then to test it all with an actual reverse proxy with path prefix stripping (what the `root_path` would be used for).

I refactored it a bit and took the opportunity to deprecate `openapi_prefix` in favor of the new `root_path` :fire: :rocket: 

And I added a bunch of docs to explain how the `root_path` works and to document how to test it all with a real proxy using Traefik.

Thanks for your contribution! :rocket: :tada: :cake:

PR #1212, opened 2020-04-04
Ignore env directory on version control

cassiobotaro:
As the [development instructions](https://fastapi.tiangolo.com/contributing/#virtual-environment-with-venv) are to create an env, we should ignore it to avoid mistakes.


tiangolo:
Good idea, thanks! :nerd_face: :cake:

Issue #1221, opened 2020-04-07
tests with pydantic failing but not details


samuelcolvin:
see [here](https://github.com/samuelcolvin/pydantic/pull/1344/checks?check_run_id=568360178) fastapi tests with pydantic are currently failing.

**Problem 1:**  do we think this is pydantic related or something to do with fastapi? I currently only have python 3.8 on my laptop which is failing, so hard to check.

**Problem 2:** the failure is on `assert response.status_code == 200` would be very helpful if there was some discipline on checks like this to do `assert response.status_code == 200, response.text` to aid in investigating problems - I'll submit a PR for this now.

samuelcolvin:
okay, after creating #1222 and running tests I can see that the error is 

```json
{"detail":"Email already registered"}
```

but it's only happening when I run `PYTHONPATH=./docs/src pytest`, not when I run `./scripts/test.sh`.

My guess is that this is caused by a brittleness in test order, e.g. that test fails if some other test has already created the user.

I'm probably not going to be the fastest person to get to the bottom of this.

@tiangolo, sorry to bug you but it would be great if you could look into this asap as it's currently breaking pydantic tests. Alternatively, if you're not going to have time, let me know and I'll temorarily pin the fastapi tests on pydantic to the previous tag. Thanks.

tiangolo:
Dang it :man_facepalming: . Thanks for the report, I'll check it ASAP.

tiangolo:
This should be fixed by #1226 , released in version `0.54.1`.

I re-triggered the Pydantic tests and it seems that it indeed fixed the problem: https://github.com/samuelcolvin/pydantic/pull/1344/checks?check_run_id=569826093

PR #1226, opened 2020-04-08
Update database setup for tests

tiangolo:
:card_file_box: Update database setup for tests.

Related to #1221


samuelcolvin:
Thanks so much.

Regarding the many "🚀 Deployed on" comments, if you setup the deploy as a separate job in github CI, you can use normal commit statuses rather than comments.

Have a look at pydantic - it's a bit more complicated as I got pissed off with netlify, but it explains the idea.

Issue #1234, opened 2020-04-10
[FEATURE] Remove *, where it's not needed


tiangolo:
### Is your feature request related to a problem

Coming from this issue: https://github.com/tiangolo/fastapi/issues/817

The docs have some sections with functions using `*,` in the parameters.

It is explained here: https://fastapi.tiangolo.com/tutorial/path-params-numeric-validations/#order-the-parameters-as-you-need-tricks

But in other sections like: https://fastapi.tiangolo.com/tutorial/response-model/ the `*,` is not really needed and would probably just cause some confusion.

### The solution you would like

Remove the `*,` from function parameters in the docs at the places where it's not needed, i.e. the places where a function parameter with a default (e.g. a `Query()` or `Depends()`) does NOT come before a function parameter without a default value.

### Describe alternatives you've considered

Leaving it as is. But, as always, I want the docs to be as easy and friendly as possible.

### Additional context

This issue: https://github.com/tiangolo/fastapi/issues/817

pankaj-giri:
@tiangolo  - Looking into it.. please assign  #1234 to me.

chris-allnutt:
@pankaj-giri any updates?

pankaj-giri:
@chris-allnutt - I had submitted this pull request some-time ago.. 
https://github.com/tiangolo/fastapi/pull/1239
Not sure why it wasn't merged.

tiangolo:
This was fixed in https://github.com/tiangolo/fastapi/pull/1239 by @pankaj-giri ! :rocket: 

So I'll close this now. :tada:

PR #1239, opened 2020-04-11
Fix for - [FEATURE] Remove *, where it's not needed #1234

pankaj-giri:
Removed * from function parameters in the docs wherever not required. Please take a look and let me know if it looks good.


tiangolo:
Awesome, thanks for your contribution! :rocket: :cake:

PR #1262, opened 2020-04-14
support add exception handler by decorator

Zheaoli:


tiangolo:
Thanks for your interest @Zheaoli ! :nerd_face: :cake: 

I'm not really sure it would help much having by default a single decorator that handles multiple exceptions that could have very different shapes and types.

If several exceptions are similar enough that they can be handled by the same decorator, I think those exceptions could inherit from the same base class, so that the decorator can handle that class.

So I'm gonna pass on this one, but thanks for your effort! :coffee:

PR #1267, opened 2020-04-15
✨ Add support for adding multiple examples in request bodies and path, query, cookie, and header params

### LINKED ISSUE #822: [FEATURE] Enable support for multiple request body "examples" in openapi / swagger ui docs
This feature is related to issue #372 

It appears to me in the Swagger UI documentation on [adding examples](https://swagger.io/docs/specification/adding-examples/) that the 'examples' key belongs to 'application/json'. It also appears to me that the key already exists in the `MediaType` model within `fastapi.openapi.models` but that only the 'schema' key is modified when writing the openapi spec through `fastapi.openapi.utils.get_openapi_operation_request_body` and then `fastapi.openapi.utils.get_open_api_path` and that the media type is nearly always passed around as a string rather than a dict or model object (please correct me if I'm wrong, I've only been using these particular tools for a few months.)

`pydantic` recently enabled `schema_extra` in [samuelcolvin/pydantic#663 ](https://github.com/samuelcolvin/pydantic/pull/663) but I can't seem to get it to work with the swagger docs. At this time it's unclear to me if this is simply a part of the json schema spec that the swagger ui doesn't recognize yet, or if it's actually a deliberate deviation from the spec, or if I just can't figure it out. From the JSON Schema documentation on ["examples"](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.5) it states that this field "MUST" be an array. The aforementioned `pydantic` PR#663 and the associated [example](https://github.com/samuelcolvin/pydantic/blob/master/docs/examples/schema_with_example.py) adhere to the spec well, but I cannot seem to build a request model using this newly minted ability that is supported by the swagger docs.

Here's a working reference swagger ui yaml file that builds the dropdown examples pick list in the [swagger editor](https://editor.swagger.io). I started with the `fastapi.docs.body_fields.tutorial_002.py` example and hand edited the swagger ui yaml to match the swagger examples docs for multiple examples:
```
openapi: 3.0.2
info:
  title: Fast API
  version: 0.1.0
paths:
  '/items/{item_id}':
    put:
      summary: Update Item
      operationId: update_item_items__item_id__put
      parameters:
        - required: true
          schema:
            title: Item Id
            type: integer
          name: item_id
          in: path
      requestBody:
        content:
          application/json:
            schema:
              title: Item
              allOf:
                - $ref: '#/components/schemas/Item'
            examples:
              test1:
                  summary: first test
                  value:
                    name: Foo
                    description: A very nice Item
                    price: 35.4
                    tax: 3.2
              test2:
                  summary: second test
                  value:
                    name: Bar
                    description: Another very nice Item
                    price: 30
                    tax: 1.1
        required: true
      responses:
        '200':
          description: Successful Response
          content:
            application/json:
              schema: {}
        '422':
          description: Validation Error
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/HTTPValidationError'
components:
  schemas:
    HTTPValidationError:
      title: HTTPValidationError
      type: object
      properties:
        detail:
          title: Detail
          type: array
          items:
            $ref: '#/components/schemas/ValidationError'
    Item:
      title: Item
      required:
        - name
        - price
      type: object
      properties:
        name:
          title: Name
          type: string
        description:
          title: Description
          type: string
        price:
          title: Price
          type: number
        tax:
          title: Tax
          type: number
    ValidationError:
      title: ValidationError
      required:
        - loc
        - msg
        - type
      type: object
      properties:
        loc:
          title: Location
          type: array
          items:
            type: string
        msg:
          title: Message
          type: string
        type:
          title: Error Type
          type: string
```

My hope is that either 1) There is another way to add multiple examples for a request body that I am missing or that 2) we can build a means for adding the "examples" dictionary to the media type so it appears in the swagger ui per their documented format, preferably building from the recent `pydantic` PR referenced above.

I recognize that a few things complicate this effort, including the fact that a `fastapi` `Route` can include multiple `Body` objects. This might mean that if we support the documented swagger format linked above for multiple examples, the attribute might have to be moved out of the `Body` function/class so that only one "examples" dict can be specified per route, or better yet, per content media type somehow.

With some direction I think I would be able to help implement this, since it's not a mission critical feature for the lead developers.

austinorr:
This PR enables multiple examples in a pick-list style dropdown in the OpenApi Docs a la the 'Request Body Examples' [here](https://swagger.io/docs/specification/describing-request-body/)

For more information and discussion see the issue this PR related to #822 

~~Note, this still needs tests to cover the 3 new lines.~~ Docs and full coverage tests added Sept 4 with 8138ff6

---

Edit by @tiangolo :nerd_face: :point_down: 

Added support for arguments `example` and `examples` in `Body()`, `Path()`, `Query()`, `Cookie()`, `Header()`, handling compatibility (and incompatibilities) between OpenAPI, JSON Schema, and OpenAPI's 3.0.x custom version of JSON Schema :zany_face: :dizzy:

More details about the standards dizzy mix in a comment below, near the end.


fastapi/openapi/utils.py
```diff
@@ -126,6 +126,9 @@ def get_openapi_operation_request_body(
     if required:
         request_body_oai["required"] = required
     request_body_oai["content"] = {request_media_type: {"schema": body_schema}}
+    examples = body_schema.get("examples")
```
Zheaoli, line 129: would you like to add some unit tests for this code in your PR?

austinorr: Yeah, i'd be happy to. I have waited to add tests and docs because I wanted to get some feedback on #822 and the proposed solution in this PR from the library maintainers, but if this looks good to you and to @tiangolo (I'm tagging him here because he had commented on the issue) then I'd be glad to add tests and even a little docs demo of the functionality.

austinorr: Rebased, added tests and a brief edit to the docs. I also added a screen grab, but my system is slightly different from the system used to create other `/docs` screenshots.

austinorr:
The docs built for this PR have a screenshot of the new multi-example capabilities [here](https://5f52f468e319ee0ebf5d7aad--fastapi.netlify.app/tutorial/schema-extra-example/#body-with-multiple-examples).

kbakk:
@austinorr The description mentions the need for more coverage but you seem to have added tests - is it ready to be merged perhaps? It would be a really useful feature! 😃

Kludex:
@kbakk `docs_src` is not being covered, maybe that's why...

check this: https://github.com/tiangolo/fastapi/pull/1904

austinorr:
@kbakk and @Kludex, I was thinking this PR is ready to go. Is there some reason merging this PR should wait on #1904? If not, I think this is fully tested (relative to the master branch).

I could also push up an empty commit to try to force a re-run of the tests/docs/coverage if that'd be useful.

Kludex:
@austinorr The goal of my message was to explain one of the possible reasons for what he saw. I didn't mean that we should wait. :sweat_smile: 

> I could also push up an empty commit to try to force a re-run of the tests/docs/coverage if that'd be useful.

I don't think that is necessary... @tiangolo will eventually see this PR, dw :+1:

mattnotmitt:
@tiangolo Is there any chance of seeing this merged soon?

austinorr:
For easier review of the capability this PR enables see the (draft) docs I edited here: https://5f52f468e319ee0ebf5d7aad--fastapi.netlify.app/tutorial/schema-extra-example/#body-with-multiple-examples

kbakk:
> Note, this still needs tests to cover the 3 new lines.

If this isn't the case anymore, maybe it should be removed from the description. :)

austinorr:
> If this isn't the case anymore, maybe it should be removed from the description. :)

Hey sorry -- we use git comments differently. I will update the original comment to indicate that this has been covered since the commits on Apr 14. The Coverage bot is current (it updates live).

adriangb:
Coming over from https://github.com/tiangolo/fastapi/issues/822#issuecomment-762555680.

I tried the implementation in this PR with Pydantic models as described in https://github.com/tiangolo/fastapi/issues/822#issuecomment-761915400 and it did not seem to work. That said, there is a good chance I made a mistake somewhere along the way.

Maybe [this](https://fastapi.tiangolo.com/tutorial/schema-extra-example/?h=+example#pydantic-schema_extra) example can be updated/extended and it can be tested that way? I'd be happy to _try_ to submit a PR to this PR with this (but I'm not very familiar with the internals, so I can't promise I'll get it to work).

austinorr:
Yeah, I see what you mean. The current implementation doesn't work with the `schema_extra`. If we think this through a bit more, the openapi examples picklist relate 1:1 with each route, since the examples are bound to the media type. This is really different from the `schema_extras` and the existing `example` (singular) method for creating reference examples; those are 1:1 for each _model_.

You're right to note that my implementation is a bit narrow in that it'll only work as expected if the developer creates a route with only one `Body` within it. `fastapi` has other `Body` params that only work if there's 1 `Body` though, like the `embed` kwarg so my hope was that this hack could sneak through since I'd still find it useful.

The `openapi.models` module has a `MediaType`  model that appears to already be configured correctly, but I don't yet see anywhere in the `fastapi` library that initializes this object or offers an entrypoint for a developer to initialize it with a value for `examples`... so in order to keep my diff down I just lifted the key off of the `Body` if it's there (since there is no valid plural `examples` key defined on a body according openapi and the `fastapi` [docs](https://fastapi.tiangolo.com/tutorial/schema-extra-example/?h=+examples#technical-details)) and I just injected into the media type dict. I'd like to avoid any solutions that require the user to define a `MediaType` model on a per-route basis. This `examples` kwarg should either be on the route or on the `Body` with the same limits as the `embed` kwarg (I've chosen the latter since the diff was just 2 lines.)

adriangb:
That makes sense, I think is it pretty sensible to support the simple use case (in terms of implementation and expected behavior, as you point out above). There is certainly a discrepancy between examples-per-route (OpenAPI) and examples-per-model (Pydantic), but that is somehow reconciled within FastAPI, since `example` in `schema_extra` does work.

That said, should also be made to work with `examples` in `schema_extra`, but maybe in a separate PR. I took a quick peek under the hood, the only thing I could see that stood out is that `SchemaBase` has `example` defined ([here](https://github.com/tiangolo/fastapi/blob/561bbfb5d23960560b1258bd5859b8b8737fe7ae/fastapi/openapi/models.py#L117)) but not `examples`. This made `examples` immediately be dropped from `schema_extra`.

austinorr:
@adriangb That's right, my understanding is that there's actually nothing we can do with `examples` on schema_base, since openapi doesn't support that key on the body, only on the media type. It's not just a different PR, it'd be a PR to the openapi repo to bring _model-level_ dropdown pickers to the swagger docs (that'd be really cool).

The `example` (singular) key works within the current paradigm because we can build up our `Body` from multiple models and create a valid overall example as long as every model has exactly one example. Otherwise we'd have a 'tree' of examples that we'd have to prune, and the behavior would likely differ from the developers intent.

I'll also admit that I'm way out over my skis here; I just use this fantastic tool, I don't know how it really does some of its magic like the core devs do. Some of this might actually be possible/easier than I thought if we can get this thread reviewed by a dev who knows `openapi`, `fastapi` and `pydantic` really well.

adriangb:
That makes a lot of sense, your explanation is excellent.

> it'd be a PR to the openapi repo to bring model-level dropdown pickers to the swagger docs

That would be cool! But I'm not sure how that fits into the larger OpenAPI picture? Like you I just use this great tool 😅 

> we'd have a 'tree' of examples that we'd have to prune

Right, and there's not valid way to select the right leaf. I'd imagine there'd need to be some FastAPI specific way for developers to designate the example to use in building the overall examples (it could just be the first one).

> Some of this might actually be possible/easier than I thought if we can get this thread reviewed by a [core] dev

Do you want to tag someone here, or would you prefer to open a separate issue? Up to you.

adriangb:
Looking at this a bit more, it looks like FastAPI does nothing special with `example`, instead, it is Swagger itself that has the logic of "if the body is just a $ref to a schema in components, and that schema in components has an `example` (singular), then pull that example and show it as a body example in the path description".

So indeed, supporting `examples` from Pydantic models would be an OpenAPI request, not a FastAPI request, and thus your PR should remain as is. I'm sorry for the diversion @austinorr .

talarari:
will this support multiple examples for a union type body?
i'm interested in having a route be able to take body that is either a single model or a list of models, so:
```
POST /users
```
could take body
```json
{
   "name": "john"
}
```
or 
```json
[{
   "name": "john"
},
{
   "name": "jack"
}]
```

currently the example value for union body types is just empty.
would be great to have this be more descriptive.

talarari:
any update on this? is there a plan to merge it?

tiangolo:
Thanks for all the effort and work on this @austinorr ! :nerd_face: And thanks for all the clarifications and arguments here and in the issue! :memo: :rocket: 

Also thanks for the discussion everyone! :coffee: 

---

I finally got the time to work on this for several consecutive days (mainly studying the standards) and I refactored it a bit to handle a couple of things.

This was quite tricky, as although the description and original change were not too complex, it touches one of the main friction points in the standards, between OpenAPI, JSON Schema, OpenAPI's 3.0.x custom version of JSON Schema, and the recently released (soon to be supported) OpenAPI 3.1.0, based on a recent version of JSON Schema, without the custom parts. :sweat_smile: 

I added internal support for the arguments `example` and `examples` in `Body()`, `Path()`, `Query()`, `Cookie()`, `Header()`, handling the compatibility issues between those standards.

## Boring Technical Details

The update/refactor adds `example` and `examples` (handling overrides from `examples` over `example` as specified in the spec) to each path operation's section in OpenAPI, not to the JSON Schema section, as the format for OpenAPI's `examples` is different than the format for `examples` in recent versions of JSON Schema. Although the currently supported OpenAPI 3.0.x doesn't support recent versions of JSON Schema, the recently release OpenAPI 3.1.0 is based on a recent JSON Schema, which includes its own definition of `examples`, so OpenAPI 3.1.0 no longer customizes its own version of JSON Schema as much as before, and the custom `example` in OpenAPI's own JSON Schema is although still supported, now deprecated. ...nevertheless, Swagger UI still doesn't support OpenAPI 3.1.0, so I won't update FastAPI yet.

Maybe all that sounds super complex :grimacing: ...it's mainly that the terms are the same here and there, but they refer to slightly different things in one standard and the other which adds to the confusion. Plus, the conflicts also involve older and newer versions, so it sounds dizzy. 🥴

---

But **in short**: this now has "first-class" support, handling all the details, making sure it's future proof, for OpenAPI 3.1.0 (:crossed_fingers: ). :sparkles:

And, now it also supports the other parameters, with a bunch of tests for all that to make sure overrides work correctly, etc. :tada: 

This will be included in the next release, `0.64.0`, in the next days. :tada:

adriangb:
Thank you @tiangolo and @austinorr!

Issue #1284, opened 2020-04-17
Class FastAPI parameters openapi_prefix and openapi_prefix with conflict


sunnnnz:
If Class FastAPI have this openapi_prefix parameter (example: /v2)
like that the actual request address is 127.0.0.1:8080/v2/openapi.json (Resources don't exist)
but the actual effective address is 127.0.0.1:8080/openapi.json

tiangolo:
Please check the new docs here: https://fastapi.tiangolo.com/advanced/behind-a-proxy/

PR #1286, opened 2020-04-18
fix open api prefix is invalid :bug:

Dustyposa:
create for #1284 
I can't make sure this is a bug. 
I'm not be familiar with openapi.
If this is not a bug, I will close this pr.
@tiangolo

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -104,9 +104,9 @@ class FastAPI(Starlette):
 
             async def openapi(req: Request) -> JSONResponse:
                 return JSONResponse(self.openapi())
-
-            self.add_route(self.openapi_url, openapi, include_in_schema=False)
             openapi_url = self.openapi_prefix + self.openapi_url
+            self.add_route(openapi_url, openapi, include_in_schema=False)
+
         if self.openapi_url and self.docs_url:
 
             async def swagger_ui_html(req: Request) -> HTMLResponse:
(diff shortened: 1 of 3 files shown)


fastapi/applications.py
```diff
@@ -105,8 +105,9 @@ def setup(self) -> None:
             async def openapi(req: Request) -> JSONResponse:
                 return JSONResponse(self.openapi())
 
-            self.add_route(self.openapi_url, openapi, include_in_schema=False)
             openapi_url = self.openapi_prefix + self.openapi_url
+            self.add_route(openapi_url, openapi, include_in_schema=False)
```
Zheaoli, line 109: I doubt it's a bug? or it's just a prefix will affect on the swagger or redoc at begin of the design

Dustyposa: Yes, I doubt it's a bug too.
If not a bug, the kwargs `openapi_prefix` has some ambiguity.
And i'm not familiar with openapi, so i can't make sure my fix is right way.
Need some review.

tiangolo:
Thanks for your effort @Dustyposa ! :cake: 

Yeah, it was not very obvious how all this should work.

I just updated all that section, and also deprecated `openapi_prefix` in favor of `root_path` that is part of the ASGI standard.

I also documented how it all should work here (all that was part/on top of a PR from someone adding support for `root_path`).

Check the new docs: https://fastapi.tiangolo.com/advanced/behind-a-proxy/

I'm gonna close this now, but thanks for the effort! :nerd_face: :coffee:

PR #1289, opened 2020-04-19
remove 'Vote to include FastAPI in awesome-python'

donhui:
remove 'Vote to include FastAPI in awesome-python', because the [PR](https://github.com/vinta/awesome-python/pull/1209) was closed


tiangolo:
Good catch, thanks! :fire:

PR #1291, opened 2020-04-20
move * to the end of line

donhui:
move * to the end of line，to be consistent with the next line


tiangolo:
Good catch, thanks! :mag: 

I just added it here: https://github.com/tiangolo/fastapi/pull/1424

Making sure it is consistent in the other languages too and the README so that the tests pass.

So I'll close this now, but thanks for your contribution! :cake:

PR #1293, opened 2020-04-20
Fix the get_db example function

bcb:
Creation of the SessionLocal should be outside the try block.


tiangolo:
Hey @bcb , thanks for the suggestion.

Why do you think that it should be outside of the try block?

tiangolo:
Great, thanks for your contribution @bcb ! :rocket: :cake:

Issue #1190, opened 2020-03-30
How to catch misspelled/unknown parameters?


bear24rw:
How can I catch misspelled or unknown query parameters? For instance:
```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/")
def root(myparam: int = 0):
    return {'myparam': myparam}
```
```bash
$ curl http://127.0.0.1:8000/?param=5
{"myparam":0}
```
It seems bad to silently accept the unknown `param` parameter. The caller wouldn't be aware that they misspelled the parameter and that it remained at its default value.

Kilo59:
I think that's pretty typical for REST api's and isn't something I would expect FastAPI to prevent.

Example
https://petstore.swagger.io/v2/store/inventory?foo=bar

bear24rw:
I suppose that's true but given fastapi's focus on parameter value validation I had expected it to also validate the parameter keys as well. I think the fact that the parameters are defined as part of the function signature led me to assume you would get an error just as if you called a regular python function with an invalid keyword argument.

Are there situations where you would _not_ want to assert that all parameter keys are valid?

Dustyposa:
You can try this one:
```python
from fastapi import FastAPI, Query

app = FastAPI()

@app.get("/")
def root(myparam: int = Query(...)):
    return {'myparam': myparam}
```
I can't understand that why set a default value for must required field?

bear24rw:
Here is a patch that I think implements what I am looking for:
```diff
diff --git a/fastapi/dependencies/utils.py b/fastapi/dependencies/utils.py
index 43ab4a0..577961f 100644
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -35,7 +35,7 @@ from fastapi.utils import (
 )
 from pydantic import BaseConfig, BaseModel, create_model
 from pydantic.error_wrappers import ErrorWrapper
-from pydantic.errors import MissingError
+from pydantic.errors import MissingError, ExtraError
 from pydantic.utils import lenient_issubclass
 from starlette.background import BackgroundTasks
 from starlette.concurrency import run_in_threadpool
@@ -529,6 +529,12 @@ async def solve_dependencies(
     values.update(header_values)
     values.update(cookie_values)
     errors += path_errors + query_errors + header_errors + cookie_errors
+
+    dependant_names = {param.name for param in dependant.query_params}
+    for param in request.query_params.keys():
+        if param not in dependant_names:
[...]

PR #1297, opened 2020-04-21
added option to also check for unknown query parameters

mortbauer:
Currently unknown query parameters are just ignored by fastapi, I feel that it should be at least possible to change the behaviour to return a validation error also for unknown.

It adresses #1190 

* not yet documented
* not yet any testing
* maybe different path to pass that option
* maybe rename to something else, maybe could be thought to generally
  check unknowns, however for models, pydantic has to be adapted

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -10,6 +10,7 @@ from typing import (
     Mapping,
     Optional,
     Sequence,
+    Set,
     Tuple,
     Type,
     Union,
@@ -31,7 +32,7 @@ from fastapi.security.open_id_connect_url import OpenIdConnect
 from fastapi.utils import create_response_field, get_path_param_names
 from pydantic import BaseModel, create_model
 from pydantic.error_wrappers import ErrorWrapper
-from pydantic.errors import MissingError
+from pydantic.errors import ExtraError, MissingError
 from pydantic.fields import (
     SHAPE_LIST,
     SHAPE_SEQUENCE,
@@ -473,6 +474,8 @@ async def solve_dependencies(
     response: Optional[Response] = None,
     dependency_overrides_provider: Optional[Any] = None,
     dependency_cache: Optional[Dict[Tuple[Callable, Tuple[str]], Any]] = None,
+    check_unknown: bool = False,
+    left_query_params: Optional[Set] = None,
 ) -> Tuple[
     Dict[str, Any],
     List[ErrorWrapper],
@@ -489,6 +492,10 @@ async def solve_dependencies(
         media_type=None,
         background=None,
     )
+    outer_level = False
+    if left_query_params is None:
+        left_query_params = set(request.query_params)
+        outer_level = True
     dependency_cache = dependency_cache or {}
     sub_dependant: Dependant
     for sub_dependant in dependant.dependencies:
@@ -523,6 +530,8 @@ async def solve_dependencies(
             response=response,
             dependency_overrides_provider=dependency_overrides_provider,
             dependency_cache=dependency_cache,
+            left_query_params=left_query_params,
+            check_unknown=check_unknown,
         )
         (
             sub_values,
@@ -571,12 +580,22 @@ async def solve_dependencies(
     values.update(header_values)
     values.update(cookie_values)
     errors += path_errors + query_errors + header_errors + cookie_errors
+
+    if check_unknown:
+        dependant_names = {param.name for param in dependant.query_params}
+        left_query_params.difference_update(dependant_names)
+        if outer_level:
+            for param in left_query_params:
+                errors.append(ErrorWrapper(ExtraError(), loc=("query", param)))
+
     if dependant.body_params:
         (
             body_values,
             body_errors,
         ) = await request_body_to_args(  # body_params checked above
-            required_params=dependant.body_params, received_body=body
+            required_params=dependant.body_params, 
+            received_body=body, 
+            check_unknown=check_unknown
         )
         values.update(body_values)
         errors.extend(body_errors)
@@ -641,9 +660,11 @@ def request_params_to_args(
 async def request_body_to_args(
     required_params: List[ModelField],
     received_body: Optional[Union[Dict[str, Any], FormData]],
+    check_unknown: bool = False,
 ) -> Tuple[Dict[str, Any], List[ErrorWrapper]]:
     values = {}
     errors = []
+    left_body_params = set(received_body)
     if required_params:
         field = required_params[0]
         field_info = field.field_info
@@ -653,6 +674,7 @@ async def request_body_to_args(
             received_body = {field.alias: received_body}
 
         for field in required_params:
+            left_body_params.discard(field.alias)
             loc: Tuple[str, ...]
             if field_alias_omitted:
                 loc = ("body",)
@@ -709,6 +731,9 @@ async def request_body_to_args(
                 errors.extend(errors_)
             else:
                 values[field.name] = v_
+    if check_unknown:
+        for field in left_body_params:
+            errors.append(ErrorWrapper(ExtraError(), loc=("body", field)))
     return values, errors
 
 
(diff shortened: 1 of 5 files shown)


rogersjw:
Is there any progress on this? I see there are some merge conflicts and missing tests. Having this optional feature would be a nice to have for some work I am doing as well

mortbauer:
No progress from my side, I'm a bit short on time and before investing into testing and docs I thought to wait for a bit of feedback about the implementation. But I guess with more than 60 pending pull requests it is quite difficult to keep up.

sebg1:
I am interested with the fix, when it will be available.

jeremyschiemann:
+1

hawkaa:
I'm interested in the fix too. I know it has been considered as an edge case by @tiangolo in a similar issue/PR, but I can explain my niece use case.

I'm implementing the "OGC API - Features - Part 1: Core" (http://docs.opengeospatial.org/is/17-069r3/17-069r3.html) which is a standard for fetching geospatial features over HTTP. In section 7.6 we read the following:

> The server SHALL respond with a response with the status code 400, if the request URI includes a query parameter that is not specified in the API definition.

This is not a big deal, but just wanted to let you guys know of an use case where this functionality would come in handy.

diogoduartec (approved):

boyaps (approved):

mortbauer:
so Í took some time to work on this again:
* rebased
* added tests
* added minimal docs
* added support to also check for unknown body params
could it be merged?

tests/test_check_unknown.py
```diff
@@ -0,0 +1,66 @@
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+from pydantic import BaseModel
+
+app_no_check = FastAPI(check_unknown=False)
+
+app_check = FastAPI(check_unknown=True)
+
+client_no_check = TestClient(app_no_check)
+client_ceck = TestClient(app_check)
```
hawkaa, line 10: Not a big deal, but should probably be `client_check` and not `client_ceck` 👍

mortbauer: yes you are right, changed that :)

philipbjorge:
@mortbauer @boyaps @diogoduartec --
We'd love to have this feature for our API product.
Do you know what it will take to move this across the finish line?

Thanks -- Appreciate all the hard work you're putting into FastAPI!

gergelyhunyady:
@mortbauer @boyaps @diogoduartec 
I also believe this is a reasonable optional feature to add. Could serve the good in many use-cases.

As far as I understand the proposed implementation would allow users to specify `check_unknown` either for the whole service (via passing it to the `FastAPI` constructor) or alternatively they could specify it for specific `APIRouter`s only. I think this is pretty nice. Next to that I think it would be also reasonable if one could go even more fine grained by specifying the `check_unknown` behavior for specific path operations exclusively if needed (`@router.get("/users", ..., check_unknown=True)`). I didn't properly investigate if it fits well to the implementation of fastapi or if it would be hard to implement, so this is just my view from the user perspective.

Another point I would stress is the name of the parameter. I feel the `check_unknown` name is a bit too general. If the behavior strictly means only that unknown parameters will be caught and and an error response will be returned to the caller (which I think is the reasonable behavior), then I think something like `check_unknown_parameters` or maybe even better `forbid_unknown_parameters` (which would be in line with the `forbid` setting of the `extra` field in the pydantic model config).

Thanks for your work! What's your opinion on my two cents?

hawkaa:
@gergelyhunyady I agree on the naming part! And if we can align the naming with `pydantic` that is even better.

As for the specific path keyword, I have no strong opinions.

gguta:
Are there any progress on this?

fastapi/dependencies/utils.py
```diff
@@ -652,7 +672,12 @@ async def request_body_to_args(
         if field_alias_omitted:
             received_body = {field.alias: received_body}
 
+        if received_body is not None:
+            left_body_params = set(received_body)
+        else:
+            left_body_params = set()
```
cikay, line 678:
```py
left_body_params = set(received_body or {})
```

fastapi/dependencies/utils.py
```diff
@@ -641,6 +660,7 @@ def request_params_to_args(
 async def request_body_to_args(
     required_params: List[ModelField],
     received_body: Optional[Union[Dict[str, Any], FormData]],
+    check_unknown: bool = False,
```
cikay, line 663: Better to name `is_unknown`.

fastapi/dependencies/utils.py
```diff
@@ -571,12 +580,22 @@ async def solve_dependencies(
     values.update(header_values)
     values.update(cookie_values)
     errors += path_errors + query_errors + header_errors + cookie_errors
+
+    if check_unknown:
+        dependant_names = {param.name for param in dependant.query_params}
+        left_query_params.difference_update(dependant_names)
+        if outer_level:
```
cikay, line 587: Merge this condition to the upper one. Avoid nested conditions if you do not have to.

cikay (changes requested):

pacoispaco:
I am also interested in this feature. However shouldn't the default status code returned be a 404 and not a 400. Query parameters are part of the URL that identify a resource. And in the case of an unexpected query parameter the URL, by definition, would be identifying a resource that does not exist.

BilalAlpaslan (approved):

danhje:
Just noticed the application I'm currently working on has been querying an internal API with an old query parameter name for months, without anyone noticing, resulting in reduced quality of it's output. Will definitely make use of this feature when it is available!

yezz123 (approved):

stat-johan:
I would also greatly appreciate this feature

grzracz:
Would also deeply appreciate this.
We have a freely available API and because of "bad" query parameters, bad actors are able to bypass cache.
Instead of 
```
/assets/X/price?currency=USD
```
they are calling
```
/assets/X/price?currencyVNO1O1X114
```
adding a random string. I'd love to return a 400 when something like this is attempted.

Toby14:
Would also appreciate this feature as users would know what was wrong if they make a typo or unknown param.

xsway:
I also definitely see a use for this. Just wanted to say that a fail on validation is not maybe needed, if there is be a way to show a warning, that would be sufficient imo.

RonC-LL:
I'm just here to lend my support for this feature.  Misspelling a param bit me in the butt as well. Fortunately, it's not production code or anything.

YuriiMotov:
@mortbauer, thanks for your contribution!

Now we can implement this logic using Pydantic models (see code example in details).
Seems that it should be enough to cover such use-cases. What do you think?

<details>

```python
from typing import Annotated
from pydantic import BaseModel, ConfigDict

app_check = FastAPI(check_unknown=True)

client_check = TestClient(app_check)


class BaseExtraForbid(BaseModel):
    model_config = ConfigDict(extra="forbid")


class Item(BaseExtraForbid):
    data: str


class BodyParams(BaseExtraForbid):
    item: Item
    item2: Item


class QueryParams(BaseExtraForbid):
    param1: str
    param2: int = 0


@app_check.post("/with-extra-forbid")
async def endpoint_check(
    body: BodyParams,
    query: Annotated[QueryParams, Query()],
):
    return {
        "item": body.item,
        "item2": body.item2,
        "param1": query.param1,
        "param2": query.param2,
    }


def test_unknown():
    response = client_check.post(
        "/with-extra-forbid",
        json={"item": {"data": "myitem"}, "item2": {"data": "item2"}, "data": "any"},
        params={"param1": "qwerty", "limit": 12},
    )
    assert response.status_code == 422, response.text
    assert response.json() == {
        "detail": [
            {
                "loc": ["query", "limit"],
                "msg": "Extra inputs are not permitted",
                "type": "extra_forbidden",
                "input": "12",
            },
            {
                "loc": ["body", "data"],
                "msg": "Extra inputs are not permitted",
                "type": "extra_forbidden",
                "input": "any",
            },
        ]
    }
```

</details>

PR #1299, opened 2020-04-22
Fixed Typo in [EN] tutorial: body-fields

### LINKED ISSUE #1296: [BUG] Typo in [EN] tutorial: body-fields
### Describe the bug

There is a typo in `fastapi/blob/master/docs/en/docs/tutorial/body-fields.md` under `## Add Extra Information:` 

> ...
> You will learn more about it later to declare examples examples.

### To Reproduce

Look for `examples examples` in https://fastapi.tiangolo.com/tutorial/body-fields/#add-extra-information

### Expected behavior

It should probably be a link to somewhere although it is not really clear where to.

chris-allnutt:
Removed duplicate examples word in body documentation

resolves #1296

### DIFF
--- a/docs/en/docs/tutorial/body-fields.md
+++ b/docs/en/docs/tutorial/body-fields.md
@@ -39,7 +39,7 @@ You can then use `Field` with model attributes:
 
 You can declare extra information in `Field`, `Query`, `Body`, etc. And it will be included in the generated JSON Schema.
 
-You will learn more about it later to declare examples examples.
+You will learn more about it later to declare examples.
 
 ## Recap
 
--- a/docs/en/docs/tutorial/body-fields.md
+++ b/docs/en/docs/tutorial/body-fields.md
@@ -39,7 +39,7 @@ You can then use `Field` with model attributes:
 
 You can declare extra information in `Field`, `Query`, `Body`, etc. And it will be included in the generated JSON Schema.
 
-You will learn more about it later to declare examples.
+You will learn more about adding extra information later in the docs, when learning to declare examples.
 
 ## Recap
 


docs/en/docs/tutorial/body-fields.md
```diff
@@ -39,7 +39,7 @@ You can then use `Field` with model attributes:
 
 You can declare extra information in `Field`, `Query`, `Body`, etc. And it will be included in the generated JSON Schema.
 
-You will learn more about it later to declare examples examples.
+You will learn more about it later to declare examples.
```
chenl, line 42: Does the new sentence make sense to you? If it is only a matter of duplicate `examples` , should this be:

> You will learn more about declaring examples later.

Was that the original meaning of this sentence? Shouldn't it have been a link to somewhere?

chris-allnutt: I'm still new to fastapi.  I think your wording is definitely clearer, IF we were to add a link it would probably go here https://fastapi.tiangolo.com/tutorial/schema-extra-example/#field-additional-arguments  However I can't find any current examples of in document linkage and would prefer not to add something that would break.  So I'm thinking the way forward is just to use your wording.  Thoughts?

chenl: Since I was unsure of the original intent of the text, and I also am unsure of where this should link to and what is the proper syntax for such a link, I created an issue and not a PR.

I think it is better to get an authoritative answer for this issue than take a guess that might not be correct.

chenl: BTW, see an example for an internal link in:

https://github.com/tiangolo/fastapi/blob/master/docs/en/docs/tutorial/dependencies/dependencies-in-path-operation-decorators.md#dependencies-for-a-group-of-path-operations:

```markdown
[Bigger Applications - Multiple Files](../../tutorial/bigger-applications.md){.internal-link target=_blank}
```

tiangolo:
Thanks! I updated it a bit to try and clarify what was the original intention for the meaning of the phrase.

Thanks for your contribution @chris-allnutt ! :rocket: :cake: 

And thanks for the review @chenl ! :coffee:

PR #1316, opened 2020-04-24
Add optional custom Swagger css/js/favicon urls for docs.

jtbaker:
I opened this PR to allow the optional passing of a custom CSS url as a theme, as there are many such themes available. http://ostranme.github.io/swagger-ui-themes/

I didn't want to clutter up the named args in the `FastAPI` constructor, so I am just testing to see if the argument exists in the `self.extra` dictionary. Let me know what you think, if you'd like changes to the PR for the functionality, or aren't willing to accept.

Thanks again for the great work in this project, I really enjoy it!

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -110,11 +110,22 @@ class FastAPI(Starlette):
         if self.openapi_url and self.docs_url:
 
             async def swagger_ui_html(req: Request) -> HTMLResponse:
+                swagger_keys = [
+                    "swagger_js_url",
+                    "swagger_css_url",
+                    "swagger_favicon_url",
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for your interest!

I think that for additional UI modifications it would be best to add a new path operation with them, as described here: https://fastapi.tiangolo.com/advanced/extending-openapi/#create-a-path-operation-to-test-it

It's fairly easy to add a custom path operation with a new UI, and supporting all the possible combinations of additional customizations to the UI would become difficult.

So I'll pass on this one. But thanks for your effort and contribution! :coffee: :cake:

PR #1326, opened 2020-04-26
Typo in [en] tutorial: exception handeling

chenl:


alexvanzyl:
Not to be a grammar troll but I believe the correct statement would be:

> These handlers are in charge **of**  returning the default JSON responses...

tiangolo:
Good catch, thanks @chenl ! :mag:

And yep, agreed @alexvanzyl , thanks for the note. I just updated it accordingly. :heavy_check_mark: 

Thanks! :rocket: :tada:

PR #1330, opened 2020-04-27
🌐 Portuguese translation of `docs/async.md`

Serrones:


cassiobotaro (approved):
Just a nit: Crush is an English word, I think it should be italic.

tiangolo:
Thanks! :rocket: 

Do you guys think "crush" is understandable? I just updated the Spanish version to use "la persona que te gusta", which is kinda long :disappointed: but there's no better alternative.

I don't know if in Portuguese there's a better word for that, would people normally say "crush"? Or would it be better to make it something like "a pessoa da que você gosta" or similar?

I also added a lot of emojis of :heart_eyes: to later refer to that person as "esa persona :heart_eyes: " instead of writing the whole thing again.

Maybe we could do the same for the Portuguese version?

Here's the new Spanish version: https://github.com/tiangolo/fastapi/blob/master/docs/es/docs/async.md

Serrones:
> Thanks! 
> 
> Do you guys think "crush" is understandable? I just updated the Spanish version to use "la persona que te gusta", which is kinda long  but there's no better alternative.
> 
> I don't know if in Portuguese there's a better word for that, would people normally say "crush"? Or would it be better to make it something like "a pessoa da que você gosta" or similar?
> 
> I also added a lot of emojis of  to later refer to that person as "esa persona  " instead of writing the whole thing again.
> 
> Maybe we could do the same for the Portuguese version?
> 
> Here's the new Spanish version: https://github.com/tiangolo/fastapi/blob/master/docs/es/docs/async.md

Hey @tiangolo , here in Brazil it's common use some words in english mixing with portuguese sentences and expressions. `Crush` is a word that is cool and natural nowadays. We have a word in portuguese that translates perfectly crush, `paquera`. But it is so old, kkkkk. I wil talk with @cassiobotaro  and @marcosmmb to see if we keep `crush` or replace with a portuguese word that reflects the exact meaning. And about the emoji  heart_eyes  is amazing.
Thanks for the notes.
Cheers

cassiobotaro:
I think we should keep _crush_ in italic. As you said, "is a word that is cool and natural nowadays".

tiangolo:
Awesome, sounds great with _crush_ then. :rocket:

stlucasgarcia (commented):
Just a small typo and correct me if I am wrong but I believe that this page isn't in mkdocs.yml yet

stlucasgarcia (approved):

tiangolo:
Thanks @Serrones for your contribution! 🚀

And thanks @cassiobotaro and @lsglucas for the reviews! :cake:

PR #1336, opened 2020-04-28
Change password recommendation in docs

skorokithakis:
I think we should discourage password storage even in examples.

I might refactor the rest of the docs to include bcrypt as well.


skorokithakis:
Hmm, I see there are multiple places where these examples are referenced, so it will take some work to change everything. Please consider this change for now and I'll open another PR later.

tiangolo:
Cool, thanks! :lock: :rocket:

PR #1344, opened 2020-04-28
Fix minor erratum in Question issue template

retnikt:
In the Question issue template, the Additional Context section reads:
>  Add any other context or screenshots about the feature request here.

but it is a question not a feature request


tiangolo:
Good point, thanks! :heavy_check_mark: :rocket: :cake:

Issue #1350, opened 2020-04-29
[BUG] "Error getting request body" should not be logged as an ERROR


dbanty:
### Describe the bug

Generally, the semantics of the "ERROR" log level seem to mean that there is an error in your code- not a client error. At least this is how I interpret it and how Sentry (popular error logging framework) interprets it by default.

I believe that a client passing a bad input is an expected case for any API, that's why we have so much validation! I believe we should either not be logging anything at all in this case (to be consistent with other input validation errors, like Pydantic errors) or downgrade the logging level to still record _that_ it happened, without suggesting it's a problem that needs to be addressed in code.

According to [this page](https://docs.python.org/3/howto/logging.html) in the Python docs, an "ERROR" levels should be used when "Due to a more serious problem, the software has not been able to perform some function." I would argue that FastAPI is performing exactly its function by validating the input structure.

### To Reproduce

1. Create a file with:

```Python
from fastapi import FastAPI
from fastapi.params import Body

app = FastAPI()


@app.get("/")
def read_root(strings: List[str] = Body(...)):
    return {"Hello": "World"}
```

3. Open the browser and call the endpoint `/` with the payload {heelo: world} (or any malformed JSON).
4. It returns a status 400 (as expected).
5. An "ERROR" is logged, indicating to me and Sentry that there was a problem in our code.

### Expected behavior

Either nothing at all would be logged (because this is an expected case that could often happen) or the log level will be lower- indicating that it happened but that it's not necessarily a problem.

### Environment

- OS: macOS and Linux
- FastAPI Version [e.g. 0.3.0], get it with:  0.52.0

- Python version, get it with: 3.8.2

### Additional context

I'm currently filtering out this log with something like the following so that Sentry will stop alerting me to it:

```python
from fastapi import logger

class NoRequestErrorFilter:
    def filter(self, record):
        if record.msg.startswith("Error getting request body"):
            return 0
        return 1

logger.logger.addFilter(NoRequestErrorFilter())
```

tiangolo:
Impeccable bug description and argument :nerd_face: :bug:

Yep, totally agreed. Thanks for implementing the fix!  :heavy_check_mark: :rocket:

This should be fixed by #1351, released in the next version, `0.56.0` (in a couple of hours).

Kludex:
@tiangolo `0.56.1` 😋

tiangolo:
Ah, yep :sweat_smile: Fixed in `0.56.1` :rocket:

PR #1351, opened 2020-04-29
Remove error log when parsing malformed JSON body

dbanty:
Addresses #1350


tiangolo:
Yep, good point. Awesome, thank you! :rocket: :tada: :coffee:

PR #1354, opened 2020-04-30
Better JSON Decode error handling

aviramha:
I was annoyed by the fact that if JSON parsing fails for the body you receive HTTP Exception 400.
I think RequestValidationError is a better suit for the case, for the client to realize what have he done wrong exactly (position of wrong character) and body to be part of the exception (so it can be logged)

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1,5 +1,6 @@
 import asyncio
 import inspect
+import json
 from typing import Any, Callable, Dict, List, Optional, Sequence, Set, Type, Union
 
 from fastapi import params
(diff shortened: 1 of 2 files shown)


aviramha:
@tiangolo Seems the new handling leaves the generic exception handling uncovered. Not sure how to trigger that exception now tho, would assume connection breaking or something similar. ideas?

aviramha:
Added missing test for full coverage, pending CR @tiangolo .

tiangolo:
Awesome! Very nice :ok_hand: :sparkles: 

Impeccable implementation and tests :heavy_check_mark: :rocket: 

PR #1361, opened 2020-05-01
upgrade starlette dependency to 0.13.4

rushton:
This  upgrade is mostly to address the issue with StreamingResponse not killing the request when the client unexpectedly disconnects from the stream.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 requires = [
-    "starlette ==0.13.2",
+    "starlette ==0.13.4",
     "pydantic >=0.32.2,<2.0.0"
 ]
 description-file = "README.md"


mumumumu:
This is also needed to support the `SameSite` flag when setting cookies (mentioned in https://github.com/tiangolo/fastapi/issues/1099 and https://github.com/tiangolo/fastapi/issues/480).

victoraugustolls:
@tiangolo is this ok? Really looking forward to use newest starlette version for the fixes.

ziglef:
If this is ok, I'm also looking forward to use the newest starlette version (mostly for samesite cookies)

euri10:
> If this is ok, I'm also looking forward to use the newest starlette version (mostly for samesite cookies)

same !

tiangolo:
Great! Thank you @rushton ! :rocket: 

This will be available in FastAPI version `0.57.0`, (to be released in a couple of hours).

PR #1365, opened 2020-05-02
Fix callable class generator dependencies

### LINKED ISSUE #1204: [BUG] Class based dependencies with __call__ don't work with yield
### Describe the bug

Class based callable dependencies (https://fastapi.tiangolo.com/advanced/advanced-dependencies/) don't work when used as generators with yield keyword.

### To Reproduce

```Python
class SessionScope:

    def __init__(self, uri):
        self.engine = create_engine(uri)
        self.session_maker = scoped_session(sessionmaker(bind=self.engine))

    def __call__(self):
        session = self.session_maker()
        try:
            yield session
        except:
            session.rollback()
            raise
        finally:
            self.session_maker.remove()

get_db = SessionScope('postgresql://postgres:mypass@127.0.0.1:5432/my_db')

@router.get('/', response_model=List[User])
def read_users(
    db: Session = Depends(get_db),
    skip: int = 0,
    limit: int = 100,
):
    """
    Retrieve users.
    """
    users = crud.user.get_multi(db, skip=skip, limit=limit)
    return users
```

### Expected behavior

Dependency must return yielded session, but now it returns \_\_call\_\_ generator instance

### Additional context

The problem lies in /fastapi/dependencies/utils.py file on line 498:
```Python
        elif inspect.isgeneratorfunction(call) or inspect.isasyncgenfunction(call):
            stack = request.scope.get("fastapi_astack")
            if stack is None:
                raise RuntimeError(
                    async_contextmanager_dependencies_error
                )  # pragma: no cover
            solved = await solve_generator(
                call=call, stack=stack, sub_values=sub_values
            )
```
Paramter _call_ passed to _inspect.isgeneratorfunction_ is instance of a class instead of it's \_\_call\_\_ method and this function always returns False.

mrosales:
Fixes/related to #1204 

Resolving dependencies uses `inspect.isgeneratorfunction` and `inspect.isasyncgenfunction` which match async methods and functions, but not a class (or other callable) that has an async generator `__call__` method.

It also adds 4 new test cases to ensure that dependencies work as expected for class methods that yield instead of return values.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -429,6 +429,14 @@ async def solve_generator(
     if is_gen_callable(call):
         cm = contextmanager_in_threadpool(contextmanager(call)(**sub_values))
     elif is_async_gen_callable(call):
+        if not inspect.isasyncgenfunction(call):
+            # asynccontextmanager from the async_generator backfill pre python3.7
+            # does not support callables that are not functions or methods.
+            # See https://github.com/python-trio/async_generator/issues/32
+            #
+            # Expand the callable class into its __call__ method before decorating it.
+            # This approach will work on newer python versions as well.
+            call = getattr(call, "__call__", None)
         cm = asynccontextmanager(call)(**sub_values)
     return await stack.enter_async_context(cm)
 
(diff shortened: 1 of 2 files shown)


mrosales:
Looks like using the `asynccontextmanager` decorator with a non function/method callable isn't supported on `python3.6`.  I'll update the PR with a workaround.

mrosales:
@tiangolo this is ready to review if you'd like to take a look

tiangolo:
Excellent job @mrosales ! :bow: :rocket: :nerd_face: 

Including all the investigation with the backport, very clever!

PR #1370, opened 2020-05-04
Replace 400 codes with 401 in security tutorial

MatthewScholefield:
[From this](https://stackoverflow.com/questions/32752578/whats-the-appropriate-http-status-code-to-return-if-a-user-tries-logging-in-wit) it looks like 401 is more appropriate.

Feel free to close if you think 400 is better (ie. to emphasize it can be customized).

Side note: Looks like PRs are piling up. Just an idea, perhaps you would benefit from giving more people merge access so you don't get bored of all these tiny docs PRs. Something I read once talked about someone's philosophy of giving repo write access to every single contributor since they argued this would increase the quality of contributions and bad actors are very sparse. Of course this wouldn't be wise in a larger repository like this, but I think it puts the risk of bad actors/suboptimal code contributions versus PR suffocation into perspective.

### DIFF
--- a/docs_src/security/tutorial003.py
+++ b/docs_src/security/tutorial003.py
@@ -66,7 +66,7 @@ async def get_current_user(token: str = Depends(oauth2_scheme)):
 
 async def get_current_active_user(current_user: User = Depends(get_current_user)):
     if current_user.disabled:
-        raise HTTPException(status_code=400, detail="Inactive user")
+        raise HTTPException(status_code=401, detail="Inactive user")
     return current_user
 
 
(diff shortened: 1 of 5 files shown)


tiangolo:
Thanks! The problem is that a 401 _must_ also return a `WWW-Authenticate` header with a "valid challenge".

So, for example, when using the integrated HTTPBasic or OAuth2 utilities, that's what they do.

But if someone returns that code, they would also have to make sure to return a valid header with a valid challenge. That could be cumbersome in many cases as it's not obvious what to put there.

On the other side, a 400 error that describes a "generic" client error is still valid.

So I'm gonna pass on this, but thanks for your effort! :cake:

PR #1385, opened 2020-05-06
Added test for repeating cookies in response headers

scottsmith2gmail:
Found this issue using fastapi.  I haven't tracked down the cause.  It seems every additional layer of redirection doubles the cookies in the header.  No clue if this would actually break anything, but it seems like incorrect behavior at any rate.

### DIFF
--- a/tests/test_repeated_cookie_headers.py
+++ b/tests/test_repeated_cookie_headers.py
@@ -1,0 +1,33 @@
+from fastapi import Depends, FastAPI, Cookie, status, Response
+from fastapi.testclient import TestClient
+
+
+app = FastAPI()
+
+
+def set_cookie(*, response: Response):
+    response.set_cookie('cookie-name', 'cookie-value')
+    return {}
+
+
+def set_indirect_cookie(*, dep: str = Depends(set_cookie)):
+    return dep
+
+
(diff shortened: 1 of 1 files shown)


obataku:
the problem is that a call to `solve_dependencies` shares a [single `Response`](https://github.com/tiangolo/fastapi/blob/master/fastapi/dependencies/utils.py#L440-L446
) with [its recursive calls for child dependencies](https://github.com/tiangolo/fastapi/blob/master/fastapi/dependencies/utils.py#L472-L491), which may cause the original to have its `headers.raw` extended redundantly:
```
            dependant=use_sub_dependant,
            ...
            response=response,
            ...
        )
        (
            ...
            sub_response,
            ...
        ) = solved_result
        sub_response = cast(Response, sub_response)
        response.headers.raw.extend(sub_response.headers.raw)
        if sub_response.status_code:
            response.status_code = sub_response.status_code
```
one obvious solution that comes to mind is to check whether `sub_response is not response` before using it to update `response`

tiangolo:
Thanks for discovering this and creating the tests @scottsmith2gmail ! :detective: :bug: :heavy_check_mark: 

I just merged the fix by @obataku including your tests, so we can close this one. :rocket:

PR #1386, opened 2020-05-07
Bugfix for duplicated headers from indirect dependencies

obataku:
addresses the issue discovered by @scottsmith2gmail in #1385 and adds the test he proposed

> It seems every additional layer of redirection doubles the cookies in the header.


tiangolo:
Awesome! Great job debugging that :bug: :detective: :mag: 

And thanks @scottsmith2gmail for discovering the problem and writing the tests :heavy_check_mark: :+1:

PR #1395, opened 2020-05-10
test function name

chenl:
We are testing that creating an **item** that already exists should fail, not that creating an item with a **token** that already exists should fail.


tiangolo:
Ah! Good point :nerd_face: :mag:

Thanks!

PR #1412, opened 2020-05-14
Make openapi models honor response_model_by_alias

ZlatyChlapec:
Fixes models in openapi using aliases even tho `response_model_by_alias` is set to `False`.


ZlatyChlapec:
I added test as well.

ZlatyChlapec:
Even after this PR `openapi["components"]["schemas"]["Model"]["properties"]` still has wrong `title` `{'name': {'title': 'Alias', 'type': 'string'}}`. It can be fixed by specifying `title` argument to `Field` object and it feels to me like bug of pydantic.

tiangolo:
Hey @ZlatyChlapec , thanks for your contribution and interest!

So, `response_model_by_alias` is only there to be used as a quick hack, but the proper to do it and have full OpenAPI support would be to create another model, possibly inheriting from the original one, and setting the new/different alias in it.

Nevertheless, I used your commits and updated the tests for `response_model_by_alias` here https://github.com/tiangolo/fastapi/pull/1642 to ensure that it works correctly.

Thanks for your effort! :coffee: :cake:

PR #1429, opened 2020-05-18
🐛 Support custom OpenAPI / JSON Schema fields in the generated output OpenAPI

jacobsvante:
Pydantic supports adding custom fields other than `example` and `examples` using `BaseModel.Config.schema_extra`. This change ensures that those extra fields exist in the FastAPI generated Open API schema as well.

This is in line with what the current docs say:
> Other info
> The same way, you could add your own custom extra info that would be added to the JSON Schema for each model, for example to customize a frontend user interface, etc.

(https://fastapi.tiangolo.com/tutorial/schema-extra-example/#other-info)


tiangolo:
Good point, thanks @jmagnusson! Thanks for your contribution! :coffee: 

And thanks for the patience :sweat_smile:

johnthagen:
Are custom fields supported by the OpenAPI spec?

I saw breakages using [OpenAPI generator](https://openapi-generator.tech/)'s schema validation when custom fields started appearing in the generated schema starting in FastAPI 0.66.0: https://github.com/collerek/ormar/issues/291

jmagnusson:
> Are custom fields supported by the [OpenAPI spec](https://swagger.io/specification/)?

Yes, see Specification Extensions here: https://swagger.io/specification/. All `x-` prefixed keys are allowed (though they mentioned that not all tooling supports it.)

Issue #1359, opened 2020-04-30
[QUESTION] Validation in the FastAPI response handler is a lot heavier than expected


grigi:
### First check

* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.

### Description

So I have built a Tortoise ORM to Pydantic adaptor, and it's about stable, so I started profiling and found some interesting.

Pydantic will validate the data I fetch from the DB, which seems redundant as the DB content is already validated. So we are doing double validation.
Further profiling I found that the majority of time is spent by FastAPI preparing the data for serialisation, and then validating it, and then actually serialising (This specific step is what https://github.com/tiangolo/fastapi/issues/1224#issuecomment-617243856 refers to)

So I am doing essentially triple validation...

I then saw that there is the orjson integration, I tried that... and it made no difference that I could tell. (I'll get to this later)

I did a few experiments (none of them properly tested, but just to get an idea) with a simple benchmark:
(The database was populated with 200 junk user profiles generated by hypothesis, response is 45694 bytes)

Key:
R1 → Using FastAPI to serialise a `List[User]` model automatically (where `User` is a Pydantic model)
R2 → Using FastAPI to serialise a `List[User]` model automatically, but disabled the validation step in `serialize_response`
R3 → Manually serialised the data using an `ORJSONResponse`
R4 → Using FastAPI to serialise a `List[User]` model automatically, bypassed the `jsonable_encoder` as I'm serialising with orjson
R5 → Using FastAPI to serialise a `List[User]` model automatically, bypassed both validation and `jsonable_encoder`
C1 → Use provided pydantic `from_orm` 
C2 → Custom constructor that doesn't validate

My results are:
R1 + C1 → 42req/s
R1 + C2 → 43req/s  (Seems the 3 FastAPI steps overpower the validation overhead of `from_orm`)
R2 + C1 → 56req/s (Disabling the validation IN FastAPI has a much bigger impact?)
R2 + C2 → 63req/s (So, no extra validation)
R3 + C1 → 75req/s
R3 + C2 → 160req/s
R4 + C1 → 53req/s (This orjson-specific optimization gave us a 26% speedup here!)
R4 + C2 → 64req/s (This orjson-specific optimization gave us a 48% speedup here!)
R5 + C1 → 74req/s (So, almost as fast as bypassing the FastAPI response handler)
[...]

PR #1434, opened 2020-05-19
[WIP] Optionally skip json_encoder for orjson & response_model_skip_validation

grigi:
This PR is to address optimizations noted in #1359 

Done:
* [x] Add a `response_model_skip_validation` parameter that is to be set on a per-request basis. This will skip validation if `response_model` is set.
* [x] Skip `jsonable_encoder` if the `Response` has `skip_jsonable_encoder = True` set and various other conditions pass.
  We set this for `orjson` because as per https://github.com/tiangolo/fastapi/issues/1224#issuecomment-617243856 it appears to handle many common python datatypes. We added `decimal.Decimal` as that was the only one missing.
* [x] If a `response_model` has been specified we require `response_model_skip_validation` before allowing skipping of `jsonable_encoder`, as it skips coercion.
  As the user just said they know what they are doing. Right?
* [ ] Allow skipping validation globally via configuring the app.
* [ ] Update docs

Motivation is based on benchmarks (more details at #1359) I performed.
Some conclusions are:
* Skipping `jsonable_encoder` can increase performance by over 35% (probably more if no `response_model` has been specified.
* Skipping validation can increase performance by 30-50%
* Skipping both can increase performance by 140% (https://github.com/tiangolo/fastapi/issues/1359#issuecomment-622998691 provides an explanation)


grigi:
After trying to write tests for this I realise that my assumptions didn't take into account non-complete or non-coerced responses.

Seems that if the response is already a pydantic object it qualifies for the jsonable_encoder to be skipped. (I did tests on this, and it seems to hold)

Looking at the pydantic `ModelField` it's not apparent how to test if the input is of type in the ModelField, of it it is a list/dict.

Seems like we can only do this skipping if validation is turned off for the response?

victoraugustolls:
Hi @grigi and @tiangolo ! Do we have any update here? Looking forward to this for a long time!

grigi:
Hi, I forgot about this sometime during having covid. Thanks for reminding me about this.

@tiangolo I have a few questions:

Assumptions (please correct me if I'm wrong):
* A response can either be a native Pydantic model object(`PMO`), OR a `dict`, OR a manual `Response` type.
* We have no control over a manual `Response` type.
* I'm assuming that `dict` responses will want to be validated and turned into a `PMO` before serialised by default. There is cases where the user wants to say "trust me" and turn off validation, and we should allow this with a global setting.

Questions:
* Can we assume that a native `PMO` is already validated? (You have to jump through serious loops to build an unvalidated `PMO`, and if you do that, one can argue that it's your perogative to ensure that things are correct). If so, can we safely skip the validation on `PMO`'s by default?
* Why did you replace the JSON encoder with your own one, instead of using the provided one from Pydantic: https://pydantic-docs.helpmanual.io/usage/exporting_models/#modeljson ? It already seems to handle all the common data type conversions correctly? Based on quick tests, it appears to be faster too.


Depending on the answer you provide, I can complete this PR. It would be lovely to have a faster response type for the cleaner code by default.

Right now returning a `PMO` (which is the cleanest in code) will result in the slowest response type.
`dict`'s are slightly faster, and `Response` is the fastest.

I would like making `PMO` perform better by default, skip validation and re-building of `PMO`, so we can just go to serialisation directly?

I would like to make the default `JSONResponse` use the provided `PMO.json()` if available so we can honour custom encoders on it.

I would like to make the `ORJSONResponse` allow skipping the datatype conversion by default, as it seems to handle all the common cases as well.

BUT, I would like to limit breakage. So what would be safe behaviour be for `PMO` and `dict` response types?

michael0liver:
+1

phuong:
+1

raphaelauv:
fastAPI is a API framework doing data validation , so it's the default behavior -> https://github.com/tiangolo/fastapi/issues/1792#issuecomment-665165674

if you want skip validation for performance purposes then it should be explicit The documentation already explain perfectly this :

https://fastapi.tiangolo.com/advanced/response-directly/

`It won't do any data conversion with Pydantic models, it won't convert the contents to any type, etc.`

kolypto:
It's nice that FastAPI does data validation, but it's best done in development and testing environments only. 
In production, validating APIs that return lists of objects is a significant performance penalty

havardthom:
> Hi, I forgot about this sometime during having covid. Thanks for reminding me about this.
> 
> @tiangolo I have a few questions:
> 
> Assumptions (please correct me if I'm wrong):
> 
> * A response can either be a native Pydantic model object(`PMO`), OR a `dict`, OR a manual `Response` type.
> * We have no control over a manual `Response` type.
> * I'm assuming that `dict` responses will want to be validated and turned into a `PMO` before serialised by default. There is cases where the user wants to say "trust me" and turn off validation, and we should allow this with a global setting.
> 
> Questions:
> 
> * Can we assume that a native `PMO` is already validated? (You have to jump through serious loops to build an unvalidated `PMO`, and if you do that, one can argue that it's your perogative to ensure that things are correct). If so, can we safely skip the validation on `PMO`'s by default?
> * Why did you replace the JSON encoder with your own one, instead of using the provided one from Pydantic: https://pydantic-docs.helpmanual.io/usage/exporting_models/#modeljson ? It already seems to handle all the common data type conversions correctly? Based on quick tests, it appears to be faster too.
> 
> Depending on the answer you provide, I can complete this PR. It would be lovely to have a faster response type for the cleaner code by default.

Based on @tiangolo's answer here: https://github.com/tiangolo/fastapi/issues/1107#issuecomment-612963659, it seems that the pydantic json utility was created after the json_encoder implementation in FastAPI. It would be a great improvement to replace the current json_encoder with pydantic's own utility.

lslunis:
@grigi Any update on this? I’d love to be able to disable heavyweight response validation in production.

grigi:
@lslunis Unfortunately due to not getting answers from FastAPI developers I abandoned the project. I still believe that my findings were valid and the proposal made sense, but its more than a year later, and things might have changed.

Anyone is welcome to take this work and continue/reimplement it.

ghandic:
@tiangolo what are you thoughts on this? There is the potential for duplicate validation running and a performance hit see https://github.com/tiangolo/fastapi/discussions/3624#discussioncomment-1084046

MaximilianFranz:
> @tiangolo what are your thoughts on this? There is the potential for duplicate validation running and a performance hit see [#3624 (comment)](https://github.com/tiangolo/fastapi/discussions/3624#discussioncomment-1084046)

Same question for me, this is a very common requirement. Using a custom Response (i.e. `PydanticJSONResponse`) like so: 

```
class PydanticJSONResponse(JSONResponse):
        if content is None:
            return b""
        if isinstance(content, bytes):
            return content
        if isinstance(content, BaseModel):
            return content.json().encode(self.charset)
        return content.encode(self.charset)
```

is a solution, but could also be provided by fastAPI directly.

tiangolo:
Thanks for the effort @grigi and the conversation everyone!

So, first, a path operation can return anything, not just the types defined above, e.g. an ORM object. The jsonable_encoder in FastAPI will do its best to encode that to something JSON-able. If a Pydantic model is returned, there's no guarantee that it's already validated, the `response_model` can be used to filter data: https://fastapi.tiangolo.com/tutorial/response-model/#add-an-output-model

I want to add a way to allow users to customize how data is serialized and provide their own functions for serialization and deserialization. That would probably help here, as people would be able to use a function that doesn't use `jsonable_encoder` and just uses Pydantic's own mechanisms.

Also, with the soon-to-be-released Pydantic v2, a lot of the duplication cloning Pydantic fields in `response_model` and similar things will be avoided.

PR #1462, opened 2020-05-23
Cannot import ''OAuth2PasswordRequestFormStrict''

RichardHoekstra:
Fixes an import error:     

from fastapi.security import OAuth2PasswordBearer, OAuth2PasswordRequestFormStrict
ImportError: cannot import name 'OAuth2PasswordRequestFormStrict'

Currently using the workaround:
from fastapi.security.oauth2 import OAuth2PasswordRequestFormStrict

Not sure whether the maintainers intended for the import to be different from the rest of the security imports. I assume this is a simple oversight.

### DIFF
--- a/fastapi/security/__init__.py
+++ b/fastapi/security/__init__.py
@@ -11,6 +11,7 @@ from .oauth2 import (
     OAuth2AuthorizationCodeBearer,
     OAuth2PasswordBearer,
     OAuth2PasswordRequestForm,
+    OAuth2PasswordRequestFormStrict,
     SecurityScopes,
 )
 from .open_id_connect_url import OpenIdConnect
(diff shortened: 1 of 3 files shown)


phy25:
You may also want to change the corresponding documentation as well.

RichardHoekstra:
> You may also want to change the corresponding documentation as well.

I would but I don't know what corresponds.

phy25:
https://github.com/tiangolo/fastapi/search?q=OAuth2PasswordRequestFormStrict

juhovh-aiven:
I always found it weird that the strict version was not exported, I think this is a good improvement. :+1:

tiangolo:
Thanks for your contribution @RichardHoekstra ! :rocket: 

So, it was actually intentional, I would expect (and suggest) everyone to use the non-strict version, as the strict version enforces clients to send a constant value doesn't add much, as it's just a constant.

Because of that, I didn't expose it directly.

But... I think this is like the 3th PR suggesting that change :joy: 

So I guess it's gonna be more intuitive if it's just there even though I don't expect many people to use it.

Thanks for the review @phy25 ! :bow:

PR #1477, opened 2020-05-25
Remove HTTPDigest as it's not helpful

### LINKED ISSUE #1476: [BUG] HTTPDigest not standard compliant
### Describe the bug

`fastapi.security.HTTPDigest` is actually not implementing HTTP Digest Access Authentication as specified by [RFC 7616](https://tools.ietf.org/html/rfc7616) or (the obsoleted) [RFC 2617](https://tools.ietf.org/html/rfc2617).

### To Reproduce

Read the relevant RFCs or maybe try it out using `curl --digest`.

### Expected behavior

Exactly as specified in [RFC 7616](https://tools.ietf.org/html/rfc7616).

macheins:
Hey,

I removed `fastapi.security.HTTPDigest` as it is not standard compliant (see [RFC 7616](https://tools.ietf.org/html/rfc7616)), very confusing for users and not useful or helpful at all. E.g. `auto_error` raises `HTTP_403_FORBIDDEN` which is not part of the standard and also misses any required `WWW-Authenticate` headers. Due to the complexity when generating `WWW-Authenticate` (i.e. a good - oftentimes even application dependent - nonce) and validating the corresponding `Authorization` header (e.g. nonces really used once) I would suggest to leave the implementation details up to the user and not giving him (including me when I first looked at FastAPI) a false sense of support by "just passing on a string".

As no one else raised an issue yet, usage seems to be quite low anyways...

Merging this would close #1476


tiangolo:
Thanks for your interest, I would rather keep the incomplete version as it has the metadata needed for the correct OpenAPI integration and accept PRs improving the implementation rather than just removing it.

As you say, usage is probably very low anyways.

So I'll pass on this, but thanks for the effort.

macheins:
Thanks for your reply @tiangolo.

What I'm most worried about is the fact that the implementation is not just incomplete, but the `auto_error` part is actually doing something unrelated. May I propose to keep the interface, but to remove the `auto_error` part and force the user to implement the correct logic himself?

Actually as we are looking at security relevant code in case of authentication, I would strongly advise against keeping such code - especially it gives users a false sense of doing the right thing.

PR #1478, opened 2020-05-25
Remove HTTPBase as it is not used anywhere

macheins:
Hey,

while it was looking at `fastapi.security.http`, I noticed that `HTTPBase` is not used anywhere and creates an unnecessary level of complexity:

- models inheriting from `HTTPBase` do never call `super` and overwrite all properties themselves
- `fastapi.security` does not expose `HTTPBase`
- the docs do not mention it
- it's not really helpful as most non-"Basic" authentication schemes (see [IANA](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml)) require much more complex `auto_error` handling than `HTTPBase` can provide
- `HTTPBasic.__call__` is incompatible with `HTTPBase.__call__` which is really bad in a typed context and requires a `#type: ignore`

Therefore I hereby propose to remove `HTTPBase`.


tiangolo:
Thanks for your interest, but it is there as a basic stub for implementing other HTTP auth protocols on top.

And it has the necessary parts to be integrated with OpenAPI.

Of course, it is not implementing any specific protocol and only gives a bare-bones stub to build upon, but it could be used to simplify that initial scaffolding.

Also, comments like:

> it's not really helpful

are subjective and derogatory. Please restrain from using those.

Thanks for your effort, but I'll pass on this one.

PR #1510, opened 2020-06-04
Remove duplicate chinese articles

waynerv:
Since the content of the 2 Chinese articles in External Links and Articles is very simple, just a brief summary of 1 or 2 document chapters, as we now have some more detailed Chinese translation of docs, I suggest temporarily removing these 2 duplicate articles.


tiangolo:
I guess that would make sense :fire:

But as I don't really know the language I would like a second opinion, just to be sure. Maybe @yanever , @Laineyzhang55 , @phy25, @Dustyposa , @RunningIkkyu , what do you think?

phy25 (approved):
The first link is a very, very simple introduction. The second one is a blog's index page containing some fastapi-related articles, some of which is purely translation and some of which is the author's learning notes.

I would not say they are "duplicate", but they are not a complete walkthrough, or insight in a specific aspect. The translation of the docs should already been better than them. Just my personal opinion.

tiangolo:
Perfect! Thanks for the opinion @phy25 ! :bow: 

Thanks for the contribution @waynerv ! :rocket:

PR #1524, opened 2020-06-06
squash __root__ in BaseModel json encoding

### LINKED ISSUE #911: Pydantic __root__ model - incorrect handling
### Describe the bug

https://pydantic-docs.helpmanual.io/usage/models/#custom-root-types
Pydantic allows to create models with only `__root__` field. In such scenario the model behaves as transparent wrapper for this single type.

When such model is used in response (request also?) fastapi does not treat it correctly and renders it as object with `__root__` field.

Object is treated correctly by pydantic itself.

### To Reproduce

```
from typing import List
from fastapi import FastAPI
from pydantic.main import BaseModel

app = FastAPI()


class RootTestClass(BaseModel):
    __root__: List[str]


@app.get("/")
async def root():
    response = RootTestClass(__root__=['a', 'b', 'c'])
    print(response.json())  # ["a", "b", "c"] so it's OK
    print(RootTestClass.schema())  # {'title': 'RootTestClass', 'type': 'array', 'items': {'type': 'string'}} this is also OK
    return response  # Wrong value in http response
```

### Expected behavior

The response should be:
```
["a", "b", "c"]
```

but at the moment is:
```
{"__root__":["a","b","c"]}
```

### Screenshots

N/A

### Environment

- OS: Linux
- FastAPI Version: 0.47.1
- Python version: Python 3.7.5

### Additional context

N/A

patrickkwang:
Fixes #911.

This corrects the specific problem described by @peku33, where the response is received by the client as `{"__root__": foo}`. This change simply converts to `foo` during json encoding.

Note that these `__root__` structures may still exist internally, e.g. `{"foo": {"__root__": foo}}`. That is a pydantic bug (https://github.com/samuelcolvin/pydantic/issues/1414) with an open PR.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -71,6 +71,8 @@ def jsonable_encoder(
                 by_alias=by_alias,
                 skip_defaults=bool(exclude_unset or skip_defaults),
             )
+        if "__root__" in obj_dict:
+            obj_dict = obj_dict["__root__"]
         return jsonable_encoder(
             obj_dict,
             exclude_none=exclude_none,
--- a/tests/test_jsonable_encoder.py
+++ b/tests/test_jsonable_encoder.py
@@ -76,6 +76,10 @@ class ModelWithDefault(BaseModel):
     bla: str = "bla"
 
 
+class ModelWithRoot(BaseModel):
+    __root__: str
+
+
 @pytest.fixture(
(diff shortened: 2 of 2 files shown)


tiangolo:
Nice, thank you! :bug: :fire: :tada:

PR #1525, opened 2020-06-06
a template for use PUT or PATCH with files

ihakh:
the basic HTTP form only supports post
with this template the new FASTAPI user can quickly use POST or PATCH ...

### DIFF
--- a/docs_src/request_files/tutorial002.py
+++ b/docs_src/request_files/tutorial002.py
@@ -20,14 +20,24 @@ async def create_upload_files(files: List[UploadFile] = File(...)):
 async def main():
     content = """
 <body>
-<form action="/files/" enctype="multipart/form-data" method="post">
-<input name="files" type="file" multiple>
-<input type="submit">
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for your interest! :nerd_face: 

The idea there was actually to show a minimum example of how it could be done. Keeping the focus on the API part, and keeping the HTML as simple as possible.

I would actually expect users to use a framework like React or Vue.js for frontend instead of plain JS, but I wouldn't want to "enforce" it in the docs, the same way, I would prefer not to add very complex examples that I would expect developers to implement in a different way anyway.

So I'll pass on this. But thanks for your effort! :cake:

ihakh:
> The idea there was actually to show a minimum example of how it could be done.
my idea was same + adding more basic and simple funtionality

tiangolo:
Yep, get it, I would still expect it to vary a lot from team to team.

Nevertheless, it could be a good idea for a blog post! :memo:

PR #1526, opened 2020-06-07
remove unused f-string

kotamatsuoka:


Kludex (approved):

tiangolo:
Good point, thanks @kotamatsuoka ! :cake: 

And thanks for the review @Kludex :coffee:

PR #1534, opened 2020-06-08
Normalise IntEnums to ints for route status codes

### LINKED ISSUE #1349: [FEATURE] Support HTTPStatus

We typically use [HTTPStatus](https://docs.python.org/3/library/http.html#http.HTTPStatus) in our code. When used as a value for `status_code` in path functions this results in the string literal in the documentation and when "trying" the response is shown as undocumented because it doesn't match the actual response code

I want to use HTTPStatus enum values and have it work exactly like the `starlette.status` pseudo enum values

retnikt:
Fixes/related to #1349

### DIFF
--- a/docs/en/docs/tutorial/response-status-code.md
+++ b/docs/en/docs/tutorial/response-status-code.md
@@ -15,7 +15,10 @@ The same way you can specify a response model, you can also declare the HTTP sta
 !!! note
     Notice that `status_code` is a parameter of the "decorator" method (`get`, `post`, etc). Not of your *path operation function*, like all the parameters and body.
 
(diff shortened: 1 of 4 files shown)


retnikt:
@tiangolo do you have any guidance on where to test this? (as in, which `tests/test_*.py` file)

Kludex:
Probably on: `test_custom_route_class.py`
As is the only test that imports the `routing` module.

retnikt:
@Kludex I decided to put it in `tests/test_application.py` because that's where the OpenAPI schema seems to be tested.

tiangolo:
Great job @retnikt ! :bow: :nerd_face: :tada: 

Thanks @Kludex for the help! :coffee:

PR #1537, opened 2020-06-08
Add __all__ definitions to main fastapi module

### LINKED ISSUE #1309: [FEATURE] Add __all__ to fastapi

We use mypy with the option `implicit_reexport = False`. This triggers on idiomatic fastapi imports because `fastapi/__init__.py` reexports everything without declaring any in `__all__`

Add a `__all__` for all types that should be imported from the main module.

Direct imports from the respective submodules is a workaround but all examples assume an import from the main package

### Additional context

https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport

gjedlicska:
Added __all__ constraints to the main fastapi __init__.py file, to:
* Fixes https://github.com/tiangolo/fastapi/issues/1309
* Enable `implicit_reexport = False` configuration value to be set in mypy configuration.
* Limit what gets imported with `form fastapi import *`


tests/test_public_exports.py
```diff
@@ -0,0 +1,31 @@
+import fastapi
+
+
+def test_fastapi_module_exports():
```
Kludex, line 4:
```suggestion
def test_fastapi_module_exports():
"""
Test the publicly exported module set.
Benefits for declaring explicitly the public exports:
    * Using __all__ sets what gets imported with `from fastapi import *`
    * Enables mypy to pass all checks with fastapi when using `implicit_reexport = False` 
"""
import yourpackage

missing = set(n for n in yourpackage.__all__
              if getattr(yourpackage, n, None) is None)
self.assertEmpty(missing, msg=f"__all__ contains unresolved names: {missing}")
```
Reason: You're not checking if something else is being imported. 
Reference: http://xion.io/post/code/python-all-wild-imports.html

Kludex (changes requested):

tests/test_public_exports.py
Kludex: I don't know how to suggest a whole block of code, sorry :(

gjedlicska: this is a valid point. I'll add it a bit later. I also have to work on some submodule __all__ definitions

gjedlicska:
Thanks @Kludex this is a good suggestion.
I'll extend the test with this check.

I also have to add some submodule __all__ definitions, cause of recursive nesting.

Converted the PR to draft for these reasons.

gjedlicska:
@Kludex 
I need some help with this, most of the starlette reexports are done via importing it to a module, and than reexporting from the main fastapi module. The only exception is the StaticFiles class. Is there any design decision behind this, or is it just by mistake?
My solution now reexports the StaticFiles class as all the others via the fastapi init, and also implements __all__ for the fastapi.staticfiles module, to not break the existing public API.
But it introduces a new public export `from fastapi import StaticFiles` .

I think this is more consistent this way, without breaking existing code.
What do you think?

Kludex:
If you do what you're doing, we'll be able to import TestClient from fastapi i.e. `from fastapi import TestClient`. I don't know the decisions made about the allowance, but I think that's not what we want. I'm not sure if works, but can you check:
```
from . import testclient

__all__ = [
   ...
   "testclient",
   ...
]
``` 
I haven't tested, so I don't know if works, you need to try. I'm too tired now, see you tomorrow.

gjedlicska:
You are correct, my changes make the `from fastapi import TestClient` a valid import. Even for mypy.
If we don`t want that, which is also fine, then it the __all__ definition containing `TestClient` should stay in the `testclient.py` file. Nothing else is needed, we don't have to import every module to the top level package.

I raised the question, because I didn`t see the logic behind, what starlette reexports are done from the top level package, and what are done in submodules.

To compare:

Top level: `status, BackgroundTasks, Request, Response, WebSocket

Submodule: `StaticFiles, Jiinja2Templates, TestClient, FileResponse, HTMLResponse, JsonResponse ... in responses.py, WebSocketDisconnect`

There is a bit of mix here, like for responses. If it is intentional, than is fine too, than __all__ should be added to the reexports in the `responses.py`.  This also clarifies how the imports should be used to the package users.

gjedlicska:
@Kludex should we work on getting this PR merged? I don't want to be pushy on this, I know there are more important issues open.

Kludex:
Is just that I don't know what we want. Maybe @tiangolo 's awesome inputs can help here. 💪

tiangolo:
Thanks for your effort @gjedlicska ! :nerd_face: :coffee: 

And thanks for the discussion here @Kludex ! :nerd_face: :memo: 

I actually prefer not to have the `__all__` as it adds code duplication that can get out of sync easily. I would prefer `import x as x`, as any errors in that would be detected by the editor.

gjedlicska:
Completely understand, it can be a burden to maintain. I'm doing it on a few projects, and to work around the no editor support, my solution is a small test function, that checks the public modules in __all__ against what is imported into __init__.py for any discrepancies either direction.

Anyhow mypy can work both ways. 
Thanks for your reply.
I'm closing this.

gjedlicska:
https://github.com/tiangolo/fastapi/issues/1309

This should be closed aswell.

gjedlicska:
Hey @tiangolo, just saw, that this was merged for Typer https://github.com/tiangolo/typer/pull/147.
Should this be attempted for FastApi aswell or is it a long shot?

tiangolo:
Thanks for all the effort here @gjedlicska ! I added support for `mypy --strict` with a lot of type annotation improvements in #2547

It's available in FastAPI 0.63.0 :tada:

gjedlicska:
Thanks @tiangolo, glad I could help.

Issue #872, opened 2020-01-16
[FEATURE] I need a way to specify servers in the openapi spec


chaitanyanettem:
### Is your feature request related to a problem

I want to be able to use the generated openapi.json doc as it is and hook it up with a document publishing flow, but i'm not able to because I have to add in information about `servers` manually.

### The solution you would like

Someway to specify at a global level what the base server url should be.

### Describe alternatives you've considered

Currently I'm doing this manually in the generated openapi.json by adding in something like - 
```
    "servers": [
        {
            "url": "http://example.com"
        }
    ]
```

I don't mind submitting a PR which enables this if someone can guide me about the changes that need to be made. One thing I saw was that the `get_openapi` method in `fastapi.openapi.utils`, doesn't expose a parameter for setting a value for the `servers` key.

dmontagu:
It's @tiangolo's decision to make, but given this *is* part of the OpenAPI spec, I personally would be in favor of adding this as a keyword argument to `FastAPI`, and as an argument to `get_openapi`, making it easier to set this.

It should be a quick PR if you want to open it.

I think eventually we should group the arguments to `FastAPI` into more nested chunks to make it a little easier to parse, but I would be fine with the approach described above for now.

origliante:
Hey there it's needed also here

tiangolo:
Thanks for the discussion! :coffee: 

This was implemented in https://github.com/tiangolo/fastapi/pull/1547 :rocket: 

It will be available in FastAPI version `0.58.0` (in a couple of hours) :tada: 

I'm re-opening the issue to allow the original author ( @chaitanyanettem ) check, confirm, and close it :nerd_face:

chaitanyanettem:
Nice! I tried this just now. This solves my problem. Thank you @tiangolo! :smile:

LoLei:
Is there a way to remove the default relative server? I need the OpenAPI spec to *only* show the servers I pass via `FastAPI(servers=)`. But the first one in the spec is still the default relative server starting with `/`.

My Python code:

```python
api = FastAPI(
    servers=[
        {"url": "http://localhost:8000/api/v1"}
    ],
)
```

The OpenAPI spec JSON:

```json
  "servers": [
    {
      "url": "/api/v1" # <-- I do not want this
    },
    {
      "url": "http://localhost:8000/api/v1"
    }
  ]
```

LdeBerried:
Hi @LoLei 

[...]

PR #1547, opened 2020-06-10
feat: add servers option for OpenAPI

mikaello:
Closes/related to #872

The missing server option has been a drawback when it comes to the swagger UI (e.g. provide prod and test servers) and portability of the openapi.json to services that is based on knowing the servers.

Example usage:
```python
from fastapi import FastAPI

server1 = Server(url="http://example.com", description="optional description")
server2 = Server(url="http://test.com")

app = FastAPI(servers=[server1, server2])

@app.get("/")
async def root():
```

It would also work to just pass a dictionary, but then the typechecker won't be happy:
```python
from fastapi import FastAPI

app = FastAPI(servers=[{"url": "http://example.com", "description": "test"}])
```


tiangolo:
Thanks! :rocket: 

I updated it a bit to receive `dict`s by default. To not force users to import another class and create instances to be able to set the parameter. And added a test for it :heavy_check_mark: 

Thanks for your contribution! :tada: :coffee:

mikaello:
Thanks for updating the branch, I agree that dict is better as default.

PR #1548, opened 2020-06-11
🐛 Fix JSON Schema "not" keyword

v-do:
The JSON Schema "not" keyword expects a JSON Schema object instead of a `List`.

As per most recent 2019-09: https://tools.ietf.org/html/draft-handrews-json-schema-02#section-9.2.1.4

(and older draft 4: https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.5.6)


tiangolo:
Great! Good catch :mag: :detective: 

PR #1549, opened 2020-06-11
Set dependant security scopes if `dependency_overrides` is set

### LINKED ISSUE #737: [BUG] dependency_overrides does not play well with scopes
**Describe the bug**
When working with `Security()` dependencies, the scopes disappear when `app.dependency_overrides` is executed. The callable dealing with the scopes gets an empty list instead of the scopes.

**To Reproduce**

```python
from fastapi import FastAPI, Header, Security, Depends
from fastapi.security import SecurityScopes

app = FastAPI()

def get_user(required_scopes: SecurityScopes):
    print(required_scopes.scopes)

    return "John Doe"

def data():
    return [1,2,3]

def other_data():
    return [3,4,5]


@app.get("/test")
def test(user: str = Security(get_user, scopes=["foo", "bar"]), data = Depends(data)):
    return data

client = TestClient(app)
response = client.get("/test")

app.dependency_overrides[data] = other_data
response = client.get("/test")

# prints: ["foo", "bar"] and [] instead of ["foo", "bar"] and ["foo", "bar"]
```

**Expected behavior**
In the above example I expect `get_user()` to print the same scopes twice. Instead, before the `dependency_overrides` it prints the correct scpoes, but an empty list afterwards.

**Environment:**
 - OS: Linux
 - FastAPI Version 0.43.0
- Python 3.7.4

andrew222651:
This change follows the precedent of https://github.com/tiangolo/fastapi/blob/master/fastapi/dependencies/utils.py#L152

Alternatively, `security_scopes` could be set within `get_dependant` (within the `Dependant` constructor).

This should fix/related to #737

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -478,6 +478,7 @@ async def solve_dependencies(
                 name=sub_dependant.name,
                 security_scopes=sub_dependant.security_scopes,
             )
+            use_sub_dependant.security_scopes = sub_dependant.security_scopes
 
         solved_result = await solve_dependencies(
             request=request,
--- a/tests/test_dependency_security_overrides.py
+++ b/tests/test_dependency_security_overrides.py
@@ -1,0 +1,65 @@
+from typing import List, Tuple
+
+from fastapi import Depends, FastAPI, Security
+from fastapi.security import SecurityScopes
+from fastapi.testclient import TestClient
+
+app = FastAPI()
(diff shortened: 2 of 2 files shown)


tiangolo:
Awesome! Great job finding the fix! :bug: :detective: :mag: 

I just added tests for it and we're now good to go :ok_hand: 

Thanks for your contribution! :rocket: :cake:

PR #1553, opened 2020-06-11
Omit name in errors that is not used in request

### LINKED ISSUE #1088: Bad `loc` on validation error, if payload represended by one model
### Describe the bug

Really like your framework, but there is, indeed, an annoying issue with `loc` on validation error with one object as payload.

### To Reproduce
Code sample
```Python
from typing import List

app = FastAPI()

class NameModel(BaseModel):
    name: str

@app.post("/test", response_model=NameModel)
def test(obj: NameModel, ):  # bad
    return obj

@app.post("/test_embed", response_model=NameModel)
def test(obj: NameModel = Body(..., embed=True)):  # ok
    return obj

@app.post("/test_multiple", response_model=List[NameModel])
def test(obj1: NameModel, obj2: NameModel):  # ok
    return obj1, obj2
```

When you make a request to endpoint (`/test`) with the wrong payload (e.g.: `{}`), it always includes the variable name into error location, despite it has no relation to request.

It makes no sense, moreover, it complicates the logic for error printing on fronted, because they just don't know and not required to know the name of the backend`s internal variable.

```json
{
    "detail": [
        {
            "loc": [
                "body",
                "obj",
                "name"
            ],
            "msg": "field required",
            "type": "value_error.missing"
        }
    ]
}
```
it should be
```json
{
    "detail": [
        {
            "loc": [
                "body",
                "name"
            ],
            "msg": "field required",
            "type": "value_error.missing"
        }
    ]
}
```

With the embedded object (`/test_embed`) or multiple objects (`/test_multiple`), it works as expected, putting the variable name into location, because it should be in the payload.

### Expected behavior

Don't include the variable name into location error, if it is not reflected in schema / not embedded / not expected to be in payload.

### Environment

- OS: macOS
- FastAPI 0.52.0
- Python 3.6.8

andrew222651:
This should fix/related to #1088

### DIFF
--- a/docs/en/docs/tutorial/handling-errors.md
+++ b/docs/en/docs/tutorial/handling-errors.md
@@ -215,7 +215,6 @@ You will receive a response telling you that the data is invalid containing the 
     {
       "loc": [
         "body",
-        "item",
         "size"
       ],
       "msg": "value is not a valid integer",
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -623,9 +623,17 @@ async def request_body_to_args(
         field = required_params[0]
         field_info = get_field_info(field)
         embed = getattr(field_info, "embed", None)
-        if len(required_params) == 1 and not embed:
+        field_alias_omitted = len(required_params) == 1 and not embed
+        if field_alias_omitted:
             received_body = {field.alias: received_body}
+
         for field in required_params:
+            loc: Tuple[str, ...]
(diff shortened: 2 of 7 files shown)


tiangolo:
Thank you! Excellent job! :rocket:
Impeccable implementation! :exploding_head: 

Thanks for taking care of updating the docs too. :ok_hand: 

Thanks for your contribution! :cake: :coffee: :tada:

PR #1554, opened 2020-06-11
Mention in docs that subapps don't fire events

### LINKED ISSUE #811: [BUG][DOCUMENTATION] Sub application startup event never triggered
### Describe the bug

Startup event handlers for sub-applications never trigger.

### To Reproduce

```python
from fastapi import FastAPI


sub_app = FastAPI(openapi_prefix="/subapp")


@sub_app.on_event("startup")
async def sub_app_startup():
    print("++++ Sub-app startup event")  # Never fires


app = FastAPI()


@app.on_event("startup")
async def app_startup():
    print("++++ App startup event")


app.mount("/subapp", sub_app)


if __name__ == "__main__":
    import uvicorn

    uvicorn.run(app, port=8000, host="0.0.0.0")  # nosec
```

Running the above shows:

```bash
$ python subapp_events.py
INFO:     Started server process [38462]
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
INFO:     Waiting for application startup.
++++ App startup event
INFO:     Application startup complete.
```

**Notice that `sub_app_startup()` is never run.**

### Expected behavior

I would expect the sub-application startup event handler to also run (although I am not sure what ordering I would expect).

### Environment

- OS: macOS
- FastAPI Version: 0.45.0
- Python version: 3.7.5

### Additional context

N/A

andrew222651:
This should fix/related to #811

### DIFF
--- a/docs/en/docs/advanced/events.md
+++ b/docs/en/docs/advanced/events.md
@@ -4,6 +4,10 @@ You can define event handlers (functions) that need to be executed before the ap
 
 These functions can be declared with `async def` or normal `def`.
 
+!!! warning
+    If `app` is an instance of `FastAPI` then
+    sub-applications registered with `app.mount` do not fire these events.
+
 ## `startup` event
 
 To add a function that should be run before the application starts, declare it with the event `"startup"`:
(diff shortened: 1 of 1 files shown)


tiangolo:
Nice! Thank you! :cake: :rocket:

PR #1555, opened 2020-06-11
Relax hard-pinning of starlette

dhirschfeld:
[`starlette`](https://github.com/encode/starlette/releases) is up to `0.13.4` now.

Unless a specific patch release is actually a hard requirement it's more user-friendly to pin to a [compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release) so as to *automatically* benefit from bug/security fixes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 requires = [
-    "starlette ==0.13.2",
+    "starlette ~=0.13.2",
     "pydantic >=0.32.2,<2.0.0"
 ]
 description-file = "README.md"


tiangolo:
Thanks! Yep, the problem is that as FastAPI is very tightly integrated with Starlette, there could be (and have been) breaking changes very easily, even for patch releases :pensive: 

For Pydantic it's possible to have more relaxed pinning mainly because Pydantic runs all of FastAPI's tests as part of its own CI tests.

On the other hand, I'm thinking of migrating from Flit to Poetry, that way I could set up Dependabot to make it send PRs for each new version.

But for the time being, I have to keep the strict pinning...

Anyway, thanks for your contribution! And for all the management in Conda Forge! :bow: :rocket:

Issue #1528, opened 2020-06-07
[BUG] Override successful response


Kludex:
### Describe the bug

On the `responses` field in the route decorator we can't override the description of the successful response. Even on the tutorial we can see it: https://fastapi.tiangolo.com/advanced/additional-responses/

### To Reproduce

![image](https://user-images.githubusercontent.com/7353520/83968295-81a19f80-a8c8-11ea-8b2e-cce027582d17.png)
![image](https://user-images.githubusercontent.com/7353520/83968299-88301700-a8c8-11ea-8d42-d0d9f2a1ec3c.png)

### Expected behavior

We should have "Item requested by ID" in the description field for the code 200.

Kludex:
We can use `response_description`, but we can't override with `responses` so the BUG still remains.

fabrizzioalco:
Hey i would like to contribute and perhaps some mentorship. This would be my first contribution to open source

tiangolo:
@fabrizzioalco that's great!

You should probably:

* Make sure to read and follow the tutorial and get familiar with using the framework and how it works
* You could help others here and in the Gitter chat, many questions just need a bit of guidance and help. And by helping others you would get a lot more expertise with FastAPI.
* If you speak another language apart from English, you could help with the translations to your language, that's something that only a few can do (and I can't, as I don't speak all the languages :sweat_smile: ): https://fastapi.tiangolo.com/contributing/#translations
    * And as by doing that you would be reading the documentation, having to understand it, and re-writing it in your language, you would probably learn a lot of the concepts and how it all works.
* If you want to contribute code to FastAPI itself, check the contributing guide: https://fastapi.tiangolo.com/contributing/

rkbeatss:
@tiangolo if there has not been any PRs yet, i have been thinking of looking into contributing to this issue. what do you think?

Kludex:
@rkbeatss I can say that there are no PRs solving it for now, I think you can safely work on it.

fabrizzioalco:
Please, be my guest. I had some troubles so i can't work on this @rkbeatss

fabrizzioalco:
Later i would contribute on something else

tiangolo:
Thanks for the report and for the discussion here everyone! :coffee: 

And thanks @rkbeatss for the initial implementation! :cake: 

This should be fixed in https://github.com/tiangolo/fastapi/pull/1577

Available in FastAPI version `0.58.0` :tada:

kevlarr:
[...]

PR #1563, opened 2020-06-12
Fix: Override successful response

rkbeatss:
first PR to fix bug #1528. shows overriden success description if specified else falls back to default response


tiangolo:
Thanks for all your work on this @rkbeatss ! :rocket: :cake: 

I was doing a small update to this PR and I realized that the problem was a lot deeper than I initially thought.

So I ended up creating https://github.com/tiangolo/fastapi/pull/1577 based on your PR, adding all the extra refactor required, and rebasing from `master`.

I just merged that one, and it includes your commits :sparkles: , so we can close this one now.

Thanks for your effort! :coffee:

PR #1583, opened 2020-06-17
Fix doc URL in root_path comment

uranusjr:
The previous URL is now returning 404.


tiangolo:
Thanks! This was fixed in https://github.com/tiangolo/fastapi/pull/1611, I'm not sure why I saw that one first :thinking: 

I'll now close this one.

But thanks for your contribution! :cake: :coffee:

PR #1585, opened 2020-06-17
Translated `index.md` and `first-steps.md` from Tutorial - User Guide to Italian

csr:
I've translated `index.md` and `first-steps.md` from Tutorial - User Guide to Italian (the first file was so short that I included both in the PR, separate PRs will be created this way forward). `docs/it/mkdocs.yml` was updated accordingly.


csr:
I'm not actively participating in the translation efforts any longer but I would encourage you to create a separate pull request with a new translation.

PR #1590, opened 2020-06-18
Add vim temporary files to gitignore

asheux:


.gitignore
```diff
@@ -16,3 +16,8 @@ Pipfile.lock
 env3.*
 env
 docs_build
+venv
+
+# vim stuff
```
Kludex, line 21:
```suggestion
# vim temporary files
```

Kludex (approved):

.gitignore
asheux: All done

Kludex (approved):

tiangolo:
Thanks! I don't normally use Vim, but I imagine this would help others using it :shrug: :heavy_check_mark: 

Thanks for the review @Kludex !

And thanks for the contribution @asheux ! :cake: :rocket:

Issue #1318, opened 2020-04-24
Tutorial [EN] speak about ordering in Union variants


chenl:
### Describe the bug

The example in `/fastapi/blob/master/docs_src/extra_models/tutorial003.py` describe how to use a `Union` in a `response_model` like so:

```python
@app.get("/items/{item_id}", response_model=Union[PlaneItem, CarItem])
async def read_item(item_id: str):
    return items[item_id]
```

However this will not work as expected if we will to flip the order of the items inside the `Union[...]`.

```python
@app.get("/items/{item_id}", response_model=Union[CarItem, PlaneItem])
async def read_item(item_id: str):
    return items[item_id]
```

In that case the `"size": 5` will not be returned for `"item2"`.

And there is no mention of the issue anywhere in the text at:

https://fastapi.tiangolo.com/tutorial/extra-models/#union-or-anyof

### To Reproduce

In the mindset of a new user, type the example flipping `ItemCar` and `ItemPlane` and try to figure out why you don't get the expected result for `item2`.

### Expected behavior

At the very least there should be a note explaining that the order of Union variants are important in this context, and say how and why.

kbanc:
I replicated the bug. It seems to only occur when the first class passed into the Union doesn't define any further class variables than what was defined in the base class. In the example below, the Union works with `CarItem` and `PlaneItem` in either order. 

```
from typing import Union

from fastapi import FastAPI
from pydantic import BaseModel

app = FastAPI()


class BaseItem(BaseModel):
    description: str
    type: str

class CarItem(BaseItem):
    type = "car"
    colour: str

class PlaneItem(BaseItem):
    type = "plane"
    size: int
    model: str

items = {
    "item1": {"description": "All my friends drive a low rider",
              "type": "car",
              "colour": "blue"
              },
    "item2": {
        "description": "Music is my aeroplane, it's my aeroplane",
        "type": "plane",
        "size": 5,
        "model": "big",
    }
}

@app.get("/items/{item_id}", response_model=Union[CarItem, PlaneItem])
async def read_item(item_id: str):
    return items[item_id]
```
It seems like this is a bug with pydantic. I'll dig into it further and file an issue.

tiangolo:
Awesome @kbanc ! :rocket: Thanks a lot for checking it and replicating the problem here, it helps a lot! :tada: :bow: 

So, yes, that's part of how Pydantic works, it's documented here: https://pydantic-docs.helpmanual.io/usage/types/#unions

[...]

PR #1591, opened 2020-06-18
Add note in docs on order in Pydantic Unions

kbanc:
Added a small note to the docs to address the confusion around the order of types in a Union. This should address #1318.

### DIFF
--- a/docs/en/docs/tutorial/extra-models.md
+++ b/docs/en/docs/tutorial/extra-models.md
@@ -162,6 +162,9 @@ It will be defined in OpenAPI with `anyOf`.
 
 To do that, use the standard Python type hint <a href="https://docs.python.org/3/library/typing.html#typing.Union" class="external-link" target="_blank">`typing.Union`</a>:
 
+!!! note
+    When defining a <a href="https://pydantic-docs.helpmanual.io/usage/types/#unions" class="external-link" target="_blank">`Union`</a>, include the most specific type first, followed by the less specific type. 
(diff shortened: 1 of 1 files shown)


docs/en/docs/tutorial/extra-models.md
```diff
@@ -162,6 +162,9 @@ It will be defined in OpenAPI with `anyOf`.
 
 To do that, use the standard Python type hint <a href="https://docs.python.org/3/library/typing.html#typing.Union" class="external-link" target="_blank">`typing.Union`</a>:
 
+!!! note
+    When defining a <a href="https://pydantic-docs.helpmanual.io/usage/types/#unions" class="external-link" target="_blank">`Union`</a>, include the most specific type first, followed by the less specific type.
```
Kludex, line 166: Good catch. I don't know if it's 100% clear without an example, do you think the same example can/need to be added here, or the note itself is already clear enough?

kbanc: I added clarification on how this behaviour can be seen in the example already within the Union section.

Kludex (approved):

docs/en/docs/tutorial/extra-models.md
Kludex: @kbanc Thank you! :)

tiangolo:
Awesome, thank you @kbanc ! :cake: :coffee: :tada:

PR #1592, opened 2020-06-18
OpenAPI docs for scopes defined on a single route

andrewsnowden:
Scopes that are defined on an individual route as part of a `Security` Dependency were not being returned on the OpenAPI JSON that is generated.

This seems to be because when the dependencies are flattened the `security_scopes` are not being flattened (unlike `security_requirements` and others which are). This sets the default scopes to an empty list and flattens the scopes of sub dependencies. I've added a test which confirms this in the OAuth2 test.

I'm brand new to the FastAPI code base so I am unsure if there are any unintended consequences with this approach - but all the tests seem to pass and the new behaviour makes sense in my understanding of what the code should do.

### DIFF
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -50,7 +50,7 @@ class Dependant:
         self.websocket_param_name = websocket_param_name
         self.response_param_name = response_param_name
         self.background_tasks_param_name = background_tasks_param_name
-        self.security_scopes = security_scopes
+        self.security_scopes = security_scopes or []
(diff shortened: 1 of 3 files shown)


tiangolo:
Hey @andrewsnowden , thanks for your contribution! Do you have an example code that shows the current problem?

I think there are things to update in the way scopes are currently handled, but I ran your test and it seems to pass with the current version, so I'm not sure this PR would fix it.

Could you write a simple self-contained example that shows the problem?

Issue #1294, opened 2020-04-20
Unable to correctly reverse proxy FastAPI


ybizeul:
### Describe the bug

When using FastAPI behind a proxy that adds path elements like `/api/1.0/`, there is no way to correctly display the swagger

### To Reproduce

1. Configure a reverse proxy to send `/api/1.0/` to your FastAPI server
2. access swagger for the API
3. Observe that all your objects are prefixed with `/api/1.0`

See screenshot

### Expected behavior

`/api/1.0` should not be part of API documentation.

OpenAPI documents this [here](https://swagger.io/docs/specification/api-host-and-base-path/)

There should be a way to configure **Server URL** to `/api/1.0` like indicated in the examples.

### Screenshots

![image](https://user-images.githubusercontent.com/182035/79806282-538fec80-8368-11ea-8cd8-f766288baf47.png)

### Environment

- OS: alpine docker image
- FastAPI Version 0.54.1
- Python version 3.7.3

### Additional context

Using Traefik as reverse proxy. Traefik like other proxies, has the ability to either :
Pass the request as-is with `/api/1.0/`, or strip the part from the URL and put it in the `X-Forwarded-Prefix` header. Maybe that could be used as the server URL for Swagger.

esomir:
You can use `openapi_prefix` (https://fastapi.tiangolo.com/advanced/sub-applications-proxy/) but that's not really viable when the application doesn't know under which prefix it's running from.

Ideally, it should be resolved dynamically based on the `X-Forwarded-Prefix`. We are using Traefik and having the same problem. We are having a PathPrefix on our route, and using the StripPrefix middleware, which will strip the prefix, and pass it to the `X-Forwarded-Prefix` header.

ybizeul:
That's exactly my situation. Traefik passes all the information but they are not leveraged by fastapi. I switched to Connexion since then

xmrdsc:
Similar issue [here](https://github.com/tiangolo/fastapi/issues/829)

I fixed it using my own crappy middleware:

```python3
async def fix_root(app: FastAPI, path: str):
    """Fixes incoming path when application is deployed under a path"""
    path_len = len(path)

    @app.middleware("http")
    async def application_root_fix_middleware(request: Request, call_next):
        if request.scope['path'].startswith(path):
            request.scope['path'] = \
                request.scope['path'][path_len:]
            request.scope['raw_path'] = \
                request.scope['raw_path'][path_len:]
        return await call_next(request)

[...]

PR #1596, opened 2020-06-19
Correctly display Swagger docs when using reverse proxy

rkbeatss:
addressing #1294: root_path is now included in the servers object instead of in the path prefix. should more tests be added?


tiangolo:
Awesome, thanks @rkbeatss ! :rocket: :cake: 

I updated it a bit and included an optional parameter `root_path_in_servers` in case someone doesn't want the auto-generated servers. And I added some docs for that.

Given this solves the generation of OpenAPI schemas and the utils no longer need the `openapi_prefix`, I removed that parameter that was recently added, so, back to its original form :man_facepalming: 

I also checked/confirmed that this doesn't break the functionality with OAuth2 tokens by the spec, and to make sure they all stay compatible, I made those `tokenUrl`s relative URLs, that way those docs will keep working as-is even when behind a reverse proxy.

Thanks for your contribution! :coffee: :rocket: :tada:

dontsovcmc:
I solved this issue using [subapi](https://fastapi.tiangolo.com/advanced/sub-applications/):
`root_path` doesn't work for me. `openapi_prefix` is deprecated.

```
from fastapi import FastAPI

app = FastAPI()

@app.get("/app")
def read_main():
    return {"message": "Hello World from main app"}

subapi = FastAPI()

@subapi.get("/sub")
def read_sub():
    return {"message": "Hello World from sub API"}

app.mount("/subapi", subapi)
```

PR #1601, opened 2020-06-20
Fix overlapping menu item in the docs

asheux:
- Before the fix
before
- After the fix
after

### DIFF
--- a/docs/en/docs/css/custom.css
+++ b/docs/en/docs/css/custom.css
@@ -16,3 +16,10 @@ a.internal-link::after {
 .md-footer-meta {
     padding-bottom: 2em;
 }
+
+/* Increase the width for menu nav-item that displays github repo contents 
+ * (number of forks and stars) 
+ * */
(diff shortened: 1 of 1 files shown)


tiangolo:
Hmm, this is strange, I can't reproduce the issue, I tried on different sizes but it seems to work correctly.

What browser are you using? Have you tried on an incognito window?

Selection_067

asheux:
I'm using Firefox and the issues is still there in incognito. See attached.
incognito

asheux:
I just tested on Chrome as well, seems ok. 
Tested my changes on both browsers. Perhaps, with the introduction of these changes, it should be fine on both browsers.

tiangolo:
Thanks @asheux ! I wasn't able to reproduce the problem in an incognito window in Firefox, so I think it might be related to something in your environment.

So I'll close this now, but thanks for your effort! :coffee: :cake:

PR #1603, opened 2020-06-20
Fix: bad link in docs

molto0504:
Fix a bad link, which return 404 previously.


Kludex (approved):

tiangolo:
Good catch, thanks! :rocket:

Issue #1598, opened 2020-06-19
Update JWT docs with python-jose


tiangolo:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

N/A, docs.

### Description

The docs should suggest and use https://github.com/mpdavis/python-jose instead of PyJWT as `python-jose` supports the same features plus some more that are required for some OAuth2 use cases.

### The solution you would like

I would like all the docs that deal with security with JWT to suggest and https://github.com/mpdavis/python-jose instead of PyJWT.

### Describe alternatives you've considered

To leave it as is, but I'm already using `python-jose` everywhere and is what I would recommend.

### Environment

* OS: [e.g. Linux / Windows / macOS]: All
* FastAPI Version [e.g. 0.3.0]: All

To know the FastAPI version use:

```bash
python -c "import fastapi; print(fastapi.__version__)"
```

* Python version: All

To know the Python version use:

```bash
python --version
```

asheux:
@tiangolo, I will be working on this issue.

tiangolo:
Thanks @asheux for your work on this! :rocket:

dineshbvadhia:
Has it been noticed python-jose has not been updated since June 2021 ?

birdhackor:
> Has it been noticed python-jose has not been updated since June 2021 ?

@tiangolo , as @dineshbvadhia mentioned, python-jose has not been updated since June 2021, maybe rollback to PyJWT is a better choice now?

swssl:
[...]

PR #1610, opened 2020-06-22
Update JWT docs with python-jose

asheux:
addressing #1598: suggesting support for python-jose as opposed to the initial PyJWT


Kludex (approved):

tiangolo:
Thanks @asheux ! :rocket: 

I just updated the format and used Python-jose for the other example, and added the dependency for the tests.

PR #1614, opened 2020-06-23
Fix typo

marciomazza:


tiangolo:
Thanks for your contribution! :coffee: 

That section actually refers to "string validations" in the same style as the previous chapter.

So I'll pass on this one, but thanks for your effort!

Issue #1273, opened 2020-04-16
[FEATURE] Use TestClient in an async fashion


bharel:
### Is your feature request related to a problem

Unfortunately, test client runs it's own event loop. This causes an issue when attempting to do other async operations in the background.
Running it inside an executor will cause it to run in a different threat, which in turn causes issues with other async libraries a they're calling coroutines from a different thread.

### The solution you would like

Make a TestClient that doesn't create it's own loop. Just have it async. 

### Describe alternatives you've considered

Other than executor or partching fastapi, there are no alternatives.

nsidnev:
There are plans (https://github.com/encode/starlette/issues/440 and https://github.com/encode/starlette/issues/652) to move `starlette.testclient.TestClient` from `requests` to `httpx` to fix this behaviour. Before this happens you can try using `httpx` [directly](https://www.python-httpx.org/async/#calling-into-python-web-apps) or try the [`async-asgi-client`](https://pypi.org/project/async-asgi-testclient/) package. But there may be some problems with the templates, if you are using them.

bharel:
As long as fastapi doesn't support it, may I suggest adding a note to the docs?

nsidnev:
Yes, good idea. I think adding an [`info`](https://squidfunk.github.io/mkdocs-material/extensions/admonition/#info) block after the introduction on [this](https://fastapi.tiangolo.com/tutorial/testing/) page will be good place, right? 

In addition, it is possible to create a new page that will show how to use `httpx` (or `async-asgi-client`) for tests. Probably in the advanced section. 

Would you submit a PR?

tiangolo:
Yep, as @nsidnev says. :point_up: 

And yeah, that could probably have a small section in the advanced tutorial.

empicano:
I've done some async testing using `httpx` in the last few days, as I wanted to verify the state of the database in my tests as well, and that works really well! I don't have experience with the `async-asgi-client` though.

Would be up to write a section in the docs on it!

aoncreed:
As @empicano mentioned using httpx directly works great. Here's a little example "workaround" by using the `httpx.AsyncClient` and it's working quite well

in your conftest.py

```python

from typing import Generator
import pytest
from httpx import AsyncClient
from asyncio import get_event_loop

from app.db.session import SessionLocal
 

@pytest.fixture(scope="session")
def db() -> Generator:

    yield SessionLocal()


@pytest.fixture(scope="module")
async def async_client() -> Generator:

[...]

PR #1619, opened 2020-06-24
Add documentation about asynchronous testing (pytest-asyncio and httpx)

empicano:
Addressing #1273 and encode/starlette#652, I added a section about asynchronous testing via pytest-asyncio and httpx to the documentation. I also made a small note (+ link) in the beginner tutorial about it.


docs_src/testing_asynchronously/test_main.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+from httpx import AsyncClient
+
+import main
+
+
+@pytest.mark.asyncio
+async def test_root():
+    async with AsyncClient(app=main.app, base_url='http://test') as ac:
```
marco-neumann-jdas, line 10: You probably need to add `trust_env=False` here as well, otherwise the async client might go through a proxy. Not sure if this is an upstream bug though.

empicano: My initial thought was to simply follow the [HTTPX example](https://www.python-httpx.org/async/#calling-into-python-web-apps) for calling into python web apps. Also, HTTPX sets [some constraints on the proxies it accepts](https://github.com/encode/httpx/blob/cfd07cc290cc610fceb8d9f025c6bebc1d4b3275/httpx/_utils.py#L287) from the environment, I don't know if that changes anything.

Otherwise:
```suggestion
    async with AsyncClient(app=main.app, base_url='http://test', trust_env=False) as ac:
```

marco-neumann-jdas: https://github.com/encode/httpx/issues/1039 I think it might be a bug.

empicano: Hm, interesting. What do you propose, we add the `trust_env=False` for now?

marco-neumann-jdas: Yes, I think that's reasonable. Users who have some default network proxy active (e.g. for privacy reasons or in company network situations) will get some bizarre test failures.

florimondmanca: Hi there, chiming in to confirm that this odd behavior is indeed due to a bug in HTTPX.

So it might be okay to go with this as-is (i.e. without the `trust_env=False` hack), keeping in mind that https://github.com/encode/httpx/issues/1039 looks to be fairly straight-forward to fix (in case anyone wants to give it a go - happy to review a PR, but no expectations of course), so hopefully a fix will be released soon and users equipped with proxies won't encounter this.

empicano:
@marco-neumann-by just to point out that the issue with httpx that we hung up on is resolved now

tiangolo:
Awesome! Great job @empicano ! Thanks for your contribution! :clap: :memo: :nerd_face: :cake: 

Thanks everyone for the discussion (and @florimondmanca for HTTPX :rocket: :blue_heart: ).

PR #1621, opened 2020-06-24
fixing docstring for oauth2.py

tomarv2:
While implementing oauth2 found issue with the doc.


tiangolo:
Good catch! :mag: :detective: Thank you! :coffee: :cake:

Issue #1599, opened 2020-06-19
Check python-multipart when using Form and raise early


tiangolo:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

Here's a self-contained [minimal, reproducible, example](https://stackoverflow.com/help/minimal-reproducible-example) with my use case:

```Python
from fastapi import FastAPI, Form

app = FastAPI()


@app.post("/login/")
async def login(username: str = Form(...), password: str = Form(...)):
    return {"username": username}
```

### Description

* Create a new venv
* Install only FastAPI and Uvicorn
* Run that example
* Try it in the docs
* It returns a 400 response with:

```JSON
{
  "detail": "There was an error parsing the body"
}
```

The problem is that it's trying to read form data (multipart), and for that it requires the package `python-multipart` installed.

It is documented in the docs in FastAPI and Starlette, but forgetting to install it and running the app ends up in that behavior that is not explicit nor obvious. Everything works normally, except that specific *path operation*, that only breaks after a receiving a request (which is worse).

To make things more complex, the package required is `python-multipart`. But there's another Python package called just `multipart`.

And both packages are imported as:

```Python
import multipart
```

This can lead to a problem that is really difficult to debug: an environment with both packages installed, in some runs could import the correct package and everything would work, but in other runs, it could import the other package, which would cause problems. That would make it one of those scary ghost bugs that disappear and appear. :bug: :ghost: 

 And even worse, the problems will only show up *after* receiving a request in a path operation that requires reading form data, not at start up.

### The solution you would like

[...]

PR #1627, opened 2020-06-25
Raise form early

chrisnguyn:
Addresses #1599.

We'll be leveraging the `[import].__package__` field as a way to determine if the CORRECT multipart import was used or not.

Oddly enough, the incorrect `multipart` import will have a `.__package__` field value of "" (empty string). The correct import, `python-multipart`, will have a `.__package__` field value of "multipart".

We'll check to see if the user will be using forms or not with `is_form_data(BodyFieldInfo)`. If this evaluates to true, then we can check to see there is a `multipart` import at all by using `importlib.util.find_spec("multipart")`. If that again evaluates to true, we can check the `.__package__` value to see if it is an empty string or not.

Alternative solution:

Instead of testing the imports' `.__package__` field, we could try to wrap a try / catch with a method exclusive to `python-multipart` that is not in `multipart` instead. If an exception was caught we know it's also the wrong import.

`parser = multipart.QuerystringParser()` might be a good method to try out.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -1,5 +1,8 @@
 import asyncio
+import importlib
 import inspect
+import logging
+
 from contextlib import contextmanager
 from copy import deepcopy
 from typing import (
@@ -776,6 +779,10 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
         ]
         if len(set(body_param_media_types)) == 1:
             BodyFieldInfo_kwargs["media_type"] = body_param_media_types[0]
+
+    if BodyFieldInfo == params.Form and importlib.util.find_spec("multipart") is None:
+        logging.error("Must import python-multipart.")
+
     return create_response_field(
         name="body",
         type_=BodyModel,
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -2,7 +2,6 @@ import asyncio
 import importlib
 import inspect
 import logging
-import pkgutil
 
 from contextlib import contextmanager
 from copy import deepcopy
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -798,8 +798,6 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
                 "[multipart] is installed and not compatible with [python-multipart]. "
                 "Uninstall [multipart] and then install [python-multipart]."
             )
-            [multipart] is installed and not compatible with [python-multipart]. 
-            Uninstall [multipart] and then install [python-multipart]."""
             logger.error(error)
             raise RuntimeError(error)
     return create_response_field(
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -740,7 +740,7 @@ def get_schema_compatible_field(*, field: ModelField) -> ModelField:
 
 
 def is_form_data(BodyFieldInfo):
-    return BodyFieldInfo == params.Form or BodyFieldInfo == params.UploadFile or BodyFieldInfo == bytes
+    return BodyFieldInfo == params.Form or BodyFieldInfo == params.File or BodyFieldInfo == bytes
 
 
 def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
@@ -783,9 +783,18 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
         ]
         if len(set(body_param_media_types)) == 1:
             BodyFieldInfo_kwargs["media_type"] = body_param_media_types[0]
-
-    if is_form_data(BodyFieldInfo) and importlib.util.find_spec("multipart") is None:
-        logging.error("Must import python-multipart.")
+    
+    if is_form_data(BodyFieldInfo):
+            if importlib.util.find_spec("multipart") is None:
+                logging.error("Import python-multipart.")
+                return
+            else:
+                import multipart as mp
+                if len(mp.__package__) == 0:
+                    logging.error("Wrong multipart import. pip3 uninstall multipart --> pip3 install python-multipart.")
+                    return
+                else:
+                    print('You have the right multipart!')
 
     return create_response_field(
         name="body",
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -785,16 +785,18 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
     
     if is_form_data(BodyFieldInfo):
         try:
-            import multipart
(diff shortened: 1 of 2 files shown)


chrisngyn:
Instead of testing the imports' `.__package__` field, we could try to wrap a try / catch with a method exclusive to `python-multipart` that is not in `multipart` instead. If an exception was caught we know it's also the wrong import.

fastapi/dependencies/utils.py
```diff
@@ -776,6 +783,19 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
         ]
         if len(set(body_param_media_types)) == 1:
             BodyFieldInfo_kwargs["media_type"] = body_param_media_types[0]
+    
+    if is_form_data(BodyFieldInfo):
+            if importlib.util.find_spec("multipart") is None:
+                logging.error("Import python-multipart.")
+                return
+            else:
```
chrisngyn, line 791: Extra tabbed by mistake!

fastapi/dependencies/utils.py
```diff
@@ -776,6 +783,19 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
...
+                import multipart as mp
+                if len(mp.__package__) == 0:
+                    logging.error("Wrong multipart import. pip3 uninstall multipart --> pip3 install python-multipart.")
+                    return
+                else:
+                    print('You have the right multipart!')
```
chrisngyn, line 797: If we're okay with this solution, we can clean it up a little bit + throw those Exceptions.

chrisnguyn (commented):
NTS

fastapi/dependencies/utils.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 import inspect
+import logging
```
kx-chen, line 3: I saw there's a logging module already in fastapi, would that be the correct one to use?

tests/test_multipart_installation.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+from fastapi import FastAPI, Form
+
+
+def test_incorrect_multipart_installed(monkeypatch):
```
kx-chen, line 5: Not too sure if this is the best way to detect it. This test will fail if the user doesn't have any kind of `multipart` installed. Maybe we skip it if we can't import `multipart`? thoughts? 

Alternatively, we add our own mock `multipart` package and change the sys path so that it gets imported when the code is run?

kx-chen (commented):
Just left a few concerns of mine. Any ideas would be greatly appreciated.

tests/test_multipart_installation.py
kx-chen: Also, I'd like to maybe add more specific assertions for the error message possibly as they both raise `RuntimeError`

fastapi/dependencies/utils.py
```diff
@@ -776,6 +781,23 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
...
+
+            QuerystringParser(  # check if correct import using python-multipart function
+                {}
+            )
+        except ModuleNotFoundError:
+            error = "Form data requires [python-multipart] to be installed."
+            logger.error(error)
+            raise RuntimeError(error)
+        except ImportError:
+            error = """Form data requires [python-multipart] to be installed. Currently
```
Kludex, line 796: If you print `error` you'll see that there are a lot of spaces on it.
```suggestion:-0+2
            error = (
                "Form data requires [python-multipart] to be installed. Currently "
                "[multipart] is installed and not compatible with [python-multipart]. "
                "Uninstall [multipart] and then install [python-multipart]."
            )
```

fastapi/dependencies/utils.py
```diff
@@ -736,6 +737,10 @@ def get_schema_compatible_field(*, field: ModelField) -> ModelField:
     return out_field
 
 
+def is_form_data(field: Type[params.Body]) -> bool:
+    return field == params.Form or field == params.File or field == bytes
```
Kludex, line 741: What do you think about this?
```suggestion
    return field in (params.Form, params.File, bytes)
```

chrisnguyn (commented):
Looks good to me; might want to ask Sebastian more about if he's satisfied with the tests / wants us to add more.

fastapi/dependencies/utils.py
chrisngyn: I didn't know you could do this to get rid of the spaces; thanks for this suggestion 😄

Kludex: I don't know how to create a multi-line suggestion, does anyone know?

fastapi/dependencies/utils.py
chrisngyn: Looks good to me, will commit this suggestion. 👍

fastapi/dependencies/utils.py
```diff
@@ -776,6 +781,27 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
...
+            error = "Form data requires [python-multipart] to be installed."
+            logger.error(error)
+            raise RuntimeError(error)
+        except ImportError:
+            error = (
+                "Form data requires [python-multipart] to be installed. Currently "
+                "[multipart] is installed and not compatible with [python-multipart]. "
+                "Uninstall [multipart] and then install [python-multipart]."
+            )
+            [multipart] is installed and not compatible with [python-multipart].
```
Kludex, line 801:
```suggestion
```

fastapi/dependencies/utils.py
chrisngyn: Oh I didn't realize the rest of the lines were there too; I'll fix it and make another push, sorry about that

fastapi/dependencies/utils.py
```diff
@@ -776,6 +781,27 @@ def get_body_field(*, dependant: Dependant, name: str) -> Optional[ModelField]:
...
+            logger.error(error)
+            raise RuntimeError(error)
+        except ImportError:
+            error = (
+                "Form data requires [python-multipart] to be installed. Currently "
+                "[multipart] is installed and not compatible with [python-multipart]. "
+                "Uninstall [multipart] and then install [python-multipart]."
+            )
+            [multipart] is installed and not compatible with [python-multipart]. 
+            Uninstall [multipart] and then install [python-multipart]."""
```
Kludex, line 802:
```suggestion
```

Kludex (commented):
Sorry for the last suggestion on it.

Kludex (approved):

tiangolo:
Thanks for all the work on this! @chrisngyn, @YKo20010, @kx-chen! :nerd_face: :cake: :coffee: 

And thanks @Kludex for the review! :coffee: 

I updated a couple of things, and as I can't push to your branch, I created a new PR including all your commits plus the new changes here: https://github.com/tiangolo/fastapi/pull/1851 , and I just merged it :tada: :rocket:

PR #1630, opened 2020-06-26
Fixed issues with docs , particularly extending OpenAPI schema

daggy1234:
The example in the docs do not work because the defualt prefix is not specified. It may confuse a lot of begginers and thats why I modified both the en docs as well as the python code for the same.

### DIFF
--- a/docs/en/docs/advanced/extending-openapi.md
+++ b/docs/en/docs/advanced/extending-openapi.md
@@ -79,6 +79,8 @@ That way, your application won't have to generate the schema every time a user o
 
 It will be generated only once, and then the same cached schema will be used for the next requests.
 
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks for your contribution! This was updated/improved in https://github.com/tiangolo/fastapi/pull/1596 , and the `openapi_prefix` parameter is no longer needed :tada: 

The new docs are here: https://fastapi.tiangolo.com/advanced/extending-openapi/

And the updated docs about `root_path`, that previously required the `openapi_prefix` parameter, are here: https://fastapi.tiangolo.com/advanced/behind-a-proxy/

This is available in FastAPI version `0.59.0` :tada: 

So I'll close this now, but thanks for your effort! :coffee:

Issue #1628, opened 2020-06-26
pass method return value as positional parameters to Response


dmig-alarstudios:
Recently I've noticed that `status_code` parameter of the next statement is useless:
```python3
@router.get('/callback/{provider}', status_code=303,
            response_description='Redirect to the application login',
            response_class=RedirectResponse)
```
Because I always have to create `RedirectResponse` objects manually:
```python3
        return RedirectResponse(target.include_query_params(error=f'{provider}_{error_reason}'),
                                status_code=303)
```

I've tried to play with return values and noticed that return value is always passed to `response_class` instance as a `content` parameter:
```python3
@app.get('/test', status_code=303, response_class=RedirectResponse)
async def test():
    return '/'
```
this snippet produces the next exception:
```
INFO:     127.0.0.1:43878 - "GET /test HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/dmig/.pyenv/versions/fastapi-sandbox/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 385, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/home/dmig/.pyenv/versions/fastapi-sandbox/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/home/dmig/.pyenv/versions/fastapi-sandbox/lib/python3.8/site-packages/fastapi/applications.py", line 171, in __call__
    await super().__call__(scope, receive, send)
  File "/home/dmig/.pyenv/versions/fastapi-sandbox/lib/python3.8/site-packages/starlette/applications.py", line 102, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/dmig/.pyenv/versions/fastapi-sandbox/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/home/dmig/.pyenv/versions/fastapi-sandbox/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/home/dmig/.pyenv/versions/fastapi-sandbox/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
[...]

PR #1643, opened 2020-06-28
Simplify usage of RedirectResponse/FileResponse

dmig:
A simple solution to #1628:

```python3
from fastapi import FastAPI
from fastapi.responses import RedirectResponse, PlainTextResponse

app = FastAPI()

@app.get('/', response_class=PlainTextResponse)
async def hello():
    return 'Hello world!'

@app.get('/redir', response_class=RedirectResponse, status_code=303)
async def redir():
    return '/'
```

```
$ curl -iL "http://127.0.0.1:8000/redir" -H "accept: */*"
HTTP/1.1 303 See Other
date: Sun, 28 Jun 2020 12:44:30 GMT
server: uvicorn
location: /
transfer-encoding: chunked

HTTP/1.1 200 OK
date: Sun, 28 Jun 2020 12:44:30 GMT
server: uvicorn
content-length: 12
content-type: text/plain; charset=utf-8

Hello world!
```

### DIFF
--- a/fastapi/responses.py
+++ b/fastapi/responses.py
@@ -1,10 +1,13 @@
-from typing import Any
+from os import stat_result
+from typing import Any, Union
 
-from starlette.responses import FileResponse  # noqa
+from starlette.background import BackgroundTask
+from starlette.datastructures import URL
+from starlette.responses import FileResponse as StarletteFileResponse  # noqa
 from starlette.responses import HTMLResponse  # noqa
 from starlette.responses import JSONResponse  # noqa
 from starlette.responses import PlainTextResponse  # noqa
-from starlette.responses import RedirectResponse  # noqa
+from starlette.responses import RedirectResponse as StarletteRedirectResponse  # noqa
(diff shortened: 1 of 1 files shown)


dmig:
This PR also adds a wrapper to `FileResponse`, so with defined `response_class=FileResponse` one can return a path string from endpoint function.

dmig:
The next step would be to change `status_code: int = 200` in `applications.py` to `status_code: int = None` and allow response classes to set their default codes if other was not specified.

tiangolo:
Thanks for the effort @dmig! :coffee: :rocket: 

I implemented this in a different way here: https://github.com/tiangolo/fastapi/pull/3457, without modifying the underlying response classes. And I also handled the default `status_code`, extracting it from the response classes (that part was tricky :sweat_smile: ).

It will be available tomorrow with FastAPI version `0.66.0` :tada: :rocket: 

Given that, I'll close this PR now. But thanks for your effort!

PR #1644, opened 2020-06-28
📝 Use Optional in docs

tiangolo:
📝 Use `Optional` in docs.

This includes all the commits of https://github.com/tiangolo/fastapi/pull/1574 , plus some small extra fixes and updates.

Sadly it seems I can't push these updates to that branch, so I have to create this new one. The good thing is that the commits and authors are preserved :rocket:


Kludex:
closes https://github.com/tiangolo/fastapi/issues/1561

Issue #1650, opened 2020-06-30
Exclude all routes from schema under an APIRouter


abdusco:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [ ] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [ ] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

I want to be able to exclude all routes of an APIRouter from documentation.

[There's a way][1] to do this for individual routes, but having a way to apply this option to all subroutes is more maintainable.

instead of this:
```Python
from fastapi import FastAPI

app = FastAPI()


foo_api = APIRouter()
app.include_router(foo_api, prefix='/foo')



@foo_api.get('/', include_in_schema=False)
async def get_foo():
    pass


@foo_api.post('/', include_in_schema=False)
async def add_foo():
    pass
```

I want this:
```Python
from fastapi import FastAPI

app = FastAPI()

foo_api = APIRouter(include_in_schema=False)
app.include_router(foo_api, prefix='/foo')


@foo_api.get('/')
async def get_foo():
    pass


@foo_api.post('/')
async def add_foo():
    pass
```

### The solution you would like

A naive solution would be just defaulting to `self.include_in_router`  if it's not specified as argument in `fastapi.routing.APIRouter.add_api_route`

https://github.com/tiangolo/fastapi/blob/e5594e860f34c939d530640b1c240ed7f64003b1/fastapi/routing.py#L369

### Environment
```
Windows 10
fastapi==0.58.0
Python 3.8.1
```

[1]: https://fastapi.tiangolo.com/advanced/path-operation-advanced-configuration/#exclude-from-openapi

HarrySky:
[...]

PR #1651, opened 2020-06-30
Exclude all routes from schema under an APIRouter

HarrySky:
Proof of Concept for issue #1650.

You can pass `include_in_schema=False` to `APIRouter` to exclude all routes in schema from this router.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -477,7 +477,7 @@ class APIRouter(routing.Router):
             ),
             response_model_exclude_defaults=response_model_exclude_defaults,
             response_model_exclude_none=response_model_exclude_none,
-            include_in_schema=include_in_schema if self.include_in_schema else False
+            include_in_schema=include_in_schema if self.include_in_schema else False,
             response_class=response_class or self.default_response_class,
             name=name,
             dependency_overrides_provider=self.dependency_overrides_provider,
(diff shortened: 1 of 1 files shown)


fastapi/routing.py
```diff
@@ -475,7 +477,7 @@ def add_api_route(
             ),
             response_model_exclude_defaults=response_model_exclude_defaults,
             response_model_exclude_none=response_model_exclude_none,
-            include_in_schema=include_in_schema,
+            include_in_schema=include_in_schema if self.include_in_schema else False,
```
Kludex, line 480: Using the same variable name makes sense, and I wouldn't do it differently but can be confusing.
```suggestion
            include_in_schema=include_in_schema and self.include_in_schema,
```
Idk if my suggestion makes it smooth to read, no need to use it if you think it doesn't. I'll approve the PR anyway.

Kludex (approved):

fastapi/routing.py
abdusco: This means individual routes under a router cannot override the router's decision, if I got that right?

HarrySky: @abdusco only if `include_in_schema` on router is set to `False` 

All routes are provided with `include_in_schema=True` by default.

HarrySky: @Kludex great suggestion, readability-wise you are right, it will take a moment to understand.

I can commit this suggestion if maintainers think it is better, both are ok with me :smile:

PR #1659, opened 2020-06-30
Fix app.extra annotation

bharel:
`**kwargs` annotation is based on the value.
ATM extra is annotated like `Dict[str, Dict[str, Any]]` which causes issues with type-checkers.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -51,7 +51,7 @@ class FastAPI(Starlette):
         openapi_prefix: str = "",
         root_path: str = "",
         root_path_in_servers: bool = True,
-        **extra: Dict[str, Any],
+        **extra: Any,
     ) -> None:
         self.default_response_class = default_response_class
         self._debug = debug


Kludex:
What issues?

bharel:
@Kludex, type-checkers will complain that `app.extra["foo"] = 123` is invalid. That's because they think `extra` is a dict inside a dict.

Kludex:
@bharel Could you post the error that you got or is there a FastAPI issue about it? Do you mind providing more information? We can't even know which "type-checkers"...

Kludex:
I just saw that all the `**extra` in the [param_functions.py](https://github.com/tiangolo/fastapi/blob/master/fastapi/param_functions.py) are using the notation you recommend.

Kludex (approved):

bharel:
> @bharel Could you post the error that you got or is there a FastAPI issue about it? Do you mind providing more information? We can't even know which "type-checkers"...

All of them.
Here's an example:
image
It appears on all of the static type checkers, such as mypy, pycharm etc...

I doubt the build fails because of the PR.

tiangolo:
Great, good catch! :mag: 

For reference, it's described here: https://mypy.readthedocs.io/en/stable/getting_started.html?highlight=kwargs#more-function-signatures

PR #1672, opened 2020-07-05
Update isort command to use the new releases of isort v5.0.2

asheux:
Apparently, isort latest release ```isort v5.0.2``` is missing the ```--recursive or -rc``` flag and since the project uses the flag in it's script, it causes the tests to fail. An alternative to this is a dot, as in ```isort .```.

#### Missing flags
```
--apply
--recursive
```
#### Replacement
```
isort .
```

For more information on these changes see:
https://github.com/timothycrosley/isort/blob/develop/CHANGELOG.md#500-penny---july-4-2020

With this change, the build should be okay.

### DIFF
--- a/scripts/format-imports.sh
+++ b/scripts/format-imports.sh
@@ -2,5 +2,5 @@
 set -x
 
 # Sort imports one per line, so autoflake can remove unused imports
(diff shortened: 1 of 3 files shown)


Kludex:
Why have you closed this issue? Is it a bad idea to update isort?

Issue #1556, opened 2020-06-12
[FEATURE] Italian translations


csr:
### Italian translations tracking
We can use this Issue to track the progress of the Italian language translations. 🇮🇹 

Please post here what page you plan on translating so we can coordinate. 🚀

**Tips**
* Just like [we're doing with the Spanish translations](https://github.com/tiangolo/fastapi/issues/1218), let's use "tu" instead of the more formal "lei" or "voi".
* If the term is a technical term, leave in English and add an explanation of the meaning of the word in Italian.

Kludex:
For Portuguese there's a telegram channel, idk if you guys want to setup that as well... Is jfyk :)

csr:
Cool, thanks for letting me know @Kludex. I've created a telegram group for the Italian translations https://t.me/fastapiita :)

csr:
I'm currently in the process of adding Italian as a new language and translating `index.md`

csr:
Up next, I'll be translating `python-types.md`.

csr:
I'll be translating `index.md` and `first-steps.md` from the [Tutorial - User Guide](https://fastapi.tiangolo.com/tutorial/) next.

csr:
Translating `tutorial/path-params` next. If you think there are other pages with higher priority let me know. Somebody else suggested they wanted to translate /features in the Telegram chat so I'm not going to work on it.

csr:
I can take https://fastapi.tiangolo.com/help-fastapi/ next.

csr:
I can take https://fastapi.tiangolo.com/contributing/ next.

benzgrant:
@csr I'd like to contribute to the Italian translation. How can I proceed?

Kludex:
@csr Pick a page, inform here that you're going to work on it and you can contribute :tada:

csr:
@benzgrant Ciao! I suggest you first start taking a look at the translation guide in the docs, join the Italian Telegram group if you need help (link above), and review the pull requests that are waiting for review.

rpalloni:
Hello @csr, I'd like to contribute to IT docs as well. 
Found in docs: "Check in the issues to see if there's one coordinating translations for your language." I suppose you are the IT coordinator right? Where should I start? (Actually cheched the IT docs and /master/docs/it/docs/index.md is still in EN. Should I start from there?)

csr:
[...]

PR #1677, opened 2020-07-06
Translated path-params.md from Tutorial User Guide to Italian

csr:
Parent thread is #1556. "Path parameter" is a technical term that was not translated to Italian, but a note was added to the first instance with a possible Italian translation.


csr:
I'm not actively participating in the translation efforts any longer but I would encourage you to create a separate pull request to translate this page.

PR #1685, opened 2020-07-08
feat: return OpenAPI content type

danielgtaylor:
According to https://github.com/OAI/OpenAPI-Specification/issues/110, the OAI TSC approved the use of an official mime/content type for OpenAPI documents in 2018:

- `application/vnd.oai.openapi`
- `application/vnd.oai.openapi+json`

This PR updates FastAPI to set the right content type header when returning the OpenAPI at `/openapi.json`. It adds a small test to ensure it gets set properly.

Setting this content type helps tools which may try to access your API description understand what type of description it is, in this case OpenAPI. It's especially useful when the URI is *not* `/openapi.json` since it could be anything by setting the FastAPI `openapi_url` setting.


tiangolo:
Thanks! I see that the referenced issue has not been closed and there's no final decision, also in the spec I can't find any reference to it... do you have more pointers?

On the other side, when browsers see something other than JSON they just download the file, which makes it quite cumbersome to explore and debug, so I'm not sure about it.

I think one option could be to send that media type when the request contains an `Accept` header with it.

tiangolo:
As this hasn't had any more activity, I'm gonna close it. But if you or anyone else is interested in this with a particular use case, we could review it again.

Thanks for your interest and effort! :coffee:

Issue #1689, opened 2020-07-08
Add HTMLResponse to Jinja2 example (docs)


frafra:
### First check

* [X] I added a very descriptive title to this issue.
* [X] I used the GitHub search to find a similar issue and didn't find it.
* [X] I searched the FastAPI documentation, with the integrated search.
* [ ] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [ ] I already read and followed all the tutorial in the docs and didn't find an answer.
* [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [X] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

Some items are not applicable, as this issue is related to documentation only.

### Example

This is what is proposed in the documentation:

```Python
from fastapi import FastAPI, Request
from fastapi.staticfiles import StaticFiles
from fastapi.templating import Jinja2Templates

app = FastAPI()

app.mount("/static", StaticFiles(directory="static"), name="static")


templates = Jinja2Templates(directory="templates")


@app.get("/items/{id}")
async def read_item(request: Request, id: str):
    return templates.TemplateResponse("item.html", {"request": request, "id": id})
```

### Description

Automatically generated docs assumes that response is a JSON response, not an HTML one.

* Open the browser and call the endpoint `/docs`.
* It lists the `read_item` endpoint as it would reply with a JSON content.
* It would make sense to flag that entry point as returning HTML content.

### The solution you would like

Just add the correct content type as proposed here: https://fastapi.tiangolo.com/advanced/custom-response/#html-response

```Python
from fastapi import FastAPI, Request
from fastapi.responses import HTMLResponse
from fastapi.staticfiles import StaticFiles
from fastapi.templating import Jinja2Templates

app = FastAPI()

app.mount("/static", StaticFiles(directory="static"), name="static")


templates = Jinja2Templates(directory="templates")


@app.get("/items/{id}", response_class=HTMLResponse)
async def read_item(request: Request, id: str):
[...]

PR #1690, opened 2020-07-08
Add HTML mimetype

frafra:
Fix issue #1689


tiangolo:
Good idea! Thanks! :coffee: :cake:

PR #1704, opened 2020-07-11
WIP: Populate summary and description of an endpoint from its docstring

jasujm:
**Note!** Please see https://github.com/tiangolo/fastapi/issues/1633 for context why I think this is a good idea.

**Note!** This is still work in progress and missing, among other things:

- Documentation
- Unit testing the new changes
- I'll need to make the way to "parse" the summary and description lines from the docstring cover all cases, like badly formatted docstrings and different line ending styles.

I'm sharing this PR to gather feedback if the maintainers agree it is a good idea and I should proceed with cleaning up this patch.

--

This commit changes how the OpenAPI summary and description for an endpoint are populated from the docstring of the function implementing the endpoint.

Old behavior: Only the description is populated from the docstring

New behavior: Both the summary and the description are populated from
the docstring unless explicitly provided as decorator arguments

The new behavior is more aligned with PEP 257 specifying that a docstring consists of summary line and an extended description, neatly mapping to the corresponding fields in OpenAPI specification.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -336,8 +336,13 @@ class APIRoute(routing.Route):
             self.dependencies = list(dependencies)
         else:
             self.dependencies = []
-        self.summary = summary
-        self.description = description or inspect.cleandoc(self.endpoint.__doc__ or "")
+        if summary:
+            self.summary = summary
+            self.description = description or inspect.cleandoc(self.endpoint.__doc__ or "")
+        else:
+            doc_parts = inspect.cleandoc(self.endpoint.__doc__ or "").split("\n\n", 1)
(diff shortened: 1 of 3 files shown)


tiangolo:
Thanks for your interest!

Hmm, I see that it wouldn't be easy to *not* use that new behavior, and it would require everyone to adopt that convention that might not work very well for all the applications.

I also prefer to have the independent `summary` parameter that can be found with autocompletion.

That on top of being a breaking change that would be incompatible with existing applications. And that the summary is currently already auto-generated from the function name when not provided.

So I'll pass on this one, but thanks for your interest! :coffee:

PR #1714, opened 2020-07-15
fix useless for

Bloodielie:
the function went through dependencies twice despite the fact that it is a waste of resources


tiangolo:
Thanks! :rocket:

PR #1720, opened 2020-07-16
Compatible with pydantic exporting model's method

xl9211:


Kludex:
@xl9211 Could you explain what you're doing exactly?

But anyhow, I've noticed that `jsonable_encoders` has a `set()` as default parameter, we should remove it:
- https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html?highlight=mutable

Code snippet examplifying problem:
```
def func(a: int, param = set()):
    param.add(a)
    print(param)

func(1) # prints "{1}"
func(2) # prints "{1, 2}"
```

tiangolo:
Thanks for your interest, it seems you are assuming Pydantic 1.x. I can't remove the backward-compatibility code until I deprecate Pydantic 0.32.0.

I should do that soon, but this can't be done before that.

So I'll pass on this for now. Thanks for your effort! :cake: :coffee:

PR #1723, opened 2020-07-17
Compatible with pydantic exporting model's method

xl9211:
Compatible with pydantic exporting model's method

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -30,7 +30,7 @@ encoders_by_class_tuples = generate_encoders_by_class_tuples(ENCODERS_BY_TYPE)
 def jsonable_encoder(
     obj: Any,
     include: Union[SetIntStr, DictIntStrAny] = None,
-    exclude: Union[SetIntStr, DictIntStrAny] = set(),
+    exclude: Union[SetIntStr, DictIntStrAny] = None,
     by_alias: bool = True,
     skip_defaults: bool = None,
     exclude_unset: bool = False,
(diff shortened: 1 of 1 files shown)


Kludex:
@xl9211 Can you close this one? It's exactly the same as #1720

fastapi/encoders.py
```diff
@@ -45,10 +45,6 @@ def jsonable_encoder(
             "exclude_unset to keep in line with Pydantic v1, support for it will be "
             "removed soon."
         )
-    if include is not None and not isinstance(include, set):
-        include = set(include)
-    if exclude is not None and not isinstance(exclude, set):
-        exclude = set(exclude)
```
MashhadiNima, line 51: Why did you remove these line?
Considering the type of variables, i think there is no possibility of removing these lines.

fastapi/encoders.py
```diff
@@ -30,7 +30,7 @@ def generate_encoders_by_class_tuples(
 def jsonable_encoder(
     obj: Any,
     include: Union[SetIntStr, DictIntStrAny] = None,
-    exclude: Union[SetIntStr, DictIntStrAny] = set(),
```
MashhadiNima, line 33: When you set "None" as default,  you will need to check condition of existence in the rest of the code.

fastapi/encoders.py
xl9211: When exclude is dict, that thing were wrong.
I user pydantic's advanced case.

Also see:
https://pydantic-docs.helpmanual.io/usage/exporting_models/#advanced-include-and-exclude

tiangolo:
Thanks for your interest, it seems you are assuming Pydantic 1.x. I can't remove the backward-compatibility code until I deprecate Pydantic 0.32.0.

I should do that soon, but this can't be done before that.

So I'll pass on this for now. Thanks for your effort! :cake: :coffee:

PR #1730, opened 2020-07-17
Remove mutable types from default arguments

SirTelemak:
As noticed by @Kludex [here](https://github.com/tiangolo/fastapi/pull/1720#issuecomment-660274476) there set and dict in defaults. Current logic for `exclude` already covers case with None, and `custom_encoder` got checks `if custom_encoder:` which cover None case but I can add additional check for dict which is unnecessary imo
Also I added Optional TypeHint for keyword arguments with None as default


SirTelemak:
3.6 build failing because of last pydantic update. I can dig a bit more to check how to mitigate this later

fastapi/encoders.py
```diff
@@ -96,7 +96,7 @@ def jsonable_encoder(
                     or (not key.startswith("_sa"))
                 )
                 and (value is not None or not exclude_none)
-                and ((include and key in include) or key not in exclude)
+                and ((include and key in include) or not exclude or key not in exclude)
```
MashhadiNima, line 99: I think your code looks wrong. When "not exclude" is true , "key not in exclude " does not check.
```suggestion
                and ((include and key in include) or (not exclude and key not in exclude))
```

MashhadiNima, line 99: I think this line needs change. When "exclude" is None, "not exclude" is true and "key not in exclude" does not check. 
```suggestion
                and ((include and key in include) or (exclude and key not in exclude))
```

blkst8 (changes requested):

fastapi/encoders.py
SirTelemak: you can check my previous commit: https://github.com/tiangolo/fastapi/pull/1730/commits/3be96857187d1e72f21777bf709455c503ead87d I did exactly the same, but it won't work:
There we check do we need this value or not, you suggestion: `If include exists and key in include we need it, or if exclude exists and key not in exists `
But what about case when exclude not exists or empty? In this case we return False and doesn't return value
My logic:
`If include exists and key in include we need it, or if exclude not exists or key not in exists` - It case when exclude not exists or empty.

MashhadiNima: Your logic was excellent.
Now I realized why my change doesn't work.
Thank you for your patch and sorry for my mistakes.

SirTelemak: No problem :) As I said, I made this mistake too

SirTelemak:
Found several more cases with mutables as default values, will make separate PR for it later

tiangolo:
Thanks for your interest!

There's actually no need for that, those container types should only be dangerous if they are mutated by the functions. But they are never mutated, their content is checked or they are copied, so it's fine to leave it as it is.

But thanks for your effort! :coffee: :cake:

SirTelemak:
Then there is unnecessary line https://github.com/tiangolo/fastapi/blob/52f0f8657e93c5ad72568e7b71dafb8c4fc9f9d6/fastapi/encoders.py#L48
since according to annotation `exclude` not Optional and can't be None
And another moment (more beauty then functionality) - include and exclude have common behavior but one have None as default and second - set.
Both makes no difference in functionality and great as it is, just a matter of taste

PR #1731, opened 2020-07-18
 Add typing.Optional to variables that accept None as value 

blkst8:


tiangolo:
Awesome, thanks! :rocket: :tada:

PR #1754, opened 2020-07-21
Simplify and improve jsonable_encoder

blkst8:


uriyyo:
Hi @MashhadiNima, 

I update this function in my PR #1753.
I used `collections.defaultdict` in order to replace usage of `dict.setdafault`.

MashhadiNima:
Hi dear @uriyyo
Thank you for informing me of this.
My goal was to eliminate the second `for` loop. In the next commit, I will use `collections.defaultdict` to make the code cleaner.

uriyyo:
> My goal was to eliminate the second `for` loop. In the next commit, I will use `collections.defaultdict` to make the code cleaner.

@MashhadiNima  What is the reason to eliminate the second `for` loop?
There can be a performance issue if we will use `tuple` concatenation in every `for` iteration.
```python
encoders_by_class_tuples[encoder] += (type_,)  # this line will create new tuple every time
```

Correct me if I missed smth.

MashhadiNima:
I eliminate the second `for` loop for better performance.
`encoders_by_class_tuples[encoder] += (type_,)` create new tuple in every time but it's faster than cascades each list to tuple.
I tested it again and found cascading got bad performance.

uriyyo:
@MashhadiNima My bad, just tested it out, you are totally right.

tiangolo:
Clever! :nerd_face: 

And thanks for testing it @MashhadiNima and @uriyyo ! :zap: 

Thanks for your contribution @MashhadiNima ! :cake: :coffee:

PR #1755, opened 2020-07-21
Remove FastAPIError class.

blkst8:


SirTelemak:
Why?

fastapi/utils.py
```diff
@@ -103,7 +103,7 @@ def create_response_field(
         else:  # pragma: nocover
             return response_field(schema=field_info)
     except RuntimeError:
-        raise fastapi.exceptions.FastAPIError(
+        raise RuntimeError(
             f"Invalid args for response field! Hint: check that {type_} is a valid pydantic field type"
         )
```
VaultVulp, line 108: Why should we stop with the `RuntimeError`? Better use `raise Exception`, for `RuntimeError` is a descendant of an `Exception` class.

MashhadiNima: `Exception` is general. Using `RuntimeError` when the exception is `RuntimeError` makes the code more intelligible.

MashhadiNima:
> Why?

Because the class ‍‍‍"FastAPIError" inherits "RuntimeError" and doesn't have anything additional.
In my opinion, it would be better to use "RuntimeError" directly.

SirTelemak:
> > Why?
> 
> Because the class ‍‍‍"FastAPIError" inherits "RuntimeError" and doesn't have anything additional.
> In my opinion, it would be better to use "RuntimeError" directly.

FastAPIError should'n add anything additional, it just helps you to find the source of mistake. RuntimeError is built-in and can be evoked by several other libraries and it will be hard to find exception source. But `FastAPIError` is specific for FastAPI. 
For example, you now your code can raise FastAPIError and what to mitigate this, you do something like this
```python
try:
    func_that_may_rise_error()
except FastAPIError:
    log.info('it\'s okay, we can handle this')
    handle_error()
except RuntimeError:
    log.exception('something weird happened, take a look')
    raise
```

fastapi/utils.py
jansedlon: So does `FastAPIError`

MashhadiNima:
Thanks for your explanation. 
My idea isn't good. sorry to all. I will close this PL.

tiangolo:
Thanks for the help here @SirTelemak ! :cake: 

And thanks @MashhadiNima for the interest. :nerd_face:

Issue #725, opened 2019-11-21
[FEATURE] Enable Dark Mode for official FastAPI website


ka-zo:
**Is your feature request related to a problem? Please describe.**
The feature request is related to the website https://fastapi.tiangolo.com/

**Describe the solution you'd like**
It would be nice to have dark mode, that could either be manually switched on, and/or it could automatically check the time, and if it is past sunset, it would automatically switch on dark mode.

**Describe alternatives you've considered**
Firefox/Edge plugins for dark/night mode.

**Additional context**
My eyes are burning :) Jokes aside, consider please the solution implemented by the docker website: https://docs.docker.com/

frankie567:
The documentation is generated using [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/). Apparently, dark mode support is scheduled for v5: https://github.com/squidfunk/mkdocs-material/issues/1305

So I think it will be available at some point, but it's not something related to FastAPI.

ka-zo:
Thanks for the feedback. Even though dark mode is not directly related to FastAPI, please consider providing the readers the dark mode option on the FastAPI homepage as soon as the feature is available.

Thanks in advance!

tiangolo:
Thanks for the help here @frankie567 !

Yep, we'll have to wait first for it to be available in Material for MkDocs before it can be integrated here.

ka-zo:
Thanks, I appreciate your proactivity!

tiangolo:
Cool, as this is currently not dependant on FastAPI itself, I'm gonna close this issue now. But feel free to add more comments or create new issues.

AngliD:
With new dark theme #1760 it's become nearly impossible to read code examples.
<img width="693" alt="image" src="https://user-images.githubusercontent.com/18736922/88180721-7a471300-cc36-11ea-9b9d-34fc179fd167.png">

Is it possible to create a switcher for light/dark theme? I don't want to change my OS theme just for one documentation   😄

Kludex:
Can we revert it until we have the switch button? We can wait for this (maybe): https://github.com/squidfunk/mkdocs-material/issues/1732

abjoker:
Hello,

How can I convert back to light theme, it is very to difficult to read now.

Appreciate your work
Thank you!

squidfunk:
Syntax highlighting was significantly improved in 5.5.0 which was just released!

[...]

PR #1760, opened 2020-07-21
💄 Use OS preference theme

adriencaccia:
Related to #725 

MkDocs [implemented a feature](https://squidfunk.github.io/mkdocs-material/getting-started/#color-scheme) to set the theme based on the user's OS preference.

This is what the website will look like if the user's preferred theme is dark:
image

For the light theme users:
image

I placed my cursor on `Languages` so that you can see that the accent is still there.

### DIFF
--- a/docs/en/mkdocs.yml
+++ b/docs/en/mkdocs.yml
@@ -4,6 +4,7 @@ site_url: https://fastapi.tiangolo.com/
 theme:
   name: material
   palette:
+    scheme: preference
     primary: teal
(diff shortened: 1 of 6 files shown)


Kludex (approved):

tiangolo:
Great, thanks for your contribution @adriencaccia ! :cake: :tada: 

And thanks for taking the time to make the detailed screenshots :dart: 

Thanks @Kludex for the review! :coffee:

PR #1769, opened 2020-07-22
Encoding subclass of a model with json_encoders

henrybetts:
Currently when encoding a model, the `json_encoders` property is ignored if it is defined in a superclass of the model. For example;

```python
from datetime import datetime, timezone


class ModelWithCustomEncoder(BaseModel):
    class Config:
        json_encoders = {
            datetime: lambda dt: dt.replace(
                microsecond=0, tzinfo=timezone.utc
            ).isoformat()
        }


class ModelWithCustomEncoderSubclass(ModelWithCustomEncoder):
    dt_field: datetime

    class Config:
        pass


model = ModelWithCustomEncoderSubclass(dt_field=datetime(2019, 1, 1, 8))

print(model.json())
# {"dt_field": "2019-01-01T08:00:00+00:00"}

print(jsonable_encoder(model))
# {'dt_field': '2019-01-01T08:00:00'}

```

To fix this, I modified `jsonable_encoder` to get the config using the `.__config__` property.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -50,7 +50,7 @@ def jsonable_encoder(
     if exclude is not None and not isinstance(exclude, set):
         exclude = set(exclude)
     if isinstance(obj, BaseModel):
-        encoder = getattr(obj.Config, "json_encoders", {})
+        encoder = getattr(obj.__config__, "json_encoders", {})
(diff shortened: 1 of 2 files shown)


tiangolo:
Excellent, thank you! :rocket: :tada:

PR #1786, opened 2020-07-24
Update external links

Kludex:
IT HAS SO MANY IMAGES! :tada:


tiangolo:
Hehehe thanks @Kludex ! :tada: :memo:

PR #1796, opened 2020-07-27
🌐 Add Portuguese translation for Python Types

izaguerreiro:


docs/pt/docs/python-types.md
```diff
@@ -0,0 +1,314 @@
+# Introdução aos tipos Python
+
+**Python 3.6 +** tem suporte para "dicas de tipo" opcionais.
```
Kludex, line 3: I'm not sure about translating "type hint" into "dicas de tipo". In the main page (https://fastapi.tiangolo.com/pt/) we have used the English version instead. What do you think?

izaguerreiro: I was in doubt, but I agree with you, I think keeping the standard in English in this case is better. I'll fix that.

izaguerreiro: Done!! 🙂

cassiobotaro (approved):
Obrigado pela contribuição @izaguerreiro !

docs/pt/docs/python-types.md
```diff
@@ -0,0 +1,316 @@
...
+
+Como o tipo, coloque a `List`.
+
+Como a lista é um tipo que contém alguns tipos internos, você os coloca entre colchetes:
+
+```Python hl_lines="4"
+{!../../../docs_src/python_types/tutorial006.py!}
+```
+
+!!! dica
```
jessicapaz, line 176: I think you need to keep the `tip` word to mark this as a tip box like this (same thing with info in other places):
image
image


```suggestion
!!! tip "Dica"
```

izaguerreiro: Well observed, I will change. Thank you ☺️

jessicapaz (approved):

tiangolo:
Awesome! Obrigado @izaguerreiro ! :rocket: :coffee: 

And thanks @Kludex , @cassiobotaro , and @jessicapaz for the reviews! :clap: :clap: :cake:

Issue #843, opened 2020-01-09
[FEATURE]: Add all parameters to the 'info' section of the OpenAPI specification


ka-zo:
### Is your feature request related to a problem
I cannot display the author details, license details, etc. nicely in the automatically generated documentation and in the OpenAPI specification. I want to be able to do this, when I call the app = FastAPI() statement.

### The solution you would like

I want to see this additional metadata in the info section of the generated OpenAPI specification. In order to achieve this, I need the FastAPI() call to support additional parameters, like termsOfService, contact and license.

### Describe alternatives you've considered

no alternatives.

### Additional context

Check OpenAPI specification.

ka-zo:
I found the answer to my own question / request. It is possible to extend the generated OpenAPI specification with new fields, if you follow the guideline described in this link: https://fastapi.tiangolo.com/tutorial/extending-openapi/

This ticket can be closed, unless the developer team wants the FastAPI() call to support all the info parameters described here: https://swagger.io/specification/#infoObject

tiangolo:
Thanks @ka-zo for reporting back your solution.

I think we could add an additional parameter to the `FastAPI` class to take the extra OpenAPI data. But I'm not sure yet about names, how many, and structure. And as this has not been commonly requested, it hasn't had a high priority.

I would accept a PR with it if someone was interested in it, though.

With respect to this issue, I think that for now we can take it as a question and close it. And if this comes up later and more people are requesting it and the current method is too cumbersome, we can think of adding it.

ka-zo:
Agreed, and the solution described in the aforementioned link (https://fastapi.tiangolo.com/tutorial/extending-openapi/) is already satisfactory.

tiangolo:
There's direct support for this now in FastAPI `0.68.0` 🔖🎉

https://fastapi.tiangolo.com/tutorial/metadata/

PR #1812, opened 2020-07-30
✨ Add additonal OpenAPI metadata parameters to `FastAPI` class, shown on the automatic API docs UI

dkreeft:
As described in #843 that I needed and stumbled upon. This PR adds extra metadata parameters to the code and docs. So now you can specify contact and license information as well as terms of service.

I tried to keep the naming of the parameters in line with the [specification](https://swagger.io/specification/), but unfortunately `license` is a Python built-in, so I went with `license_info` instead.


tests/test_tutorial/test_metadata/test_tutorial001.py
```diff
@@ -7,9 +7,19 @@
 openapi_schema = {
     "openapi": "3.0.2",
     "info": {
-        "title": "My Super Project",
-        "version": "2.5.0",
-        "description": "This is a very fancy project, with auto docs for the API and everything",
+        "title": "My API",
+        "version": "0.0.1",
+        "description": "This is my API, with auto docs for the API and everything",
```
Kludex, line 12: Is there a reason for changing the fields `title`, `version` and `description` here?

fastapi/applications.py
```diff
@@ -36,6 +36,9 @@ def __init__(
         title: str = "FastAPI",
         description: str = "",
         version: str = "0.1.0",
+        terms_of_service: Optional[str] = "",
+        contact: Optional[Dict[str, Union[str, Any]]] = None,
```
Kludex, line 40: Don't we need to validate the `keys` on the `contact` and `license_info`?

tests/test_tutorial/test_metadata/test_tutorial001.py
dkreeft: Yes, the title and description did not reflect what they should be (`My Super Project`) imo.

fastapi/applications.py
dkreeft: We could, but that is not really done for other arguments either.

fastapi/applications.py
```diff
@@ -36,6 +36,9 @@ def __init__(
         title: str = "FastAPI",
         description: str = "",
         version: str = "0.1.0",
+        terms_of_service: Optional[str] = "",
```
Kludex, line 39:
```suggestion
        terms_of_service: str = "",
```

fastapi/openapi/utils.py
```diff
@@ -330,13 +330,22 @@ def get_openapi(
     version: str,
     openapi_version: str = "3.0.2",
     description: str = None,
+    terms_of_service: str = None,
```
Kludex, line 333:
```suggestion
    terms_of_service: Optional[str] = None,
```

dkreeft:
@Kludex do you mind approving this PR? Thanks!

dkreeft:
@tiangolo any ETA for approving this request? In my team we are waiting for this change.

tiangolo:
Thanks @dkreeft! :rocket: 

I updated it a bit, updated the type annotations, order of arguments, and the docs.

This will be available later today in FastAPI version `0.68.0` 🔖🎉

Issue #1814, opened 2020-07-30
Empty message about raised exception in documentation code example (Handling Errors)


manlix:
### Issue description

I have tried to run example code from documentation (section **[Handling Errors](https://fastapi.tiangolo.com/tutorial/handling-errors/#re-use-fastapis-exception-handlers)**) and got empty message in console logs about raised exception (**HTTPException**).

### Actual and unexpected result in logs — no details about raised exception:

```
OMG! An HTTP error!: 
INFO:     127.0.0.1:50364 - "GET /items/3 HTTP/1.1" 418 
```

### Expected result — with details about raised exception:

```
OMG! An HTTP error!: HTTPException(status_code=418, detail="Nope! I don't like 3.")
INFO:     127.0.0.1:50364 - "GET /items/3 HTTP/1.1" 418
```

### Code from documentation:

```Python
from fastapi import FastAPI, HTTPException
from fastapi.exception_handlers import (
    http_exception_handler,
    request_validation_exception_handler,
)
from fastapi.exceptions import RequestValidationError
from starlette.exceptions import HTTPException as StarletteHTTPException

app = FastAPI()


@app.exception_handler(StarletteHTTPException)
async def custom_http_exception_handler(request, exc):
    print(f"OMG! An HTTP error!: {exc}")
    return await http_exception_handler(request, exc)


@app.exception_handler(RequestValidationError)
async def validation_exception_handler(request, exc):
    print(f"OMG! The client sent invalid data!: {exc}")
    return await request_validation_exception_handler(request, exc)


@app.get("/items/{item_id}")
async def read_item(item_id: int):
    if item_id == 3:
        raise HTTPException(status_code=418, detail="Nope! I don't like 3.")
    return {"item_id": item_id}
```

### How to reproduce:

Run in console:
```Bash
manlix@lab:~$ curl http://localhost:8000/items/3
{"detail":"Nope! I don't like 3."}
```

### The cause of the problem:

**Starlette**'s **HTTPException** hasn't **\_\_str\_\_()** method:

```Python
class HTTPException(Exception):
    def __init__(self, status_code: int, detail: str = None) -> None:
        if detail is None:
            detail = http.HTTPStatus(status_code).phrase
        self.status_code = status_code
        self.detail = detail

    def __repr__(self) -> str:
        class_name = self.__class__.__name__
        return f"{class_name}(status_code={self.status_code!r}, detail={self.detail!r})"
```

tiangolo:
[...]

PR #1815, opened 2020-07-30
Fix empty message about raised exception

manlix:
Fix for issue in documentation example code #1814

### DIFF
--- a/docs_src/handling_errors/tutorial006.py
+++ b/docs_src/handling_errors/tutorial006.py
@@ -11,7 +11,7 @@ app = FastAPI()
 
 @app.exception_handler(StarletteHTTPException)
 async def custom_http_exception_handler(request, exc):
-    print(f"OMG! An HTTP error!: {exc}")
+    print(f"OMG! An HTTP error!: {repr(exc)}")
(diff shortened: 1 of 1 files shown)


Kludex (approved):
Do we have the same problem on next exception handler (`RequestValidationError`)? 

But nice catch! :tada:

manlix:
> Do we have the same problem on next exception handler (`RequestValidationError`)?
> 
> But nice catch!

@Kludex No, we don't. Next exception handler is correct:

```Bash
manlix@lab:~$ curl http://localhost:8000/items/incorrect
{"detail":[{"loc":["path","item_id"],"msg":"value is not a valid integer","type":"type_error.integer"}]}
```

Output in logs has expected details in plain text:

```Bash
OMG! The client sent invalid data!: 1 validation error for Request
path -> item_id
  value is not a valid integer (type=type_error.integer)
INFO:     127.0.0.1:50902 - "GET /items/incorrect HTTP/1.1" 422 Unprocessable Entity
```

tiangolo:
Awesome, good catch @manlix ! :rocket: :cake: 

Thanks for the help @Kludex ! :coffee:

PR #1824, opened 2020-08-02
pencil2: Fix awkward pronoun usage

Mause:


Kludex (approved):
Nice catch! :tada:

acnebs:
A lot of these are talking about "the client". Is that not more of an "it"?

tiangolo:
Thanks a lot! :mag: :memo: 

I want to have a welcoming space in the FastAPI community and I don't want to have language that might feel unwelcoming. :rocket: 

At the same time, one of my main objectives in the docs is to reduce ambiguity to make it as easy to learn and understand as possible, even for people that might not be native English speakers (like myself), so I updated those examples a bit to make them non-gender specific without having to deal with the ambiguity of possible plurals.

PR #1827, opened 2020-08-03
add support for injecting HTTPConnection

nsidnev:
Hi! 

**TL;DR**:
This PR adds support for injecting `starlette.requests.HTTPConnection` into FastAPI views/dependencies.

**Long reasoning**:
I'm building an application that defines a REST + websocket interface to the outside world. And as storages for this application, I use Redis + PostgreSQL directly throught the drivers (aioredis + asyncpg). Connections are managed through pools that I store in application's state (`app.state.pg_pool`/` app.state.redis_pool`) and queries are made using custom gateways that I initialize in dependencies. Currently FastAPI does not allow using` HTTPConnection` in views, so I need to define duplicate dependencies to access pools via HTTP `Request` and` Websocket`, or use dynamic attributes that break typing/autocompletion, so I don't like this approach. With this change, it is possible to define a single dependency for accessing common request attributes such as app/state/url/headers/etc.


tiangolo:
Thanks for all the clarification and arguments, excellent job with the PR @nsidnev ! :bow: :clap:

Thanks for your contribution. :cake: :coffee:

Issue #1165, opened 2020-03-25
ORJSONResponse options


plankter:
Please add a functionality to pass `orjson` options (https://github.com/ijl/orjson#option) to ORJSONResponse. For instance, `OPT_SERIALIZE_NUMPY` is quite useful in many scenarios.

raphaelauv:
it's possible and easily by doing this :

#1846 (comment)

You can probably close this issue, thank you

raphaelauv:
@plankter 

Can you close this issue since it's easily possible by extending the current class. Thank you

tiangolo:
Thanks for the interest @plankter! And thanks a lot for the help here @raphaelauv! :bow: 

I added this section to the docs explaining how to do that: https://fastapi.tiangolo.com/advanced/custom-response/#custom-response-class

Issue #1137, opened 2020-03-19
Allow setting orjson options in ORJSONResponse


acnebs:
This is a problem and a requested feature.

`orjson` takes a lot of options for dumping data, such as allowing UUID dumping with `option=orjson.OPT_SERIALIZE_UUID`. It would be nice if we could use these options when using `ORJSONResponse`.

As far as I'm aware, with the existing implementation of `ORJSONResponse`, there is not a way to do this, as orjson cannot be configured globally, and `ORJSONResponse` does not pass any args except the content to dump to `orjson.dumps`.

gmelodie:
@tiangolo this was also mentioned on #1165, is it something you'd want done? I'm up for it if so :)

Enacero:
@acnebs Do you want to set option, when you define default_response_class?

raphaelauv:
it's possible and easily by doing this :

https://github.com/tiangolo/fastapi/pull/1846#issuecomment-688553504

You can probably close this issue, thank you

tiangolo:
Thanks for the interest @acnebs! And thanks for the pointer @raphaelauv!

I added this section to the docs explaining how to make a custom response: https://fastapi.tiangolo.com/advanced/custom-response/#custom-response-class

PR #1846, opened 2020-08-07
Add orjson response factory with ability to pass option and default

Enacero:
Add orjson response factory, so that it will be possible to pass option property into ORJSONResponse.
Related to #1137 and #1165

### DIFF
--- a/fastapi/responses.py
+++ b/fastapi/responses.py
@@ -24,7 +24,7 @@ class ORJSONResponse(JSONResponse):
         return orjson.dumps(content, option=self.option)
 
 
-def create_or_json_class(options: int = 0) -> Type[ORJSONResponse]:
+def create_orjson_class(options: int = 0) -> Type[ORJSONResponse]:
     """
     Create a new ORJsonResponse class with additional options.
     """
--- a/fastapi/responses.py
+++ b/fastapi/responses.py
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, Type
 
 from starlette.responses import FileResponse  # noqa
 from starlette.responses import HTMLResponse  # noqa
@@ -17,7 +17,19 @@ except ImportError:  # pragma: nocover
 
 class ORJSONResponse(JSONResponse):
     media_type = "application/json"
+    option: int = 0
 
     def render(self, content: Any) -> bytes:
         assert orjson is not None, "orjson must be installed to use ORJSONResponse"
-        return orjson.dumps(content)
+        return orjson.dumps(content, option=self.option)
+
+
+def create_or_json_class(options: int = 0) -> Type[ORJSONResponse]:
+    """
+    Create a new ORJsonResponse class with additional options.
+    """
+
+    class _ORJSONResponse(ORJSONResponse):
+        option = options
+
+    return _ORJSONResponse
--- a/tests/test_responses.py
+++ b/tests/test_responses.py
@@ -1,0 +1,42 @@
+import orjson
+from fastapi import FastAPI
+from fastapi.responses import create_orjson_class
+from fastapi.testclient import TestClient
+
+app = FastAPI()
+
+PAYLOAD = {"payload": {"message": "Hello World"}}
+
+
+@app.get("/orjson", response_class=create_orjson_class())
+def get_orjson():
+    return PAYLOAD
+
+
+@app.get("/orjson_with_option", response_class=create_orjson_class(orjson.OPT_INDENT_2))
+def get_orjson_with_option():
+    return PAYLOAD
+
+
+@app.get("/json")
+def get_json():
+    return PAYLOAD
+
+
+client = TestClient(app)
+
+
+def test_orjson_response():
+    response = client.get("/orjson")
+    assert response.status_code == 200
+    assert response.content == b'{"payload":{"message":"Hello World"}}'
+
+    response = client.get("/orjson_with_option")
+    assert response.status_code == 200
+    assert (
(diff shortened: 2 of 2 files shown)


victoraugustolls:
@Enacero great PR! Was looking for something like this in `FastAPI`. Do you think it is possible to include in the factory the `default` parameter to `orjson`?

Enacero:
@victoraugustolls yes, I will add this with next commit for this pull request.

Enacero:
@victoraugustolls I've tried a lot of variants, but found out, that it's unnecessary to add default to orjson factory, since every response in fast_api is being pre encoded, so that all possible types are already converted by json_encoder and if type is not pre serializable  by json_encoder, error with be raised. So for now, it's useless to add default to orjson factory.

victoraugustolls:
Hi! Why useless? For example, asyncpg returns an instance of "Record" class when you fetch a row from the database. This class is not json encodable but with by passing a default to orjson by casting it to a dict, it is.

Enacero:
@victoraugustolls The thing is that the response is fisrt encoded by json_encoder and than passed to orjson.dumps. So if it's not encoded by json_encoder, it won't be passed to orjson.dumps

victoraugustolls:
@Enacero now I understand! Thanks for your explanation!

fastapi/responses.py
```diff
@@ -17,7 +17,19 @@
 
 class ORJSONResponse(JSONResponse):
     media_type = "application/json"
+    option: int = 0
```
raphaelauv, line 20: why not just extend the `__init__` ?

```python
    def __init__(self, option, **kwds):
        self.option = option
        super().__init__(**kwds)
```

havok2063:
@Enacero I fully support this PR as I need to be able to pass options into the `orjson.dumps` method.  In particular I need to be able to serialize numpy arrays.  I tried your code changes but it didn't seem to work.  Can you verify that serialization of numpy arrays works with your suggested changes?  

This was my tested route based off your file changes and your tests:
```
import numpy as np

@app.get("/file/{ext}", response_class=create_orjson_class(orjson.OPT_SERIALIZE_NUMPY))
async def fits_ext(ext: str):
    hdr = f'this is a string test {ext}'
    data = np.array([[1,2,3],[4,5,6]])
    results = {'header':hdr, 'data': data}
    return results
```
A `requests.get` returns a 500 response and I get the following error in the server terminal  `ValueError: [ValueError('dictionary update sequence element #0 has length 3; 2 is required'), TypeError('vars() argument must have __dict__ attribute')]`

If I bypass your code changes completely and return a standard `Response` I can get something to work, but it's not as elegant as your solution, e.g.
```
@app.get("/file/{ext}")
async def fits_ext(ext: str):
    hdr = f'this is a string test {ext}'
    data = np.array([[1,2,3],[4,5,6]])
    results = {'header':hdr, 'data': data}
    return Response(orjson.dumps(results, option=orjson.OPT_SERIALIZE_NUMPY), media_type='application/json')
```

raphaelauv:
@havok2063 

```python
class ORJSONResponseCustom(JSONResponse):
    media_type = "application/json"

    def render(self, content: Any,option=Optional[int]=None) -> bytes:
        assert orjson is not None, "orjson must be installed to use ORJSONResponse"
        return orjson.dumps(content,option=option)
```

and do this

```python

    return ORJSONResponseCustom(results, option=orjson.OPT_SERIALIZE_NUMPY))
```

fastapi/responses.py
Enacero: @raphaelauv I need somehow to store option so that when ORJson response is being initiated, it must get option from context not from constructor

havok2063:
@raphaelauv This doesn't work as you have it written.  I get `    return ORJSONResponseCustom(results, option=orjson.OPT_SERIALIZE_NUMPY)
TypeError: __init__() got multiple values for argument 'option'`

raphaelauv:
Well it work perfectly on my with python 3.7 and latest version of fastAPI.

You need to set **content=results** ! Do not pass it directly to ORJSONResponseCustom

raphaelauv:
Seems easy to extends the intern logic , I don't see the purpose of this PR

havok2063:
@raphaelauv Thanks, that did the trick.  

I think the purpose is that it shouldn't be necessary for every user to customize the class themselves.  The `ORJSONResponse` should be designed to allow someone to take full advantage of what `orjson` offers.  It's a pretty straightforward PR to implement and merge in.

tiangolo:
Thanks for the effort @Enacero! :coffee: 

Thanks for the discussion here everyone. :cake: 

I prefer not to add more responses and ways to generate responses but instead to explain how to create them. I agree with @raphaelauv that the best solution would be to customize the response in your code rather than figuring out ways to allow many combinations of parameters in FastAPI, because in the end users would end up having to learn that, and they could have just learned how to create a custom response instead.

I added [this PR](https://github.com/tiangolo/fastapi/pull/5331) with a section in the docs to explain how to create a custom orjson response: https://fastapi.tiangolo.com/advanced/custom-response/#custom-response-class

And yes, if you want to avoid the automatic data encoding with the `jsonable_encoder` you should return a response directly. I hope to add a way to customize how data is encoded. Maybe directly in the responses, maybe allowing to provide a function to the FastAPI class, routers, and path operations to do the encoding (that by default would be the `jsonable_encoder`). This probably also would enable using other content data types more easily, like XML (not that I like XML, YAML, TOML, etc).

Given all that, I'll close this PR now. But thanks for the effort! :coffee:

PR #1855, opened 2020-08-08
change "methods" assignment in APIRoute class

blkst8:
For better performance, I've removed cascading and replaced it with creating a set manually.
I've tested it for safety and found my idea has better performance ESPECIALLY when the input `methods` is `None`. It creates a set in about half the previously required time for cascading.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -305,7 +305,6 @@ class APIRoute(routing.Route):
         self.methods = (
             {"GET"} if methods is None else {method.upper() for method in methods}
         )
-
         self.unique_id = generate_operation_id_for_path(
             name=self.name,
             path=self.path_format,
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -302,11 +302,14 @@ class APIRoute(routing.Route):
         self.endpoint = endpoint
         self.name = get_name(endpoint) if name is None else name
         self.path_regex, self.path_format, self.param_convertors = compile_path(path)
-        if methods is None:
-            methods = ["GET"]
-        self.methods = set([method.upper() for method in methods])
+        self.methods = (
+            {"GET"} if methods is None else {method.upper() for method in methods}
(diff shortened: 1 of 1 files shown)


fastapi/routing.py
```diff
@@ -302,11 +302,13 @@ def __init__(
         self.endpoint = endpoint
         self.name = get_name(endpoint) if name is None else name
         self.path_regex, self.path_format, self.param_convertors = compile_path(path)
-        if methods is None:
-            methods = ["GET"]
-        self.methods = set([method.upper() for method in methods])
+        self.methods = (
+            {"GET"} if methods is None else {method.upper() for method in methods}
```
ycd, line 306: @MashhadiNima afaik iterating a set significantly slower than  than iterating a list and tuple.

```python
def test_tuple():
    for i in tuple(range(10000)):
        pass

def test_list():
    for i in list(range(10000)):
        pass

def test_set():
    for i in set(range(10000)):
        pass

print(timeit(test_tuple, number=10000))      -    3.80763795600069
print(timeit(test_list, number=10000))       -    3.8430613660020754
print(timeit(test_set, number=10000))        -    4.660357303000637
```

Why did you used set for here? I remember #1754 and it was pretty clever, did i miss smth?

MashhadiNima: In the original code , `self.methods`'s type is set .
In my opinion, this is done to build a unique set of input parameters.

in this part `{method.upper() for method in methods}` , I changed inputs and used set to create a unique set instead of creating a `list` of changed parameters and cascading it to set.

tiangolo:
Thanks for the interest @nimctl ! :nerd_face: 

And thanks for the comment @ycd ! :mag: 

This optimization would not really affect much, as that code is only called at application startup, once per path operation. It is not called again later per request. So I prefer to keep it simpler, as it is now.

But thanks for the interest! :coffee:

PR #1864, opened 2020-08-09
🔥 Remove old/unused parameter sqlalchemy_safe from jsonable_encoder

tiangolo:
🔥 Remove old/unused parameter `sqlalchemy_safe` from `jsonable_encoder`.

It was an early hack to allow returning SQLAlchemy models, but it was never documented, and the recommended way is using Pydantic's `orm_mode` as described in: https://fastapi.tiangolo.com/tutorial/sql-databases/

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -32,7 +32,6 @@ def jsonable_encoder(
     exclude_defaults: bool = False,
     exclude_none: bool = False,
     custom_encoder: dict = {},
-    sqlalchemy_safe: bool = True,
 ) -> Any:
     if include is not None and not isinstance(include, set):
         include = set(include)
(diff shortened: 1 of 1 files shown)


shawnwall:
I can't quite figure out if this was the commit that did it, but if you try to do the following now you get a stack overflow:

`obj_data = jsonable_encoder(some_sqla_model_instance)`

This breaks the update pattern found in https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/crud/base.py

mvilanova:
All our Pydantic models inherit from the same [base model](https://github.com/Netflix/dispatch/blob/develop/src/dispatch/models.py#L52) which has `orm_mode = True` and we're also getting a stack overflow.

tiangolo:
Thanks for pointing this out @shawnwall and @mvilanova ! :mag: :bug: 

I reverted it in #1987 , available in FastAPI version `0.61.1` (released today in a couple of hours).

PR #1872, opened 2020-08-10
✨ Import and re-export data structures from Starlette, used by Request properties, on `fastapi.datastructures`

### LINKED ISSUE #1871: Add imports of Starlette datastructures for type annotating Request
### First check

### Description

As a user of FastAPI, I want to type annotate functions that return properties of `fastapi.Request` without needing to import `starlette`, in keeping with the [documentation that states](https://fastapi.tiangolo.com/deployment/#about-starlette):

> You shouldn't pin the version of starlette.

Instead of the current state, where Starlette must be a pinned dependency to be able to use type annotations for the properties of `Request`:

```python
from fastapi import Request
from starlette.datastructures import URL  # Can only import URL from starlette; it isn't exposed in FastAPI

def example_get_url_from_request(request: Request) -> URL:
    return request.url
```

Instead, import all necessary properties in `fastapi.requests` such that the example could be updated to:

```python
from fastapi import Request
from fastapi.requests import URL

def example_get_url_from_request(request: Request) -> URL:
    return request.url
```

jamescurtin:
Closes https://github.com/tiangolo/fastapi/issues/1871


SirTelemak:
Why not import it in fastapi.datastructures for more compatible syntax? Imo it's unintuitive that starlette.datastructures models imported from fastapi.requests.
Moreover in your example:
```python3
def example_get_url_from_request(request: Request) -> URL:
```
there URL is not request, it's response which, again, unintuitive in my opinion.

jamescurtin:
> Why not import it in fastapi.datastructures for more compatible syntax? Imo it's unintuitive that starlette.datastructures models imported from fastapi.requests.

Good point; restructured such that the imports mirror the structure of starlette

fastapi/datastructures.py
```diff
@@ -1,5 +1,11 @@
 from typing import Any, Callable, Iterable, Type
 
+from starlette.datastructures import URL  # noqa: F401
```
Kludex, line 3: I don't know how to create a multiline suggestion, but:
```suggestion
from starlette.datastructures import (
    Address,
    FormData,
    Headers,
    QueryParams,
    State,
    UploadFile as StarletteUploadFile,
    URL
)
```

fastapi/__init__.py
```diff
@@ -1,7 +1,6 @@
 """FastAPI framework, high performance, easy to learn, fast to code, ready for production"""
 
 __version__ = "0.61.0"
-
```
Kludex, line 4: Readd this line, please

fastapi/types.py
```diff
@@ -0,0 +1 @@
+from starlette.types import Receive  # noqa: F401
```
Kludex, line 1: Repeated?

Kludex (commented):
My suggestions are only focused on the syntax. I'm not sure about the idea of the changes tho. I saw the issue about this, but I'm still not sure.

fastapi/__init__.py
jamescurtin: This diff was generated by running `bash scripts/format-imports.sh`. I just confirmed by re-adding it and running the format script; it removes the line again.

fastapi/datastructures.py
jamescurtin: In general I would agree, but the change you suggest is incompatible with the `bash scripts/format-imports.sh` script.

That script invokes `isort fastapi tests docs_src scripts --force-single-line-imports`, which splits each import to a single line, but does not propagate the `noqa` hint to each line. It then invokes `autoflake --remove-all-unused-imports ...` which deletes the imports.

fastapi/types.py
jamescurtin: Yes--thank you! (Though it was actually the one in `fastapi/requests.py` that was repeated)

jamescurtin:
👋  Checking in on this PR: would someone please be able to review? Thank you!

tiangolo:
Great, thanks @jamescurtin for your contribution! :rocket: 

Thanks @Kludex and @SirTelemak for the discussion!

I updated it a bit to keep the current import style that improves typing support in editors (e.g. autoimports).

This will be available later today in FastAPI version `0.68.0` 🔖🎉

PR #1880, opened 2020-08-12
➖ Do not require backports in Python >= 3.7

FFY00:
Signed-off-by: Filipe Laíns <lains@archlinux.org>


FFY00:
This should avoid some extra package installation for some users, as well as act as a reminder to remove the dependency and code when you move to Python 3.7 as the minimum version :blush:

aviramha (approved):

tiangolo:
Awesome, thanks @FFY00! :rocket:

PR #1895, opened 2020-08-14
📝 Add note in CORS tutorial about allow_origins with ["*"] and allow_credentials

dsmurrell:
Couldn't get credentials working with `allow_origins` set to `['*']`. Thought this might help others.

I was lucky to find this after a while: https://stackoverflow.com/questions/46288437/set-cookies-for-cross-origin-requests

which explains: "Also, make sure the HTTP header Access-Control-Allow-Origin is set and not with a wildcard *."

Specifying some origins, allows my axios request with `{ withCredentials: true }` set to work.


Kludex (approved):

tiangolo:
Good point! Thanks for the contribution @dsmurrell ! :rocket: :cake: 

And thanks @Kludex for the review :coffee:

PR #1904, opened 2020-08-16
✅ Add  the `docs_src` directory to test coverage and update tests

Kludex:
I've just spotted a typo in the test script:
```bash
pytest --cov=fastapi --cov=tests --cov=docs/src --cov-report=term-missing --cov-report=xml tests ${@}
```
`docs/src` is a typo, we want to cover `docs_src`.

If you don't trust me :trollface: you can check at the end of the test:
```bash
Coverage.py warning: Module docs/src was never imported. (module-not-imported)
``` 

I'll be increasing the coverage for those tests on this PR.

Report on the missing coverage:
image

**Changes:**
- Add `:skip-covered` in the terminal Pytest report. It makes it easy to read missing coverage.
- Fix test script.
    - Remove `--cov=tests` from test script
    - Fix typo `docs/src` -> `docs_src` 
- Fix double test call. 
    - If a function that starts with `test` is imported in the `tests`, there's no need for wrapping it. If you do so, it calls the test function twice.
- Add missing tests to be covered.

**After the fixes:**
image

EDIT:
November 4, 2020: This PR is still useful. And I've added the coverage on the `tests` once again.
November 10, 2020: Rebase it, everything is fine!


scripts/test.sh
```diff
@@ -7,4 +7,4 @@ bash ./scripts/lint.sh
 # Check README.md is up to date
 diff --brief docs/en/docs/index.md README.md
 export PYTHONPATH=./docs_src
-pytest --cov=fastapi --cov=tests --cov=docs/src --cov-report=term-missing --cov-report=xml tests ${@}
+pytest --cov=fastapi --cov=docs_src --cov-report=term-missing:skip-covered --cov-report=xml tests ${@}
```
Kludex, line 10:
```suggestion
pytest --cov=fastapi --cov=tests --cov=docs_src --cov-report=term-missing:skip-covered --cov-report=xml tests ${@}
```

Kludex:
@tiangolo I've rebased it, I'm not sure why the docs stage is failing on the CI, but I saw another PR failing with the same reason.

tiangolo:
Great, thanks @Kludex! :rocket: :heavy_check_mark: 

I updated it a bit, updated the docs where necessary to match the changed examples, and added a couple of missing tests for full coverage. Thanks! :cake: :coffee:

tests/test_tutorial/test_testing/test_main_b.py
```diff
@@ -0,0 +1,10 @@
+from docs_src.app_testing import test_main_b
+
+
+def test_app():
+    test_main_b.test_create_existing_item()
+    test_main_b.test_create_item()
+    test_main_b.test_create_item_bad_token()
+    test_main_b.test_read_inexistent_item()
+    test_main_b.test_read_item()
+    test_main_b.test_read_item_bad_token()
```
Kludex, line 10: This doesn't duplicate the tests as well?

tiangolo: If I remember correctly, the tests from `docs_src` are not run by pytest, so the way to run them and have coverage for that is to do this.

Kludex: I remember doing this because before my changes here, the tests were running twice, and I think (I didn't check) this makes them run twice, because (I think) the import is enough to make pytest find the tests. Anyway... :eyes:

tiangolo: Hum, maybe I have to review this again.

Although, one possible explanation is if you were running `pytest` and I was running `pytest tests`.

But anyway, yeah, maybe for later. 😅

PR #1912, opened 2020-08-17
✏️ Fix typo on dependencies utils and cleanup unused variable

Kludex:
Small PR.

**Changes**:
- Removed unused `param_supported_types = (str, int, float, bool)`
- Fix typo on `multlipart` -> `multipart`


tiangolo:
Nice! Good catch, thanks! :mag: :cake:

Issue #1879, opened 2020-08-12
How to access route parameter in a middleware?


mpdevilleres:
### First check

* [X] I added a very descriptive title to this issue.
* [X] I used the GitHub search to find a similar issue and didn't find it.
* [X] I searched the FastAPI documentation, with the integrated search.
* [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [X] I already read and followed all the tutorial in the docs and didn't find an answer.
* [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [X] After submitting this, I commit to one of:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

I am trying to add a custom parameter in the route and like to access it in a middleware.
eg.
```Python
from fastapi import FastAPI
app = FastAPI()

@app.get("/", action=['search'])  # I am interested in accessing the `action` parameter in a middleware
def read_root():
    return {"Hello": "World"}
```

Here's what I have done so far.

```Python
from typing import Callable, List, Any

import uvicorn
from fastapi.routing import APIRoute
from starlette.requests import Request
from starlette.responses import Response

from starlette.middleware.base import (
    BaseHTTPMiddleware,
    RequestResponseEndpoint
)

from fastapi import FastAPI, APIRouter


class AuditHTTPMiddleware(BaseHTTPMiddleware):
    # I want the `action` parameter from the route to be accessible in this middleware
    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
        response = await call_next(request)
        return response


class CustomAPIRoute(APIRoute):
    def __init__(self, path: str, endpoint: Callable, *, action: List[str] = None, **kwargs) -> None:
        super(CustomAPIRoute, self).__init__(path, endpoint, **kwargs)
        self.action = action


class CustomAPIRouter(APIRouter):
    def __init__(self):
        super(CustomAPIRouter, self).__init__()
        self.route_class = CustomAPIRoute

[...]

PR #1917, opened 2020-08-18
✨ Add support for custom attributes on custom APIRoute

Kludex:
I'm not sure if this PR will be accepted. :cry:
Because I've implemented without waiting @tiangolo 's answer about "if it's a good feature". :eyes: 

If you're reading this, and you think is a nice feature, please leave a :+1: 

**Changes:**
- `APIRoute` setup now withstand custom attributes. It makes possible the following code:
```python
@router.get("/", permissions=["Potato", "Tomato"])
def home(request: Request):
    return request.state.permissions
```

Issue: https://github.com/tiangolo/fastapi/issues/1879

Notes:
- If needed, I can add docs about it. :book: 
- If needed, I can add more tests. :white_check_mark: 

Edit (Aug 19): 
- Commit https://github.com/tiangolo/fastapi/pull/1917/commits/753330344a13d448e03a07f223cc9d2866ce5a71 changes slightly the idea, as you don't need to override the `http method` methods anymore. 
- MyPy ignores were added due to known MyPy issue: https://github.com/python/mypy/issues/1969

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -205,6 +205,7 @@ class FastAPI(Starlette):
         include_in_schema: bool = True,
         response_class: Optional[Type[Response]] = None,
         name: Optional[str] = None,
+        **kwargs: Dict[str, Any],
     ) -> None:
         self.router.add_api_route(
             path,
@@ -229,6 +230,7 @@ class FastAPI(Starlette):
             include_in_schema=include_in_schema,
             response_class=response_class or self.default_response_class,
             name=name,
+            **kwargs,  # type: ignore
         )
 
     def api_route(
@@ -255,6 +257,7 @@ class FastAPI(Starlette):
         include_in_schema: bool = True,
         response_class: Optional[Type[Response]] = None,
         name: Optional[str] = None,
+        **kwargs: Dict[str, Any],
     ) -> Callable:
         def decorator(func: Callable) -> Callable:
             self.router.add_api_route(
@@ -280,6 +283,7 @@ class FastAPI(Starlette):
                 include_in_schema=include_in_schema,
                 response_class=response_class or self.default_response_class,
                 name=name,
+                **kwargs,  # type: ignore
             )
             return func
 
@@ -341,6 +345,7 @@ class FastAPI(Starlette):
         response_class: Optional[Type[Response]] = None,
         name: Optional[str] = None,
         callbacks: Optional[List[routing.APIRoute]] = None,
+        **kwargs: Dict[str, Any],
     ) -> Callable:
         return self.router.get(
             path,
@@ -364,6 +369,7 @@ class FastAPI(Starlette):
             response_class=response_class or self.default_response_class,
             name=name,
             callbacks=callbacks,
+            **kwargs,  # type: ignore
         )
 
     def put(
@@ -390,6 +396,7 @@ class FastAPI(Starlette):
         response_class: Optional[Type[Response]] = None,
         name: Optional[str] = None,
         callbacks: Optional[List[routing.APIRoute]] = None,
+        **kwargs: Dict[str, Any],
     ) -> Callable:
         return self.router.put(
             path,
@@ -413,6 +420,7 @@ class FastAPI(Starlette):
             response_class=response_class or self.default_response_class,
             name=name,
             callbacks=callbacks,
+            **kwargs,  # type: ignore
         )
 
     def post(
@@ -439,6 +447,7 @@ class FastAPI(Starlette):
         response_class: Optional[Type[Response]] = None,
         name: Optional[str] = None,
         callbacks: Optional[List[routing.APIRoute]] = None,
+        **kwargs: Dict[str, Any],
     ) -> Callable:
         return self.router.post(
             path,
@@ -462,6 +471,7 @@ class FastAPI(Starlette):
             response_class=response_class or self.default_response_class,
             name=name,
             callbacks=callbacks,
+            **kwargs,  # type: ignore
         )
 
     def delete(
@@ -488,6 +498,7 @@ class FastAPI(Starlette):
         response_class: Optional[Type[Response]] = None,
         name: Optional[str] = None,
         callbacks: Optional[List[routing.APIRoute]] = None,
+        **kwargs: Dict[str, Any],
     ) -> Callable:
         return self.router.delete(
             path,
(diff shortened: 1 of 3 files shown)


podhmo:
Honestly, it is not good implementation, I think.
Because, this changes leads to lost typo-tolerance.

For example, consider the following typo when writing code:

```python
@router.post("/something", status=201) # not status, status_code is correct name.
def create(request: Request):
    ...
```

This code is valid, and mypy check is also OK. but return code is 200.

podhmo:
Instead of this, adding one parameter receiving metadata is better, I think.

```python
# "metadata" may not be a good name

@router.get("/", metadata=dict(permissions=["Potato", "Tomato"]))
def home(request: Request):
    return request.state.permissions
```

Kludex:
> Because, this changes leads to lost typo-tolerance.

Yeah, it's a trade-off. It can be hard to catch that.
Just to let people know:
```python
@router.post("/something", status=201) # not status, status_code is correct name.
def create(request: Request):
    ...
```
Leads to:
```bash
TypeError: post() got an unexpected keyword argument 'status'
```

On the other hand I want the freedom I propose. 

> Instead of this, adding one parameter receiving metadata is better, I think.
```python
# "metadata" may not be a good name

@router.get("/", metadata=dict(permissions=["Potato", "Tomato"]))
def home(request: Request):
    return request.state.permissions
```
The solution you propose is better than not having it at all. But it's not the same.

I'll propose something different.

```python
router = APIRouter(route_class=CustomAPIRoute, metadata=["permissions"])

@router.get("/", permissions=["Potato", "Tomato"]))
def home(request: Request):
    return request.state.permissions
```
This solution only allows the metadata fields and none of the `type-tolerance` is lost. 
An alternative, instead of `metadata` receiving a list, it can be a Pydantic model.

What do you think?

podhmo:
> What do you think?

In my opinion, I feel the proposed solution is too complicated, a bit.
Probably, we will need new mypy plugin to handle it. if possible, I want to skip implementing mypy plugin (usually, this is a heavy task) .

> On the other hand I want the freedom.

If freedom is needed, the request or request's state has an untyped dict, which can be set through the decorator. This is enough.

Also, if you want to limit the range of values, I think it's enough to define a function like this:

```py
def my_metadata(*, permissions: List[str]) -> Dict[str, Any]:
    return {"permissions": permissions}

@get(metadata=my_metadata(permissions=["Potato", "Tomato"]))  # too long?
def home(request: Request):
    return request.state.metadata["permissions"] # or request.state.permissions ?
```

podhmo:
Come to think of it, maybe we're talking too much about the implementation details, right? Before judging an idea?

On my side, if keeping typo-tolerance, just OK, I mean.

Kludex:
> In my opinion, I feel the proposed solution is too complicated, a bit.

Same complexity level as creating your own custom `APIRouter`. Which is allowed and described on our documentation.

> Probably, we will need new mypy plugin to handle it.

Why?

> If freedom is needed, the request or request's state has an untyped dict, which can be set through the decorator. This is enough.

I don't think the same way and the proposed code snippet is not something I would enjoy to use as a developer.

mpdevilleres:
first of all thanks for considering this issue# [1879](https://github.com/tiangolo/fastapi/issues/1879) a feature.

I think having below is more concise and easier to use for the developer.
`@router.get("/", permissions=["Potato", "Tomato"]))`

and introducing pydantic models to validate it is a good approach.

I hope this gets to the next release.

sedatcagdas (approved):

nberktumer (approved):
Tested with FastAPI 0.63.0 and it is working

Marc-Roig:
This would be a really nice feature 🙏

lawsong:
nice feature.we need.

tiangolo:
Thanks for the effort @Kludex ! :nerd_face: 

I think this use case can be solved simply with a dependency, and it doesn't require developers to create a custom router with custom logic that needs to fiddle with internal parts that get quite complex.

The equivalent of the example snippet in the description above would look like:

```python
@router.get("/", dependencies=[Depends(Permissions(["Potato", "Tomato"]))])
def home(request: Request):
    return request.state.permissions
```

---

And the full implementation is a simple [dependency class](https://fastapi.tiangolo.com/tutorial/dependencies/classes-as-dependencies/):

```python
from typing import List
from fastapi import FastAPI, Depends, Request


class Permissions:
    def __init__(self, permissions: List[str]):
        self.permissions = permissions

    def __call__(self, request: Request):
        request.state.permissions = self.permissions
        return self.permissions


app = FastAPI()


@app.get("/", dependencies=[Depends(Permissions(["Potato", "Tomato"]))])
def home(request: Request):
    return request.state.permissions
```

In fact, another path operation can be added with direct access to the permissions from the dependency:

```python
@app.get("/permissions")
def show_permissions(
    permissions: List[str] = Depends(Permissions(["Potato", "Tomato"]))
):
    return permissions
```

In this last example, the editor would be able to provide type checks and autocompletion inside of the function. :tada: 

---

Now, about the referenced issue #1879 , I understand that the feature request was to be able to access the data in the middleware, so, a middleware like this would have to work:

```python
@app.middleware("http")
async def add_permissions_headers(request: Request, call_next):
    params_str = ' - '.join(request.state.params)
    response: Response = await call_next(request)
    response.headers["x-params"] = params_str
    return response
```

But as the router is handled only after any middleware, this middleware breaks (with the current implementation or with this PR).

So, I imagine the intention with this PR would be an alternative implementation for the same use case, without using middlewares.

---

I would prefer not to add more ways to achieve something that can be easily achieved with dependencies (or at least with a comparable complexity), as it would create more maintenance burden and more complexity and options to learn in the framework.

Nevertheless, before closing this, I wanted to wait and see if anyone has any other use case that would be solved by this and that isn't easily solvable with the normal dependencies.

tiangolo:
Cool, thanks for the explanation!

Then, if someone later has a clear use case that couldn't be solved with the available tools and ideas (like dependencies), I would be happy to review this again (or an alternative).

But for now, I prefer to close this one, as there are already some things that I put there at some point "just in case" and ended up only adding confusion and cognitive load that wasn't necessary. And now they mainly mean work maintaining them, explaining them and explaining why they are there when another approach would be recommended, etc.

But thanks for the effort! :coffee: :cake:

PR #1919, opened 2020-08-18
✏️ Add descriptive error message when model that has no json encoders gets encoded

tomgrin10:
It took me hours of debugging to realize my error was not including `json_encoders` in the `Config` class of my Model.
Twice.

Changed:
* The error `TypeError: 'MyModel' object is not iterable` becomes `TypeError: 'MyModel' object has no json encoder`

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -118,7 +118,10 @@ def jsonable_encoder(
 
     errors: List[Exception] = []
     try:
-        data = dict(obj)
+        try:
+            data = dict(obj)
+        except TypeError:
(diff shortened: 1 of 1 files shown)


tomgrin10:
Hello? Anybody here? Is this gonna get merged??

tomgrin10:
@tiangolo Can this please be merged? This is a really minimal change.

tiangolo:
The problem is that `jsonable_encoder` is used in a lot more places that don't necessarily have to do with Pydantic models. So, there's no certainty that a `TypeError` is always for the same reason.

Given that, I'll pass on this, but thanks for your effort @tomgrin10 ! :cake:

Issue #1440, opened 2020-05-20
[BUG] Overwriting HTTPException and RequestValidationError does not work via FastAPI exceptions handlers init


timbmg:
When initilizing FastAPI, an exception_handlers dictionary can be passed, mapping Exceptions/Error Codes to callables. When using this method to set the exception handlers, the HTTPException and RequestValidationError can not be overwritten.

### Code to Reproducte

```Python

import logging

import uvicorn
from fastapi import FastAPI, Request
from fastapi.responses import JSONResponse
from fastapi.exceptions import RequestValidationError
from fastapi.exception_handlers import request_validation_exception_handler


logger = logging.getLogger(__name__)

# app = FastAPI()
# @app.exception_handler(RequestValidationError)
async def my_request_validation_exception_handler(
    req: Request, exc: RequestValidationError
):
    logger.info("my_request_validation_exception_handler")
    return await request_validation_exception_handler(req, exc)

app = FastAPI(exception_handlers={RequestValidationError: my_request_validation_exception_handler})

@app.get("/test")
def test(param: int):

    return JSONResponse(
        status_code=200,
        content={"message": "successful test with param {}".format(param)},
    )


if __name__ == "__main__":
    uvicorn.run("api:app", host="localhost", port=2000, reload=True)

```

Now a `curl "http://localhost:2000/test"` will not show the `logger.info("my_request_validation_exception_handler")` log statement. However, when using the commented out code it works fine. I want to use the other way to keep the handlers in a separate file.

### Environment

- OS: Windows/WSL
- FastAPI Version: 0.54.2
- Python version: 3.6.9

### Origin
As far as i can see this is coming from the [setup method](https://github.com/tiangolo/fastapi/blob/601d8eb8098591df7b2b95d83d319a6f8931402d/fastapi/applications.py#L137-L140)
```Python
self.add_exception_handler(HTTPException, http_exception_handler)
self.add_exception_handler(
    RequestValidationError, request_validation_exception_handler
)
``` 
Starlette will overwrite the custom handlers in the [add_exception_handler method](https://github.com/encode/starlette/blob/41ad780073b3a0a815e0083ecf1a0cea38cb8105/starlette/applications.py#L128-L134)
```Python
def add_exception_handler(
        self,
        exc_class_or_status_code: typing.Union[int, typing.Type[Exception]],
        handler: typing.Callable,
    ) -> None:
        self.exception_handlers[exc_class_or_status_code] = handler
[...]

PR #1887, opened 2020-08-12
Bug/fix exception handlers

mortezaPRK:
Closes #1886

PR #1924, opened 2020-08-19
🐛 Fix bug overriding custom HTTPException and RequestValidationError from exception_handlers

uriyyo:
Fix issue/related to #1440.

A fix is simple, set `HTTPException` and `RequestValidationError` exception handlers only in case when they don't set.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -66,6 +66,10 @@ class FastAPI(Starlette):
         self.exception_handlers = (
             {} if exception_handlers is None else dict(exception_handlers)
         )
+        self.exception_handlers.setdefault(HTTPException, http_exception_handler)
+        self.exception_handlers.setdefault(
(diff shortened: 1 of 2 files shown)


uriyyo:
This PR solve same issue as #1887 solve, I didn't see it because they point to two different issues related to the same problem.

Basically my idea was to reduce the number of calls to `Starlette.build_middleware_stack` method. 

Previously there can be 3 calls to this method because in  `FastAPI.setup` will call `FastAPI.add_exception_handler` method which will call `Starlette.build_middleware_stack`.

With this bugfix, there will be always only one call to the `Starlette.build_middleware_stack` method.

tiangolo:
Awesome! Great job @uriyyo ! :coffee: :cake:

Thanks for pointing to the previous PR. I ended up merging this one as the code was slightly simpler :nerd_face: 

This is included in FastAPI `0.61.2` (released in some minutes).

PR #1939, opened 2020-08-20
Add default parameter for Router path parameter

uriyyo:
Add default parameter to the Router `path` parameter.

Code like this:
```python
@router.get("", response_model=Item)
def route():
    ...
```
Can be rewritten to:
```python
@router.get(response_model=Item)
def route():
    ...
```

In case when call to `router` method doesn't have any arguments than a parenthesis can be skipped:
```python
@router.get
def route():
    ...
```

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -62,6 +62,7 @@ class FastAPI(Starlette):
             dependency_overrides_provider=self,
             on_startup=on_startup,
             on_shutdown=on_shutdown,
+            default_path="/",
         )
         self.exception_handlers = (
             {} if exception_handlers is None else dict(exception_handlers)
@@ -319,7 +320,7 @@ class FastAPI(Starlette):
 
     def get(
         self,
-        path: str,
+        path: Optional[str] = None,
         *,
         response_model: Optional[Type[Any]] = None,
         status_code: int = 200,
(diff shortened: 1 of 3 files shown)


ssw99:
Wouldn't it be better if you just give default value to path parameter of router?

uriyyo:
@ssw99 You mean to leave only the default parameter for path and remove the ability to use router functions as decorator without parameters?

ssw99:
What i mean is, rather than passing None as default value to router functions and checking if it is None, we can just pass empty string as default value.

uriyyo:
> What i mean is, rather than passing None as default value to router functions and checking if it is None, we can just pass empty string as default value.

It won't work with root `APIRouter`, because the root router should start with `/`. Correct me if I miss smth.

tiangolo:
Thanks for the interest @uriyyo ! :coffee: 

But I would actually prefer to leave it as is and have it more explicit. I feel it could end up being too magical and could be confusing.

Also, having support for:

```Python
@router.get("", response_model=Item)
def route():
    ...
```

is only to add support for those edge cases, but is actually discouraged. I would prefer not to add more magic to that.

So I'll pass on this one. But thanks for the effort! :cake:

PR #1991, opened 2020-08-31
handle list changed size during iteration in websockets broadcast

graingert:
if you get a connect or disconnect during broadcast

### DIFF
--- a/docs_src/websockets/tutorial003.py
+++ b/docs_src/websockets/tutorial003.py
@@ -66,6 +66,7 @@ class ConnectionManager:
             )
         )
 
+
 manager = ConnectionManager()
 
 
--- a/docs_src/websockets/tutorial003.py
+++ b/docs_src/websockets/tutorial003.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import List
 
 from fastapi import FastAPI, WebSocket, WebSocketDisconnect
@@ -58,9 +59,15 @@ class ConnectionManager:
         await websocket.send_text(message)
 
     async def broadcast(self, message: str):
-        for connection in self.active_connections.copy():
-            await connection.send_text(message)
-
+        return await (
+            asyncio.gather(
+                *(
+                    connection.send_text(message)
+                    for connection in self.active_connections
+                ),
+                return_exceptions=True
+            )
+        )
 
 manager = ConnectionManager()
 
--- a/docs_src/websockets/tutorial003.py
+++ b/docs_src/websockets/tutorial003.py
@@ -61,10 +61,7 @@ class ConnectionManager:
     async def broadcast(self, message: str):
         return await (
             asyncio.gather(
-                *(
-                    connection.send_text(message)
-                    for connection in self.active_connections
-                ),
+                *(conn.send_text(message) for conn in self.active_connections),
                 return_exceptions=True
             )
         )
--- a/docs_src/websockets/tutorial003.py
+++ b/docs_src/websockets/tutorial003.py
@@ -46,7 +46,7 @@ html = """
 
 _T = TypeVar("_T")
 
-@attr.s(frozen=True, slots=True, auto_attribs=True, hash=False, eq=False)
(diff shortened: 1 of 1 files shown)


docs_src/websockets/tutorial003.py
```diff
@@ -58,9 +59,12 @@ async def send_personal_message(self, message: str, websocket: WebSocket):
...
     async def broadcast(self, message: str):
-        for connection in self.active_connections:
-            await connection.send_text(message)
-
+        return await (
+            asyncio.gather(
+                *(conn.send_text(message) for conn in self.active_connections),
+                return_exceptions=True
+            )
+        )
```
graingert, line 67:
```suggestion
        )

```

docs_src/websockets/tutorial003.py
```diff
@@ -42,24 +44,39 @@
...
+        return isinstance(other, _IdEq) and self.v is other.v
+
+    def __hash__(self):
+        return hash(id(self.v))
+
 
 class ConnectionManager:
     def __init__(self):
-        self.active_connections: List[WebSocket] = []
+        self.active_connections: MutableSet[_IdEq(WebSocket)] = set()
```
Kludex, line 61:
```suggestion
        self.active_connections: MutableSet[_IdEq[WebSocket]] = set()
```

Kludex: I've suggested this, but you probably want to create a variable for it. Because on the add you want to use that new variable to something like: `_newIdEq(websocket)`

Kludex (commented):
Do you mind explaining what you're trying to achieve? :tada:

docs_src/websockets/tutorial003.py
graingert: Might be better to drop the __eq__ from HTTPConnecttion

docs_src/websockets/tutorial003.py
```diff
@@ -42,24 +44,39 @@
 </html>
 """
 
+_T = TypeVar("_T")
+
+@attr.s(frozen=True, slots=True, auto_attribs=True)
```
graingert, line 49:
```suggestion
@attr.s(frozen=True, slots=True, auto_attribs=True, hash=False, eq=False)
```

docs_src/websockets/tutorial003.py
```diff
@@ -42,24 +44,39 @@
 </html>
 """
 
+_T = TypeVar("_T")
+
+@attr.s(frozen=True, slots=True, auto_attribs=True, hash=False, eq=False)
```
graingert, line 49:
```suggestion
@attr.s(frozen=True, auto_attribs=True, hash=False, eq=False)
```

graingert:
> 
> 
> Do you mind explaining what you're trying to achieve? 🎉

if you get a disconnect/connect during broadcast you can skip over sockets in the broadcast queue:

```
>>> x = [1, 2, 3, 4, 5, 6, 7]
>>> y = iter(x)
>>> next(y)
1
>>> next(y)
2
>>> next(y)
3
>>> x.remove(1)
>>> next(y)
5
```

docs_src/websockets/tutorial003.py
graingert: @Kludex how did this pass the tests?

graingert:
Starlette has a fix for this in 0.16.0 and so this PR needs to change

tiangolo:
So, should anything else be done here @graingert or can we close it?

---

If you think there's something that still needs to be done here, I feel the code change and complexity is a bit daunting, I would try to find a way to make it as simple as possible, and easy to read for newcomers. And for example, I wouldn't use attrs but Pydantic, as everything is using Pydantic here, and FastAPI doesn't depend on attrs directly.

graingert:
The tutorial just needs to be updated to use a set instead

tiangolo:
As this won't be handled in this PR, I'm gonna go ahead and close it to clean up things. Thanks! :coffee:

Issue #2007, opened 2020-09-03
Casting dictionary in response_model_include/response_model_exclude to set.


Rubikoid:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to one of:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

Here's a self-contained, [minimal, reproducible, example](https://stackoverflow.com/help/minimal-reproducible-example) with my use case:

```Python
from fastapi import FastAPI
from pydantic import BaseModel

class Test(BaseModel):
    foo: str
    bar: str

class Test2(BaseModel):
    test: Test
    baz: str

app = FastAPI()
@app.get(
    "/",
    response_model=Test2,
    response_model_include={'baz': ..., 'test': {'foo'}},
    response_model_exclude={'test': {'bar'}}
    )
def item():
    return Test2(test=Test(foo="visible field", bar="invisible field"), baz="also visible field")
```

### Description

Dictionary, passed to response_mode_include/exclude casts to set.

* Open the browser and call the endpoint `/`.
* It returns a JSON with `{"baz":"also visible field"}`.
* But I expected it to return `{"baz":"also visible field", "test": {"foo": "visible field"}}`.

### Environment

* OS: Windows
* FastAPI Version: 0.60.1
* Python version: Python 3.7.3

### Additional context

So, this happens due to cast of include and exclude params to `set` at https://github.com/tiangolo/fastapi/blob/master/fastapi/encoders.py#L37 and https://github.com/tiangolo/fastapi/blob/master/fastapi/encoders.py#L39.

[...]

PR #2016, opened 2020-09-05
🐛 Fix include/exclude for dicts in `jsonable_encoder`

Rubikoid:
Related to #2007


Kludex (approved):

tiangolo:
Good catch, thank you @Rubikoid! :cake: :nerd_face: 

I updated the tests a bit to verify combined and nested use cases.

This is now available in FastAPI `0.66.0`. :rocket:

PR #2042, opened 2020-09-12
Fixed error 422 when creating Pydantic model from FormData

vobinics:
Fixed error 422 when creating Pydantic model from FormData. [Example temporal workaround on StackOverflow](https://stackoverflow.com/questions/60127234/fastapi-form-data-with-pydantic-model)

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -1,5 +1,6 @@
 import asyncio
 import inspect
+import json
 from contextlib import contextmanager
 from copy import deepcopy
 from typing import (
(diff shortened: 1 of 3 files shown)


ghandic:
@Vobinics I think the project makes use of `flit` so no need to add in the `setup.py`, you can check that out [here](https://github.com/tiangolo/fastapi/blob/master/pyproject.toml) and instructions on how to use it [here](https://fastapi.tiangolo.com/contributing/)

Vobinics:
@ghandic Yes, I know the Fast API uses Flit. setup.py I mistakenly added a branch to the master branch, but I need it in order to install the patched version for my project since I cannot wait for the developer to add my patch to the main branch and upload it to PyPi

Vobinics:
No need to add setup.py to the main Fast API repository, only my first commit needs to be added

PR #2047, opened 2020-09-14
Feature/pagination

BruceWW:
this is the feature that supports pagination, 
with the param with_page_split as True, the method will calculate the length of result, split the result by page_num and page_size params and finally return previous url and next_url
pyandtic.BaseModel, sqlalchemy object and normal type are supported


e-belair:
Hi did you plan to pull request this feature soon?

PR #2051, opened 2020-09-15
Fix tags's output format

hitrust:


jcwilson:
That's not valid JSON. Have you tested the tutorial changes in your other PRs that seem to be related to these documentation updates? (also, why aren't all of these PRs a single PR?)

hitrust:
> 
> 
> That's not valid JSON. Have you tested the tutorial changes in your other PRs that seem to be related to these documentation updates? (also, why aren't all of these PRs a single PR?)

Yep, you are right about the JSON format, sorry about that, just doing it in a hurry.
After doing some tests, the output is no need to be changed. It's good.
Just the beginning of learning FastApi and reading the documentation, so edit the other three PRs in github.com separately.

PR #2052, opened 2020-09-15
📝 Fix example code with sets in Tutorial - Body - Nested Models

hitrust:


tiangolo:
Good catch, thank you! :rocket: :cake:

PR #2054, opened 2020-09-15
📝 Fix example code with sets in Tutorial - Body - Nested Models 3

hitrust:


ycd:
Duplicate of #2052

tiangolo:
Good catch, thank you! :tada:

PR #2056, opened 2020-09-15
Feature/pagination

BruceWW:
this is the feature that supports pagination, 
with the param with_page_split as True, the method will calculate the length of result, split the result by page_num and page_size params and finally return previous url and next_url
pyandtic.BaseModel, sqlalchemy object and normal type are supported


ycd (commented):
I suggest you to use Typing's Dict and List for type-hinting since whole codebase based on Typing.

PR looks good overall :tada: :rocket:

BruceWW:
@ycd pertty agree with your suggestion, and it's been updated

Shark009:
Any idea when this will be merged and released?

Kludex:
@Shark009 There's no confirmation that it will be merged...

jrversteegh:
Despite its usefulness, this pagination option is not part of OpenAPI (as far as I am aware) and there are also competing options like page[number] and page[size] like used in json::api, so I don't think this should be in fastapi. Maybe in fastapi_utils or something?

BruceWW:
Rest api do not suggest use json body while the request method is GET, so i think  page[number] and page[size] is not a competitor. And fastapi_utils could be discussed

tiangolo:
I see you did a lot of work here @BruceWW ! :nerd_face: 

Nevertheless, that would be a very tight/strict compromise in how FastAPI expects to be used. And I would prefer to not have it that way. That's the same reason why it is not tightly coupled with any specific ORM.

For example, you could imagine someone wanting to use an opaque "cursor" object instead of pages. Or someone querying a NoSQL DB like CouchDB with an index range based on dates, etc.

So I'm gonna pass on this. Nevertheless, if this is a feature that you, your team, and/or others are frequently needing, I encourage you to create an independent Python package that can be used with FastAPI.

And if you add the `fastapi` topic to the GitHub repo it will show up here: https://fastapi.tiangolo.com/external-links/#projects

Given that, I'm gonna pass on this. But thanks for your effort! :cake: :coffee:

jrversteegh:
@BruceWW page[number] and page[size] are not body parameters, but query parameters in json::api. See e.g. this discussion if you're interested: https://github.com/OAI/OpenAPI-Specification/issues/1706

PR #2078, opened 2020-09-19
Add support for query parameter serialization styles

jakub-bacic:
It adds support for query parameter serialization styles (see: https://swagger.io/docs/specification/serialization/).

For more details, please follow reported enhancement issue https://github.com/tiangolo/fastapi/issues/2077.


adriangb:
Question: does this cover the use case of CSV-like quoted strings: `GET /items?foo="a,bc",d` (example taken from [here](https://specs.openstack.org/openstack/api-wg/guidelines/pagination_filter_sort.html#filtering)). A simple `.split` would result in `['"a', 'bc"', 'd']` which means no parameter values could have delimiters in them.

jakub-bacic:
@adriangb 
You're right - it won't handle such use case. I think the easiest way to handle it is to reuse built-in `csv` python module and replace:

```python
value = value.split(delimiter)
```
with
```python
value = list(csv.reader([value], delimiter=delimiter))[0]
```

The only problem I can see is that we can go a step further and ask: what about quotes in quotes? Actually, this is handled by `csv` module as well but there are different strategies and AFAIK, OpenAPI does not specify which one should be used:

```python
import csv
from typing import List, Optional

def parse_query_field(value: str, doublequote: bool = True, escapechar: Optional[str] = None) -> List[str]:
   return list(csv.reader([value], delimiter=',', doublequote=doublequote, escapechar=escapechar))[0]

EXPECTED = ['a', '"b,c,d"', 'e,f']

# default behaviour of csv module
assert(EXPECTED == parse_query_field('a,"""b,c,d""","e,f"', doublequote=True))

# example from the website you linked where escape chars are used
assert(EXPECTED == parse_query_field('a,"\\"b,c,d\\"","e,f"', doublequote=False, escapechar='\\'))
```

Ideally, it should be up to the FastAPI user to define it (more parameters?) but I'm not sure we want to go that way. Anyway, I rebased my changes on the top of current master and added simple fix with `csv` module with default values.

@tiangolo did you have a change to look at this?

maciej-trebacz:
@tiangolo Sorry for pinging, any chance this will be merged any time soon?

fastapi/openapi/utils.py
```diff
@@ -109,6 +109,10 @@ def get_openapi_operation_parameters(
         if field_info.deprecated:
             parameter["deprecated"] = field_info.deprecated
         parameters.append(parameter)
+        if isinstance(field_info, Query):
+            if not field_info.explode:
```
cikay, line 113: What is the point of nested if statement here?

cikay (changes requested):

fastapi/openapi/utils.py
jakub-bacic: I assume you're suggesting to merge these two conditions like this?
```python
if isinstance(field_info, Query) and not field_info.explode:
   parameter["explode"] = field_info.explode
   parameter["style"] = field_info.style.value
```

I can't see a reason why I chose nested version over this one. Maybe it was left after some refactoring. I can change that and update the branch as there are some conflicts with master.

cikay: > I assume you're suggesting to merge these two conditions like this?
Yes

fastapi/dependencies/utils.py
```diff
@@ -583,16 +591,24 @@ def request_params_to_args(
...
+            field_info, params.Param
+        ), "Params must be subclasses of Param"
+
         if is_scalar_sequence_field(field) and isinstance(
             received_params, (QueryParams, Headers)
         ):
-            value = received_params.getlist(field.alias) or field.default
+            if isinstance(field_info, params.Query) and not field_info.explode:
+                value = received_params.get(field.alias)
+                if value is not None:
```
cikay, line 604: You can also use [the walrus operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions) to merge this condition as well if the python version is compatible.

jakub-bacic: That's true but AFAIK FastAPI tries to be compatible with python 3.6+ so I guess walrus is out of the question.

tests/test_tutorial/test_query_params_str_validations/test_tutorial015.py
```diff
@@ -0,0 +1,112 @@
...
+            },
+        }
+    },
+}
+
+
+def test_openapi_schema():
+    response = client.get("/openapi.json")
+    assert response.status_code == 200, response.text
+    assert response.json() == openapi_schema
```
BilalAlpaslan, line 84: test failed in here and test_tutorial015_py39.py same function

nidhishgajjar:

## Summary
This PR adds support for OpenAPI query parameter serialization styles (`explode` and `style`) to FastAPI, enabling custom delimiters (comma, space, pipe) for array query parameters when `explode=False`.

## Architecture
The implementation cleanly follows the OpenAPI 3.0 specification by:
- Adding `QueryStyle` enum with form, space_delimited, and pipe_delimited options
- Extending `Query()` with `style` and `explode` parameters (defaulting to backward-compatible values)
- Parsing delimited values using Python's `csv.reader` when `explode=False`
- Correctly emitting `explode` and `style` in the generated OpenAPI schema
- Adding comprehensive tests covering basic parsing, quoted values, and escaped quotes

## Issues

**fastapi/dependencies/utils.py:139 — [suggestion] — Empty string edge case**

```python
value = list(csv.reader([value], delimiter=delimiter))[0]
```

When an empty string is passed (`?q=`), this produces `[]` (empty list) instead of `['']` (list with one empty string). While this might be intentional (no values = empty list), it's worth documenting this behavior since users might expect the empty string to be preserved.

**fastapi/dependencies/utils.py:139 — [suggestion] — Consider adding error handling**

The CSV parsing is wrapped directly without try-except. If malformed input causes a CSV parsing error (e.g., unmatched quotes in edge cases), the exception will propagate. Consider wrapping in a try-except to provide a clearer validation error message.

**docs/en/docs/tutorial/query-params-str-validations.md:33 — [suggestion] — Document CSV escaping rules**

The documentation shows basic examples but doesn't explain how to include the delimiter character in values (e.g., to pass a literal pipe character in a pipe_delimited field). Consider adding a note about CSV-style quoting: use double quotes to escape delimiters and double-double-quotes for literal quotes.

## Cross-file Impact

**Public API changes:**
- `fastapi/__init__.py`: Exports new `QueryStyle` enum
- `fastapi/param_functions.py`: `Query()` now accepts `style` and `explode` parameters with backward-compatible defaults
- No breaking changes (default values maintain existing behavior)

**Tests:**
- Two new test files covering both Python 3.6+ and 3.9+ variants
- Tests cover: OpenAPI schema generation, default values, basic parsing, quoted values, and escaped quotes

**Dependencies:**
- Uses Python's built-in `csv` module (standard library)

## Assessment
✅ Approve

This is a well-implemented feature that correctly follows the OpenAPI specification with proper backward compatibility, comprehensive tests, and clean code structure. The minor suggestions above are for documentation and robustness improvements but are not blockers.

nidhishgajjar:

## Summary
This PR adds support for query parameter serialization styles (`form`, `space_delimited`, `pipe_delimited`) in FastAPI, aligning with the OpenAPI 3.0 specification. It allows developers to customize how list-type query parameters are parsed from URL query strings using the `style` and `explode` parameters.

## Architecture
The implementation is well-architected and follows FastAPI's existing patterns:

1. **New types**: Introduces `QueryStyle` enum in `fastapi/params.py` to represent the supported serialization styles
2. **Extended parameters**: Adds `style` and `explode` to both `Query` class and `Query()` function with sensible defaults (form, True) for backward compatibility
3. **Parsing logic**: Updates `request_params_to_args()` in `fastapi/dependencies/utils.py` to handle non-exploded query parameters by parsing them with appropriate delimiters using Python's `csv.reader` module
4. **OpenAPI integration**: Modifies `get_openapi_operation_parameters()` to include `explode` and `style` in the generated OpenAPI schema
5. **Public API**: Exports `QueryStyle` from `fastapi/__init__.py` for public use

The use of `csv.reader` for parsing is a clever choice—it properly handles quoted values, allowing delimiters within quoted strings to be treated as literal characters.

## Issues

**fastapi/dependencies/utils.py:160** — [suggestion] — Potential edge case with unknown query styles

The code uses `query_style_to_delimiter.get(field_info.style)` which could return `None` if an unknown style is somehow passed. While the type system (`QueryStyle` enum) makes this unlikely in practice, it might be safer to add a fallback:

```python
delimiter = query_style_to_delimiter.get(field_info.style)
if delimiter is None:
    # Fallback to default form style for safety
    delimiter = ","
value = list(csv.reader([value], delimiter=delimiter))[0]
```

However, this is a minor concern given that `QueryStyle` is an enum and cannot have arbitrary values.

**fastapi/dependencies/utils.py:159-161** — [suggestion] — Empty string handling

When parsing an empty string with `csv.reader`, the result is a list containing one empty string `['']` rather than an empty list. This might be unexpected behavior for users. Consider documenting this or adding a check:

```python
if value is not None:
    delimiter = query_style_to_delimiter.get(field_info.style)
    parsed = list(csv.reader([value], delimiter=delimiter))[0]
    # Handle empty string case
    value = parsed if parsed != [''] else []
```

However, the current behavior might be intentional—treating an empty string as a list with one empty element could be the desired behavior in some contexts.

## Cross-file Impact

- **Public API changes**: `QueryStyle` is now exported from the main `fastapi` module
- **Breaking changes**: None—all new parameters have defaults that maintain existing behavior
- **Backward compatibility**: Excellent—default values (`style=QueryStyle.form`, `explode=True`) preserve the existing comma-separated list behavior
- **OpenAPI schema**: Generated schemas will now include `explode` and `style` for query parameters when explicitly set
- **Tests**: Comprehensive test coverage added for both Python 3.6+ and 3.9+ variants, including edge cases with quoted delimiters

## Assessment
✅ Approve

This is a well-implemented feature that adds valuable OpenAPI 3.0 compliance to FastAPI. The code is clean, follows existing patterns, includes comprehensive tests, and maintains full backward compatibility. The suggested improvements are minor and optional.

tiangolo:

Thank you! 🍰

PR #2083, opened 2020-09-21
Add option for custom validator for Path params.

aviramha:
Nowadays, the only supported validation for Path params are the given ones (le, ge, etc).
I think there are many use cases where we want to sanitize and validate input of Path parameters that is "smarter" than the given options.
I will add documentation once I get approval of concept :)

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -30,6 +30,7 @@ from fastapi.security.oauth2 import OAuth2, SecurityScopes
 from fastapi.security.open_id_connect_url import OpenIdConnect
 from fastapi.utils import create_response_field, get_path_param_names
 from pydantic import BaseModel, create_model
+from pydantic.class_validators import Validator
 from pydantic.error_wrappers import ErrorWrapper
 from pydantic.errors import MissingError
 from pydantic.fields import (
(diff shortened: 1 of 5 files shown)


aviramha:
@tiangolo I hope things are going well, hopefully you can check this out :)

M-e-r-c-u-r-y:
@aviramha Any chance you can add this feature for query parameters too, Currently [query parameters](https://github.com/tiangolo/fastapi/blob/f495d98fae187beeacd9d0fcaf75838402a4b6fa/fastapi/param_functions.py#L39) have similar validation options as path parameters and they should be similar to setup in Pydantic I think.

@tiangolo Please consider this merge request, having custom validators would be a nice feature when one is doing more complex validations.

aviramha:
@M-e-r-c-u-r-y Yes I can add but I'd like initial approval of current implementation before adding it to any more places :)

tiangolo:
Thanks @aviramha! :nerd_face: 

I think this would make sense at the Pydantic level, `Field()` could take a validator directly, right?

Then that functionality would only be inherited here instead of extended.

That would probably need to wait for Pydantic v2 or be part of it, as I guess no new PRs will be accepted for v1.

What do you think?

> Sorry for the long delay! 🙈  I wanted to personally address each issue and they piled up through time, but now I'm checking each one in order.

aviramha:
Hi @tiangolo 
I have pretty much lost the context and can't really answer correctly as I'm not following nor using FastAPI in the last few months (just not using Python, still love FastAPI :))

tiangolo:
Thanks for coming back! And yep, makes sense.

I'll close this one then. Thanks! 🍰

PR #2085, opened 2020-09-23
📝 Simplify intro to Python Types, all currently supported Python versions include type hints 🎉

ninjaaron:
Function annotation syntax was proposed in [pep 3107](https://www.python.org/dev/peps/pep-3107/) in 2006 and was a feature of Python 3.0. Type annotations were already a proposed usecase at this time. Mypy supports Python 3.2 and newer. A standard vocabulary for type hints (and the typing module itself) based on conventions used in Mypy and other sources was added in 3.5 with [pep 484](https://www.python.org/dev/peps/pep-0484/).

The only thing new for type hints in 3.6 is _variable annotations_, which allows types to be specified outside of function signatures. [pep 526](https://www.python.org/dev/peps/pep-0526/)

### DIFF
--- a/docs/en/docs/python-types.md
+++ b/docs/en/docs/python-types.md
@@ -1,8 +1,8 @@
 # Python Types Intro
 
-**Python 3.6+** has support for optional "type hints".
+**Python 3.5+** has support for optional "type hints".
 
-These **"type hints"** are a new syntax (since Python 3.6+) that allow declaring the <abbr title="for example: str, int, float, bool">type</abbr> of a variable.
+These **"type hints"** are a new syntax (since Python 3.0+) that allow declaring the <abbr title="for example: str, int, float, bool">type</abbr> of a variable.
 
 By declaring types for your variables, editors and tools can give you better support.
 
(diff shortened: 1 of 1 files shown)


ArcLightSlavik:
I think this proposal is *technically* correct but I find some drawbacks.

1. Typing module was provisional in 3.5. It become stable in either 3.6.4 or 3.7.
2. Pydantic uses 3.6 in it's documentation and it has way more weight in on types.
3. Some people might read it as 'fastapi can run on python 3.5' which might be possible but is not suggested at all.

ninjaaron:
Is it possible to rephrase this in such a way that it is both technically correct and doesn't give people the wrong idea? (a possible option would be to omit details about when type annotations were introduced altogether)

ArcLightSlavik:
Considering 3.5 reached end of life this month, completely removing the version sounds the best.

ArcLightSlavik (approved):

Kludex (approved):

ycd (approved):

tiangolo:
Awesome, thanks! It's so nice that type hints are now "mainstream" and standard in all the Python versions supported! :tada: 

Thanks for your contribution! :rocket: :tada: 

And thanks @ArcLightSlavik for those points :heavy_check_mark: :mag:

PR #2089, opened 2020-09-24
📌 Relax Swagger UI version pin

jmriebold:
Removes Swagger UI pin, as the [Swagger UI bug](https://github.com/swagger-api/swagger-ui/issues/6249) was fixed July 22.


Kludex:
Maybe we could do something similar to what dependabot does here?

tiangolo:
Great, thank you! :rocket: :coffee: 

This is available in FastAPI version `0.61.2` :rocket:

PR #2098, opened 2020-09-25
Update tutorial003.py, fixed bad import in websockets docs

polamayster:
Fixed ImportError: cannot import name 'WebSocketDisconnect' from 'fastapi' in websockets example

### DIFF
--- a/docs_src/websockets/tutorial003.py
+++ b/docs_src/websockets/tutorial003.py
@@ -1,8 +1,8 @@
 from typing import List
 
 from fastapi import FastAPI, WebSocket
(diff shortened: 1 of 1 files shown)


tiangolo:
Hey, maybe you have an older FastAPI version?

The original version:

```Python
from fastapi import FastAPI, WebSocket, WebSocketDisconnect
```

should work correctly as it is exported here:

https://github.com/tiangolo/fastapi/blob/master/fastapi/__init__.py#L25

I'm gonna close this, but if you keep having problems please create an issue.

Thanks for your effort! :cake: :coffee:

PR #2114, opened 2020-09-30
Add Stale bot configuration

### LINKED ISSUE #2049: Add a Stale bot for the issues
There is a lot of issues ( almost 300 ) and a lot of them are just question of how use fastAPI and not real current bugs. 

I like to use the issues on github to know the current bugs on the latest version of a project. But here is complicated to have an overview of the "situation".

We could add a github bot [Stale](https://github.com/marketplace/stale) to manage and close old very inactive issues.

What do you think about it ? :)

bene64:
The configuration is basically the same like the example config of https://probot.github.io/apps/stale/.
Differences:
- 90 instead of 60 days: for the start of the bot we should allow issues to be stale a little bit longer. This can be decrease in the future.
- exemptLabels:
  - pinned and security: are not part of the project, but could be useful in the future.
  - help-wanted: is probably something which the authors are willing to work on.
- markComment: make it more clear what "recent" means

Fixes #2049.
Additionally stale bot has to be installed into the repository. This can only be done by the maintainers.

Sorry if stale bot is not something which is wanted for this project. I probably should've checked with the maintainers before.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - help-wanted
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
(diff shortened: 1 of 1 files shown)


raphaelauv (approved):

tiangolo:
Thanks for your contribution @bene64 ! :nerd_face: 

There's actually a similar bot/GitHub Action already in place, with slightly different behavior that I prefer: https://github.com/tiangolo/issue-manager#motivation

So I'm gonna pass on this, but thanks for your effort! :cake:

PR #2118, opened 2020-10-02
🐛 Check Content-Type request header before assuming JSON

### LINKED ISSUE #1018: [BUG] Body is always parsed as JSON regardless of media_type
### Describe the bug

This is to continue the discussion at https://github.com/tiangolo/fastapi/issues/579#issuecomment-589263249. `Body` accepts an argument `media_type`, but regardless of the media_type Body is parsed as JSON. This is a problem for other media_types such as `plain/text`, `application/sql`, etc. And this is not compliant to [OpenAPI content](https://swagger.io/docs/specification/describing-responses/). Body should be able to represent plain text as `str` too.

Basically I wish I could do this:
```python
request: str = Body(..., media_type='text/plain')
```
However, this produces 400 parsing error because Body is parsed as JSON anyway.

### To Reproduce

1. Create a file with:

```Python

app = FastAPI()

@app.post("/")
def read_root(body: str = Body(..., media_type='text/plain')):
    return {"I received:": body}
```

2. Run the server.

3. Then execute: `curl -d 'I want to print this' -H 'Content-Type: plain/text' http://localhost:8000` 

4. It returns {"detail":"There was an error parsing the body"}

5. But I expected it to return `{"I received": "I want to print this"}`.

### Expected behavior

My expectation is that the Body should be parsed based on the Content-Type header always. 

### Screenshots

I think the information is quite comprehensive without screenshots.

### Environment

- OS: Ubuntu 18.04
- FastAPI Version: 0.49.0
- Python 3.7.4

### Additional context
The associated part is: https://github.com/tiangolo/fastapi/blob/9c3c9b6e78768374868d690bc05918d58481e880/fastapi/routing.py#L114

This does not check Content-Type. My proposal is to fix it as 
```python
if body_bytes and request.headers['Content-Type'] == 'application/json':
# if body_bytes and request.headers.get('Content-Type', 'application/json'): # if the content type should be assumed.
    body = await request.json()
else:
    body = body_bytes
```

I can make a PR for this, but I would like to learn the maintainers' opinion on this.
@tiangolo @dmontagu @phy25, thanks for the great project!

patrickkwang:
🐛 Check Content-Type request header before assuming JSON

Related to #1018.

---

**Edit** by @tiangolo: Updated tests to include extra corner cases, and updated implementation to parse content-type headers using Python's standard library `email` module to cover more corner cases.


fastapi/routing.py
```diff
@@ -155,16 +154,20 @@ def get_request_handler(
...
                 if is_body_form:
                     body = await request.form()
                 else:
                     body_bytes = await request.body()
                     if body_bytes:
-                        body = await request.json()
-        except json.JSONDecodeError as e:
-            raise RequestValidationError([ErrorWrapper(e, ("body", e.pos))], body=e.doc)
+                        if (
+                            request.headers.get("Content-Type", "application/json")
```
ArcLightSlavik, line 165: Is Content-Type not being there a concern? I believe RFC says it should always be filled.

tests/test_tutorial/test_body/test_tutorial001.py
```diff
@@ -178,16 +178,9 @@ def test_post_broken_body():
...
-                    "doc": "name=Foo&price=50.5",
-                    "lineno": 1,
-                    "msg": "Expecting value",
-                    "pos": 0,
-                },
-                "loc": ["body", 0],
-                "msg": "Expecting value: line 1 column 1 (char 0)",
-                "type": "value_error.jsondecode",
+                "loc": ["body"],
+                "msg": "value is not a valid dict",
```
ArcLightSlavik, line 182: Would be considered a breaking change. Think even my work API has a test similar to this.

fastapi/routing.py
patrickkwang: [RFC 7231](https://tools.ietf.org/html/rfc7231#section-3.1.1.5) suggests it is not strictly required, and that we should assume "application/octet-stream" if it's not present.

ArcLightSlavik (approved):
Find the formatting a bit odd but if that's how flake8(?) did it then okay.

Mause:
I think this needs to be thought about a bit more - there are other mime types that are JSON but which don't use the application/json minetype. application/hal+json is one such common example of this

patrickkwang:
Good point, @Mause. Based on [this list](https://www.iana.org/assignments/media-types/media-types.xhtml) of registered mimetypes, it seems safe to interpret as JSON any mimetype that _ends_ with "json". Notably, this excludes the "json-seq" type, which is not itself a subset of JSON.

Mause:
Another thing - you're also breaking content-type headers with charset or boundary values, such as `application/json; charset=UTF-8`

patrickkwang:
Perhaps some new tests are called-for at this point.

patrickkwang:
How do we like conflicts to be handled? I can rebase and force-push to my feature branch...

rdpravin1895:
Hi @patrickkwang. Are you planning to merge this?

patrickkwang:
I've rebased and everything is green, but I do not have the necessary permissions to merge.

rdpravin1895:
Hi @tiangolo, sorry to disturb you, but can you have a look at this? We've been using FastAPI, so far has been an excellent framework, but only because of strict json application type, we are facing troubles in documenting the examples.

tiangolo:
Thanks @patrickkwang ! :rocket: :nerd_face: 

And thanks for the thorough reviews @ArcLightSlavik and @Mause :bow: :cake: 

I updated the tests to include extra corner cases and updated the implementation to parse the content-type headers using Python's standard library `email` module to cover more corner cases.

jolynch:
For what it's worth this appears to be a pretty big breaking change for clients who were previously not sending `Content-Type` headers. For example, some of our production clients started receiving a validation error from a route where previously it was working fine:
```
{"message":"Error processing request","error":"RequestValidationError","details":[{"loc":["body"],"msg":"value is not a valid dict","type":"type_error.dict"}]}
```

I understand the original default was incorrect but this fix breaks functional code and I believe the CVE would be resolved by assuming json in any case that is not exempted from CORS preflight (so don't assume json when the content-type is explicitly set to something not `application/json`)? Alternatively could the server author configure the default assumed content type?

kpflugshaupt:
This just cost me two days of debugging. Not so pleased right now.

I fully see and appreciate the reasoning behind stricter checks, mind you. Go ahead with that!
But the error message is rather misleading.  _'value is not a valid dict'_ does **not** make me suspect my HTTP headers!

Is it possible to implement an exception that's more to the point here? I have the feeling I'm not the only one tearing their hair out over this...

Sorry for the venting -- FastAPI is and remains my favourite framework. Just feeling a bit betrayed by that error message?

hademircii:
> For what it's worth this appears to be a pretty big breaking change for clients who were previously not sending Content-Type headers. 
experiencing a similar problem right now (content-type is plain text). Bumping Fast API to 0.65.2 to breaks their integration, and we are not able to update all clients right away. So we either need to hold off on this and future FastAPI updates, or find a server side fix to go around it.

tiangolo:
That's a very good point @jolynch! Thanks for idea.

I just released version `0.65.3` that does exactly that. If there's no Content-Type header, it is now assumed as JSON.

This still provides protection from CSRF because browsers automatically set Content-Type headers. And if an attacker removes the header explicitly, it triggers the CORS preflight, which provides the main protection using CORS.

So, you should all be able to upgrade to `0.65.3` more easily now. :tada:

Issue #2026, opened 2020-09-08
HTTPBearer security scheme is returning 403 instead or 401


aaaaahaaaaa:
HTTPBearer security scheme enabled as a dependency is returning a `403` when a request is unauthenticated because of a missing or a malformed `authorization` header. In those scenarios, a `401` should be returned instead.

ArcLightSlavik:
Could you provide an example of how you are getting that?
Docs say:
```
If it doesn't see an Authorization header, or the value doesn't have a Bearer token, it will respond with a 401 status code error (UNAUTHORIZED) directly.
```

raphaelauv:
Yes if the authorization is missing then it's a 403 https://github.com/tiangolo/fastapi/blob/e77ea635777d2494690ba3eb62bd005b9edeefde/fastapi/security/api_key.py#L27

aaaaahaaaaa:
@raphaelauv @ArcLightSlavik 
I believe the status code is coming from here: https://github.com/tiangolo/fastapi/blob/55b9faeb48f9c8676cd56adc8f8d75040e3f1010/fastapi/security/http.py#L114

My setup is the following:
```
...
from fastapi.security import HTTPBearer

security = HTTPBearer()

@app.get("/test")
async def test(bearer_token: str = Depends(security)):
    ...
```

Then if you don't provide a bearer token via the `authorization` header, this produces a 403 by default.

aaaaahaaaaa:
To be clear here: I only want to support Authorization header to do the JWT verification. I don't want to support the full OAuth2 flows which is why I'm not using the `OAuth2PasswordBearer` scheme.

raphaelauv:
@aaaaahaaaaa You could open a PR to correct this

OYE93:
> For API_KEY when the authorization is missing then it's a `403`
> 
> https://github.com/tiangolo/fastapi/blob/e77ea635777d2494690ba3eb62bd005b9edeefde/fastapi/security/api_key.py#L27
> 
> but for bearer it's a `401` if it's missing
> 
> https://github.com/tiangolo/fastapi/blob/e77ea635777d2494690ba3eb62bd005b9edeefde/fastapi/security/oauth2.py#L158

I'd like to know why for API_KEY, the error should be `403`, but `401` for bearer, thanks

dorinclisu:
It makes total sense to have 401 returned, I'm sure tiangolo did not mean 403 and it was just a small mishap

iantimmis:
[...]

PR #2120, opened 2020-10-02
🐛 Fix exception status codes in security http schemes

gostekk:
Hi!

**_TL;DR:_**
This PR changes exception response status codes in security http schemes from 403_FORBIDDEN to 401_UNAUTHORIZED.

**__Long__**
HTTPBearer and HTTPDigest security schemes used as a dependency are returning a 403_FORBIDDEN status on scheme Exception. This should be 401_UNAUTHORIZED status because we can't identify user without token or with invalid credentials.

There's issue mentioning this: #2026


aksel:
I was just about to open a similar PR myself, and noticed yours here :smile: 

In general, the status code for invalid authorization seems to depend on the application. The use case I am building would require 401 for a missing (or malformed, i.e. not `/Bearer [a-z0-9]+/i`)  Authorization header.

But I have seen other status codes being used.

**400**: If header was malformed. In my opinion, this is bad form. But I've seen it, to make client-side error handling easier; any request parameter - be it header, body, query, whatever - must match _this format_, and if not, 400 is sent. 400 is then handled client side as something like `Your request did not match the specification blablabla`

**403**: For example if an auth header was provided, but malformed, you may wish to respond _as if_ it was valid auth, but insufficient scope. This also matches the [RFC](https://tools.ietf.org/html/rfc7231#section-6.5.3), in my opinion:

> If authentication credentials were provided in the request, the server considers them insufficient to grant access.  The client SHOULD NOT automatically repeat the request with the same credentials.  The client MAY repeat the request with new or different credentials.  However, a request might be forbidden for reasons unrelated to the credentials.

**404**: E.g. if you wish to obfuscate the endpoint.

This was a long winded comment, in order to justify the following: I think the constructor should accept a parameter `status_code`, which defaults to 401.

zolaemil:
I'd like to add my support for this PR.

In response to @aksel:
>403: For example if an auth header was provided, but malformed, you may wish to respond as if it was valid auth, but insufficient scope. This also matches the RFC, in my opinion

It is true that the quoted definition of 403 fits the scenario you describe. However, the definition of 401 in the same [RFC](https://tools.ietf.org/html/rfc7235#section-3.1) fits it _better_:
> The 401 (Unauthorized) status code indicates that the request has not
   been applied because it lacks valid authentication credentials for
   the target resource.

aksel:
> However, the definition of 401 in the same [RFC](https://tools.ietf.org/html/rfc7235#section-3.1) fits it _better_

I agree.

smacpher-myst:
@gostekk Thanks for putting this change together. Are there plans to merge / release this soon?

adriangb:
Bumping this. Would be great to get this merged.

ericrdgz:
Would love to see this fixed. Any update on this?

dreinon:
Hi @tiangolo! Sorry to tag you but just doing it in case you missed this one, since this PR is supported by a few people already.
Thanks!

BilalAlpaslan (approved):

HansBambel:
Would be nice to have this fixed

chrysaor:
Recently, I found this problem. I think it's a good point to separate error codes. When it will be merged to a master branch?

alexmaurizio:
Hello! Pinging in to this PR. We've just had this discussion with colleagues and we agree that the 403 when the auth is missing is not a correct default. 

We agree all applications have peculiar standards and needs and it's okay to return different codes, but this is about default values, not custom behaviour.

We all agree 401 is a better fit in the unauth case instead of a 403.

arthurio (approved):
I agree that it should be `401` instead of `403` as well.
Side question, @tiangolo is there a reason why we are asserting against `401`, `403`, etc. in the tests instead of `HTTP_401_UNAUTHORIZED`, `HTTP_403_FORBIDDEN`?

yezz123 (approved):

acutaia (approved):

tiangolo:
Hey there! Yep, this makes sense and I've been wanting to fix this, actually, since the first moment I was writing it for the first time while reading all the standards and specs. :sweat_smile: 

But here's the thing:

> The server generating a 401 response MUST send a WWW-Authenticate header field containing at least one challenge applicable to the target resource.

Ref: https://www.rfc-editor.org/rfc/rfc7235#section-3.1

I haven't been able to put the time to figure out what are the correct challenges for each security scheme, with the right spec references, etc. ...only for some of them.

That's why I haven't updated the rest, because 403 is not as precise as 401, but it's *not wrong*. While a 401 without `WWW-Authenticate` header would be, by the spec, wrong.

If you (or anyone else) can find the right references to the related specs that define the specific challenges that should be used for each one of those security schemes, I would love to have this in. It could also be independent PRs, assuming you could find only references for one of the schemes but not the others, I could take that in a single PR.

dhilazo (approved):

danclaytondev:
This is now [a discussion](https://github.com/tiangolo/fastapi/discussions/9130) but I think shoud remain an open issue.

creyD (approved):
lgtm!

ordinary-jamie:
I think RFC 6750 "The OAuth 2.0 Authorization Framework: Bearer Token Usage", specifically section 3 "The WWW-Authenticate Response Header Field" enumerates the required `WWW-Authenticate` challenge for the bearer auth-scheme.

In summary:

- **MUST** use the auth-scheme value "Bearer".
- **MUST** have one or more auth-param values following auth-scheme
- **SHOULD** include the `error` attribute to provide the client with the reason why the access request was declined.
- **MAY** include the `error_description` attribute to provide developers a human-readable explanation that is not meant to be displayed to end-users.
- **MAY** include the `error_uri` attribute with an absolute URI identifying a human-readable web page explaining the error.
- **MAY** use `realm` attribute to indicate the scope of protection in the manner described in HTTP/1.1 [[RFC2617](https://datatracker.ietf.org/doc/html/rfc2617)].
- **MAY** use `scope` field, a space-delimited list of case-sensitive scope values indicating the required scope of the access token for accessing the requested resource.

An example would be

```
     HTTP/1.1 401 Unauthorized
     WWW-Authenticate: Bearer realm="example",
                       error="invalid_token",
                       error_description="The access token expired"
```

Ref: https://datatracker.ietf.org/doc/html/rfc6750#section-3

Kludex (approved):

pnambiar-signos (approved):

svlandeg:
Thanks for the contribution @gostekk and everyone for the discussion 🙏. I'm closing this one in favour of https://github.com/fastapi/fastapi/pull/13786, which is more extensive and should (hopefully) address this issue once and for all.

aksel:
🫡

Haven't used FastAPI in three years. I can't believe the nostalgia I am feeling, seeing a little quirk from five years ago being addressed haha
Can't imagine what it must be like to have worked in open-source since the '90s! Like *can you imagine* seeing a 30 year-old issue being brought back up!

PR #2135, opened 2020-10-05
📝 Fix renamed parameter `content_type` typo

TeoZosa:
The `content_type` parameter was renamed to `response_class` in release `0.19.0`

### Tangential question to maintainers:

This is the 3rd documentation-related PR I've opened this week. They all fall into different categories, but all have fairly small changes. 
- Is this disruptive? 
- Would it be better to condense any near-future PRs into a single "documentation" PR? 

Thanks in advance!

### DIFF
--- a/docs/en/docs/advanced/custom-response.md
+++ b/docs/en/docs/advanced/custom-response.md
@@ -40,7 +40,7 @@ Import the `Response` class (sub-class) you want to use and declare it in the *p
 To return a response with HTML directly from **FastAPI**, use `HTMLResponse`.
 
 * Import `HTMLResponse`.
-* Pass `HTMLResponse` as the parameter `content_type` of your *path operation*.
+* Pass `HTMLResponse` as the parameter `response_class` of your *path operation*.
 
 ```Python hl_lines="2  7"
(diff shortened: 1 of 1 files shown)


tiangolo:
Awesome! Thank you @TeoZosa ! :mag: :coffee: 

About the question, it's perfect like this, in different PRs. That way if for example, I didn't want to accept one of them but I wanted the other 3, I could just merge the other 3.

But if it was all in a single PR, I wouldn't be able to merge all the things from the other 3, waiting for changes to that little thing I didn't want.

Thanks for your contribution! :cake: :tada:

TeoZosa:
> Awesome! Thank you @TeoZosa ! 🔍 ☕
> 
> About the question, it's perfect like this, in different PRs. That way if for example, I didn't want to accept one of them but I wanted the other 3, I could just merge the other 3.
> 
> But if it was all in a single PR, I wouldn't be able to merge all the things from the other 3, waiting for changes to that little thing I didn't want.
> 
> Thanks for your contribution! 🍰 🎉

Perfect! Thanks for the feedback!

PR #2158, opened 2020-10-10
📝 Fix code consistency in examples for Tutorial - User Guide - Path Parameters

nukopy:
I found typos in the chapter "Path Parameters" in "Tutorial - User Guide".

## What

Endpoint's path `model` in [example code](https://fastapi.tiangolo.com/tutorial/path-params/#create-an-enum-class) should be a plural form, `models`.

I fixed below:

- `docs_src/path_params/tutoial005.py`
- image associated with this example: `docs/en/docs/img/tutorial/image03.png`


Kludex:
The pipeline is failing because the tests are calling `/model` instead of `/models`.

nukopy:
Thanks. I fixed tests.

tiangolo:
Nice! Good catch, thank you! :rocket: :tada: 

Thanks @Kludex for the help! :muscle: :cake:

PR #2187, opened 2020-10-17
✨ Set response model defaults at application and router levels

frjonsen:
This would allow for setting the default value of `response_model_by_alias`, `response_model_exclude_unset`, `response_model_exclude_defaults` and `response_model_exclude_none` at an application or at a router level, rather than on each individual endpoint


philipbjorge:
@frjonsen --
This would be useful for our team as we're having to set this on every API route.

What needs to happen to get this PR reviewed/considered for main?

Thanks for all your hard work 🙇 !

fastapi/routing.py
```diff
@@ -563,8 +596,25 @@ def include_router(
                     )
         if responses is None:
             responses = {}
+
         for route in router.routes:
             if isinstance(route, APIRoute):
```
cikay, line 601: Reverse the condition and add the continue statement to the if statement to decrease the indentation level.

cikay (changes requested):

dkatzan:
@cikay @frjonsen 
will be happy to see this PR merged, as my team as well need to set these flags in every API route
looks like this PR is also a bit outdated and has conflicts

will be happy to pick it up from here and open a new PR

frjonsen:
> will be happy to pick it up from here and open a new PR

I have since moved on to different projects that don't use FastAPI, so feel free finish this yourself, if you want.

tiangolo:

Thank you! 🍰

PR #2242, opened 2020-10-25
⚡️ Build docs for languages in parallel in subprocesses to speed up CI

tiangolo:
⚡️ Build docs for languages in parallel in subprocesses to speed up CI


Kludex:
cool

PR #2244, opened 2020-10-25
✨ Add instant docs deploy previews for PRs from forks

tiangolo:
✨ Add instant docs deploy previews for PRs from forks.

This uses the new GitHub Actions event [`workflow_run`](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_run).

The GitHub Action event is currently not very well documented, and it ended up requiring a new custom action to extract the original PR for that docs deployment. But this should help with previews for changes in docs (e.g. translations). Instead of the current workaround of previews triggered every hour, this should have a preview in a comment ready in about 1 minute.

### DIFF
--- a/.github/actions/comment-docs-preview-in-pr/Dockerfile
+++ b/.github/actions/comment-docs-preview-in-pr/Dockerfile
@@ -1,0 +1,7 @@
+FROM python:3.7
+
+RUN pip install httpx "pydantic==1.5.1" pygithub
+
+COPY ./app /app
+
+CMD ["python", "/app/main.py"]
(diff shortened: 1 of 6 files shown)


adriangb:
Hey @tiangolo, I was able to get this to work using pure workflows ([here](https://github.com/adriangb/scikeras/blob/master/.github/workflows/deploy_docs.yml) and [here](https://github.com/adriangb/scikeras/blob/master/.github/workflows/docs.yaml)) and some existing actions. Not that this is any better than what you've got, just sharing options 😄 

The main thing that helped in figuring out the `workflow_run` event (which as you say, isn't very well documented) is that you can do `echo ${{ toJSON(github.event.workflow_run) }}` to get all of the key/values in there.

tiangolo:
Thanks @adriangb! Very interesting! 🤓 

PR #2249, opened 2020-10-26
♻️ Update black to 20.8b1 and apply new ruleset to codebase.

a-recknagel:
Just a quick update on black, and a more permissive update scheme (`>=` over `==`). Right now, having fastapi[tests] in our dev requirements nails our black to whichever version they have set. Having it like this allows us to update to newer versions quicker if we want to do so.


a-recknagel:
closes #2250

a-recknagel:
Obsolete, I misunderstood what the test-extra was supposed to do.

tiangolo:
Thanks for coming back to close it! 🍰 

PR #2297, opened 2020-11-04
Enable the use of nested structures in peewee

bratao:
Hello, 

I love your lib, it is such a produtivity booster to me!

Another lib that I love is Peewee. One little problem that I have is the conversion from peewee to Pydantic models. When you have a nested model, the proposed getter on the documentation fails to convert it. With the proposed path, it convert the nested reference to model to json, and Pydantic take care of the right conversion.

### DIFF
--- a/docs_src/sql_databases_peewee/sql_app/schemas.py
+++ b/docs_src/sql_databases_peewee/sql_app/schemas.py
@@ -1,6 +1,7 @@
 from typing import Any, List, Optional
 
 import peewee
+from playhouse.shortcuts import model_to_dict
 from pydantic import BaseModel
 from pydantic.utils import GetterDict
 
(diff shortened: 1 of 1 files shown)


tiangolo:
Hey @bratao, thanks for your contribution!

I no longer work with Peewee, and I wouldn't recommend it because it doesn't play well with async/await as it depends on thread locals, and it also doesn't work well with type annotations.

I see the change you are proposing is simple, but I'm not sure how nested models work in Peewee, not sure if that saves them to JSON types, or if it makes automatic relationships with other tables. I fear if it's the second option it would easily end up in infinite recursion when serializing models.

Also, if I accept this I would need to understand it and then maintain it and answer questions about it, etc. And I would actually prefer to recommend other libraries that are made to be compatible with async and await, and that play better with type annotations.

Nevertheless, I think this could be a good candidate for an external blog post, with the tricks you have used to make it work properly for you. :sunglasses: If you write an article like that, you can add it to the external links section in the docs. :memo: 

Given that, I'll now close this PR, but thanks for your effort! :cake:

PR #2316, opened 2020-11-07
two small fixes

hitrust:


docs/en/docs/tutorial/dependencies/index.md
```diff
@@ -48,8 +48,8 @@ And it can return anything you want.
 In this case, this dependency expects:
 
 * An optional query parameter `q` that is a `str`.
-* An optional query parameter `skip` that is an `int`, and by default is `0`.
-* An optional query parameter `limit` that is an `int`, and by default is `100`.
```
ycd, line 52: I do not think changing these is a good idea. Every parameter is optional in the example like @tiangolo states.

```python
async def common_parameters(q: Optional[str] = None, skip: int = 0, limit: int = 100):
    return {"q": q, "skip": skip, "limit": limit}


@app.get("/items/")
async def read_items(commons: dict = Depends(common_parameters)):
    return commons
```

tiangolo: Thanks for the help @ycd! Yep, agreed.

tiangolo:
Thanks for the help here @ycd ! 👏  🙇 

Thanks for coming back and closing the issue @hitrust  👍

PR #2335, opened 2020-11-10
⬆️  Upgrade Starlette to 0.14.2, including internal `UJSONResponse` migrated from Starlette

hanneskuettner:
starlette 0.14.1 includes a [fix](https://github.com/encode/starlette/pull/942) for users that want to use the `@requires` decorator from the `AuthenticationMiddleware` on FastAPI routes.

Breaking in starlette changes as far as I can see are:
 -  `UJSONResponse` was removed from the core [as per this issue](https://github.com/encode/starlette/issues/901), they added [documentation](https://www.starlette.io/responses/#custom-json-serialization) on how to implement custom JSON serialization. `UJSONResponse` is explicitly exported in FastAPI and should either be removed to follow starlette or added back in as done with `ORJSONResponse`. For this PR I chose to add a custom `UJSONResponse` and add test cases.

Is there any reason it was hard locked to 13.6 I am not aware of or is it as straightforward as updating the requirement?

### DIFF
--- a/tests/test_tutorial/test_custom_response/test_tutorial001.py
+++ b/tests/test_tutorial/test_custom_response/test_tutorial001.py
@@ -1,0 +1,36 @@
+from fastapi.testclient import TestClient
+
+from docs_src.custom_response.tutorial001 import app
+
+client = TestClient(app)
+
+openapi_schema = {
+    "openapi": "3.0.2",
+    "info": {"title": "FastAPI", "version": "0.1.0"},
+    "paths": {
+        "/items/": {
+            "get": {
+                "responses": {
+                    "200": {
+                        "description": "Successful Response",
+                        "content": {"application/json": {"schema": {}}},
+                    }
+                },
+                "summary": "Read Items",
+                "operationId": "read_items_items__get",
+            }
+        }
+    },
+}
+
+
+def test_openapi_schema():
+    response = client.get("/openapi.json")
+    assert response.status_code == 200, response.text
+    assert response.json() == openapi_schema
+
+
+def test_get_custom_response():
+    response = client.get("/items/")
+    assert response.status_code == 200, response.text
+    assert response.json() == [{"item_id": "Foo"}]
--- a/fastapi/responses.py
+++ b/fastapi/responses.py
@@ -7,12 +7,23 @@ from starlette.responses import PlainTextResponse as PlainTextResponse  # noqa
 from starlette.responses import RedirectResponse as RedirectResponse  # noqa
 from starlette.responses import Response as Response  # noqa
 from starlette.responses import StreamingResponse as StreamingResponse  # noqa
-from starlette.responses import UJSONResponse as UJSONResponse  # noqa
+
+try:
+    import ujson
+except ImportError:  # pragma: nocover
+    ujson = None  # type: ignore
+
 
 try:
     import orjson
 except ImportError:  # pragma: nocover
     orjson = None  # type: ignore
+
+
+class UJSONResponse(JSONResponse):
+    def render(self, content: Any) -> bytes:
+        assert ujson is not None, "ujson must be installed to use UJSONResponse"
+        return ujson.dumps(content, ensure_ascii=False).encode("utf-8")
 
 
 class ORJSONResponse(JSONResponse):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 requires = [
-    "starlette ==0.13.6",
+    "starlette ==0.14.2",
     "pydantic >=1.0.0,<2.0.0"
 ]
 description-file = "README.md"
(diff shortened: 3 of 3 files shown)


miguescri:
Will this me merged soon? It is problematic that server breaks when installing both FastAPI and Starlette latest versions.

Kludex:
This needs to be refactored when the PR is accepted: https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Python/fastapi

Just posting here as a note for people in the future :open_mouth:

miguescri:
@tiangolo 

> Will this be merged soon? It is problematic that server breaks when installing both FastAPI and Starlette latest versions.

hanneskuettner:
Hey @tiangolo, anything that needs to be resolved before we can merge this?
As far as I can see only the one docs failed to build due to an unrelated issue.

miguescri:
@hanneskuettner try force the CI to re-run so that the tests pass. It is weird, but maybe it is not being reviewed because of it.

Kludex:
The issue on the `build docs` stage was fixed some commits ago, so a rebase will probably fix the issue. :+1:

yareyaredesuyo:
Still need for supporting ujson?

The reason why starletter removed ujson from core looks like reducing 3rd party dependency.

Supporting a lot of 3rd libraries maybe lead to messsy situation.

How about writing custom response in docs (maybe not documented)? like this https://github.com/encode/starlette/pull/1047

hanneskuettner:
@yareyaredesuyo Sure thing. I personally agree with removing `UJSONResponse` for good and pointing at to the custom response documentation. But then I saw that @tiangolo added a `ORJSONResponse` to FastAPI so I thought might as well since it seems to be in the spirit of this library to offer these two out of the box and I didn't want to break backwards compatibility with this PR. 

We could open up an issue and discuss removing both of them maybe.

tests/test_tutorial/test_custom_response/test_tutorial001.py
```diff
@@ -0,0 +1,36 @@
+from fastapi.testclient import TestClient
```
Mause, line 1: Why was this file added in this pr?

hanneskuettner: Since this PR also re-adds `UJSONResponse` that was removed in starlette 0.14.1 and the response code should be tested as well. So I followed suite by adding a test case similar to the ORJSONResponse.

mweinelt:
This change, reimplementing `UJSONResponse`, is what is preventing us in nixpkgs, and probably other distros, from upgrading starlette, as it breaks the whole fastapi testsuite on 0.63.0. Please consider fast-tracking this or something like it.

vinayinvicible:
Please consider upgrading starlette to 0.14.2.
it fixed an issue when run in debug mode.

lwinkler-cloudflight:
Seems fairly straight forward, any reason this is not merged and released yet?

soundstripe (changes requested):
I recommend going ahead with this after pinning starlette to 0.14.2 instead of 0.14.1 -- there is a fix for Python 3.8+ when debug is set (starlette does not show/log tracebacks properly without the fix).

juntatalor:
@tiangolo hi! for version 0.63 debug response is broken in python 3.8 due to starlette bug. Could you please schedule this PR for release?

hanneskuettner:
Rebased onto current master and bumped starlette to 0.14.2.
Let's go!

Mause (approved):

juntatalor (approved):

jcwilson:
This is awesome. Thanks @hanneskuettner!

Any chance we could change the version pin to not be so strict? For instance, use `starlette>=0.14.2,<0.15.0` so we don't have to wait for an update in FastAPI to pick up bugfixes? It would still let users pin to 0.14.2 in their own projects if necessary.

miguescri:
Still not merging?

hanneskuettner:
@jcwilson I also thought about relaxing the pinned version, I found however [this previous](https://github.com/tiangolo/fastapi/pull/1555#issuecomment-643678310) attempt and @tiangolo advocated against relaxing due to tight coupling to starlette internals.

@tiangolo any change in opinion on that front?

jcwilson:
> @jcwilson I also thought about relaxing the pinned version, I found however [this previous](https://github.com/tiangolo/fastapi/pull/1555#issuecomment-643678310) attempt and @tiangolo advocated against relaxing due to tight coupling to starlette internals.
> 
> @tiangolo any change in opinion on that front?

Thanks for pointing that out. I'd still like to advocate for not being so strict, but if doing would delay this PR at all, I'd be glad to argue for it in a follow-up PR. However, I can see how the strict version pinning can make it easier to ensure that the docs remain consistent with expected behavior since much of the FastAPI documentation is actually starlette documentation.

drainer-cloudflight:
Any news on this? Would be really nice to have this merged. @hanneskuettner @jcwilson

hanneskuettner:
TBH doesn't look like there's been a lot of activity in the repo at all in the last few months. 
Anything we can do to help along @tiangolo?

Kludex:
**There's no need for spamming.**

@tiangolo will eventually see this PR. 

Also, for those who don't know, he announced publically that he'll dedicate more time to FastAPI (and other OS projects) starting this month. So, you can expect this to be reviewed by him in the next weeks.

L1ghtn1ng:
@tiangolo any update on this? As this is causing issues for so many people and holding things up for people and distros. If this could get merge ASAP even without a release least people could then point to the master git branch

Issue #2091, opened 2020-09-25
FastAPI ignores json_encoders in the Model


mikegrima:
### First check

* [*] I added a very descriptive title to this issue.
* [*] I used the GitHub search to find a similar issue and didn't find it.
* [*] I searched the FastAPI documentation, with the integrated search.
* [*] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [*] I already read and followed all the tutorial in the docs and didn't find an answer.
* [*] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [*] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [*] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [*] After submitting this, I commit to one of:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

Below is an example that showcases the issue:

```Python
from typing import *

from pydantic import BaseModel
from fastapi.encoders import jsonable_encoder

def sort_func(x: Set[str]) -> List[str]:
   """This function should be called on every serialization. This will convert a Set of strings to a sorted List of strings."""
    print("i was called!")
    return sorted(x)


class SomeModel(BaseModel):
    stuff: Set[str] = ...

    class Config:
        """Define a custom encoder for Sets such that they are sorted lists."""
        json_encoders = {
            set: sort_func
        }

print(SomeModel(stuff={"z", "a", "f", "d"}).json())
# This prints:
# "i was called!"  # GOOD
# '{"stuff": ["a", "d", "f", "z"]}'  # GOOD
# The Pydandic serializer does what it's supposed to do.

print(jsonable_encoder(SomeModel(stuff={"z", "a", "f", "d"})))
# This prints:
# {'stuff': ['z', 'd', 'f', 'a']}  # BAD
# The fastapi encoder is not calling the `sort_func`
```

### Description

* The primary issue is that when a Pydantic model defines special serialization behavior, fastapi is not obeying it.
* In my example, I'm trying to store a set of values internally, but then serialize it into a sorted list.
* The desired behavior is for fastapi to be consistent and obey Pydantic's conventions.

I also looked at other issues that are similar, but not quite the same:
[...]

PR #2061, opened 2020-09-16
🐛 Prefer custom encoder over defaults if specified in `jsonable_encoder`

viveksunder:
In `encoders.py/jsonable_encoder`, default encoders for each type are specified first & are used in preference to custom encoders. 

This affects custom encoders specified for certain types like enums such that the default encoder is always executed and any custom encoders specified are ignored. See - https://github.com/tiangolo/fastapi/issues/1986 - for a description of this issue and example code as well that illustrates the problem.

This PR adds a change to prefer custom encoders if specified in preference to default encoders.

Disclaimer - I am completely new to this code base :-)

Issue #1986, opened 2020-08-28
Unable to get custom json serialization for enums to work


viveksunder:
Hi there! We are using pydantic to define our API models and have a use-case where certain fields stored in our database (mongodb) are enums and need to be translated into an API/user friendly representation. We tried the following in order to solve this problem -

* Define a common enum base class
* Define a common pydantic model base class
* Define a `class Config` section in the above pydantic base class that will implement serialization for our enums.

```
import enum


import pydantic
from fastapi.encoders import jsonable_encoder


class BaseEnum(enum.Enum):
    """
    A base class for all enums that require custom serialization.
    """

    def get_api_repr(self):
        """
        All inheriting classes must implement `def get_api_repr`
        """
        raise NotImplementedError


class SomeStatusEnum(BaseEnum):
    ACTIVE = "ACTIVE"
    INACTIVE = "INACTIVE"

    def get_api_repr(self):
        """
        Return an API friendly value.
        """
        if self.value == self.ACTIVE.value:
            return "Active"

        return "Inactive"


class AppBaseModel(pydantic.BaseModel):
    """
    All app models should inherit from this base class
    so they can benefit from common configuration.
    """

    class Config:
        json_encoders = {
            BaseEnum: lambda v: v.get_api_repr(),
        }


class CustomModel(AppBaseModel):
    """ This does something custom ;) """

    status: SomeStatusEnum


example = CustomModel(status=SomeStatusEnum.ACTIVE)
print(jsonable_encoder(example))

```

Output - 
```
{'status': 'ACTIVE'}
```

Would expect -
```
{'status': 'Active'}
```

### Environment

* OS: Linux
* Python version: 3.8.5
* FastAPI Version: 0.61.0
* Pydantic Version: 1.6.1

PS - We only recently started using FastAPI at work and we love it!!

viveksunder:
Also FYI - I've identified that in https://github.com/tiangolo/fastapi/blob/cdba8481c2773bfca903c439b052199ebea1eda0/fastapi/encoders.py#L60 because for enums this returns .value instead of first checking if there is a custom encoder defined for that type that this explains why the custom encoder I added in the example above doesn't get picked up. 

Does it make more sense to check if any custom encoders for the type have been defined first before applying the default for the type? If so, I'm happy to look into contributing a PR to explore that - just wanted to check if my thinking is reasonable first.

tiangolo:
Thanks for the report and for the fix in the PR @viveksunder! :rocket: 

[...]

PR #2347, opened 2020-11-12
Serialization optimization when using Pydantic for return types

aviramha:
Skip validation and jsonable_encoder in case of a Pydantic's BaseModel in order to speedup


ycd (commented):
I think we should prefer custom encoders first, see #1986, and #2091 and there is a PR that closes both #2061, but as far as I have seen this will override the changes of 2061, am i right, or am i missing smth?

aviramha:
It will override it, but I think it's the correct solution as none of the validation should occur when you use a pydantic model (it's already validated)

aviramha:
@ycd I took a different approach, as it's easier implementation in current design. LMKWYT

tests/test_response_model.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Model(BaseModel):
+    name: str
+
+
+app = FastAPI()
+
+
+@app.get("/", response_class=Model)
```
ycd, line 14: Awesome, having a `PydanticResponse` definitely sounds interesting and sweet.

I think you mean `response_model` here.

ycd: or `response_class=PydanticModel`, not sure but let's have both!

aviramha: Thanks, I fixed it now :)

fastapi/responses.py
```diff
@@ -21,3 +22,10 @@ class ORJSONResponse(JSONResponse):
     def render(self, content: Any) -> bytes:
         assert orjson is not None, "orjson must be installed to use ORJSONResponse"
         return orjson.dumps(content)
+
+
+class PydanticResponse(JSONResponse):
```
raphaelauv, line 27: Why not name this class ModelResponse ?

Since it's response_model= ... and not response_pydantic= in the definition of the endpoint.

aviramha: Fair point.

raphaelauv:
Would be nice to update at the same time the documentation https://fastapi.tiangolo.com/advanced/response-directly/

So it explicit that it's possible to skip validation with ModelResponse

fastapi/responses.py
```diff
@@ -21,3 +22,10 @@ class ORJSONResponse(JSONResponse):
     def render(self, content: Any) -> bytes:
         assert orjson is not None, "orjson must be installed to use ORJSONResponse"
         return orjson.dumps(content)
+
+
+class ModelResponse(JSONResponse):
```
raphaelauv, line 27: Also add a stringdoc to the class , cause it could be confusing with a "response_model" and this class exist in the framework

"""
This class let you respond with a pydantic model without the data validation ...
"""

aviramha: Done!

docs/en/docs/advanced/response-directly.md
```diff
@@ -10,6 +10,12 @@ But you can return a `JSONResponse` directly from your *path operations*.
 
 It might be useful, for example, to return custom headers or cookies.
 
+# Return a Model directly
+
+In case you're already initializing a `pydantic`'s `BaseModel`, you can return `ModelResponse` directly to increase performance
+by skipping re-validation and calls to `jsonable_encoder` by FastAPI.
```
raphaelauv, line 16: why **"re"** validation ? Should it be **" by skipping validation and calls to `jsonable_encoder` by FastAPI."**
	
nothing protect from this case

```python
@app.get("/", response_model=Model)
async def get_pydantic() -> ModelResponse:
    a = Model(arg_str="toto", arg_int = 5)
    a.arg_int="tata"
    return ModelResponse(a)
```

aviramha: Done.

fastapi/responses.py
```diff
@@ -21,3 +22,15 @@ class ORJSONResponse(JSONResponse):
...
+    """
+    A Response that skips re-validation of Pydantic models and calls directly the json method of the pydantic model
```
raphaelauv, line 29: > why "re" validation ?

Same remark

aviramha: done

raphaelauv (approved):
LGTM

falkben:
This looks really nice. I especially appreciate the ability to skip the `jsonable_encoder` invocation.

To be clear, without a custom encoder, `ModelResponse` _should_ be roughly equivalent to:

```python
return JSONResponse(Model(arg_str="toto", arg_int = 5).dict())
```

Which to me isn't such a huge improvement.

By explicitly returning `JSONResponse` we will also skip over the `jsonable_encoder` piece. 

The benefit, as I see it, to this PR is that it makes it relatively easy to define a custom encoder in your pydantic model and get that used automatically. In my case, I have extended `JSONResponse` and defined some common encoding specifications for all my responses (like datetime), instead of customizing it on a "per model" basis, as this then allows it to apply to all json responses.

However, I see the benefit to this, especially if different models require different custom encoders.

It might be nice to show in the docs how one might define and use a custom encoder on a pydantic model. Maybe something like this:

```python
class Model(BaseModel):
    name: str
    birthday: datetime

    class Config:
        json_encoders = {
            datetime: lambda v: v.strftime("%m/%d/%Y, %H:%M:%S"),
        }
```

And then maybe also an example showing it's return value...

```python
model = Model(name="test_model", birthday=datetime.now())
app_model = ModelResponse(content=model)
print(app_model.body)
# b'{"name": "test_model", "birthday": "01/03/2021, 16:37:38"}'
```

aviramha:
@tiangolo any chance to merge this?

raphaelauv:
would be great to check at least that the return model is the one in the decorator ( something with an instanceof )

so something like this must be impossible

```python
class Model(BaseModel):
    name: str

class Model2(BaseModel):
    name: str

app = FastAPI()

@app.get("/", response_model=Model)
async def get_pydantic() -> ModelResponse:
    return ModelResponse(Model2(name="test_model"))
```

YuriiMotov:
@aviramha, thanks for your interest and efforts!

It was probably different at the time this PR was created, but for now it seems that the code example should look like:
```python
@app.get("/", response_model=Model)
async def get_pydantic():
    return ModelResponse(Model(name="test_model"))
```

What I don't like in this approach is that we don't ensure the model returned is correct (the same as model specified in `response_model`).

I personally like the idea of skipping the response model validation in special cases when the instance of correct model is returned, but I think we should do it in a different way. So that the code would look something like:
```python
@app.get("/", response_model=Model, skip_same_model_response_validation=True)
async def get_pydantic():
    return Model(name="test_model")
```
and the following would still give `ResponseValidationError`:
```python
class MyDerivedModel(Model):
    pass

@app.get("/", response_model=Model, skip_same_model_response_validation=True)
async def get_pydantic():
    return MyDerivedModel(name="test_model")  # Should raise ResponseValidationError
```

We could also make it possible to configure `skip_same_model_response_validation` globally for the app or router.

What do you think?

aviramha:
hey Yurii - I didn't use FastAPI (or Python) for the last 3 years so my opinion would be null but I think that automatically skipping makes sense assuming none constructed the model manually..

tiangolo:
Thanks all for the interest and discussions! I think this use case was solved with Pydantic v2, if I remember correctly, with that migration, when the return is a Pydantic model defined in `response_model` (or the return type), it is automatically used to serialize it without going through the old double serialization.

So, the original issue should be solved now. I'll close this one. Thanks! :coffee:

PR #2351, opened 2020-11-14
🐛 Fix edge case with repeated aliases names not shown in OpenAPI

klaa97:
Currently, routes' parameters defined in multiple dependencies are not duplicated since only one parameter per position (path, query, headers, cookie) is retrieved. **The current solution has the caveat of not identifying parameters with the same alias, but different position.**

**Problem example**
Example of parameters shown in the current FastAPI Schema for the given route: 

```
@app.get("/{repeated_alias}")
def get_parameters_with_repeated_aliases(
    path: str = Path(..., alias="repeated_alias"),
    query: str = Query(..., alias="repeated_alias"),
):
    return {"path": path, "query": query}
```

```
'parameters': [{'in': 'query',
                                               'name': 'repeated_alias',
                                               'required': True,
                                               'schema': {'title': 'Repeated '
                                                                   'Alias',
                                                          'type': 'string'}}]
```

**Solution**
To fix the name clashing, it is sufficient to consider parameters duplicated only if they have the same alias and the same type.

The OpenAPI generated by the PR solution is:

```
"parameters": [
                    {
                        "in": "path",
                        "name": "repeated_alias",
                        "required": True,
                        "schema": {"title": "Repeated " "Alias", "type": "string"},
                    },
                    {
                        "in": "query",
                        "name": "repeated_alias",
                        "required": True,
                        "schema": {"title": "Repeated " "Alias", "type": "string"},
                    },
                ]
```


falkben:
maybe i'm a bit dense, but why do you need the same alias to a query and a path param?

klaa97:
> maybe i'm a bit dense, but why do you need the same alias to a query and a path param?

Thank you, the path param and query is a corner case indeed and easily solvable by just changing the name of the query param (which is an implementation detail). On the other hand, consider a combination of query/cookie/header parameter. I agree it can be a corner case, but it is not an implementation detail and may be a particular requirement for an API to expose

tiangolo:
Thanks @klaa97! :cake: 

I don't see an obvious use case that needs this, but it also doesn't hurt, so I'm accepting it. :smile:

I updated the branch and the test with some recent updates.

Thanks for the contribution! :coffee: 

And thanks for the input @falkben! :nerd_face:

PR #2393, opened 2020-11-22
✏ Fix typo/clarify docs for SQL (Relational) Databases

kangni:
By way of code example, I think this is a typo. btw, the tutorial is very well written.


ycd (commented):
Instead of deleting you can change because i think @tiangolo meant **read a single user** here.

* Read multiple users.
* Read a single user.
* Read multiple items.

tiangolo:
Good catch! Thanks @kangni ! :mag: :rocket: 

And thanks @ycd for the help. :coffee: :cake:

Issue #2041, opened 2020-09-12
Persian translations


saeedesmaili:
### Description

I want to start translating the docs to Persian language and will track the progress here.

mosi-kha:
sounds good!

mohsen-mahmoodi:
nice

Kludex:
I'm going to ping the translations' coordination issue here, just in case. #1497

tiangolo:
Awesome! I just added this issue to the pinned translations issue.

meysam81:
@saeedesmaili 
Need any help buddy? I'm game if you want a partner ✌️

lnxpy:
@saeedesmaili I'm also ready to contribute.

onionj:
I'm also ready to contribute. @saeedesmaili

aminalaee:
@Kludex Do you think you can merge this #2395 through so people can start the translations?

And for anyone else, I think when #2395 is merged, we can just pick one or some `markdown` files from English docs and start the translations and put them in the Persian docs folder, and create PRs for them.
I'm happy to contribute in translations and reviews.

Kludex:
I don't have merge rights on FastAPI. We'd need to wait for Sebastián to merge it.

What I recommend (so you don't are blocked) is to create branches on top of #2395, and when that is merged, rebase your branch.

aminalaee:
Thanks for the reply. Honestly I thought about that too, but since the base branch is already a year old, if that's not approved and needs changes then we'll need some modifications in all new branches.

Kludex:
Usually the first commit on a language translation doesn't have any translated pages. This PR has `index.md` translated, so that's maybe why it was not merged.

Even so, if that PR changes, each PR is minimal, only rebase will be needed without conflicts. If you feel like that PR is not going to be merged, for some reason (maybe changes are needed and author away), knowing that you're active on the community, maybe you should create the PR yourself (without the index.md being translated, only in English, like the other languages).

zarinpy:
@tiangolo what about the messages how we can translate them and make a pull request ?

tiangolo:
@zarinpy hmm, I think there are no messages inside the framework... what do you refer to?

NFEL:
> @zarinpy hmm, I think there are no messages inside the framework... what do you refer to?

Error Handler messages, i assume .

PR #2395, opened 2020-11-23
🌐 Add Persian translation for `docs/fa/docs/index.md` and tweak right-to-left CSS

mohsen-mahmoodi:
Hi 

This is a PR to start the Persian documentation translation (lang=fa).
It includes only files generated by running `python ./scripts/docs.py new-lang fa` according to the [Development - Contributing > Docs > Translations > New Language](https://fastapi.tiangolo.com/contributing/#new-language)
You can track Persian translations on this issue: #2041

Thanks a lot.


mohsen-mahmoodi:
Added index page translation to this PR.

rabbitix (approved):

onionj (approved):
I've read the codes and the translations , and i see no problem in it.

SepehrRasouli (approved):

TorhamDev (commented):
A very good idea 👌

TorhamDev (approved):

mohsen-mahmoodi:
> A very good idea 👌

Whe  it can be merged?

docs/fa/docs/css/rtl.css
```diff
@@ -0,0 +1,15 @@
...
+    display: inline-block;
+}
+.external-link{
+    direction: ltr;
+    display: inline-block;
+}
+
+.md-content__inner h1{
+    direction: ltr !important;
+}
```
tiangolo, line 15: As I see that most of these settings only set some elements to always be `ltr`, I think it would be safe to move them to the main CSS (at `docs/en/docs/css/custom.css`). That file is then copied to all the languages.

Also because when the MkDocs configs are regenerated the custom CSS for Persian would be overridden, so, if these CSS styles can be put in a way that applies to everything, it will all work correctly. But if it's absolutely needed to have custom CSS styles only for right-to-left languages, then I would need to figure out a way to support it in the scripts that generate the multi-language docs. 😬

But that would also depend on if the `display: inline-block;` are necessary or not...

tiangolo (commented):
Awesome, thank you! :bow: 

I have a couple of questions about the CSS styles, that we would need to figure out a way to make them work with all the multi-language docs versions.

docs/fa/docs/css/rtl.css
```diff
@@ -0,0 +1,15 @@
+.termy{
+    direction: ltr;
+}
+code{
+    direction: ltr;
+    display: inline-block;
+}
```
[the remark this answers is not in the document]
mohsen-mahmoodi: Hey Sebastián,

I added this because Persian language is RTL and I made the whole document RTL. But some parts of the doc including links and codes should always be LTR, no matter what the translation language is, otherwise they will be rendered in a bad way (Not readable actually). 

Also the `display: inline-block` is required because otherwise, some inline codes will not accept the `ltr` style.

So the answer is yes, please keep it as is.

docs/fa/docs/css/rtl.css
```diff
@@ -0,0 +1,15 @@
...
+.external-link{
+    direction: ltr;
+    display: inline-block;
+}
```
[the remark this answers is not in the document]
mohsen-mahmoodi: Please read the above comment. It applies equally to `external-links`.

docs/fa/docs/css/rtl.css
mohsen-mahmoodi: Yes, the changes that make some styles `ltr` are safe in every other language and it makes sense because they are applied to codes and links and these two items are always ltr, no matter what the language is. But I will pull the latest changes and check if everything is OK before resolving this thread.

mohsen-mahmoodi: I checked it. Yes it is totally safe to move these `ltr` changes to the main CSS.

mohsen-mahmoodi: I tried to check the latest changes and then reverted my branch to the final changes you made.

tiangolo:
Thank you @mohsen-mahmoodi! :bow: :rocket: 

And thanks for the reviews @rabbitix, @onionj, @SepehrRasouli, @TorhamDev :cake: :coffee:

PR #2402, opened 2020-11-24
Add django-ninja to alternatives

vitalik:
Please consider adding django-ninja to alternatives.
Thank you

### DIFF
--- a/docs/en/docs/alternatives.md
+++ b/docs/en/docs/alternatives.md
@@ -36,6 +36,17 @@ It was one of the first examples of **automatic API documentation**, and this wa
 
 !!! check "Inspired **FastAPI** to"
     Have an automatic API documentation web user interface.
+ 
+### <a href="http://django-ninja.rest-framework.com/" class="external-link" target="_blank">Django Ninja</a>
+
+**Django Ninja** is a project that was inspired by **FastAPI**. It has almost all the features from FastAPI and as well brings even more:
+
+ - Deep django integration
+ - ORM support in schemas output
+ - Async views
+ - Multiple API versions
+ - Multiple Response types
+ - High performance thanks to Pydantic (2x faster then Django Rest Framework on serialization/validation, and 100x in specific cases wehere async views do the profit)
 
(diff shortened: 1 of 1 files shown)


ycd:
@vitalik thanks for the contribution, but @tiangolo shared his inspirations before creating FastAPI there, not the alternatives to FastAPI.

vitalik:
@ycd , but page title says "Alternatives, ..." ;)

ycd:
Yes, but it also says 

> **What inspired FastAPI, how it compares to other alternatives and what it learned from them.**

Under each alternative, it has an **Inspired FastAPI to**  section, please consider adding that then.

vitalik:
@ycd, well it's currently other way around django-ninja was created under FastAPI inspiration

raphaelauv:
django-ninja could be part of the alternative like sanic is ( also ASGI ) , but since fastapi is prod-ready , maybe you could wait until you release a stable and battle tested version of django-ninja.

@vitalik ?

vitalik:
@raphaelauv thank you for your feedback

I consider django-ninja production ready at this moment: it is used by multiple companies of different size for 3-5 month now with positive feedback.

Kludex:
@tiangolo always have the last word on those matters, but based on what I have seen, I share the same thoughts as @ycd and @raphaelauv. 

This doesn't look the right place and to me, it looks more a propaganda than a fair description "_It has almost all the features from FastAPI and as well brings even more_". Besides, and the most important thing, even if the title says "Alternatives", the first sentence is _"What **inspired** **FastAPI**, how it compares to other alternatives and what it **learned** from them."_ "Inspired" and "learned" are words in the past, so it's not ambiguous.

Once again, the last word comes from Sebastián. :+1:

tiangolo:
Thanks for the suggestion @vitalik, and thanks for the discussion here everyone. :coffee: 

Yep, I think this would be better placed in External Links. In fact, if you add the `fastapi` GitHub topic to the repo it will show up here automatically: https://fastapi.tiangolo.com/external-links/#projects

Given that, I'm gonna close this PR, but thanks for the interest.

PR #2427, opened 2020-11-27
🌐 Add Chinese translation for Tutorial - Path Operation Configuration

maoyibo:
🌐 Add Chinese translation for Tutorial - Path Operation Configuration


tiangolo:
It seems this was translated in another PR, so I'll close this one, thanks again! 🍰

PR #2439, opened 2020-11-30
🌐 Add Chinese translation for Tutorial - Dependencies in path operation decorators

maoyibo:
🌐 Add Chinese translation for Tutorial - Dependencies in path operation decorators


tiangolo:
It seems this was translated in a different PR, so I'll close this one, but thanks for the effort! ☕

PR #2459, opened 2020-12-02
Fix typo on docs/contributing

CyanBook:


Kludex:
@CyanBook It's not a typo: https://dictionary.cambridge.org/us/dictionary/english/clash

CyanBook:
Anyway, "clash" isn't the proper verb to use. "Crash" is better imo.

docs/en/docs/contributing.md
```diff
@@ -259,7 +259,7 @@ $ uvicorn tutorial001:app --reload
 
 </div>
 
-as Uvicorn by default will use the port `8000`, the documentation on port `8008` won't clash.
+as Uvicorn by default will use the port `8000`, the documentation on port `8008` won't crash.
```
Kludex, line 262: Based on your comment I've investigated and it looks like the most known term seems to be `overlap`. :thinking: 
```suggestion
as Uvicorn by default will use the port `8000`, the documentation on port `8008` won't overlap.
```

CyanBook: I agree with you. "overlap" is better than "crash"

ycd (approved):

tiangolo:
Thanks for your interest! But I actually think "clash" would be more appropriate here, as it wouldn't be possible for applications to take the same port twice, ad the same time, and overlap. One of the two applications will crash, so I think the word "clash" conveys that message better.

Given that, I'll pass on this one. But thanks for the effort! :coffee:

PR #2473, opened 2020-12-05
Propagate callable's return type annotation for Depends/Security

kernc:
I was struck by the following paragraph in [tutorial for `Depends`](https://fastapi.tiangolo.com/tutorial/dependencies/classes-as-dependencies/#type-annotation-vs-depends):
https://github.com/tiangolo/fastapi/blob/51e920e2fcf3e2a715081f28e3187549f2bf531a/docs/en/docs/tutorial/dependencies/classes-as-dependencies.md#L138-L150

I though that's easily solvable with [`typing.TypeVar`](https://docs.python.org/3/library/typing.html#typing.TypeVar), and this PR serves as a untested proof-of-concept, further corroborated only by the fact that the following short snippet passes mypy checks:
```py
from typing import Callable, TypeVar

T = TypeVar('T')

class C:
    ...

def foo(x: Callable[..., T]) ->  T:
    return bar()

def bar():
    return ...

c: C = foo(C)   # <-- no error here
```
This alleviates the need for repeating the dependency (e.g. `CommonQueryParams`) in the annotation. :crossed_fingers:


kernc:
While an obvious example of laziness and poor judgement, with tests still passing, the proof-of-concept seems to work. :sweat_smile:

kernc:
Finally took time to make the tests pass. Please have a look. :grin:

fastapi/param_functions.py
```diff
@@ -239,15 +241,19 @@ def File(  # noqa: N802
 
 
 def Depends(  # noqa: N802
-    dependency: Optional[Callable[..., Any]] = None, *, use_cache: bool = True
-) -> Any:
-    return params.Depends(dependency=dependency, use_cache=use_cache)
+    dependency: Optional[Callable[..., _T]] = None, *, use_cache: bool = True
+) -> _T:
+    return params.Depends(dependency=dependency, use_cache=use_cache)  # type: ignore
```
Kludex, line 246: I'm curious. Why not propagate the _T to the Depends? I mean, define a T (or another name there and import on this file as well) 

I'm asking this because of the ignore comment.

kernc: I'm no expert. `Depends` is a class:
https://github.com/tiangolo/fastapi/blob/33be5fc8baa02bae949e1986f1821dae8f84f487/fastapi/params.py#L316
How do I type-override a class?

GeorchW: I think you can just make the class generic:
```py
from typing import TypeVar, Generic

T = TypeVar("T")


class Depends(Generic[T]):
    def __init__(
        self, dependency: Optional[Callable[..., T]] = None, *, use_cache: bool = True
    ):
        self.dependency = dependency
        self.use_cache = use_cache
    ...
```

GeorchW:
I'd love to see this change land. Let me know if I can help.

patrick91:
I think this can be closed now as we can use `Annotated`:

```python
@app.get("/items/")
async def read_items(commons: Annotated[dict, Depends(common_parameters)]):
    return commons
```

tiangolo:
Thanks all! :rocket: 

Yep, we can probably achieve these ideas with `Annotated` now :sunglasses: 

I'll close this one now. Thanks! :coffee:

PR #2526, opened 2020-12-17
Add newsletter directory for Chinese translation

waynerv:
@tiangolo Since last commit #2509 , all existing translations need to add `overrides` directory manually, otherwise there will be an error when previewing the translation:
```
$ python ./scripts/docs.py live zh
ERROR   -  Config value: 'theme'. Error: The path set in theme.custom_dir ('/Users/waynerv/Repos/fastapi/docs/zh/overrides') does not exist.
Error: Aborted with 1 Configuration Errors!
```


tiangolo:
Thanks for pointing it out! :mag:  I fixed it here, to not require translations for each language, at least to allow using previews for translations: https://github.com/tiangolo/fastapi/pull/2541

---

Now the question is, should there be a notification translation in each language? I feel it could be misleading because people could subscribe expecting a newsletter in their language, and the newsletter would arrive in English.

Although the subscription form is in English, I'm not sure. What do you think @waynerv ?

blt232018:
> Thanks for pointing it out! 🔍 I fixed it here, to not require translations for each language, at least to allow using previews for translations: #2541
> 
> Now the question is, should there be a notification translation in each language? I feel it could be misleading because people could subscribe expecting a newsletter in their language, and the newsletter would arrive in English.
> 
> Although the subscription form is in English, I'm not sure. What do you think @waynerv ?

you may write some words like ‘the content of newsletter would be written in English' above the `SUBSCRIBE` if you decide to add it to all languages . By the way, people who studying and learning here should know some English more or less, so it won't be a problem.

blt232018 (approved):

waynerv:
@tiangolo I think @blt232018  has got a point. People choose to read documents in their native language just for convenience, but they may also be interested in newsletter in English.

tiangolo:
Maybe it would make sense to translate the file `newsletter.md` 🤓 

For this one, as this is specifically for the override, as I would change this more frequently, for example when a new sponsor comes or leaves, taking the translation would mean that I would have to manually update each translation, and I don't speak the languages. 😅 

So I'll pass on this file. But I could take the one for `newsletter.md`. Thank you! 🤓 ☕

PR #2531, opened 2020-12-17
Update the Security Example to prevent Timing Attacks

urjeetpatel:
Added a dummy verify password check to the examples to prevent timing attacks on the JWT example.


vrslev:
This method [is deprecated](https://passlib.readthedocs.io/en/stable/lib/passlib.context.html#passlib.context.CryptContext.dummy_verify). Is this still relevant and do you know an alternative?

violentmagician:
@vrslev I looked at the example code provided by the passlib project and updated the code to reflect the new recommended way now that dummy_verify() is deprecated.

[Docs Link](https://passlib.readthedocs.io/en/stable/narr/context-tutorial.html?highlight=%E2%80%9Cnever%20verifying%E2%80%9D#verifying-migrating-existing-passwords)

tiangolo:
Thanks for the interest!

In this case, the information that the attacker would get is only if a user exists or not, but no information about the password.

And probably in most cases, the attacker could get that information more easily by trying to create a new account with that username, to detect when a username already exists.

So I would think this would depend more on the particular application, I don't think it's strictly necessary to add it here.

Also because adding it would require explaining what are timing attacks to the docs. And I'm already explaining it in another section. I wouldn't think it's necessary to add a long explanation about that for this section as there's no information about the password that can be exposed, only about the existence of users.

Given that, I'll close this one. But thanks for the effort! :coffee:

PR #2564, opened 2020-12-23
📝 Add link to article in Russian "FastAPI: знакомимся с фреймворком"

trkohler:
Hi there! 👋 

Hope it would be considered as useful addition!

In my article in Russian I covered how to:
- get familiar with FastAPI framework
- set up tests
- connect to async Tortoise ORM 
- make async test for the previous feature

Sorry in advance for the ugly link :( Gatsby generates it based on Russian headline and I still hadn't fix this.

### DIFF
--- a/docs/en/data/external_links.yml
+++ b/docs/en/data/external_links.yml
@@ -211,6 +211,10 @@ articles:
       title: Почему Вы должны попробовать FastAPI?
       author_link: https://github.com/prostomarkeloff
       author: prostomarkeloff
+    - link: https://trkohler.com/fast-api-%D0%B7%D0%BD%D0%B0%D0%BA%D0%BE%D0%BC%D0%B8%D0%BC%D1%81%D1%8F-%D1%81-%D1%84%D1%80%D0%B5%D0%B9%D0%BC%D0%B2%D0%BE%D1%80%D0%BA%D0%BE%D0%BC
+      title: "FastAPI: знакомимся с фреймворком"
+      author_link: https://www.linkedin.com/in/trkohler/
+      author: Troy Köhler
   german:
     - link: https://blog.codecentric.de/2019/08/inbetriebnahme-eines-scikit-learn-modells-mit-onnx-und-fastapi/
       title: Inbetriebnahme eines scikit-learn-Modells mit ONNX und FastAPI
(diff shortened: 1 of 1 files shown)


ArcLightSlavik (changes requested):
1. Screenshots for postman aren't loading 
Screenshot 2020-12-23 at 23 40 35

2. Think the link should be just renamed to an english one.

3. I severely disapprove the usage of `response_model_exclude=["password"], # new !`, you can get the same result by splitting your pydantic models, example: https://fastapi.tiangolo.com/tutorial/sql-databases/.

trkohler:
Thank you for noticing @ArcLightSlavik seems like I fixed images from postman and slugs -- the whole time I thought it requires much more work than it actually was.

on third point I'll change that according to recommendation, but wondering why? Making separate in and out models and also separate ORM Model seems like a tons of code repeating almost the same thing every time. Is there any way to avoid this?

ArcLightSlavik:
Your confusing orm model and pydantic model. Orm model should include every single field in a table, but you wouldn't want to return every piece of information in the response all the time, this is where pydantic comes in. 
Say you have 100 fields in your table but you only want to return 2, your `response_model_exclude` would have to include every field in there and then it would be repeated for any other endpoint you might make(!). By making multiple pydantic models you don't have to duplicate it. For me response_model_exclude is more of `only **this** endpoint doesn't need this field`.

In tortoise it would be something like this? (pseudo code)
```
SoftwareEngineer_Pydantic = pydantic_model_creator(
    SoftwareEngineers, name="SoftwareEngineer"
)
SoftwareEngineer_Out_Pydantic = pydantic_model_creator(
    SoftwareEngineers, name="SoftwareEngineerOut", exclude=('password')
)
```
<https://github.com/tortoise/tortoise-orm/blob/develop/tortoise/contrib/pydantic/creator.py#L117>

tiangolo:
Thanks for the help and suggestions @ArcLightSlavik !

Thanks for the article @trkohler ! :nerd_face: 

Nevertheless, when I try to access it I'm getting a 404. Maybe you updated the URL?

trkohler:
yes, I have!
Sorry, forgot about it. Let me update the url in PR!

[While I'm in process, this is a new url which should work](https://trkohler.com/fast-api-introduction-to-framework)

trkohler:
I fixed the link in PR, should work correctly now

tiangolo:
Great, thanks! :coffee:

PR #2568, opened 2020-12-24
✨ Enable configuring Swagger UI parameters

jmriebold:
At present there are a number of [Swagger UI parameters](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/) that are not configurable when using FastAPI. Under normal circumstances most of these wouldn't need to be used at all, however there are some cases where you might need to override some of these. For example, Swagger UI's syntax highlighting is [known to be slow](https://github.com/swagger-api/swagger-ui/issues/3832), and being able to easily disable this would be very useful in certain situations.


docs/en/docs/advanced/extending-openapi.md
```diff
@@ -248,3 +248,16 @@ Now, to be able to test that everything works, create a *path operation*:
 Now, you should be able to disconnect your WiFi, go to your docs at <a href="http://127.0.0.1:8000/docs" class="external-link" target="_blank">http://127.0.0.1:8000/docs</a>, and reload the page.
 
 And even without Internet, you would be able to see the docs for your API and interact with it.
+
+## Configuring Swagger UI
+
+Extra [Swagger UI parameters](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration) can be set using the `swagger_ui_parameters` argument either in the `FastAPI()` constructor directly in the `
+get_swagger_ui_html()` method. This can be used to disable Swagger UI's syntax highlighting, change the theme, etc.
+
+`swagger_ui_parameters` must a dict mapping strings to strings, as described in the Swagger UI documentation. For example, to disable syntax highlighting:
```
pablogamboa, line 257: typo: `must be a dict`

jmriebold: Good catch, thanks!

jmriebold:
@pablogamboa, any other changes needed to wrap this one up?

pablogamboa (approved):
lgtm great work @jmriebold

jmriebold:
Thanks, appreciate the review @pablogamboa!

long2ice:
Look forward for this merge!

fastapi/openapi/docs.py
```diff
@@ -14,6 +14,7 @@ def get_swagger_ui_html(
     swagger_favicon_url: str = "https://fastapi.tiangolo.com/img/favicon.png",
     oauth2_redirect_url: Optional[str] = None,
     init_oauth: Optional[Dict[str, Any]] = None,
+    swagger_ui_parameters: Optional[Dict[str, str]] = None,
```
akihiro-inui, line 17: I think the second parameter of this line `Dict[str, str]` can be also an integer so I suggest to change it into  `Dict[str, any] `

akihiro-inui (changes requested):
swagger_ui_parameters argument type

fastapi/openapi/docs.py
jmriebold: Good point, changed.

akihiro-inui: Nice!

akihiro-inui (approved):

ycd (approved):

jmriebold:
@pablogamboa, @akihiro-inui, @ycd, this PR's been approved for several weeks now, could we get this merged? Do you need anything else from me?

ycd:
Nope, the change looks good, but only @tiangolo has access to master branch. We need to wait his input.

jmriebold:
Understood, thanks @ycd!

c-shekhar:
Hey, @tiangolo, any plans to get this PR merged soon?
Also @jmriebold great addition! :100: 
We are having a use-case where we are exposing alpha release of our APIs to customers and `tryItOutEnabled` display configuration of [swagegr-ui](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md#display) will be very helpful and handy there for a particular group of end users.

stvnksslr:
Bumping this as it would be a huge help to my teams everyday workflow specifically the syntax highlighting being slow for even fairly small datasets is a killer.

jziolkowski:
another :+1: for this bump from our team as we needed to add some ugly hacks to achieve the same result, while we could have this builtin so nicely. Great job @jmriebold !

fastapi/openapi/docs.py
```diff
@@ -34,6 +35,10 @@ def get_swagger_ui_html(
         url: '{openapi_url}',
     """
 
+    if swagger_ui_parameters:
+        for key, value in swagger_ui_parameters.items():
+            html += f"{key}: {value},"
```
elrik75, line 40: To follow the current JS rendering, you should add an `\n` at the end.
I don't think it's a good idea to pass JS code in argument `{value}`. It's easy to break all the javascript rendering in the browser. Why not using JSON to escape?
```
>>> json.dumps(False)
'false'
>>> json.dumps("string")
'"string"'
>>> json.dumps(1)
'1'
```

jmriebold: Great point, I've made the changes and rebased onto master.

fastapi/openapi/docs.py
```diff
@@ -34,6 +35,10 @@ def get_swagger_ui_html(
         url: '{openapi_url}',
     """
 
+    if swagger_ui_parameters:
+        for key, value in swagger_ui_parameters.items():
+            html += f"{key}: {json.dumps(value)},\n"
```
elrik75, line 40: Maybe, we could catch the `TypeError` we will have ifthe value is not serializable in JSON (datetime, object...) to facilitate the dev experience. Here a proposition:

```python
if swagger_ui_parameters:
    try:
        ...
    except TypeError:
        print("swagger_ui values are serialized to Javascript code, please use only basic types like strings, numbers and booleans"")
        raise

jmriebold: Good idea, change made.

aperullo:
What's holding this MR up?

jmriebold:
@aperullo, as I understand it @tiangolo is the only one with write-access, so it's blocked on him reviewing and/merging it.

elrik75:
And @tiangolo seems to be focused on other projects now. SQLModel and Typed. FastAPI needs more maintainers.

Armster15:
@tiangolo is there a specific reason this hasn't gotten merged yet? This PR has been approved by the reviewers, and this is something many people want (personally me as well 😄). I find it a little odd that there is no official response from you yet.

I hope this didn't come off as condescending by the way, thanks so much for your work on FastAPI :)

ghandic:
I've put in a PR for a plug-in system too, I think it would be better to have an approach that is more general and flexible

long2ice:
This project probably don't like new feature, check the recent activity

alexjolig:
I'm having a big issue with large response JSONs and because I'm not able to deactivate syntax highlights from Swagger UI, my project flow is freezing. Merging this would really keeps me going. @tiangolo I hope you're somewhere to be found!

tiangolo:
Thanks for your work @jmriebold! :cake: 

And thanks for the discussion everyone.

I fixed an error in the implementation that would break Swagger UI if parameters with dots like `syntaxHighlight.theme` were used.

I also refactored the other default configurations included by FastAPI to allow overriding them, and documented it.

I refactored and updated the docs, to include source code examples that are now tested, explain defaults and overrides, and added some screenshots.

This will be available in FastAPI version `0.72.0`, I'll release it in a couple of hours. 🔖

jmriebold:
Thanks @tiangolo!

ghandic:
This will now require rework to the plug-in system, is there still interest in having that? Don't want to reinvest the time if there isn't the appetite https://github.com/tiangolo/fastapi/pull/4060

PR #2569, opened 2020-12-24
Update starlette to `~=0.14`

dhirschfeld:
Just seeing if there is any appetite to:
  1. Update the `starlette` pin
  2. Loosen the exact pinning

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 requires = [
-    "starlette ==0.13.6",
+    "starlette ~=0.14.1",
     "pydantic >=1.0.0,<2.0.0"
 ]
 description-file = "README.md"


Mause:
This is a duplicate of https://github.com/tiangolo/fastapi/pull/2335

dhirschfeld:
Looks like it's not compatible due to the removal to `UJSONResponse`:
```
fastapi/responses.py:10: error: Module 'starlette.responses' has no attribute 'UJSONResponse'; maybe "JSONResponse" or "Response"?
```
***xref:*** https://github.com/encode/starlette/releases/tag/0.14.1

dhirschfeld:
> This is a duplicate

So, it is - sorry should have searched first!

I still think there's value in relaxing the exact pinning so wanted to check if @tiangolo had changed his position since #1555?

tiangolo:
Thanks for the discussion! 🍰 

And nope, I can't relax the pin yet, there have been constant/recent breaking changes. And in fact, the current pin on Starlette allows Starlette to move more easily without worrying a lot about breaking stuff between versions and having to deal with a lot of backwards compatibility between versions. That problem will probably come later after the first 1.x versions, but for now, it's better to try to structure the code the best way possible before getting into those new problems... 🤓 

PR #2571, opened 2020-12-26
Update config.yml

Spixer:
For disabling


Mause:
This does not appear to be a valid config option?

Mause (changes requested):
Explain why you've added an option that doesn't exist

tiangolo:
Thanks for the help @Mause ! 🍰

Issue #1218, opened 2020-04-07
[FEATURE] Spanish translations


mariacamilagl:
### Spanish translations tracking

We can use this Issue to track the progress of the Spanish language translations. 🤓 

Please let us know here what page you plan on translating so we can coordinate. 🚀 

#### Tips

- We chose to use "tu" instead of "usted" or "vos"

- If the term is a technical term, leave in English and add an explanation of the meaning of the word in Spanish.  You can do it with a note or an `abbr` element. This should help us avoid confusion with confusing Spanish translations, for example using "instancia" for "instance". 

Thank you for your help! Gracias por tu ayuda! 🎉

mariacamilagl:
I'm currently translating the Features page. 

https://fastapi.tiangolo.com/es/features/

jorgerpo:
Great
Count on me.
Let me grasp the procedure

mariacamilagl:
I'm currently translating the Python Type Intro page.

https://fastapi.tiangolo.com/python-types/

jorgerpo:
ok
I'll do the features page
https://fastapi.tiangolo.com/es/features/

MartinEliasQ:
Can i take https://fastapi.tiangolo.com/tutorial/ ?

mariacamilagl:
> ok
> I'll do the features page
> https://fastapi.tiangolo.com/es/features/

Hi @jorgerpo, thank you for your help. However, the translation is done and it's up for review. If you can please help us review it would be most welcome. The pull request is here: https://github.com/tiangolo/fastapi/pull/1220

mariacamilagl:
> Can I take https://fastapi.tiangolo.com/tutorial/ ?

@MartinEliasQ  That would be awesome! Thank you 🎉 👍

jorgerpo:
> 
> 
> > ok
> > I'll do the features page
> > https://fastapi.tiangolo.com/es/features/
> 
> Hi @jorgerpo, thank you for your help. However, the translation is done and it's up for review. If you can please help us review it would be most welcome. The pull request is here: #1220

ok
great

blaisep:
¡ Saludos @mariacamilagl , @jorgerpo , @MartinEliasQ !
I want to help also. I would like to make a suggestion (and I recognize that I don't yet have any credibility in this project):
Python has a number of tools that make documenting large projects much easier. I believe that we have reached the point where we should consider using those tools.
Signals that we have reached the point: 

[...]

PR #2607, opened 2021-01-05
🌐 Spanish translation of project-generation.md

pedabraham:
Giving continuity to #1218.
In this pull request I made the spanish translation of **project-generation.md**

@mariacamilagl , I will wait for revision


tiangolo:
Hello there! Unfortunately, this file changed in the original, the content is now different, so this translation is no longer applicable. Given that I'll close it, but thanks for the effort! :coffee:

PR #2613, opened 2021-01-07
fix(deps): install asyncio backports only on python<3.7

kumaraditya303:


Kludex (approved):

ycd (approved):

ArcLightSlavik (approved):

raphaelauv (approved):

kumaraditya303:
@tiangolo Can you review this PR ?

musicinmybrain:
So do I understand correctly that this was approved by four reviewers and then closed without merging? It seems like this would still be a valuable change.

PR #2615, opened 2021-01-07
📝 Update docs for `ORJSONResponse` with details about improving performance

falkben:
Update docs to mention to explicitly return a `Response` when thinking about performance. Currently the docs say to use `orjson` for speed, but then demonstrate returning a dictionary, which in my experience can be very slow. The main speed improvement I have found is in returning Responses directly, which avoids the call to `jsonable_encoder`.

Below are results from some tests I did with 7MB+ json data. These are real world and include round trips to the database so to see such a huge effect is rather amazing to me.

To be clear, the difference here is just returning an explicit `ORJSONResponse` object, vs. `JSONResponse` object vs. a `dict` and letting fastapi inspect it. The data returned are exactly the same in all cases.

For more examples w/ code demonstrating the differences see here: 
https://github.com/falkben/fastapi_experiments/blob/master/experiments/orjson_response.py

and some tests here: 
https://github.com/falkben/fastapi_experiments/blob/master/experiments/test_orjson_response.py

In those tests I observed a 6x slowdown when returning a dictionary vs. returning a Response directly

This was also discussed in the following issue: https://github.com/tiangolo/fastapi/issues/360

At some point it would be good to try to optimize `jsonable_encoder` but for now I think it would be good to at least tell people about the performance impact of returning large dictionaries.


ycd (approved):

ArcLightSlavik (approved):

docs/en/docs/advanced/custom-response.md
```diff
@@ -21,6 +21,8 @@ For example, if you are squeezing performance, you can install and use <a href="
 
 Import the `Response` class (sub-class) you want to use and declare it in the *path operation decorator*.
 
+For large responses, returning a `Response` directly is much faster than returning a dictionary and having FastAPI inspect every item inside. However, you be certain that the dictionary you pass to the Response is json serializable (or subclass the JSONResponse class and add your own custom serialization).
```
garry-jeromson, line 24: Missing a word here - probably should be "you _must_ be certain...".

Issue #2625, opened 2021-01-10
Class-based Router Encapsulation


Kojiro20:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Preamble

Please add support for class-based encapsulated route definitions. This is similar to fastapiutils `cbv` but could be much simpler if we make a small change to `routing.py`.

There are many motivations for this, but the biggest to me are:
 - modularization of independently testable router definitions using dependency-injection style testing
 - re-use of shared objects within the route (i.e. a database connection, management of a stateful task)

This is achievable using global variables or re-importing files, and maybe that is the pythonic way, but that always becomes problematic in large projects. I would really love to use FastAPI at my company but can't quite justify it given the lack of encapsulation. Django makes this possible, but I prefer not to use Django because of all the legacy support baggage they carry.

Testing FastAPI projects without this requires global definitions of shared resources and there is no nice way run dependency injection per environment etc without attaching a global config (again, maybe this is the pythonic way and I need to stop thinking in Java terms... lmk).

Wouldn't it be great if we could define a router like this:

```Python
from fastapi import FastAPI, SelflessAPIRouter
from pydantic.main import BaseModel
from databases import Database
from uvicorn import Config
from uvicorn import Server as UvicornServer


late_bound_router = LateBoundAPIRouter()


class Thing(BaseModel):
    is_happy: bool = True
    value: str


class MyNicelyEncapsulatedRoutes():

    def __init__(self, db_connection, stateful_thing):
        self.db_connection = db_connection
        self.stateful_thing = stateful_thing
[...]

PR #2626, opened 2021-01-10
Add support for route encapsulation (#2625)

Kojiro20:
Implementation of route encapsulation per issue #2625.

This adds a `ViewAPIRouter` which can be bound to a class, decorating class-level methods. It works by tagging the routes as `instance_based` and stripping out the instance parameter from reflective document definitions. Then, it wraps the base class to initialize a new router along with each instantiation of the base view class.

### Example usage:
```python
from fastapi import FastAPI
from fastapi.routing import ViewAPIRouter

view_router = ViewAPIRouter()
app = FastAPI()


class Item(BaseModel):
    message_update: str


@view_router.bind_to_class()
class Messages(ViewAPIRouter.View):
    def __init__(self, message: str):
        self.message = message

    @view_router.get("/message")
    def get_message(self) -> str:
        return self.message

    @view_router.post("/message")
    def post_message(self, item: Item) -> str:
        self.message = item.message_update


em_instance = Messages(message="👋")
pt_instance = Messages(message="olá")
en_instance = Messages(message="hello")
app.include_router(em_instance.router, prefix="/em")
app.include_router(pt_instance.router, prefix="/pt")
app.include_router(en_instance.router, prefix="/en")
client = TestClient(app)


def test_view_api_router():

# verify route inclusion
response = client.get("/em/message")
assert em_instance.message in response.text
response = client.get("/pt/message")
assert pt_instance.message in response.text
response = client.get("/en/message")
assert en_instance.message in response.text

# change state in an instance
item = Item(message_update="✨")
response = client.post("/em/message", json=item.dict())
response = client.get("/em/message")
assert item.message_update in response.text
```


tiangolo:
Thanks for the interest! As discussed on the issue/discussion, I would consider there are currently better ways to achieve these results, so for now, I'll pass on this one. But thanks for the effort! :cake:

PR #2646, opened 2021-01-13
Fixed incorrect default value for type Set

lukasgee:
Fixed incorrect default value for the `tags` of type `Set` from the list `[]` to the actual set `set()` according to initial type.

### DIFF
--- a/docs_src/body_nested_models/tutorial004.py
+++ b/docs_src/body_nested_models/tutorial004.py
@@ -16,7 +16,7 @@ class Item(BaseModel):
     description: Optional[str] = None
     price: float
     tax: Optional[float] = None
-    tags: Set[str] = []
+    tags: Set[str] = set()
     image: Optional[Image] = None
 
 


ycd (approved):
Nice catch!

cikay (commented):
The commit message subject should be "Make the default data type of tags value set"
You can also explain the reason in [the body of the commit message](https://robertcooper.me/post/git-commit-messages)

lukasgee:
@cikay Changed it as you requested. Hope it's fine now. Thanks for the hint!

BilalAlpaslan (approved):

tiangolo:
Thanks for the effort @lukasgee! :cookie: 

And thanks for the help @ycd! :coffee: 

Thanks for the interest @cikay, but the commit message is not too important, when I merge I use squash, and I update the PR title before that to keep it consistent with the other commits.

This was covered in a recent PR, so I'm gonna close this one now.

PR #2697, opened 2021-01-25
FEATURE - yield dependencies exit code before response

zoliknemet:
This adds a possibility for dependencies with `yield` to execute any code done after yielding, but before returning response.
Now it is possible to handle and raise exceptions in exit code of dependencies `yield`.
Still, there is no option to change response but we don't want that.

Example
```Python
from fastapi import Depends

async def dependency(value):
    print("1 - Before yield")
    yield value
    print("2 - After yield")

@app.get("/dependency")
async def endpoint(value=Depends(dependency, exit_before_response=True)):
    return JSONResponse(status_code=200, content={"response": "3 - This is my response"})


test_client = TestClient(app)

response = test_client.get("/dependency")
print(response.json().get("response"))


```
Output:
```
1 - Before yield
2 - After yield
3 - This is my response
```

There is further information about exception handling in combination with rest of the framework in documentation under /tutorial/dependencies/dependencies-with-yield/


adriangb:
> Still, there is no option to change response but we don't want that

Could you help me understand (feel free to point me to an existing issue) why we would not want to modify the response from within a dependency (aside from any implementation / backwards compatibility problems)?

zoliknemet:
> > Still, there is no option to change response but we don't want that
> 
> Could you help me understand (feel free to point me to an existing issue) why we would not want to modify the response from within a dependency (aside from any implementation / backwards compatibility problems)?
> 
> Regarding the feature itself, would it be possible to optionally yield one more time, to get the ability to both operate before the response is sent _and_ as a background task within the same dependency?
> 
> ```python
> def dep():
>     yield  # to the path operation
>     # raise an HTTPException or anything else
>     yield  # send the response
>     # background task
>     return
> ```

zoliknemet:
> > Still, there is no option to change response but we don't want that
> 
> Could you help me understand (feel free to point me to an existing issue) why we would not want to modify the response from within a dependency (aside from any implementation / backwards compatibility problems)?
> 
> Regarding the feature itself, would it be possible to optionally yield one more time, to get the ability to both operate before the response is sent _and_ as a background task within the same dependency?
> 
> ```python
> def dep():
>     yield  # to the path operation
>     # raise an HTTPException or anything else
>     yield  # send the response
>     # background task
>     return
> ```
1. 
The idea is to not change the response. The power of this feature is to have handling of dependency operations only on one place.
For example: working with database
```python

async def get_session():
    session = DBSession()
    transaction = session.begin_transaction()
    try:
        yield session
        transaction.commit()
    except DBSessionException:
        transaction.rollback()
    finally:
        session.close()

@app.get("/commit-may-fail")
async def read_items(
    session: DBSession = Depends(get_session, exit_before_response=True)
):
    session.add(Table())
```
In this case you don't have to wright same lines of code on every endpoint
```python
try:
    session.commit()
except(SomeDBException):
    session.rollback()
finally:
    session.close()
```
Those lines of code will be executed before response is returned and you will always have commited or rollbacked transaction when returning response.

Now (without this feature), if you use this kind of dependency client will get response, and then the transaction will be commited. We do not want that.

Maybe my sentence `Still, there is no option to change response but we don't want that` is not so clear here :)

2. Yielding two times is not available in current architecture of FastAPI.
Dependencies use `AsyncExitStack` which is async context manager for dynamic management of a stack
The problem under the hood is `when?` to call next, next on generators

fastapi/dependencies/utils.py
```diff
@@ -536,7 +544,11 @@ async def solve_dependencies(
         if sub_dependant.use_cache and sub_dependant.cache_key in dependency_cache:
             solved = dependency_cache[sub_dependant.cache_key]
         elif is_gen_callable(call) or is_async_gen_callable(call):
-            stack = request.scope.get("fastapi_astack")
+            if sub_dependant.exit_before_response:
+                stack = request.scope.get("fastapi_astack_before_response")
+            else:
+                stack = request.scope.get("fastapi_astack")
+
```
cikay, line 551:
```
key = 'fastapi_astack_before_response' if sub_dependant.exit_before_response else 'fastapi_astack'
stack = request.scope.get(key)
```

fastapi/routing.py
```diff
@@ -151,6 +155,32 @@ async def run_endpoint_function(
...
+    solved_result = await solve_dependencies(
+        request=request,
+        dependant=dependant,
+        body=body,
+        dependency_overrides_provider=dependency_overrides_provider,
+    )
+    values, errors, background_tasks, sub_response, _ = solved_result
+    if errors:
+        raise RequestValidationError(errors, body=body)
+    else:
```
cikay, line 175: Remove else statement here. No need for that. It just causes decreasing indentation level.

zoliknemet: fixed

fastapi/dependencies/utils.py
zoliknemet: usually, I do it like you commented, but this is very sesitive matter

when doing stuff on sensitive matter, I prefer to be explicit, it is easier to see what is going on

i will fix if this will ever be a candidate to merge :)

zoliknemet:
this PR is not relevant anymore: https://fastapi.tiangolo.com/release-notes/#0740

it is not the same feature, but purpose is fulfilled

PR #2704, opened 2021-01-26
Added debugging example/solution where relative imports will break th…

golf-mike:
The debugging examples in the tutorial will not work when using relative imports in your main fastapi python executable. Added an example of how to create a Pycharm run configuration that will work in all this situation as well. Unfortunately not a user of Visual Studio so not able to add an example for that.


Kludex:
I'm not sure if relevant for this PR: https://github.com/Kludex/fastapi-docker-debug

golf-mike:
That is interesting, though not sure how relevant it is for this PR. It does not detail how to accomplish this. Also not a docker user so I don’t really know how to use that info for a tutorial/documentation section.

Kludex:
Yeah, I was not clear enough with my previous message. Maybe rephrasing it: "This is not related, but if someone gets here, know that you can use with docker and VSCode with this."

Sorry for the confusion.

PR #2706, opened 2021-01-26
Add alias parameters to Form Data tutorial

Cleptomania:
This is more or less the same documentation that appears at https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#alias-parameters

However this may make it easier for people to find in the context of using Form Data.

This PR was brought about by https://github.com/tiangolo/fastapi/issues/2705


tiangolo:
Thanks for the effort @Cleptomania! :coffee: 

I prefer to not duplicate the information in the tutorial and instead point to the section that explains it, to not make people read the same thing twice.

I added this PR with a small note in the section about forms to suggest an `alias` for those cases but without duplicating the whole section and the docs.

Given that, I'll close this PR now. But thanks for the interest! :cake:

PR #2714, opened 2021-01-28
Fix section tree in navigation bar for missing translated documentations

tokusumi:
This fixes section tree in a navigation bar of translated documentations.

Ideally, if a part of sections is translated, they and missing translations (EN docs) must be listed on same section tree.
The current document creates an additional section tree for translation, but the original section tree may remain ("Advanced User Guide" is translated into "高度なユーザーガイド" on the just bottom of "Advanced User Guide").

スクリーンショット 2021-01-28 212732

This happens where missing "index.md" but other subsections is translated.

Here improves logic for collecting "nav" section name.
(Sorry for inconvenience that my example is Japanese translation because I can read only that 😢 .


tokusumi:
Docs previewe shows successfully as follows ("Advanced User Guide" is unified with "高度なユーザーガイド" (translation for "Advanced User Guide")): 

スクリーンショット 2021-01-28 221457

scripts/docs.py
```diff
@@ -439,5 +430,51 @@ def get_sections(nav: list) -> Dict[Tuple[str, ...], str]:
     return sections
 
 
+def get_sections_all(nav: list) -> Dict[Tuple[str, ...], Set[str]]:
+    sections = {}
+    for item in nav:
+        if type(item) is str:
+            continue
+        elif type(item) is dict:
```
cikay, line 438: You added the continue keyword to the above if condition. Make it if and [reverse the condition](https://szymonkrajewski.pl/why-should-you-return-early/) and add continue. You can also merge two conditions like the following.
```py
if type(item) is str or type(item) is not dict:
    continue
```

scripts/docs.py
```diff
@@ -439,5 +430,51 @@ def get_sections(nav: list) -> Dict[Tuple[str, ...], str]:
...
+    # Translate English section name, when translated section contains
+    # subset of files in original section.
+    section_en_to_lang = {section: section for section in sections_set.keys()}
+    for en_section, file_set in sections_set.items():
+        for lang_section, lang_file_set in lang_sections_set.items():
+            if lang_file_set.issubset(file_set):
+                section_en_to_lang[en_section] = lang_section
+
+    for en, lang in section_en_to_lang.items():
+        if len(en) >= 2 and en == lang:
```
cikay, line 465: Reverse the condition and add the continue statement to decrease the indentation level.

```py
if len(en) < 2 or en != lang:
    continue

# your code goes here
```

tiangolo:
Thank you @tokusumi! I updated how the docs are generated somewhat recently, and now that there's no content in any additional `mkdocs.yml`, and with the new setup structure, this is fixed. 🚀 

Given that, I'll now close this one. But thanks for all your help with the translations! 🍰

PR #2722, opened 2021-01-29
fix: #624 allow returning a dataclass from a handler

### LINKED ISSUE #624: [FEATURE REQUEST] KeyError using Pydantic dataclass as request body parameter
**Describe the bug**
If a user attempts to use the dataclass from Pydantic to decorate the class used in a request body, a KeyError is thrown when loading the OpenAPI docs.

**To Reproduce**
1. Create a file with this example:
```python
from fastapi import FastAPI
from pydantic.dataclasses import dataclass


@dataclass()
class Item:
    name: str
    price: float
    description: str = None
    tax: float = None


app = FastAPI()


@app.post("/items/")
async def create_item(item: Item):
    return item
```
2. Start the app server
3. Open the OpenAPI docs in a browser (e.g. http://127.0.0.1:8000/docs)
4. See error in terminal `KeyError: <class 'src.main.Item'>`

**Expected behavior**
I expect this to work the same as if I had declared my `Item` class with Pydantic's BaseModel.
Here's an example that works as expected:
```python
from fastapi import FastAPI


class Item(BaseModel):
    name: str
    price: float
    description: str = None
    tax: float = None


app = FastAPI()


@app.post("/items/")
async def create_item(item: Item):
    return item
```

**Environment:**
 - OS: macOS
 - FastAPI Version: `0.42.0`
- Python version, get it with: `3.7.3`

**Additional context**
I stumbled across this while trying to adapt [this example from the docs](https://fastapi.tiangolo.com/tutorial/body/) to use `Pydantic.dataclasses` instead of `Pydantic.BaseModel`.

amitlissack:
A prior PR added support for `pydantic.dataclasses.dataclass` response models. This PR allows returning an instance of a dataclass from a handler.

closes #624 

# Changelog

- `_prepare_response_content` can convert a `dataclass` into a `dict`.


christhekeele:
This would be very useful for my project, anything I can do to help move this along?

amitlissack:
> This would be very useful for my project, anything I can do to help move this along?

Good question. What do you have in mind?

christhekeele:
> Good question. What do you have in mind?

Not sure, I'm really asking the maintainers what's blocking this (if anything) from the next release. :)

This is an open PR with a passing build, submitted several months ago, associated with a pretty good discussion and a decent number of :+1:'s for this `fastapi`. So mostly checking in on what's keeping it from hitting `master` and if there's any way I can volunteer to assist with that process.

george-ycc:
Just encountered another bug not mentioned in #624: directly returning a `dataclass` (NOT pydantic dataclass) will be converted to `dict`, but with an additional field `__initialised__`. (added by pydantic, but not correctly handled by fastapi)

This PR should also fix this problem.

tiangolo:
Great, thanks @amitlissack! :cake: 

As I was not allowed to push commits to your branch, I continued it in this PR: https://github.com/tiangolo/fastapi/pull/3576

I added a couple of tests to include returning `dataclasses` without a `response_model`, and added the fix for it in `jsonable_encoder`.

As this is included in that PR, I'll close this one.

This will be available in the next version of FastAPI, `0.67.0`, released in a couple of hours. :tada: 

Thanks for your contribution! :coffee: :rocket:

PR #2725, opened 2021-01-29
🐛 Fix `response_model` not invalidating `None`

hukkin:
Related to https://github.com/tiangolo/fastapi/issues/2719


Kludex:
I don't think this is going to be accepted. It introduces a breaking change. 😗

ycd:
I agree with @Kludex, but this idea can be evaluated something like:

```python
response_model_skip_validation: bool = False
```

By implementing something like this we can also be able to skip the response validation, and we will still have documentation for that object in our OpenAPI schema.

Kludex:
> I don't think this is going to be accepted. It introduces a breaking change. 😗

Oh _me_ from the past, this comment is not very useful! Glad you learned over time! 👀

tiangolo:
Thanks for the contribution @hukkin! :rocket: 

And thanks for the input @Kludex and @ycd! :nerd_face: I'll make sure I make a bump in the version as this will be a potentially breaking change.

This makes sense, as `response_model` should take anything that is a type annotation, and that Pydantic would receive in a field. Including a `Union[Something, None]`. And it also make sense that if the type annotation doesn't include `None` as a possible type, it should be a server error, the same it would be with a Pydantic field that doesn't allow a `None` type.

This will be available in FastAPI `0.80.0` :tada:

PR #2762, opened 2021-02-05
✨ Support HEAD method when GET is supported [1]

Kludex:
First approach towards closing https://github.com/tiangolo/fastapi/issues/1773.

Considerations:
- This approach has the problem of "order matters" between `get()` and `head()` methods, meaning that if the `get()` is implemented first, the `head()` is ignored.  
- It does not make it optional adding the `HEAD` to OpenAPI, it just doesn't include it. If we want to make it optional we can add a `include_head_schema` on the _HTTP method_ methods.

TO DO:
- [X] Implement logic around `APIRouter` methods.
- [ ] Implement logic around `FastAPI` methods.
- [ ] Add tests.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -310,7 +310,10 @@ class APIRoute(routing.Route):
         self.name = get_name(endpoint) if name is None else name
         self.path_regex, self.path_format, self.param_convertors = compile_path(path)
         if methods is None:
-            methods = ["GET"]
+            methods = ["GET", "HEAD"]
+        elif "GET" in methods:
+            methods = list(methods)
+            methods.append("HEAD")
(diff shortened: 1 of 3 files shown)


fastapi/routing.py
```diff
@@ -310,7 +310,10 @@ def __init__(
         self.name = get_name(endpoint) if name is None else name
         self.path_regex, self.path_format, self.param_convertors = compile_path(path)
         if methods is None:
-            methods = ["GET"]
+            methods = ["GET", "HEAD"]
+        elif "GET" in methods:
```
Kludex, line 314: We can add `HEAD` either here, or on the method. One of those needs to be removed.

Issue #1773, opened 2020-07-22
Automatically support HEAD method for all GET routes, as Starlette does


matthewlloyd:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

Here's a self-contained [minimal, reproducible, example](https://stackoverflow.com/help/minimal-reproducible-example) with my use case:

```Python
from fastapi import FastAPI
import pytest
from fastapi.testclient import TestClient

app = FastAPI()

@app.get("/")
def read_root():
    return {"Hello": "World"}

client = TestClient(app)

def test_index_head():
    response = client.head("/")
    assert response.status_code == 200
```

### Description

* The above test fails because the HEAD request returns a 405.
* This is counterintuitive.

### The solution you would like

* To better support the HTTP standard, all routes that handle GET methods should automatically handle HEAD methods, too.
* This is similar to what Starlette's router already does: https://github.com/encode/starlette/issues/45
* This should happen without requiring the developer to do any additional work.

### Describe alternatives you've considered

* It's currently possible to do this manually by adding `@app.head` and a helper method for each route, but it's cumbersome:

```
@app.head("/")
def read_root_head():
    return Response()
```

* It could also be done using middleware, but that would incur a performance overhead.

### Environment

* OS: [e.g. Linux / Windows / macOS]: Linux
* FastAPI Version [e.g. 0.3.0]: 0.60.1
* Python version: 3.8.5

phy25:
encode/starlette#45 is only about static FileResponse. I don't know how a default should be set for a dynamic API.

[...]

PR #2763, opened 2021-02-06
✨ Support HEAD method when GET is supported [2]

Kludex:
This is the second approach towards closing #1773.

Slower performance compared to #2762 at startup. 

TO DO:
- [X] Support `HEAD` on `GET` methods by default.
- [ ] Modularize the core logic of this PR.
- [ ] Add tests.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1,6 +1,7 @@
 import asyncio
 import enum
 import inspect
+import itertools
 import json
 from typing import (
     Any,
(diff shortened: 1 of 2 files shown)


fastapi/routing.py
```diff
@@ -627,6 +628,16 @@ def include_router(
             responses = {}
         for route in router.routes:
             if isinstance(route, APIRoute):
+                if "GET" in route.methods and "HEAD" not in route.methods:
```
cikay, line 631: Reverse the condition and add the continue statement to decrease the indentation level like the following.

```py
if "GET" not in route.methods or "HEAD" in route.methods:
    continue

#your code goes here
```

cheesycod:
I disagree, I use HEAD internally in my API to do a simple does resource exist check instead of a complete fetch so can we have a way to disable this too?

fastapi/routing.py
```diff
@@ -627,6 +628,16 @@ def include_router(
...
+                    routes_in_path = [
+                        r
+                        for r in router.routes
+                        if isinstance(r, APIRoute) and r.path == route.path
+                    ]
+                    methods_in_path = [route.methods for route in routes_in_path]
+                    all_methods = set(itertools.chain.from_iterable(methods_in_path))
+                    if "HEAD" not in all_methods:
+                        route.methods.add("HEAD")
```
juantoca, line 640: There is an edge case that might cause troubles with the current implementation:

```python

from fastapi import FastAPI
from fastapi.routing import APIRouter

router = APIRouter(prefix="/router")
router2 = APIRouter(prefix="/router")


@router.get("/")
def read_root():
    return {"Hello": "World"}

@router2.head("/")
def modified():
    print("Custom head implementation")
    return {"Hello": "World"}


app.include_router(router)
app.include_router(router2)
```

In this case, the custom endpoint will be ignored as the builtin HEAD will be appended earlier to the `routes` list at the FastAPI class

PR #2769, opened 2021-02-08
Fixed #2768 - Cache depends errors so it runs only once

### LINKED ISSUE #2768: Input Validation run separately for each dependent (depends not cached when input validation fails resulting in duplicated errors)
### First check

* [x] After submitting this, I commit to one of:
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.

### Example

Here's a self-contained, [minimal, reproducible, example](https://stackoverflow.com/help/minimal-reproducible-example) with my use case:

```Python

import logging
from fastapi import FastAPI, Depends, Request


logger = logging.Logger(__name__)


class UserModel(BaseModel):
    username: str
    name: str = None


app = FastAPI()


@Depends
def get_user_model_data(model: UserModel):
    print("Running function get_user_model_data")
    return model.dict(exclude_unset=True)


@Depends
def log_new_users(request: Request, data: dict = get_user_model_data):
    logger.info(f'Got request from {request.client.host} with username {model.username}')


@app.post('/users', dependencies=[log_new_users])
def create_account(data: dict = get_user_model_data):
    return {
        "status": "OK",
        "data": data,
    }

```

### Description

Above is a simple application with a /users endpoint using a pydantic model which requires a username in the post body.
when the model validation passes all is fine. But when model validation fails, the model is validated for each depends separately resulting in duplicated errors.

I would expect that it should be cached even when there are errors so it is validated only once.

* Run the app
* execute `curl -X POST "http://127.0.0.1:8000/api/users" -H  "Content-Type: application/json" -d '{"username":null, "name":"string"}'`
  it will result in duplicate error as follows
```
{
  "detail": [
    {
      "loc": [
        "body",
        "username"
      ],
      "msg": "none is not an allowed value",
      "type": "type_error.none.not_allowed"
    },
    {
      "loc": [
        "body",
        "username"
      ],
      "msg": "none is not an allowed value",
      "type": "type_error.none.not_allowed"
    }
  ]
}
```

### Environment

* OS: macOS:
* FastAPI Version 0.63.0:

* Python version: 3.9.1

ar45:


ar45:
Needs some more work to cover additional use cases.

tiangolo:
Thanks for coming back and closing it. 👍 

PR #2779, opened 2021-02-09
♻️ Cache errors for dependents

ar45:
No need to re-evalute dependencies which already failed.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -462,6 +462,10 @@ async def solve_generator(
     return await stack.enter_async_context(cm)
 
 
+class _CachedError:
+    pass
+
+
 async def solve_dependencies(
     *,
     request: Union[Request, WebSocket],
@@ -477,9 +481,11 @@ async def solve_dependencies(
     Optional[BackgroundTasks],
     Response,
     Dict[Tuple[Callable[..., Any], Tuple[str]], Any],
+    bool,
 ]:
     values: Dict[str, Any] = {}
     errors: List[ErrorWrapper] = []
+    has_error: bool = False
     response = response or Response(
         content=None,
         status_code=None,  # type: ignore
@@ -513,41 +519,52 @@ async def solve_dependencies(
             )
             use_sub_dependant.security_scopes = sub_dependant.security_scopes
 
-        solved_result = await solve_dependencies(
-            request=request,
-            dependant=use_sub_dependant,
-            body=body,
-            background_tasks=background_tasks,
-            response=response,
-            dependency_overrides_provider=dependency_overrides_provider,
-            dependency_cache=dependency_cache,
-        )
-        (
-            sub_values,
-            sub_errors,
-            background_tasks,
-            _,  # the subdependency returns the same response we have
-            sub_dependency_cache,
-        ) = solved_result
-        dependency_cache.update(sub_dependency_cache)
-        if sub_errors:
-            errors.extend(sub_errors)
+        if (
+            sub_dependant.cache_key in dependency_cache
+            and dependency_cache[sub_dependant.cache_key] is _CachedError
+        ):
+            has_error = True
             continue
-        if sub_dependant.use_cache and sub_dependant.cache_key in dependency_cache:
+        elif sub_dependant.cache_key in dependency_cache and sub_dependant.use_cache:
             solved = dependency_cache[sub_dependant.cache_key]
-        elif is_gen_callable(call) or is_async_gen_callable(call):
-            stack = request.scope.get("fastapi_astack")
-            if stack is None:
-                raise RuntimeError(
-                    async_contextmanager_dependencies_error
-                )  # pragma: no cover
-            solved = await solve_generator(
-                call=call, stack=stack, sub_values=sub_values
-            )
-        elif is_coroutine_callable(call):
-            solved = await call(**sub_values)
         else:
-            solved = await run_in_threadpool(call, **sub_values)
+            solved_result = await solve_dependencies(
+                request=request,
+                dependant=use_sub_dependant,
+                body=body,
+                background_tasks=background_tasks,
+                response=response,
+                dependency_overrides_provider=dependency_overrides_provider,
+                dependency_cache=dependency_cache,
+            )
+            (
+                sub_values,
+                sub_errors,
+                background_tasks,
+                _,  # the subdependency returns the same response we have
+                sub_dependency_cache,
+                sub_has_error,
+            ) = solved_result
+            dependency_cache.update(sub_dependency_cache)
+
+            if sub_errors or sub_has_error:
+                has_error = True
+                errors.extend(sub_errors)
+                dependency_cache[sub_dependant.cache_key] = _CachedError
+                continue
+            elif is_gen_callable(call) or is_async_gen_callable(call):
+                stack = request.scope.get("fastapi_astack")
+                if stack is None:
+                    raise RuntimeError(
+                        async_contextmanager_dependencies_error
+                    )  # pragma: no cover
+                solved = await solve_generator(
+                    call=call, stack=stack, sub_values=sub_values
+                )
+            elif is_coroutine_callable(call):
+                solved = await call(**sub_values)
+            else:
+                solved = await run_in_threadpool(call, **sub_values)
         if sub_dependant.name is not None:
             values[sub_dependant.name] = solved
         if sub_dependant.cache_key not in dependency_cache:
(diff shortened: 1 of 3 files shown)


ar45:
Fixes #2768

dltacube:
I like this a lot! If I'm reading this correctly, it always caches errors right?

ar45:
Correct

dltacube:
The only thing I would suggest then is to alter the feature so that it continues behaving the way it does now and add a flag that would allow a particular dependency to be cached on failure. I say this because in general then it wouldn't be a breaking change but also because that's how most implementations (I've seen) work that way...nginx for instance will only cache a result if a 2XX response is returned.

_However_, I personally wouldn't put anymore work into it without a maintainer's opinion as to whether this is something they would consider. You haven't gotten any feedback yet, have you?

fastapi/dependencies/utils.py
```diff
@@ -513,41 +519,52 @@ async def solve_dependencies(
...
+                sub_has_error,
+            ) = solved_result
+            dependency_cache.update(sub_dependency_cache)
+
+            if sub_errors or sub_has_error:
+                has_error = True
+                errors.extend(sub_errors)
+                dependency_cache[sub_dependant.cache_key] = _CachedError
+                continue
+            elif is_gen_callable(call) or is_async_gen_callable(call):
```
cikay, line 555: Make it if. You added the continue statement the above if condition.

YuriiMotov:
Related discussion: https://github.com/fastapi/fastapi/discussions/13828

tests/test_dependency_error_caching.py
```diff
@@ -0,0 +1,45 @@
...
+class Model(BaseModel):
+    name: str
+
+
+@Depends
+def get_model_data(model: Model):
+    return model
+
+
+@Depends
```
YuriiMotov, line 17: Let's use `Depends` in a way it's documented to make it more readable

YuriiMotov (commented):
@ar45, thanks for your efforts!

For now we only cache dependency result in case it has been successfully resolved.
If we use the same dependency multiple times per route and this dependency fails due to request validation error, we will have duplicated errors in error response.
This PR aims to fix this.

Related discussion: https://github.com/fastapi/fastapi/discussions/13828

Now we need to validate the main idea by Sebastián an then work on it a bit more (resolve conflicts, simplify tests, probably add more tests)

nidhishgajjar:

## Summary
This PR implements error caching for FastAPI dependencies to prevent re-evaluation of dependencies that have already failed validation. When a dependency fails validation, it's cached with a sentinel `_CachedError` marker, preventing subsequent dependent dependencies from attempting to resolve the same failed dependency. This optimization avoids redundant validation checks and potential cascading errors.

## Architecture
The PR introduces error caching to FastAPI's dependency resolution system:

- **Sentinel Class**: `_CachedError` acts as a marker for cached error states
- **Modified solve_dependencies**: Returns additional `has_error` boolean and caches errors when they occur
- **Early Exit Logic**: When a dependency's cache key maps to `_CachedError`, skip re-evaluation and propagate the error state
- **Error Propagation**: Errors and error state are tracked and returned through the dependency resolution chain
- **Test Coverage**: New test file validates that dependencies with errors are properly cached and dependent dependencies are not executed

The implementation follows FastAPI's existing caching patterns and maintains backward compatibility while adding performance benefits.

## Issues

### Critical

**No critical issues found** - The implementation is correct and follows FastAPI patterns.

### Warnings

**1. fastapi/dependencies/utils.py:9-10** - **Empty sentinel class**:

```python
class _CachedError:
    pass
```

While the empty class works as a sentinel, consider making it more explicit:

```python
class _CachedError:
    """Marker class for cached dependency errors."""
    def __repr__(self):
        return "<CachedError>"
```

This improves debugging when inspecting the cache and makes the intent clearer.

**2. fastapi/dependencies/utils.py:51-56** - **Cache key collision risk**:

```python
if (
    sub_dependant.cache_key in dependency_cache
    and dependency_cache[sub_dependant.cache_key] is _CachedError
):
    has_error = True
    continue
```

The cache check uses `is` comparison for the sentinel but if `dependency_cache[sub_dependant.cache_key]` is an instance of `_CachedError` instead of the class itself, this check would fail. Consider using `isinstance()`:

```python
if (
    sub_dependant.cache_key in dependency_cache
    and isinstance(dependency_cache[sub_dependant.cache_key], _CachedError)
):
```

**3. fastapi/dependencies/utils.py:95** - **Potential cache pollution**:

```python
dependency_cache[sub_dependant.cache_key] = _CachedError
```

This caches the class itself rather than an instance. Consider creating an instance:

```python
dependency_cache[sub_dependant.cache_key] = _CachedError()
```

### Suggestions

**1. Type annotation improvement** - fastapi/dependencies/utils.py:20 has an incomplete type annotation:

```python
] -> Tuple[  # Missing generic types
    Dict[str, Any],
    List[ErrorWrapper],
    Optional[BackgroundTasks],
    Response,
    Dict[Tuple[Callable[..., Any], Tuple[str]], Any],
    bool,
]
```

The return type should be a proper Tuple with all generic types.

**2. Variable naming** - fastapi/routing.py:131, 139 use vague variable names:

```python
values, errors, background_tasks, sub_response, _, _2 = solved_result
values, errors, _, _2, _3, _4 = solved_result
```

Consider more descriptive names:

```python
values, errors, background_tasks, sub_response, _, sub_has_error = solved_result
values, errors, _, _, _, _has_error = solved_result
```

**3. Documentation** - Consider adding docstrings or comments explaining:
- Why error caching is needed (performance, avoiding cascading errors)
- The expected behavior when errors are cached
- How this interacts with existing caching mechanisms

**4. Test coverage** - The test in `test_dependency_error_caching.py` could be expanded:
- Test that error cache doesn't interfere with successful dependencies
- Test that error cache is cleared appropriately between requests
- Test that error caching works with nested dependencies
- Test error caching with async and sync dependencies

**5. Edge cases** - Consider testing:
- What happens if `cache_key` is None? (The check assumes it's present)
- Behavior with multiple concurrent requests to the same endpoint
- Interaction with dependency overrides
- Behavior when cache is manually cleared

**6. Performance validation** - Consider adding benchmarks to validate the performance improvement:
- Measure time before and after error caching
- Test with deeply nested dependency chains
- Compare error handling performance for different scenarios

**7. Backward compatibility** - The signature change to `solve_dependencies` is internal, but ensure:
- All call sites are updated correctly
- No breaking changes in public APIs
- Existing tests still pass

**8. Code organization** - The `_CachedError` class is defined in `utils.py` but is only used within that module. Consider moving it closer to where it's used or making it a module-level constant if it's truly a singleton.

**9. Error message improvement** - The test assertion could be more informative:

```python
raise AssertionError(
    f"Should have not reached this function, data: {data!r}, ran: {ran}"
)
```

Consider using pytest.raises with expected exception instead.

**10. Cache invalidation** - The PR doesn't address cache invalidation. Consider documenting:
- When error cache entries should be cleared
- Whether error cache persists across requests
- How to manually clear the error cache if needed

## Cross-file Impact

- **Breaking changes**: None (internal implementation change)
- **Public API**: No changes to public interfaces
- **Internal API**: `solve_dependencies` signature changed to return additional `has_error` boolean
- **Performance**: Improves performance by avoiding redundant dependency validation
- **Caching**: Extends existing dependency caching to include error states
- **Test Coverage**: Adds test file for error caching functionality
- **Error Handling**: Improves error propagation through dependency chains

## Assessment
✅ Approve

This is a well-implemented performance optimization for FastAPI's dependency system. The error caching logic is sound and addresses a real inefficiency where failed dependencies were being re-evaluated in dependent dependencies.

**Strengths:**
- Follows FastAPI's existing caching patterns
- Clean implementation with minimal code changes
- Good test coverage for the new functionality
- Improves performance for cases with dependency validation failures
- Maintains backward compatibility with existing code
- Proper error propagation through dependency chains

**Minor improvements needed:**
- Consider making `_CachedError` an instance-based sentinel
- Add proper type annotations for Tuple return types
- Improve variable naming for ignored return values
- Expand test coverage for edge cases
- Add documentation explaining the error caching behavior

**Implementation quality:**
- Correct use of cache checking and early exit
- Proper error state tracking through the dependency chain
- Clean separation of error caching logic
- Good integration with existing dependency resolution

**Code style:**
- Follows FastAPI conventions
- Minimal changes to existing code
- Clear logic flow in the caching implementation
- Appropriate use of sentinel pattern

**Testing:**
- Test correctly validates that dependencies with errors are cached
- Test ensures dependent dependencies are not executed when dependencies fail
- Test file follows FastAPI's test structure

**Performance impact:**
- Reduces redundant validation checks for failed dependencies
- Avoids potential cascading errors in complex dependency chains
- Minimal overhead added to successful dependency resolution

The implementation is correct and provides a meaningful performance improvement. The suggested improvements are minor and don't block merging. The PR successfully addresses the issue of re-evaluating dependencies that already failed validation.

tiangolo:
Thanks for the interest here! :coffee: 

There are several refactors I have planned that will affect how this code works, plus there's a risk of edge cases breaking. So for now, I prefer to pass on this one until I get that done. Thanks! :coffee:

PR #2790, opened 2021-02-10
🔧 Configure strict pytest options and update/refactor tests

graingert:
currently upstream is having to override pytest options to ignore these warnings https://github.com/samuelcolvin/pydantic/pull/2337#discussion_r574145761


docs_src/custom_response/tutorial008.py
```diff
@@ -7,5 +7,8 @@
 
 @app.get("/")
 def main():
-    file_like = open(some_file_path, mode="rb")
```
graingert, line 10: @tiangolo pytest 6.2 includes support for a new plugin [unraisableexception](https://docs.pytest.org/en/stable/usage.html#warning-about-unraisable-exceptions-and-unhandled-thread-exceptions) that can be used to detect the ResourceWarnings raised when files are garbage collected without being closed explicitly

tests/test_tutorial/test_request_files/test_tutorial001.py
```diff
@@ -152,35 +150,35 @@ def test_post_body_json():
     assert response.json() == file_required
 
 
-def test_post_file(tmpdir):
-    path = os.path.join(tmpdir, "test.txt")
-    with open(path, "wb") as file:
-        file.write(b"<file content>")
+def test_post_file(tmp_path):
+    path = tmp_path / "test.txt"
+    path.write_bytes(b"<file content>")
```
tiangolo, line 155: I love these refactors using `pathlib` with `tmp_path`, thanks for that! :rocket:

docs_src/custom_response/tutorial008.py
tiangolo: Get it, thanks! I think I'll have to update something in the docs to account for these changes, as this file is injected in the examples. But I can do that later, for now we can check the other questions I have.

pyproject.toml
```diff
@@ -98,3 +98,18 @@ all = [
...
 profile = "black"
 known_third_party = ["fastapi", "pydantic", "starlette"]
+
+[tool.pytest.ini_options]
+addopts = [
+  "--strict-config",
+  "--strict-markers",
+]
+xfail_strict = true
+junit_family = "xunit2"
```
tiangolo, line 108: I was just going to ask about each of these options but after a bit of research now I understand the reason for them.

So, in short, the strict configs (as you describe it in the title) and the future version of files with the `junit_family`. Right?

pyproject.toml
```diff
@@ -98,3 +98,18 @@ all = [
...
+filterwarnings = [
+    "error",
+    'ignore:"@coroutine" decorator is deprecated since Python 3\.8, use "async def" instead:DeprecationWarning',
```
tiangolo, line 111: I see that without this ignore it fails on Python 3.8 because `aiofiles` uses `@coroutine`. I imagine this will be temporary as (IIRC) with `anyio` there will be no need for `aiofiles`, right?

pyproject.toml
```diff
@@ -98,3 +98,18 @@ all = [
...
+    'ignore:The explicit passing of coroutine objects to asyncio\.wait\(\) is deprecated since Python 3\.8:DeprecationWarning',
```
tiangolo, line 112:
```suggestion
```

I removed this ignore locally and tried with Python 3.8 and 3.9, but it seems it was not breaking. Is there something else I'm missing?

pyproject.toml
```diff
@@ -98,3 +98,18 @@ all = [
...
+    'ignore:int_from_bytes is deprecated, use int\.from_bytes instead:cryptography.utils.CryptographyDeprecationWarning',
```
tiangolo, line 113:
```suggestion
```

I just upgraded `python-jose` in https://github.com/tiangolo/fastapi/pull/3468 to the latest version solving this deprecation, so I think this can now be removed.

pyproject.toml
```diff
@@ -98,3 +98,18 @@ all = [
...
+    'ignore:Exception ignored in. <socket\.socket fd=-1:pytest.PytestUnraisableExceptionWarning',
```
tiangolo, line 114:
```suggestion
```

I wasn't able to trigger this warning. Maybe it was from Starlette <= `0.13.6` (I upgraded recently)?

tiangolo (changes requested):
Thanks for all your work on this! :bow: :nerd_face:

Thanks for improving the file handling and for the refactor to `pathlib`, I love it. :star_struck: 

I have mainly a couple of questions about the configs.

pyproject.toml
graingert: yep, I think pytest 7 will make `--strict` an alias to `--strict-config --strict-markers` but currently `--strict` is a deprecated alias to `--strict-markers`

pyproject.toml
graingert: yep we can use anyio.Path and anyio.AsyncFile but we need to sort out the type annotations for anyio.AsyncFile before it can be used properly

pyproject.toml
graingert: I think this might be only in the Proactor

graingert: also PytestUnraisableExceptionWarning are susceptible to being flakey

graingert: I believe `<socket\.socket fd=-1` comes from closing the ssl wrapped without closing the underlying socket

docs_src/custom_response/tutorial008.py
graingert: I'm going to change how StreamingResponse works - and make it take an async context manager factory instead, deprecating the old aiter parameter

graingert: so it would work like this instead:

```python
return StreamingResponse(iterfile, media_type="video/mp4")
```

graingert:
@tiangolo Hello thanks for enabling my build!
There's a number of test failures, and I'd like to push fixes to them
I see you're currently working on my branch please can you let be know if you've finished

tiangolo:
Hey @graingert! Sure thing, go ahead! I just updated the docs that use that tutorial example to account for the changes, but that's done. And thanks a lot!

BTW, I commented out the pytest ignores to see if they would explode or not.

Also, I think the errors are because some recent tests used Pydantic models with `Test` in the name that were added in a recent PR. If you wanna fix those, that's awesome, but otherwise, you can just let me know and I can do it. :nerd_face:

graingert:
Yep that looks like it to me, I'm happy for you to keep pushing to my
branch!

PR #2796, opened 2021-02-11
Update tutorial004.py

gsganden:
Use `set()` instead of `[]` as default value for set field.

The code seems to run fine as it is, but `mypy` complains.


gsganden:
Never mind, repeat of https://github.com/tiangolo/fastapi/pull/2646

tiangolo:
Thanks for reporting back and closing it 👍

PR #2803, opened 2021-02-12
📝 Remove note about (now supported) feature from Swagger UI in `docs/en/docs/tutorial/request-files.md`

gsganden:
It appears to me that this note is no longer needed because Swagger UI now does support multiple file uploads.

Screen Shot 2021-02-12 at 4 39 34 PM


JonasKs (approved):
Agree, just did the tutorial and noticed this.

PR #2816, opened 2021-02-15
Update tutorial001.py

gsganden:
Replace default `[]` with `set()` for set field. Same as https://github.com/tiangolo/fastapi/pull/2646/files but for a different file.

### DIFF
--- a/docs_src/body_nested_models/tutorial004.py
+++ b/docs_src/body_nested_models/tutorial004.py
@@ -16,7 +16,7 @@ class Item(BaseModel):
     description: Optional[str] = None
     price: float
     tax: Optional[float] = None
-    tags: Set[str] = []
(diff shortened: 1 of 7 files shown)


gsganden:
I did some `grep`ping and caught a few more cases.

ycd (changes requested):
This line conflicts #2646.

gsganden:
Sorry, I'm not following what's happening. I think this PR makes the same change as #2646 and also makes the same change in a few other places. Would you like me to take out the one change that matches #2646?

tiangolo:
Thanks for your contribution @gsganden! :bow: 

And thanks @ycd for the help! :cake: 

It seems this was also covered in another PR, so I'm gonna close this one. But thanks for the effort! :coffee:

PR #2833, opened 2021-02-18
automatically handle 204 status code when not overriding twith custom…

aviramha:
… response class
related to https://github.com/tiangolo/fastapi/issues/1458
related to https://github.com/tiangolo/fastapi/issues/2832
related to: https://github.com/encode/httpx/issues/1474

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -168,8 +168,9 @@ def get_request_handler(
     is_coroutine = asyncio.iscoroutinefunction(dependant.call)
     is_body_form = body_field and isinstance(body_field.field_info, params.Form)
     if isinstance(response_class, DefaultPlaceholder):
+        actual_response_class: Type[Response]
         if status_code == HTTP_204_NO_CONTENT:
-            actual_response_class: Type[Response] = Response
+            actual_response_class = Response
         else:
             actual_response_class = response_class.value
     else:
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -50,7 +50,7 @@ from starlette.routing import (
     request_response,
     websocket_session,
 )
-from starlette.status import WS_1008_POLICY_VIOLATION, HTTP_204_NO_CONTENT
+from starlette.status import HTTP_204_NO_CONTENT, WS_1008_POLICY_VIOLATION
 from starlette.types import ASGIApp
 from starlette.websockets import WebSocket
 
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -50,7 +50,7 @@ from starlette.routing import (
     request_response,
     websocket_session,
 )
-from starlette.status import WS_1008_POLICY_VIOLATION
+from starlette.status import WS_1008_POLICY_VIOLATION, HTTP_204_NO_CONTENT
 from starlette.types import ASGIApp
 from starlette.websockets import WebSocket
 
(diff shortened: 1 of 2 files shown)


fastapi/routing.py
```diff
@@ -168,7 +168,10 @@ def get_request_handler(
     is_coroutine = asyncio.iscoroutinefunction(dependant.call)
     is_body_form = body_field and isinstance(body_field.field_info, params.Form)
     if isinstance(response_class, DefaultPlaceholder):
-        actual_response_class: Type[Response] = response_class.value
+        if status_code == HTTP_204_NO_CONTENT:
+            actual_response_class: Type[Response] = Response
+        else:
+            actual_response_class = response_class.value
```
falkben, line 174: the type annotation was dropped here

fastapi/routing.py
```diff
@@ -168,7 +168,10 @@ def get_request_handler(
     is_coroutine = asyncio.iscoroutinefunction(dependant.call)
     is_body_form = body_field and isinstance(body_field.field_info, params.Form)
     if isinstance(response_class, DefaultPlaceholder):
-        actual_response_class: Type[Response] = response_class.value
+        if status_code == HTTP_204_NO_CONTENT:
+            actual_response_class: Type[Response] = Response
```
falkben, line 172: This seems to work as expected, even when you override the response status code, as documented here: https://fastapi.tiangolo.com/advanced/response-change-status-code/#response-change-status-code

I'm not opposed to this, but I wish we didn't have to add another conditional here, and it seems like we might get some unintended behavior, as we're really just taking advantage of the fact that Response converts `None` to an empty byte string, not that we're actually doing any validation of the result (I'm not proposing that we should be... either).

fastapi/routing.py
aviramha: Pylance (Pyright) called this a redefinition, I'll check if mypy throws the same errors and if not I'll add type annotation to the else clasue

fastapi/routing.py
aviramha: I understand your concern, yet I didn't find any more gentle solution.. feel free to propose if you have any :)

fastapi/routing.py
aviramha: I found some solution, not best either but maybe better?

dongfangtianyu:
Hi,
I just met "too much data for declared content length"  
What is the current status of this PR? What can I do somethings ?

kekel87:
Hello. Some news about this MR ? 😅 (we lost a lot of time because of this bug)

funkindy:
Hi, any updates on this PR?

TBBle:
With only a brief glance, it seems this won't work if you set the status code using the [temporary `Response` object](https://fastapi.tiangolo.com/advanced/response-change-status-code/) approach, as you'd have to repeat the check after checking `sub_response.status_code` later in the same function.

So for example, this idiom would still slip content out in violation of the RFC.

```python
@app.get(
    "/d",
    status_code=200,
    responses={
        200: {}
        204: {}
    }
)
async def d(response: Response):
    if do_not_return_content():
        response.status_code = 204
        return
    return {"field": "value" }
```

BilalAlpaslan (approved):

TBBle:
Is this PR superseded by #5145?

tiangolo:
It seems this was fixed in https://github.com/tiangolo/fastapi/pull/5145, right? That fix is available since FastAPI `0.79.0`.

I just tried the test you have here and it seems to pass with the current code, so I would think this is already handled. :tada: 

But let me know if you think there's something else missing.

PR #2841, opened 2021-02-19
Fix/incorrect jsonable encoder arguments cast

juntatalor:
Due to https://pydantic-docs.helpmanual.io/usage/exporting_models/#advanced-include-and-exclude it is possible to pass dicts into Pydantic model `dict()` method for `include` and `exclude` params. 
Types in jsonable_encoder were already correct, but the cast inside function supported only sets.

PR includes:

- The fix itself
- Docs for passing dicts into jsonable_encoder with new part of tutorial
- Tests

### DIFF
--- a/docs_src/response_model/tutorial006.py
+++ b/docs_src/response_model/tutorial006.py
@@ -25,10 +25,10 @@ class Item(BaseModel):
 items = {
     "foo": {
         "name": "Foo",
-        "image": {"url": "http://test.com/img.jpg", "name": "Some image"},
+        "image": {"url": "https://example.com/img.jpg", "name": "Some image"},
         "videos": [
-            {"url": "http://test.com/video_1.mp4", "duration": 40},
-            {"url": "http://test.com/video_2.mp4", "duration": 120},
+            {"url": "https://example.com/video_1.mp4", "duration": 40},
+            {"url": "https://example.com/video_2.mp4", "duration": 120},
         ],
     },
 }
--- a/docs_src/response_model/tutorial006.py
+++ b/docs_src/response_model/tutorial006.py
@@ -1,39 +1,74 @@
-from typing import Optional
+from typing import List, Optional
 
 from fastapi import FastAPI
-from pydantic import BaseModel
+from pydantic import BaseModel, HttpUrl
 
 app = FastAPI()
 
 
+class Image(BaseModel):
+    url: HttpUrl
+    name: str
+
+
+class Video(BaseModel):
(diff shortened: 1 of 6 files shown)


juntatalor:
This fixes #2007

docs_src/response_model/tutorial006.py
```diff
@@ -1,39 +1,74 @@
...
-        "name": "Baz",
-        "description": "There goes my baz",
-        "price": 50.2,
-        "tax": 10.5,
+    "foo": {
+        "name": "Foo",
+        "image": {"url": "http://test.com/img.jpg", "name": "Some image"},
+        "videos": [
+            {"url": "http://test.com/video_1.mp4", "duration": 40},
+            {"url": "http://test.com/video_2.mp4", "duration": 120},
```
SnkSynthesis, line 31: For `"url"` use https://example.com instead of https://test.com since https://example.com is a reserved domain meant for examples in documentation, etc. 

For more information please visit: https://www.iana.org/domains/reserved

juntatalor: fixed that

SnkSynthesis (approved):

iamnotaprogrammer:
Very usefull bugfix which allow to use native pydantic logic!!

jimhartmanknc:
@juntatalor  Was wondering if there was some issue with the `build-docs` pipeline.  I recently found this bug and saw that this PR was in for it.  Last comment was in May and I wanted to see if there was something that could be done.  Thanks

tiangolo:
Thanks for your effort!

The type annotations for `include` and `exclude` in `jsonable_encoder()` were fixed in some recent PR.

Thanks for all the effort with the documentation and examples too. Nevertheless, I prefer not to include more information about `response_model_include` and `response_model_exclude` because I consider those a bad practice. So I'll pass on this PR.

---

If you are curious, let me explain why my point of view.

Those parameters are there only because some people asked for them, but using them means that the generated JSON Schema and OpenAPI will be incorrect, because the models are documented as the original model, while some fields are being excluded. This breaks the documentation, automatically generated clients, etc.

The right way would be to have a different model that doesn't include the fields that are to be excluded and use it in those parameters. And if duplication is a concern, then there could be a base model with only the fields that are always required, and have some models inherit from that (e.g. for creation, read, etc.) and use inheritance to avoid duplication. It would be roughly equivalent in the amount of code, but the result would be much more robust.

---

Thanks for the interest and work! :cake:

PR #2850, opened 2021-02-20
Update benchmarks.md

Jedore:
fix wrong spelling in Chinese translation benchmarks.md


waynerv (approved):

Jedore:
What's next step? Is that all? I have no idea about PR in Github.-_-@waynerv

waynerv:
Yeah, you just need to wait until the creator merges it.

tiangolo:
I see this PR is closed, not sure if there was something else needed here, but I imagine it's no longer needed in some way. Anyway, thanks for the effort! 🍰 

PR #2888, opened 2021-03-02
WIP: ✨ Add probes endpoints

Kludex:
## Summary

More like a prototype of `/liveness` and `/readiness` endpoints supported natively. 

### Proposed Solution :sunglasses: 

This solution proposes storing the `probes_state`. In that state, we define two attributes: `alive` and `ready`. Based on those two attributes, we're able to change the response of the `liveness` and `readiness` probes. 

Those two attributes can be changed at any moment using the `request.app.probes_state.<alive|ready>` for **now**. This is obviously a solution that @tiangolo will not accept, neither I think is a good one. :sweat_smile: I think we should have a duck typed solution here. That being said, some ideas:
- Global variable imported instead of storing the state inside the FastAPI class (which doesn't look that good and it will be exceptional way to do something in FastAPI), for instance:
```python
from fastapi import probes_state
``` 
- Provide a new object the same way BackgroundTasks and Request is inserted, for instance:
```python
from fastapi import FastAPI, LivenessState, ReadinessState

app = FastAPI()

@app.get("/")
async def potato(liveness_state: LivenessState, readiness_state: ReadinessState):
    liveness_state.broken() # Turn alive to False
    liveness_state.correct() # Turn alive to True
    readiness_state.accepting_traffic() # Turn ready to True
    readiness_state.refusing_traffic() # Turn ready to False  
```

### Pros :heavy_check_mark: 
- Looks like people want this feature (check related issue thumbs up).

### Cons :x: 
- Easy to implement i.e. doesn't need to be on FastAPI natively.

### Related issues :nerd_face: 
- https://github.com/tiangolo/fastapi/issues/1907

### References
- https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
- https://spring.io/blog/2020/03/25/liveness-and-readiness-probes-with-spring-boot
- https://www.baeldung.com/spring-liveness-readiness-probes

### Personal note for curious readers
Let your like here if you want this feature, but don't assume this will be merged, please.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -42,6 +42,8 @@ class FastAPI(Starlette):
         servers: Optional[List[Dict[str, Union[str, Any]]]] = None,
         dependencies: Optional[Sequence[Depends]] = None,
         default_response_class: Type[Response] = Default(JSONResponse),
+        liveness_probe_url: Optional[str] = None,
+        readiness_probe_url: Optional[str] = None,
         docs_url: Optional[str] = "/docs",
         redoc_url: Optional[str] = "/redoc",
         swagger_ui_oauth2_redirect_url: Optional[str] = "/docs/oauth2-redirect",
@@ -117,6 +119,10 @@ class FastAPI(Starlette):
         self.extra = extra
         self.dependency_overrides: Dict[Callable[..., Any], Callable[..., Any]] = {}
 
+        self.liveness_probe_url = liveness_probe_url
+        self.readiness_probe_url = readiness_probe_url
+        self.probes_state: State = State({"alive": True, "ready": True})
+
         self.openapi_version = "3.0.2"
 
         if self.openapi_url:
@@ -189,6 +195,24 @@ class FastAPI(Starlette):
                 )
 
             self.add_route(self.redoc_url, redoc_html, include_in_schema=False)
+
+        if self.liveness_probe_url:
+
+            async def liveness(req: Request) -> JSONResponse:
+                if not self.probes_state.alive:
+                    return JSONResponse({"status": "DOWN"}, status_code=503)
+                return JSONResponse({"status": "OK"})
+
+            self.add_route(self.liveness_probe_url, liveness, include_in_schema=False)
+
+        if self.readiness_probe_url:
+
+            async def readiness(req: Request) -> JSONResponse:
+                if not self.probes_state.ready:
+                    return JSONResponse({"status": "OUT_OF_SERVICE"}, status_code=503)
+                return JSONResponse({"status": "UP"})
+
+            self.add_route(self.readiness_probe_url, readiness, include_in_schema=False)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         if self.root_path:
--- a/tests/test_probes.py
+++ b/tests/test_probes.py
@@ -1,0 +1,8 @@
+from fastapi import FastAPI, Request
+
+app = FastAPI(liveness_probe_url="/health/liveness")
+
+
+@app.get("/")
+def change_state(request: Request):
+    request.app.probes_state.alive = not request.app.probes_state.alive


Mause:
I don't see why this is something that needs to be built into fastapi itself, instead of an externally available package?

Divyadarshee:
@Mause  one reason being even external libraries like fastapi-health or fastapi-healthcheck are not working as any other fastapi module when integrated like they are missing out some advantages or features which fastapi has

Kludex:
@Mause You're right. It's even mentioned in the "Cons" section.

But I'm not sure when we cross the line between "should be on FastAPI or not". A clear example is the swagger generation, or even better, the redoc. Why having redoc if we already have the swagger?

I think we should take in consideration for what FastAPI is being used.

Kludex:
@Divyadarshee That's a completely different issue 😅

If a good package that does that doesn't exist already, maybe the community don't need it that much. 😗

But if you want something better than what is around, you should at least open an issue on those packages. It has far higher probability of being attended there than here. 🧐

Divyadarshee:
Yes @Kludex  I agree with that thought process my naïve thought was it would have been a good to have feature but yeah your point is better👍

ArcLightSlavik:
What are the benefits of this solution over doing:

```
@api.get('/notify/v1/health') 
def get_health():
    return dict(msg='OK')
```
```
        readinessProbe:
          httpGet:
            path: /notify/v1/health
            port: 80
        livenessProbe:
          httpGet:
            path: /notify/v1/health
            port: 80
```

ycd:
As addition to comment above, FastAPI's main feature is being modern, i can't see anything modern than built-in probes.

fastapi/applications.py
```diff
@@ -117,6 +119,10 @@ def __init__(
         self.extra = extra
         self.dependency_overrides: Dict[Callable[..., Any], Callable[..., Any]] = {}
 
+        self.liveness_probe_url = liveness_probe_url
+        self.readiness_probe_url = readiness_probe_url
+        self.probes_state: State = State({"alive": True, "ready": True})
```
klaa97, line 124: What is the reason behind using the State object from `Starlette` and not a Pydantic `dataclass`, or anything that constrains the types of this object to `alive` and `ready`?

Kludex: I don't have a good answer. I think the moment I choose to do that I was thinking about "more customizable attributes" (some changes are needed for this to be true). 

About the dataclass, I'm not sure because we'll need to install the dataclasses backport package for 3.6. It's not really a problem by itself, is just that it will be the first here. 😅

But anyway, your comment is relevant. Thank you :)

I'll apply modifications on the PR once I have more feedbacks about the whole idea. Because as stated in the description, it's kind of a prototype.

dimaqq:
It only took 1 g search to discover that these endpoints are used in kubernetes...
May I propose adding a `k8:` to the PR title?

My 2c: I wish there was a more elegant way to extend or mix in additional functionality (such as 2x docs) into `FastAPI` than the current keyword arguments.

fastapi/applications.py
staticdev: @klaa97 I don't see a problem here. If you check on fastapi's code you will not see any dataclasses declared. So I would say LGTM.

staticdev (approved):

tests/test_probes.py
```diff
@@ -0,0 +1,8 @@
+from fastapi import FastAPI, Request
+
+app = FastAPI(liveness_probe_url="/health/liveness")
```
staticdev, line 3: You probably need to add a test for readiness, since this PR is making code coverage drop from 100%.

staticdev (changes requested):

Kludex:
relevant https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check-06

BilalAlpaslan (approved):

Kludex:
This PR has a lot of 👍 , but I don't think it will land. I wouldn't accept it.

staticdev:
> This PR has a lot of 👍 , but I don't think it will land. I wouldn't accept it.

@Kludex I don't see why, only if you don't want this anymore. If you want, first thing I would make it pass the CI checks.. it is a good step in making things approved from my experience.

Kludex:
> @Kludex I don't see why, only if you don't want this anymore. If you want, first thing make it pass the CI checks.. it is a good step in making things approved from my experience.

... ???

In any case, I'm not going to rebase, add tests, or whatever here without @tiangolo's feedback.

fastapi/applications.py
```diff
@@ -190,6 +196,24 @@ async def redoc_html(req: Request) -> HTMLResponse:
...
+            async def liveness(req: Request) -> JSONResponse:
+                if not self.probes_state.alive:
+                    return JSONResponse({"status": "DOWN"}, status_code=503)
+                return JSONResponse({"status": "OK"})
+
+            self.add_route(self.liveness_probe_url, liveness, include_in_schema=False)
+
+        if self.readiness_probe_url:
+
+            async def readiness(req: Request) -> JSONResponse:
```
Siyet, line 210: As an example, I have the following case: 
_the readiness of the endpoint is polled constantly, not only until the service is ready to receive traffic._

So this handler contains logic to check the readiness of the module to receive traffic.
Probably it goes against the rules, but this is a life case and ops engineers say that this is a frequent case.
Perhaps it is worth giving the opportunity to redefine the readiness method.

What do u think?

Siyet:
It looks like there are not enough **startup** probes: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

tiangolo:
Thanks for the proposal @Kludex! 🤓 

And thanks for the discussion everyone! ☕ 

Here's my rationale (more or less) to decide these things:

I consider two options:

* Teaching people how to build something directly, explaining the needed concepts, code, and teaching them how to do it themselves.
* Include some of that functionality in the library (in this one or an additional library, etc), teaching them the needed concepts, and how to use this new feature/library (instead of teaching them how to do things directly themselves).

If the effort required from users to learn the new functionality in the library is a lot less than what wold be needed to learn how to build it themselves, I would put it in the library, if not, I would rather teach them the underlying concepts and code.

At some point, they might still need to learn the underlying concepts and code anyway, and if they had to learn an abstraction before learning the actual concepts, in the end, they had to learn the double.

Also, if the new feature/library adds some noticeable improvement in some way, e.g. developer experience (autocompletion, type annotations) compared to doing things by hand, or makes things particularly easier, then I add the feature/library.

The other thing is, if I document how to add probes by hand, people that come searching for that can find it. But if it's part of the library as a feature, and more keyword arguments, everyone will end up seeing them, and maybe wondering what that's for, and having to learn it or feel they don't know a part of the tool.

I probably consider other things that I don't remember right now from the top of my head, but you get the point.

---

In this particular case, I don't think having this as part of the library is saving users that much effort compared to teaching them how to write the code themselves. And also "liveness" and "readiness" will be quite dependent on the application. So, adding all the needed parameters and configs would require quite some complexity on the library and complexity people would have to learn. So I would prefer to teach them how to add the endpoints themselves and what things to consider.

So, I'll pass on this PR. But I would take a PR adding docs about creating probes by hand. 🤓 ☕

PR #2898, opened 2021-03-05
Fix concurrency.py test coverage

SnkSynthesis:
The coverage of concurrency.py was 79% which made the entire test coverage be 99%  with the coverage given by `./scripts/test.sh`. To fix this, this PR moves the `# pragma: no cover` (In 2 places) in concurrency.py to the upper `except ImportError` block to fix the test coverage of that file.


tiangolo:
Hey there! Thanks for the interest. :coffee:

I think if you were having lower coverage locally it's because the full coverage is done from the multiple versions of Python.

If you check the project's coverage in Codecov, it should currently still be 100%.

But if you feel there's still something to be handled, please create a new PR to avoid the conflicts. Thanks! :cookie:

PR #2926, opened 2021-03-10
initial commit for indonesian translations

pace-noge:


tiangolo:
Thank you! :nerd_face: 

But it seems here: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes that the code for Indonesian would be `id`, instead of `in`.

Could you please update it? Or if it's easier you can also close this PR and create a new one.

Kludex:
@tiangolo This `github-action` is closing also PRs after your replies :thinking: Is this on purpose?

tiangolo:
Yep, it's on purpose @Kludex. After 10 days of (manually) marking something with a label of `answered`.

Issue #1561, opened 2020-06-12
Update docs with parameters that default to None to use Optional


tiangolo:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

Here's a self-contained [minimal, reproducible, example](https://stackoverflow.com/help/minimal-reproducible-example) with my use case:

```Python
from fastapi import FastAPI, Query

app = FastAPI()


@app.get("/")
def read_root(q: str = Query(None)):
    return {"q": q.lower()}
```

### Description

* Open the editor with that code, similar to many examples in the docs.
* It's quite probable that the editor won't complain that `q` can be `None` and that `q.lower()` could raise an error.

### The solution you would like

I would like the examples in the docs to use `Optional`:

```Python
from typing import Optional

from fastapi import FastAPI, Query

app = FastAPI()


@app.get("/")
def read_root(q: Optional[str] = Query(None)):
    return {"q": q.lower()}
```

By using `Optional` the editor would be able to warn that `q.lower()` could raise an error if `q` is `None`.

### Describe alternatives you've considered

To leave it as is.

### Environment

* OS: [e.g. Linux / Windows / macOS]: All (docs)
* FastAPI Version [e.g. 0.3.0]: 0.56.0

To know the FastAPI version use:

```bash
python -c "import fastapi; print(fastapi.__version__)"
```

* Python version: All (docs)

To know the Python version use:

```bash
python --version
```

### Additional context

[...]

PR #2930, opened 2021-03-11
Tiny doc change to improve clarity of Optional type hint note in tutorial

mboratko:
When reading the documentation, the note regarding Optional type hints was somewhat confusing to me - it made it seem like FastAPI was ignoring relevant type information, which goes against all the rest of the beautiful type awareness. This is my attempt to clarify that note.

(Somewhat related to #1561, which seems as though it can be closed.)

### DIFF
--- a/docs/en/docs/tutorial/body.md
+++ b/docs/en/docs/tutorial/body.md
@@ -156,9 +156,11 @@ The function parameters will be recognized as follows:
 * If the parameter is declared to be of the type of a **Pydantic model**, it will be interpreted as a request **body**.
 
 !!! note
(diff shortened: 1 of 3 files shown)


tiangolo:
Thanks for the help and interest @mboratko! :cookie: 

I recently updated all those sections to use `Union[str, None]`, to try and avoid confusion about the word "optional".

Could you give all that a check and see if it makes more sense now?

If it still seems confusing, could you please make a new PR starting from the main branch so that it takes into account those new changes?

Given that, I'll close this PR now. Thank you! :cake:

PR #2945, opened 2021-03-13
🐛 Fix cached dependencies when using a dependency in `Security()` and other places (e.g. `Depends()`) with different OAuth2 scopes

laggardkernel:
```python
class Dependant:
    def __init__(...):
        ...
        # Save the cache key at creation to optimize performance
        self.cache_key = (self.call, tuple(sorted(set(self.security_scopes or []))))
```

To make `.security_scopes` contribute to `.cache_key`, parameter `security_scopes` must be passed to `Dependant.__init__()`. We forgot to do it in this way. Reassignment after init `dependant_instance.security_scopes = ["dummy"]` doesn't change the `.cache_key` anymore. Main related logic is at `get_sub_dependant()`.

This makes `Security` with different scopes ([OAuth2 scopes][oauth2-scopes]) share the same cache. If we have `Security(func)` and `Security(func, scopes=['scope1'])` in a dependency tree. Only one of them will be really called.

The bug could be traced back to bff5dbb on 6/6/19 when the `dependency_cache` is first introduced. Seems a serious bug for anyone using `Security` in dependencies, better fix it soon.

[oauth2-scopes]: https://fastapi.tiangolo.com/advanced/security/oauth2-scopes/


laggardkernel:
I've also considered implementing `Dependant.cache_key` as a `property`. But current fix is enough

- it keeps the designed made by the author to pass `scopes` during `__init__()`
- `Dependant` is an internal class, and `Dependant.__init__()` is  only called few times in the source code

laggardkernel:
Updated once after tweaking the related test.

tiangolo:
Thanks for the detailed explanation and for your contribution @laggardkernel! :bow: :rocket: :cake:

qmorek:
Hi guys,
it looks that changes from this PR broke my privs checking mechanism. 
Simplest code to demonstrate it is below:
```python
from fastapi import FastAPI, Depends, Security
from fastapi.security import SecurityScopes


def get_obj(
    required_scopes: SecurityScopes,
):
    print(f'get_obj called with scopes: "{required_scopes.scope_str}"')
    return 'object'


def check_user_role_for_obj(
    required_scopes: SecurityScopes,
    obj=Depends(get_obj),
):
    print(f'check_user_role_for_obj called for {obj} with scopes:, "{required_scopes.scope_str}"')
    return 'object role'


app = FastAPI()


@app.get(
    '/',
    dependencies=[Security(check_user_role_for_obj, scopes=['read_scope'])],
)
def get_obj_with_scopes(
    obj=Depends(get_obj),
):
    return obj


@app.put(
    '/',
    dependencies=[Security(check_user_role_for_obj, scopes=['edit_scope'])],
)
def put_obj_with_scopes(
    obj=Depends(get_obj),
):
    return obj


@app.get(
    '/child_obj',
    dependencies=[Security(check_user_role_for_obj, scopes=['read_child_scope'])],
)
def get_obj_with_scopes():
    return 'child object'
```

Here is output from calling get_obj_with_scopes endpoint:
```
INFO:     127.0.0.1:58537 - "GET / HTTP/1.1" 200 OK
get_obj called with scopes: "read_scope"
check_user_role_for_obj called for object with scopes:, "read_scope"
get_obj called with scopes: ""
`
``
as you can see get_obj was called twice....
I understand that this change was necessary as I was also suprised that I can not use different scopes for the same dependencies, but thought that it is by desing ;]
Maybe my understanding of this is not correct, but in such case _get_obj_ itself is only used with _Depends_, so no specific scopes are requested for it, so maybe it should be cached?
Especially that I don't even need/use required_scopes: SecurityScopes in _get_obj_ function, so why differentiate cache key by it? Maybe it should depend if get_obj has SecurityScopes in its parameters list?

I am just throwing it for consideration.
As for now I will fix my FastAPI version to 0.79.1 and maybe I need to rethink how my privs model is designed.
Good Job guys!

**Comment by nparley** [comment]
Using a database session inside a `Security` and `Depends` also will open two sessions after this PR. Thus if the security method if returning a sqlalchemy the object return from the two methods will be attached to different sessions.

**Comment by qmorek** [comment]
Hi @tiangolo,
maybe my case is kind specific to how I designed my privileges model, but creating multiple db sessions may lead to unexpected errors... I am not sure if I fully understand how dependency cache is working, but maybe it should be divided to 2 parts first scopes dependant and another not?
Can you please have a look on this?

**Comment by laggardkernel** [comment]
To understand dependency cache, we may start what features are added by FastAPI upon Starlette.

- When a request is received from the client at the server, it's received as bytes
- The ASGI server parse the bytes to `environ, recv(), send()` (ASGI standard)
- Starlette abstract the `environ` dict and `recv()` as a `Request`, which is a more developer-friendly Python object.

In the view/router function, we parse the `Request` (e.g. it's query param, json body, form body, etc), get the info we want, and compose a `Response` object.

`FastAPI` is capable to let developers declare the query param, or form, or json body as arguments of the view function. It helps us use the more finer part of `Request` directly without using the whole `Request`.

The introduction of dependency (`Depends`, `Security`) makes `FastAPI` more powerful. We can declare some composite value based on the primary data (query param, json body, form element, etc). e.g.

```python
async def sum_c(a: int = 0, b: int = 0):
    return a + b

@app.get("/sum")
async def calc_sum(c: int = Depends(sum_c)):
    return c
```

In the above example, the view func doesn't accept query param `a` or `b`, but a composite param named `c`, which is the sum of `a + b`.

To avoid duplicated, high cost calculation in some `Depends`, `FastAPI` caches the result. 

- For `Depends`, the cache key is based on the func wrapped by `Depends`
- For `Security` (subclass of `Depends`), the cache key is based on the wrapped func and param `scopes`. (Formerly `scopes` are ignored cause a bug in the implementation, and this PR fixed it.)

--- 

@nparley @qmorek For your questions, you may separate db session initialization out of existing `Security`s as standalone `Depends`.
This makes the new `Depends`, namely db session initialization process, has its own cache key, being executed only once, and not affected by `scopes` of `Security` anymore.

qmorek:
Hi @laggardkernel,
thank you very much for your answer.

could you eplain what you mean by:
> you may separate db session initialization out of existing Securitys as a standalone Depends, and re-used it as an argument in the existing multiple Securitys.

qmorek:
I took example from: https://fastapi.tiangolo.com/advanced/security/oauth2-scopes/#dependency-tree-and-scopes
and modified it slightly (paswords are in cleartext, but it is not important here) with db dependency as in:
https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/api/deps.py
and
https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/api/api_v1/endpoints/items.py

```python
from typing import Dict
from datetime import datetime, timedelta
from typing import List, Union

from fastapi import Depends, FastAPI, HTTPException, Security, status
from fastapi.security import (
    OAuth2PasswordBearer,
    OAuth2PasswordRequestForm,
    SecurityScopes,
)
from jose import JWTError, jwt
from pydantic import BaseModel, ValidationError

# to get a string like this run:
# openssl rand -hex 32
SECRET_KEY = "09d25e094faa6ca2556c818166b7a9563b93f7099f6f0f4caa6cf63b88e8d3e7"
ALGORITHM = "HS256"
ACCESS_TOKEN_EXPIRE_MINUTES = 30


fake_dict_db = {
    "users": {
        "johndoe": {
            "username": "johndoe",
            "full_name": "John Doe",
            "email": "johndoe@example.com",
            "cleartext_password": "password",
            "disabled": False,
        },
        "alice": {
            "username": "alice",
            "full_name": "Alice Chains",
            "email": "alicechains@example.com",
            "cleartext_password": "password",
            "disabled": True,
        },
    },
    "items": [
        {
            "name": "First"
        },
        {
            "name": "Second"
        }
    ]
}


class Token(BaseModel):
    access_token: str
    token_type: str


class TokenData(BaseModel):
    username: Union[str, None] = None
    scopes: List[str] = []


class User(BaseModel):
    username: str
    email: Union[str, None] = None
    full_name: Union[str, None] = None
    disabled: Union[bool, None] = None


class UserInDB(User):
    cleartext_password: str

oauth2_scheme = OAuth2PasswordBearer(
    tokenUrl="token",
    scopes={"me": "Read information about the current user.", "items": "Read items."},
)

app = FastAPI()


def verify_password(plain_password, db_password):
    return plain_password == db_password


def get_user(db, username: str):
    users_db = db['users']
    if username in users_db:
        user_dict = users_db[username]
        return UserInDB(**user_dict)


def authenticate_user(fake_db, username: str, password: str):
    user = get_user(fake_db, username)
    if not user:
        return False
    if not verify_password(password, user.cleartext_password):
        return False
    return user


def create_access_token(data: dict, expires_delta: Union[timedelta, None] = None):
    to_encode = data.copy()
    if expires_delta:
        expire = datetime.utcnow() + expires_delta
    else:
        expire = datetime.utcnow() + timedelta(minutes=15)
    to_encode.update({"exp": expire})
    encoded_jwt = jwt.encode(to_encode, SECRET_KEY, algorithm=ALGORITHM)
    return encoded_jwt


def get_db() -> Dict:
    print('get_db called')
    return fake_dict_db


async def get_current_user(
    security_scopes: SecurityScopes,
    db: Dict = Depends(get_db),
    token: str = Depends(oauth2_scheme),
):
    print('required scopes are:', security_scopes.scope_str)
    if security_scopes.scopes:
        authenticate_value = f'Bearer scope="{security_scopes.scope_str}"'
    else:
        authenticate_value = f"Bearer"
        headers={"WWW-Authenticate": authenticate_value},
    )
    try:
        username: str = payload.get("sub")
        if username is None:
            raise credentials_exception
        token_scopes = payload.get("scopes", [])
        token_data = TokenData(scopes=token_scopes, username=username)
    except (JWTError, ValidationError):
        raise credentials_exception
    user = get_user(db, username=token_data.username)
    if user is None:
        raise credentials_exception
    for scope in security_scopes.scopes:
        if scope not in token_data.scopes:
            raise HTTPException(
                detail="Not enough permissions",
                headers={"WWW-Authenticate": authenticate_value},
            )
    return user


async def get_current_active_user(
    current_user: User = Security(get_current_user, scopes=["me"])
):
    if current_user.disabled:
        raise HTTPException(status_code=400, detail="Inactive user")
    return current_user


@app.post("/token", response_model=Token)
async def login_for_access_token(
    db: Dict = Depends(get_db),
    form_data: OAuth2PasswordRequestForm = Depends(),
):
    user = authenticate_user(db, form_data.username, form_data.password)
    if not user:
        raise HTTPException(status_code=400, detail="Incorrect username or password")
    access_token_expires = timedelta(minutes=ACCESS_TOKEN_EXPIRE_MINUTES)
    access_token = create_access_token(
        data={"sub": user.username, "scopes": form_data.scopes},
        expires_delta=access_token_expires,
    )
    return {"access_token": access_token, "token_type": "bearer"}


@app.get("/users/me/", response_model=User)
async def read_users_me(current_user: User = Depends(get_current_active_user)):
    return current_user


@app.get("/users/me/items/")
async def read_own_items(
    db: Dict = Depends(get_db),
    current_user: User = Security(get_current_active_user, scopes=["items"])
):
    return {
        'items': db['items'],
        'current_user': current_user
    }


@app.get("/status/")
async def read_system_status(current_user: User = Depends(get_current_user)):
    return {"status": "ok"}
```
if you run this example code and call endpoint: /users/me/items/

you will see in console:
```
get_db called
get_db called
required scopes are: items me
```
so as you see get_db dependency is called twice...

Maybe my modifications are not correct, but it consistent with project template creted by tiangolo: https://github.com/tiangolo/full-stack-fastapi-postgresql

I am aware that it was created before this bug was discovered, so may also not be correct in this matter, but I would love to know how to do this correctly....

laggardkernel:
@qmorek Sorry, my former solution is wrong. I forgot something, the `scopes` is inherited by children in the dependency tree. In your example,

```python
@app.get("/users/me/items/")
async def read_own_items(
    db: Dict = Depends(get_db),
    current_user: User = Security(get_current_active_user, scopes=["items"])
):

async def get_current_active_user(
    current_user: User = Security(get_current_user, scopes=["me"])
):

async def get_current_user(
    security_scopes: SecurityScopes,
    db: Dict = Depends(get_db),
    token: str = Depends(oauth2_scheme),
):
```

`Depends(get_db)` param in `read_own_items()` and `Depend(get_db)` from `get_current_user() <- get_current_active_user() <- read_own_items()` has different cache keys. The later inherits scopes `"items"` and `"me"` as part of the cache key. It makes `get_db()` executed twice during one request.

<details>

<summary>confirmed by printing dependency_cache</summary>

Two different cache keys

- `(<function get_db at 0x10eeebf70>, ())`
- `(<function get_db at 0x10eeebf70>, ('items', 'me'))`

```log
get_db called: 2
dependency_cache: {(<function get_db at 0x10eeebf70>, ()): {'items': [{'name': 'First'},
                                                    {'name': 'Second'}],
                                          'users': {'alice': {'cleartext_password': 'password',
                                                              'disabled': True,
                                                              'email': 'alicechains@example.com',
                                                              'full_name': 'Alice '
                                                                           'Chains',
                                                              'username': 'alice'},
                                                    'johndoe': {'cleartext_password': 'password',
                                                                'disabled': False,
                                                                'email': 'johndoe@example.com',
                                                                'full_name': 'John '
                                                                             'Doe',
                                                                'username': 'johndoe'}}},
 (<function get_db at 0x10eeebf70>, ('items', 'me')): {'items': [{'name': 'First'},
                                                                 {'name': 'Second'}],
                                                       'users': {'alice': {'cleartext_password': 'password',
                                                                           'disabled': True,
                                                                           'email': 'alicechains@example.com',
                                                                           'full_name': 'Alice '
                                                                                        'Chains',
                                                                           'username': 'alice'},
                                                                 'johndoe': {'cleartext_password': 'password',
                                                                             'disabled': False,
                                                                             'email': 'johndoe@example.com',
                                                                             'full_name': 'John '
                                                                                          'Doe',
                                                                             'username': 'johndoe'}}}}
```

</details>

---

@tiangolo It seems making `scopes` part of cache key may have a side effect because of the inheritance and merging of `scopes` in the dependency tree.
But this very inheritance also makes sense when we do this out of security concerns. What's your idea?

laggardkernel:
@tiangolo I have a solution but not sure if this contradicts the original security design.

- `scopes` is still passed down in the dependency tree
- but `scopes` is taken into cache key if the child node is still `Security` while not `Depends` in the dependency tree

qmorek:
@tiangolo would you be able to have a look on this please?

nyte-owl:
This issue of scopes affecting `Depends` cache_key bit me as well in #9513.  Curious what the solution to this could be as it doesn't seem intended. 🤔

DurandA:
I am also affected by this issue and opened a discussion (https://github.com/tiangolo/fastapi/discussions/6024).

> @tiangolo I have a solution but not sure if this contradicts the original security design.
> 
>     * `scopes` is still passed down in the dependency tree
> 
>     * but `scopes` is taken into cache key if the child node is still `Security` while not `Depends` in the dependency tree

@laggardkernel This is the way I wanted to implement a fix when poking with the cache code. I could also make a PR if we get some indication from @tiangolo that this approach is acceptable. Also we could be on the conservative side and introduce a new parameter to `Security` to allow dropping the scopes of non-security `Depends` in cache keys. I am however not sure of how this parameter should be passed down the dependency chain.

DurandA:
I created #9790 to fix this issue. @laggardkernel Let me know what you think of it.

Barnyclark:
Is there a timeline for this to be PR to be merged? This bug is causing a few issues with running FastAPI in production because it halves our available db connections - every request requires two db sessions.

Barnyclark:
@DurandA we have been doing some more testing on this and have found that if you set up the DB using middleware you don't have this issue. This is because although you still have two entries in the cache for `get_db` they both point to the same db connection - e.g:

```
@app.middleware("http")
async def db_session_middleware(request: Request, call_next):
    response = Response("Internal server error", status_code=500)
    try:
        request.state.db = SessionLocal()
        response = await call_next(request)
    finally:
        request.state.db.close()
    return response
```

```
def get_db(request: Request):
    return request.state.db
```

This approach also limits the amount of refactoring needed because you are still using `get_db` as a dependancy.

DurandA:
> @DurandA we have been doing some more testing on this and have found that if you set up the DB using middleware you don't have this issue.

Thank you for sharing your solution. IMHO, this is clean enough to be put in production while waiting for an update on this. The only downside I see is that a session is established regardless of whether a DB session is required by the request.

brian-montgomery:
>  The only downside I see is that a session is established regardless of whether a DB session is required by the request.
@DurandA: You can easily mitigate that concern by making a wrapper class. Here is one example:

```
class LazyDBSessionContextManager:
    """
    Lazy instantiating database session designed to be use with middleware.
    Will not invoke a database session until used as a context manager.
    """

    __slots__ = ('session', '_session_cls', '_reentries')

    def __init__(self, session=None, session_cls=Session):
        """
        Can pass in an instance of a session or a mock class for testing
        """
        self.session = session
        self._session_cls = session_cls
        self._reentries = 0

    def __enter__(self):
        print("session", self.session)
        if (self.session is None):
            self.session = self._session_cls()
            self.session.begin()
        else:
            self._reentries += 1

        return self.session

    def __exit__(self, exc_type, exc_value, _):
        if (self._reentries):
            self._reentries -= 1
        else:
            try:
                if self.session and self.session.in_transaction():
                    if (exc_type is None):
                        self.session.commit()
                    else:
                        self.session.rollback()
            except Exception as e:
                # Shouldn't need this, but just in case something
                # goes wrong in the commit or rollback.
                raise e from exc_value  
            finally:
                self.session.close()
                self.session = None

    def __repr__(self):
        return f"{self.__class__.__name__}({self.session!r}, {self._session_cls!r})"
```

This both keeps track of your database session instantiation, as well as providing reentries, transaction handling and proper closing. You would use it very similarly to the one @Barnyclark mentioned:
```
@app.middleware("http")
async def db_session_middleware(request: Request, call_next):
    try:
        request.state.db = LazyDBSessionContextManager()
        response = await call_next(request)
    except:
        return Response("Internal server error", status_code=500)
``` 

``` 
def get_db(request: Request):
    with request.state.db as session:
        return session
```

DurandA:
> The only downside I see is that a session is established regardless of whether a DB session is required by the request.

Hi @brian-montgomery. You provided an interesting solution to this problem in a comment here but it somehow disappeared. Was there any particular issue with your proposal?

YuriiMotov:
We continue to discuss the mentioned problem (about dependency caching) caused by this PR.
Could you take a look and give your opinion?
[https://github.com/tiangolo/fastapi/discussions/6024#discussioncomment-8541913](https://github.com/tiangolo/fastapi/discussions/6024#discussioncomment-8541913)

@laggardkernel , @qmorek , @nyte-owl , @Barnyclark

PR #2954, opened 2021-03-15
WIP: [Docs] - Fix OpenAPI version in extending openapi

Leletir:
Hi,

Here is a small fix on the documentation.
As shown in this issue: tiangolo#1457 if we use openapi `2.5.0` as mentioned in the doc today we have a rendering problem. We need to use at least openapi `3.0.0` in order to get the Swagger UI displayed.

Have a good day.

PS: Thanks again for this framework ! It's just 👌


tiangolo:
Hey there! Thanks for the interest. This `version` is the version of your app, not of the OpenAPI specification. That one is `openapi_version`.

Given that I'll close this PR, but thanks for the interest! :coffee:

PR #2958, opened 2021-03-16
sparkles: Support custom openapi coroutine

Kludex:


Kludex:
There's really no need for this.

PR #2963, opened 2021-03-19
Swap light mode and dark mode toggle tooltip on docs

jcf-dev:
The current one shows "switch to light mode" while and light mode and "switch to dark mode" while on dark mode. This commit will correct it.


tiangolo:
Hey there! Thanks for the interest @jcf-dev! :cake: 

I think this was solved at some point in another PR. Could you check if it's working well for you now?

Otherwise, please make a new PR starting from the current main branch to avoid the conflicts. Given that, I'll close this one now. Thank you! :cookie: :coffee:

PR #2968, opened 2021-03-20
Translated Turkish.

sezginibis:
But it's not all over.


ycd:
We already have a PR for `index.md` https://github.com/tiangolo/fastapi/pull/1908

See: https://github.com/tiangolo/fastapi/issues/1892 for future translations.

tiangolo:
Thanks for your effort @sezginibis! :rocket: 

I would prefer to keep https://github.com/tiangolo/fastapi/pull/1908 as it was first. Nevertheless, it still needs a couple of approvals from people that understand the language. So it would be of great help if you could review that translation.

Given that, I'm gonna close this one, but thanks for your interest!

PR #3001, opened 2021-03-27
📌 Pin SQLAlchemy range for tests, as it doesn't use SemVer

tiangolo:
📌 Pin SQLAlchemy range for tests, as it doesn't use SemVer and I forgot (or didn't know?) that when I pinned it the first time.

This only affects FastAPI internal tests... and by accident, [Pydantic tests](https://github.com/samuelcolvin/pydantic/pull/2584). :grimacing: :see_no_evil:


Kludex:
`databases` already solved this issue, discussions: https://github.com/tiangolo/fastapi/issues/2996 and https://github.com/encode/databases/pull/299

tiangolo:
Thanks @Kludex for the heads up! I was convinced FastAPI was also affected as it has a direct dependency on SQLAlchemy and it had a too-open range (given SQLAlchemy doesn't follow SemVer). But now I realize that probably none of the code in FastAPI itself was really affected by the new SQLAlchemy release. :thinking:

Issue #1708, opened 2020-07-13
Is there a way to hide some fields from OpenAPI specs?


mclate:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to one of:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

Here's a self-contained, [minimal, reproducible, example](https://stackoverflow.com/help/minimal-reproducible-example) with my use case:

```Python
from fastapi import FastAPI

app = FastAPI()


@app.get("/")
def read_root(hidden_param=Query(None, description="This parameter should not be visible in OpenApi specs")):
    if hidden_param:
        return {"Hello": "Master!"}
    else:
        return {"Hello": "World"}
```

### Description

I'd like to have some request parameters (query or body ones) to be available in the python code, but to be hidden from the OpenAPI specs. Wonder if it is possible to do?

tiangolo:
Hmm, interesting...

I guess we could have a parameter `include_in_schema` for `Query`, `Body`, etc, something like:

```Python
from fastapi import FastAPI, Query, Body
from pydantic import BaseModel

app = FastAPI()


@app.get("/")
def read_root(hidden_param: str = Query(None, include_in_schema=False)):
    if hidden_param:
        return {"Hello": "Master!"}
    else:
        return {"Hello": "World"}


class Item(BaseModel):
    name: str
    price: float


@app.post("/items/")
def create_item(item: Item = Body(None, include_in_schema=False)):
    return item
```

---

[...]

PR #3012, opened 2021-03-29
Add include_in_schema parameter to headers to allow exclusion from OpenAPI spec

astraldawn:
Partially addresses #1708

Have the same use case as described by @mnorrsken 

```
We have a header with auth info thats always sent from our API manager to our backend. 
I want to consume this header in my API but I don't want it to show up in the docs.
```


astraldawn:
Closing in favour of #3144

tiangolo:
Thanks for coming back and closing the issue 👍

PR #3016, opened 2021-03-30
Fixed use of parameters in serialize_response

nanocfw:
Fixed use of parameters in serialize_response method when calling jsonable_encoder


tiangolo:
Thanks! Could you add a test for this?

PR #3033, opened 2021-04-01
👷 Only trigger build-docs CI in case of doc updates

Kludex:
:kissing:


tiangolo:
Oh, nice! :sunglasses: :tada: 

Thanks @Kludex! :cake:

PR #3038, opened 2021-04-03
⬆️ Upgrade OpenAPI from 3.0.2 to 3.1.0

Kludex:
Following [this document](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0).

## Changes
- Bump version (and check if the links are valid on the docs)
- Support `examples`

## Observations
Nothing to do on `exclusiveMinimum` and `exclusiveMaximum`, as we were already compliant based on the JSON Schema version that Pydantic uses.

Update: I got blocked by https://github.com/swagger-api/swagger-ui/issues/5891. The latest swagger doesn't support OpenAPI 3.1.0 yet. 
image

### DIFF
--- a/docs/en/docs/advanced/additional-responses.md
+++ b/docs/en/docs/advanced/additional-responses.md
@@ -36,7 +36,7 @@ For example, to declare another response with a status code `404` and a Pydantic
     **FastAPI** will take the Pydantic model from there, generate the `JSON Schema`, and put it in the correct place.
 
     The correct place is:
-    
+
     * In the key `content`, that has as value another JSON object (`dict`) that contains:
         * A key with the media type, e.g. `application/json`, that contains as value another JSON object, that contains:
             * A key `schema`, that has as the value the JSON Schema from the model, here's the correct place.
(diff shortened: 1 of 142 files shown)


a-akimov:
@Kludex, this is terrific! 

In general, migration to OpenAPI 3.1.0 should give us a lot of improvements (like full compatibility with JSON Schema) and fixing the problem with `exclusiveMinimum` and `exclusiveMaximum` (https://github.com/tiangolo/fastapi/issues/240 - because of it, currently generated OpenAPI files are invalid and cause errors).

However, it seems that SwaggerUI will not support OpenAPI 3.1.0 anytime soon (at least not clear from https://github.com/swagger-api/swagger-ui/issues/5891). Do you consider supporting OpenAPI 3.1.0 regardless SwaggerUI, and simply relying on the fact that Redoc already supports 3.1.0?

There are already a lot of tools supporting 3.1.0 - https://openapi.tools/ - and in my opinion the slow speed of SwaggerUI development shouldn't affect that much the development speed of FastAPI.

Alternatvely, of course, we can focus on fixing at least https://github.com/tiangolo/fastapi/issues/240 to make OpenAPI files valid for 3.0.3 and usable with the ecosystem of other tools (like Postman, Stoplight, Spectral, etc.).

Kludex:
> Do you consider supporting OpenAPI 3.1.0 regardless SwaggerUI, and simply relying on the fact that Redoc already supports 3.1.0?

Well... Yeah, generating a valid schema is something I'm interested in, as there are tools/services like https://readme.com/ that need a valid schema. Also, having it for ReDoc, and wait for https://github.com/swagger-api/swagger-ui/issues/5891 on Swagger's side looks fine.

In any case, I'll not perform further work without @tiangolo's input.

raphaelauv (approved):

thatguysimon:
@Kludex @a-akimov Following the [latest update](https://github.com/swagger-api/swagger-ui/issues/5891#issuecomment-1172274595) from SmartBear, it looks like it will take many more months before we see OpenAPI 3.1 support in Swagger UI.
In the meantime, maybe it's worth having an opt-in option for developers to allow FastAPI to generate OpenAPI 3.1 (even though it will break Swagger UI)? 🤷🏻‍♂️

mrlubos:
Hey everyone, any updates on this?

Kludex:
No.

acnebs:
I think FastAPI should probably push ahead with supporting OpenAPI 3.1. The spec was finalized 2 years ago at this point, so it seems unclear when if ever Swagger UI will support OAI 3.1.

Can we not do both? Allow people to choose if they want a 3.0 or 3.1 spec generated (or both!) with the knowledge that opting for 3.1 means no Swagger UI (but ReDoc has compat so I don't think that should be a dealbreaker).

Stopping progress in the name of a dep that doesn't seem interested in progress hardly seems like the right move.

Kludex:
FastAPI will only support OpenAPI 3.1 after pydantic v2.

PR #3051, opened 2021-04-07
🌐 Fix Chinese translation of Tutorial - Query Parameters, remove obsolete content

louis70109:
## Description
The contents in [EN page](https://fastapi.tiangolo.com/tutorial/query-params/) do not exist.
So remove them to fit EN page.


tiangolo:
Good catch, thank you! :rocket: :nerd_face:

PR #3058, opened 2021-04-08
added missing info to install 'aiofiles' when using 'FileResponse'

maces:
Just noticed that using `FileResponse` needs `aiofiles` to be installed: https://github.com/encode/starlette/blob/master/starlette/responses.py#L247
Added a hint in the docs in the PR.


tiangolo:
Thank you! This was a good PR! :rocket: 

Nevertheless, as FastAPI and Starlette are now based on AnyIO, `aiofiles` is no longer needed. :tada: 

Given that I'll close this one one, but thanks for your effort! :cake:

PR #3068, opened 2021-04-10
🔧 Replace Flit by Poetry

Kludex:
Given your message at https://github.com/tiangolo/fastapi/pull/1555#issuecomment-643678310:

> On the other hand, I'm thinking of migrating from Flit to Poetry, that way I could set up Dependabot to make it send PRs for each new version.

You said "thinking", but I went ahead, and prepared this PR in order to enable you to add [dependabot](https://dependabot.com/) or [renovate](https://docs.renovatebot.com/).

I didn't get your confirmation, so I'll understand if this is closed. :kissing: :+1: 

P.S.: This PR has 2.2k lines, but 2.1k are from `poetry.lock` :sweat_smile: 

## Plus

In case you want to enable [renovate](), you'll need to go to [GitHub Apps](https://github.com/apps/renovate) and configure the bot for FastAPI. Besides, I've also prepared a basic [renovate](https://docs.renovatebot.com/) config compatible with gitemoji: https://github.com/Kludex/fastapi/pull/1 . But in case you want to check more configuration possibilities, just click [here](https://docs.renovatebot.com/configuration-options/).

Oh, and that PR already shows you the real scenario on what PRs will be opened in case you configure this bot. But as a preview:
image

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -16,10 +16,10 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: "3.7"
-      - name: Install Flit
-        run: python3.7 -m pip install flit
+      - name: Install Poetry
+        run: python3.7 -m pip install poetry
       - name: Install docs extras
-        run: python3.7 -m flit install --extras doc
+        run: python3.7 -m poetry install --extras doc
       - name: Install Material for MkDocs Insiders
         if: github.event.pull_request.head.repo.fork == false
         run: pip install git+https://${{ secrets.ACTIONS_TOKEN }}@github.com/squidfunk/mkdocs-material-insiders.git
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -24,7 +24,7 @@ jobs:
         if: github.event.pull_request.head.repo.fork == false
         run: pip install git+https://${{ secrets.ACTIONS_TOKEN }}@github.com/squidfunk/mkdocs-material-insiders.git
       - name: Build Docs
-        run: python3.7 ./scripts/docs.py build-all
+        run: poetry run python3.7 ./scripts/docs.py build-all
       - name: Zip docs
         run: bash ./scripts/zip-docs.sh
       - uses: actions/upload-artifact@v2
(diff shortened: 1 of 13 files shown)


pyproject.toml
```diff
@@ -31,68 +35,104 @@ classifiers = [
...
-requires-python = ">=3.6"
 
-[tool.flit.metadata.urls]
-Documentation = "https://fastapi.tiangolo.com/"
+[tool.poetry.dependencies]
+python = "^3.6"
+starlette = "0.13.6"
+pydantic = "^1.0.0"
+
+requests = { version = ">=2.24.0,<3.0.0", extras = ["all"] }
```
hukkinj1, line 44: I think this should be
```suggestion
requests = { version = ">=2.24.0,<3.0.0", optional = true }
```

And the same change for all optional dependencies in `[tool.poetry.dependencies]` and `[tool.poetry.dev-dependencies]`

.github/workflows/build-docs.yml
```diff
@@ -16,15 +16,15 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: "3.7"
-      - name: Install Flit
-        run: python3.7 -m pip install flit
+      - name: Install Poetry
+        run: python3.7 -m pip install poetry
```
hukkinj1, line 20: I recommend pinning Poetry version here to get truly reproducible builds, and avoid issues if/when lockfile format changes or similar issues.

pyproject.toml
hukkinj1: To clarify: in the current state this line would mean requiring `requests[all]` but we dont want that. We want to install requests given `fastapi[all]`

Kludex: Thanks. I'll apply later. :)

.github/workflows/build-docs.yml
Kludex: Here I'm not sure, because I followed the same idea that was already in place.

pyproject.toml
Kludex: Applied.

Kludex:
everybody is going with hatchling ;)

PR #3069, opened 2021-04-11
Adding url: How to build a newsletter using FastAPI

duarteocarmo:


ycd (changes requested):
Can you please move this to the end? There is FCFS (First Come First Serve) rule

Smlep (approved):

PR #2606, opened 2021-01-05
🐛 Fix `jsonable_encoder` using `include` and `exclude` parameters for non-Pydantic objects

xaviml:
This fixes #2594. I added more few cases with include and exclude parameters, which were missing. Now the parameters are consistent with the way that it works with Pydantic.

When I first added the new tests in `test_jsonable_encoder.py` (without changing the codebase), these were the errors I got:
```
================================= test session starts =================================
platform linux -- Python 3.8.5, pytest-5.4.3, py-1.10.0, pluggy-0.13.1
rootdir: /home/xaviml/work/fastapi
plugins: cov-2.10.0, asyncio-0.14.0
collected 15 items                                                                    

tests/test_jsonable_encoder.py FFF............                                  [100%]

====================================== FAILURES =======================================
__________________________________ test_encode_dict ___________________________________

    def test_encode_dict():
        pet = {"name": "Firulais", "owner": {"name": "Foo"}}
        assert jsonable_encoder(pet) == {"name": "Firulais", "owner": {"name": "Foo"}}
>       assert jsonable_encoder(pet, include={"name"}) == {"name": "Firulais"}
E       AssertionError: assert {'name': 'Fir...name': 'Foo'}} == {'name': 'Firulais'}
E         Omitting 1 identical items, use -vv to show
E         Left contains 1 more item:
E         {'owner': {'name': 'Foo'}}
E         Use -v to get the full diff

tests/test_jsonable_encoder.py:99: AssertionError
__________________________________ test_encode_class __________________________________

    def test_encode_class():
        person = Person(name="Foo")
        pet = Pet(owner=person, name="Firulais")
        assert jsonable_encoder(pet) == {"name": "Firulais", "owner": {"name": "Foo"}}
>       assert jsonable_encoder(pet, include={"name"}) == {"name": "Firulais"}
E       AssertionError: assert {'name': 'Fir...name': 'Foo'}} == {'name': 'Firulais'}
E         Omitting 1 identical items, use -vv to show
E         Left contains 1 more item:
E         {'owner': {'name': 'Foo'}}
E         Use -v to get the full diff

tests/test_jsonable_encoder.py:112: AssertionError
________________________________ test_encode_dictable _________________________________

    def test_encode_dictable():
        person = DictablePerson(name="Foo")
        pet = DictablePet(owner=person, name="Firulais")
        assert jsonable_encoder(pet) == {"name": "Firulais", "owner": {"name": "Foo"}}
[...]

Issue #2594, opened 2021-01-03
`jsonable_encoder` function not working as expected with different type of variables


xaviml:
### First check

* [X] I added a very descriptive title to this issue.
* [X] I used the GitHub search to find a similar issue and didn't find it.
* [X] I searched the FastAPI documentation, with the integrated search.
* [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [X] I already read and followed all the tutorial in the docs and didn't find an answer.
* [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [X] After submitting this, I commit to one of:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

Here's a self-contained, [minimal, reproducible, example](https://stackoverflow.com/help/minimal-reproducible-example) with my use case:

#### Example 1 (BaseModel)

```python
import pytest
from attr import dataclass
from fastapi.encoders import jsonable_encoder
from pydantic.main import BaseModel

class MyBaseModel(BaseModel):
    foo: int
    bar: int


@pytest.mark.parametrize(
    "obj, include, exclude, expected",
    [
        (MyBaseModel(foo=1, bar=2), None, None, {"foo": 1, "bar": 2}),  # Passed
        (MyBaseModel(foo=1, bar=2), {}, {}, {}),  # Passed
        (MyBaseModel(foo=1, bar=2), {"foo"}, {}, {"foo": 1}),  # Passed
        (MyBaseModel(foo=1, bar=2), {}, {"foo"}, {}),  # Passed
        (MyBaseModel(foo=1, bar=2), {"foo"}, None, {"foo": 1}),  # Passed
        (MyBaseModel(foo=1, bar=2), None, {"foo"}, {"bar": 2}),  # Passed
    ],
)
def test_jsonable_encoder_include_exclude_base_model(obj, include, exclude, expected):
    assert jsonable_encoder(obj, include=include, exclude=exclude) == expected
```

#### Example 2 (Python dictionary)
```python
import pytest
[...]

PR #3078, opened 2021-04-12
fix jsonable_encoder exclude include encode model obj invalid

shan-guo:
Hi, When i use the `jsonable_encoder` expect covert model to dict obj，use the `exclude` or `include` options  was invalid

```
data = jsonable_encoder(queryset, exclude={'id'})
```

or

```
data = jsonable_encoder(queryset, include={'name'})
```

The `id` always in the data

So i created this  pr to solve this problem, Maybe there are better ways to solve this problem

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -56,6 +56,8 @@ def jsonable_encoder(
             obj_dict = obj_dict["__root__"]
         return jsonable_encoder(
             obj_dict,
+            include=include,
+            exclude=exclude,
             exclude_none=exclude_none,
             exclude_defaults=exclude_defaults,
             custom_encoder=encoder,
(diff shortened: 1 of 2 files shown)


xaviml:
For someone reviewing this PR: this was raised in #2594, and the open PR for the fix is #2606. There is more than just including the "include" and "exclude".

As of this fix @shan-guo, did you considered nested objects? With your proposed solution, the include and exclude will be recursively passed to the children and they will also be included and excluded, and I am not sure if that is the expected behaviour. For this reason in #2606, I just included "include" and "exclude" to the last "jsonable_encoder". Pydantic solves this recursivity problem by allowing a nested dictionary for "include" and "exclude", but this is not solved either in this PR and in the #2606. It is a different problem to solve.

For the first "include" and "exclude" you passed (BaseModel) is not needed since it is passed to the BaseModel.dict function just above, so it is redundant.

For the second one (dict), it is not needed for what I mentioned before, it will include and exclude children, which I don't think it is exptected.

For the third one (default) is needed since it will be needed for the dict case. This is why this was included in #2606.

Furthermore, the #2606 fixes the problem with include and exclude for dict objects to work like they work in Pydantic BaseModel.

You can see the added tests in #2606 for the use cases.

Regards,
Xavi M.

shan-guo:
Sorry i ignore the Pydantic work way ,Thank you answer

PR #3127, opened 2021-04-24
Add a tip for using dataclasses for "Using Classes as Dependencies" tutorial

### LINKED ISSUE #3114: Recommend using dataclass in "Classes as dependencies" Docs
### First check

## Idea

Currently in the Classes as Dependencies docs: https://fastapi.tiangolo.com/tutorial/dependencies/classes-as-dependencies/#classes-as-dependencies_1

The sample class is:

```py
class CommonQueryParams:
    def __init__(self, q: Optional[str] = None, skip: int = 0, limit: int = 100):
        self.q = q
        self.skip = skip
        self.limit = limit
```

I believe that it would be better to recommend the cleaner use of a `dataclass`, to remove boilerplate that can be error prone. New users are likely to simply copy this style directly and not realize they can use a cleaner syntax. It also matches more similarly to Pydantic usage elsewhere in FastAPI:

```py

@dataclass
class CommonQueryParams:
    q: Optional[str] = None
    skip: int = 0
    limit: int = 100

@app.get("/items/")
async def read_items(commons: CommonQueryParams = Depends(CommonQueryParams)):
    response = {}
    if commons.q:
        response.update({"q": commons.q})
    items = fake_items_db[commons.skip : commons.skip + commons.limit]
    response.update({"items": items})
    return response
```

This is somewhat discussed in the closed issue #318 from two years ago.

johnthagen:
Closes #3114

### DIFF
--- a/docs/en/docs/tutorial/dependencies/classes-as-dependencies.md
+++ b/docs/en/docs/tutorial/dependencies/classes-as-dependencies.md
@@ -55,6 +55,17 @@ class Cat:
 fluffy = Cat(name="Mr Fluffy")
 ```
 
+!!! tip
+    If you are using Python 3.7+ or install the PyPI backport of [`dataclasses`](https://pypi.org/project/dataclasses/) for Python 3.6, you can use a `dataclass` as a dependency to significantly reduce the boilerplate involved:
+    
+    ```Python
+    from dataclasses import dataclass
+    
+    @dataclass
+    class Cat:
+        name: str
+    ```
+
 In this case, `fluffy` is an instance of the class `Cat`.
 
 And to create `fluffy`, you are "calling" `Cat`.


johnthagen:
@tiangolo Would you mind reviewing this PR?

tiangolo:
Thanks for the interest and effort! But I prefer to leave it as is, because dataclasses require assuming some knowledge about dataclasses and how they work. And they would only work for some specific cases. They wouldn't necessarily take things like `Query()`, `Body()` as default values, or at least the behavior is not certain. And it would require explaining how they work.

I prefer to keep the explanation simple and pure, that makes it obvious that the thing that makes it callable is what would be used to get the information (in this case the `__init__()`. But in a dataclass that is hidden, so it would require explaining that it's hidden but created automatically, and then it would require explaining that because the `__init__()` is autocreated, some things would work and some wouldn't.

And although the code is shorter, the explanation and the cognitive load would be higher, explaining what's everything that is happening underneath.

So I'll close this one, but thanks for the effort! :cake:

PR #3131, opened 2021-04-25
📝 Update docs with pip install calls when using extras with brackets, use quotes for compatibility with Zsh

tomwei7:
Square brackets in zsh have special meanings. Copy and run `pip install uvicorn[standard]` in zsh will get error.


Kludex:
It will not.

waynerv:
I had this problem before and now I finally know how to solve it.zsh uses [square brackets for globbing / pattern matching](http://zsh.sourceforge.net/Guide/zshguide05.html#l137).

I've never seen it (add quotation mark) in any other documentation, but add a notice about this problem should be nice.

tomwei7:
@waynerv [python-language-server](https://github.com/palantir/python-language-server)  add quotation mark on install command `pip install 'python-language-server[yapf]'` in its documentation

Kludex:
Just adding the quotes as @tomarv2 suggested seems fine.

Why I don't have this problem here tho? :sweat_smile:

tomwei7:
run `zsh -f` start zsh without any config that can reproduce the problem. `setopt nonomatch` in zsh will avoid this problem.

Kludex (approved):
I see... Ok. :)

waynerv (approved):

tiangolo:
Thanks @tomwei7! :cake: 

I updated it to include all the other calls of `pip install` that were using packages with extras.

PR #3138, opened 2021-04-29
♻ Change a `dict()` for `{}` in `fastapi/utils.py`

ShahriyarR:
In general `dict()` call is considered slower than direct `{}`.
The internal representations are also different, there is no extra function call while constructing the map with `{}`
```
>>> import dis
>>> dis.dis('{}')
  1           0 BUILD_MAP                0
              2 RETURN_VALUE
>>> dis.dis('dict()')
  1           0 LOAD_NAME                0 (dict)
              2 CALL_FUNCTION            0
              4 RETURN_VALUE
```

There are bunch of similar cases in FastAPI source code, I will send PRs accordingly.

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -76,7 +76,7 @@ def create_cloned_field(
 ) -> ModelField:
     # _cloned_types has already cloned types, to support recursive models
     if cloned_types is None:
-        cloned_types = dict()
+        cloned_types = {}
     original_type = field.type_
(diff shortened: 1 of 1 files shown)


Smlep (approved):

cikay (commented):
The commit message should be "Replace empty dict literal with built-in dict function"

You can also specifiy the reason[ in the body of the commit](https://robertcooper.me/post/git-commit-messages) which should be performance optimization

yezz123 (approved):

BilalAlpaslan (approved):

PR #3139, opened 2021-04-29
♻ Refactor dict value extraction to minimize key lookups `fastapi/utils.py`

ShahriyarR:
There are 3 extra key/value lookups for update_dict, it can be eliminated by unpacking value in for loop directly.


ycd (approved):

tiangolo:
Nice, thank you! 😎🚀

PR #3140, opened 2021-04-29
Replaced if clauses with if expression for more clarity

ShahriyarR:
Replaced several `if`s with more Pythonic expressions.


tiangolo:
Thanks for your interest! I think the code is fine as it is, and I consider it's more clear with the current structure, I guess it depends on personal taste. :shrug: Maybe later when doing performance profiling I find that some of these things improve performance I'll add them, but for now, I'll pass. Nevertheless, thanks for your effort! :coffee:

PR #3141, opened 2021-04-29
Removed redundant variable assignments; replaced with direct return

ShahriyarR:
The anti-pattern spotted as returning variable after assignment, which is just redundant.
Instead, return it directly.


tiangolo:
Thanks, I don't think it's strictly necessary and having the value in a variable before returning helps sometimes while debugging, to set a breakpoint before returning. So, for now, I'll pass on this one. Maybe in a future, when I'm doing some benchmarks and trying to optimize the code I'll optimize some places like this, but for now, I think it's fine as it is, so I'll pass on this one. But thanks for your interest! :coffee:

PR #3142, opened 2021-04-29
♻️ Move duplicated code portion to a static method in the `APIKeyBase` super class

ShahriyarR:
The code block below was used 3 times:
```
        if not api_key:
            if auto_error:
                raise HTTPException(
                )
            else:
                return None
        return api_key
 ```
 
 I have moved this portion to the `staticmethod` into the super class as it will be shared by inheritance to all 3 subclasses.

### CODE REVIEW

File: fastapi/security/api_key.py [hunk 1]
```diff
@@ -8,7 +8,16 @@
...
 class APIKeyBase(SecurityBase):
-    pass
+    @staticmethod
+    def check_api_key(api_key: Optional[str], auto_error: bool) -> Optional[str]:
+        if not api_key:
+            if auto_error:
+                raise HTTPException(
+                    status_code=HTTP_403_FORBIDDEN, detail="Not authenticated"
+                )
+            else:
```
cikay, line 18: Remove else statement.

cikay (changes requested):

fastapi/security/api_key.py
ShahriyarR: Done

yezz123 (approved):

BilalAlpaslan (approved):

patrick91 (approved):
Nice!

svlandeg:
@patrick91 : could you have another look at this PR, as it's not succeeding the linting step? 🙏

svlandeg (approved):

tiangolo (approved):
Nice, thank you! :rocket: 

And thanks for the reviews and help! :coffee:

PR #3144, opened 2021-04-29
✨ Allow hiding from OpenAPI (and Swagger UI) `Query`, `Cookie`, `Header`, and `Path` parameters

astraldawn:
Add parameter `include_in_schema` to partially address #1708 for subclasses of `Param`: `Cookie`, `Header`, `Path` and `Query`

Implements the `Param` portion of https://github.com/tiangolo/fastapi/issues/1708#issuecomment-663366079, relevant parts of code sample reproduced below:

> ```python
> from fastapi import FastAPI, Query
> 
> app = FastAPI()
> 
> 
> @app.get("/")
> def read_root(hidden_param: str = Query(None, include_in_schema=False)):
>     if hidden_param:
>         return {"Hello": "Master!"}
>     else:
>         return {"Hello": "World"}
> ```


jordantshaw:
Any idea when this will be merged, and what version it will be released with?

astraldawn:
@jordantshaw no idea but I have updated the request with a code sample from the original issue. Hope that will speed up the review.

HIRANO-Satoshi:
This is a useful PR for me, too.

marlenekoh:
Please merge this!! This is a super useful feature! 😊😊

mamicahencarnacion:
hello, any updates on this? this will be a really helpful additiona feature 👍

jordantshaw:
@tiangolo  Any updates on when this will merged?  Looks like a lot of people would find this very helpful.

sirtux:
@tiangolo Ping? I would also be very delighted to see this feature :3

phuong:
@tiangolo Any update for this, really helpful

yezz123 (approved):
This is a great Pull request LGTM!
@Kludex @tiangolo this one need to be approved 🌟

BlueMagma2:
Joining the gang, I would find this very useful too

fastapi/openapi/utils.py
```diff
@@ -91,23 +91,24 @@ def get_openapi_operation_parameters(
...
-        if field_info.description:
-            parameter["description"] = field_info.description
-        if field_info.examples:
-            parameter["examples"] = jsonable_encoder(field_info.examples)
-        elif field_info.example != Undefined:
-            parameter["example"] = jsonable_encoder(field_info.example)
-        if field_info.deprecated:
-            parameter["deprecated"] = field_info.deprecated
-        parameters.append(parameter)
+        if field_info.include_in_schema:
```
tokikanno, line 94: Maybe 

```python
if not field_info.include_in_schema:
    continue
```

This will decrease 1 indent level to the rest of codes in loop.
Also people can skip finding the `else` block of this if while reading these codes

fastapi/params.py
```diff
@@ -93,6 +96,7 @@ def __init__(
             deprecated=deprecated,
             example=example,
             examples=examples,
+            include_in_schema=include_in_schema,
```
tokikanno, line 99: [discussion] don't think should/could hide path parameter, the path parameter already shown during route declaration. For example:


```python
@app.get('/user/{uid}')
def get_user_info(uid: str = Path(...)):
    pass
```

Though you could hide the validation info of this path parameter, but people who have access to document already know there is a path parameter called `uid`.

tiangolo: Agreed @astraldawn, it's fine to leave it as is. I don't see why someone would want to hide the path param, but at the same time if they can hide the others, they should be able to hide this as well.

fastapi/openapi/utils.py
tiangolo: Thanks for the discussion! Yep, agreed it makes sense to switch it around. I just did in a commit on top. Thanks!

tiangolo:
Awesome, thank you @astraldawn! :coffee: 

Thanks everyone for the discussion! :cake: 

I updated a bit the tests and added source examples (and tests) for Python 3.10.

I was thinking that maybe it could make sense to support also body/form-data, but the implementation would be non-trivial, or it would have to be a per-route `include_body_in_schema` or something like that. But at the same time, I don't easily see a use case where hiding the entire body would be wanted without just hiding the entire route, so let's just have this for now and I'll see if anyone finds later a use case that would benefit from hiding the body before implementing that. 🤓

This is available in FastAPI version `0.73.0` (released in a couple of hours). 🎉

mangiahsan:
This is a fantastic feature addition—thank you for this! 🎉

I wanted to share an alternative approach I’ve used in a similar context where I needed to exclude certain fields from the OpenAPI schema for a cleaner API documentation. My use case involved building an OAuth2-like authentication schema where fields like `grant_type` and `access_token` are essential for internal processing but irrelevant for API consumers to see in the documentation.

Here’s the implementation I used, leveraging `pydantic.json_schema.SkipJsonSchema`:
`from pydantic import BaseModel, Field, root_validator

class UserSignInSchema(BaseModel):
    grant_type: SkipJsonSchema[GrantType] = GrantType.PASSWORD
    access_token: SkipJsonSchema[str] = None
    username: str = Field(..., alias="email")
    password: str

    @root_validator(pre=True)
    def validate_fields(cls, values):
        grant_type = values.get('grant_type')

        if grant_type == GrantType.PASSWORD:
            values['access_token'] = 'Null'  
        elif grant_type in (GrantType.GOOGLE, GrantType.FACEBOOK):
            values['username'] = 'null@null.com'
            values['password'] = 'Null1234567'

        return values
`

In this approach, `SkipJsonSchema` ensures that fields like `grant_type` and `access_token` are excluded from the OpenAPI spec, allowing us to focus the documentation on the fields users need to interact with, like username and password.

I’m curious about how this compares to the` include_in_schema` parameter introduced here. Both approaches achieve field exclusion but might cater to slightly different needs. For instance, does include_in_schema offer better control in dynamically configurable scenarios, or are there limitations when using `SkipJsonSchema`?

Thank you again for this valuable addition—it’s going to make handling such use cases a lot more intuitive! 😊

PR #1757, opened 2020-07-21
✨ Add `description` parameter to all the security scheme classes, e.g. `APIKeyQuery(name="key", description="A very cool API key")`

hylkepostma:
My suggestion is to add a description parameter to the `__init__` methods of `APIKeyQuery(SecurityBase):`, `APIKeyHeader(SecurityBase):` and `APIKeyCookie(SecurityBase):` and use it to initialize the `APIKey` object. This allows for a description in the `securitySchemes` objects of the `openapi.json` and therefore in the Swagger and Redoc documentation.

`class SecurityBase(BaseModel):` already has an attribute `description: Optional[str] = None`.
`class APIKey(SecurityBase):` inherits from SecurityBase.

The classes `APIKeyQuery(SecurityBase):`, `APIKeyHeader(SecurityBase):` and `APIKeyCookie(SecurityBase):` initialize an `APIKey` object in their `__init__` method. They initialize it with some kwargs and the `name` parameter, but currently without a description parameter.

Example: using `APIKeyQuery(name="X-Auth-Key", auto_error=False, description="My description")` would look likt this in Redoc:
image w:400

PR #3153, opened 2021-05-01
Add description attribute to security schemes

### LINKED ISSUE #2840: Specify securitySchemes description in OpenAPI
### First check

### Description

According to the OpenAPI scheme, there is an optional `description` field that can be specified for any authentication type. For example, here it is for `OAuth2`: https://github.com/OAI/OpenAPI-Specification/blob/42a9e3d4eddade52363a5c4fac852e80681c2fe5/schemas/v3.0/schema.yaml#L831-L832.

If we look at FastAPI's own code, we also find it: https://github.com/tiangolo/fastapi/blob/33be5fc8baa02bae949e1986f1821dae8f84f487/fastapi/openapi/models.py#L258.

Unfortunately, there is no easy way to fill this field for pre-defined security dependencies because none of them accept the `description` parameter.

### Example

```Python
from fastapi.security import OAuth2PasswordBearer

AUTH_DESCRIPTION = """
To get a fully-functional access token, you need to pass the `scope`
and `audience` parameters as described below..."""

# what I want to do, but it doesn't work this way
auth = OAuth2PasswordBearer(tokenUrl="...", description=AUTH_DESCRIPTION)

# what I have to do at the moment as a workaround
auth = OAuth2PasswordBearer(tokenUrl="...")
auth.model.description = AUTH_DESCRIPTION
```

What I want to achieve as a result is the following OpenAPI spec fragment:

```yaml
  securitySchemes:
    OAuth2PasswordBearer:
      description: '

        To get a fully-functional access token, you need to pass the `scope`

        and `audience` parameters as described below...'
      flows:
        password:
          scopes: {}
          tokenUrl: '...'
      type: oauth2
```

All child classes of `SecurityBase` (i.e. `HTTPBase`, `OAuth2`, `OpenIdConnect`,  etc) should accept `description` as an argument in the `__init__` method and properly configure its `self.model` attribute using this description.

### Environment

* OS: macOS
* FastAPI Version: `0.62.0`
* Python version: `3.8.6`

### Additional context

Note: there is also [a bug in Swagger UI](https://github.com/swagger-api/swagger-ui/issues/5230) that prevents the description from being shown up. But we at least need to support it on the FastAPI side.

Itay4:
fixes https://github.com/tiangolo/fastapi/issues/2840

### DIFF
--- a/fastapi/security/http.py
+++ b/fastapi/security/http.py
@@ -24,9 +24,14 @@ class HTTPAuthorizationCredentials(BaseModel):
 
 class HTTPBase(SecurityBase):
     def __init__(
-        self, *, scheme: str, scheme_name: Optional[str] = None, auto_error: bool = True
+        self,
+        *,
+        scheme: str,
+        scheme_name: Optional[str] = None,
+        auto_error: bool = True,
+        description: Optional[str] = None,
     ):
-        self.model = HTTPBaseModel(scheme=scheme)
+        self.model = HTTPBaseModel(scheme=scheme, description=description)
         self.scheme_name = scheme_name or self.__class__.__name__
         self.auto_error = auto_error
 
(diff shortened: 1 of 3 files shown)


codespearhead (approved):

codespearhead:
This was already handled by #9658 and #1757 .

@Itay4 Can you close this issue and #9013 ?

PR #3171, opened 2021-05-04
🌐 Fix Korean translation for `docs/ko/docs/tutorial/first-steps.md`

SueNaEunYang:
Hello, I carefully pulled a request on the already uploaded page. 
While I was quickly going through the docs to study, I found out a few trivial points to suggest. Please take a look if there's any part you like.

Here are what I suggested :
- corrected a trivial typo
- changed the tone that suit #2017 
- suggested some alternatives that seem to be closer to the author's original intent.

Plus, I have one thing to ask for your opinion. I personally think that putting the original terminoloy right after the translated word inside the parentheses seems to be a good way to reduce confusion for Korean readers. I am asking this because I saw some of your reviews on other PRs removing those English versions in parentheses, and still having the one within this file (동작(operation) <- this one).
If there's any translation guide on this matter, I would be very grateful if you let me know :)

Thanks always for your contributions and efforts!!


0417taehyun:
I think it would be better to remove the supplementary explanation by using `()` like you said, `(operation)` because if we accept that rules there are so many things to write.

Also, we follow the rules `격칙체` (check in #2017 ) so it is right to use `-기 바립니다.` or `-십시오` instead of `-세요.`.

I agree to translate the docs to be closer to the author's original intent, however if you want to discuss it, I think it would be better to add your opinion in #3167 !

tiangolo:
As this is quite old and it seems it's not necessarily agreed to follow with these changes, I'll pass on this one, but thanks for the effort @SueNaEunYang! :coffee:

PR #3173, opened 2021-05-04
use empty set instead of empty list

yaroverg:
Replaces empty list with empty set.


tiangolo:
Good catch, thank you! :mag: 

Sadly, as this branch was removed, I can't update it or commit on top. Most of these changes were already in place, but there was one missing, so I replicated it here: https://github.com/tiangolo/fastapi/pull/4899

Thanks for your help! :coffee:

PR #3233, opened 2021-05-15
Allow override of `BackgroundTasks` implementation.

nickswebsite:
It would be nice to override the default implementation of `BackgroundTasks`.

My particular use case came about when looking for a way to easily push a number of background tasks to Celery in a fairly large code base. Splicing in a bunch of `return JSONResponse(background=...)` everywhere would be rather clunky.

This is a feature that has been asked for here:

* https://github.com/tiangolo/fastapi/issues/805

And would be helpful for the use case here:

* https://github.com/tiangolo/fastapi/issues/3133
* https://github.com/tiangolo/fastapi/issues/3115
* https://github.com/tiangolo/fastapi/issues/1637

My approach was to look for `app.dependency_overrides[BackgroundTasks]` for a callable that would instantiate a `BackgroundTasks` instance. If it isn't overridden, we default the usual `BackgroundTasks` class.

Currently this is only overridable at the `FastAPI` app level and the constructor must not take any arguments.


YuriiMotov (changes requested):
@nickswebsite, thanks for your interest and your efforts!

I agree that it might be useful to provide the ability to specify the actual class for Background Tasks to be used, but `dependency_overrides` is not the right way to specify it (`dependency_overrides` is for testing, not for production).

Why don't we just add the `background_tasks_class` parameter to `FastAPI()` and use it?

Another option (maybe even better) is to use the actual type from the function signature to inject the background tasks instance. We can add the check to ensure that only one type is used across all dependencies of this particular endpoint.

What do you think?

PR #3241, opened 2021-05-17
📝 Add external link to article: How-to deploy FastAPI app to Heroku

jarmos-san:
Related to https://github.com/tiangolo/fastapi/issues/802 & will act as a central place for anyone looking to deploy their FastAPI application to Heroku.


ArcLightSlavik (changes requested):
We have an unwritten rule where new articles go to the bottom of the list, not the top.

Also I would argue to change uvicorn to gunicorn with uvicorn workers, but wouldn't consider it as a hard requirement.

ArcLightSlavik (approved):

PR #3246, opened 2021-05-18
🔧 Swap light/dark theme button icon

eddsalkield:
Just a very minor PR for the FastAPI docs that swaps the light/dark theme icons around.  Currently the dark mode icon has the bulb filled (on) in the dark theme, and outlined (off) in the light theme.

Although not particularly important, most UI elements display the current state of affairs, rather than represent what will change when activated - for example, buttons are depressed when selected, not depressed prior to selection.  This PR simply aims to make the UX more consistent.


frinzekt (approved):
I came here to look for this as well

PR #3276, opened 2021-05-22
feat: add contributor over time graph to README

guoqqqi:
Hi, community!

To better present how our community grows, we develop a tool to show contributors growing history on [https://github.com/api7/contributor-graph](https://github.com/api7/contributor-graph). Since we found it helpful, we think maybe if it could help some other community.

## WHAT IT IS

Basically, it just shows the contributors growth over time, just like the stargazers over time on the README. It would be the same with stars that we would update the graph each day, so the link would always present the real-time data. There is some other stuff to play around with if you would like to give it a try~

image

## HOW IT WORKS 

We use Github API to get all commits, try to find the “Github way” to filter commits so the result data would be similar to Github, and then get the first commit time of each user.

Don't hesitate to tell us if there is a better place to present this graph other than this, or there are some other worries or other features you would like to have~🍻

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,6 +454,10 @@ Used by FastAPI / Starlette:
 
 You can install all of these with `pip install fastapi[all]`.
 
+## Contributor over time
+
+[![Contributor over time](https://contributor-graph-api.apiseven.com/contributors-svg?chart=contributorOverTime&repo=tiangolo/fastapi)](https://www.apiseven.com/en/contributor-graph?chart=contributorOverTime&repo=tiangolo/fastapi)
+
 ## License
 
 This project is licensed under the terms of the MIT license.


ArcLightSlavik (changes requested):
Adding a link to a completely different service not controlled by fastapi is a horrible idea.

The graph itself can be useful, but it would need to be hosted on a service controlled by fastapi (tiangolo) or calculated monthly, similar to FastAPI people.

tiangolo:
Hey, thanks for the suggestion, but I prefer to keep it as it is. If someone is interested in seeing a graph like that they can probably use your service directly or check out the contributors section: https://github.com/tiangolo/fastapi/graphs/contributors

But thanks for your interest! I'll close this now.

PR #2640, opened 2021-01-12
🐛 Fix support for prefix on APIRouter WebSockets

Kludex:
Related to #2639

PR #3280, opened 2021-05-24
Fix: websocket route path not inherited  prefix value

AngusWG:
websocket route path not inherited  prefix value

In my case:
```
router = APIRouter(prefix = '/asset')
@router.websocket_route("/ws")
async def ws_func(ws: WebSocket):
    ....
app.include_router(router)
```
`ws_func` path should be  `/asset/ws` .
but `ws_func` path is `/ws` .
I found starlette WebSocketRoute not have prefix  attribute.

we need override add_websocket_route method for add router prefix .

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -49,6 +49,7 @@ from starlette.routing import (
     get_name,
     request_response,
     websocket_session,
+    WebSocketRoute,
 )
 from starlette.status import WS_1008_POLICY_VIOLATION
 from starlette.types import ASGIApp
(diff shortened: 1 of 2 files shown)


Kludex:
I guess that issue is solved by https://github.com/tiangolo/fastapi/pull/2640/

AngusWG:
> I guess that issue should also be solved by #2640

Thank you for answer.
Please release as soon as possible.  :two_men_holding_hands:

AngusWG:
https://github.com/tiangolo/fastapi/blob/0.65.1/fastapi/routing.py#L583


@Kludex 

It seems that the latest version does not have https://github.com/tiangolo/fastapi/pull/2640/commits/8843fe8009389123dbe398de5034c78583ea40e0  code

Kludex:
I mean, it was not merged, but it solves the issue :sweat_smile:

AngusWG:
Hope to be release soon. :joy:

genagurbanguliyev:
use `websocket` not `websocket_route`
try this:
```python
router = APIRouter(prefix = '/asset')
@router.websocket("/ws")
async def ws_func(ws: WebSocket):
    ....

```

This works for me

PR #3285, opened 2021-05-24
✏ Fix typo about file path in `docs/en/docs/tutorial/bigger-applications.md`

HolyDorus:
The file path has been changed to the correct one because the ```app/routers/dependencies.py``` is an imaginary file.
We should replace ```app/routers/dependencies.py``` to ```app/dependencies.py``` in this two cases.

image


HolyDorus:
Why don't you accept the change?

PR #3305, opened 2021-05-28
🛠 Native JWT support

k4black:
### Relevance
About half of the modern internet uses raw JWT (the other half uses OAuth2, which is already supported).  
I think that in such a modern library as `fastapi`, native support of JWT will be a big advantage.

### Existed solutions 
Only 1 external lib: [github:fastapi-jwt-auth (192 stars)](https://github.com/IndominusByte/fastapi-jwt-auth/) - an flask-jwt-auth customization. 

The following drawbacks are significant:
1. poorly supported (last updated in 2020)
2. depends on the original library
3. DOES NOT match the `fastapi` security style (`Security`/`Depends` function args wrapper)
4. no `openapi` specification  generation

### Existed Issues and Pull Requests:
* Searching 'jwt' or 'json web tokens' - 'No results matched' (except docs example)

---

### Features 

- [x] bearer/cookie token place
- [x] access/refresh tokens 
- [x] time expiring tokens 
- [x] `auto_error=False` support
- [x] `openapi` specification generation
- [ ] fresh token demand
- [ ] token revocation
- [ ] cookies name selecting
- [ ] docs examples (after approval)

### Requrements 
`python-jose` for tokens interaction and `pytest-mock` for testing time-dependent functions.
 

### Usage example 
All available use cases can be found in the test files. The simplest example is provided below:

```python
app = FastAPI()

access_security = JwtAccessBearer(secret_key="secret_key")

@app.post("/auth")
def auth():
    subject = {"username": "username", "role": "user"}
    access_token = access_security.create_access_token(subject=subject)
    return {"access_token": access_token, "refresh_token": None}

@app.get("/users/me")
def read_current_user(credentials: JwtAuthorizationCredentials = Security(access_security)):
    return {"username": credentials["username"], "role": credentials["role"]}
```

---

Any suggestions for improvement are gladly discussed!


acutaia:
@k4black  in security docs @tiangolo  in the past used  Pyjwt but later he moved to python-jose  PR  [#1610] so I think we should consider using python-jose for your pull request 🤔

k4black:
@acutaia 
I saw this warning in docs, however, there were no arguments in favour of a younger library. Initially I used `python-jose` 😃, but migrate to `PyJWT`, as the most popular library according to [jwt.io :libraries:python](https://jwt.io/).  
Although I agree that to comply with the general style, there may be a need to rewrite it with `python-jose`. I'll work on it

acutaia:
@k4black 
Maybe it's better use it in an agnostic way without being tied directly to PyJWT or   python-jose  So the end user could use the one he/she prefers. If you want i can help you 👷🏽

k4black:
@tiangolo Can you approve or decline this PR, please

fastapi/security/jwt.py
```diff
@@ -0,0 +1,444 @@
...
+        return None
+
+    def _get_payload(
+        self, bearer: Optional[HTTPBearer], cookie: Optional[APIKeyCookie]
+    ) -> Optional[Dict[str, Any]]:
+        refresh_token = self._get_token(bearer, cookie)  # TODO: del function
+
+        # Check token exist
+        if refresh_token is None:
+            if not self.auto_error:
```
cikay, line 138: Do not reverse the condition add the raise keyword to the if block.

fastapi/security/jwt.py
```diff
@@ -0,0 +1,444 @@
...
+        )
+
+    def _get_credentials(
+        self,
+        bearer: Optional[JwtAuthBase.JwtAccessBearer],
+        cookie: Optional[JwtAuthBase.JwtAccessCookie],
+    ) -> Optional[JwtAuthorizationCredentials]:
+        payload = self._get_payload(bearer, cookie)
+
+        if payload:
```
cikay, line 249: This is where you reverse the condition the success case should be at the bottom. [It is called early return or fast fail.](https://szymonkrajewski.pl/why-should-you-return-early/)

JanBeelte:
Is there any chance that this PR or comparable functionality will be merged soon?
What are the next steps here @tiangolo @cikay @k4black @acutaia , I might be willing to take over some work.
Would be very handy to have jwt support out of the box!

and-sm:
@tiangolo, any update?

jpmckinney:
python-jose is unmaintained. Please do not use it.

nsteinmetz:
> python-jose is unmaintained. Please do not use it.

latest version of fastapi-jwt allows to use authlib instead of python-jose (still present as fallback)

Pandede:
Any update?

YuriiMotov (commented):
I think the problem with adding such native support for JWT to FastAPI is that in this case FastAPI will become responsible for this part. But security is really complex subject and it will add a lot of work and improve the maintenance burden.

There are quite good docs on this topic and it should be enough for newcomers to implement simple security for their app.

Also, it it were simple, there would be multiple third-party libraries that add such support. And I think this is the way to go (implement this as a third-party library).

tiangolo:
Thanks for the interest @k4black! 

So, JWT and OAuth2 are two separate things. OAuth2 is about the exchange of communication via HTTP to handle auth stuff and exchange of tokens, without specifying any format for the tokens.

JWT is a format for the tokens. Using OAuth2 it's quite common to exchange JWT tokens.

"Bearer" is a header format to transfer tokens. Normally, the 3 go together.

For example, if you check the security docs, they are about using Bearer tokens in JWT format using OAuth2.

Parsing and handling tokens is not really dependent of any web stuff, so I think it would be better in an external package.

This also adds a strong dependency on python-jose that I wouldn't want to take right now.

For now, I'll pass on this one, but thanks for the interest! :cake:

LiorA1:
I got here because I'm new to FastAPI, but not sure in my implementation of the access & refresh tokens..
Maybe its worth to include in the documentation, because its a common practice

PR #3313, opened 2021-05-31
🌐 Add Chinese translation for docs/tutorial/encoder.md 

jaystone776:
as title


tiangolo:
It seems this was translated at some point in another PR. So I'll close this one, but thanks for the effort! ☕

PR #3322, opened 2021-06-02
Add fastapi.middleware.sessions

br-kim:
get SessionMiddleware from starlette.middleware.sessions for convenience.

I'm learning python web framework using fastapi.
Thank you for providing a good Framework and Docs.

I'm adding it because I think it's missing.
It's simple, but I haven't done much pull request, so I'd appreciate it if you could check if it's the right pull request.


tiangolo:
Thanks for the interest! But the best way to handle sessions in FastAPI would be with dependencies and the security/auth utilities, so I'll pass on this one, but thanks! ☕

PR #3343, opened 2021-06-08
fix: async partial function will not awaited in Depends

lbhsot:
Hi, it seems wrong when using async partial function in Depends. Here's an example code:

```
def partial_func(q: str):
    async def inner(data):
        return data

    return partial(inner, q + " extras")

@app.get("/")
def test(data: str = Depends(partial_func("test"))):
    return {"data": data}
```

In `test` method, data here will be a coroutine object.

I just add `asyncio.iscoroutinefunction` to check if depends function is awaitable in this pr, I have no idea whether this is the right solution. Or maybe I need to change `is_coroutine_callable` method?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .mypy_cache
 .vscode
 __pycache__
+.python-version
 .pytest_cache
 htmlcov
 dist
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -546,6 +546,8 @@ async def solve_dependencies(
             )
         elif is_coroutine_callable(call):
             solved = await call(**sub_values)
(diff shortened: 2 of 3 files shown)


lbhsot:
I found a workaround for this issue: Using [A callable instance](https://fastapi.tiangolo.com/advanced/advanced-dependencies/#a-callable-instance) instead of partial function.

```
class PartialCls:
    def __init__(self, q: str):
        self.q = q

    async def __call__(self):
        return self.q + "extras"


@app.get("/")
def test(data: str = Depends(PartialCls("test"))):
    return {"data": data}
```

So I don't know whether we should use partial function or not, but since there is a solution here, I'll close this pr for now.
If we need use async partial function, maybe open another issue to track this bug(or feature?)

PR #3345, opened 2021-06-08
⬆️ Bump Uvicorn max range to 0.15.0

Kludex:
EDIT: We probably want to add `asgiref` as dependency and use its types.


jedidiahhorne:
awesome, thanks guys!

musicinmybrain:
I have no influence over the FastAPI project, but:

- Consider bumping the maximum uvicorn dependency in the `dev` extra to match the `all` extra.
- Also, uvicorn 0.15.0 is now released. Can it be safely allowed too?

Kludex:
Thanks for the review @musicinmybrain :)

tiangolo:
Great, thanks @Kludex! :cake: 

For now, I would prefer not to add a dependency on `asgiref` only for the types (`asgiref` as a bunch of other tools apart from the types), and as ASGI is just dicts everywhere I would think we should be able to get the annotations right.

Thanks everyone else for the comments too! :coffee:

Kludex:
The thing is that Starlette is too generic on its ASGI types, and uvicorn is strict now.

In any case, FastAPI shouldn't care about asgiref, I'm going to take care on Starlette.

tiangolo:
> In any case, FastAPI shouldn't care about asgiref, I'm going to take care on Starlette.

Woohoo! Nice :tada: :sunglasses: :nerd_face:

PR #3346, opened 2021-06-09
✏ Re-word and combine 2 paragraphs into 1 for readability

jalvaradosegura:
👋 Hi everyone! 

This is a very little change that I think it improves the readability of the introduction text (even though I think we don't even need the last 2 paragraphs of the introduction, because the second paragraph already says the point of the others 2)

### Reasons of the change
- **You can read the first 3 paragraphs independently** and they give you a good idea about what each of them is talking about.
- But **with the last paragraph you can't do that** in my opinion. Because it is tightly related with the paragraph 3. That's why I combined paragraph 3 and 4.
- I removed the word `diseñado` from paragraph 3, because I think we don't actually need it. Paragraph 2 already says how the Tutorial is designed (gradually...) and the main focus of paragraph 3 is the functionally the Tutorial may have in the future for the user.

**Old version:**
```
Este tutorial te muestra cómo usar FastAPI con la mayoría de sus características paso a paso.

Cada sección se basa gradualmente en las anteriores, pero está estructurada en temas separados, así puedes ir directamente a cualquier tema en concreto para resolver tus necesidades específicas sobre la API.

También está diseñado para funcionar como una referencia futura.

Para que puedas volver y ver exactamente lo que necesitas.
```

**2 last paragraphs combined**
```
Funciona también como una referencia futura, para que puedas volver y ver exactamente lo que necesitas.
```

Please let me know what you think 🙏

### DIFF
--- a/docs/es/docs/tutorial/index.md
+++ b/docs/es/docs/tutorial/index.md
@@ -4,9 +4,7 @@ Este tutorial te muestra cómo usar **FastAPI** con la mayoría de sus caracter�
 
 Cada sección se basa gradualmente en las anteriores, pero está estructurada en temas separados, así puedes ir directamente a cualquier tema en concreto para resolver tus necesidades específicas sobre la API.
 
-También está diseñado para funcionar como una referencia futura.
-
-Para que puedas volver y ver exactamente lo que necesitas.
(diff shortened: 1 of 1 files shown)


tiangolo:
Yep, that makes sense. Thank you @jalvaradosegura! :coffee: :cake:

PR #3355, opened 2021-06-10
Use Decimal for certain OpenAPI params to avoid losing precision

the-moonwitch:
Currently, the schema parameters `multipleOf`, `maximum`, `exclusiveMaximum`, `minimum`, and `exclusiveMinimum` are forcibly cast to `float`.

This causes a couple problems with OpenAPI:
* From [the docs on `multipleOf`](https://swagger.io/docs/specification/data-models/data-types/#multipleOf): 
  > `multipleOf` may be used with floating-point numbers, but in practice this can be unreliable due to the limited precision or floating point math.
* It loses precision with large numbers. For example, if we wanted to have a field that accepts a 64-bit integer (in the range `-9223372036854775808` to `9223372036854775807`), casting that to float converts it to +/-`9.223372036854776e+18`, resulting in +/-`9223372036854776000` when cast back to int. This causes a significant difference between what the developer intends to be a valid value, and what actually gets validated against (and displayed in error messages).

Declaring those fields as `Decimal` avoids losing precision through unwanted `float` conversions.


tiangolo:
I'm seeing that this would only work to override the generated OpenAPI, but it wouldn't alter the internal implementation in Pydantic and the actual validation, right?

In that case, I feel it would be dangerous to make it look like some validation is done, but it's actually no exactly that. :thinking:

PR #3367, opened 2021-06-12
Tip to install email-validator package to use email validation

manlix:
Fix for issue in documentation example code #3368

### DIFF
--- a/docs/en/docs/tutorial/response-model.md
+++ b/docs/en/docs/tutorial/response-model.md
@@ -31,6 +31,20 @@ But most importantly:
 !!! note "Technical Details"
     The response model is declared in this parameter instead of as a function return type annotation, because the path function may not actually return that response model but rather return a `dict`, database object or some other model, and then use the `response_model` to perform the field limiting and serialization.
 
+## Install `email-validator`
+
+First you need to install `email-validator`:
+
+<div class="termy">
+
+```console
+$ pip install email-validator
+
+---> 100%
+```
+
+</div>
+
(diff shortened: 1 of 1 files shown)


docs/en/docs/tutorial/response-model.md
```diff
@@ -33,6 +33,11 @@ But most importantly:
 
 ## Return the same input data
 
+!!! info
+    To use email validator (`EmailStr` type), first install <a href="https://github.com/JoshData/python-email-validator" class="external-link" target="_blank">`email-validator`</a> package.
+
+    E.g. `pip install email-validator`.
+
```
Bluenix2, line 40: Personally I rather see this as a note under the code worded more like:
```suggestion
!!! note
    `EmailStr` uses <a href="https://github.com/JoshData/python-email-validator" class="external-link" target="_blank">`email-validator`</a>, you will have to install this package to use `EmailStr`.

```

Bluenix2 (changes requested):

Bluenix2 (changes requested):
This is misleading, now the user is let on to belive that to use `response_model` you need to install `email-validator`.

vrslev:
I don’t think the instructions belong here. Pydantic uses `email-validator`. Imagine what it would be like if FastAPI would follow and add all remarks from its dependencies 😃

tiangolo:
Thank you! This was solved recently in another PR.

Given that, I'll close this one now. But thanks for the effort! :cake:

PR #3370, opened 2021-06-14
Add default for `default` argument in Path

Bluenix2:
This pull request means that the following the code from [Path Parameters and Numeric Validations - FastAPI](https://fastapi.tiangolo.com/tutorial/path-params-numeric-validations/#declare-metadata) can omit its pointless default argument:
```py
from typing import Optional

from fastapi import FastAPI, Path, Query

app = FastAPI()


@app.get("/items/{item_id}")
async def read_items(
    # item_id: int = Path(..., title="The ID of the item to get"),
    item_id: int = Path(title="The ID of the item to get"),
    q: Optional[str] = Query(None, alias="item-query"),
):
    if q:
        results.update({"q": q})
    return results
```

As explained by the note under it, and by looking at the source, we can make out that this argument never had any meaning.
> A path parameter is always required as it has to be part of the path.
> 
> So, you should declare it with `...` to mark it as required.
> 
> Nevertheless, even if you declared it with None or set a default value, it would not affect anything, it would still be always required.

Looking at the source (simplified as to not take up too much space):
```py
class Path(Param):
    def __init__(
        self,
        default: Any,
        *,
        **extra: Any,
    ):
        super().__init__(
            ...,
            **extra,
        )
```

### DIFF
--- a/docs/en/docs/tutorial/path-params-numeric-validations.md
+++ b/docs/en/docs/tutorial/path-params-numeric-validations.md
@@ -12,19 +12,15 @@ First, import `Path` from `fastapi`:
 
 ## Declare metadata
 
-You can declare all the same parameters as for `Query`.
+You can declare all the same parameters as for `Query` except for the default. A path parameter is always required as it has to be part of the path. You should declare it with `...` to correctly mark it as required, for convenience FastAPI already does this if you leave it out.
 
-For example, to declare a `title` metadata value for the path parameter `item_id` you can type:
+As an example of what you can pass to `Path`, to declare a `title` metadata value for the path parameter `item_id` you can type:
 
 ```Python hl_lines="10"
 {!../../../docs_src/path_params_numeric_validations/tutorial001.py!}
 ```
 
 !!! note
-    A path parameter is always required as it has to be part of the path.
-    
-    So, you should declare it with `...` to mark it as required.
-
     Nevertheless, even if you declared it with `None` or set a default value, it would not affect anything, it would still be always required.
 
 ## Order the parameters as you need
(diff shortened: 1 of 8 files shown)


Bluenix2:
I have now also edited the documentation.

docs/en/docs/tutorial/path-params-numeric-validations.md
```diff
@@ -12,19 +12,15 @@ First, import `Path` from `fastapi`:
 
 ## Declare metadata
 
-You can declare all the same parameters as for `Query`.
+You can declare all the same parameters as for `Query` except for the default. A path parameter is always required as it has to be part of the path. You should declare it with `...` to correctly mark it as required, for convenience FastAPI already does this if you leave it out.
```
ArcLightSlavik, line 15: Maybe just leave out the "You should declare it with ..." part?

ArcLightSlavik (approved):

docs/en/docs/tutorial/path-params-numeric-validations.md
Bluenix2: It still has meaning, we're explaining that compared to Query parameters, Path parameters are always required and that's why you should declare it with `...` to mark it as such (in relation to Query parameters). Then adding that since this is the case, for Path parameters FastAPI defaults to `...` if you leave it out.

Bluenix2:
> Can't we completely remove the `default` parameter from the `Path` object's signature? It is a bit misleading/confusing as it is. Path has no default value, period.

I believe the reason it is there is because it has to implement implement and replicate the Param class types, which does have a `default` parameter.

It is also a breaking change to remove now, what would be your suggestion?

tiangolo:
Thanks for the effort @Bluenix2! :cake: 

And thanks for the conversation and discussion @ArcLightSlavik and @frerikandriessen. :coffee: 

This was in another PR, now no parameter needs a default value unless one is actually defined. If it's not defined, it's taken as required. This works for `Query`, `Path`, `Cookie`, etc.

Given that, I'll close this PR now, but thanks for the interest! :cookie:

PR #3380, opened 2021-06-16
Fix attempt to relative import in 'Testing' code examples

manlix:
Fix for issue in documentation example code #3381

### DIFF
--- a/docs_src/app_testing/test_main.py
+++ b/docs_src/app_testing/test_main.py
@@ -1,6 +1,6 @@
 from fastapi.testclient import TestClient
 
-from .main import app
+from main import app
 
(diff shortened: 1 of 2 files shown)


ArcLightSlavik (changes requested):
We use relative explicit relative import everywhere in the docs, so we need to either change it everywhere or add a diagram like here: <https://fastapi.tiangolo.com/tutorial/sql-databases/#file-structure>

Also would require fixing a lot of tests.

tiangolo:
Thanks for the effort! I handled it here: https://github.com/tiangolo/fastapi/pull/5302

I'll close this one now. :cake:

PR #3384, opened 2021-06-18
Added support to x-enum-varnames and x-enum-descriptions

nanocfw:


fastapi/openapi/models.py
```diff
@@ -116,6 +116,8 @@ class SchemaBase(BaseModel):
     externalDocs: Optional[ExternalDocumentation] = None
     example: Optional[Any] = None
     deprecated: Optional[bool] = None
+    x_enum_varnames: Optional[List[str]] = Field(None, alias="x-enum-varnames")
+    x_enum_descriptions: Optional[List[str]] = Field(None, alias="x-enum-descriptions")
```
Bluenix2, line 120: Why was an alias needed?

Kludex:
This is not in the [OpenAPI specs](https://swagger.io/specification/).

fastapi/openapi/models.py
nanocfw: Because I need to export this information to OpenAPI. 
Currently I have in my enum the following code to generate the "x-enum-varnames" automatically, but without this variable/alias this information is ignored. 
The alias is required due to the variable name, just like the @Kludex commented.

My code:
```python
    @classmethod
    def __modify_schema__(cls, field_schema):
        field_schema.update({"x-enum-varnames": [item.name for item in cast(Iterable[Enum], enum)]})
```

Bluenix2: Ah yes I see now, the attribute is `x_enum_varnames` and the alias points to `x-enum-varnames` (with a dash, not an underscore).

Mause:
Is this still required now that https://github.com/tiangolo/fastapi/pull/1429 is merged?

tiangolo:
Sorry for the delay in handling this! 😅

As @Mause said, this is now covered by that other feature. So I'll close this one, but thanks! :coffee:

PR #3406, opened 2021-06-24
fix-3405: fix broken Starlette Middleware URL

DavidBord:


ArcLightSlavik:
@DavidBord This was fixed by Starlette, could you close this?

PR #3440, opened 2021-07-01
Used default value if empty string in request

dmitrijkir:
Some problems with default value

For example, we have route like this

```python
@router.get("/all")
async def get_all(
    itemId: Optional[int] = Query(None),
    page: int = Query(1),
    size: int = Query(10),
):
    return {"itemId": itemId, "page": page, "size": size}
```

and I try to send request with empty value `GET /all?itemId`. I hope, I'll get default value from `Query(None)`, but in `starlette` into request empty value always is empty string. If I send this request `GET /all?itemId`, I'll get validation error, because we compare string with int.

I decided the best way equate empty string to empty value.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -69,6 +69,7 @@ sequence_shape_to_type = {
     SHAPE_TUPLE_ELLIPSIS: list,
 }
 
+EmptyString = ""
 
 multipart_not_installed_error = (
     'Form data requires "python-multipart" to be installed. \n'
@@ -614,7 +615,7 @@ def request_params_to_args(
         assert isinstance(
             field_info, params.Param
         ), "Params must be subclasses of Param"
-        if value is None:
+        if value is None or value == EmptyString:
             if field.required:
                 errors.append(
                     ErrorWrapper(


vrslev:
I think what you’re trying to do is `/all?itemId=`.

kevindaffaarr:
I think I want to upvote this issue and add more cases. Sometimes the user/frontend sends us the backend API with these parameters:
- `itemId=`
- `itemId=null` (because frontend mostly javascript)
- `itemId=undefined` (because frontend mostly javascript)

It would be nice if FastAPI filtered those parameters to be a default parameter or `None`

kevindaffaarr (changes requested):
Line 618
`if value in (None, "", "null", "undefined"):`

ThirVondukr:
> Line 618 `if value in (None, "", "null", "undefined"):`

"", "null" and "undefined" are valid strings and shouldn't be cast to None or raise an error when used with a required parameter.



> Some problems with default value
> 
> For example, we have route like this
> 
> ```python
> @router.get("/all")
> async def get_all(
>     itemId: Optional[int] = Query(None),
>     page: int = Query(1),
>     size: int = Query(10),
> ):
>     return {"itemId": itemId, "page": page, "size": size}
> ```
> 
> and I try to send request with empty value `GET /all?itemId`. I hope, I'll get default value from `Query(None)`, but in `starlette` into request empty value always is empty string. If I send this request `GET /all?itemId`, I'll get validation error, because we compare string with int.
> 
> I decided the best way equate empty string to empty value.

You might want to just filter empty values out in your frontend.

tiangolo:
Hey there! Thanks for the interest! :cake: 

The problem here is that query parameters are actually more of a concept than a type. In the end, they are part of the URL, which is just a string, it doesn't have any notion of types, so all query parameters are naturally strings, and if they don't have a value, they are by default just strings. For example, it's not really possible to explicitly send a query parameter with a value of `None`. So, a query parameter that is provided but doesn't have a value, like `?itemId=` would indeed be a string (an empty string). And then a query parameter without the equal sign, as in `?itemId` would be something a bit weird, as those parameters are expected to be key-value pairs (https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL#parameters).

Before deciding that parameters that don't have an equal sign should be treated as `None`, it would probably be better to find a reference in a spec (e.g. an IETF RFC, a W3C spec) specifying that.

If that was the case, then it would probably make sense to fix this in Starlette itself.

---

Now, if that was the case, this implementation would have a problem, that a URL with `?itemId=""`, which is actually passing the required value (it's just that it's an empty string, but the "required" value is given) would be taken as invalid for "not passing a required value".

It's possible to do the check with `None` because if the parameter is ever in the URL, the value will never be `None`, it would only be `None` when the parameter is not sent in the URL.

---

Given that, I'll close this PR now. But if you think this still should be done and you find some references to some specs that define that, please create a new PR (and some tests please).

Thanks for the effort! :coffee:

kevindaffaarr:
Thanks for not only a comment, but the whole knowledge. Keep a good work, Sir

PR #3443, opened 2021-07-01
Add Swagger UI persistAuthorization configuration parameter

otsuka:
When developing APIs that requires authentication using Swagger UI, if you reload the browser window of Swagger UI to reflect a change in the code, the authentication is lost and you have to log in again, which is troublesome.

Swagger UI has a `persistAuthorization` parameter that, if enabled, will persist the authentication and keep you logged in even if you reload or close the browser.
[Swagger Documentation](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/)

In this PR, an argument `swagger_ui_persist_auth` is added to `FastAPI` constructor to set the `persistAuthorization` parameter. The default value is False, but setting this argument to True will also set the `persistAuthorization` parameter of Swagger UI to true.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -46,6 +46,7 @@ class FastAPI(Starlette):
         redoc_url: Optional[str] = "/redoc",
         swagger_ui_oauth2_redirect_url: Optional[str] = "/docs/oauth2-redirect",
         swagger_ui_init_oauth: Optional[Dict[str, Any]] = None,
+        swagger_ui_persist_auth: bool = False,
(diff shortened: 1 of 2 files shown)


elrik75:
I think the PR https://github.com/tiangolo/fastapi/pull/2568 is more generic to give any parameters to swagger UI.

RayBB:
@otsuka   I think since #2568 is merged and released you can probably close this ticket!

otsuka:
@RayBB 
Thank you for your comment. I closed the PR.

PR #3456, opened 2021-07-03
♻ Assume request bodies contain JSON when no Content-Type header is provided

tiangolo:
♻ Assume request bodies contain JSON when no Content-Type header is provided

This is related to https://github.com/tiangolo/fastapi/pull/2118 and the CVE https://github.com/tiangolo/fastapi/security/advisories/GHSA-8h2j-cgx8-6xv7 about CSRF.

Some clients already assume that request bodies are read as JSON, even when no content-type header is set: https://github.com/tiangolo/fastapi/pull/2118#issuecomment-858155064

And the risk of CSRF only applies for specific Content-Type headers, so, no Content-Type header is already protected from CORS preflights: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -184,7 +184,9 @@ def get_request_handler(
                     if body_bytes:
                         json_body: Any = Undefined
                         content_type_value = request.headers.get("content-type")
-                        if content_type_value:
+                        if not content_type_value:
+                            json_body = await request.json()
+                        else:
                             message = email.message.Message()
                             message["content-type"] = content_type_value
                             if message.get_content_maintype() == "application":
--- a/tests/test_tutorial/test_body/test_tutorial001.py
+++ b/tests/test_tutorial/test_body/test_tutorial001.py
@@ -229,6 +229,20 @@ def test_geo_json():
     assert response.status_code == 200, response.text
 
 
+def test_no_content_type_is_json():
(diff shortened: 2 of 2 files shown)


LironSherk:
Hi,
I'm sending a request with no Content-Type header but got the same error before this fix.
when I put a breakpoint on the line 
` if not content_type_value`

it equals: 'text/plain;charset=UTF-8'

meaning something between from the client to this line of code adds defaults.

It was working before the upgrade to 0.66.3

falkben:
> Hi,
> I'm sending a request with no Content-Type header but got the same error before this fix.
> when I put a breakpoint on the line
> ` if not content_type_value`
> 
> it equals: 'text/plain;charset=UTF-8'
> 
> in node-fetch, if developer didn't provide Content-Type the default is : 'text/plain;charset=UTF-8'
> 
> so all node servers that sending JSON requests to fastapi without providing Content-Type will fail.

I think you should either change the default you are sending w/ `node-fetch` or make sure the content-type is specified on every request. Assuming json when the content type was sent as text/plain is not correct. CORS specifically checks text/plain, so to do so we'd be back in the same situation as https://github.com/tiangolo/fastapi/security/advisories/GHSA-8h2j-cgx8-6xv7.

PR #3462, opened 2021-07-05
✏ Fix typo in `body-nested-models.md`

jaystone776:
Typo in line 201, *where* should be *were*


tiangolo:
Good catch, thanks @jaystone776! :mag:

PR #3463, opened 2021-07-05
✏ Fix code highlighted line in: `body-nested-models.md`

jaystone776:


tiangolo:
Good catch with the highlighting too, thanks! :cake: :mag:

PR #3362, opened 2021-06-11
✏ Fix typo "might me" -> "might be" in `docs/en/docs/tutorial/schema-extra-example.md`

dbrakman:

PR #3465, opened 2021-07-05
fix typo line 91 in schema-extra-example.md

jaystone776:
line 91: "might me" should be "might be"


JuanBenitezDev:
Approved pull request with the same changes waiting to be merged #3362

tiangolo:
Good catch, thanks!

This was handled in https://github.com/tiangolo/fastapi/pull/3362, so I'll close this now. But thanks for the contribution! :coffee: :cake:

PR #3476, opened 2021-07-07
🌐 Update Chinese translation for `docs/python-types.md`

jaystone776:
also add new translation for optional and generics


tiangolo:
It seems this translation was updated at some point, could you please check if the changes are still needed in a new PR? For now I'll close this one, thanks! 🍰

PR #3484, opened 2021-07-07
🌐 Update Chinese translation for `docs/tutorial/body-nested-models.md`

jaystone776:
also fixed code highlight


tiangolo:
It seems this translation was updated at some point. Could you check if the changes are still needed? If that's the case, please create a new PR. For now, I'll close this one. Thank you! 🍰

PR #3485, opened 2021-07-07
🌐 Update Chinese translation for `docs/tutorial/schema-extra-example.md`

jaystone776:
also add translation for new technical detail


tiangolo:
It seems this translation was updated at some point, could you please check if there are still changes needed? If so, please create a new PR. For now, I'll close this one. Thank you! ☕

PR #3499, opened 2021-07-07
🌐 Update Chinese translation for `docs/tutorial/path-params-numeric-validations.md`

jaystone776:
…ns.md

as title


tiangolo:
It seems this was updated at some point. Could you please check if there are still changes needed? If that's the case, please create a new PR. For now, I'll close this one. Thank you! ☕

Issue #2057, opened 2020-09-15
Dependency Overrides in startup event gives TypeError: 'Depends' object is not callable


timbmg:
How can dependencies in the startup event be overwritten for testing?

Below a minimal example. When executing the test, `Model` is a `Depends` object and not a `MagicMock`. When putting the depends in normal endpoints it works just fine.

api.py
```python 
from my_package.model import Model
from fastapi import Depends, FastAPI
app = FastAPI()

def model_provider():
    return Model

model = None

@app.on_event("startup")
def startup(Model = Depends(model_provider)):
    global model
    model = Model()

@app.get("/prediction")
def predict(x):
    return model.predict(x)

```

tests/test_api.py
````python
from my_package import api
from unittest.mock import MagicMock
from fastapi.testclient import TestClient

def test_predict():
    model_mock = MagicMock()
    model_mock.predict = MagicMock(return_value=lambda x: x)
    api.app.dependency_overrides[api.model_provider] = lambda: model_mock
    with TestClient(api.app) as client:
        prediction = client.get("/prediction?x=1")
        assert prediction.text == "1"

```

wichert:
The documents seems to hint that you can only use `Depends` for request functions. Perhaps this could be extended to app events as well?

oxfn:
+1 this would be great to use dependencies in event handlers. Now it is Starlette events which know nothing about FastAPI DI

Faylixe:
+1, the DI is a very important design feature of FastAPI that should be usable in event handler IMO.

Unless already planned for future release, I would like to contribute to this, but this doesn't seems that easy to implement it in a clean way. By reading DI implementation currently used, I realize that it is deeply connected to request / response game, with no separation of concern allowing to just refactor `fastapi.dependencies.utils.solve_dependencies` function to handle an event only use case.

A quick intuition is that it should have a more granular `Dependent` / `Depends` hierarchy like having a base `Dependent` definition without any `Body`, `Form`, `Request`, `Response`, etc ... support and adding those to a child `RequestDependent` before to finally break solver function to support either both use case or smartly trigger errors when unsupported request deps are required in an "non request" context.

But maybe there is a simpler way ?

adriangb:
@Faylixe I took a look, and I agree with your assessment. It looks doable without breaking changes, but yeah a _lot_ of refactoring would be needed.

Issue #617, opened 2019-10-13
[FEATURE] Further develop startup and shutdown events


sm-Fifteen:
While the documentationn for FastAPI is in general extremely solid, there's a weakpoint that I feel hints at some underdevelopped feature within the framework, and that's [startup and shutdown events][1]. They are briefly mentionned (separately) with the startup event in particular being demonstrated like this :

```py
items = {}


@app.on_event("startup")
async def startup_event():
    items["foo"] = {"name": "Fighters"}
    items["bar"] = {"name": "Tenders"}


@app.get("/items/{item_id}")
async def read_items(item_id: str):
    return items[item_id]
```

...which could very well be written like this:

```py
items = {
    "foo": {"name": "Fighters"},
    "bar": {"name": "Tenders"}
}

@app.get("/items/{item_id}")
async def read_items(item_id: str):
    return items[item_id]
```

...and therefore makes the feature look useless. The example for `shutdown` instead uses logging as an example, which makes it look like this would be the primary purposes for those events, while in reality, it's not.

**Is your feature request related to a problem? Please describe.**
The problem is that, throughout the entire documentation, things like database connections are created in the global scope, at module import. While this would be fine in a regular Python application, this has a number of problems, especially with objects that have a side-effect outside the code itself, like database connections. To demonstrate this, I've made [a test structure that creates a lock file when initialized and deletes it when garbage collected][2].

Using it like this:
```py
from fastapi import FastAPI
from lock import FileLock

app = FastAPI()
lock = FileLock("fastapi")

@app.get("/")
async def root():
    return {"message": "Hello World"}
```

...does not work and the lock is not deleted before shutdown (I was actually expecting it to be closed properly, like SQLAlchemy does with its connections, but clearly there's a lot of extra magic going on with SQLAlchemy that I don't even come close to understanding). This is also extremely apparent when using the `--reload` option on Uvicorn, bcause the lock is *also* not released when the modules are reloaded, causing the import to fail and the server to crash. This would be one thing, but I've had a similar incident occur some time ago when, while developping in reload mode, I've actually managed to take up every connection on my PostgreSQL server because of that problem, since while SQLAlchemy is smart enough to cleanup on exit where my `FileLock` cannot, the same does not happen when hot-reloading code.

[...]

PR #3516, opened 2021-07-11
Add startup/shutdown dependencies and dependency caching lifespan control

adriangb:
## Issues this addresses
- #2057
- #617
- #2697 

To address these issues, this PR provides for more advanced dependency injection concepts.
In particular:
- Gives users the ability to set a cache scope so that values can be cached for the app's lifetime instead of just a single request.
- Give users control over when teardown is run for dependencies with yield. It can now be run during app shutdown, after the response is sent (current default) or right before the response is sent (#2697)

Finally, this PR gives lifespan events (startup/shutdown) the dependency injection.

These changes are presented together for cohesiveness (they aren't very useful individually), but they can easily be split into multiple smaller PRs for easier review.

Here's a small motivating example:

```python
from asyncpg import create_pool, Connection, Pool


class Config(BaseSettings):
    dsn: str


def get_config() -> Config:
    return Config()


async def get_connection_pool(cfg: Config = Depends(get_config)) -> Pool:
    async with create_pool(dsn=cfg.dsn) as pool:
        yield pool


async def get_connection(pool: Pool = Depends(get_connection_pool, lifetime="app", cache="app")) -> Connection:
    async with pool.acquire() as conn:
        yield conn


async def check_db_connection(conn: Connection = get_connection()) -> None:
    assert (await conn.execute("SELECT 1;")) == 1


app = FastAPI(on_startup=[check_db_connection])


@app.get("/")
async def root(conn: Connection = get_connection()) -> int:
    return await conn.execute("SELECT 2;")  # or some other real query

```


fastapi/applications.py
```diff
@@ -93,6 +94,11 @@ def __init__(
             [] if middleware is None else list(middleware)
         )
         self.middleware_stack: ASGIApp = self.build_middleware_stack()
+        
+        def clear_lifetime_dependencies():
+            self.lifetime_dependencies = {}
```
adriangb, line 99: This stuff I'm a bit unclear about. I'm pretty sure the desired behavior is indeed for these dependencies to be re-initialized on every app startup, but I'm not sure if there's any precedent in the codebase for something like this. This is what I came up with. Open to ideas.

fastapi/applications.py
```diff
@@ -93,6 +94,11 @@ def __init__(
...
+
+        self.on_event("shutdown")(clear_lifetime_dependencies)
```
adriangb, line 101: Clearing the dependencies at shutdown aught to help with garbage collection by releasing the refs.

adriangb:
@Faylixe @sm-Fifteen, based on your input in the linked issues, I'd appreciate if you could glance over this and give any feedback. Thanks!

adriangb:
> This even supports sync and async yield dependencies, so that you can easily do startup/shutdown handling

Realized this isn't working. Will think of how to fix it, I hope it's doable

fastapi/dependencies/models.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         security_scopes_param_name: Optional[str] = None,
         security_scopes: Optional[List[str]] = None,
         use_cache: bool = True,
+        lifetime: str = "request",
```
Mause, line 36: Why not add a `Literal["request", "app"]` here?

fastapi/applications.py
```diff
@@ -125,6 +131,16 @@ def __init__(
         self.openapi_schema: Optional[Dict[str, Any]] = None
         self.setup()
 
+    @property
+    def lifetime_dependencies(self) -> Dict[Tuple[Callable[..., Any], Tuple[str]], Any]:
```
Mause, line 135: Rather than repeating the same type annotation over and over, define an alias?

fastapi/applications.py
Mause, line 101:
```suggestion
        self.on_event("shutdown")(self.lifetime_dependencies.clear)
```

fastapi/applications.py
```diff
@@ -125,6 +131,16 @@ def __init__(
...
+        if not hasattr(self, "_lifetime_dependencies"):
+            self._lifetime_dependencies = {}
+        return self._lifetime_dependencies
+
+    @lifetime_dependencies.setter
+    def lifetime_dependencies(self, value: Dict[Tuple[Callable[..., Any], Tuple[str]], Any]) -> None:
```
Mause, line 141: What's the reason behind this getter and setter? Why not just assign the variable in the `__init__` method?

Mause (changes requested):

docs/en/docs/tutorial/dependencies/lifetime-dependencies.md
```diff
@@ -0,0 +1,47 @@
+# Lifetime dependencies
+
+FastAPI supports dependencies that get creted once, when the app starts, and are cached thereafter.
+This can be used, for example, to load your config from enviroment variables or initialize database connections.
+
+Let's say we have a `DataBase` object which usees a `.connect` context manager to establish a connection.
```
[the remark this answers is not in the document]
adriangb: I'm referring to a class name. Shouldn't it be camel cased?

fastapi/applications.py
adriangb: That sounds like a good idea.

fastapi/dependencies/models.py
adriangb: I made an alias `DependencyLifetime = Literal["request", "app"]`

fastapi/applications.py
adriangb: Probably works as well, what I have here is over complicated and doesn't fully work.

Really I'm just struggling to come up with a good way to deal with binding the lifetime dependency stack `__aenter__`/`__aexit__` to the app lifetime. Do you have any suggestions in that regard?

docs/en/docs/tutorial/dependencies/lifetime-dependencies.md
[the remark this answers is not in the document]
Mause: `Database` is one word, not two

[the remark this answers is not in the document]
adriangb: Fair enough, I'll apply the suggestion.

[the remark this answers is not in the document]
adriangb: Pushed the change. Thank you for the suggestion!

fastapi/applications.py
adriangb: Here is the alias I made, also the one referred to in https://github.com/tiangolo/fastapi/pull/3516#discussion_r667614049: 
https://github.com/tiangolo/fastapi/blob/cecd77fdd12ac58c66631e9abac25e8c5149eccc/fastapi/dependencies/models.py#L7-L11

fastapi/applications.py
adriangb: I completely reworked this part. I got rid of the setters/getters, and it is now handling lifespans as expected. However, I'm not convinced that creating 3 async closures in `__init__` and binding them to `on_start`/`on_shutdown` is the cleanest way to do this.

fastapi/params.py
```diff
@@ -1,5 +1,8 @@
 from enum import Enum
-from typing import Any, Callable, Dict, Optional, Sequence
+from typing import Any, Callable, Dict, Optional, Sequence, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from fastapi.dependencies.models import DependencyLifespan
```
adriangb, line 5: I was getting circular imports, I'm not via what path. Since these are just type hints, this solves it, but I'm open to moving the aliases or another solution.

sm-Fifteen:
I'm not personally sold on having the dependency's lifespan be specified on the caller's end instead of the dependency's end. On one hand, it means any dependency marked as a lifetime dep anywhere causes all of its transitive dependencies to be run and cached as lifetime deps at the same time, which may not be what the user intended. It's also unclear to me what happens if the same function ends up called both as a request-scoped and a lifetime-scoped dependency at different places throughout the application. Would the request-scoped dep short-circuit to using the already available lifetime dep or would it create a separate instance of it? What if the lifetime dep hasn't been initialized yet?

> The main difference between regular startup events and this system is that this system is lazy: nothing is run until the first request that needs it comes in.
> 
> Or more generally, this is not DI for startup/shutdown events (which is what the above linked issues were asking for) but rather startup/shutdown events for the DI system. Maybe it's not what we wanted, but perhaps it's just as good?

More generally, I'm not sure of the benefit of lazily-loaded shared resources compared to having dependencies context-managed by the lifetime protocol (the main reason I can think of being parametrized dependencies, which weren't the sort of thing I was considering using for lifetime deps in the first place).

> In most real world scenarios, the first request will come from a load balancer, Kubernetes, etc to a healthcheck endpoint, and the app won't even take outside traffic until it consistently is returning 200.
If one wanted to, they could make a Startup dependency that collects all lifetime dependencies and binds them to the healthcheck to ensure that they get initialized before any user requests:

I can't speak for most FastAPI users, but the sort of setup I'm using (just a plain reverse proxy for a small API server connected to a bunch of heterogeneous databases) wouldn't fit under that scenario.

Really, my main use case for lifetime deps when I opened #617 was to not have active object handles littering the global namespace outside of functions (see https://github.com/tiangolo/fastapi/issues/726#issuecomment-557687526, and imagine something like that, but minus the underscored global variables). I figured those could just be wrapped in CM-like functions and annotated for FastAPI to know that those functions, when used as dependencies, are not supposed to be run, but to have the eagerly singleton'd result (hopefully a *thread-safe* result, I don't think that can reliably be checked by FastAPI, so it's up to the user to make sure) returned instead.

It also generally makes sense to me that your server should fail to startup if the database connection your application requires to run is unreachable, or if your core config file can't be found, which is why I would have imagined these as not being lazy, but we might just have different use cases for these.

adriangb:
> we might just have different use cases for these

This is very possible, I may have been shortsighted with my use cases, but I'll try to answer some of good questions you bring up

> It's also unclear to me what happens if the same function ends up called both as a request-scoped and a lifetime-scoped dependency at different places throughout the application

> Would the request-scoped dep short-circuit to using the already available lifetime dep or would it create a separate instance of it?

I think this is an important question. To me it boils down to "is an app lifespan dep == to the same dep but with a request lifespan?". Let me know if this interpretation is incorrect.
We can make it work either way, I think we'd just have to add a tuple element to the cache key to make them unequal.
In other words, we can have either of these two work:

```python
class Config:
    ...


def get_config() -> Config:
    return Config()


@app.get("/")
def root(cfg1 = Depends(get_config, lifetime="app"), cfg2 = Depends(get_config)):
    assert cfg1 is cfg2  # option 1, current implementation
    assert cfg1 is not cfg2  # option 2, adding the `lifetime` parameter as a cache key element
```

> On one hand, it means any dependency marked as a lifetime dep anywhere causes all of its transitive dependencies to be run and cached as lifetime deps at the same time, which may not be what the user intended

The intention is that only the top level (let's call it root) dependency gets marked as a lifetime dependency and cached as such. If a transitive dependency (`get_config` in this example) is used by both a lifetime dependency and a non-lifetime dependency it will only be "shared" the first call:

```python
class Config:
    ...


def get_config() -> Config:
    return Config()


def lifetime(cfg: Config = Depends(get_config)):
    return cfg

def request(cfg: Config = Depends(get_config)):
   return cfg

calls = 0

@app.get("/")
def root(cfg1 = Depends(lifetime, lifetime="app"), cfg2 = Depends(request)):
    global calls
    calls +=1
    if calls == 1:
        assert cfg1 is cfg2
    else:
        assert cfg1 is not cfg2
```

>  I'm not sure of the benefit of lazily-loaded shared resources compared to having dependencies context-managed by the lifetime protocol 

I agree, it would be nice to have the opposite (DI for the startup/shutdown system) as well. Like you say, it often makes sense to not even start the application unless the database, etc. are able to connect. But maybe there is a use case for both?

> just a plain reverse proxy for a small API server

As far as I know, even simple reverse proxies like NGINX have healthcheck functionality.

In your use case, how do you know a deploy was successful if you don't have some sort of healthcheck / rediness?

> I'm not personally sold on having the dependency's lifespan be specified on the caller's end instead of the dependency's end

This is not necessarily a solution to this, but maybe something like the following can be used to clean up things:

```python
def DBConnection() -> Connection:
    def real_dep(cfg: Config = Depends(get_config)):
        ...
    return Depends(real_dep, lifetime="app")


@app.get("/")
def root(conn: Connection = DBConnection()):
    ...
````

Now the caller doesn't have to specify `Depends(..., lifetime="app")`.

**Comment by Faylixe** [comment]
> I'm not personally sold on having the dependency's lifespan be specified on the caller's end instead of the dependency's end. 

+1 on this, not a big fan as well.

> Now the caller doesn't have to specify Depends(..., lifetime="app").

In that case, if dependency behavior should be defined by the user I would prefer having a factory function handling this:

```python
def AppDepends(*args, **kwargs) -> Any:
    return Depends(*args, **kwargs, lifespan="app")
```

Which then can be use like a traditional one:

```python
@app.get("/")
def root(cfg1 = AppDepends(get_config), cfg2 = Depends(get_config)):
```

adriangb:
Yeah, I totally understand on:
- Wanting this to be tied to startup/shutdown
- Wanting to specify the lifetime on the dependency side

The _main_ reason I'm not doing it like that is that:
- There is no DI integration into startup/shutdown
- The DI system currently relies on specifying things on the caller's side (eg. `use_cache`)

And so doing it this way allowed me to do it in relatively few lines of code, with no major refactors, while still satisfying all of _my_ needs.
I think trying to hit either of those other two features would require more extensive refactors (although maybe I'm just not familiar enough with the codebase).

I do think there's some use for both, I can see scenarios where it is desirable to lazily initialize lifetime dependencies.

graingert:
hello! I'd highly recommend using the `Starlette(lifespan=...)` interface for this so people can create `anyio.TaskGroup`s in their lifespan dependencies.

To support this, `lifespan` was changed in 0.16.0 to only support async context manager factories, eg using `@contextlib(2).asynccontextmanager`, and importantly the `yield` statement can now throw an instance of `anyio.get_cancelled_exc_class()` eg:


```python
@contextlib.asynccontextmanager
async def TaskGroup() -> anyio.TaskGroup:
        yield tg  <- this might throw asyncio.CancelledError or trio.CancelledError!
```

adriangb:
100% agreed, thank you for the suggestion.

The only reason I didn't use `Starlette(lifespan=...)` is because FastAPI does not support that yet (in particular, FastAPI's router class doesn't). And so adding those keyword arguments to `Router`, etc. would have been out of scope for this PR. It may be possible to skirt around adding any public parameters, but there's an even bigger problem: because Statlette can't have both `Starlette(startup=[...])` and `Starlette(lifespan=...)` at the same time, if we add a `lifespan` argument we would disable users existing startup/shutdown hooks, which would break a lot of code.

sm-Fifteen:
> The second bigger problem is that Statlette can't have both `Starlette(startup=[...])` and `Starlette(lifespan=...)` at the same time, if we add a `lifespan` argument we would disable users existing startup/shutdown hooks, which would break a lot of code.

The startup and shutdown hooks are merely shimmed by the default starlette lifespan context manager (see encode/starlette#799), now. There's nothing stopping FastAPI from adding a custom lifespan handler to take care of lifespan dependencies that also shims the old startup and shutdown events.

adriangb:
@graingert I pushed a version using `Starlette(lifespan=...)` in 
629e465
@sm-Fifteen you're right, thanks for the recommendation. I ended up with this shim: https://github.com/tiangolo/fastapi/blob/629e465d7afe8fac442066c62f2897455f5198f8/fastapi/routing.py#L454-L479 I think the only tricky bit left is that whenever FastAPI enables `FastAPI(lifespan=...)` we'll have to handle adding the user's lifespan into this glue lifespan (I suppose checking if it is a context manager or generator, etc.).

docs_src/dependencies/tutorial013.py
```diff
@@ -0,0 +1,38 @@
...
+
+class AppConfig(BaseSettings):
+    database_hostname: str = "localhost"
+
+
+def Config() -> AppConfig:
+    return AppConfig()
+
+
+def DBConnection(config: AppConfig = Depends(Config)) -> Database:
```
graingert, line 26:
```suggestion
@contextlib.asynccontextmanager
def DBConnection(config: AppConfig = Depends(Config)) -> Generator[Database, None, None]:
```

Pushing the responsibility of managing the generator down to the user solves a number of weird edge cases around Exceptions/StopIteration and Traceback preservation

adriangb: makes sense, thanks!

graingert:
I think it's worth waiting for the Starlette 0.16.0 upgrade in this pr because so much changed about how lifespan works

adriangb:
Are there more changes beyond having to wrap the lifespan func in `@contextlib.asynccontextmanager`?

Do you know if FastAPI has a plan / upgrade path for Starlette versions?

adriangb:
Looks like https://github.com/tiangolo/fastapi/pull/3372 is what we'd be waiting for

adriangb:
Alrighty folks, some exciting news, I reworked this to support dependencies in startup/shutdown events (and also in lifespan managers, whenever we update to `starlette=0.16.0`)

I decided to keep this in the same PR since:
1. The underlaying implementation using the lifespan AsyncExitStack is shared.
2. The question of caching/sharing deps between lifespan/startup events and endpoints naturally arises, and is addressed using the original mechanism of this PR.

To elaborate on (2), I renamed the parameter `Depends(..., lifespan="app")` to `Depends(..., cache_lifespan="app")`, which not only more accurately reflects what it is doing, but also gives users the ability to choose if they want to cache startup/lifespan dependencies to be shared with endpoints or not. Endpoints also always have the option of using `Depends(..., cache=False)` to independently disable caching of a particular dependency for a particular endpoint.

I still need to update the docs, but I have basic tests for most of the behavior described above.

sm-Fifteen:
The wording change goes a long way to make the changes fit more cleanly in the current model, since they would allow deprecating/replacing the current `use_cache` parameter and establishes a clear mental model of what should happen if the same dependency is called with or without that parameter (or with different values). While it doesn't change my reservations about how caller-side vs function-side lifespan specifications, this has the potential to work well enough in more varied use cases and would be easy to tutorialize on top of the existing dependency stuff, so I'm willing to give it a go.

A few things I should note, though:
* Like with all forms of shared resources (whether passed as dependencies or from global scope), users should ensure that any resource managed this way is both thread-safe and coroutine-safe (one does not always imply the other, see the issues raised with `scoped_session` in https://github.com/tiangolo/fastapi/issues/726#issuecomment-868750382)
* Like I mention in the "Additional context" part of my original post for https://github.com/tiangolo/fastapi/issues/617#issue-506258325, the user should be aware that the shutdown phase is not strictly guaranteed to run if the server doesn't get to exit cleanly (so that they can make sure that database connections time out properly and such).
* Starlette currently has a problem with lifetime on mounted subapplications (encode/starlette#649) so you might want to either have a backup mechanism in case lifetime didn't fire properly or somehow disallow lifetime deps on sub-applications somehow.

graingert:
Thanks for your work on this feature, I'm really exited to use it.

There's some issues with lifespan and multiple threads exacerbated by the starlette.TestClient: https://github.com/django/asgiref/pull/277

And there's some additional issues with Trio and AsyncExitStack: https://github.com/python-trio/trio/issues/1243

graingert:
> the user should be aware that the shutdown phase is not strictly guaranteed to run if the server doesn't get to exit cleanly

This should no longer be true with Starlette 0.16.0. if the lifespan startup hook completes the lifespan shutdown hook will certainly be called. (Unless the thread is aborted with os._exit or get_event_loop().run_until_complete is used)

adriangb:
> The wording change goes a long way to make the changes fit more cleanly in the current model

👍 glad it helps!

> allow deprecating/replacing the current use_cache

Hmm I was thinking these would work in tandem. One establishing the lifespan, another if caching should be used at all or not.

> users should ensure that any resource managed this way is both thread-safe and coroutine-safe

Agreed. I think there's not much we can do about this other than write good documentation, right?

> user should be aware that the shutdown phase is not strictly guaranteed to run if the server doesn't get to exit cleanly

Agreed again. But also documentation right?

If I am misunderstanding and either of these have implementation implications, please let me know.

> Starlette currently has a problem with lifetime on mounted subapplications (...) you might want to either have a backup mechanism in case lifetime didn't fire properly

This seems like something where we should let Starlette resolve it (maybe holding off on implementing this feature until they do if it's a deal breaker), not something FastAPI should be working around.

graingert:
>  If those two parts could be welded wold put back together, I feel like we'd have a much more solid mechanism for managing lifetime-scoped ressources than the current one. 

Afaik this is exactly the fix I did here https://github.com/encode/starlette/pull/1227

adriangb:
> There's some issues with lifespan and multiple threads exacerbated by the starlette.TestClient

> And there's some additional issues with Trio and AsyncExitStack

To be honest @graingert, these are a bit too deep into ASGI and anyio for me to fully understand. If you have the time, I would appreciate a bit of guidance on the implication to this PR.

> > If those two parts could be welded wold put back together, I feel like we'd have a much more solid mechanism for managing lifetime-scoped resource than the current one.

(from https://github.com/tiangolo/fastapi/pull/3516#issuecomment-884333414)

I couldn't find where this quote is coming from, maybe it was from an edited comment? Could you please link me to the context? Thanks

graingert:
https://github.com/tiangolo/fastapi/issues/617#issuecomment-548465697

adriangb:
Thank you @graingert

I do agree that the `give me a context manager as a lifespan handler` approach is much, much better than startup/shutdown events. And, as far as I can tell, this PR will be compatible with it with minimal changes (once FastAPI updates it's Starlette version).

sm-Fifteen:
> https://github.com/tiangolo/fastapi/issues/617#issuecomment-548465697
> > If those two parts could be welded wold put back together, I feel like we'd have a much more solid mechanism for managing lifetime-scoped ressources than the current one.
> 
> Afaik this is exactly the fix I did here [encode/starlette#1227](https://github.com/encode/starlette/pull/1227)

Ah, I thought encode/starlette#799 had already managed this, but I guess forcing the use of context managers instead of generators could make everything more robust in a bunch of cases (hard to say from the code diff alone).



> > the user should be aware that the shutdown phase is not strictly guaranteed to run if the server doesn't get to exit cleanly
> 
> This should no longer be true with Starlette 0.16.0. if the lifespan startup hook completes the lifespan shutdown hook will certainly be called. (Unless the thread is aborted with os._exit or get_event_loop().run_until_complete "Pending task was destroyed!" is used)

That's what I was referring to. VSCode will immediately kill your process [if stopped via the debugger's stop button](golang/vscode-go#120) (I'm not sure whether or not the Python extension still does it), and I've had that issue bite me more than once. Non-graceful shutdowns happen, so I figure that's something to be aware of.


> > allow deprecating/replacing the current use_cache
> 
> Hmm I was thinking these would work in tandem. One establishing the lifespan, another if caching should be used at all or not.

Ah, I figured that's what ~~`cache_lifespan=None`~~ `cache_lifespan=False` (`None` seems like a better parameter type to use) was supposed to be for.

> > users should ensure that any resource managed this way is both thread-safe and coroutine-safe
> 
> Agreed. I think there's not much we can do about this other than write good documentation, right?
> 
> > user should be aware that the shutdown phase is not strictly guaranteed to run if the server doesn't get to exit cleanly
> 
> Agreed again. But also documentation right?

Yes, both of these were highlighted more in regards to documentation than to code, given how there's not really anything that can realistically be done at runtime for either of those issues.

> > Starlette currently has a problem with lifetime on mounted subapplications (...) you might want to either have a backup mechanism in case lifetime didn't fire properly
> 
> This seems like something where we should let Starlette resolve it (maybe holding off on implementing this feature until they do if it's a deal breaker), not something FastAPI should be working around.

Yeah, I'm not sure what to do in that regard here either. On one hand, it may lead to unexpected application breakage, on the other hand I don't know if that use-case is enough of a roadblock to put this feature on hold until the upstream bug gets resolved, given it's been sitting untouched for almost 2 years.

adriangb:
> I figured that's what cache_lifespan=None cache_lifespan=False (None seems like a better parameter type to use) was supposed to be for.

I actually like the idea of replacing `cache=False` with `cache_lifespan=None`. But I think it would break backwards compatibility / we'd have to figure out what to do if `cache_lifespan=None, cache=True`.

graingert:
Seems like something akin to pytest.fixture scope might work? Maybe with an Enum instead?

```
class Scope/(enum.Enum):
    lifespan = enum.auto()
    request = enum.auto()
```

?

sm-Fifteen:
> I actually like the idea of replacing `cache=False` with `cache_lifespan=None`. But I think it would break backwards compatibility / we'd have to figure out what to do if `cache_lifespan=None, cache=True`.

From what I've seen elsewhere, you usually check for the new parameter (`cache_lifespan=None`) first and use it if it's set, and then either ignore the old one, emit a deprecation warning or throw if both happen to be set. Given how `use_cache` currently has a default value of `True`, `cache_lifespan` could default to `Ellipsis` (unset, aliasing to `'request'` by default) so FastAPI can check if it has been set intentionally or not, and have it take priority over `use_cache` if it's is set (by the application programmer) to something else.

adriangb:
> Seems like something akin to pytest.fixture scope might work? Maybe with an Enum instead?

That could work, but it would add an extra import.

In any case, it seems like the use of enum and/or replacing `cache: bool = {True,False}` seem totally doable, albeit minor things that don't change the fundamentals of the proposed change. I'm going to hold off on those changes until we can get review from @tiangolo  / more clarity on the viability of the PR in general.

graingert:
This documentation goes into more details about the sort of shared startup and shutdown that I think FastAPI will need to implement: https://pytest-trio.readthedocs.io/en/stable/reference.html#an-important-note-about-yield-fixtures

adriangb:
> This strategy of using a task for each fixture (read Dependency lifespan) has a number of benefits, it means task group and cancel scopes work as expected and it means that you can execute startup and shutdown handlers concurrently

Interesting. I will have to do a bit more reading / give this some more thought, I'm not super familiar with some of this stuff, especially anyio (I'm just a vanilla asyncio user). If you prefer, please feel free to jump in and commit changes instead of holding my hand.

graingert:
It looks like something that can be handled by anyio for you. Eg you pass it a DAG of async context manager factory functions and you get back a single async connect manager that schedules them all for you

adriangb:
On another note, I pushed doc/example updates (just examples, I still have to rewrite the docs). There are now 2 examples, that are really just rewrites of current examples:
- A rewrite of the [Settings as a dependency](https://fastapi.tiangolo.com/advanced/settings/?h=sett#settings-in-a-dependency) example that replaces `@lru_cache` with a `cache_lifespan ="app"` dependency.
- A rewrite of the [Async Databases example](https://fastapi.tiangolo.com/advanced/async-sql-databases/?h=async+data) that instead of hardcoding the DB URL as a global variable pulls it from a Pydantic config (this is typical for a real world use I think) and connects to the database as a startup event. The startup event dependency is marked with `cache_lifespan="app"` so that endpoints get the same instance of the result, and context management of the database connection is handled via the app's lifespan AsyncExitStack. It then uses dependency overrides to override the Postgres URL with an SQLite url (the original example just comments out the global variable and replaces it)

GabeMedrash:
Watching from afar with keen interest...

May also be helpful to see how other web frameworks have approached pieces of this question. Spring, for example, is an interesting comparison. Related to dependency scoping, it allows for beans--depending on the application's context--to have one of 6 scopes: https://www.baeldung.com/spring-bean-scopes. Perhaps "request" and "app" scoping is enough for FastAPI for now, but, may be helpful to see how this feature has grown in a longer-standing framework.

adriangb:
That's very interesting, thank you for sharing.

It seems to me that, as currently implemented, this PR provides the equivalent of several of those scopes:
- `cache=True, cache_lifetime="request"`: equivalent to Spring's `Request` and `WebSocket` scopes.
- `cache=False, cache_lifetime="request"`: equivalent to Spring's `Prototype` scope.
- `cache=True, cache_lifetime="app"`: equivalent to Spring's `Singleton` scope.
- `cache=False, cache_lifetime="app"`: this one doesn't make much sense. If used in an endpoint, it would just keep adding to the exit stack every time it's called, and they would all exit when the app stops. I'd lean towards explicitly prohibiting this behavior.

There is no real equivalent for Springs `Application` scope, if I'm understanding it correctly. You'd be better off just using a module level global.

@GabeMedrash please correct me if this interpretation is flawed (it probably is in some way, I'm not a Spring user).

So maybe we should just have one parameter like `cache={request/True,prototype/False,app}`?

GabeMedrash:
Yeah, that's a good point: there's a decent amount of flexibility built into the combination of `cache` and `cache_lifespan` as you've implemented here, and your comparisons look about right to me on first glance, though I'm sure there are "details" (there always are with Spring). 

Of course, the challenge is making sure that it's clear/explicit how to accomplish a task given the API. I'm just some random guy on the internet--no need to give me too much attention--but, I tend to find, e.g., `scope="prototype"` a little clearer than `cache=False, cache_lifespan="request"`. Perhaps though something like `scope="prototype"` (and the like) is ultimately syntactic sugar over these more low-level args.

adriangb:
I think you're making a very good point in that `scope` (in the Spring sense) is syntactic sugar on top of a combination of how long the value is stored (`cache` in our case) and when the bean is created/destroyed (`lifespan/lifetime` in our case).

I think the nice thing is that because of how generator dependencies and the cache are implemented here, we have full control over both aspects independently. This got me thinking about all of the options we have that users might want. I'm going to leave aside _creation_ of dependencies since that is governed by when they are first needed (which simplifies understanding of startup/shutdown events: they just move the dependency creation to happen before the first request).

## Dependency Lifespans

### Request
This is the current default / only option.
Once the endpoint function runs, the dependency is destroyed in the background.

### App
The original proposal in this PR.
The dependency gets destroyed when the app is destroyed.

### Endpoint
The dependency gets destroyed when the endpoint finishes running (i.e. immediately before the request is done processing). Notably, this would allow the dependency to raise `HTTPExceptions` while being torn down. This is has already been proposed and implemented in #2697.

### None/ Background
The dependency gets destroyed in the background immediately after its value is retrieved.
This could be useful if for example you retrieve a scalar value or something and need to do some teardown, but the endpoint doesn't depend on the teardown itself, it only needs the value.
Instead of waiting until the endpoint finishes running to start the teardown (which is the behavior with the current / `Request` lifespan), the teardown starts immediately after the the value is retrieved.

## Caching

### True / Request
This is the current default behavior.
The dependency's returned value is cached for the lifetime of the request.

### False
This is current optional behavior
Every time this dependency is requested, it is re created.

### App
Proposed in this PR.
The _value_ of the dependency is cached for the lifetime/lifespan of the App (note that the dependency itself could be torn down immediately, eg. if using `lifespan="background"`).
This acts as a default for request cached dependencies: if an app-cached value exists, it is returned and the dependency is not re-constructed.

adriangb:
I pushed a series of commits implementing the options proposed in https://github.com/tiangolo/fastapi/pull/3516#issuecomment-885107031.

The only one I didn't implement is `lifetime="background"` because the implementation is dissimilar to the others and requires a bit more technical thought; that can be a future PR / feature.

I'd like to acknowledge @zoliknemet for the excellent implementation in #2697, which I partially borrowed to implement `lifetime="endpoint"`.

Additionally:
- The implementation is now using enums, as proposed in https://github.com/tiangolo/fastapi/pull/3516#issuecomment-884397298. `cache={True,False}` gets converted to the Enum names if passed, making the change backwards compatible (although it might break static analysis tools unless we add a `Union[...]` type hint).
- Any startup/shutdown dependencies that depend on `Request`, `Response`, etc. are caught and a user-friendly error is given.

fastapi/dependencies/utils.py
```diff
@@ -462,10 +468,91 @@ async def solve_generator(
...
+        dependant.request_param_name,
+        dependant.query_params,
+        dependant.header_params,
+        dependant.cookie_params,
+        dependant.body_params,
+        dependant.request_param_name,
+        dependant.websocket_param_name,
+        dependant.http_connection_param_name,
+        dependant.response_param_name,
+        dependant.path
```
adriangb, line 543: I hope this is a sensible set of exclusions. I dunno what to do with the security dependencies.

adriangb:
Some more doc edits. I've gone with editing the existing examples referenced in https://github.com/tiangolo/fastapi/pull/3516#issuecomment-885334657 instead of duplicating them.

adriangb:
In case any of your are interested, I made a toy repo where I've been prototyping what it would look like to generalize the DI system to be less request/response coupled and to support the features proposed in this PR.

Here's an example of how it might integrate into FastAPI internally (the goal is near zero user facing changes): https://github.com/adriangb/anydep/blob/main/comparisons/test_anydep_request_params.py

I was also able to get some other nice features working:
- Anyio compatibility
- Pretty good speedups for highly nested / branched dependencies by parallelizing execution using anyio task groups
- Full typing support (i.e. `Depends` tells mypy that it returns the type that the callable you gave it is annotated with instead of `Any`)

adriangb:
I'm going to close this in favor of discussion in #3641

PR #3529, opened 2021-07-14
Add a run function to the base FastAPI for running the server via uvicorn from the base class.

Ixyk-Wolf:
Just added a run function to the base so it can run it using uvicorn without having the user need to import it and do it themself. Keeps all original arguments and raises an error if uvicorn isn't installed. Just wanted to do this because Flask has this and I have wanted it in FastAPI since I switched over.


fastapi/applications.py
```diff
@@ -737,3 +737,10 @@ def trace(
             name=name,
             callbacks=callbacks,
         )
+    def run(self, **args):
+        try:
+            import uvicorn
+            uvicorn.run(self, **args)
+        except:
```
Kludex, line 744:
```suggestion
        except ImportError:
```

graingert:
```suggestion
        except ModuleNotFoundError:
```

Ixyk-Wolf:
Sorry if I'm slow to respond in the next few weeks because I'll be really busy.

fastapi/applications.py
```diff
@@ -769,6 +769,7 @@ def trace(
     def run(self, **args):
         try:
             import uvicorn
-            uvicorn.run(self, **args)
         except ImportError:
```
graingert, line 772:
```suggestion
        except ModuleNotFoundError:
```

graingert: Using this suggestion, this will raise a more informative diagnostic if uvicorn is installed, but fails to import

Ixyk-Wolf: Traceback (most recent call last):
  File "C:\Users\Isaac\stdinmedia\uvicorn.py", line 11, in <module>
    app.run()
  File "C:\Users\Isaac\stdinmedia\venv\lib\site-packages\fastapi\applications.py", line 771, in run
    import uvicorn
  File "C:\Users\Isaac\stdinmedia\uvicorn.py", line 11, in <module>
    app.run()
  File "C:\Users\Isaac\stdinmedia\venv\lib\site-packages\fastapi\applications.py", line 775, in run
    uvicorn.run(self, **args)
AttributeError: partially initialized module 'uvicorn' has no attribute 'run' (most likely due to a circular import)


This is the error that is raised when I name the project file uvicorn. It is neither an ImportError or a ModuleNotFoundError. I really don't see it.

graingert: Here's some more details behind the motivation of `ModuleNotFoundError`: https://bugs.python.org/issue15767

if `import uvicorn` raises an instance of `ImportError` that's not an instance of `ModuleNotFoundError` then installing uvicorn won't fix it

Ixyk-Wolf: I have just tested it, and see no difference. I'll just switch it I guess.

Ixyk-Wolf:
I wonder if this should be a pull request for Scarlette so it is also in FastAPI. I think it would make more sense.

vrslev:
I don’t think changes belong to a _ASGI framework_. `uvicorn` is one of many ASGI servers. And more importantly fine grained hierarchy would be broken if we’d couple ASGI servers and apps.

BilalAlpaslan (approved):

tiangolo:
Thanks for the interest @Ixyk-Wolf! 🍰 

I prefer to teach how to use Uvicorn, and run it directly. If the added abstraction (in this case `app.run()`) and explaining how to use it doesn't simplify significantly using the lower level part (Uvicorn), I prefer to teach how to use the low level part than to create an abstraction on top.

Given that, I'll close this one, but thanks for the effort! ☕

PR #3536, opened 2021-07-15
📝 Update docs for handling HTTP Basic Auth with `secrets.compare_digest()` to account for non-ASCII characters

lewoudar:
Updated by @tiangolo:

📝 Update docs for handling HTTP Basic Auth with `secrets.compare_digest()` to account for non-ASCII characters.

---

> I changed the basic auth security example to encode data in bytes using the `idna` format. This is because `compare_digest` only checks **ascii** characters but in the real world, many people have a name with non-ascii characters :smiley:

> I also removed useless docstring comment.


tiangolo:
Thank you @lewoudar! :nerd_face: 

I updated the example and docs a bit to make it more explicit. I used UTF-8 instead of IDNA, as this is not related to domains.

I reverted the change in the docstring to put it in a different PR as that is a different thing. Also because only removing that line also left the example code broken. But let's fix that in another PR later.

lewoudar:
@tiangolo perfect!

PR #3551, opened 2021-07-19
Enable configuration of Swagger UI #2568 - updated

r-downing:
Piggybacking off of this one - updated it to ensure proper json formatting for the args https://github.com/tiangolo/fastapi/pull/2568/commits/c85e97782599986ae41548c4b4f714353e7066d4


tiangolo:
Thanks for the effort! I think this was already handled in another PR, handling JSON and also encoding other possible object types with `jsonable_encoder`.

So I'll close this one now, but if you feel there's something else to do, please create a new PR to avoid conflicts. Thanks! :cake:

PR #3560, opened 2021-07-20
enable branch coverage

graingert:


Kludex:
We lost some coverage over time on the `docs/`, you can see more about it here: #1904.

tiangolo:
I imagine this is done, right? You are not planning anything else with this, right?

I'll close this one, but if you wanted to work more on something around this, please create a new/fresh PR, thanks! :cake:

PR #3583, opened 2021-07-22
🙈 Add more environment related files to `.gitignore`

mawassk:
We should add `.venv` to `.gitignore`, because it would be contributor friendly cause `.venv` is a common name for python virtualenvs (it is also used by poetry) and many devs configured there dev tools to use `.venv` as a default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ env3.*
 env
 docs_build
 venv
+.venv
 docs.zip
 archive.zip
 
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ env
 docs_build
 venv
 .venv
+.env
+ENV/
 docs.zip
 archive.zip
 


elrik75:
Seems not relevant as every developer that use `.venv` are able to create a global `gitignore` to exclude all `.venv`.
http://egorsmirnov.me/2015/05/04/global-gitignore-file.html

mawassk:
> Seems not relevant as every developer that use `.venv` are able to create a global `gitignore` to exclude them all.
> http://egorsmirnov.me/2015/05/04/global-gitignore-file.html

You are right, that there are other easy ways to handle this issue, but for new contributors, who doesn't know about these options, this could be a little bit obstructive and from my point of view we should aim contributor friendliness. Also as i mentioned above, many developers are affected by these, because it's a common name for the virtual environment.

dukkee:
@elrik75 if you check current `.gitignore` file, you will find `venv`, why then we don't need to add `.venv` variant? :smiley:

dukkee:
@mawassk if you solved to update .gitignore file, then maybe it makes sense to add all variants to not generate similar PRs (related to env folders I mean) in the future. E.g. in the one of my `.gitignore` files I see the next variants: https://github.com/boardpack/cookiecutter-pyboardpack/blob/main/%7B%7Bcookiecutter.dist_name%7D%7D/.gitignore#L105

mawassk:
> @mawassk if you solved to update .gitignore file, then maybe it makes sense to add all variants to not generate similar PRs (related to env folders I mean) in the future. E.g. in the one of my `.gitignore` files I see the next variants: https://github.com/boardpack/cookiecutter-pyboardpack/blob/main/%7B%7Bcookiecutter.dist_name%7D%7D/.gitignore#L105
@dukkee That's a nice idea! I will definitely add them, but currently i see no use-case for the backup files `venv.back/` and `env.bak`. Can you provide more information about working with backup files?

dukkee:
> > @mawassk if you solved to update .gitignore file, then maybe it makes sense to add all variants to not generate similar PRs (related to env folders I mean) in the future. E.g. in the one of my `.gitignore` files I see the next variants: https://github.com/boardpack/cookiecutter-pyboardpack/blob/main/%7B%7Bcookiecutter.dist_name%7D%7D/.gitignore#L105
> 
> @dukkee That's a nice idea! I will definitely add them, but currently i see no use-case for the backup files `venv.back/` and `env.bak`. Can you provide more information about working with backup files?

Honestly, I also didn't hear about such variants :smile:

dukkee (approved):

PR #3585, opened 2021-07-22
Typographical Fix: Change Default Title Separator from Hyphen to En Dash

ChristianMertes:
In usages such as "App Title – Swagger UI", an en dash (–) or an em dash (—) can be used, but never a hyphen (-). So this tiny PR changes the hyphens in these cases to en dashes. The difference between - and – might seem subtle depending on the font used but it's definitely easier on the eye to use the typographically correct dash here.

See also https://en.wikipedia.org/wiki/Dash#Parenthetic_and_other_uses_at_the_sentence_level

This PR doesn't change usages in the docs as this would be a whole nother can of worms.

### DIFF
--- a/docs_src/extending_openapi/tutorial002.py
+++ b/docs_src/extending_openapi/tutorial002.py
@@ -15,7 +15,7 @@ app.mount("/static", StaticFiles(directory="static"), name="static")
 async def custom_swagger_ui_html():
     return get_swagger_ui_html(
         openapi_url=app.openapi_url,
-        title=app.title + " - Swagger UI",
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks for your interest, but I would prefer to keep it as it is. I wouldn't be able to type the en dash on my keyboard, it's not encodable with ASCII, it would require/assume UTF-8, but if someone is on Windows that assumes windows-1252 (extension of iso-8859-1/latin-1) then it would not work correctly, etc.

So, anyway, I'll pass on this one, but thanks for your effort!

PR #3598, opened 2021-07-23
Feature: Inject the final path operation response into dependencies

### LINKED ISSUE #3500: Allow access to the path operation's Response from within Dependencies using yield
## Checks

## Problem

FastAPI's dependency injection system supports generators as dependencies ([docs](https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/)).

But there is currently no way to access a response generated by the path function in dependencies.
Well, you _can_ access _a_ response object, but it doesn't get merged with the response from the path operation.
Here's a minimal example:

```python
from fastapi import Depends, FastAPI, Response


def dependency(response: Response):
    yield
    assert response.status_code is None  # so not the same response object!


app = FastAPI(dependencies=[Depends(dependency)])

@app.get("/")
def root():
    return Response(status_code=400)


client = TestClient(app)
res = client.get("/")
```

If your path operation accepts a Response object and modifies and returns the same object,
then you can access data from the path operation (like status code) in the dependency.
Otherwise, you cannot.

## Use case

The main use case I have for accessing this data is logging, but I imagine there may be others.

## Proposed solution

I _think_ that access to the final Response object could be enable via the generator's `send` functionality:

```python
def dependency():
    response = yield
    assert response.status_code == 400
```

I think the main complication with this will be how FastAPI is using these generators internally.
I believe they are being wrapped in context managers, which means the ability to use `.send()` is lost.

### Changing how responses are merged
I know somewhere within FastAPI Response objects get merged together.
I'm not sure if it's possible, but an alternative might be to merge into the dependency's response object instead of creating a new object or merging into the path operation's response (I am not sure which is the current behavior).
This may or may not be a breaking change, I'm not sure.

### Middleware
For the particular use case of logging, it is possible to use a middleware.
But middleware's are in some ways not as flexible as dependencies.
For example, you can't apply a middleware to just some paths (for example, to avoid logging data on a path that might handle sensitive information) unless you resort to "give me a list of path regexes as a constructor parameter" in your middleware.

adriangb:
**Background**

Currently, there is no way for dependencies to access the final response sent by the path operation function.
Even if the path operation itself depends on `Response`, the connection will be lost unless it returns the same object (which is often the case, eg. if returning a dict or Pydantic model).

**Goal**

To be able to access the final response from dependencies.

This makes a good replacement for logging middlewares: instead of coming up with some regex path matching system (eg. [here](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-fastapi#exclude-lists) ), users can apply this dependency at the app level (all routes), router level (some routes) or route level (individual route).

If combined with #3516 , dependencies would become just as powerful as middleware (in the sense that they can access and edit the final response) while still retaining the advantages of the DI system.

**Proposal**

Take advantage of generator's ability to receive a value from yield to give dependencies access to the final response being returned to the client.

**Associated issues**
Closes #3500 

**Sample usage**

```python
def log_request_response(request: Request):
    print(request.status_code)  # or any other request data
    response: Response = yield
    print(response.status_code)  # or any other response data

app = FastAPI(dependencies=[Depends(log_request_response)])
```

**Alternatives**

A simpler alternative is presented in #3600 .

Personally I am indifferent to which one is chosen.


adriangb:
Since this likely won't get merged, I'm going to close this to avoid contributing to PR pile up and let the maintainers focus on PRs fixing long standing bugs

WolfDWyc:
Hey, should this be re-opened? I saw that a few months ago you tiangolo said he now intends to add this at some point.

Happy to help with it if it's missing something :)

PR #3603, opened 2021-07-25
Add context manager syntax to dependency_overrides

adriangb:
This adds syntactic sugar so that `dependency_overrides` can used as a context manager.

Requiring users to set something and then remember to reset it is a typical use case for context managers, especially when not catching errors or _forgetting_ to reset it might have very hard to debug consequences in seemingly unrelated tests.


adriangb:
Since this likely won't get merged, I'm going to close this PR to avoid contributing to PR pile up

PR #3607, opened 2021-07-26
🐛 Fix `jsonable_encoder` for dataclasses with pydantic-compatible fields

himbeles:
This PR fixes an issue (#3608) I encountered when returning `dataclasses` responses with pydantic-compatible (but incompatible with json encoder of Python standard library) fields, e.g. `datetime` objects or `pathlib` file paths.

This was achieved by adding a call to `jsonable_encoder(dict)` after converting the dataclass to a dict in `jsonable_encoder`.

For testing, I have added a `datetime` field to the  `Item` class in [tests/test_serialize_response_dataclass.py](tests/test_serialize_response_dataclass.py). 
Without this PR, the last two tests in this file fail (`test_no_response_model_objectlist`, `test_no_response_model_object`).
Note that the tests also pass with a pre 0.67 version of FastApi (before official dataclass response support).


zroger:
I am having this same issue, and was going to propose the same fix. 

I think test cases should be added to [test_jsonable_encoder.py](https://github.com/tiangolo/fastapi/blob/master/tests/test_jsonable_encoder.py) since the `jsonable_encoder` is the affected component, and it can be used for other purposes than just response serialization.

tiangolo:
Amazing, thanks @himbeles! And thanks for adding several tests! :rocket: :cake:

Issue #3620, opened 2021-07-27
Automatically run `db.commit()` in session dependency BEFORE returning request


anguspmitchell:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
# orm.py
def fastapi_session_scope() -> sqlalchemy.orm.session.Session:
    session = session_maker(bind=engine)
    try:
        yield session
        session.commit()
    except:
        session.rollback()
        raise
    finally:
        session.close()

# routers.py
@router.post("/v1/organizations", response_model=Organization)
def create_org(session: Session = Depends(orm.fastapi_session_scope)):
    org = sqlalc.Organization(name="test")
    session.add(org)
    return org

@router.get("/v1/organizations", response_model=List[Organization])
def get_orgs(session: Session = Depends(orm.fastapi_session_scope)):
    return session.query(sqlalc.Organization).all()
```

### Description

(I realize this is diverging slightly from the [SQLAlchemy example](https://fastapi.tiangolo.com/tutorial/sql-databases/), which explicitly uses `session.commit()` in the body of the function and not in the session dependency.)

I'm creating a row with a POST and then immediately afterwards fetching all rows with a GET. I've got `session.commit()` in the session scope dependency, which is convenient because it runs a `commit()` at the end no matter what and doesn't require us to always explicitly commit at the end of an endpoint body.

As written, if an Organization is created via a POST and then all the organizations are immediately afterwards fetched via a GET, the GET will not include the recently created Organization. This is not a JS async issue. What's happening, as best I can tell, is that the POST starts a transaction, `org` is returned, and then, before the `commit()` actually happens, the GET is called and the list of organizations is returned without the organization that was just created (or so it seemed). 

```
t  POST            GET

1  request
2  session.add
3  return
4                  request
[...]

Issue #639, opened 2019-10-22
Execute dependencies in parallel instead of sequentially awaiting them


sm-Fifteen:
**Is your feature request related to a problem? Please describe.**
FastAPI's Dependency injection system, while extremely useful, tends to run into a problem when several long-running dependencies are needed for a single route, such as with the example below.

```py
from fastapi import Depends, FastAPI
from asyncio import sleep as async_sleep

app = FastAPI()

async def get_big_data():
    await async_sleep(5)
    return "𝐃𝐀𝐓𝐀"

async def compute_pi():
    await async_sleep(5)
    return "About 3, I think?"

async def do_nothing_for_a_bit():
    await async_sleep(5)
    return "Gotta let the CPU spin down for a bit, otherwise it could explode."

@app.get("/do_long_thing")
async def do_long_thing(
    big_dep1: str = Depends(get_big_data),
    big_dep2: str = Depends(compute_pi),
    big_dep3: str = Depends(do_nothing_for_a_bit)
):
    return {
        'big': big_dep1,
        'pi': big_dep2,
        'sleep': big_dep3,
    }
```

Running that application and thentrying to access `/do_long_thing` takes 15 seconds to load, even though none of those dependencies are reliant on each other and could very well be ran in parallel.

**Describe the solution you'd like**
Don't await coroutines and threadpools here. Collect them into a list instead and gather them.

https://github.com/tiangolo/fastapi/blob/3f9f4a0f8f7f5db50fbdc3157fe692cf4e68e13f/fastapi/dependencies/utils.py#L442-L456

**Describe alternatives you've considered**
Asking that the user take care of running and gathering those dependencies themselves is pretty much the only solution to this problem available right now (that I can see, anyway, there might be another one), which undermines the usefulness of the dependency injection system.

euri10:
that's indeed a good idea

tiangolo:
This seems like a good idea. But I think it would be quite non-trivial to implement. It would require sorting and finding the dependencies _before_ running them, to be able to know what can be parallelizable that doesn't require something else first.

But if you feel like trying to take a stab at it, a PR would be more than welcome. :muscle:

adriangb:
Hey @tiangolo , just a heads up it looks like @ariloulaleelay has a PR up to implement this in #3902, and I've also worked out how to do it in [di](https://github.com/adriangb/di) (which is based on FastAPI's DI system), so there's at least a couple ideas on how to move forward with this 😄

[...]

PR #3641, opened 2021-08-01
Enhancement Proposal: Generalized DI system

adriangb:
A high-level proposal for enhancements to the DI system.


adriangb:
I believe tbe original request in #3620 also be made to work by having a dependency scope that goes out of scope before the request is returned

ghandic:
@adriangb does this PR help this issue too? https://github.com/tiangolo/fastapi/pull/3953

adriangb:
> @adriangb does this PR help this issue too? #3953

I can't tell just from the description of the issue, but if it is about raising exceptions after `yield` in dependencies, then yes, this could potentially help certain use cases

adriangb:
Seems like #3902 and #639 are also related. `di` is currently able to parallelize both sync and async dependencies: https://github.com/adriangb/di/tree/main/benchmarks

PR #3654, opened 2021-08-04
Proposal: Replace plain Dict parameters w/ TypeDict when there is an existing BaseModel

RileyMShea:
# Context
Some parameters for config in the framework are typed with generic dicts when there are defined models available:

https://github.com/tiangolo/fastapi/blob/25a48287ed6043f7ee81417cef3c4d1f1dc56f72/fastapi/applications.py#L41-L42

https://github.com/tiangolo/fastapi/blob/25a48287ed6043f7ee81417cef3c4d1f1dc56f72/fastapi/openapi/models.py#L66-L69

## Proposal

Replace instances of `Dict[str,Any]` with [TypedDict ](https://pydantic-docs.helpmanual.io/usage/types/#typeddict)instances where there is a corresponding Pydantic model

```python
class ServerD(TypedDict, total=False):
    url: Union[AnyUrl, str]
    description: Optional[str]
    variables: Optional[Dict[str, ServerVariable]
```

Optionally these TypedDicts can be used as fields in [Pydantic Models](https://pydantic-docs.helpmanual.io/usage/types/#typeddict) and Pydantic will validate the field types of the TypedDict.

## Goals
Retain the ease of use for dicts but give end-users additional feedback from editors and static type checkers.

Vscode can provide autocompletion for TypedDict string keys.  
image

Static type checkers can provide feedback on missing/extra fields, wrong types:
image

### Integration

Support of `Typeddict` is generally available in python3.8, but full support for usage with pydantic is only available with either `typing_extensions` or python3.9.  Fastapi depends on pydantic which automatically installs `typed_extensions` so no new dependencies are needed.  For full compatibility with python versions <3.9 FastAPI would need to use the `TypedDict` from `typed_extensions`.

I have the same import setup Starlette currently uses:

https://github.com/encode/starlette/blob/537ab6afd110b79dca95f3c8ecc6980710b1de1c/starlette/testclient.py#L22-L25

### Minimizing code duplication

Ideally, there would be a way to define a class with some fields and then create corresponding `Typedict` and `BaseModel` Classes that could just inherit from that.  Unfortunately, I don't know a way to do so without either breaking IntelliSense or copy-pasting the BaseModel fields into the `TypedDict`.

### Naming

I have named the provisional example `ServerD` and placed that in the same models module as its corresponding BaseModel, but there is probably a better approach to naming and placement of the TypedDict instance if they were implemented.


YuriiMotov:
I like the idea, but having to maintain 2 duplicated schemas (Pydantic models and TypedDict) is not good.
As an alternative, we could modify methods to accept `Union[list[dict[...]], list[Server]]` and then unpack Pydantic models into `dict`s inside.

This way we could write:
```python

app = FastAPI(servers=[Server(url="google.com")])
```

Not as perfect as with `TypedDict`, but less hard to maintain.

For now we have to add `.model_dump()`:
```python
app = FastAPI(servers=[Server(url="google.com").model_dump()])
```

@RileyMShea, would you like to work on implementing this in case the idea has been accepted?

PR #3667, opened 2021-08-07
Refactor & Enhance multiple Functionality in FastAPI

yezz123:
- Replace if statement with if expression, Simplify if expression by using or
- Merge nested if conditions
- Simplify conditional into return statement
- Inline variable that is immediately returned
- Replace if statement with if expression, Simplify boolean if expression
- Simplify comparison to boolean, Simplify logical expression using De Morgan identities
- Replace if statement with if expression
- Merge nested if conditions, Replace unneeded comprehension with generator.
- Invert any/all to simplify comparisons
- Convert for loop into list comprehension, Inline variable that is immediately returned, Remove unnecessary else after guard condition
- Add guard clause
- Replace if statement with if expression, Replace unneeded comprehension with generator, Replace `list()`, `dict()` or `set()` with comprehension
- Replace `dict()` with `{}`
- Use `items()` to directly unpack dictionary values


pinchXOXO (approved):

Bernando32 (approved):

GuilleQP (approved):

fastapi/openapi/utils.py
```diff
@@ -296,11 +297,9 @@ def get_openapi_path(
                     deep_dict_update(openapi_response, process_response)
                     openapi_response["description"] = description
             http422 = str(HTTP_422_UNPROCESSABLE_ENTITY)
-            if (all_route_params or route.body_field) and not any(
-                [
-                    status in operation["responses"]
-                    for status in [http422, "4XX", "default"]
-                ]
+            if (((all_route_params or route.body_field))) and all(
```
louis-jaris, line 300:
```suggestion
            if (all_route_params or route.body_field) and all(
```

louis-jaris (approved):
Syntax wise and functionality wise, this seems to be identical, but performance wise I cannot tell

yezz123:
> Syntax wise and functionality wise, this seems to be identical, but performance wise I cannot tell

@louis-jaris Thank you 🚀

fastapi/openapi/utils.py
```diff
@@ -296,11 +297,9 @@ def get_openapi_path(
...
+                status not in operation["responses"]
+                for status in [http422, "4XX", "default"]
             ):
```
cikay, line 303: You can also encapsulate the conditions by a util function which will make it easier to understand the conditions by the function name and provide reusability.

BilalAlpaslan (approved):

tiangolo:
Thanks @yezz123! 🍰 

If you or someone still wants to have some of these changes, please make an individual PR for each change, including the reason why it would be needed. That way it will be easier to review, discuss, and merge each of the changes.

For now, and also as there are already a bunch of conflicts, I'll close this one. But thanks for the interest and effort! ☕

PR #3668, opened 2021-08-07
Enhancement: ResponseValidationError instead of ValidationError

### LINKED ISSUE #3520: ResponseValidationError with body
### First check

# Response Validation

Documentation suggests to implement pydantic's `ValidationError` handler to catch response validation errors. Problem is, there's no access to actual response, input of pydantic model, via this exception.

On the other hand, these errors may be much harder to spot, even to reproduce. Ideally, I'd want to log/sentry both request and failed response in this case, but as far as I know, there's no way I can do it via exception_handler mechanism currently. The only option -- manually validate and handle errors in every request handler, instead of:
```python
return raw_dict

@app.exception_handler(ResponseValidationError)
def on_invalid_resp(req: Request, exc: ResponseValidationError):
   # once per project
```

I need to:
```python
# copy&paste to each handler
try: 
      return MyResponseModel.parse_obj(raw_dict)
except ValidationError:
     ...
```

So maybe it'll make sense to create `ResponseValidationError` subclass of ValidationError, with `body` containing failed object, just like it is implemented for request validation?

gistart:
Closes #3520 

Sometimes it's hard to reproduce response validation error, especially if your app proxies or collects data from several external sources.

With new `ResponseValidationError` and it's `request_body` and `response_body` fields, users can install their own exception handlers to log/sentry both request and failed response.

This change is backward compatible with previously recommended `ValidationError` approach.


gistart:
#3520

tiangolo:
Thanks for the interest! This was handled in another PR at some point, now there's a `ResponseValidationError` that you can catch and inspect, it's not documented yet, but you can already use it. :nerd_face: 

Given that, for now, I'll close this one. But thanks for the effort! :cake:

seperman:
@tiangolo 
It doesn't seem we can catch the `ResponseValidationError` . In my case for example, the response object doesn't fit the response schema.
I get:

```py
ipdb> type(exc)
<class 'pydantic_core._pydantic_core.ValidationError'>
ipdb> isinstance(exc, ResponseValidationError)
False
```

```
fastapi==0.115.5
pydantic==2.9.2
pydantic_core==2.23.4
```

seperman:
@gistart Haha, I didn't realize this PR was from 3 years ago. Yeah that would be great. Thanks.

PR #3669, opened 2021-08-07
Join dependency execution paths before executing

adriangb:
This goes hand in hand with https://github.com/encode/starlette/pull/1260

The idea is to defer all execution and binding of arguments to dependencies to a central place using a series of closures/wrappers.

This provides better separation of the responsibilities of the DI system (executing the dependency w/ it's values) from concurrency management (binding sync deps to a thread pool, managing contexts, etc). I think evidence to this is that this system _can_ be annotated to carry dependency return type information through all of the wrapping (see [here](https://github.com/adriangb/anydep/blob/b18d45ade01d8993d348e72284dde43b8eb612c2/anydep/concurrency.py#L66-L93)).

This change should be 100% backwards compatible (unless we remove the now unused `contextmanager_in_threadpool` method, in which case this would not be backwards compatible, but it would be an easy fix for users).

### DIFF
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -1,7 +1,17 @@
-from typing import Any, Callable
+import contextvars
+import functools
+from typing import (
+    Any,
+    AsyncContextManager,
+    AsyncGenerator,
+    Awaitable,
+    Callable,
+    ContextManager,
+    TypeVar,
+)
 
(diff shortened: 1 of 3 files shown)


adriangb:
Since this likely won't get merged, I'm going to close those to avoid contributing to PR pile up

PR #3680, opened 2021-08-10
Add fine grained dependency support.

charlesoconor:
FastAPI has a number of optional dependencies that need to be installed
based on the use case. This change will create a extras-require for each
for of them. This will allow for users to install everything they need
for their service while keeping the requirements as slim as possible.

For example if a user need form parsing and the test client they can
install with `pip install fastapi[requests,python-multipart]`.

This will also help ensure the packages FastAPI depends on are in the
desired range.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ And now, go to <a href="http://127.0.0.1:8000/redoc" class="external-link" targe
 
 ### Recap
 
-In summary, you declare **once** the types of parameters, body, etc. as function parameters. 
+In summary, you declare **once** the types of parameters, body, etc. as function parameters.
 
 You do that with standard modern Python types.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -453,7 +453,7 @@ Used by FastAPI / Starlette:
 * <a href="https://www.uvicorn.org" target="_blank"><code>uvicorn</code></a> - for the server that loads and serves your application.
 * <a href="https://github.com/ijl/orjson" target="_blank"><code>orjson</code></a> - Required if you want to use `ORJSONResponse`.
 
-You can install all of these with `pip install fastapi[all]`.
+You can install all of these with `pip install fastapi[all]`. If you need a subset of dependencies from this list, you can also request them by name with `pip install fastapi[request,aiofiles,pyyaml]`. This will ensure that dependence installed will conform to the rages FastAPI expects.
```
embooglement, line 456: Typo in the last sentence: `This will ensure that dependence installed will conform to the rages FastAPI expects.` -> `This will ensure that the dependencies installed with conform to the ranges FastAPI expects`.

Also, I think you want to replace `request` with `requests` in the `pip install` example.

tiangolo:
I assume you solved it, thanks for closing the PR 👍

I want to do something like this, but first I have to solve and implement some things.

PR #3681, opened 2021-08-10
Upgrade Starlette to version 0.16.0

jerluc:
## Overview

This small change upgrades FastAPI's internal dependency on Startlette to version 0.16.0, which incorporates several important bugfixes (see release notes for [0.15.0](https://github.com/encode/starlette/releases/tag/0.15.0) and [0.16.0](https://github.com/encode/starlette/releases/tag/0.16.0)) that would be relevant for FastAPI users.

## Testing done

- [x] Run unit tests per the [contribution docs](https://fastapi.tiangolo.com/contributing/#tests)
- [x] Run a few examples from the FastAPI docs locally to ensure some "real-world" examples still work as expected

## Other notes

- During the upgrade, there were only a couple of unit tests that failed due to upstream breakage relating to websocket connections, which were pretty straightforward to resolve
- It's worth noting that Starlette now requires [AnyIO](https://anyio.readthedocs.io/) as of version 0.15.0; I don't claim to fully understand 100% of the reasons behind this switch, but from the commits I've read in Starlette, it seems to provide upstream support for different async runtimes (e.g. [Trio](https://trio-asyncio.readthedocs.io/)) along with other nice async I/O primitives such as [task groups](https://anyio.readthedocs.io/en/stable/tasks.html)

Apologies in advance if version upgrades aren't a typical part of outside contribution, but I figured this would be one way to get my feet wet!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 requires = [
-    "starlette ==0.14.2",
+    "starlette == 0.16.0",
     "pydantic >=1.6.2,!=1.7,!=1.7.1,!=1.7.2,!=1.7.3,!=1.8,!=1.8.1,<2.0.0"
 ]
 description-file = "README.md"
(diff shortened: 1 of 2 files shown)


Kludex:
Duplicated https://github.com/tiangolo/fastapi/pull/3594/

brnosouza:
any updates on this?

jerluc:
It looks like this is has been done in https://github.com/tiangolo/fastapi/pull/4016 and is available as of FastAPI v0.70.0

PR #3684, opened 2021-08-11
correctly calculate cache key for sub dependencies

### LINKED ISSUE #469: [BUG] Router level Security dependency blocks route level dependency
**Describe the bug**
If I define a route level `Security` dependency in a router, then define a `Security` dependency in the dependencies when adding the router to an app, then the route level dependency is ignored.

**To Reproduce**

Excerpted example.

A router defines a dependency with a scope.

```
data_router = APIRouter()                                                           
@router.get(                                                                   
    '/sales',                                                                  
    dependencies=[Security(get_current_user, scopes=['sales'])],               
    )                                                                          
def get_sales():
  pass
```

Another router includes that router, defining another dependency

```
from .data import data_router
api_router = APIRouter()                                                           
api_router.include_router(                                                      
    data_router,                                                                
    prefix='/data',                                                             
    dependencies=[Security(get_current_user, scopes=[])],                       
    )                                                                           
```

The `sales` scope does not appear in the `SecurityScopes` scopes when validating the scopes.  Defining `Depends(get_current_user)` instead has the same problem (my original version; not all routes require the same scope, but I have a fallback requirement for a signed in user at least).  Removing the dependency _entirely_, e.g. below, does get the correct scope in `SecurityScopes`.

```
from .data import data_router
api_router = APIRouter()                                                           
api_router.include_router(                                                      
    data_router,                                                                
    prefix='/data',                                                             
    )                                                                           
```

**Expected behavior**

The scopes from all `Security` definitions in the dependency tree should be required (e.g. any defined when including the router, and any on the routes that the router provides).

**Screenshots**
N/A

**Environment:**
 - OS: Ubuntu 19.04
 - FastAPI Version: 0.35.0
- Python version: 3.7.3

ojii:
This fixes https://github.com/tiangolo/fastapi/issues/469  (at least partially).

When sub dependencies are derived, the cache key would be calculated incorrectly if the sub dependency is a Security dependency with scopes. This change ensures the `cache_key` is re-calculated whenever its constituent parts are modified.

This solves the problem of `SecurityScopes` not having the correct value, but now the dependency passed to `Security` can be invoked multiple times (in the provided test it is invoked twice).


ojii:
@tiangolo is there anything I can do to have this looked at?

dimaqq:
Inquisitive minds want to know...

tiangolo:
Thanks for the interest and work @ojii! I'm trying this and I see that the test passes even without the changes. I'm also trying several examples locally copying things from that issue, but it seems to work, or I'm not understanding the problem correctly.

But I have the feeling that this was solved in https://github.com/tiangolo/fastapi/pull/2945

Could you check that one? That fix is available in FastAPI `0.80.0`

ojii:
@tiangolo that fix looks okay to me.

PR #3689, opened 2021-08-12
🌐  Add French translation for `docs/fr/docs/deployment/index.md`

rjNemo:
🤗 

Here is the PR to translate the deployment intro.
See the french translation tracking issue [here](https://github.com/tiangolo/fastapi/issues/1972).

Thanks for the reviews


Smlep (approved):

arthurio (approved):
Looks good 👍🏻 🇫🇷 
However, I just checked https://github.com/tiangolo/fastapi/blob/master/docs/en/docs/deployment/index.md and looks like there is more to translate now?

arthurio (changes requested):
I didn't see that it was out of sync with current english version, but the translation was great!

rjNemo:
@arthurio I review the recent version as soon as possible.
Thanks for pointing it out

docs/fr/docs/deployment/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+Le déploiement d'une application **FastAPI** est relativement simple.
+
+## Que signifie le déploiement
+
+**Déployer** une application signifie effectuer les étapes nécessaires pour la rendre **disponible pour les
+utilisateurs**.
+
+Pour une **API Web**, cela implique normalement de la placer sur une **machine distante**, avec un **programme serveur**
+qui offre de bonnes performances, une bonne stabilité, _etc._, afin que vos **utilisateurs** puissent **accéder** à
```
arthurio, line 11: Why `_ect._` instead of just `etc.`?

docs/fr/docs/deployment/index.md
```diff
@@ -0,0 +1,28 @@
...
+Ceci contraste avec les étapes de **développement**, où vous êtes constamment en train de modifier le code, de le casser
+et de le réparer, d'arrêter et de redémarrer le serveur de développement, _etc._
+
+## Stratégies de déploiement
+
+Il existe plusieurs façons de procéder, en fonction de votre cas d'utilisation spécifique et des outils que vous
+utilisez.
+
+Vous pouvez **déployer un serveur** vous-même en utilisant une combinaison d'outils, vous pouvez utiliser un **service
+en nuage** qui fait une partie du travail pour vous, ou d'autres options possibles.
```
arthurio, line 23: I think we should keep `"service cloud"`, WDYT?
Also, I would add `encore` at the end?

```suggestion
cloud** qui fait une partie du travail pour vous, ou encore d'autres options possibles.
```

arthurio (commented):
2 questions, otherwise looks great 🤩

docs/fr/docs/deployment/index.md
rjNemo: I learned that Latin terms should be written in italics.

arthurio: Thanks for the info 😃

arthurio (approved):
Great work! 💯

tiangolo:
Awesome, thanks @rjNemo! :cake: 

And thanks for the reviews @arthurio and @Smlep! :coffee:

PR #3698, opened 2021-08-13
✏ Tweak wording in `docs/en/docs/advanced/dataclasses.md`

pfackeldey:
Dear FastAPI developers,

this small PR improves the docs for data classes by adding a missing adjective.

Best, Peter

### DIFF
--- a/docs/en/docs/advanced/dataclasses.md
+++ b/docs/en/docs/advanced/dataclasses.md
@@ -8,7 +8,7 @@ But FastAPI also supports using <a href="https://docs.python.org/3/library/datac
 {!../../../docs_src/dataclasses/tutorial001.py!}
 ```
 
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for the interest! I'm actually not sure about this. I don't think that "valid" fully conveys the message here, I feel it makes it look like the code is valid or not, but it's not about having the code valid, but the fact that Pydantic has support for it. I would think the word "supported" could work better, but it is also used in the same sentence afterwards.

What do you think?

pfackeldey:
Dear @tiangolo,

yes, I fully agree! I will update the PR with your suggestion asap.

Best, Peter

tiangolo:
Cool, thanks @pfackeldey! :coffee:

PR #3724, opened 2021-08-19
Remove unnecesarry branch for default empty list

EdwardBlair:
```python
list(dependencies or [])
```

is identical behaviour to

```python
list(dependencies or []) or []
```

Is it not?

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -460,7 +460,7 @@ class APIRouter(routing.Router):
             ), "A path prefix must not end with '/', as the routes will start with '/'"
         self.prefix = prefix
         self.tags: List[str] = tags or []
-        self.dependencies = list(dependencies or []) or []
+        self.dependencies = list(dependencies or [])
         self.deprecated = deprecated
         self.include_in_schema = include_in_schema
         self.responses = responses or {}


fastapi/routing.py
```diff
@@ -460,7 +460,7 @@ def __init__(
             ), "A path prefix must not end with '/', as the routes will start with '/'"
         self.prefix = prefix
         self.tags: List[str] = tags or []
-        self.dependencies = list(dependencies or []) or []
+        self.dependencies = list(dependencies or [])
```
cikay, line 463: `self.dependencies = dependencies or []` this is good enough.

cikay (changes requested):

EdwardBlair:
I disagree

EdwardBlair:
self.dependencies was previously a List, I’d argue it’s more ‘safe’ to
explicitly convert the passed param to list than assume its type? (it’s
exposed by a less strict interface). Can make the change if you wish makes
no difference to me any more, I was relying on it being a List as I was
messing around creating shivs for the dependency system but I’ve since
given up.

tiangolo:
Hey there! It seems this was handled recently in another PR, so I'm gonna close this one. But thanks for the effort! :cake:

PR #3753, opened 2021-08-23
🔧 enable ansi in cmd

SayHelloRoman:
The Windows Console did not support ANSI escape sequences, nor did Microsoft provide any method to enable them.
But in 2016, Microsoft released the Windows 10 version 1511 update which unexpectedly implemented support for ANSI escape sequences, over two decades after the debut of Windows NT.
Windows Console Host used by Command Prompt support for character escape codes used by terminal-based software for Unix-like systems. This is not the default behavior and must be enabled programmatically with the Win32 API via `SetConsoleMode(handle, ENABLE_VIRTUAL_TERMINAL_PROCESSING)`

Before
Before
After
After

### DIFF
--- a/fastapi/logger.py
+++ b/fastapi/logger.py
@@ -1,3 +1,7 @@
 import logging
 
 logger = logging.getLogger("fastapi")
+
+if platform.system() == "Windows":
+    kernel32 = ctypes.windll.kernel32
+    kernel32.SetConsoleMode(kernel32.GetStdHandle(-11), 7)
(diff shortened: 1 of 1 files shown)


Kludex:
That's on uvicorn. You just need to install `colorama`, and it will be colored.

SayHelloRoman:
At the expense of colorama, this is an additional addiction that can be dispensed with. Also no matter what ASGI server we are using (gunicorn, Hypercorn, daphne, uWSGI, uvicorn), ansi escape will always be enabled.

ArcLightSlavik:
FastAPI shouldn't touch os related behaviours, this should be dealt with by servers. For example how uvicorn detects that it's on windows and doesn't run uvloop.

PR #3762, opened 2021-08-25
路由列表重复添加

LiNoZues:
app.include_router(xxx.router) 
多次调用 app.include_router且路由中存在 file:UploadFile = File(...)  会导致生成openapi.json的时候报 KeyError  报错位置位于 (fastapi/utils.py--> line 17  get_model_definitions() )


tiangolo:
Hey! Could you please create an issue with the example code that is causing you problems?

tiangolo:
Let's continue the conversation in the discussion to first figure out what's the problem and discover if/how this change would solve it. And then if it's the case, I can re-open this one. For now, I'll close it while we figure out what you're intending to solve. ☕

PR #3770, opened 2021-08-25
Add ability to set example=Undefined for models

### LINKED ISSUE #2149: [BUG] NoneType model value examples are not supported
### Description
I'm not able to pass a `None`/ as an example for a model value created with `pydantic.Field`

Looks like parser can't find a difference between  `example=None` as a NoneType variable and "_missing example field_"

```Python
class SomeResponse(BaseModel):
       city: Optional[str] = Field(..., nullable=True, example=None)
```

### Expected
```json
"city": {
"title": "City",
"type": "string",
"nullable": true,
"example": null
}
```

### Occurred 
```json
"city": {
"title": "City",
"type": "string",
"nullable": true
}
```

### Environment

* OS: macOS
* FastAPI Version: 0.61.1
* pydantic Version: 1.6.1

* Python version: 3.7.9

four43:
Allows models to have a "null" value in output Schema

Fixes #2149


four43:
yeah it was up to date but this project is a bit slow to review PRs ;)

tiangolo:

How different would it be this to setting `example=None`?

four43:
Hey @tiangolo - Thanks for stopping by! See the use case in the attached issue that this solves, #2149. This is an example of of "unknown/default" vs. "explicit null".

furiousteabag:
Hey everyone! What's the status of this fix? Would love to have explicit `null`s in the OpenAPI schema. How can I help?

YuriiMotov (changes requested):
@four43, thanks for contributing to FastAPI!

I think the general approach of this PR is nice, but it shouldn't be `Undefined`.
I think we should introduce a new value like `ExplicitNone` for this reason.

Also, the approach only works with Pydantic V1 and should be adjusted to work with V2.

What do you think?
Would you like to work on it?

PR #3789, opened 2021-08-30
Change "files" to "file" in tutorial001.py, because "files" is also used in tutorial002.py and causes confusion if you combine both..

mborus:
Call the endpoint "file" instead of "files" because it clashes with the multipart 
example 2, where the endpoint is also "files".

Since this example doesn't tell you how to write the form needed to test it, 
FastAPI users trying to upload files for the first time are likely to combine both examples 
into one project and then run into problems.

Also: For the simple, one file example, it might be helpful to add a curl example on how to upload a single file against this interface, because it's not easy to guess how to do that.I was thrown off the path by curl not automatically following the HTTP redirect I got from FastAPI, which required to add a "-L" parameter or an extra "/" at the end.

### DIFF
--- a/docs_src/request_files/tutorial001.py
+++ b/docs_src/request_files/tutorial001.py
@@ -3,7 +3,7 @@ from fastapi import FastAPI, File, UploadFile
 app = FastAPI()
 
 
-@app.post("/files/")
+@app.post("/file/")
 async def create_file(file: bytes = File(...)):
     return {"file_size": len(file)}
 


ArcLightSlavik (approved):
>Also: For the simple, one file example, it might be helpful to add a curl example on how to upload a single file against this interface, because it's not easy to guess how to do that.I was thrown off the path by curl not automatically following the HTTP redirect I got from FastAPI, which required to add a "-L" parameter or an extra "/" at the end.

It is assumed you would use the /docs endpoint and not curl directly.

mborus:
> It is assumed you would use the /docs endpoint and not curl directly.

I did not expect that the "/docs" endpoint provided a working curl example for file upload
(and chose to waste time researching how to do it manually instead...)

I just checked "/docs" and - of course - it has a working example there..,

cikay (commented):
The commit message subject should sum up the changes only. You can explain the reason in the body of the commit message. [Go ahead.](https://robertcooper.me/post/git-commit-messages)

tiangolo:
Thanks for your interest @mborus!

But I prefer to keep the consistency of making API paths with plurals to create a new item using POST there. An item by ID would go after the plural, e.g. `/files/{file_id}`, and reading all files would be `/files/` with GET, but a single file would be GET `/files/{file_id}`, and updating a single file would be PUT `/files/{file_id}`, etc.

Given that, I'll pass on this one, but thanks for the effort! :coffee:

Issue #3790, opened 2021-08-30
validation_error_definition schema is incorrect when detecting an error in an array


silversurfer34:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Optional, List

from fastapi import FastAPI
from pydantic import BaseModel

class SubItem(BaseModel):
    price: float

class Item(BaseModel):
    price_list: Optional[List[SubItem]] = None


app = FastAPI()


@app.post("/items/")
async def create_item(item: Item):
    return item
```

### Description

Run the sample program and call the /items/ api with this body:
```
{
  "price_list": [
    {
      "price": "zero"
    }
  ]
}
```
You will get a 422 like this 

```
{
  "detail": [
    {
      "loc": [
        "body",
        "price_list",
        0,
        "price"
      ],
      "msg": "value is not a valid float",
      "type": "type_error.float"
    }
  ]
}
```

But this 422 response does not follow the validation_error_definition schema :
```
validation_error_definition = {
    "title": "ValidationError",
    "type": "object",
    "properties": {
        "loc": {"title": "Location", "type": "array", "items": {"type": "string"}},
        "msg": {"title": "Message", "type": "string"},
        "type": {"title": "Error Type", "type": "string"},
    },
    "required": ["loc", "msg", "type"],
}
```

So the 422 response from fastapi does not respect the schema of the response... 

The issue comes from the error in the array, pydantic returns 0 as index of the error, but it is not a string, it is an integer. And it makes sense to return an integer, otherwise we wouldn't know if it is a key in a dict or an index.

As a developer I can workaround this issue by redefining validation_error_definition in my own code

```
from fastapi.openapi.utils import validation_error_definition

validation_error_definition["properties"] = {
    "loc": {
[...]

PR #3810, opened 2021-08-31
🐛 Fix JSON Schema for `ValidationError` at field `loc`

dconathan:
See #3790

Basically, Pydantic's schema for the `loc` field of the error detail is `Union[str, int]` not `str` for e.g. cases when the invalid data is at a particular index of an array:

See the type defined in Pydantic's source here:
https://github.com/samuelcolvin/pydantic/blob/5ccbdcb5904f35834300b01432a665c75dc02296/pydantic/error_wrappers.py#L12

The `openapi.json` schema for a `ValidationError` should match this.

This is particularly important for automated testing using https://schemathesis.readthedocs.io/en/stable/ because ValidationErrors that don't match the schema will lead to failing tests.

I've updated the `validation_error_definition` (h/t to @silversurfer34 for creating the issue and supplying the fixed schema 😄 ).
Also I had to update all the tests since the `openapi.json`s are hardcoded in that.


fastapi/openapi/utils.py
```diff
@@ -37,7 +37,11 @@
     "title": "ValidationError",
     "type": "object",
     "properties": {
-        "loc": {"title": "Location", "type": "array", "items": {"type": "string"}},
+        "loc": {
+            "title": "Location",
+            "type": "array",
+            "items": {"oneOf": [{"type": "string"}, {"type": "integer"}]},
```
Stranger6667, line 43: I'd suggest using `anyOf` instead (as Open API 3.0 doesn't support `type` as arrays) - it is slightly faster on the validation side. It doesn't require validating the input instance against all sub-schemas + these types are mutually exclusive (e.g. `number` vs `integer` might give different results under `anyOf` and `oneOf`).

silversurfer34 (approved):
Looks good to me

fastapi/openapi/utils.py
Stranger6667: Btw, the "Dog vs. Cat" example on that page is wrong :) Neither schema enforces `required` properties or restricts extra ones (via `additionalProperties: false`). E.g. `{"bark": true, "breed": "Dingo"}` is a valid `Cat` instance too!

tiangolo:
Awesome, thanks @dconatha for the work and explanation! :nerd_face: :cake: 

This will be available in FastAPI `0.75.2`, released in a couple of hours. :rocket:

PR #3840, opened 2021-09-01
🌐 Delete old Chinese translation for `docs/deployment.md`

jaystone776:
as title


tiangolo:
Good point, thanks @jaystone776! :rocket: :fire:

PR #3843, opened 2021-09-01
🌐 Update Chinese translation for `docs/tutorial/security/index.md`

jaystone776:
as title


tiangolo:
It seems this was updated at some point in another PR. Could you please check if there are changes still needed? If that's the case, please create a new PR. For now, I'll close this one, but thanks for the effort! 🍰

PR #3849, opened 2021-09-01
feat: support exclude_none in dataclass response models

sbelzile-nexapp:


fastapi/routing.py
```diff
@@ -57,6 +57,12 @@
 from starlette.websockets import WebSocket
 
 
+def get_dataclass_dict_factory(exclude_none: bool = False) -> Callable[[Any], Any]:
+    if exclude_none:
+        return lambda data: dict(x for x in data if x[1] is not None)
```
cikay, line 62: `{x for x in data if x[1] is not None}` is faster.

cikay (changes requested):

BilalAlpaslan (approved):

YuriiMotov (commented):
@sbelzile-nexapp, thanks for your efforts!

The usage of `response_model_exclude_none` is discouraged.
I think we shouldn't fix it and instead deprecate this parameter in the future

tiangolo:
Yep, as @YuriiMotov said. So, for now I'll pass on this one, but thanks for the interest! :cake:

PR #3851, opened 2021-09-01
fix a small mistake in en docs (bigger-applications)

goojal:
There is a mistake in some paths (probably caused by copy-paste from its previous section :grin:).


adam-tokarski (approved):
LGTM

szrg (approved):
Found it too.

PR #3874, opened 2021-09-08
🐛 Fix bug preventing to use OpenAPI when using tuples

victorbenichoux:
This PR is meant to fix issues (https://github.com/tiangolo/fastapi/issues/3665, https://github.com/tiangolo/fastapi/issues/3782) encountered generating `openapi.json` with nested models that have iterable-typed attributes (e.g. `Tuple` or `List`).

Briefly, this behavior did _not_ occur with 0.67.0, and was introduced in 0.68.0. Bisecting the issue led me to the following commit https://github.com/tiangolo/fastapi/commit/97fa743ecb5a716fa0223a7463cb83f3f0df494c. 

As per the error raised when using nested schemas of iterables:
```
pydantic.error_wrappers.ValidationError: 2 validation errors for OpenAPI
components -> schemas -> BuggyItem -> properties -> items -> items -> items
  value is not a valid dict (type=type_error.dict)
components -> schemas -> BuggyItem -> $ref
  field required (type=value_error.missing)
``` 
 I realized that there must be an issue when initializing the `OpenAPI` model, and specifically the `items` field of the `Schema`. This makes sense, since the commit introduced changes to the specification of the `OpenAPI` model.

And indeed, consider that we have the following `Model` to add to the openAPI specification:
```
class BuggyItem(pydantic.BaseModel):
    items: List[Tuple[str, str]]
```

We have to instantiate a `Schema` with 
```
{
  "items": {
    "title": "Items",
    "type": "array",
    "items": {
      "type": "array",
      "items": [{ "type": "string" }, { "type": "string" }]
    }
  }
}
```
In other words, the definition of the `Schema.items` attribute cannot be
```
    items: Optional["Schema"] = None
```
But must be
```
    items: Optional[Union["Schema", List["Schema"]]] = None
```
Which is the change introduced in this commit. 

Hopefully this is helpful, let me know if there is anything that I can add to this PR!

### DIFF
--- a/fastapi/openapi/models.py
+++ b/fastapi/openapi/models.py
@@ -123,7 +123,7 @@ class Schema(BaseModel):
     oneOf: Optional[List["Schema"]] = None
     anyOf: Optional[List["Schema"]] = None
     not_: Optional["Schema"] = Field(None, alias="not")
-    items: Optional["Schema"] = None
+    items: Optional[Union["Schema", List["Schema"]]] = None
     properties: Optional[Dict[str, "Schema"]] = None
     additionalProperties: Optional[Union["Schema", Reference, bool]] = None
     description: Optional[str] = None
--- a/tests/test_tuples.py
+++ b/tests/test_tuples.py
@@ -1,0 +1,267 @@
+from typing import List, Tuple
+
+from fastapi import FastAPI, Form
+from fastapi.testclient import TestClient
+from pydantic import BaseModel
+
+app = FastAPI()
+
+
+class ItemGroup(BaseModel):
+    items: List[Tuple[str, str]]
+
+
+class Coordinate(BaseModel):
+    x: float
+    y: float
+
+
+@app.post("/model-with-tuple/")
+def post_model_with_tuple(item_group: ItemGroup):
+    return item_group
+
+
+@app.post("/tuple-of-models/")
+def post_tuple_of_models(square: Tuple[Coordinate, Coordinate]):
+    return square
+
+
+@app.post("/tuple-form/")
+def hello(values: Tuple[int, int] = Form(...)):
+    return values
+
+
+client = TestClient(app)
+
+openapi_schema = {
+    "openapi": "3.0.2",
+    "info": {"title": "FastAPI", "version": "0.1.0"},
+    "paths": {
+        "/model-with-tuple/": {
+            "post": {
+                "summary": "Post Model With Tuple",
+                "operationId": "post_model_with_tuple_model_with_tuple__post",
+                "requestBody": {
+                    "content": {
+                        "application/json": {
+                            "schema": {"$ref": "#/components/schemas/ItemGroup"}
+                        }
+                    },
+                    "required": True,
+                },
+                "responses": {
+                    "200": {
+                        "description": "Successful Response",
+                        "content": {"application/json": {"schema": {}}},
+                    },
+                    "422": {
+                        "description": "Validation Error",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/HTTPValidationError"
+                                }
+                            }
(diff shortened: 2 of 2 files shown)


mrtolkien:
Had the issue and am waiting for a patch right now, would love to see this merged

RileyMShea:
I am also experiencing this. I don't see a clear workaround without this PR so I am remaining on 0.67 for the time being.

victorbenichoux:
Thanks for the comments @RileyMShea! 

Indeed, the generated `openapi.json` fails to render the `items` correctly in this case, and the `/docs` does not mention that `Tuple` is expected to be of fixed length (although it remains correctly validated when calling the endpoint, raising 422s appropriately). 

It does appear that this would in addition require a fix on the `pydantic` side which confuses fixed-length `Tuple` and variable length `List` in fields. 

That is, the following models have the same `pydantic.BaseModel.schema()` (up to the name):
```
class ListModel(pydantic.BaseModel):
    items: List[str]

class TupleModel(pydantic.BaseModel):
    items: Tuple[str]
```

When one would expect that the latter has a schema with `minLength` and `maxLength` set. I will raise an issue with them as I could not find anything related to this, and suggest a fix there.

Mause:
@victorbenichoux that actually isn't correct - openapi has a special syntax for tuples as you correctly detailed in your original post

victorbenichoux:
My bad then, I had not correctly read the initial issue (which I did not write though 😄 ). 

Then the solution would be to implement the [tuple validation for openapi](https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation) with `prefixItems`. 

I still feel, however, that it would be best to be implemented on the `pydantic` side, since `fastapi` does not generate this aspect of the schema.

victorbenichoux:
Something worth noting, as I mention in https://github.com/samuelcolvin/pydantic/issues/3210, is that the specification for Tuples in the `openAPI` spec is only available for versions `>3.1.0`, which, in turn, is not yet supported by swagger yet. As a result, I do not believe that we can really treat the `Tuple` case until swagger supports the new version of the openAPI spec.

This PR at least reverts the behavior to what it was in `0.67.0`: `/openapi.json` is generated correctly, although `Tuple` fields don't render correctly in the swagger.

HansBrende:
@RileyMShea @Mause : 
I agree with @victorbenichoux: even though the behavior in 0.67.0 was not _optimal_, it was better than it is _now_. The behavior _now_ is that the Swagger page CRASHES... making it completely useless and we (anyone using tuples) are all stuck on 0.67.0 until this PR gets merged & released.

tiangolo:
Thanks @victorbenichoux! :cake:

Good detective work! 🚀

And thanks everyone for the comments.

I added a lot of tests to check that using tuples doesn't break the rest of OpenAPI.

---

For anyone else paying attention to this, have in mind that Swagger UI doesn't support OpenAPI 3.1.0 yet. That is the first version that supports using tuples (also the most recent one), because it's based on JSON Schema draft 07.

They recently changed how JSON Schema defines tuples in the latest drafts as @victorbenichoux pointed out (with `prefixItems`). But the first thing that Swagger UI will support will be OpenAPI 3.1.0, using the way to define tuples from JSON Schema draft 07 (what this PR enables).

So, whenever Swagger UI supports tuples at some level it will be first with the current results. But again, have in mind it's not really supported by the docs UI yet and they won't be properly rendered.

That aside, because the internal models were defining the OpenAPI models so tightly, they didn't allow using tuples with OpenAPI and Swagger UI (because, technically it's not supported by the spec :sweat_smile: ). But with this more relaxed version now you can continue using tuples, even if they are not properly rendered in Swagger UI, but you can still benefit from the rest of OpenAPI and Swagger UI.

---

This is available in FastAPI `0.73.0`, released in the next hours. :tada:

PR #3892, opened 2021-09-11
✨ Update `ORJSONResponse` to support non `str` keys and serializing Numpy arrays

baby5:
the detail see: https://github.com/ijl/orjson/issues/197

my case is `<class 'sqlalchemy.sql.elements.quoted_name'>`


adam-tokarski (changes requested):
Please provide any tests for that.

baby5:
@adam-tokarski done

tests/test_default_response_class.py
```diff
@@ -31,6 +32,12 @@ def get_root():
     return {"msg": "Hello World"}
 
 
+@app.get("/orjson_non_str_keys")
+def get_orjson_non_str_keys():
+    key = quoted_name("msg")
+    return {key: "Hello World"}
```
adam-tokarski, line 38: Isn't that `key` still a string? It seems like it should be working even before the fix (I didn't check that though).

Also, you could provide a test with some integer as a key maybe?

adam-tokarski (approved):

tiangolo:
Great, thank you @baby5! And thanks for adding tests! :cake: 

And thanks @adam-tokarski for the help and review! :coffee:

NikosAlexandris:
What is the status of this issue regarding support for NumPy arrays ?

PR #3902, opened 2021-09-14
#639 Run dependencies in parallel

ariloulaleelay:
Fix for this issue: https://github.com/tiangolo/fastapi/issues/639


yezz123 (approved):
Great Work 🙏🏻, looks like a good PR that gonna fix this issue relates to executing dependencies in parallel.

adriangb:
I only just learned that Python 3.9 added some graph tooling to the standard library: https://docs.python.org/3/library/graphlib.html Just in case you hadn't seen it @ariloulaleelay, I certainly hadn't. If you had, sorry for the noise!

fastapi/dependencies/utils.py
```diff
@@ -433,184 +516,217 @@ def is_gen_callable(call: Callable[..., Any]) -> bool:
...
-            value, values, loc=(field_info.in_.value, field.alias)
-        )
-        if isinstance(errors_, ErrorWrapper):
-            errors.append(errors_)
-        elif isinstance(errors_, list):
-            errors.extend(errors_)
-        else:
-            values[field.name] = v_
-    return values, errors
+    async with anyio.create_task_group() as task_group:
```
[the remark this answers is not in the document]
ariloulaleelay: Thanks for benchmarking. I've made some optimization for simple cases.
Also i added special case for sync functions, to avoid threadpool execution (for simple cases without IO).
Here is your benchmark example updated:
```python
try:
    from fastapi.dependencies.utils import sync_dependency
except ImportError:
    def sync_dependency(f):
        return f

app = FastAPI()

Dep0 = Depends(sync_dependency(lambda: "dep0"))
Dep1 = Depends(sync_dependency(lambda: "dep1"))
Dep2 = Depends(sync_dependency(lambda: "dep2"))
Dep3 = Depends(sync_dependency(lambda: "dep3"))
Dep4 = Depends(sync_dependency(lambda: "dep4"))
Dep5 = Depends(sync_dependency(lambda: "dep5"))
Dep6 = Depends(sync_dependency(lambda: "dep6"))
Dep7 = Depends(sync_dependency(lambda: "dep7"))
Dep8 = Depends(sync_dependency(lambda: "dep8"))
Dep9 = Depends(sync_dependency(lambda: "dep9"))


@app.get("/")
async def read_main(
    dep0: str = Dep0,
    dep1: str = Dep1,
    dep2: str = Dep2,
    dep3: str = Dep3,
    dep4: str = Dep4,
    dep5: str = Dep5,
    dep6: str = Dep6,
    dep7: str = Dep7,
    dep8: str = Dep8,
    dep9: str = Dep9
):
    return {"msg": "Hello World"}


client = TestClient(app)


def test_read_main():
    response = client.get("/")
    assert response.json() == {"msg": "Hello World"}


if __name__ == '__main__':
    import timeit
    print(timeit.timeit("test_read_main()", globals=globals(), number=3000))
    
# mine: 3.93
# 0.70.0: 8.98
# 0.69.0: 8.98
# 0.68.2: 9.06
```
I got 2x performance gain for synthetic case :)

fastapi/dependencies/utils.py
```diff
@@ -84,7 +85,22 @@
 )
 
 
+class CallType(IntEnum):
+    ASYNC = 0
+    GENERATOR = 1
+    THREADPOOL = 2
+    SYNC = 3
+
+
```
cikay, line 94: I suggest assigning string values rather than integers to evade conflicts.

fastapi/dependencies/utils.py
```diff
@@ -135,7 +167,10 @@ async def task(
 
             parent_node.refcount -= 1
             if parent_node.refcount == 0:
-                task_group.start_soon(parent_node.task, nodes, request, task_group)
+                if task_group is not None:
+                    task_group.start_soon(parent_node.task, nodes, request, task_group)
+                else:
+                    await parent_node.task(nodes, request, task_group)
```
cikay, line 173: Decrease the indentation level as much as possible. 

```py
if parent_node.refcount != 0:
    return

if task_group is not None:
    task_group.start_soon(parent_node.task, nodes, request, task_group)
else:
    await parent_node.task(nodes, request, task_group)
```

fastapi/dependencies/utils.py
```diff
@@ -168,6 +203,17 @@ def __init__(
         self.security_scopes_param_list = security_scopes_param_list
         self.body_params_list = body_params_list
 
+        self.sync_tasks_count = sum(
+            1 if node.call_type == CallType.SYNC else 0
```
cikay, line 207: Evade if statement if it is possible.
`int(node.call_type == CallType.SYNC)`

cikay (changes requested):

fastapi/dependencies/utils.py
ariloulaleelay: This is internal enum, no one will use it. IntEnum might be slightly faster.

ariloulaleelay:
Hey @tiangolo just checking status of this PR. Is there any suggestions for it.

fastapi/dependencies/utils.py
```diff
@@ -86,6 +85,138 @@
...
+        self.background_tasks_param_list = background_tasks_param_list
+        self.http_connection_param_list = http_connection_param_list
+        self.request_param_list = request_param_list
+        self.websocket_param_list = websocket_param_list
+        self.response_param_list = response_param_list
+        self.security_scopes_param_list = security_scopes_param_list
+        self.body_params_list = body_params_list
+
+        self.sync_tasks_count = sum(
+            1 for node in nodes
```
cikay, line 209: Use variable binding to make 1 clear what it is.

fastapi/dependencies/utils.py
```diff
@@ -435,184 +566,221 @@ def is_gen_callable(call: Callable[..., Any]) -> bool:
...
+                call=overrided_call,
+                name=dependant.name,
+                security_scopes=security_scopes,
+            )
+            dependant.security_scopes = security_scopes
+
+        dependant.cache_key = cast(Tuple[Callable[..., Any], Tuple[str]], dependant.cache_key)
+
+        if dependant.use_cache and dependant.cache_key in node_cache:
+            node = node_cache[dependant.cache_key]
```
cikay, line 619: Move it to the below if block.

fastapi/dependencies/utils.py
ariloulaleelay: Can you give example?
1 is just 1, I need to count nodes with some condition.

cikay: Then I suggest filtering sync_call_type_node and passing its length to the function.

ariloulaleelay: Such technique will create unneeded array:
```
>>> import resource
>>> val = len([1 for i in range(int(1e8)) if i % 2 == 0])
>>> resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
401112
```
vs
```
>>> import resource
>>> val = sum(1 for i in range(int(1e8)) if i % 2 == 0)
>>> resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
10280
```

cikay: Okay.

PR #3913, opened 2021-09-17
Add support for Iterator in jsonable_encoder

### LINKED ISSUE #3911: FastAPI response_model unable to handle Iterable types -- returns only the model keys
### First Check

### Commit to Help

### Example Code

```python
from typing import Any, Iterable, List

# Define simple model
class Simple(BaseModel):
    spam: int
    eggs: str

# instantiate fast api and route
loader = FastAPI(
    version="0",
)

requests_router = APIRouter(prefix="/requests", tags=["requests"])

@requests_router.get("/", response_model=Iterable[Simple])
def get_requests() -> Iterable[Simple]:
    return [
        Simple(
            spam=1,
            eggs="a"
        )
    ]

loader.include_router(requests_router)

###
### Output at http://127.0.0.1:8000/requests/ is 
### {
###     "spam": "eggs"
### }
###
```

### Description

To reproduce

* Run uvicorn api2:loader --reload
* Go to http://127.0.0.1:8000/requests/
* Observe only the keys in the data model are returned

To fix
* Change the `response_model` to `response_model=List[Simple]`
* Go to http://127.0.0.1:8000/requests/
* Observe you get the correct response:
```json
[
    {
        "spam": 1,
        "eggs": "a"
    }
]
```

I haven't been able to find this in any documentation or bugs, but maybe my google skills are lacking

```
uvicorn            0.15.0
fastapi            0.68.1
```

### Operating System

Windows

### Operating System Details

Windows 11 21H2 (OS Build 22000.160))

### FastAPI Version

0.68.1

### Python Version

3.9.1

### Additional Context

This has a workaround, but I don't know the code base well enough to understand why an `Iterable` type would cause this problem

klaa97:
This PR adds support for `Iterator` in the `jsonable_encoder`.

Currently, FastAPI allows for `response_model` such as `Iterable[BaseModel]` . In those cases:
1. The Pydantic Model is used to retrieve the validated data from the response object https://github.com/tiangolo/fastapi/blob/20d48345465e5b09756cc9ef62fb940c63255a13/fastapi/routing.py#L129
2. The output of the validation is a List Iterator
3. The `jsonable_encoder` doesn't encode the list, but just calls `dict()` and returns a list of the fields of the `BaseModel`

By adding the support for `Iterator` in the `jsonable_encoder`, we avoid that the response object is passed to `dict()`, losing the values of the fields of the BaseModel.

An example of this issue is described in #3911 

Fixes #3911


fastapi/encoders.py
```diff
@@ -95,7 +95,7 @@ def jsonable_encoder(
                 )
                 encoded_dict[encoded_key] = encoded_value
         return encoded_dict
-    if isinstance(obj, (list, set, frozenset, GeneratorType, tuple)):
+    if isinstance(obj, (list, set, frozenset, GeneratorType, tuple, Iterator)):
```
interifter, line 98: Could we remove `list`, `set`, `frozenset`, `GeneratorType`, and `tuple`? I believe they are all iterables.

interifter: Local testing verifies **list, set, and tuple** are all Iterators. But `frozenset` is not. I haven't been able to check on `GeneratorType`. Operating with a very restricted setup at the moment.

interifter, line 98: More-proper question: Why don't we just make this `isinstance(obj, Iterable)`?

```python
from typing import Iterable

def gen(arr):
    for x in arr:
        yield x

l = [1, 2, 3, 4]
g = gen(l)
s = set(l)
t = tuple(l)
f = frozenset(l)

print(isinstance(g, Iterable))
print(isinstance(l, Iterable))
print(isinstance(s, Iterable))
print(isinstance(t, Iterable))
print(isinstance(f, Iterable))

# Output is all True
```

klaa97: The problem is exactly the one you guessed - Iterable makes some tests fail (https://github.com/tiangolo/fastapi/blob/22528373bba6a654323de416ad5c867cbadb81bb/tests/test_jsonable_encoder.py#L102). On the other `Iterator` maintains the current interface and fix the issue you opened, and in general the behaviour with Iterable[] types.

interifter:
What do we need to do in order to get this approved and merged?

klaa97:
> What do we need to do in order to get this approved and merged?

Tagging randomly and asking a review to @Kludex @dmontagu?
I have 2 PRs opened for some months here, and it's not clear to me who to ask for a review for bug fixes.

Thank you 🙏

Kludex:
@tiangolo is the only one who can have this merged. Nothing to do besides wait. :kissing:

iudeen (approved):
Looks good, and I working fine.

Just a curious question:

Why don't we consider all iterable, sequence instead of explicitly specifying each types?

interifter:
> Looks good, and I working fine.
> 
> Just a curious question:
> 
> Why don't we consider all iterable, sequence instead of explicitly specifying each types?

See https://github.com/tiangolo/fastapi/pull/3913#discussion_r712343889

> Why don't we just make this `isinstance(obj, Iterable)`? Is that going to be too generic and pick up bad inputs?
> 
> ```python
> from typing import Iterable
> 
> def gen(arr):
>     for x in arr:
>         yield x
> 
> l = [1, 2, 3, 4]
> g = gen(l)
> s = set(l)
> t = tuple(l)
> f = frozenset(l)
> 
> print(isinstance(g, Iterable))
> print(isinstance(l, Iterable))
> print(isinstance(s, Iterable))
> print(isinstance(t, Iterable))
> print(isinstance(f, Iterable))
> 
> # Output is all True
> ```

yezz123 (approved):

tiangolo:
Thanks @klaa97!

I'm struggling to find what use case would this solve that would not be solved by setting the response model to `list[X]`, as described in the same discussion.

As there are many other things that could be `Iterable` and `Iterator` and it has nuanced behaviors, I would be inclined to leave just `list` and the others. Also, it's normally better to be more explicit in the output/return about the exact type.

It would be better to start with a use case that shows why this is needed and why using a `list[X]` is not enough, filling the GitHub Discussion question form: https://github.com/tiangolo/fastapi/discussions/new?category=questions 🙏

tiangolo:
Assuming the original need was handled, this will be automatically closed now. But feel free to add more comments or create new issues or PRs.

interifter:
> Assuming the original need was handled, this will be automatically closed now. But feel free to add more comments or create new issues or PRs.
I think so.

I ended up using a list, but the ask to support Iterator seems pretty no-nonsense, given it allows for more flexibility. After two years, I had figured this would never get merged and a list seems to work well enough.

PR #3925, opened 2021-09-21
quote pip install command for zsh compatibility

zeke:
Hey there. 👋🏼 

I'm a zsh user, making my way through the tutorial. I noticed that typing `pip install fastapi[all]` doesn't work in my shell because square brackets have some special meaning in zsh.

This PR updates the tutorial to wrap the commands in quotes, which should work on all shells.

Without quotes:

```
$ pip install fastapi[all]
zsh: no matches found: fastapi[all]
```

With quotes:

```
$ pip install "fastapi[all]"
Collecting fastapi[all]
  Downloading fastapi-0.68.1-py3-none-any.whl (52 kB)
  ...
```

### DIFF
--- a/docs/en/docs/tutorial/index.md
+++ b/docs/en/docs/tutorial/index.md
@@ -43,7 +43,7 @@ For the tutorial, you might want to install it with all the optional dependencie
 <div class="termy">
 
 ```console
-$ pip install fastapi[all]
+$ pip install "fastapi[all]"
 
 ---> 100%
 ```


Kludex (approved):

ArcLightSlavik (approved):

tiangolo:
Thanks @zeke! This was covered in https://github.com/tiangolo/fastapi/pull/3131, so I'll close this one. But thanks for the effort! :coffee:

PR #3948, opened 2021-09-27
option to hide curl from swagger ui

sadakmed:
what?
The following PR adds the option to hide the curl response.

why?
For long requests, either long vectors, or just multiple entries. The curl response covers the whole page, it's unpleasant for a show case, a client testing the api, or as a developer u need to scroll.

How?
adding a wrap component as a pluging in SwaggerUIBundle

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -65,6 +65,7 @@ class FastAPI(Starlette):
         callbacks: Optional[List[BaseRoute]] = None,
         deprecated: Optional[bool] = None,
         include_in_schema: bool = True,
+        hide_curl: bool = False,
         **extra: Any,
     ) -> None:
         self._debug: bool = debug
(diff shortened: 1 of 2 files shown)


ghandic:
@sadakmed could you mark the discussion as answered if my solution was acceptable?

https://github.com/tiangolo/fastapi/discussions/3853

ghandic:
I suggest that the PR be slightly more generic, so people can add pre_js and then add in custom options into swaggerui  similar to how Folium wraps leafletjs https://github.com/python-visualization/folium

sadakmed:
> add pre_js and then add in custom options

didn't quite understand this, do u suggest that the input should be a str of js code?

ghandic:
The idea I was suggesting is to allow someone to inject custom js (to allow plugins) then they could refer to them in the swaggerui js initialization options.

yezz123 (approved):
LGTM+1 🚀, its a good one can be used to reduce some panels in the SwaggerUI

BilalAlpaslan (approved):

tiangolo:
Thanks for the interest @sadakmed! :cake: 

And thanks for the discussion everyone. :coffee: :cookie: 

This is something that not necessarily everyone would need, and it would probably be easier to teach how to implement it manually than to have internal support for it, which also means that everyone that wants to learn FastAPI would have another parameter to learn and understand.

And especially because Swagger UI could be so customizable and there are so many things that could be modified, I would rather not add more abstractions on top to tweak it through FastAPI. I think that for cases that need customization there, it makes more sense to handle them directly.

Given that, I'll pass on this one, but thanks for the interest!

PR #3949, opened 2021-09-27
Include frozenset in tutorial

theo-brown:
The tutorial section on 'Extra Data Types' did not include an example of frozenset.


tiangolo:
Thanks for the interest and effort! :nerd_face: 

I wasn't trying to include all the possible data types supported by Pydantic.

But still, I'm curious, in which cases do you think a frozen set would be required or particularly useful?

I haven't found a use case where I need it, also that justifies explaining it in the docs. :shrug: 

PR #3953, opened 2021-09-28
Catching exceptions in the dependancy call - eg for pydantic validati…

ghandic:
…on error to return a 422

Just a first pass, will need tests adding etc, but this aims to solve a few issues for deepObject encoding (where you need to use Depends currently), also in response to the issue raised in the discussion https://github.com/tiangolo/fastapi/discussions/3426#discussioncomment-1391836

### DIFF
--- a/docs/en/docs/release-notes.md
+++ b/docs/en/docs/release-notes.md
@@ -19,6 +19,11 @@ This way, in case there was a breaking change for your code in one of the releas
 
 * ⬆️ Upgrade Starlette to 0.16.0. PR [#4016](https://github.com/tiangolo/fastapi/pull/4016) by [@tiangolo](https://github.com/tiangolo).
 
+Also upgrades the ranges of optional dependencies:
+
+* `"jinja2 >=2.11.2,<4.0.0"`
+* `"itsdangerous >=1.1.0,<3.0.0"`
+
 ## 0.69.0
 
 ### Breaking Changes - Upgrade


fastapi/dependencies/utils.py
```diff
@@ -535,25 +535,29 @@ async def solve_dependencies(
...
+                )
+            elif is_coroutine_callable(call):
+                solved = await call(**sub_values)
+            else:
+                solved = await run_in_threadpool(call, **sub_values)
+            if sub_dependant.name is not None:
+                values[sub_dependant.name] = solved
+            if sub_dependant.cache_key not in dependency_cache:
+                dependency_cache[sub_dependant.cache_key] = solved
+        except Exception as e:
```
alukach, line 559: Is catching all `Exception` errors overly greedy? Perhaps we should only handle `ValueError` and `TypeError`, as those seem to be the two bases for all of Pydantic's validation errors (I may be wrong on that, I'm not that well versed in Pydantic):

https://github.com/samuelcolvin/pydantic/blob/c256dccbb383a7fd462f62fcb5d55558eb3cb108/pydantic/errors.py#L128-L133
```suggestion
        except (ValueError, TypeError) as e:
```

alukach:
@ghandic What are your thoughts on this feature idea?  Do you still think it's valid but just not a priority for you, or was there a particular reason to abandon this PR?

ghandic:
I need to reopen this from a feature branch as it was on my master branch on my fork... will aim to do it in the next few days

PR #3959, opened 2021-09-28
memo: fix typo in requests path

rpalloni:


Kludex:
Nop. 😗

rpalloni:
@Kludex is the proposed edit wrong? Is /items/{item_id}  correct also in the second path?

Kludex:
Yes and yes. It's a reference to the above endpoint, which has "q" as query parameter.

Issue #3359, opened 2021-06-10
example argument not used in docs ui when using multiple Body() params


tobiasdam:
### Example

```python
from fastapi import Body, FastAPI

app = FastAPI()


@app.post("/")
def test(a: str = Body(..., example="A"), b: str = Body(..., example="B")):
    pass
```
This results in the following example value for the request body in the docs ui (`/docs`):
```json
{
  "a": "string",
  "b": "string"
}
```

### Description
* Open the browser and call the endpoint `/docs`.
* When using multiple `Body()` params, the example value for the request body consists of the individual schemas and ignores the provided examples.
* But I expected it to create the following example value:
```json
{
  "a": "A",
  "b": "B"
}
```

### Environment
* Docker image: tiangolo/uvicorn-gunicorn-fastapi:python3.8-2021-06-09
* OS: Linux
* FastAPI Version: 0.65.2
* Python version: 3.8.6

johnLenon1l083:
![286FAF2A-4C24-48FD-A92D-921CA2F880F9](https://user-images.githubusercontent.com/83523194/121815717-7114c980-cc80-11eb-8878-0c66877ad9cc.png)

johnLenon1l083:
Вы следующие

cointreau:
it still remains on 0.85.0. When I have a string parameter and its example, it works. 
But if I have multiple string parameter and examples, it just shows "string".

tiangolo:
Ah, yeah, if you want one example you should have one single Pydantic model and not several independent `Body()` parameters, that would create independent example objects but Swagger UI can only show one example object.

> Sorry for the long delay! 🙈  I wanted to personally address each issue/PR and they piled up through time, but now I'm checking each one in order.

PR #3971, opened 2021-10-01
Fix OpenAPI examples from Body

jmriebold:
FastAPI documentation states that examples can be added to the OpenAPI docs in three different ways. Two of these ways ([schema_extra](https://fastapi.tiangolo.com/tutorial/schema-extra-example/?h=exampl#pydantic-schema_extra) and the [`example` kwarg](https://fastapi.tiangolo.com/tutorial/schema-extra-example/?h=exampl#field-additional-arguments)) are currently broken, as identified in #3359, a bug introduced by PR #1267 (support for multiple named examples in OpenAPI docs). This PR fixes this behavior by passing the `example` kwarg through to the Pydantic superclass. Note that this still allows for the possibility of overriding model examples using the [`examples` kwarg](https://fastapi.tiangolo.com/tutorial/schema-extra-example/?h=exampl#body-with-multiple-examples).

### DIFF
--- a/fastapi/params.py
+++ b/fastapi/params.py
@@ -246,6 +246,8 @@ class Body(FieldInfo):
         self.embed = embed
         self.media_type = media_type
         self.example = example
+        if self.example != Undefined:
+            extra["example"] = self.example
         self.examples = examples
         super().__init__(
             default,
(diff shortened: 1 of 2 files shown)


jmriebold:
Any thoughts on this PR? This is still an issue in the latest release, and I don't see any other way to restore previous functionality.

PR #3978, opened 2021-10-03
Python

RChandana:
Want applications of MONAI using Python


ghandic:
I'm confused what this PR is intended for? Could you describe it in more detail?

tiangolo:
I'm not sure what you intend with this PR, I imagine it was probably a mistake, so I'm closing it.

PR #3979, opened 2021-10-03
💚 Fix badges in README and main page

ghandic:
Fixing workflow badge to point to the master branch and not forks with PR. Looking into how to do the same for codecov


tiangolo:
Awesome, thank you @ghandic! :rocket: 

I just updated the main page as well, the README is generated from it, so now tests should pass. Thanks for your contribution! :cake:

ghandic:
You'll be measuring the PR coverage and displaying that on the main page though. I did have a look to see if there was some filtering that could be done for that badge but couldn't see anything at a first pass.

I'll continue to look 👍🏻

ghandic:
We could add flags conditional of whether it is a PR or a push and then filter by that flag for the badge. I'll make another PR today to fix, that way you'll keep the codecov run for PR to check but also have the coverage % accurate on the readme

https://community.codecov.com/t/badges-for-flags/270/5

ghandic:
@tiangolo see PR - https://github.com/tiangolo/fastapi/pull/3982

PR #3981, opened 2021-10-03
📝 Update GraphQL docs, recommend Strawberry

tiangolo:
📝 Update GraphQL docs, recommend [Strawberry](https://strawberry.rocks/)


ghandic:
@tiangolo there were some related questions for how to handle dependencies when using graphQL as `add_route` just calls the starlette `add_route` and skips out the dependencies https://github.com/tiangolo/fastapi/issues/3968

PR #3982, opened 2021-10-03
Added flags to coverage upload depending if it is a PR or a push and …

ghandic:
…updated badge icon to point only to master flag

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,4 +35,12 @@ jobs:
       - name: Test
         run: bash scripts/test.sh
       - name: Upload coverage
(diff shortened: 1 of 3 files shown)


tiangolo:
Thanks for your interest @ghandic! I was checking that there's no need to use flags, it's possible to use the same current badges pointing to `master`.

But still, it seems that leaving them as they are they already point to master. I just created a broken PR here: https://github.com/tiangolo/fastapi/pull/4887 removing lots of tests, that would reduce the coverage. But the badges still seem to show 100%, so it seems that by default they show the percentage of `master`.

Given that, I'll close this PR now, but thanks for the effort!

PR #3987, opened 2021-10-05
⬆️ Increase dependency ranges for tests and docs: pytest-cov, pytest-asyncio, black, httpx, sqlalchemy, databases, mkdocs-markdownextradata-plugin

tiangolo:
⬆️ Upgrade dependency ranges


ghandic:
@tiangolo youre killin it today with the merges 💪🏻

PR #4015, opened 2021-10-07
📝 Update `docs/tutorial/dependencies/classes-as-dependencies`: Add type of query parameters in a description of `Classes as dependencies`

0417taehyun:
The [description](https://fastapi.tiangolo.com/tutorial/dependencies/#create-a-dependency-or-dependable) of `common_parameters` of parameters in `Dependecies - First steps` has types like "An optional query parameter `q` that is a `str`."

I think it is better to add type in a description of [Class as dependencies](https://fastapi.tiangolo.com/tutorial/dependencies/classes-as-dependencies/#classes-as-dependencies_1) in `Classes as Dependencies` too, like "An optional `q` query parameter that is `str`".


ArcLightSlavik (approved):

TestRights (commented):
thanks

TestRights (approved):

PR #4028, opened 2021-10-09
Fix specifying custom_encoder on a model with json_encoders modifies json_encoders

segevfiner:
Check the added test to see the issue. Basically using `custom_encoder` with a model that has `json_encoders` causes it to update the model's `json_encoders` with the `custom_encoder`, rather than making a copy.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -44,7 +44,7 @@ def jsonable_encoder(
     if isinstance(obj, BaseModel):
         encoder = getattr(obj.__config__, "json_encoders", {})
         if custom_encoder:
-            encoder.update(custom_encoder)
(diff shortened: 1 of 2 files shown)


fastapi/encoders.py
```diff
@@ -44,7 +44,7 @@ def jsonable_encoder(
     if isinstance(obj, BaseModel):
         encoder = getattr(obj.__config__, "json_encoders", {})
         if custom_encoder:
-            encoder.update(custom_encoder)
+            encoder.copy().update(custom_encoder)
```
tiangolo, line 47: If I'm reading the code correctly, `encoder.copy()` would create another object. Then it would update that new object, but it would not be used anywhere else. Right?

PR #4042, opened 2021-10-12
Add default API error schema

### LINKED ISSUE #3650: Provide the means to globally override the default 422 OpenAPI schema
### First Check

### Commit to Help

### Example Code

```python
from starlette import status


class Example(BaseModel):
    value: int


class Error(BaseModel):
    message: str


app = FastAPI()


async def validation_exception_handler(request: Request, exc: RequestValidationError):
    error = Error(message="Validation failed")
    return JSONResponse(error.json(), status_code=status.HTTP_422_UNPROCESSABLE_ENTITY)


@app.post("/", response_model=Example)
def root(example: Example = Body(...)):
    return example
```

### Description

Overriding the default exception handlers is easy and [well documented](https://fastapi.tiangolo.com/tutorial/handling-errors/#override-the-httpexception-error-handler). Unfortunately this has a significant caveat in regards to the generated OpenAPI specifications, the schema does not match the actual responses.

### Wanted Solution

Provide a parameter or function that can override the default 422 schema model.

### Wanted Code

```python
from starlette import status


class Example(BaseModel):
    value: int


class Error(BaseModel):
    message: str


# Provide error model as parameter
app = FastAPI(
    validation_error_model=Error,
    validation_error_description="Returns an object with a simple error message",
    validation_error_status=400,
)


# or set via fields
app.validation_error_model = Error
app.validation_error_description = "Returns an object with a simple error message"
app.validation_error_status = 400


async def validation_exception_handler(request: Request, exc: RequestValidationError):
    error = Error(message="Validation failed")
    return JSONResponse(error.json(), status_code=status.HTTP_422_UNPROCESSABLE_ENTITY)


@app.post("/", response_model=Example, responses={422: {"model": Error}})
def root(example: Example = Body(...)):
    return example
```

### Alternatives

Manually add a response model for 422 to every endpoint that can raise a RequestValidationError. This is suitable for a small API but will lead to problems as the application grows and maintenance is required.

### Operating System

Linux

### Operating System Details

DISTRIB_ID=ManjaroLinux
DISTRIB_RELEASE=21.0.7
DISTRIB_CODENAME=Ornara
DISTRIB_DESCRIPTION="Manjaro Linux"

### FastAPI Version

0.67.0

### Python Version

Python 3.9.5

### Additional Context

_No response_

caiofauza:
Closes #3650 

We implemented a way to define a default API error schema, as suggested on #3650. 
The error object can be passed in the constructor of the FastAPI class (optional parameter) or defined by variable attribution.

Example code:
```py from fastapi import FastAPI

class Response(BaseModel):
    status: bool
    
default_error = {
    'status': 400,
    'name': 'DefaultError',
    'description': 'DefaultError',
    'properties': {
        'message': {'title': 'Message', 'type': 'string'},
        'metadata': {'title': 'Metadata', 'type': 'string'}
    },
    'required': ['message']
}

app = FastAPI(default_error_schema=default_error)

@app.get('/status', response_model=Response)
def retrieve_status():
    return { 'status': True }
```

image

@zMendes @Gustavobb


YuriiMotov (changes requested):
@caiofauza, thanks for your work!

I really like the idea of this PR and it works, but there are a few things that need to be done to get it approved and merged:
* the type of `default_error_schema` is declared as `Optional[DefaultErrorSchema]`, but in fact it's `Optional[dict]` (we pass it as `dict` and it's handled as `dict` in code). Typing needs to be fixed. It can be done by converting `DefaultErrorSchema` into `TypedDict`. Another option is to accept both `dict[str, Any]` or `DefaultErrorSchema` and handle it in code.
* `DefaultErrorSchema` should align with the real schema of expected `dict`. Now `name`, `properties` and `required` are missing, but `model` exists instead.
* I think we should add the default values for `status` (maybe rename to `status_code`?), `name` and `description`.
* Feature should be covered by tests

@caiofauza, would you like to work on this?

mvanderlee:
Not sure this is needed anymore since we can define custom response models already, and we can do so by using Pydantic models.

```python
app = FastAPI(
    responses={422: {"model": YourErrorModel}}
)
```

YuriiMotov:
> Not sure this is needed anymore since we can define custom response models already, and we can do so by using Pydantic models.
> 
> ```python
> app = FastAPI(
>     responses={422: {"model": YourErrorModel}}
> )
> ```

This PR aims to simplify the process if you need to change the status code used for error responses.
For example, if you want it to be `400` instead of `422`.
For now it's easy to add response, but to remove default response you need to use hack:
```py
app = FastAPI(
    responses={400: {"model": YourErrorModel}, 422: None}
)
```

Ideally this PR should also change the status code FastAPI uses to send error response in case of request validation error

mvanderlee:
> > Not sure this is needed anymore since we can define custom response models already, and we can do so by using Pydantic models.
> > ```python
> > app = FastAPI(
> >     responses={422: {"model": YourErrorModel}}
> > )
> > ```
> 
> This PR aims to simplify the process if you need to change the status code used for error responses. For example, if you want it to be `400` instead of `422`. For now it's easy to add response, but to remove default response you need to use hack:
> 
> ```python
> app = FastAPI(
>     responses={400: {"model": YourErrorModel}, 422: None}
> )
> ```
> 
> Ideally this PR should also change the status code FastAPI uses to send error response in case of request validation error

I disagree. This PR specifically talks about the eror schema. 
I also think it's fine for FastAPI to have opinions, but I may be biased because I strongly agree with them. A validation error should be a 422. If a developer feels differently he can overwrite it already. 

1. If they define any of `422, "4XX", or "default"`, then FastAPI does not add the default schema
    ```python
    responses={422: {"model": YourErrorModel}}
    responses={"4XX": {"model": YourErrorModel}}
    responses={"default": {"model": YourErrorModel}}
    ```
    
    Note that neither of these work
    ```python
     # Your example of a hack raises an error!
    responses={422: None} 
    # 422 still shows up in the OpenAPI schema.
    responses={422: {"model": None}}
    ```
    
2. To overwrite the actual response status code the developer can add a custom error handler
    ```python
    async def my_request_validation_exception_handler(
        request: Request, exc: RequestValidationError
    ) -> JSONResponse:
        return JSONResponse(
            status_code=400,
            content={"detail": jsonable_encoder(exc.errors())},
        )
        
    fastapi = FastAPI(exception_handlers={RequestValidationError: my_request_validation_exception_handler})
    ```

The only issue left is that the user can't simply disable the 422 documentation. They must define either a default, or `4XX` both of which may be inaccurate as they cover too many other status codes. 
This could be solved with a simple boolean. e.g.: `include_default_validation_response_in_schema`

YuriiMotov:
> This could be solved with a simple boolean. e.g.: `include_default_validation_response_in_schema`

I agree with this. `include_default_validation_response_in_schema` is simpler

Issue #3993, opened 2021-10-05
Preserve json.JSONDecodeError when handling invalid JSON in request body


declaresub:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
not applicable
```

### Description

I think that the response to a request with syntactically invalid JSON should be 400, not 422.  This is the feature request I would prefer.  But I would be willing to write a custom exception handler for now, if that were feasible.

### Wanted Solution

   In FastAPI/routing.py line 210-211, we have 

        except json.JSONDecodeError as e:
            raise RequestValidationError([ErrorWrapper(e, ("body", e.pos))], body=e.doc)

I propose that the original exception not be discarded, by raising from -- 

        except json.JSONDecodeError as e:
            raise RequestValidationError([ErrorWrapper(e, ("body", e.pos))], body=e.doc) from e

With the original exception included, it would be simple enough to write a custom exception handler to handle the request as I see fit.  This change should not break anyone's code, and I think it is generally the better practice to use raise from in such situations.

### Wanted Code

```python
not applicable
```

### Operating System

Linux, macOS

### FastAPI Version

0.63.0

### Python Version

3.9.5

UKnowWhoIm:
Is this currently being worked on @declaresub ?

declaresub:
I don't know.  I would be happy to do it, but this is the first issue I've submitted to fastapi, and with 389 open pull requests, I didn't see the point spending time on a pull request that would never be looked at.

tiangolo:
Thanks for posting it @declaresub! And thanks @UKnowWhoIm for the PR! :tada: :cake: 

[...]

PR #4057, opened 2021-10-16
✨ Preserve `json.JSONDecodeError` information when handling invalid JSON in request body, to support custom exception handlers that use its information

UKnowWhoIm:
Save `JSONDecodeError` to context by adding a from clause. The feature was requested because the user wants to use status code 400 for invalid response body.

Usage

```
from fastapi.exception_handlers import request_validation_exception_handler

app = FastAPI()

async def handle(request: Request, exc: Exception):
    if isinstance(exc.__context__, JSONDecodeError):
        return Response(str(exc),status_code=400)
    return await request_validation_exception_handler(request, exc)
```

Enhancement requested #3993


tiangolo:
Awesome, thank you @UKnowWhoIm! :bow:

PR #4060, opened 2021-10-19
OpenAPI Plugin System - with Snippet Support

ghandic:
A first pass at implementing a Plugin system for FastAPI OpenAPI showing adding snippets in openapi and redoc. Keen to gather thoughts before putting in any more work into this.

Proposed user usage:

```python
from fastapi.openapi.plugins.snippet import SnippetPlugin, PremadeSnippets

app = FastAPI(openapi_plugins=[SnippetPlugin(snippets=PremadeSnippets.get(["curl", "python"]))])

@app.post("/a")
def do_a(demo: str = Query(..., example="lol")):
    return {"answer": "a"}

##### OR Custom Snippets

from fastapi.openapi.plugins.snippet import SnippetPlugin, CodeSampleFactory

PERL = CodeSampleFactory(language="perl", title="PERL", syntax="perl")

app = FastAPI(openapi_plugins=[SnippetPlugin(snippets=PERL.snippet())])

@app.post("/a", openapi_extra={"x-codeSamples": [PERL.sample("this is how you do it in perl")]})
def do_a(demo: str = Query(..., example="lol")):
    return {"answer": "a"}
```

Example of custom snippets using "x-codeSamples" as supported in redoc

Screenshot 2021-10-19 at 12 38 43

Example of auto-generated snippets (note you can select as few or as many as you wish, I added all to showcase here)

Screenshot 2021-10-19 at 12 39 54

Check out `example.py` for how it currently looks to the user to spin up 

Still waiting on the openapi-snippet to add CDN support in this issue raised last week https://github.com/ErikWittern/openapi-snippet/issues/74 which would remove the need to build it yourself

Junk files to be removed:
* `example.py`
* `public/openapisnippet.min.js`

TODO:
* Will need linting on the right rules
* Will need to add unit tests
* Documentation on how to use the plugin system
* Would we prefer ABC or protocols?
* Additional plugins built to show that the template generalizes well (in talks to add date picker and checkbox plugins)


tiangolo:
Thanks for the effort @ghandic! :cake: 

This is a very interesting idea! Nevertheless, the code complexity increase and the maintenance burden would be increased a lot with it. And I'm not sure that it's justified to have to maintain and teach people how each of the parts work, to be able to use it. 

There are many ways to extend Swagger UI, but I think trying to cover each one as part of FastAPI would mean a lot of additional code to maintain and explain. And in those cases I prefer to teach people how to do those things themselves.

I think this is one of the cases where it's probably easier for people to follow https://fastapi.tiangolo.com/advanced/extending-openapi/ to change OpenAPI and/or Swagger UI however they find necessary. Adding a layer of abstraction on top means people have to learn it, and it still limits them to what the layer can do, and they will probably still have to learn how it works underneath.

But something I think you could do that would probably be very useful is something like a blog post, explaining how to follow that same tutorial in the docs to remove the default OpenAPI and Swagger UI, and how to add them again adding custom snippets, without requiring an internal plugin system.

Given that, I'll pass on this PR, but thanks a lot for all the effort and interest! :coffee: :cookie: 

iudeen:
This would be cool if it were a separate package! Easy to maintain, those who want can use, and FastAPI remains light and cool!

PR #4061, opened 2021-10-19
Catching exceptions in the dependancy call

ghandic:
eg for pydantic validation error to return a 422

Just a first pass, will need tests adding etc, but this aims to solve a few issues for deepObject encoding (where you need to use Depends currently), also in response to the issue raised in discussion #3426 

NOTE: Moved from https://github.com/tiangolo/fastapi/pull/3953 as it was in my master branch on my fork

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -116,18 +116,12 @@ def get_param_sub_dependant(
     else:
         dependency = param.annotation
     return get_sub_dependant(
-        depends=depends,
-        dependency=dependency,
-        path=path,
-        name=param.name,
-        security_scopes=security_scopes,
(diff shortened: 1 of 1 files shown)


idmitrievsky:
Hey, @ghandic! I see that this PR is not active, were there any obstacles beyond lack of time?

ghandic:
@tiangolo js only person with merge rights

tiangolo:
Thanks! I think this was covered in another PR, now it's possible to catch exceptions in dependencies with `yield` :tada: 

https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/#dependencies-with-yield-and-httpexception

Given that, I'll close this one. But if there's some other use case not covered, please create a Discussion wth a minimal self contained example that I can copy paste to understand the use case better. :coffee:

PR #4076, opened 2021-10-21
🌐 Add Korean translation for `docs/ko/docs/tutorial/dependencies/dependencies-in-path-operation-decorators.md`

0417taehyun:
- Add Korean translation for `docs/ko/docs/tutorial/dependencies/dependencies-in-path-operation-decorators.md`
- Complete to check that it works on test server using `python ./scripts/docs.py live ko`.
- Related #2017


DevDae (commented):
Big appreciate your works 🚀
I  want to check the word, 애플리케이션.
It may be translated as 응용 프로그램 and you also mentioned about this word in my PR. 
 I look forward to your reply. Thank u 😄

PR #4089, opened 2021-10-23
🐛 Fix issue with double model parsing

uriyyo:
Fix caching issue when instead of cache lookup dependency solving was performed all the time:

Code example:
```py
from asyncio import run

from httpx import AsyncClient
from pydantic import BaseModel, root_validator

app = FastAPI()


class Model(BaseModel):
    @root_validator
    def __root_validator__(cls, values):
        print("parsing")
        return values


def dependency(request: Model):
    return request


@app.post("/")
def foo(
        a: Model = Depends(dependency),
        b: Model = Depends(dependency),
):
    return {}


async def main():
    async with AsyncClient(app=app, base_url="http://testserver.com/") as client:
        await client.post("/", json={})


if __name__ == '__main__':
    run(main())
```

Expected behaviour - parsing printed only one time
Actual behaviour - parsing printed two times


yezz123 (approved):
LGTM+1 🚀, Waiting for this in the next version this one gonna help me fix a lot of issues related to my package 👌🏻
Thanks @uriyyo ✨

uriyyo:
@tiangolo Could you please review this PR? It would my nice to fix this issue

BilalAlpaslan (approved):

YuriiMotov (approved):
Can confirm that:

- the issue is still relevant to FastAPI 0.110.0
- new added test fails before changes are made and passes after
- solution looks nice and reasonable

uriyyo:
@Kludex Any chase you can take a look?)

tests/test_dependency_cache.py
```diff
@@ -1,9 +1,31 @@
 from fastapi import Depends, FastAPI, Security
+from fastapi._compat import PYDANTIC_V2
 from fastapi.testclient import TestClient
+from pydantic import BaseModel
 
 app = FastAPI()
 
-counter_holder = {"counter": 0}
+counter_holder = {"counter": 0, "parsing_counter": 0}
```
Jaza, line 8: IMHO this should be set in pytest session-level context or similar, it shouldn't just be a global variable, I feel that there's too high a risk that the test could end up running with the counter initialised to a non-zero value (e.g. if for some reason the same test runs twice in one pytest invocation, if we end up adding another test in the future that depends on the same counter, etc).

Jaza:
Mentioning this perf improvement somewhere in the docs might also not be a bad idea?

svlandeg (commented):
Apologies for the late follow-up on this one. 

PR #11323 has a very similar fix, and has been updated to include the unit test from this PR, and is in sync with current `master`. As such, I'm going to close this PR so we can hopefully wrap this up relatively quickly in #11323. 

Thanks to everyone who contributed and helped review this one! If you have time, please consider having a quick look at #11323 as well. 🙏

Issue #3616, opened 2021-07-27
FastAPI 0.67.0 new dataclass related features conflict with some aspects of new SQLAlchemy dataclass-style model mapping


AlexanderPodorov:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from __future__ import annotations

from dataclasses import dataclass, field

from fastapi import Depends, FastAPI
from pydantic import BaseModel
from sqlalchemy import Column, ForeignKey, Integer, String, create_engine
from sqlalchemy.orm import Session, registry, relationship, sessionmaker
from uvicorn import run

SQLALCHEMY_DATABASE_URL = "sqlite:///./sql_app.db"

engine = create_engine(
    SQLALCHEMY_DATABASE_URL, connect_args={"check_same_thread": False}
)
SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)

mapper = registry()


@mapper.mapped
@dataclass
class User:
    __tablename__ = "users"
    __sa_dataclass_metadata_key__ = "sa"

    id: int = field(
        metadata={
            "sa": Column(Integer, primary_key=True, index=True)
        }
    )
    email: str = field(
        metadata={
            "sa": Column(String, unique=True, index=True)
        }
    )
    items: list[Item] = field(
        init=False,
        metadata={
            "sa": relationship("Item", lazy="raise")
        }
    )

    @property
    def greeting(self):
        return f"Hello {self.email}"


@mapper.mapped
@dataclass
class Item:
    __tablename__ = "items"
    __sa_dataclass_metadata_key__ = "sa"

    id: int = field(
        metadata={
            "sa": Column(Integer, primary_key=True, index=True)
        }
    )
    title: str = field(
        metadata={
            "sa": Column(String, index=True)
[...]

PR #4094, opened 2021-10-26
Fix issue for SQLAlchemy Dataclasses post 0.67

Lunrtick:
*edited* due to change from flag based approach in the route to an optional property on the dataclasses.

As detailed in #3616, SQLAlchemy dataclasses don't work nicely with the new defaults introduced in FastAPI 0.67. This pull request adds the option for dataclasses to include a `read_with_orm_mode` flag, matching the options available to Pydantic models.

Currently this is specified as a special dunder property: `__read_with_orm_mode__` on each dataclass, though it could take a different form. In my codebases, all my SQLAlchemy dataclasses inherit from the same base class, so I would only be setting this property on the base class.

Besides disabling the new dataclass parsing feature in some manner, it seems as though it'd be impossible to use the SQLAlchemy dataclass pattern as described in their docs [here](https://docs.sqlalchemy.org/en/14/orm/mapping_styles.html#declarative-mapping-with-dataclasses-and-attrs).


Lunrtick:
Another option would be to check if a response model was declared - I'd imagine that automatically using the dataclass doesn't make sense when an explicit Pydantic response model is given, though I may have misunderstood.

tiangolo:

I wanted to re-write the SQL docs with SQLModel, I just finished that here: https://github.com/fastapi/fastapi/pull/12285

The new docs are here: https://fastapi.tiangolo.com/tutorial/sql-databases/

SQLModel would be the way to go for using SQLAlchemy mixed with Pydatnic, so it would be the way to go with FastAPI (also said in SQLAlchemy's docs).

Given that, I'll pass on this one, but thanks for the effort! :cake:

PR #4097, opened 2021-10-27
⬆ Upgrade version pin accepted for Flake8, for internal code, to `flake8 >=3.8.3,<6.0.0`

jamescurtin:
Flake8 released a new major version, and the pinned upper bound in `pyproject.toml` needs to be increased to reflect this.


tiangolo:
Thanks @jamescurtin! :rocket: 

I updated it a bit, now that there's a new version, etc. Thanks for your contribution! :rocket: :cake:

PR #4105, opened 2021-10-29
Cache cloning of pydantic types globally for 2x+ startup speed-up

huonw:
If `response_model=...` is specified in a route handler, FastAPI creates a copy of the type to avoid accidental leakage of secrets. This is done in a pure way, but is only cached within a single route. This means that that if there's models that occur several times (for instance, multiple routes that return the same model), identical work is done and re-done. This PR change adds a cache for the conversion: when converting a type that's already been seen, just use the cache.

This dramatically improves the start-up time of the server, making for a nicer developer experience when running locally, or running tests, and reducing cold-start times if run in FaaS/serverless platforms like AWS Lambda.

For instance, we have an app with 144 routes, defined in the usual way with an `app = FastAPI(...)` in app.py and various routers across other files.

I measured the impact on this change in two ways.

- Just the time to initialise all the routes etc (command below)
- The time to reach "Application startup complete" when actually running (`uvicorn.run(..., port=8000, reload=True, debug=True, workers=1)`)

| measurement    | 0.85.0 (s) | this patch (s) |
|----------------|-----------:|---------------:|
| initialisation |       4.65 |           1.43 |
| startup complete   |       11.45 |           3.38 |

Initialisation measurement command:

    python -c 'import time; start = time.perf_counter(); import .app; print(time.perf_counter() - start)'

---

Thanks for FastAPI!


huonw:
I tried this in our AWS Lambda infrastructure. The cold start initialisation drops from ~6s to ~2s, not including the time to actually serve the request, which is typically milliseconds. That is, 4s/two-thirds of the lambda start-up time is this repeated redundant type conversion.

(We do realise that FastAPI works great as a standalone server, not in lambda too.)

yezz123 (approved):
LGTM+1! I will use it locally ⚡️, waiting for Sebastián to merge it!
Thanks, @huonw 🚀

ghandic:
Likewise LGTM, massive improvement for serverless 🚀

huonw:
Hello @tiangolo, when you have chance, could you take a look at this PR? Thank you!

fastapi/utils.py
```diff
@@ -69,28 +69,22 @@ def create_response_field(
...
 
-def create_cloned_field(
-    field: ModelField,
-    *,
-    cloned_types: Optional[Dict[Type[BaseModel], Type[BaseModel]]] = None,
-) -> ModelField:
-    # _cloned_types has already cloned types, to support recursive models
-    if cloned_types is None:
-        cloned_types = dict()
+_CLONED_TYPES_CACHE: Dict[Type[BaseModel], Type[BaseModel]] = dict()
```
cikay, line 72: Use literal empty dict instead of built in function `dict`. Is is optimized.

cikay (changes requested):

huonw:
Hello @tiangolo, I've now updated this to merge cleanly again, as well as updated the performance numbers.

In addition to the local server and FaaS/serverless use-cases, we're affected by this in our local testing that uses `fastapi.testclient.TestClient`, since that requires initialising the app. The current initialisation times means that a test runs spends 5-10s on setup before running any actual tests.

As always, thank you for FastAPI.

iudeen (approved):

orfisko:
Is there a timeline on when this will be shipped? This change makes a huge difference in terms of startup speed for my project. On AWS lambda with cold starts this is an important change.

madkinsz:
As noted by @orfisko there's a duplicate of this with a smaller diff at #4645. This also addresses the issue at https://github.com/tiangolo/fastapi/issues/4644

BilalAlpaslan (approved):

michaelgmiller1:
This is a great addition and sorely needed for the application I'm working on. Please let me know if there's anything I can do to help push this over the finish line!

orfisko:
It's all nice and dandy to put security fixes in fastapi, but I am not getting those because I am condemned to use a branched version of fastapi due to this bug. I don't see why this is not moving forward?

madkinsz:
@orfisko I've updated https://github.com/tiangolo/fastapi/pull/4645 with `master`.

anton-shauchenia:
We've tried this patch in our FastAPI application on AWS Lambda and were able to shave quite some cold start latency. I wonder if we can do anything to help push this PR forward.

tiangolo:
Thanks a lot for the work on this @huonw! 🙇 

And thanks everyone for the comments and the patience, sorry for the long delay. 🙈 

Special thanks to those of you that tested it locally and commented it, that helps a lot.

I've been working on the migration to Pydantic v2, which will make this `clone_field` logic unnecessary altogether, as everything will be done on the Pydantic-core (Rust) side. Nevertheless, I'll first make a release fixing this.

This will be available in FastAPI 0.96.0, released in the next hours. 🚀 

---

I'm taking the alternative implementation in https://github.com/tiangolo/fastapi/pull/4645 by @madkinsz just for the simplification in the changes/diff, the weakref, and keeping the signature, it's pretty much equivalent. But I'm making you @huonw an honorary co-author of that PR for your work here. 🤓

PR #4119, opened 2021-10-31
fix for #4118 issue

### LINKED ISSUE #4118: `dependency_overrides` requires unnecessary arguments with Protocol/ABC in Swagger
### First Check

### Commit to Help

### Example Code

```python

import dataclasses
from typing import Protocol

import uvicorn
from fastapi import FastAPI, Depends, Response

app = FastAPI()


@dataclasses.dataclass()
class Foo:
    id: int


class BaseRepo(Protocol):
    def get_foo(self, some_id: int) -> Foo: ...


class ConcreteRepo:
    def get_foo(self, some_id: int) -> Foo:
        return Foo(id=some_id)


@app.get("/hello/world")
async def view(repo: BaseRepo = Depends(BaseRepo)):
    return Response(200)


app.dependency_overrides[BaseRepo] = lambda: ConcreteRepo()  # also without lambda works incorrect

if __name__ == '__main__':
    uvicorn.run(app)
```

### Description

* Open /docs route and see that API method requires *args and **kwargs for request (probably from metaclass of Protocol)

### Operating System

Linux

### Operating System Details

_No response_

### FastAPI Version

0.70.0

### Python Version

3.9.7

### Additional Context

_No response_

GLEF1X:


tiangolo:
Hey there! I just replied to your issue. I'll copy the message here as well for completeness.

---

I see that in your examples `BaseRepo` and `ConcreteRepo` don't have an `__init__` method, so they don't require other dependencies, don't require data from the request, don't integrate with OpenAPI or OAuth2. Why do you want to put them in dependencies instead of creating the instances in the code?

Using things in dependencies is normally for when there's a reason to use them, because otherwise, it's simpler and more explicit to call the code directly in your function.

I want to understand your use case to see why it would make sense to support this, to achieve what, and see if there's another way to do it.

tiangolo:
@jbmusso thanks for writing. It's still not super clear to me, please create a new issue with a self-contained example that I can run to see your problem.

You could also copy this PR and add tests to it that show how you would intend to use it.

jbmusso:
I think the issue and code sample provided by @GLEF1X here https://github.com/tiangolo/fastapi/issues/4118#issue-1040587961 highlights the issue. Protocol allow programming against interfaces rather than concrete implementations.

Issue #3920, opened 2021-09-18
I hope fastapi can support TypeVar


normidar:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import TypeVar, Generic

from fastapi import Depends
from sqlalchemy.orm import Session

from app.core.database_engine.db_core import get_db
from app.core.table_class import HasIdTable, DateCreateTable, DateCreateUpdateTable

T = TypeVar('T', bound=HasIdTable)


class DbAdaptor(Generic[T]):
    def __init__(self, db: Session = Depends(get_db)):
        self.db = db

    def add(self, data_element: T, is_commit: bool = True) -> dict:
        if isinstance(data_element, DateCreateTable):
            data_element.create_stamp()
        self.db.add(data_element)
        if is_commit:
            self.db.commit()
        return data_element.get_dict()

    def read_by_id(self, id: int) -> T:
        rt: T = self.db.query(T).filter_by(id=id).first()
        return rt

@bp.post('/create', description='create data to table')
def create(dba: DbAdaptor[SampleTable] = Depends(), data: str = Body(..., embed=True)):
            """create data into the table"""
            data = SampleTable(data=data,
                               link=str(random.randint(0, 99999)),
                               title=data,
                               content=data)
            return dba.add(data)
```

### Description

im sorry, im no english user but I want to achieve this function.

This is the code in my project, but guess what happened when I opened swagger.

![スクリーンショット 2021-09-18 22 48 27](https://user-images.githubusercontent.com/36730656/133891074-8ef8e5c2-0a41-4f58-bdbb-4c29c138127f.jpg)

This shouldn't happen.

I think you know what I want to do.

I just want to simplify the operation of the database.

### Wanted Solution

I hope it can show a normal swagger.

[...]

PR #4126, opened 2021-11-01
fix(generic): generic in depends

JeanArhancet:
Hello,

This is my first contribution for this project :) 

Fix issues #3920


yezz123 (approved):
Overall LGTM+1 🚀, This one could support the issue relate to support TypeVar. 👌🏻

fastapi/dependencies/utils.py
```diff
@@ -243,29 +243,69 @@ def is_scalar_sequence_field(field: ModelField) -> bool:
...
+        args_to_replace = []
+        for arg in annotation_args:
+            if arg in provided_type:
+                arg = provided_type[arg]
+                updated_args_annotation = True
+            args_to_replace.append(arg)
+        if updated_args_annotation:
+            setattr(annotation, "__args__", tuple(args_to_replace))
+    return annotation
+
```
cikay, line 308: Decrease the indentation level if it is possible.

```py
if annotation in provided_type:
    annotation = provided_type[annotation]

annotation_args = get_args(annotation)
if annotation_args is None:
    return annotation

args_to_replace = []
for arg in annotation_args:
    arg = provided_type.get(arg, arg)
    args_to_replace.append(arg)

updated_args_annotation = any(
    arg in provided_type for arg in annotation_args
)
if updated_args_annotation:
    setattr(annotation, "__args__", tuple(args_to_replace))

return annotation
```
In my view this is better.

JeanArhancet: Thanks for your feedback @cikay.

Yes it's clearer but we go through the loop twice. it's seem less efficient to me, what do you think ?

cikay: Then you can merge the loops. It would still be less indentation level. by the way the built-in function `any` is break the loop for the first `True` value. It is efficient.

cikay: There is nothing to discuss to apply the early return pattern to decrease the indentation level.

JeanArhancet:
> Overall LGTM+1 🚀, This one could support the issue relate to support TypeVar. 👌🏻

Hello @yezz123 , I changed the indentation level, what do you think ?

PR #4220, opened 2021-11-29
✨Add `Rollbar` to the middleware examples list

bxsx:
This PR adds `Rollbar` to the [middleware examples list](https://fastapi.tiangolo.com/advanced/middleware/#other-middlewares).
Rollbar is a notifier for reporting exceptions, errors, and log messages. Full docs: https://docs.rollbar.com/docs/fastapi

I also sorted the list alphabetically in f0968dc. I can revert it if it's not desired.


tiangolo:
Thanks @bxsx! I think it makes more sense for this to be in the awesome-asgi repo, or now that it's on Starlette, over there. But otherwise it makes more sense to me for it to be in the Rollbar product docs. So I'll pass on this one, but thanks for the interest! 🍰

PR #4267, opened 2021-12-11
🐛 Fix support for extending `openapi_extras` with parameter lists

orilevari:
# Issue
When specifying parameters in openapi_extras for an endpoint, they will overwrite the existing query parameters with the extra parameters. This is due to deep_dict_update in utils.py overwriting list types rather than merging them together.

# PR Content
This PR updates deep_dict_update to merge arrays instead of overwriting. I have also added a test case that demonstrates and verifies the desired behavior.

# Testing
Ran bash scripts/test-cov-html.sh  with 100% test pass including new test case


tiangolo:
Awesome, thanks for the contribution @orilevari! :rocket: :tada:

Darep:
Is there a way to fall back to the old functionality somehow? 🤔 We have `Query(default=datetime.utcnow())` and we are trying to override it like `openapi_extra={"parameters": [{"schema": {"default": "Current timestamp in UTC"}}]}` 😄 And this change now simply appends a new parameter in the list.

We might do `include_in_schema=False` and write it completely if there isn't, so no worries... :blush: (Other solutions for changing the default in the OpenAPI-schema also welcome!)

PR #4314, opened 2021-12-21
fix unnecessary `elif`-`else`

ghost:
The `else` / `elif` block here is unnecessary because the `if` block has `return` / `raise` statement or the leading block has `continue` statement which is trivial.


aarizat:
I have seen a lot of this kind of `if-else` (with return), I think that it is used because to some people is more readable.

BilalAlpaslan (approved):

PR #4315, opened 2021-12-21
make use of faster equality check

ghost:
This is faster and more readable, than checking for equality against each of the values


cikay (commented):
The commit message should be "Use the operator `in` to check multiple cases to enhance performance"

PR #4316, opened 2021-12-21
use literal to create data structure

ghost:
Using literal gives better performance than calling functions

```Python
timeit.timeit(stmt="tuple()", number=1000000000)
33.760535042
timeit.timeit(stmt="()", number=1000000000)
4.175541042000006
timeit.timeit(stmt="dict()", number=1000000000)
56.924091708000105
timeit.timeit(stmt="{}", number=1000000000)
11.097609916000238
```

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -76,7 +76,7 @@ def create_cloned_field(
 ) -> ModelField:
     # _cloned_types has already cloned types, to support recursive models
     if cloned_types is None:
-        cloned_types = dict()
+        cloned_types = {}
     original_type = field.type_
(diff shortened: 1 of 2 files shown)


fastapi/utils.py
```diff
@@ -76,7 +76,7 @@ def create_cloned_field(
 ) -> ModelField:
     # _cloned_types has already cloned types, to support recursive models
     if cloned_types is None:
-        cloned_types = dict()
+        cloned_types = {}
```
cikay, line 79: This is another way
```py
cloned_types = cloned_types or {}
```

PR #4318, opened 2021-12-21
Inline Returned Variables

yezz123:
Inlines a variable to a `return` in the case when the variable being declared is immediately returned.


cikay (commented):
The commit message should be "Remove redundant variable binding".

tiangolo:
Thanks for the interest! But I prefer to keep it as is because it allows me to debug more easily with breakpoints. So I'll pass on this one, but thanks! ☕

PR #4319, opened 2021-12-21
🐛 Make sure a parameter defined as required is kept required in OpenAPI even if defined as optional in another dependency

chardigio:
**Summary**
In the case where a parameter is looked for by multiple dependencies, but only required by some of them, we should reflect that in the openapi spec as `required=True`.

**Currently**
Right now the openapi spec indeterminately displays the parameter being required/optional. However, omitting the parameter will always result in a `422`.

**Suggested**
We ought not change anything about how the server acts, however we should show that the parameter is required in the openapi spec.

**Rationale**
As is, this is misleading and confusing for anyone trying to work with OpenAPI where this kind of situation arises. This becomes especially important when using inheritance with [class-based views](https://fastapi-utils.davidmontague.xyz/user-guide/class-based-views/). Often times you may want to look for a certain field in a base view, but only require it in _some_ of the views that inherit from it.


cikay (commented):
The commit message should be "Simplify conditions"

cd17822:
@tiangolo any chance you can check this out? It would help clear up our docs a lot 🙏🏼

tiangolo:
Thanks for the contribution @cd17822! :cake: 

I updated the test to have in mind the new schemas for validation errors. And I refactored the implementation to account for the location of parameters, from a recent PR.

Thanks for the interest here @cikay, nevertheless, it's not necessary to ask for some type of commit messages. I make sure the final commit message is consistent with the rest and then squash and merge. It's better to check if the implementation is correct, if it has tests, if the tests break before the fix and pass after the fix. Thanks!

This will be available in the next FastAPI version, `0.82.0`, probably in the next hours. :tada:

PR #4323, opened 2021-12-23
use `localhost` as an example than `0.0.0.0`

ghost:
In my opinion, using explicit binding to `localhost` / `127.0.0.1` or host IP you want to run on is safe example to provide than bind `0.0.0.0` / `internet`


dolfinus:
I'll fail if someone runs FastAPI-based app in the docker container

PR #4324, opened 2021-12-23
fix default value to an arguments

ghost:
I would use safe default value  than use mutable value to an args

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -34,9 +34,11 @@ def jsonable_encoder(
     exclude_unset: bool = False,
     exclude_defaults: bool = False,
     exclude_none: bool = False,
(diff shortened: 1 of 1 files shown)


fastapi/encoders.py
```diff
@@ -34,9 +34,11 @@ def jsonable_encoder(
     exclude_unset: bool = False,
     exclude_defaults: bool = False,
     exclude_none: bool = False,
-    custom_encoder: Dict[Any, Callable[[Any], Any]] = {},
+    custom_encoder: Optional[Dict[Any, Callable[[Any], Any]]] = None,
     sqlalchemy_safe: bool = True,
 ) -> Any:
+    if custom_encoder is None:
+        custom_encoder = {}
```
cikay, line 41:
```py
```
This is better.

PR #4329, opened 2021-12-25
fix useless import alias; group & sort them

ghost:
Certain imports do not require alias and when removed can be grouped as per PEP8, while these changes are done `isort` made  sure imports are sorted too. Also increases readability.

### DIFF
--- a/fastapi/__init__.py
+++ b/fastapi/__init__.py
@@ -2,23 +2,24 @@
 
 __version__ = "0.70.1"
 
-from starlette import status as status
+from starlette import status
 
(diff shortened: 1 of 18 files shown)


Kludex:
I guess this has something to do with starlette as it's a reimport?

I'm not sure why `mypy --strict` needs it. I've never checked... But find more on the PR that introduced it: https://github.com/tiangolo/fastapi/pull/2547

prnam:
Neither me. 😅

Anyway, thanks for looking into this issue @kludex and sharing more details. 🙂

PR #4341, opened 2021-12-28
Dynamic parameter specification

borisoid:
### Prehistory
I was trying to create a little personal framework that would auto-generate standard CRUD endpoints for me (just like DjangoRestFramework does). The huge problem was that there was no way to specify the number, names, types of parameters other than using function signature. Apparently some people were also wondering how to define the interface dynamically ([here](https://github.com/tiangolo/fastapi/issues/2511) for example). Following the answer from the link, I had to use ``makefun.with_signature`` decorator to make it somewhat work, like shown below.
Screenshot from 2021-12-28 18-24-47

### The actual thing
So I made ``@extra_parameters()``  decorator and ``class Kwargs(Dependency)``; modified your util function responsible for signature parsing, so that it takes these things into consideration as well.

Here is how that code from the picture looks now
Screenshot from 2021-12-28 19-37-35

Kwargs usage
Screenshot from 2022-01-13 18-39-28


ariloulaleelay:
Hi, @Borisoid.
Good PR! 
I have met same issue. I found non invasive solution inspired by [merge-args](https://pypi.org/project/merge-args/).
redefine_defaults changes the signature as you need it

Here the code:
```python
#!/usr/bin/env python3

import functools
import inspect
import sys
import types

PY38 = sys.version_info >= (3, 8)


def _blank():  # pragma: no cover
    pass


def redefine_defaults(f, remove_varkw=False, **kwargs):
    """"
      f: function
      remove_varkw: remove **kwargs from signature
      kwargs: defalts to redefine
    """"
    spec = inspect.getfullargspec(f)

    if spec.defaults:
        pos_args = spec.args[:-len(spec.defaults)]
        kw_args = spec.args[-len(spec.defaults):]
        defaults = spec.defaults
    else:
        pos_args = spec.args
        kw_args = []
        defaults = []

    new_posargs = []
    new_kwargs = []
    new_defaults = []
    new_kwonlyargs = []

    for arg_name in pos_args:
        if arg_name not in kwargs:
            new_posargs.append(arg_name)

    assert len(kw_args) == len(defaults)
    for arg_name, arg_default in zip(kw_args, defaults):
        if arg_name in kwargs:
            arg_default = kwargs[arg_name]
        new_kwargs.append(arg_name)
        new_defaults.append(arg_default)

    for arg_name, arg_default in kwargs.items():
        if arg_name not in new_kwargs and arg_name not in new_posargs:
            new_kwargs.append(arg_name)
            new_defaults.append(arg_default)

    for arg_name in spec.kwonlyargs:
        if arg_name in new_posargs or arg_name in new_kwargs:
            continue
        new_kwonlyargs.append(arg_name)

    varnames = list(new_posargs)
    if spec.varargs:
        varnames.append(spec.varargs)

    varnames.extend(new_kwargs)
    if spec.varkw and not remove_varkw:
        varnames.append(spec.varkw)

    passer_args = [len(new_posargs) + len(new_kwargs)]
    if PY38:
        passer_args.append(f.__code__.co_posonlyargcount)

    co_flags = f.__code__.co_flags
    if remove_varkw:
        co_flags = co_flags & (~inspect.CO_VARKEYWORDS)

    passer_args.extend([
        len(new_kwonlyargs),
        _blank.__code__.co_nlocals,
        _blank.__code__.co_stacksize,
        co_flags,
        _blank.__code__.co_code, (), (),
        tuple(varnames), f.__code__.co_filename,
        f.__code__.co_name,
        f.__code__.co_firstlineno,
        f.__code__.co_lnotab
    ])
    passer_code = types.CodeType(*passer_args)
    passer = types.FunctionType(passer_code, globals())
    passer.__defaults__ = tuple(new_defaults)

    g = types.FunctionType(
        f.__code__,
        f.__globals__,
        name=f.__name__,
        argdefs=tuple(new_defaults),
        closure=f.__closure__
    )
    g = functools.update_wrapper(g, f)
    g.__kwdefaults__ = f.__kwdefaults__
    g.__wrapped__ = passer
    g.__defaults__ = tuple(new_defaults)
    # TODO rewrite annotations
    return g


def some_func(x='will_depend', **kwargs):
    print(f'{kwargs}')


def print_sig(f):
    signature = inspect.signature(f)
    print(f'signature for {f.__name__} {signature}')


def main():
    some_func()
    other_func = redefine_defaults(some_func, remove_varkw=True, y='new_depends', m='mmm', x='redefine_depends_for_x')
    other_func(m='mmmm')
    print(some_func is other_func)

    print_sig(some_func)
    print_sig(other_func)


if __name__ == '__main__':
    main()
```

Borisoid:
Hi, @ariloulaleelay!
Yes, this is another way to do it, however both this solution and ``makefun.with_signature`` that I wrote about are just "hacs" that trick ``inspect.signature()`` into returning the signature you want instead of the actual one. I just find such approach too convoluted and straight up ugly. Especially for FastAPI that tempts to be as developer-friendly as possible. The thing that confuses me the most is that FastAPI claims that it can help you develop faster, but forces you to explicitly type out each endpoint not letting you generate them with code. And if you want to make it more flexible, you have to spend time "hacking" signatures. There must be the native way of doing this, I think.

fastapi/dependencies/utils.py
```diff
@@ -266,6 +267,79 @@ def get_typed_annotation(param: inspect.Parameter, globalns: Dict[str, Any]) ->
...
+    merged_params = filter(
+        lambda p: p.kind != inspect.Parameter.VAR_KEYWORD, merged_params
+    )
+    return inspect.Signature(merged_params)
+
+
+def get_signature_extra_parameters(
+    call: Callable[..., Any]
+) -> Dict[str, inspect.Parameter]:
+    return getattr(call, "__endpoint_extra_parameters__", dict())
```
cikay, line 294: Use empty dict literal instead of the built-in function `dict`. It is faster.

Borisoid: Yeah, I thought they were the same, but apparently not. ``{}`` is more optimized. Thanks for the hint)

PR #4350, opened 2021-12-31
wrench: Use casefold for caseless comparison of schemes

Abdur-rahmaanJ:
[str.casefold()](https://docs.python.org/3/library/stdtypes.html?highlight=casefold#str.casefold) in the docs.

### DIFF
--- a/fastapi/security/http.py
+++ b/fastapi/security/http.py
@@ -78,7 +78,7 @@ class HTTPBasic(HTTPBase):
             detail="Invalid authentication credentials",
             headers=unauthorized_headers,
         )
-        if not authorization or scheme.lower() != "basic":
+        if not authorization or scheme.casefold() != "basic":
             if self.auto_error:
                 raise HTTPException(
                     status_code=HTTP_401_UNAUTHORIZED,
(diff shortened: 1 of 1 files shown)


yezz123 (approved):
LGTM! 🎉

ycd (approved):

harunyasar:
@Abdur-rahmaanJ I can't quite sure that what you were fixing here. Could you briefly explain why we need this change?  I am thinking that `str.lower()` will always work here perfectly because there shouldn't be any case of that we are supposed to support caseless matching for possible `Authorization` headers.

Abdur-rahmaanJ:
Well, `.casefold` is recommended for caseless matching. That's what we are doing with .lower which works but is not recommended by the docs. `.lower` is for converting to lowercase.

harunyasar:
@Abdur-rahmaanJ I can see where you are coming from but it doesn't add a value to replace `lower()` with `casefold()` because there won't be any case that we need to have a caseless matching for `Authorization` header schemes.

Abdur-rahmaanJ:
Thanks for your comments, i don't think i have more to add.

TheOnlyWayUp (approved):

tiangolo:
Thanks for the interest @Abdur-rahmaanJ! :coffee: 

As in this case the strings are plain ASCII, it wouldn't make much of a difference, so I'll just leave it as it is and I'll close this PR.

But thanks for the effort! :cake:

Abdur-rahmaanJ:
@tiangolo ASCII for now ^^/ btw thanks for cleaning the PRs

PR #4353, opened 2021-12-31
fix issues related to `assert` when running on optimised mode

ghost:
After making this change you will not see python skipping `assert` statement when running python on optimised byte code mode `.pyo` . Also we should limit `assert` to test cases and always raise exception in app / business logic, like for this case `AssertionError`.

Note: Use of `python -O` or `python -OO` depends on business use cases, scale of the application and objectives such as optimisation for performance, reduce memory footprint, build time reduction, image size reduction, security etc.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -153,7 +153,8 @@ async def run_endpoint_function(
 ) -> Any:
     # Only called by get_request_handler. Has been split into its own function to
     # facilitate profiling endpoints, since inner functions are harder to profile.
-    assert dependant.call is not None, "dependant.call must be a function"
+    if dependant.call is None:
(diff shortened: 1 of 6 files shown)


fastapi/routing.py
```diff
@@ -379,12 +383,14 @@ def __init__(
         self.responses = responses or {}
         response_fields = {}
         for additional_status_code, response in self.responses.items():
-            assert isinstance(response, dict), "An additional response must be a dict"
+            if not isinstance(response, dict):
+                raise AssertionError("An additional response must be a dict")
             model = response.get("model")
             if model:
```
cikay, line 389: Reverse the condition and add the continue statement to decrease the indentation level.

PR #4376, opened 2022-01-06
✨ Export `WebSocketState` in `fastapi.websockets`

matiuszka:
Trivial improvement about exposing `WebSocketState`. It will eliminate the need of importing it from `starlette` directly.

Previously:
```python
from fastapi.websockets import WebSocket
from starlette.websockets import WebSocketState
```
After:
```python
from fastapi.websockets import WebSocket, WebSocketState
```


TheOnlyWayUp (approved):

Kludex:
How do you use `WebSocketState`?

matiuszka:
So there is an endpoint that is sending some data, usually very rarely but still, and is receiving nothing. So while we are not sending any data and the WebSocket connection was closed by the client, the task is still running. So to add functionality that if the connection is closed task will be closed also we are monitoring the WebSocket status, by simple `.recevie`.

The endpoint more or less looks like that:
```python
@router.websocket("/subscribe/some/data/changes")
async def subscribe_connectors_status(
    websocket: WebSocket,
) -> None:
    async def send_some_data() -> None:
        last_data = await get_data()
        await websocket.send_json({"data": last_data})
        while True:
            await asyncio.sleep(1)
            new_data = await get_data()
            if new_data != last_data:
                # This happens very rarely that's why I would like to close all of these unused tasks.
                last_data = new_data
                await websocket.send_json({"foo": last_data})

    async def watch_status() -> None:
        while websocket.client_state == WebSocketState.CONNECTED:
            # We are not expecting any data, but this will change to the client state if the connection will be closed.
            await websocket.receive()

        tg.cancel_scope.cancel()

    await websocket.accept()
    async with create_task_group() as tg:
        tg.start_soon(watch_status)
        tg.start_soon(send_some_data)

tiangolo:
Thanks for the contribution @matiuszka! :cake: 

And thanks @Kludex for the help! :bow:

Knowing how you are using it made me feel it makes sense to have this in, so I'll take it. :rocket: 

Thanks! :coffee:

PR #4381, opened 2022-01-06
zap: Make the code smaller

ghost:


fastapi/routing.py
```diff
@@ -336,8 +336,7 @@ def __init__(
         self.endpoint = endpoint
         self.name = get_name(endpoint) if name is None else name
         self.path_regex, self.path_format, self.param_convertors = compile_path(path)
-        if methods is None:
-            methods = ["GET"]
+        methods = methods or ["GET"]
```
dimaqq, line 339: I wonder if perhaps `methods=[]` has a valid user case somehow...
For example top-level `[]` followed by specialisation of `["PATCH"]`.

jetvolkov: Hmm, how about making `methods=["GET"]` the default?

dimaqq: my 2c: let's not make these changes at all.
they concern the most used API where any subtle change may break someone's code.

PR #4401, opened 2022-01-09
Docs: fix incorrect example for dicts in 3.6+

joakimnordling:
As part of the new python 3.9+ examples (commit d08a062ee2121f446537f06c8425cdeb1209e4ea) there's been a small error in the example for dicts. Instead of adding a new example, the old code was replaced and linked to be used for both 3.6+ and 3.9+.

This merely restores the old example for 3.6 and moves the new example to a new file with the "_py39" suffix (so it's in line with the other similar examples) and fixes the link in the markdown.

### DIFF
--- a/docs/en/docs/python-types.md
+++ b/docs/en/docs/python-types.md
@@ -252,7 +252,7 @@ The second type parameter is for the values of the `dict`:
 === "Python 3.9 and above"
 
     ```Python hl_lines="1"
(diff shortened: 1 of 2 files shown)


joakimnordling:
Seems that the same work was done in https://github.com/tiangolo/fastapi/pull/4434, which was merged, so closing this.

PR #4404, opened 2022-01-10
fix definition param name for step installs material

solomein-sv:
Step: "Install Material for MkDocs Insiders"
Definition: `github.event.pull_request.head.repo.fork == false && steps.cache.outputs.cache-hit != 'true'`

At step "Dump GitHub context" there is no such field -- `github.event.pull_request.head.repo.fork`, but there is the field `github.event.repository.fork`


tiangolo:
Thanks for your interest! This was handled in https://github.com/tiangolo/fastapi/pull/4410 , so I'll close this one.

PR #4410, opened 2022-01-10
👷 Disable CI installing Material for MkDocs in forks

dolfinus:
A fix introduced in #2340 is not working properly if Github actions are enabled in the forked repo, and repo's author pushed a commit:
https://github.com/dolfinus/fastapi/runs/4766625638?check_suite_focus=true

This is because the if clause is checking `github.event.pull_request.head.repo.fork` context variable, but if workflow is triggered by commit push, there is no pull_request variables at all. Let's add a check for the event name


tiangolo:
Awesome, thank you @dolfinus! :rocket:

PR #4411, opened 2022-01-10
Fix `Response` `status_code` value on `solve_dependencies()`

Kludex:
`None` is not allowed, and now there's a conditional that checks `status_code` on the `Response` initialization. 

Either here or https://github.com/encode/starlette/pull/1402


Kludex:
Error is mainly because of the `RedirectResponse`. The culprit: https://github.com/tiangolo/fastapi/blob/672c55ac626b20cc6219696023cdcc5871e5141d/fastapi/routing.py#L248-L249

Which are assuming `status_code` can be `None`...

Kludex:
More context can be found on https://github.com/encode/starlette/pull/1397

PR #4412, opened 2022-01-11
Fixed the `Dict`/`dict` example in the python typing tutorial.

Luttik:
# Fixed the `Dict`/`dict` example in the python tutorial.
It seemed like previously the tutorial008.py was modified to the py3.9+ version without making a copy of the file.

the current live documentation uses `dict[str, float]` in python versions 3.6+ while this is actually only possible from py3.9 onwards.


Luttik:
@tiangolo could you take a quick look at this? It requires only a very minor check but will definitely help the target group for these pages.

PR #4416, opened 2022-01-12
✨ Don't revalidate the response content if it is of the same type as `response_model`

kaiix:
There's no need to use [`secure_cloned_response_field`](https://github.com/tiangolo/fastapi/blob/0.71.0/fastapi/routing.py#L361) to revalidate the [`response_content`](https://github.com/tiangolo/fastapi/blob/0.71.0/fastapi/routing.py#L127) if the response object type is exactly the same as `response_model`

related issues:

- https://github.com/tiangolo/fastapi/issues/2359
- https://github.com/tiangolo/fastapi/issues/2689
- https://github.com/tiangolo/fastapi/issues/3021
- https://github.com/tiangolo/fastapi/issues/3454
- https://github.com/tiangolo/fastapi/issues/3503

---

Updated comment: https://github.com/fastapi/fastapi/pull/4416#issuecomment-2691027881


fastapi/routing.py
```diff
@@ -116,25 +117,28 @@ async def serialize_response(
     is_coroutine: bool = True,
 ) -> Any:
     if field:
```
cikay, line 119:
```py
if not field:
    return jsonable_encoder(response_content)
```
To decrease the indentation level. The pattern is called early return.

kaiix: good point

fastapi/routing.py
```diff
@@ -116,40 +116,41 @@ async def serialize_response(
...
-            value = response_content
-        return jsonable_encoder(
-            value,
-            include=include,
-            exclude=exclude,
-            by_alias=by_alias,
+    if not field:
+        return jsonable_encoder(response_content)
+
+    if type(response_content) is not response_model:
```
cikay, line 122: Reverse the condition and return early.

```py
if type(response_content) is response_model:
    return jsonable_encoder(
        response_content,
        include=include,
        exclude=exclude,
        by_alias=by_alias,
        exclude_unset=exclude_unset,
        exclude_none=exclude_none,
    )

# put all logics that are in your if statement implementation here.
```

kaiix: Put this case together with `type(response_content) is not response_model` is more clear, they should share the same encoding (`jsonable_encoder`) process.

dolfinus:
@tiangolo any chance this PR will be reviewed?

Tishka17:
Rebuilding a response model is a real problem and cannot actually be done correctly because of aliases and other setting (I guess I've already left a comment about this). In normal code your create it by your own from other sources that could not be matched directly so it is actually not needed. 

Can we expect this to be review and fixed?

nkhitrov:
@tiangolo hi! Can you answer on this comment?

https://github.com/tiangolo/fastapi/pull/4416#issuecomment-1455178994

levsh:
Hi!
I have this problem too (
This can greatly degrade performance for the complex models with validators or endpoints that return a large number of items.
In example below `get_items` endpoint already returns a list of items so it is no need to rebuild response again.
And finally we have response in ~2 second instead of ~1 second. 
```
from typing import List
from uuid import UUID, uuid4

from fastapi import FastAPI
from pydantic import BaseModel, validator

app = FastAPI()

class Data(BaseModel):
    i: int

    @validator("i")
    def validate_i(cls, v):
        sleep(0.1)
        return v

class Item(BaseModel):
    id: UUID
    data: Data

@app.get(
    "/items",
    response_model=List[Item],
)
def get_items():
    return [Item(id=uuid4(), data=Data(i=i)) for i in range(10)]
```
Thank you!

Tishka17:
@tiangolo are there any plans on merging or reimplementing this?

Danipulok:
@tiangolo is there any way this PR could be merged?

tests/test_serialize_response_model.py
```diff
@@ -152,3 +153,32 @@ def test_validdict_exclude_unset():
...
+        @field_validator("switch")
+        def toggle_switch(cls, switch: bool):
+            return not switch
+
+    else:
+        from pydantic import validator
+
+        @validator("switch")
+        def toggle_switch(cls, v):
+            return not v
```
patrick91, line 173: not sure if there's a better way to test this 😊

@kaiix I know this is an old PR, but feel free to suggest something :D

patrick91 (approved):
Looks good to me, I rebased it and update the tests for Pydantic 2 :)

tests/test_serialize_response_model.py
kaiix: happy to see this PR finally has a chance to be merged

I think its enough to illustrate the issue, any better suggestions?  @patrick91

tiangolo: Thanks! Currently, this test passes in `master` :sweat_smile:

We would need a test that doesn't pass in `master` but passes with this PR. :nerd_face:

kaiix: I see, this test worked fine when this PR was created (v0.71.0), years of updates have affected the results.

I'll try to fix the test.

mastizada: @kaiix did you manage to check it? Would you like some help?

svlandeg: @mastizada: we would need a test that fails on `master` and succeeds on this PR, as that will showcase the added value of this PR. If you could come up with such a minimal test, that would certainly help!

kaiix:
When users are using Pydantic v2 (fortunately, most applications should be using v2 by now), this issue doesn’t need much attention.

In v2, [`field.validate`](https://github.com/fastapi/fastapi/blob/68074badcc036b908479b146bd64d61e6d953aa2/fastapi/routing.py#L166) will call [`type_adapter.validator`](https://github.com/fastapi/fastapi/blob/68074badcc036b908479b146bd64d61e6d953aa2/fastapi/_compat.py#L129). If you check its schema, you’ll see that the validator will not revalidate instances of `response_model` (see reference:  [`revalidate_instances`](https://docs.pydantic.dev/latest/api/pydantic_core_schema/#pydantic_core.core_schema.CoreConfig)).

[This is part of the validator schema for the `AutoIncrement` object in the test case]
```
 SchemaValidator(title="AutoIncrement", validator=Model(
    ModelValidator {
        revalidate: Never,
        validator: ModelFields(
            ModelFieldsValidator {
                fields: [
                    Field {
                        name: "count",
                        lookup_key: Simple {
                            key: "count",
...
```

Although there are some additional function calls, it won’t trigger the validator in a way that would modify the original response_content value in certain cases.

Therefore, this fix is only relevant for applications that are still using Pydantic v1.

tests/test_serialize_response_model.py
```diff
@@ -152,3 +153,23 @@ def test_validdict_exclude_unset():
...
+            return count + 1
+
+    @app.post("/increment", response_model=AutoIncrement)
+    async def increment():
+        return AutoIncrement(count=0)
+
+    def test_response_model_should_not_revalidate_response_content_if_they_had_same_type():
+        response = client.post("/increment")
+        response.raise_for_status()
+        assert response.json() == {"count": 1}
```
svlandeg, line 175: Just to confirm: on `master` and with Pydantic 1.10.22, this fails with 
```
AssertionError: assert {'count': 2} == {'count': 1}
```

svlandeg:
The changes from https://github.com/fastapi/fastapi/pull/14099 are interfering with this PR... 🕵️‍♀️ 

[UPDATE]: solved ✔️

svlandeg (commented):
We've recently dropped support for Pydantic v1, so I'm afraid this PR is no longer really useful. @kaiix, thanks again for all the work you've put into this and I'm sorry we weren't able to merge it sooner 🙏

creative-being:
Preparing PR description...

creative-being:
Preparing PR description...

creative-being:
Preparing PR description...

creative-being:
Preparing PR description...

creative-being:
Preparing PR description...

creative-being:
Preparing PR description...

PR #4418, opened 2022-01-12
Support MermaidJS diagrams in Alternative API Docs

devova:
We actively use MermaidJS diagrams for API docs. 
These ones are already supported for FastAPI docs site (https://github.com/tiangolo/fastapi/pull/985).
Given PR add support to ReDoc version of auto API docs.
Screenshot 2022-01-12 at 21 49 05
index-07-redoc-mermaid


fastapi/openapi/docs.py
```diff
@@ -96,8 +97,30 @@ def get_redoc_html(
...
+                        mermaid.render(`mermaid-${{id++}}`, mermaidNode.textContent, svg => {{
+                            mermaidNode.parentNode.outerHTML = svg;
+                        }});
+                    }}
+                    props.dangerouslySetInnerHTML.__html = markdown.innerHTML;
+                    return "";
+                }}
+            }}
+        }}, document.getElementById("redoc-container"));
+    </script>
```
devova, line 123: Snippet is taken from https://github.com/Redocly/redoc/issues/266#issuecomment-631252318

tiangolo:
Hey @devova, this is very clever! Thanks for the interest.

I see that the comment that mentions this says it's a hacky way to support it and it seems it's not officially supported by ReDoc.

I think it would be better, rather than including it in FastAPI itself by default, to publish this trick in a blog post that you write, explaining how to override the exposed ReDoc. :sunglasses: 

As this is a sort-of hacky workaround/script and I'm not deeply involved in it (I don't know the inner workings and what it would take to maintain it, etc), I would rather not inherit the responsibility of maintaining it and making sure it keeps working, it doesn't break anything else, etc.

Given that, I'll close this issue now. But thanks for the effort! :cake:

PR #4431, opened 2022-01-15
Generate a single tag when the parameter is a string

duarte-pompeu:
# Summary

Generates a single tag when the parameter is a string, such that `app.include_router(router, tags="test")` generates tag *test* instead of *t*, *e* and *s*.

An alternative may be to reject such parameter: see #4525.

# Motivation

I often use a single tag in my routers, which even the documentation states is very common.

When I include a router in my app, I often do the mistake of using a string instead of a list, which generates a very weird result, as seen below: 

```python

app = FastAPI()
router = APIRouter()

@router.get("")
def test():
    return {"msg" : "Hello, world."}

app.include_router(router, prefix="/test", tags="test")
```

image

This lead me to try implement a fix for this itch which often annoys me. I don't know how often it happens to other users, but I find it hard to see a use case for the current behavior - hence the PR.

# Description

In this PR, I test if the object is an instance of string, and if such convert it to a list with one element. This results in a single tag, which is the correct behavior in my opinion. See example below for the same code:

```python

app = FastAPI()
router = APIRouter()

@router.get("")
def test():
    return {"msg" : "Hello, world."}

app.include_router(router, prefix="/test", tags="test")
```

image

I also added tests and updated some single-tag examples.

# Request for comments

I'm a first time contributor, so comments are welcome. Specifically:
- is this a desirable improvement?
- could there be unwanted behavior which I'm unaware of?
- should we update documentation to mention that strings are allowed?
- do these tests make sense?

I would also consider the alternative of raising an exception when the parameter is a string, to force the user to use another iterable such as a list or a tuple.

### DIFF
--- a/docs_src/bigger_applications/app/main.py
+++ b/docs_src/bigger_applications/app/main.py
@@ -12,7 +12,7 @@ app.include_router(items.router)
 app.include_router(
     admin.router,
     prefix="/admin",
-    tags=["admin"],
+    tags="admin",
     dependencies=[Depends(get_token_header)],
     responses={418: {"description": "I'm a teapot"}},
 )
--- a/docs_src/bigger_applications/app/routers/items.py
+++ b/docs_src/bigger_applications/app/routers/items.py
@@ -4,7 +4,7 @@ from ..dependencies import get_token_header
 
 router = APIRouter(
     prefix="/items",
-    tags=["items"],
+    tags="items",
     dependencies=[Depends(get_token_header)],
     responses={404: {"description": "Not found"}},
 )
@@ -27,7 +27,7 @@ async def read_item(item_id: str):
 
 @router.put(
     "/{item_id}",
-    tags=["custom"],
+    tags="custom",
     responses={403: {"description": "Operation forbidden"}},
 )
 async def update_item(item_id: str):
--- a/docs_src/bigger_applications/app/routers/users.py
+++ b/docs_src/bigger_applications/app/routers/users.py
@@ -3,16 +3,16 @@ from fastapi import APIRouter
 router = APIRouter()
 
 
-@router.get("/users/", tags=["users"])
+@router.get("/users/", tags="users")
 async def read_users():
     return [{"username": "Rick"}, {"username": "Morty"}]
 
 
-@router.get("/users/me", tags=["users"])
+@router.get("/users/me", tags="users")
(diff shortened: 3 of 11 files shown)


AlrasheedA:
Hi @duarte-pompeu , thank you for your contribution. As a user of FastAPI, I lean towards your alternative solution, which is to raise an exception if a string is passed to tags. Here is my rationale: 

**For raising an exception:**
Iterating through a string to generate character level tags is an unexpected behavior, and as you mentioned, I don't see a use case where such behavior is required. If a string is entered, it will pass silently, without raising an exception or at least a warning, making debugging for new users difficult, and potentially could pass auto testing. 

**Against accepting a string:**
So why not accept a single string as an argument? It may be a desirable behavior, but it is implicit, and implicit behaviors are not the best for developers experience and thus require good justification and should be implemented as a last resort. I don't see enough justification for this use case, since `tags` parameter is plural, so an iterable is explicitly expected. Also, the effort of wrapping a single string into an iterable is minimal.

duarte-pompeu:
Thank you for your feedback @AlrasheedA .

My philosophy is that *graceful recovery* is better than *fast failure*, which is better than *unexpected behavior*.

With that being said, perhaps failing fast with an informative message is more intuitive (as you said, the plural in `tags` points to a group rather than a single one) and requires less changes in documentation and tests. 

I implemented that behavior in #4525, let's see if it gets some attention 😄

acnebs:
What linter are you using where `tags="mytag"` is not being caught?

duarte-pompeu:
> What linter are you using where `tags="mytag"` is not being caught?

Strangely enough, I created a new project in VS code and it appears under problems, which I don't remember happening previously. In this case, it may not be as necessary as I had thought before.

duarte-pompeu:
I still defend that trying to catch the user error and fixing it (this PR) or blocking it (see #4525) would be more user-friendly, which would align with the philosophy of FastAPI.

BilalAlpaslan (approved):

patrick91 (commented):
I'm also in favour of https://github.com/fastapi/fastapi/pull/4525 

especially since the param is called `tags`, and passing a string doesn't feel right. Though I appreciate it being a nice shortcut :D

duarte-pompeu:
Abandoned in favor of #4525

PR #4440, opened 2022-01-17
📝 Add external link to article: Building an API with FastAPI and Supabase and Deploying on Deta

aUnicornDev:
Building an API with FastAPI and Supabase and Deploying on Deta


tiangolo:
Cool, thanks @aUnicornDev !

Issue #3608, opened 2021-07-26
Returning `dataclass` with pydantic-compatible fields (e.g. datetime) fails for FastAPI >= 0.67


himbeles:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to one of:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example & Description

I have encountered an issue when returning `dataclass` responses with some pydantic-compatible (but incompatible with json encoder of Python standard library) fields, e.g. `datetime` objects or `pathlib` file paths.

Here's a self-contained, [minimal, reproducible, example](https://stackoverflow.com/help/minimal-reproducible-example)  where I have added a `datetime` field to the  `Item` class in [tests/test_serialize_response_dataclass.py](tests/test_serialize_response_dataclass.py):

```Python
from datetime import datetime
from fastapi import FastAPI
from fastapi.testclient import TestClient
from dataclasses import dataclass

app = FastAPI()


@dataclass
class Item:
    name: str
    date: datetime


@app.get("/items/no-response-model/object")
def get_no_response_model_object():
    return Item(name="object", date=datetime(2021, 7, 26))


client = TestClient(app)


def test_no_response_model_object():
    response = client.get("/items/no-response-model/object")
    response.raise_for_status()
    assert response.json() == {
        "name": "object",
        "date": datetime(2021, 7, 26).isoformat(),
    }

```

I would 
* Run the test code snippet with Pydantic 0.67
* I would expect it pass because datetime objects are compatible with Pydantic models
* It fails with the following stack trace:
```
[...]

PR #4454, opened 2022-01-20
🐛 Recurse into dataclass when encoding

ConorStevenson:
Context:
PR #3576 added support for dataclasses as response objects.

Problem:
`json_encoder` doesn't recurse into dataclasses as it does for pydantic models, dictionaries, and other collections. Custom encoders aren't applied to attributes on dataclasses.

Solution:
Recurse into dataclasses when encoding. Also change dataclass tests to use dataclasses rather than pydantic models.

Example:
```
from datetime import datetime
from fastapi import FastAPI
import dataclasses
import pydantic

app = FastAPI()

@pydantic.dataclasses.dataclass
class PydanticDataclass:
    timestamp: datetime = datetime.now()

@dataclasses.dataclass
class ClassicDataclass:
    timestamp: datetime = datetime.now()

@app.get("/pydantic", response_model=PydanticDataclass)
async def pydantic_dataclass():
    return PydanticDataclass()


@app.get("/classic", response_model=ClassicDataclass)
async def classic_dataclass():
    return ClassicDataclass()

```

```
$ curl localhost:8000/pydantic
-> {"timestamp":"2022-01-24T20:51:42.316967"}
$ curl localhost:8000/classic
-> TypeError: Object of type datetime is not JSON serializable
```

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -63,7 +63,14 @@ def jsonable_encoder(
             sqlalchemy_safe=sqlalchemy_safe,
         )
     if dataclasses.is_dataclass(obj):
-        return jsonable_encoder(dataclasses.asdict(obj))
+        obj_dict = dataclasses.asdict(obj)
+        return jsonable_encoder(
+            obj_dict,
+            exclude_none=exclude_none,
+            exclude_defaults=exclude_defaults,
+            custom_encoder=custom_encoder,
+            sqlalchemy_safe=sqlalchemy_safe,
+        )
     if isinstance(obj, Enum):
         return obj.value
     if isinstance(obj, PurePath):
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -1,6 +1,7 @@
 import dataclasses
 from collections import defaultdict
 from enum import Enum
+from functools import partial
 from pathlib import PurePath
 from types import GeneratorType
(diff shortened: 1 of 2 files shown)


himbeles:
See also my existing (nearly identical) PR #3607 and issue #3608 from last year. No response until now unfortunately.

don4of4 (approved):
Looks good -- encountered this issue in our projects as well.

himbeles:
Shouldn’t the keyword arguments of `jsonable_encoder` also be passed on into the recursion? See also #3607.

ConorStevenson:
> Shouldn’t the keyword arguments of `jsonable_encoder` also be passed on into the recursion? See also #3607.

Yep, updated it now. Sorry to have duplicated your existing PR!

himbeles:
No problem, just hope it gets merged :)

himbeles:
@don4of4 Thanks for the review. Do you see any chances, that this PR will be merged in the near future?

himbeles:
My PR #3607 got merged. I think you can close this one now! 🚀

tiangolo:
Thanks for the effort @ConorStevenson! :coffee: 

It seems this was solved by https://github.com/tiangolo/fastapi/pull/3607, right? Given that, I'll close this one.

But if you feel there's something else that needs to be done, feel free to create a new PR.

Thanks! :smile:

PR #4455, opened 2022-01-20
draft fastcli

jonatasoli:
## FastApi CLI

The goal is to create a CLI similar to the flask CLI where it is possible to extend more commands through a plugin and have some basic commands, at the moment I just did it and a shell that definitely needs a better solution.
* The CLI was done  using typer
* I'm create fastcli prefix to start commands for example. `fastcli run`
* Create run command
* Create draft shell command


Kludex:
This PR crashed my browser. Be careful when opening the `Files changed` tab.

austinorr:
@jonatasoli this PR looks pretty substantial (+200k lines, whoa!). I cannot load your whole diff, but I did take a look at your fork and it looks like you've committed your whole venv with a ./bin and a ./lib directory. I suspect this is from overuse of `git add .` rather than being more selective to just commit the files with your changes.

If you're willing to try to clean this up, and maybe some more narrative on the purpose of the PR, I bet a maintainer here could take a closer look at your contribution.

jonatasoli:
> This PR crashed my browser. Be careful when opening the `Files changed` tab.

After I created the PR I realized that I had uploaded the lib build files when I did the tests, but I already fixed that.

jonatasoli:
> his PR looks pretty substantial (+200k lines, whoa!). I cannot load your whole diff, but I did take a look at your fork and it looks like you've committed your whole venv with a ./bin and a ./lib directory. I suspect this is from overuse of `git add .` rather than being more selective to just commit the files with your changes.
> 
> If you're willing to try to clean this up, and maybe some more narrative on the purpose of the PR, I bet a maintainer here could take a closer look at your contribution.

Yes thank you for your comments, I saw that I had made this mistake so I left the PR as draft but I only managed to adjust it today. Thanks for the tip I'll try to describe the PR better.

vrslev:
Hey there! This seems to be an add-on to the main package. I’m sure author wouldn’t want to include something like that.

This PR makes two optional dependencies, `uvicorn` and `typer`, not optional. This means lack of flexibility. 

And the main point here. FastAPI is ASGI framework — it is generic way to do async Python web apps. Why would a ASGI framework want to include ASGI server and reimplement whole CLI functionality?

jonatasoli:
> Hey there! This seems to be an add-on to the main package. I’m sure author wouldn’t want to include something like that.
> 
> This PR makes two optional dependencies, `uvicorn` and `typer`, not optional. This means lack of flexibility.
> 
> And the main point here. FastAPI is ASGI framework — it is generic way to do async Python web apps. Why would a ASGI framework want to include ASGI server and reimplement whole CLI functionality?

Hi @vrslev ,

Do you believe then that you should remove the PR and implement it as a library?
I had thought about doing this initially, but just as flask has a simple CLI, I would like to have a CLI to work on top of the fast api, but as you said it creates a dependency with uvicorn which by the way I discovered that the CLI says much more of the asgi server than the framework itself.

vrslev:
Werkzeug contains unoptimized WSGI server for testing historically. Flask is built on top of Werkzeug and `app.run()` method has zero dependency overhead.
Using ASGI server apart from ASGI — it gives more flexibility for developer. 

I don’t think we need separate package that couples FastAPI and Uvicorn. Uvicorn’s CLI is pretty straightforward, and custom scripts will do for further simplification.

jonatasoli:
> Werkzeug contains unoptimized WSGI server for testing historically. Flask is built on top of Werkzeug and `app.run()` method has zero dependency overhead. Using ASGI server apart from ASGI — it gives more flexibility for developer.
> 
> I don’t think we need separate package that couples FastAPI and Uvicorn. Uvicorn’s CLI is pretty straightforward, and custom scripts will do for further simplification.

Honestly, I lost interest in this CLI a little, I started doing it because I had some issues from people asking for something in this sense.
But, today I prefer to use make as a CLI for my projects, I understand what you say to uvicorn but, it only covers the execution, it wouldn't have a part for tests, routes, starting a project and etc as well as the CLIs of flask and django.
But with the makefile I have a simpler CLI, without messing with the framework and covering the external libs of my service stack.

https://github.com/jonatasoli/star-wars-api/blob/main/Makefile

Then I'll look for the issues and pass on this discussion because I think it's valid for everyone who asked about it to understand why a CLI might not make sense in this case.

PR #4463, opened 2022-01-23
✨ Add support of endpoint return type as response_model

Tradunsky:
# Why this PR is open?
Tl;Dr; Generate endpoint response OpenAPI schema based on the endpoint return type.

In order to declare endpoint params' schema, you can just add typing to a param like this:
```python
@app.get("/items/")
async def read_item(skip: int = 0, limit: int = 10):
    return fake_items_db[skip : skip + limit]
```
and it naturally generates OpenAPi spec for parameter types
image

The same applies to the request body, which is very convenient and feels natural and expected by the [principal of least surprise](principles-wiki.net/principles:principle_of_least_surprise) 👍🏻 

However, when you add return type - no response schema generated, return schema stays an arbitrary string.
```python
@app.get("/items/")
async def read_item(skip: int = 0, limit: int = 10) -> List[Item]:
    return Item(fake_items_db[skip : skip + limit])
```
image

Unless you specify `response_model` to generate OpenAPI spec:
```python
@app.get("/items/", response_model=List[Item])
async def read_item(skip: int = 0, limit: int = 10):
    return Item(fake_items_db[skip : skip + limit])
```
image

# What this PR does?
It aligns behaviour of an endpoint parameter and return type to generate OpenAPI spec.
```python
@app.get("/items/")
async def read_item(skip: int = 0, limit: int = 10) -> List[Item]:
    return Item(fake_items_db[skip : skip + limit])
```
image

It is also backward compatible with existing code that has to put `response_model` in FastAPI route decorator.

# How this PR was tested?
`pytest tests/test_response_class_is_return_type.py`
`./scripts/test.sh`
and run modified `docs_src/query_params/tutorial001.py` locally.

WDYT?

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -35,6 +35,7 @@ from fastapi.utils import (
     create_response_field,
     generate_operation_id_for_path,
     get_value_or_default,
+    return_type,
 )
 from pydantic import BaseModel
 from pydantic.error_wrappers import ErrorWrapper, ValidationError
@@ -513,6 +514,7 @@ class APIRouter(routing.Router):
         current_response_class = get_value_or_default(
             response_class, self.default_response_class
         )
+        current_response_model = response_model or return_type(endpoint)
         current_tags = self.tags.copy()
         if tags:
             current_tags.extend(tags)
@@ -525,7 +527,7 @@ class APIRouter(routing.Router):
         route = route_class(
             self.prefix + path,
             endpoint=endpoint,
-            response_model=response_model,
+            response_model=current_response_model,
             status_code=status_code,
             tags=current_tags,
             dependencies=current_dependencies,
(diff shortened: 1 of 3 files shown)


Tradunsky:
oh.. I should have looked at closed PRs, issues and `response_model` design choice documentation first...

https://github.com/tiangolo/fastapi/issues/101
In general, I upvoted for the same arguments people provided in the comments. I think the Principle of least surprise is violated and it is important to fix it, even considering the tradeoff of having both `response_model` and return type hint as. it opens the opportunity for better UX and keeps the original solution for many more sophisticated cases like specifying a model with status and description.

Data-wise, it may help to analyze how often people just need anything more sophisticated than just one model type. 🤔 

Feel free to close this PR though.

Tradunsky:
Found a lot more discussions about this already. 

My opinion still stays the same, but it may be annoying to go over these issues, again and again, so closing this PR.

If anyone finds this, I would recommend looking at the small utility to get this as you need:
https://fastapi-utils.davidmontague.xyz/user-guide/inferring-router/

PR #4474, opened 2022-01-24
♻ Internal small refactor, move `operation_id` parameter position in delete method for consistency with the code

hiel:


tiangolo:
Makes sense, thanks @hiel! :coffee:

PR #4480, opened 2022-01-26
error handler use the response_class of application

ponytailer:
Use the same `response_class` from application setting

### DIFF
--- a/docs/en/docs/release-notes.md
+++ b/docs/en/docs/release-notes.md
@@ -2,6 +2,7 @@
 
 ## Latest Changes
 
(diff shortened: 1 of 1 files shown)


fastapi/exception_handlers.py
```diff
@@ -1,25 +1,34 @@
...
+    *,
+    response_class: Type[Response] = Default(JSONResponse)
+) -> Type[Response]:
     headers = getattr(exc, "headers", None)
     if headers:
-        return JSONResponse(
+        return response_class(
             {"detail": exc.detail}, status_code=exc.status_code, headers=headers
         )
     else:
```
cikay, line 22: Remove else statement here. No need that.

ponytailer: 👌

PR #4482, opened 2022-01-26
📝 Replace HTTP code returned in case of existing user error in docs for testing

TristanMarion:
Change the HTTP code from 400 to 409 to indicate a Conflict


Kludex (approved):

tiangolo:
Nice, thank you @TristanMarion! 🍰

Issue #4477, opened 2022-01-25
Update Starlette to 0.18.0 to solve some issues fastapi will have on FIPs enabled systems


tonybutt:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
No code needed see starlette's release notes.
```

### Description

Please bump starlette version to 0.18.0 to get the haslib.md5() function fixes for FIPs enabled systems

### Operating System

Linux

### Operating System Details

Ubuntu 20.04

### FastAPI Version

0.73.0

### Python Version

3.9

### Additional Context

Starlette recently fixed a FIPs system issue in version 0.18.0. Would like to see starlette bumped a version up so we don't have to fork fastapi.

newsman2:
I also ran into the same problem. Hope to bump to 0.18.0

tiangolo:
Thanks! This was handled in #4483 by @Kludex :rocket: 

This will be available in the next release in a couple of hours, FastAPI `0.76.0` :label:

PR #4483, opened 2022-01-26
⬆ Upgrade Starlette from 0.17.1 to 0.18.0

Kludex:
For reference, set of changes that broke FastAPI:
- https://github.com/encode/starlette/pull/1383 (mypy)
- https://github.com/encode/starlette/pull/1397
- https://github.com/encode/starlette/pull/1395

Related to #4477


fastapi/dependencies/utils.py
```diff
@@ -462,13 +462,10 @@ async def solve_dependencies(
...
-        content=None,
-        status_code=None,  # type: ignore
-        headers=None,  # type: ignore # in Starlette
-        media_type=None,  # type: ignore # in Starlette
-        background=None,  # type: ignore # in Starlette
-    )
+    if response is None:
+        response = Response()
+        del response.headers["content-length"]
+        response.status_code = None
```
Kludex, line 468: @tomchristie FastAPI uses **a** `Response` object that travels through the dependencies, but that `Response` is not the response that will be returned at the end. There's a "cleanup" after the dependencies are solved. You can see the "cleanup" logic below:

https://github.com/tiangolo/fastapi/blob/291180bf2d8c39e84860c2426b1d58b6c80f6fef/fastapi/routing.py#L216-L254

I've selected the whole snippet, from the point we get the `Response` object used in the dependencies (`solve_dependencies` call) until the last 5 lines (on which the values from `Response` are passed to the actual response object that will be used.

This comment is JFYK, I don't expect any action to be taken on Starlette. Also, considering that FastAPI tweaks Starlette internally for some features it has.

fastapi/routing.py
```diff
@@ -126,7 +126,7 @@ async def serialize_response(
         if is_coroutine:
             value, errors_ = field.validate(response_content, {}, loc=("response",))
         else:
-            value, errors_ = await run_in_threadpool(
+            value, errors_ = await run_in_threadpool(  # type: ignore[misc]
```
Kludex, line 129: We added `ParamSpec` to `run_in_threadpool` on Starlette. The error here was:
```
fastapi/routing.py:129: error: "<nothing>" object is not iterable  [misc]
```

yezz123 (approved):
@Kludex 🌟

L1ghtn1ng:
can this get merged soon and released please?

ShadowLNC:
[Starlette 0.19.0](https://github.com/encode/starlette/releases/tag/0.19.0) has now been released, not sure if it's worth performing a further upgrade here, or leaving that for another time.

Kludex:
As per history, Sebastián prefers to bump progressively. Also, those two releases have a lot of changes.

BilalAlpaslan (approved):

mdczaplicki (approved):

Kludex:
> Hello ! I would like to know more regarding the status of this PR. Is there a reason not to merge it ?

Let's wait for @tiangolo :)

tiangolo:
Awesome @Kludex, thank you! :bow: 

I added a filterwarning for pytest complaining about Cryptography so we can run tests while reaching the version that drops support for Python 3.6.

This will be available in the next release in a couple of hours, FastAPI `0.76.0` :label:

PR #4488, opened 2022-01-27
⬆ Upgrade Starlette from 0.18.0 to 0.19.0

Kludex:
Set of breaking changes:
- https://github.com/encode/starlette/pull/1435 (mypy)
- https://github.com/encode/starlette/pull/1431

~Starlette 0.19.0 is not released yet. Feel free to ignore this for now.~


pyproject.toml
```diff
@@ -129,8 +129,8 @@ check_untyped_defs = true
 
 [tool.pytest.ini_options]
 addopts = [
-  "--strict-config",
-  "--strict-markers",
+    "--strict-config",
```
Kludex, line 132:
```suggestion
  "--strict-config",
```

fastapi/exceptions.py
```diff
@@ -1,20 +1,8 @@
...
-    def __init__(
-        self,
-        status_code: int,
-        detail: Any = None,
-        headers: Optional[Dict[str, Any]] = None,
-    ) -> None:
-        super().__init__(status_code=status_code, detail=detail)
-        self.headers = headers
-
+from starlette.exceptions import HTTPException as HTTPException  # noqa
```
rabinadk1, line 5: Is the renaming of import, i.e., `as HTTPException` needed at the last?

tests/test_extra_routes.py
```diff
@@ -32,12 +32,12 @@ def delete_item(item_id: str, item: Item):
 
 @app.head("/items/{item_id}")
 def head_item(item_id: str):
-    return JSONResponse(headers={"x-fastapi-item-id": item_id})
+    return JSONResponse(None, headers={"x-fastapi-item-id": item_id})
```
paulo-raca, line 35: If there is not body, it shouldn't be a JSONResponse:

```suggestion
    return Response(headers={"x-fastapi-item-id": item_id})
```

Kludex: I'm not changing the behavior of the test.

BilalAlpaslan (approved):

Kludex:
I need to update this PR. I'll do it today.

musicinmybrain:
Still works for me, including the newly-ignored warning. Thanks!

fastapi/exceptions.py
mdczaplicki:
```suggestion
from starlette.exceptions import HTTPException
```

What's the reason for `noqa`?

fastapi/routing.py
```diff
@@ -126,7 +126,7 @@ async def serialize_response(
         if is_coroutine:
             value, errors_ = field.validate(response_content, {}, loc=("response",))
         else:
-            value, errors_ = await run_in_threadpool(
+            value, errors_ = await run_in_threadpool(  # type: ignore[misc]
```
mdczaplicki, line 129: How about adding proper type hints for these names?

fastapi/exceptions.py
```diff
@@ -1,20 +1,8 @@
...
 from pydantic import BaseModel, ValidationError, create_model
 from pydantic.error_wrappers import ErrorList
-from starlette.exceptions import HTTPException as StarletteHTTPException
-
-
-class HTTPException(StarletteHTTPException):
-    def __init__(
-        self,
-        status_code: int,
-        detail: Any = None,
```
tiangolo, line 12: There are several examples and places in the code base that expect `detail` to be any JSON-able data, not only a string (as is with Starlette).

So, only for this little type hint we still need the custom `HTTPException` class in FastAPI.

fastapi/exceptions.py
mdczaplicki: @tiangolo, why leave it dangling?

tiangolo:
Thanks for the work @Kludex! :rocket: :tada: 

This is now available in FastAPI version `0.77.0` :bookmark:

PR #4490, opened 2022-01-29
recycle: Move errors list to relevant block

GuilleQP:
Moved the declaration of `errors` list to the first except clause in `fastapi/encoders.py`

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -132,10 +132,10 @@ def jsonable_encoder(
         if isinstance(obj, classes_tuple):
             return encoder(obj)
 
+    errors: List[Exception] = []
     try:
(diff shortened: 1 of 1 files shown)


cikay (commented):
The commit message should be "Move errors list to relevant block"

cikay:
Why do you make three commit. It should be only one. History should not contain redundant changes as much as possible.

four43:
A quick Git tip to condense messages, you can move your git index back but leave the files alone:
```
git reset --soft HEAD~3
```
will back it up 3 commits, you can sub `HEAD~3` for a specific revision hash too.

Then you can commit all your changes at once with a fresh message and force up (because you're over writing the git history by condensing things)

PR #4525, opened 2022-02-06
Reject single string in tags to avoid unexpected behavior

duarte-pompeu:
# Summary

Using a string as a value for parameter `tags` leads to strange results. For example `app.include_router(router, tags="test")` results in tags `t`, `e` and `s`. This PR attempts to remove this unexpected behaviors and guide developers to use lists instead.

This is an alternative to #4431, which converted a string to a list with one string.

# Motivation

I often use a single tag in my routers, which even the documentation states is very common.

When I include a router in my app, I often do the mistake of using a string instead of a list, which generates a very weird result, as seen below: 

```python

app = FastAPI()
router = APIRouter()

@router.get("")
def test():
    return {"msg" : "Hello, world."}

app.include_router(router, prefix="/test", tags="test")
```

image

This lead me to try implement a fix for this itch which often annoys me. I don't know how often it happens to other users, but I find it hard to see a use case for the current behavior - hence the PR.

# Description

In this PR, I assert that the parameter `tags` is not a `string`, to fail fast rather than accepting with unexpected results. See example below for the same code:

```python

app = FastAPI()
router = APIRouter()

@router.get("")
def test():
    return {"msg" : "Hello, world."}

app.include_router(router, prefix="/test", tags="test")


if __name__ == "__main__":
    import uvicorn

    uvicorn.run("test:app", host="127.0.0.1", port=8000, log_level="info", debug=True, reload=True)
```

Output:

```
$ python app3.py 
  File "E:\dev\git\3rdparty\fastapi\app3.py", line 10, in <module>
    app.include_router(router, prefix="/test",tags="test")
  File "E:\dev\git\3rdparty\fastapi\fastapi\applications.py", line 350, in include_router
    self.router.include_router(
  File "E:\dev\git\3rdparty\fastapi\fastapi\routing.py", line 674, in include_router
    assert not isinstance(tags, str), """tags should not be a string: please use ["tag"] instead of "tag"."""
AssertionError: tags should not be a string: please use ["tag"] instead of "tag".
```

I also added tests to assert this behavior.

# Request for comments

I'm a first time contributor, so comments are welcome. Specifically:
- is this a desirable improvement?
- did I miss any method with the `tags` parameter? 
- do these tests make sense?

My first approach was to convert a string to a list (see #4431), but after feedback from @AlrasheedA this seems simpler.


iudeen (commented):
Just a curious question...

The `inlcude_router` method already is type annotated this `tags: Optional[List[Union[str, Enum]]] = None`. Ideally mypy or your IDE should warn.

Do we still need to reject the input?

duarte-pompeu:
Hi @tiangolo ! I noticed your recent activity: is there interest in merging this PR?

I removed an error reported by the linter, not sure how to proceed with the coverage.

duarte-pompeu:
On a second look, I guess the error message could be more specific:

```python

app = FastAPI()
router = APIRouter()

@router.get("", tags="hello")
def test():
    return {"msg" : "Hello, world."}

app.include_router(router, prefix="/test")


if __name__ == "__main__":
    import uvicorn

    uvicorn.run("app2:app", host="127.0.0.1", port=8000, log_level="info", debug=True, reload=True)
```

It currently is:

```
$ python app2.py 
  File "E:\dev\git\3rdparty\fastapi\app2.py", line 7, in <module>
    def test():
  File "E:\dev\git\3rdparty\fastapi\fastapi\routing.py", line 633, in decorator
    self.add_api_route(
  File "E:\dev\git\3rdparty\fastapi\fastapi\routing.py", line 559, in add_api_route
    assert not isinstance(
AssertionError: tags should not be a string: please use ["tag"] instead of "tag".
```

But it could easily be:
```
$ python app2.py 
  File "E:\dev\git\3rdparty\fastapi\app2.py", line 7, in <module>
    def test():
  File "E:\dev\git\3rdparty\fastapi\fastapi\routing.py", line 633, in decorator
    self.add_api_route(
  File "E:\dev\git\3rdparty\fastapi\fastapi\routing.py", line 559, in add_api_route
    assert not isinstance(
AssertionError: tags should not be a string: please use ["hello"] instead of "hello".
```

iudeen (changes requested):
In case we agree this is needed. 

The better way would be to rise a proper Exception rather than assertion. 

>Ref: [Google Style Guide](https://google.github.io/styleguide/pyguide.html#244-decision)
>assert is used to ensure internal correctness, not to enforce correct usage nor to indicate that some unexpected event occurred. If an exception is desired in the latter cases, use a raise statement.

duarte-pompeu:
> Just a curious question...
> 
> The `inlcude_router` method already is type annotated this `tags: Optional[List[Union[str, Enum]]] = None`. Ideally mypy or your IDE should warn.
> 
> Do we still need to reject the input?

Hi @iudeen , thanks for the review.

Previously I didn't get reports for this error and I got it multiple times.

Now I do get a warning, so I'm not sure if it's so useful.

duarte-pompeu:
> In case we agree this is needed.
> 
> The better way would be to rise a proper Exception rather than assertion.
> 
> > Ref: [Google Style Guide](https://google.github.io/styleguide/pyguide.html#244-decision)
> > assert is used to ensure internal correctness, not to enforce correct usage nor to indicate that some unexpected event occurred. If an exception is desired in the latter cases, use a raise statement.

Sorry, I wasn't fully aware of the differences. I updated it, hopefully it's better now.

tests/test_tags.py
```diff
@@ -0,0 +1,29 @@
...
+            ...
+
+
+def test_string_is_invalid_in_include_router_tags():
+    app = FastAPI()
+    router = APIRouter()
+
+    @router.get("")
+    def test():
+        ...
```
duarte-pompeu, line 26: Seems this line causes code coverage to go down - it's not supposed to execute anyway. Would adding `# pragma: nocover` help?

tests/test_tags.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def test_string_is_invalid_in_router_route_tags():
+    router = APIRouter()
+
+    with pytest.raises(TypeError):
+
+        @router.get("", tags="test")
+        def test():
+            ...
```
duarte-pompeu, line 17: Same as other comment.

tests/test_tags.py
duarte-pompeu: Another option would be to send a request and asset it returns a 404, to assure the endpoint was not created 🤔

duarte-pompeu:
Hi @tiangolo : is there interest in merging this feature? Let me know if I can help.

patrick91 (commented):
I wonder if it is still worth doing this now 🤔  since type checkers and linters are quite common these days 😊

duarte-pompeu:
> I wonder if it is still worth doing this now 🤔 since type checkers and linters are quite common these days 😊

I agree, let's close it.

PR #4534, opened 2022-02-08
✨ Add support for `dependencies` in WebSocket routes

paulo-raca:
I've been using dependencies to handle authentication.
But imagine my surprise when I realized my websocket endpoint wasn't authenticated at all?

This commit cherry-picks the `dependencies` chunks from `APIRoute` into `APIWebSocketRoute`

I also made a few minor style nit-picks.

Here is a much-simplified code that demonstrates the issue:

```
import uvicorn
from fastapi import Depends, FastAPI, HTTPException,  WebSocket
from starlette.requests import Request, HTTPConnection
from starlette.responses import PlainTextResponse, Response

def check_permission():
    print("check_permission")
    raise HTTPException(status_code=401, detail="Unauthorized")

app = FastAPI(dependencies=[Depends(check_permission)])

@app.get("/get")
def get_endpoint() -> Response:
    return PlainTextResponse("GET")

@app.websocket("/ws")
async def websocket_endpoint(ws: WebSocket):
    await ws.accept()
    await ws.send_text("Hello")

if __name__ == "__main__":
```

As expected, accessing "/get" results in 401 error:
```
$ curl "http://localhost:8000/get"
{"detail":"Unauthorized"}
```

However the websocket endpoint remains accessible:

```
$ python -m websockets "ws://localhost:8000/ws"
Connected to ws://localhost:8000/ws.
< Hello
Connection closed: 1000 (OK).
```

After applying the patch, the dependencies execute successfully:

```
$ python -m websockets "ws://localhost:8000/ws"
Failed to connect to ws://localhost:8000/ws: server rejected WebSocket connection: HTTP 500.
```

Interestingly the websocket connection fails with code 500 instead of 401. I have addressed it into [a separate PR to starlette](https://github.com/encode/starlette/pull/1478)


fastapi/routing.py
```diff
@@ -366,10 +373,7 @@ def __init__(
             self.secure_cloned_response_field = None
         self.status_code = status_code
         self.tags = tags or []
-        if dependencies:
-            self.dependencies = list(dependencies)
-        else:
-            self.dependencies = []
+        self.dependencies = list(dependencies or [])
```
cikay, line 376: Could you please amend it like below?

```py
self.dependencies = list(dependencies) or []
```

paulo-raca: Sure

cikay: Yeah.

cikay:
```py
self.dependencies = dependencies if dependencies else []
```
I think this is more clear.

cikay: Or maybe
```py
self.dependencies = dependencies or []
```

paulo-raca: This looks better, thanks!

I had to do a minor update because of this change: 
This `dependencies` are now any kind `Sequence`, and `dependencies.copy().extend(...)` only works with lists.
I have updated it to `list(dependencies).extend(...)` instead

cikay: Why do you make new commit instead of update the one and make force push?

paulo-raca: When I'm the reviewer I find it easier if my comments are addressed in separate commits, so that I don't need to read all the changes again.

However that is a matter of personal preference, I will squash everything in a single commit for you

cikay: I just do want to be available reduntant changes in the history. I am not fimiliar with the FastApi convention for this topic. I encourage you to follow the convention of the FastApi.

paulo-raca:
Hello!

Is there anything missing on this PR?

cikay (approved):

BilalAlpaslan (approved):

paulo-raca:
Thank you for the reviews, @BilalAlpaslan @cikay.

I just fixed the merge conflicts, it should be ready for merging

paulo-raca:
@tiangolo, sorry to call you directly, but can I get your eyes on this?

This is a straightforward issue, but it has been bugging me a lot in the last month :sweat_smile: 

Thank you :pray:

yezz123 (approved):

paulo-raca:
ping?

paulo-raca:
Merged master branch again :+1: 

@tiangolo, @Kludex, can I get eyes on this one?

Thanks!

fastapi/routing.py
```diff
@@ -655,21 +660,37 @@ def decorator(func: DecoratedCallable) -> DecoratedCallable:
         return decorator
 
     def add_api_websocket_route(
-        self, path: str, endpoint: Callable[..., Any], name: Optional[str] = None
+        self,
+        path: str,
+        endpoint: Callable[..., Any],
+        *,
+        dependencies: Optional[Sequence[params.Depends]] = None,
+        name: Optional[str] = None,
```
Kludex, line 668: Moving the `name` parameter around is a breaking change (here and on the other places).

paulo-raca: :thinking:  It seemed right to make it a kw-only argument and move it after `dependencies` for consistency with `add_api_route`.

However you are right, this is not the point of this PR, so I'm reverting it

paulo-raca:
ping?

lclbm:
Is there any progress about this issue?
have the same problem

tiangolo:
📝 Docs preview for commit a46341443da7c0a2569d4de735f4c5ccaebe6cea at: https://64861d40ff845b421d97876c--fastapi.netlify.app

fastapi/routing.py
tiangolo: Thanks @cikay for the interest here!

Nevertheless, there's no need to request changes that depend more on taste. I actually prefer the original proposed version, so I'm gonna revert it to that state.

Also, no need to worry about commits, which ones, how many, or the messages... I'll squash everything and update the message when merging.

tiangolo:
Great, thank you @paulo-raca! 🚀 🙇 

This will be available in the next version, FastAPI 0.97.0, released in the next hours (or tomorrow). 🎉

paulo-raca:
Neat, thank you

bastianh:
For me this was a breaking change because some dependencies I have on the router itself hat `Request` as a dependency which is not available on the websocket endpoint.

gjanezic:
I have the same problem as @bastianh. How can we resolve this? Maybe related to #8503?

tiangolo:
@bastianh and/or @gjanezic, please create a new discussion with a simple example I can copy paste and replicate your issue: https://github.com/tiangolo/fastapi/discussions/categories/questions 🙏

PR #4546, opened 2022-02-09
Update tutorial001.py

github4ansh:
Dear team,
Line "return {"token": token}" is giving misinterpretation that read_items path operation is responsible for returning the token. Instead it is name of endpoint defined in tokenUrl (e.x. "login" or "token" etc..) while creating instance of OAuth2PasswordBearer.


four43:
This example seems to be to show how you can retrieve the actual token from the authorizer. It is already obvious the user can do whatever they want in the controller.

github4ansh:
> This example seems to be to show how you can retrieve the actual token from the authorizer. It is already obvious the user can do whatever they want in the controller.

Agreed with you on some part, but still target audience i.e. the developers who are new to fastAPI may not able be grab this concept at first.

github4ansh:
Hi @tiangolo, @four43, thanks for supporting me in my beginner journey of contributing into my all time favorite FastAPI. 
Kindly look into the recent commit made to fix failing unit test. 
Sincere thanks, 
@github4ansh

tiangolo:
Thanks @github4ansh, but as the tutorial is focused on the auth part and handling tokens, the name of the function is only for illustration, but the focus is not really on that part. Given that, I'll pass on this one, but thanks for the interest! ☕

Issue #4557, opened 2022-02-12
When using postponed annotations (from __future__ import annotations), class based dependencies fail to resolve their dependencies with NameErrorr


waweber:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from __future__ import annotations

from fastapi import FastAPI, Depends

app = FastAPI()

CoolType = int

def get_cool_type() -> int:
    return 123

class MyClass:
    def __init__(self, val: CoolType = Depends(get_cool_type)):
        self.val = val

def get_my_class(val: CoolType = Depends(get_cool_type)):
    return MyClass

@app.get("/")
def my_view(mycls: MyClass = Depends(MyClass)):
    return mycls.val

@app.get("/test2")
def my_view_working(mycls: MyClass = Depends(get_my_class)):
    return mycls.val
```

### Description

Run the example app, and you will get a NameError because CoolType cannot be resolved. If you comment out the first view, the program will work correctly. The second view just uses a function to construct and return MyClass instead using MyClass as a dependency directly.

A possible root of the problem is in `fastapi.dependency.utils:get_typed_signature`:

```py3
def get_typed_signature(call: Callable[..., Any]) -> inspect.Signature:
    signature = inspect.signature(call)
    globalns = getattr(call, "__globals__", {})
    ...
```

To resolve the annotations, this function uses the dependency's __globals__ property to look up the names. When `call` is a function, this works just fine. When `call` is a class, it doesn't have a __globals__, so we get an empty namespace as a fallback.

However, when we specify a class as a dependency, what we really care about is `MyClass.__init__`. This is a function, and it has a __globals__ containing all the types we need to resolve its dependencies.

A possible fix for this would be to make this function check if `call` is a class and use it's __init__ method instead.

### Operating System

[...]

PR #4558, opened 2022-02-12
Example fix to postponed annotations for #4557

waweber:
Example fix for https://github.com/tiangolo/fastapi/issues/4557

This works strictly for the problem described in that issue, I didn't test to see if e.g. a class instance with `__call__` defined works or breaks. I imagine there might even be a better way to fix it higher up in the dependency resolution code somewhere, this is just an example.


fastapi/dependencies/utils.py
```diff
@@ -244,7 +244,12 @@ def is_scalar_sequence_field(field: ModelField) -> bool:
 
 def get_typed_signature(call: Callable[..., Any]) -> inspect.Signature:
     signature = inspect.signature(call)
-    globalns = getattr(call, "__globals__", {})
+    
+    if isinstance(call, type) and hasattr(call, "__init__"):
+        globalns = getattr(call.__init__, "__globals__", {})
+    else:
+        globalns = getattr(call, "__globals__", {})
+
```
cikay, line 252:
```py
lookup_object = call.__init__ if isinstance(call, type) and hasattr(call, "__init__") else call
globalns = getattr(lookup_object, "__globals__", {})
```

BilalAlpaslan (approved):

waweber:
I had to re-work this, because it didn't work with dependencies that are class instances with `__call__`.

The full fix is a bit more involved: use the callable's `__globals__` if it exists. Otherwise, if it is a class, use `call.__init__.__globals__`. If not, try `call.__call__.__globals__`.

This works correctly even if you do something weird, like manually re-assign a class's `__init__`

YuriiMotov:
This was fixed, code example from mentioned discussion works well with current version (0.124.0).

https://github.com/fastapi/fastapi/discussions/8588#discussioncomment-15185298

@waweber, thank you for your interest and efforts!

PR #4560, opened 2022-02-13
♻ Move internal variable for errors in `jsonable_encoder` to put related code closer

GuilleQP:
Moved the declaration of `errors` list to the first except clause in `fastapi/encoders.py`


tiangolo:
Cool, thanks! :coffee:

PR #4567, opened 2022-02-14
Add echo to live command to inform that the server is running.

ss-o-furda:
Echo in serve command don't work for me, if run live command.
I decided to add the output of the message, because after running
the command, nothing is displayed in the console,
and this confused me for the first time.
Therefore, IMHO, it will be better to deduce some information message
that the server was started.


tiangolo:
Thanks! I refactored how this works recently because I was using MkDocs incorrectly, and now it shows the live address the way MkDocs does it. So I'll pass on this one, but thanks for the effort! ☕

PR #4573, opened 2022-02-14
🔧 Keep description when endpoint depends schema

jujumilk3:
When endpoint depends schema, Field's descriptions are gone Because when custom schema class extends `pydantic.BaseModel`, class's members turn into `pydantic.fields.ModelField`. So descriptions are omitted from schema class members.

This commit keep their descriptions when it uses as query parameters.


fastapi/dependencies/utils.py
```diff
@@ -245,15 +245,31 @@ def is_scalar_sequence_field(field: ModelField) -> bool:
...
+                default=param.default,
+                annotation=get_typed_annotation(param, globalns),
+            )
+            for param in signature.parameters.values()
+        ]
+    else:
+        from fastapi import Query
+        fields = {}
+        if call.__dict__.get("__fields__"):
+            for param in call.__fields__:
```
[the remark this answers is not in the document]
odiseo0: getattr, right?

fastapi/dependencies/utils.py
```diff
@@ -245,15 +245,31 @@ def is_scalar_sequence_field(field: ModelField) -> bool:
...
+                fields[param] = dict((call.__fields__[param].field_info.__repr_args__()))
+        typed_params = [
+            inspect.Parameter(
+                name=param.name,
+                kind=param.kind,
+                default=Query(fields[param.name].get("default"), description=fields[param.name].get("description")),
+                annotation=get_typed_annotation(param, globalns),
+            )
+            for param in signature.parameters.values()
+        ]
```
cikay, line 272: Also, you check the reversed condition and else. It makes it hard to read. Another thing is typed params should be defined about of if-else statement because only one parameter is different, the different parameter should be defined in the conditions. I have amended it as below.
```py
if inspect.isclass(call):
    from fastapi import Query
    fields = {}
    for param in getattr(call, '__fields__', []):
        fields[param] = dict((call.__fields__[param].field_info.__repr_args__()))

    default = Query(fields[param.name].get("default"), description=fields[param.name].get("description"))
else:
    default = param.default

typed_params = [
    inspect.Parameter(
        name=param.name,
        kind=param.kind,
        default=Query(fields[param.name].get("default"), description=fields[param.name].get("description")),
        annotation=get_typed_annotation(param, globalns),
    )
    for param in signature.parameters.values()
]

```

fastapi/dependencies/utils.py
```diff
@@ -245,15 +245,31 @@ def is_scalar_sequence_field(field: ModelField) -> bool:
...
+        from fastapi import Query
+        fields = {}
+        if call.__dict__.get("__fields__"):
+            for param in call.__fields__:
+                fields[param] = dict((call.__fields__[param].field_info.__repr_args__()))
+        typed_params = [
+            inspect.Parameter(
+                name=param.name,
+                kind=param.kind,
+                default=Query(fields[param.name].get("default"), description=fields[param.name].get("description")),
```
cikay, line 268: What if the `fields` is empty? I recommend the following implementation

```py
# rest of code
default=Query(fields.get(param.name, {}).get("default"), description=fields.get(param.name, {}).get("description")
# rest of code
```

jujumilk3: thx for your review

cikay:
> I just fixed vulnerabilities you guys said. thx for the comments and review.

Could you please define typed_params out of conditions?

cikay:
> > I just fixed vulnerabilities you guys said. thx for the comments and review.
> 
> Could you please define typed_params out of conditions?

`default` parameter needs to be defined in the conditions and passed to `inspect.Parameter`.

jujumilk3:
oh, sure. I just wanted to record this history. i'm gonna force commit it at this night(in my country) with `typed_params`fix.
Thx!

fastapi/dependencies/utils.py
jujumilk3: I've checked this code.
It makes no sense. `param` is inner variable of `for statement` but you're using this out of scope. This code is totally disordered.

And, if variable `call` is a class, `__fields__` returns dict not list.

I think you never test this code.
And most of cases, `call` is not a class so, giving condition at `default` is a bad way.

jujumilk3: Additionally, this code doesn't work.

cikay: > And, if variable call is a class, __fields__ returns dict not list.

OK. Then it should be

```py
for param in getattr(call, '__fields__', {}):
```



> param is inner variable of for statement but you're using this out of scope. This code is totally disordered.
> And most of cases, call is not a class so, giving condition at default is a bad way.

I do not get it.

cikay: > param is inner variable of for statement but you're using this out of scope. This code is totally disordered.

Okay, I got it. 

Please do not move `typed_params` out of conditions.

cikay: If `parameters` dict contain `param.name` value all the time, then go ahead with the following 

```py
default=Query(
    fields.get(param.name, {}).get("default"),
    description=fields.get(param.name, {}).get("description")
)
```

jujumilk3: > Does `parameters` dict contain `param.name` value all the time?

Yes it always contains `param.name`.

Do you know what things are happening when endpoint depends schema? Would you please reply after test yourself and how it works exactly?

cikay: Okay, I tried to test but could not reproduce how `call ` can be a class. 

By the way, [Sourcery](https://sourcery.ai/) AI suggests the below implementation about `parameters` dict.

```py
parameters = {
    param: dict((fields[param].field_info.__repr_args__()))
    for param in fields
}
```

tiangolo:
Thanks for the interest and effort @jujumilk3! :cake: 

I'm not sure I understand the problem. Could you add a test that breaks without this but is solved by this? That way I can understand what is it that you need.

Thanks everyone else for the discussion and interest here! :coffee: 

@cikay thanks for all the interest! Just so you know, there's no need to request changes about commits or messages. I handle the final commit message and I squash before merging, so it's fine having any commit messages. It's also not necessary to ask changes about style that depend on some specific taste or some convention used by others, because I will format everything with Black either way, and I will refactor things if necessary to adjust to the code style and structure here, and to keep everything consistent. So it can end up being a lot of constraining requests to contributors that are not really necessary.

The best help when reviewing PRs would be about checking if the implementation is correct, if there are tests, if the tests are broken before the PR and fixed by the PR, that the explanation of the PR is clear and makes sense, etc.

Thanks all! :cake:

jujumilk3:
@tiangolo

Below capture is current version.
Screen Shot 2022-09-03 at 4 04 19 PM


Below capture is this PR applied version.
Screen Shot 2022-09-03 at 3 56 07 PM


I managed many routes as a schema but description field is being omitted because of some reasons what i wrote in past comments. So i PRed this.


But like you said i found some break tests and bugs. I'm gonna fix this and add new tests for this. I will ask for a review when I finish. Thx.

jujumilk3:
@tiangolo 

I applied it. Can you review it?
Sorry for the frequent mentions.

dotX12:
@tiangolo, https://github.com/dotX12/pyfa-converter
I think you should look at this implementation, there is no need to specify something other than Field (pydantic) in the pydantic schema.

dotX12:
> @tiangolo, https://github.com/dotX12/pyfa-converter I think you should look at this implementation, there is no need to specify something other than Field (pydantic) in the pydantic model.
> 
> ```python
> from datetime import datetime
> from typing import Optional
> 
> from fastapi import FastAPI, UploadFile, File, Form
> from pydantic import BaseModel, Field
> 
> from pyfa_converter import FormDepends, PyFaDepends
> 
> app = FastAPI()
> 
> 
> class PostContractBodySchema(BaseModel):
>     title: str = Field(..., description="Description title")
>     date: Optional[datetime] = Field(
>         None, description="Example: 2021-12-14T09:56:31.056Z"
>     )
> 
> 
> @app.post("/form-data-body")
> async def example_foo_body_handler(
>     data: PostContractBodySchema = FormDepends(PostContractBodySchema),
>     # data1: PostContractBodySchema = PyFaDepends( # OR
>     #         model= PostContractBodySchema, _type=Form
>     #     ),
>     document: UploadFile = File(...),
> ):
>     return {"title": data.title, "date": data.date, "file_name": document.filename}
> ```

and perhaps this functionality can be transferred to fastapi.

dotX12:
and perhaps this functionality can be transferred to fastapi.

lspgn:
Hello @jujumilk3,
I may have found a bug with this implementation:
```python
foo_bar: str = Field(alias="foobar, title="My field")
```
Will cause `param.name` to be the value of the alias. There will be a `KeyError` when getting `query_extra_info`.

Thank you!

jujumilk3:
@lspgn 
Thank you for the review. I will fix the bug, add the corresponding test code, and commit it again.

dflock:
I would really like to be able to add descriptions to query params in our fastapi project, without extensive workarounds. [This](https://github.com/tiangolo/fastapi/issues/4700), which turned into [this](https://github.com/tiangolo/fastapi/discussions/8634), which lead to this PR seems to be the closest to actually getting this upstream.

What else needs to be done to get this - or something else that allows simple query parameter descriptions to be shown in the OpenAPI docs for a FastAPI project - upstream? What can we do to help get this done?

tiangolo:
📝 Docs preview for commit fb7fbb2fcf723be602758d4e2ecd67ae55b891d3 at: https://6488578b51ef870a3d1fb269--fastapi.netlify.app

jujumilk3:
Just applied latest release version (0.97.0)

lspgn:
@jujumilk3 
Thank you again for your work!
I just tested with `pydantic==2.1.1` and `fastapi==0.100.1`.

The move to v2 of Pydantic got rid of `Required`.

It can be redefined with:

```python
Required: Any = Ellipsis
```

And also getting the following error:

```
AttributeError: 'FieldInfo' object has no attribute 'field_info'
```

I looked at the following, the move to Annotations changed quite a few things. Haven't had time to fix entirely yet

```python
# new version with "fix"
dict(fields[param].__repr_args__())
# {'annotation': str, 'required': True, 'title': 'ABC', 'description': 'ABC'}

# current version
dict(fields[param].field_info.__repr_args__())
# [('default', PydanticUndefined), ('title', 'ABC'), ('description', 'ABC'), ('extra', {})]
```

fastapi/dependencies/utils.py
```diff
@@ -208,15 +208,57 @@ def get_flat_params(dependant: Dependant) -> List[ModelField]:
...
+    if len(fields):
+        alias_dict = {}
+        query_extra_info = {}
+        for param in fields:
+            if PYDANTIC_V2:
+                query_extra_info[param] = dict(fields[param].__repr_args__())
+            else:
+                query_extra_info[param] = dict(fields[param].field_info.__repr_args__())
+
+            if "alias" in query_extra_info[param]:
```
YuriiMotov, line 224: There is also `validation_alias` attribute. I think it should also be considered and tested

fastapi/dependencies/utils.py
```diff
@@ -208,15 +208,57 @@ def get_flat_params(dependant: Dependant) -> List[ModelField]:
...
+                if PYDANTIC_V2:
+                    query_extra_info[query_extra_info[param]["alias"]] = dict(
+                        fields[param].__repr_args__()
+                    )
+                else:
+                    query_extra_info[query_extra_info[param]["alias"]] = dict(
+                        fields[param].field_info.__repr_args__()
+                    )
+                alias_dict[query_extra_info[param]["alias"]] = param
+            query_extra_info[param]["default"] = (
```
YuriiMotov, line 234: I think `default_factory` should also be considered

YuriiMotov (commented):
@jujumilk3 did a great job working on this PR!
But I think it's still not enough..
I think this problem should be solved another way (probably on Pydantic's side), because it's really complex issue and there are a lot of things we have to consider to make it fully work without bugs.. 
I think this PR should be closed

YuriiMotov (commented):
As I see it, the original idea of this PR was to make the following work (show the description of `f` field in openapi):

```python

app = FastAPI()

class Item(BaseModel):
    f: str = Query("default", description="This is F.")

@app.get("/item")
async def get_item(item: Item = Depends()):
    return item

def test_():
    openapi = app.openapi()
    parameters = openapi["paths"]["/item"]["get"]["parameters"]

    assert "description" in parameters[0]
    assert parameters[0]["description"] == "This is F."
```
That was a trick with `Depends` that people used in order to define query parameters as Pydantic model.
But since FastAPI 0.115.0 [we can officially use Query parameter models](https://fastapi.tiangolo.com/tutorial/query-param-models/).
We just need to change `Depends` to `Query` (as it's shown in docs) to make the test above working.

Since we now have official way of using Pydantic models to define Query parameters, that trick with `Depends` should be discouraged.
I suggest we close this PR.

There are still some problems with Query parameter models support, but I suggest we create new PRs to address them in order to keep it simple for review.

@jujumilk3, thanks for your efforts!

tiangolo:
Thank you for the interest and work on this @jujumilk3! :rocket: 

And thanks a lot @YuriiMotov for the review, investigation and summary of all the context. Given all that, I'll pass on this one, but thanks everyone for the interest! :cake:

PR #4575, opened 2022-02-14
✨ Update internal `AsyncExitStack` to fix context for dependencies with `yield`

tiangolo:
This also allows catching exceptions like `HTTPException` inside of dependencies with `yield`. :sparkles: 

---

Dependencies with `yield` are converted to `asynccontextmanagers` internally and then are run with `AsyncExitStack`.

But before this change, the `AsyncExitStack` was created on the application's `__call__()`, right before custom user middlewares.

But custom user middlewares are created using a new AnyIO task group, and that creates a new contextvars context.

The call to enter the `asynccontextmanager` is done when calling the internal app, inside of the custom middlewares. But the call to exit the `AsyncExitStack` is done outside, so, the exit code of dependencies with `yield` (the code after `yield`) was run in a different context than the code above `yield`.

Because of this, it was not possible to set a context variable in a dependency with `yield` before `yield` and then reset it or use its value afterwards:

```Python
from contextvars import ContextVar
from typing import Any, Dict, Optional


legacy_request_state_context_var: ContextVar[Optional[Dict[str, Any]]] = ContextVar(
    "legacy_request_state_context_var", default=None
)

async def set_up_request_state_dependency():
    request_state = {"user": "deadpond"}
    contextvar_token = legacy_request_state_context_var.set(request_state)
    yield request_state
    legacy_request_state_context_var.reset(contextvar_token)
```

...that example would throw an error at `legacy_request_state_context_var.reset(contextvar_token)`.

This PR sets up the `AsyncExitStack` in a new middleware that is run inside of the custom user middlewares and is run directly, without a new task group. This way the same contextvar context used to enter the asynccontextmanager created for a dependency with `yield` is also used to run the exit code (after `yield`) of the dependency.

---

**Note**: `contextvars` are very powerful, but also complex and the way they work can be a bit magical and hard to debug. If you can, try **not** to use them, avoid them, pass parameters to functions directly, and save `contextvars` for only the advanced use cases that really need them.

An example of an advanced use case that would need `contextvars` is if you are migrating from Flask to FastAPI (or anything else) and you are using Flask's `g` semi-global variable to store data. And then you use that `g` in some function hidden deep down, and passing parameters directly through all the functions up to that point would be a refactor so big and cumbersome that would prevent the migration entirely. In that case, you can replace `g` with a context var, independent of Flask. This would apply to a migration to FastAPI or to anything else. In the case of FastAPI, you could set any of that `g` data in a context variable instead, in a dependency with `yield`, and then use the context variable in that function deep down instead of Flask's `g` object. You can also do this *with* Flask, and that would make your internal function independent of Flask.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -237,7 +237,7 @@ class FastAPI(Starlette):
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         if self.root_path:
             scope["root_path"] = self.root_path
-            await super().__call__(scope, receive, send)
+        await super().__call__(scope, receive, send)
 
     def add_api_route(
         self,
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -155,7 +155,26 @@ class FastAPI(Starlette):
                 Middleware(
                     ExceptionMiddleware, handlers=exception_handlers, debug=debug
                 ),
-                # Add FastAPI-specific AsyncExitStackMiddleware for dependencies
+                # Add FastAPI-specific AsyncExitStackMiddleware for dependencies with
+                # contextvars.
+                # This needs to happen after user middlewares because those create a
+                # new contextvars context copy by using a new AnyIO task group.
+                # The initial part of dependencies with yield is executed in the
+                # FastAPI code, inside all the middlewares, but the teardown part
+                # (after yield) is executed in the AsyncExitStack in this middleware,
+                # if the AsyncExitStack lived outside of the custom middlewares and
+                # contextvars were set in a dependency with yield in that internal
+                # contextvars context, the values would not be available in the
+                # outside context of the AsyncExitStack.
+                # By putting the middleware and the AsyncExitStack here, inside all
+                # user middlewares, the code before and after yield in dependencies
+                # with yield is executed in the same contextvars context, so all values
+                # set in contextvars before yield is still available after yield as
+                # would be expected.
+                # Additionally, by having this AsyncExitStack here, after the
+                # ExceptionMiddleware, now dependencies can catch handled exceptions,
+                # e.g. HTTPException, to customize the teardown code (e.g. DB session
+                # rollback).
                 Middleware(AsyncExitStackMiddleware),
             ]
         )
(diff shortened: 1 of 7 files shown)


fastapi/applications.py
```diff
@@ -134,6 +136,36 @@ def __init__(
...
+                Middleware(
+                    ExceptionMiddleware, handlers=exception_handlers, debug=debug
+                ),
+                # Add FastAPI-specific AsyncExitStackMiddleware for dependencies
+                Middleware(AsyncExitStackMiddleware),
+            ]
+        )
+
+        app = self.router
+        for cls, options in reversed(middleware):
```
cikay, line 164: Rename `cls` to `klass`, `cls` is usually used to refer to the current class. This is a convention. [I encourage you to go ahead.](https://github.com/django/django/blob/7119f40c9881666b6f9b5cf7df09ee1d21cc8344/django/contrib/gis/sitemaps/views.py#L18)

tiangolo: Thanks for your interest. But I prefer to keep the code as-is, the same way it is in Starlette.

fastapi/middleware/asyncexitstack.py
```diff
@@ -0,0 +1,28 @@
...
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+
+class AsyncExitStackMiddleware:
+    def __init__(self, app: ASGIApp, context_name: str = "fastapi_astack") -> None:
+        self.app = app
+        self.context_name = context_name
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if AsyncExitStack:
```
dimaqq, line 13: It seems to me that `AsyncExitStack` is always true, it’s imported and concurrency provides it from either of two locations, but it’s always provided.

fortzi:
Is it (or will it be) possible to access the contextvars in the view that depends on said dependency? For example:

```python
def my_dep():
    # Define some context var here
    yield

@router.get(/, dependencies=[Depends(my_dep)])
def view():
    # Use context var set in my_dep
```

In my case It's done for carrying log context fields (using structlog)

fastapi/middleware/asyncexitstack.py
Kludex: yeah... :thinking:

PR #4602, opened 2022-02-20
Adds code as query parameter to token request when using OpenIdConnect in docs.

ejsyx:
fix for https://github.com/tiangolo/fastapi/issues/4601


frankie567:
Also experienced the same problem and came to the same solution. Could be worth to take this opportunity to update the whole `oauth-redirect.html` page from the official Swagger one: https://github.com/swagger-api/swagger-ui/blob/3.x/dev-helpers/oauth2-redirect.html

kamikaze:
any success?

ejsyx:
fixed in version 0.80.0

PR #4618, opened 2022-02-25
camel_case_unique_id

sulthonzh:
I think the snake case in SchemaName is not standard compared to another generated schema. Please look at the screenshot below. The schema name with Body_ prefix with another schema name like ProfileInResponse etc., 

image

Thanks, Sulthon


tiangolo:
Hey there! Thanks for the interest.

Yeah, the operation ID is generated from multiple things, that's why it doesn't look that pretty. There's no standard that enforces the casing, it's more about conventions.

It seems this PR was failing, and it seems it now has some conflicts. You might want to solve the conflicts and the tests here, but if you prefer, it could also be done in another PR.

Thanks again! :coffee:

PR #4620, opened 2022-02-25
✨ Preserve path parameters order in openapi.json

poul1x:
This fix preserves path parameters order in `openapi.json` which might be broken in some situations (one of those is shown in my issue). All tests passed.

**Before**:

**After**:


poul1x:
Relates to https://github.com/tiangolo/fastapi/issues/4606

jujumilk3:
This is good but i think you have to remove committed testfile `tests/test_path_parameters_order.py`.

BilalAlpaslan (approved):

poul1x:
@tiangolo I've noticed there are 3 failing tests in master branch (without my code). These errors cause CI/CD unsuccessful checks in my pull request

image

gaborfekete-greehill:
> @tiangolo I've noticed there are 3 failing tests in master branch (without my code). These errors cause CI/CD unsuccessful checks in my pull request
> 
> image

These tests fail because of the indentation of the compared objects, they seem to be the same, it's mostly a linebreak difference. Isn't there an option to compare them without the whitespaces?

tiangolo:
I'm wondering about the implications of this and what else could it affect. :thinking: 

I understand there's nothing on OpenAPI specifying the order that parameters should have in the schema related to the parameters in the path string, they are supposed to be matched by name, not by position/order. I would expect generated clients to have keyword arguments instead of positional ones.

I'm also thinking if there's a different way to achieve this.

For now, it seems there are some conflicts and tests not passing, would you be able to fix those?

PR #4624, opened 2022-02-27
📝 Add link to german article: REST-API Programmieren mittels Python und dem FastAPI Modul

fschuermeyer:
A beginner article in German to get started with the FastAPI with a small Todo API as an example.


tiangolo:
Great, thanks!

PR #4629, opened 2022-03-01
✏ Fix typo in deployment

raisulislam541:


tiangolo:
Good catch, thank you! :mag:

PR #4632, opened 2022-03-01
🌐Translated the documentation page index.html into Russian

filarty:
Hi all, I translated the first page into Russian, if I have time I'll translate it further.


h0rn3t:
I think need to delete russian documentation because of 
fascist war against Ukraine, 38 days of war was died more then 150 children and many civilian, many towns was destroyed

filarty:
> I think need to delete russian documentation because of fascist war against Ukraine, 38 days of war was died more then 150 children and many civilian, many towns was destroyed

I do not treat Russia with love, because I was not born there.
I'm sorry about what happened.
But in other countries there are speaking people, moreover, h0rn3t himself left a message in Russian in his profile, and it is somehow strange to disassemble politics in it.

PR #4637, opened 2022-03-02
Fix test_extra_routes on latest Starlette

paulo-raca:
This modified some endpoints to use raw `Response` instead of JSONResponse on methods that do not have response bodies.

This is necessary on latest starlette versions, as [JSONResponse contents is now a required argument](https://github.com/encode/starlette/pull/1431)


paulo-raca:
Nevermind, this is redundant with #4488 :man_facepalming:

PR #4652, opened 2022-03-05
feat: Allow custom http header name instead of default "Authorization"

### LINKED ISSUE #4651: Allow custom http header name instead of default "Authorization"
### First Check

### Commit to Help

### Example Code

```python
# see wanted code
```

### Description

The http header for the different authentication protocols is currently hardcoded to "Authorization", which on itself is accurate according to the standards ((ex. MDN definition)[https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization]). There are, however, reasons in practice you would want to have a different, custom named header. 

For example, in my case where the FastAPI is hosted behind a proxy which already requires the Authorization header to be set. I would like it to be set in the FastAPI to something else (e.g. like "x-bearer-key" or "foo").

Is it possible to define a custom http header name for authorization, instead of the default "Authorization" header name?

### Wanted Solution

# example code of what would be great
auth_scheme = HTTPBearer(authorization_header='MY-OWN-KEY-INSTEAD-OF-AUTHORIZATION')

### Wanted Code

```python
# example code of what would be great
auth_scheme = HTTPBearer(authorization_header='MY-OWN-KEY-INSTEAD-OF-AUTHORIZATION')
```

### Alternatives

_No response_

### Operating System

Linux, Windows, macOS

### Operating System Details

_No response_

### FastAPI Version

0.74

### Python Version

3.10.x

### Additional Context

_No response_

hanssens:
Fixes #4651 feature request.

Feature allows to set (override the default) http header name from "Authorization" to any other header name.

This is now supported:

* HTTPBase
* HTTPBearer
* HTTPDigest
* OAuth2 / OAuth2AuthorizationCodeBearer / OAuth2PasswordBearer
* OpenIdConnect

**Not supported** is:

* HTTPBasic, because external lib requests.HTTPBasicAuth still requires it to be set as "Authorization".

In general, there are no breaking changes, as the default http header name is by convention still `Authorization`. It simply allows this to be overridden, when needed.

A separate test file has been added for each of the supported authorization types, as with the header changes these could not be appended to each existing test set without introducing breaking changes.


yezz123 (approved):

BilalAlpaslan (approved):

tiangolo:
Thanks @hanssens!

Using the header `Authorization` is not really an arbitrary decision, it's part of the standards, and they depend on it. For example, all the integration with OAuth2 in OpenAPI and the "Authorize" button in Swagger UI depend on it.

If you need to have some type of authorization mechanism that uses a custom header, then it would be a custom solution, it wouldn't really adhere to the standards. And of course, you can do that if you need it, but I would rather not allow and encourage that at the FastAPI level as a default option.

But for your use case, you can instead use an APIKeyHeader: https://github.com/tiangolo/fastapi/discussions/8613#discussioncomment-5153067

Given that, I'll pass on this one, but thanks for your interest! ☕

PR #4657, opened 2022-03-06
Fix bug: Sub app with custom openapi

### LINKED ISSUE #4656: Wrong path at sub app with custom open api
### First Check

### Commit to Help

### Example Code

```python
from fastapi import FastAPI
from fastapi.openapi.utils import get_openapi


app = FastAPI()
sub_app = FastAPI(
    docs_url=None,
    redoc_url=None,
    openapi_url=None,
)


@sub_app.get("/items/")
async def read_items():
    return {"id": "foo"}


@sub_app.get("/openapi.json")
async def openapi(
    # Custom Deps
):
    return get_openapi(
        title="Custom OpenAPI",
        version="0.1",
        routes=sub_app.routes,
        prefix="/sub_app",
    )

app.mount("/sub_app", sub_app)
```

### Description

Open API not including `/sub_app` in the paths

### Operating System

Linux

### Operating System Details

_No response_

### FastAPI Version

0.75.0

### Python Version

Python 3.7.12

### Additional Context

_No response_

LorhanSohaky:
Close #4656


fastapi/openapi/utils.py
```diff
@@ -393,6 +393,7 @@ def get_openapi(
     terms_of_service: Optional[str] = None,
     contact: Optional[Dict[str, Union[str, Any]]] = None,
     license_info: Optional[Dict[str, Union[str, Any]]] = None,
+    prefix: Optional[str] = None,
```
cikay, line 396: Why not empty string as default value? That will make if statement redundant.

LorhanSohaky, line 396:
```suggestion
    prefix: Optional[str] = "",
```

fastapi/openapi/utils.py
```diff
@@ -421,6 +422,8 @@ def get_openapi(
             )
             if result:
                 path, security_schemes, path_definitions = result
+                if prefix:
+                    route.path_format = prefix + route.path_format
```
LorhanSohaky, line 426:
```suggestion
                route.path_format = prefix + route.path_format
```

fastapi/openapi/utils.py
LorhanSohaky: Thanks!

fastapi/openapi/utils.py
```diff
@@ -393,6 +393,7 @@ def get_openapi(
     terms_of_service: Optional[str] = None,
     contact: Optional[Dict[str, Union[str, Any]]] = None,
     license_info: Optional[Dict[str, Union[str, Any]]] = None,
+    prefix: Optional[str] = "",
```
cikay, line 396: No nees optional. Just do this. 
```py
prefix: str=''
```

fastapi/openapi/utils.py
```diff
@@ -421,6 +422,7 @@ def get_openapi(
             )
             if result:
                 path, security_schemes, path_definitions = result
+                route.path_format = prefix + route.path_format
```
cikay, line 425: I suggest using the f-string feature to concatenate strings. Do not create new commits for requested changes. Just amend commits that are the changes requested.

cikay: Go ahead :) https://stackoverflow.com/questions/7947322/preferred-github-workflow-for-updating-a-pull-request-after-code-review

LorhanSohaky: What changes if I use f-string?

fastapi/openapi/utils.py
LorhanSohaky: Why don't you need "Optional"?

fastapi/openapi/utils.py
cikay:
```py
route.path_format = f'{prefix}{route.path_format'
```

LorhanSohaky: Yes, but is there any reason to use this instead of +?

fastapi/openapi/utils.py
cikay: Optional means it can be None. No need to be None.

fastapi/openapi/utils.py
cikay: Yeah. It makes it easy to read.

BilalAlpaslan (approved):

fastapi/openapi/utils.py
LorhanSohaky: Done

yezz123 (approved):

LorhanSohaky:
@tiangolo , can you review?

LorhanSohaky:
@tiangolo , can you see this bug?

tiangolo:
📝 Docs preview for commit 09497db773759152c86e63338739372a2619ba95 at: https://64848a9eaab7f46b156614be--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit 488bd15f10e09fb22311fdd4c0b3a3856d87b1be at: https://64865ebb6038ce684e4f639a--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit b9fbd272535cfa5bab9bc1c3ed9d532e20588a65 at: https://6491dc4fd8f6f600c9c4503d--fastapi.netlify.app

grreeenn:
I see it. Messes with client generator

LorhanSohaky:
This problem prevents the generation of protected documentation

YuriiMotov (commented):
@LorhanSohaky, thanks for your efforts!

I think the situation described in https://github.com/fastapi/fastapi/discussions/8614 is not a bug, but expected behavior (see my reply there).
And there is no need to change something.

YuriiMotov:
> This problem prevents the generation of protected documentation

Could you please elaborate on this? Can you give an example?

tiangolo:
As @YuriiMotov says, this is expected behavior, if you want to have multiple path operations in a single OpenAPI, they should be in the same FastAPI app, using `include_router`, not mounts.

Mounts are actually a lower level feature from Starlette that is not related to OpenAPI and how FastAPI connects everything together.

For now, I'll pass on this one, but thanks for the interest! :cake:

LorhanSohaky:
@YuriiMotov 

Let me describe my problem better...
I have an API with several sub-apps within it. Some of these sub-apps have public documentation, while others only allow access to documentation through authentication.
When the documentation for these sub-apps is generated, it doesn't include the subpath where it's located, so when I try to execute a request through Swagger, it gives an error.

YuriiMotov:
> @YuriiMotov
> 
> Let me describe my problem better... I have an API with several sub-apps within it. Some of these sub-apps have public documentation, while others only allow access to documentation through authentication. When the documentation for these sub-apps is generated, it doesn't include the subpath where it's located, so when I try to execute a request through Swagger, it gives an error.

Could you provide a minimal example of such app?

LorhanSohaky:
> > @YuriiMotov
> > Let me describe my problem better... I have an API with several sub-apps within it. Some of these sub-apps have public documentation, while others only allow access to documentation through authentication. When the documentation for these sub-apps is generated, it doesn't include the subpath where it's located, so when I try to execute a request through Swagger, it gives an error.
> 
> Could you provide a minimal example of such app?

Of course!

Public doc:
image

Private doc:

image

image

**Notice that in the image above, the curl command does not include the path where the sub-app is located.**

```py
from typing import Union, Annotated

from fastapi.openapi.docs import get_swagger_ui_html
from fastapi.openapi.utils import get_openapi
from fastapi.security import HTTPBasic, HTTPBasicCredentials
security = HTTPBasic()

app = FastAPI(
    title="EXAMPLE",
    version="0.1.0",
)


private_subapp = FastAPI(
    title="EXAMPLE PRIVATE",
    version="0.1.0",
    docs_url=None,
    redoc_url=None,
    openapi_url=None,
)


@private_subapp.get("/openapi.json")
async def openapi(
    credentials: Annotated[HTTPBasicCredentials, Depends(security)],
):
    return get_openapi(
        title=f"PRIVATE: OpenAPI",
        version="0.1.0",
        routes=private_subapp.routes,
    )


@private_subapp.get("/docs")
async def get_documentation(
    credentials: Annotated[HTTPBasicCredentials, Depends(security)],
):
    return get_swagger_ui_html(
        openapi_url="/private/openapi.json",
        title=f"PRIVATE: Swagger webhook docs",
    )

@private_subapp.post("/items/{item_id}")
async def create_item(
    item_id: int,
    item: str,
    credentials: Annotated[HTTPBasicCredentials, Depends(security)],
):
    return {"item_id": item_id, "item": item}


app.mount("/private", private_subapp)


@app.get("/")
def read_root():
    return {"Hello": "World"}


@app.get("/items/{item_id}")
def read_item(item_id: int, q: Union[str, None] = None):
    return {"item_id": item_id, "q": q}
```

YuriiMotov:
@LorhanSohaky, thanks for the code example!

The problem here is related to the fact that you are overriding the `openapi.json` endpoint, and your implementation is simplified compare to default implementation.

If you look at the default implementation of `openapi.json` endpoint, you will see that before calling `get_openapi` (it's inside `self.openapi()`) it updates `servers` by adding `url` equal to `root_path`.

https://github.com/fastapi/fastapi/blob/d5d302c3da11262f4e320a9af5dbb9d13386e1b5/fastapi/applications.py#L999-L1011

The implementation shown in docs misses this part. It works well for apps that are mounted at `'/'`, but fails in other case.

So, to fix your problem you need to change the implementation of `/openapi.json` endpoint to something like:
```py
@private_subapp.get("/openapi.json")
async def openapi(
    credentials: Annotated[HTTPBasicCredentials, Depends(security)],
    req: Request,
):
    # The following 3 lines may be moved to lifespan event
    app = cast(FastAPI, req.app)
    urls = (server_data.get("url") for server_data in app.servers)
    server_urls = {url for url in urls if url}

    root_path = req.scope.get("root_path", "").rstrip("/")
    if root_path not in server_urls:
        if root_path and app.root_path_in_servers:
            app.servers.insert(0, {"url": root_path})
            server_urls.add(root_path)

    return get_openapi(
        title=f"PRIVATE: OpenAPI",
        version="0.1.0",
        routes=private_subapp.routes,
        servers=app.servers,  # <- added
    )
```

I think we should probably mention this somewhere in docs.
Thanks for bringing this to our attention!

PR #4670, opened 2022-03-09
✏ Fix links to Pydantic docs

kinuax:
`pydantic-docs.helpmanual.io/#json-serialisation` is not redirecting to the expected content on Pydantic docs. `pydantic-docs.helpmanual.io/#` links are also updated to avoid redirections and to keep homogeneity.


tiangolo:
Awesome, thank you! :rocket:

PR #4710, opened 2022-03-21
Add SPDX style license identifier

ccorbacho:
Many automated scanning tools look for an SPDX style license identifier to
find the license being used by a package in an automated fashion
(even though PEP 639 is still in draft).


tiangolo:
Thanks! This was already handled in a recent change moving to Hatch. So I'll close this one, but thanks for the interest! :cake:

PR #4711, opened 2022-03-21
deleting a route from APIRouter

sobhansaf:
There are some scenarios in which someone may need to delete a route from APIRouter. For example, you may use a third-party library that provides a router for you. But you do not want to use all of the provided endpoints. In this case, it is useful to remove some of the endpoints


BilalAlpaslan (changes requested):
it could be better with test

BilalAlpaslan (approved):

fastapi/routing.py
```diff
@@ -645,6 +645,13 @@ def decorator(func: DecoratedCallable) -> DecoratedCallable:
 
         return decorator
 
+    def delete_api_route(self, path: str):
```
cikay, line 648: I would name it `exclude_routes` which should accept an iterable.

fastapi/routing.py
```diff
@@ -645,6 +645,13 @@ def decorator(func: DecoratedCallable) -> DecoratedCallable:
...
+        for i in range(len(self.routes)):
```
cikay, line 649: Use built-in `enumerate` function.

fastapi/routing.py
sobhansaf: I also changed the data structure I use. Because if it should accept iterable, it is more efficient to use a dictionary to check if a path exists in self.routes or not.

tiangolo:
Thank you @sobhansaf! 🍰

I haven't seen a use case like this... I would think it would make more sense for the creator of the router to provide multiple routers so that users can include the functionality they want, or a way to customize the router, not at the FastAPI level... as it would add more complexity in handling and understanding when is a router and all its routes included or not.

PR #4725, opened 2022-03-25
Workaround to fix Pyright/Pydantic missing parameters issue

jtv8:
Users of Pyright, or the Pylance extension in Visual Studio Code, may have noticed that when type checking is activated certain models in `fastapi.openapi.models` complain about missing parameters that are actually optional.

Most notably, the `Schema` object cannot be created programmatically without a type checking error as Pyright/Pylance believes that it is missing an required keyword argument, `$ref`, that cannot be legally provided in Python.

> `Arguments missing for parameters "$ref", ...`

This is due to a known compatibility issue with a workaround that is documented here:
https://github.com/samuelcolvin/pydantic/issues/3753

This PR implements that workaround for the models in `fastapi.openapi.models`.


yezz123 (approved):

jtv8:
Resolved by https://github.com/tiangolo/fastapi/pull/4906

Issue #3910, opened 2021-09-16
Would be nice to be able to route request using header's Accept field (or generic header's field)


mcosta74:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
N.A.
```

### Description

In some cases would be nice to specify header's field as routing rules.

One important example is to support API versioning based on Header's Accept field

### Wanted Solution

Ability to specify some header's fields in the `.get()`, `.post()` .... decorators

### Wanted Code

```python
from fastapi import FastAPI

app = FastAPI()


@app.get("/", accept="application/json;version=1.0")
async def root():
    return {"message": "Hello World v1.0"}

@app.get("/", accept="application/json;version=1.1")
async def root():
    return {"message": "Hello World v1.1"}
```

### Alternatives

from fastapi import FastAPI

app = FastAPI()

@app.get("/", headers={"accept": "application/json;version=1.0"})
async def root():
    return {"message": "Hello World v1.0"}

@app.get("/", headers={"accept": "application/json;version=1.1"})
async def root():
    return {"message": "Hello World v1.1"}

### Operating System

macOS

### FastAPI Version

python -c "import fastapi; print(fastapi.__version__)"

### Python Version

Python 3.9.7

ycd:
You still have access to request object, couldn't you use that to retrieve `content-type` and implement your logic as if there are two different endpoints?

mcosta74:
> You still have access to request object, couldn't you use that to retrieve `content-type` and implement your logic as if there are two different endpoints?

well, it's not so easy.
Usually new version of the API means breaking changes in the request body and/or response body.

Handle it in a single endpoint will end-up in a "messy" code.

For instance how to define different input body depending of the version?

ghandic:
[...]

PR #4727, opened 2022-03-25
✨ Merge openapi descriptions for routes with same path and method.

plankthom:
This feature allows merging the open-api specification for routes
with the same path and method.
The primary motivation was to allow separate route definitions
per response content (routed on the Accept header), resulting in separate sections per accepted mediatype in the openapi.json at
> `paths.<path>.<method>.responses.<status_code>.content.<media-type>`

See #4694 and #3910


plankthom:
The current merge will not cover all cases, e.g. one could image that two schemas that need to be merged in the same request parameters or response content should be merges with a `oneOf` json schema keyword.

However, a user of fast api could register such adapted or new merge methods  in
https://github.com/tiangolo/fastapi/blob/d77a6720149436a2cdc0a8460d30efbc411d27e0/fastapi/openapi/utils.py#L565-L575

plankthom:
@tiangolo would this PR be welcome?

devoredevelops:
Very interested in seeing this as a separate package if not in core fastapi.

fastapi/openapi/utils.py
```diff
@@ -437,3 +453,123 @@ def get_openapi(
...
+    summary: Optional[str] = None
+    for update_summary in summary_by_id.values():
+        if update_summary and summary != update_summary:
+            summary = f"{summary} / {update_summary}" if summary else update_summary
+    return summary
+
+
+def merge_description(desc_by_id: Dict[str, Optional[str]]) -> Optional[str]:
+    desc: Optional[str] = None
+    for update_id, update_desc in desc_by_id.items():
```
mvdbeek, line 489: Since you're not accessing update_id you could do
```suggestion
    for update_desc in desc_by_id.values():
```

mvdbeek:
This is very nice and also works well for documenting routes that accept request bodies with different encodings, e.g application/json and multipart/form-data.

BryanEddy5:
Any update on this?

YuriiMotov (commented):
@plankthom, thanks for your contribution! I find the idea interesting.

I think we should start from the question "Whether FastAPI is going to support routing by content type?".
The way it's implemented in the code example in this PR is complex and not FastAPIsh.

If we want to support this - we should simplify this process. If not - we should probably provide some recommendations on how to achieve this in clearer way.

Let's wait for Sebastian to review the idea

tiangolo:
Thanks for the interest here! ☕

There are several refactors I have planned that will affect how this code works, plus there's a risk of edge cases breaking. So for now, I prefer to pass on this one until I get that done. Thanks! ☕

PR #4754, opened 2022-04-01
➖ Temporarily remove typer-cli from dependencies and upgrade Black

tiangolo:
➖ Temporarily remove typer-cli from dependencies until it supports Click 8.x.x, to unblock pydantic CI

Also upgrade Black, as it has the same type of Click conflicts as Typer


78877i (commented):
#Bitcoin1e-12/50btc[https://cointelegraph.com/news/secure-bitcoin-self-custody-balancing-safety-and-ease-of-use](lawup@rada.gov.ua) @tiangolo

PR #4758, opened 2022-04-04
🔧 Allow rough response model

jujumilk3:
for example

```python
import uvicorn

from typing import Optional
from fastapi import FastAPI


app = FastAPI()


class Item(BaseModel):
    name: Optional[str]
    price: Optional[float]


@app.get("/items/{item_id}", response_model=Item)
async def get_item(item_id: int):
    return {
        "name": "this is name",
        "price": "this is price"
    }
```

This return error by pydantic model's validator. But In many cases, the response model and the actual return value were different. So i made this PR.

This method can use like this.

```python
import uvicorn

from typing import Optional
from fastapi import FastAPI


app = FastAPI()


class Item(BaseModel):
    name: Optional[str]
    price: Optional[float]


@app.get("/items/{item_id}", response_model=Item, allow_rough_response_model=True)
async def get_item(item_id: int):
    return {
        "name": "this is name",
        "price": "this is price"
    }
```
This returns
```json
{
  "name": "this is name",
  "price": "this is price"
}
```

Look at the `allow_rough_response_model`. Default value of this is False but if set it as True, This endpoint ignore pydantic model validator.


odiseo0:
I like this.
To be clear, the way it works is that when the Pydantic model fails, instead of raising an exception, what it does is to send the dictionary as it is returned?

tjgmlfbs124:
I like this too.

damon-at-aidkr:
I have encountered many cases where this function is necessary.

ddanier:
I'm not getting it.

The idea of setting the response model in the API endpoint is to define the contract you as the backend developer has with the client using the API. This contract in this case defines there is a field `price` thats either a `float´ or `None`. If you just return a string instead your are actually breaking that contract and by doing so possibly breaking the client.

Why would you like to do that?

And if you need to have a different return type for some reason: Why shouldn't you define a new DTO response model (DTO=data transfer object) just for that case - so the API spec matches what you are really doing. Could be like so:
```python
class ItemDTO(Item):
    price: Optional[Union[float, str]]
```

Could you please explain?

damon-at-aidkr:
Exactly right what you said all. But you know There's no Always good for conditions to dev as-is.

Like you said, define new DTO is the right way If i need little changed from original DTO. But There will be so many duplicated code if i make DTO every single time when i have to define new DTO.

For example like above, if i have to make PUT method endpoint with that DTO, i have to make new DTO without `Optional`.
like this

```python
class PutItem(BaseModel):
    name: str
    price: float
```

so i get two very similar DTO. Of course, in this case i can make every attrs to optional by use metaclass. But if i have to add some attr in specific endpoint, Then i have to define new class too. I can solve it class inheritance. Like

```python
class MoreAttr(Item):
    additional_attr: int
```

problem solved. But, if i have to define new response model with very complicated hierachy DTO? It includes n:n fields and n:1 fields. Then i have to define new DTO per every model it uses.


I know the best way but, not so good for it at every cases. I just suggest a new option for developers to handle this problem.

jujumilk3:
@ddanier 

Exactly right what you said all. But you know There's no Always good for conditions to dev as-is.

Like you said, define new DTO is the right way If i need little changed from original DTO. But There will be so many duplicated code if i make DTO every single time when i have to define new DTO.

For example like above, if i have to make PUT method endpoint with that DTO, i have to make new DTO without `Optional`.
like this

```python
class PutItem(BaseModel):
    name: str
    price: float
```

so i get two very similar DTO. Of course, in this case i can make every attrs to optional by use metaclass. But if i have to add some attr in specific endpoint, Then i have to define new class too. I can solve it class inheritance. Like

```python
class MoreAttr(Item):
    additional_attr: int
```

problem solved. But, if i have to define new response model with very complicated hierachy DTO? It includes n:n fields and n:1 fields. Then i have to define new DTO per every model it uses.


I know the best way but, not so good for it at every cases. I just suggest a new option for developers to handle this problem.

ddanier:
And I just have to add this:  
"Explicit is better than implicit."  
(https://peps.python.org/pep-0020/)

So using an explicit model definition just make sense and is the more pythonic way I think. ;-)

jujumilk3:
Thanks for your Sincere comment. I read it couple of times. "Explicit is better than implicit." is perfectly fit in this case. 
I have nothing to say anymore because you are creating every cases of DTOs in your work actually.

Thank you again. 😊

ddanier:
@jujumilk3 I published https://github.com/team23/pydantic-partial to allow creating partial models where not all fields are required any more. This way you can create the necessary DTO's just be converting your existing models to being partials. It's still not perfect and may miss some things and I'm not sure it will really help you in your case, but feel free to give it a try. ;-)

tiangolo:
Thanks for the interest @jujumilk3 and everyone! :cake: 

The problem is that as @ddanier says, this would break the contract, which is the only point of having a `response_model` there.

If you need to return arbitrary data as in this case, you can just omit the `response_model`. That way the data you return won't be validated with the response model, serialized, filtered, nor documented in OpenAPI, Swagger UI. And that's what you need, because otherwise you would be validating, documenting, etc. something, so that the client can expect that, but in the end you are not returning it.

Now, if for some reason you want to include the response model, maybe for the docs as a possible response (even if the client can't/shouldn't rely on that), you can declare it with `response_model=Union[Item, Any]`. And that will tell FastAPI/Pydantic to validate it as an `Item`, or as `Any`, and `Any` will always be valid, because it can just be anything. :sweat_smile: 

I think that should probably achieve what you want. And the extra benefit is that OpenAPI, automatic client generators, etc. All those things will still have the information that this could have those types but also that it **could not have them and be anything**. So your API is not promising something that it's not fulfilling. It's just giving an example of what it could send back, but not committing to it.

Here's an updated version of your same example with that:

```Python
from typing import Any, Optional, Union
from fastapi import FastAPI


app = FastAPI()


class Item(BaseModel):
    name: Optional[str]
    price: Optional[float]


@app.get("/items/{item_id}", response_model=Union[Item, Any])
async def get_item(item_id: int):
    return {
        "name": "this is name",
        "price": "this is price"
    }
```

That one accepts the data as is without server errors. And it also documents it like that.

---

Given all that, I'll close this PR. But thanks for the interest and effort! :coffee:

PR #4784, opened 2022-04-11
Fix data param for tests

mbrancato:
This PR fixes use of `client.post` in the testing documentation which incorrectly passes a dictionary to the `json` parameter instead of the `data` parameter.


mbrancato:
I was wrong about this. This issue I hit was `bytes` objects in the dictionary, and I thought this was the cause as the ode comments indicated that `data` should be a dictionary and not `json`.

Issue #4635, opened 2022-03-02
Incorrect openapi.json when using multiple APIKeyHeaders with different names


nightbluepn:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from fastapi import FastAPI, Security
from fastapi.security import api_key

app = FastAPI()

api_key_header_a = api_key.APIKeyHeader(name='key_a')
api_key_header_b = api_key.APIKeyHeader(name='key_b')


@app.get("/a")
async def a(auth_token: str = Security(api_key_header_a)):
    pass


@app.get('/b')
async def b(auth_token: str = Security(api_key_header_b)):
    pass
```

### Description

- Start the app with `uvicorn main:app`
- Go to http://127.0.0.1:8000/openapi.json
- Note that the specification only contains one of the two api key headers
- Note that http://127.0.0.1:8000/docs hence doesn't work properly: any authorization sets the header 'key_b'

I would expect that the openapi specification contains both api key headers, and that that in the Swagger UI, clicking the lock on endpoint 'a' sets header 'key_a' and the lock on endpoint 'b' sets header 'key_b'.

I also want to mention that the code itself works fine, only the openapi specification is incorrect.

### Operating System

Linux

### FastAPI Version

0.73.0

### Python Version

3.9.10

### Additional Context

![screenshot_2022-03-02-115416](https://user-images.githubusercontent.com/26491995/156348396-d1d3d7af-b209-480d-9f1c-129ef623c721.png)
As we can see, the Swagger UI wants to authorize against endpoint 'a' by setting 'key_b', which is not what is defined in the code.

BilalAlpaslan:
you should set schema name for openapi
```py
from fastapi import FastAPI, Security
from fastapi.security import api_key

[...]

PR #4791, opened 2022-04-12
🐛 fix APIKey scheme name default

BilalAlpaslan:
```py
from fastapi import FastAPI, Security
from fastapi.security import api_key

app = FastAPI()

api_key_header_a = api_key.APIKeyHeader(name='key_a')
api_key_header_b = api_key.APIKeyHeader(name='key_b')


@app.get("/a")
async def a(auth_token: str = Security(api_key_header_a)):
    pass


@app.get('/b')
async def b(auth_token: str = Security(api_key_header_b)):
    pass
```
before schema name default is class name and second APIKey crush firsth APIKey


BilalAlpaslan:
related issue #4635

yezz123 (approved):
Its look clean and work as expected 🚀

JESSIEALOPAY (approved):

yusufberad (approved):

BilalAlpaslan:
related discussion: https://github.com/tiangolo/fastapi/discussions/6335

BilalAlpaslan:
@Kludex can u review

hasansezertasan (approved):

esrefzeki (approved):

hasansezertasan:
I think it's stuck.

BilalAlpaslan:
Hi @tiangolo, it almost 2 year and still this little bug in fastapi: https://github.com/tiangolo/fastapi/discussions/9012#discussioncomment-7553951
can u review it :grin:

karakuss (approved):

BilalAlpaslan:
Hi @tiangolo. People can not create a multiple APIKey. Actually this pr report a bug not an feature request.  I mentioned that because this pr has a feature label :upside_down_face:  .

zekiyasin (commented):
nice

emrahyurttutan:
nice @BilalAlpaslan

emrahyurttutan (approved):

zekiyasin (approved):

YuriiMotov (changes requested):
This solution will fail if you use `APIKeyHeader` and `APIKeyQuery` (two different classes) with the same token key name:
```
from fastapi import FastAPI, Security
from fastapi.security import api_key

app = FastAPI()

api_key_header = api_key.APIKeyHeader(name="key_a")
api_key_query = api_key.APIKeyQuery(name="key_a")


@app.get("/a")
async def a(auth_token: str = Security(api_key_header)):
    pass


@app.get("/b")
async def b(auth_token: str = Security(api_key_query)):
    pass
```

`name` attribute of `APIKeyHeader` (and others) is a name of parameter that contain token.

I think this will solve the problem:
```
        self.scheme_name = scheme_name or f"{self.__class__.__name__}_{name}"
```

BilalAlpaslan:
> This solution will fail if you use `APIKeyHeader` and `APIKeyQuery` (two different classes) with the same token key name:
> 
> ```
> from fastapi import FastAPI, Security
> from fastapi.security import api_key
> 
> app = FastAPI()
> 
> api_key_header = api_key.APIKeyHeader(name="key_a")
> api_key_query = api_key.APIKeyQuery(name="key_a")
> 
> 
> @app.get("/a")
> async def a(auth_token: str = Security(api_key_header)):
>     pass
> 
> 
> @app.get("/b")
> async def b(auth_token: str = Security(api_key_query)):
>     pass
> ```
> 
> `name` attribute of `APIKeyHeader` (and others) is a name of parameter that contain token.
> 
> I think this will solve the problem:
> 
> ```
>         self.scheme_name = scheme_name or f"{self.__class__.__name__}_{name}"
> ```

Actually no one can create two variables with the same name. The same way, people doshould not be able to create two schema names with same name.  Maybe we can throw an error when someone used the same names more than two.

YuriiMotov:
Sorry, I don't quite understand what you mean.

In the example I mentioned before, endpoint "/a" expects the authorization token to be passed as a Header parameter with the name `key_a`. Endpoint "/b" expects the authorization token to be passed as a Query parameter with the name `key_a`. It's absolutely OK.
And talking about 'schema name', in our case schema name is not specified by user (parameter `scheme_name` is omitted), so it's FastAPI who sets this name. And it should be unique for all combinations of `Security scheme` and `parameter name`.

YuriiMotov:
I've looked at this PR again, and now I think it should be one of two solutions:
* "class name" + "param name" + "auto error"
* "class name" + index for duplicates. For example, for initial code example we will have `APIKeyHeader` and `APIKeyHeader_2`

I think that second is more preferable since it's not braking change (it will not change scheme name if there is only one instance of this class in the project).
To implement this I think we can use counter stored in class attribute.

@BilalAlpaslan, what do you think?

PR #4794, opened 2022-04-13
♻️ Refactor `include_router` to mount sub-routers

tiangolo:
This is a work in progress. :construction: 

## Description

Currently when a FastAPI app or a router uses `router.include_router(some_sub_router)` each of the routes in `some_sub_router` is re-created in `router`, so it is recreated in the parent router.

The intention of this PR is to mount sub-routers using Starlette's regular mount.

This would allow several things:

* Use custom router classes.
* Use custom route classes.
* Add dependencies to the router that could be executed even if there's no matching route, for example for logging.
* It would allow adding middleware per router, which currently isn't supported, not even by Starlette.
* It could possibly allow a future refactor to include a router at the top of the file before adding routes to it, currently using `.include_router()` requires calling it after all the routes are added, at the end of the file, or at another file after importing it (but this is still not possible in this PR).

## Alternative implementations and problems

There are several (too many) particular corner cases and exotic use cases that are currently supported by FastAPI, which makes this refactor/implementation non-trivial when trying to preserve compatibility for everything currently supported.

For that, I considered several approaches, but most of them have drawbacks and caveats.

## General Caveats

Adding dependencies to a router means that it could require additional data from a request, e.g. more headers, or additional body fields. And when generating the OpenAPI for a specific route, it would need to have access to all that to generate the schema properly.

At the same time, ideally, the body should be validated/parsed by pydantic once, in a single point, and collect all the validation errors. So the body field should be generated once with all the fields, it wouldn't be great if the router with its dependencies has some body field and/or some additional request requirements, validates them, returns an early error, but doesn't include any other validation errors from the actual route.

Right now, for each route, a pydantic *field* is created for each route with a body. It's a pydantic field and not a pydantic model because a field can also contain a `List[X]` (a JSON array), not only JSON objects. But that field is created once when the route is created.

All this is complicated by the fact that in pure Starlette each route doesn't have knowledge of its router, and in each router it doesn't have knowledge of any mount, or parent router. This information is (or could be) available at runtime from the ASGI scope, per request, but it would be ideal to build/compile the pydantic model once at the beginning, when starting the app, not every time there's a request.

## Features to support

The idea is to be able to support the features described above. But on top of that, ideally, to support currently supported corner-case features, like:

* Define a router prefix when creating the router (e.g. `APIRouter(prefix="/items")`).
* Define a router prefix when including it (e.g. `router.include_router(sub_router, prefix="/items")`).
* Support empty paths, as described in: https://github.com/tiangolo/fastapi/issues/414, to support a router with prefix `/items` that has a path `/items` (instead of only supporting `/items/` with the trailing slash)

```Python
router = APIRouter(prefix="/items")

@router.get("")
def get_all_items():
    return []
```

**Note**: this particular feature that is currently supported makes using the current Starlette mounts quite problematic, and that's where the current failing tests in this PR are.
* Support including a router more than once under different prefixes. For example, here the router is used at `/items` and `/v2/items`:

```Python

app = FastAPI()

router = APIRouter()


@router.get("/")
def get_all_items():
    return []


@router.get("/{item_id}")
def get_item(item_id: int):
    return item_id


app.include_router(router, prefix="/items")
app.include_router(router, prefix="/v2/items")
```

* Support including a router more than once, under different prefixes **with different custom dependencies, tags, etc**:

```Python
from fastapi import APIRouter, Depends, FastAPI, Header

app = FastAPI()

router = APIRouter()

def new_auth(x_header: str = Header(...)):
    return {"x_header": x_header}


@router.get("/")
def get_all_items():
    return []


@router.get("/{item_id}")
def get_item(item_id: int):
    return item_id


app.include_router(router, prefix="/items")
app.include_router(router, prefix="/v2/items", dependencies=[Depends(new_auth)])
```

* Support including a router without a prefix. That would make it live in the top level, with any other routes, without concluding them. For example, two routers could share the same path `/items/`, one with a dependency and one without. This would not be supported by standard Starlette mounting.

```Python
from typing import List
from fastapi import APIRouter, Depends, FastAPI, Header

app = FastAPI()


def new_auth(x_header: str = Header(...)):
    return {"x_header": x_header}


protected_router = APIRouter(dependencies=[Depends(new_auth)])
simple_router = APIRouter()


@simple_router.get("/items/")
def get_all_items():
    return []


@protected_router.post("/items/")
def create_items(item_ids: List[int]):
    return item_ids


app.include_router(simple_router)
app.include_router(protected_router)
```

## Recreate each route

Recreate each route and add it to the current router when using `router.include_router()` instead of mounting, this is the current behavior.

### Problems

* Doesn't support custom classes for routers and routes, as it creates new ones
* Doesn't support router dependencies, each dependency in the router is actually just added to the route dependencies
* Doesn't support router middleware (more precisely, doesn't support adding this as a feature), as the router itself is never called, the routes are called directly

## Mount each router

Mount the router, add an attribute to each route with the `.router` it belongs to. And add an attribute to each router with the `.parent_router`. That would allow finding the the parent dependencies and other metadata for each route.

### Problems

* This doesn't support including a router more than once as the route is strongly attached to one single chain of routers, so including it multiple times would override any other parent router's dependencies.
* It doesn't support multiple routers sharing the same path (prefix).
* It doesn't support routers that handle things at the root without trailing slash. E.g. `APIRouter(prefix="/items")` can't have a route `/items`, it can only have `/items/`.
* Extracting path parameters set in the prefix for OpenAPI could be cumbersome or not properly supported.

## Call dependencies in router

One of the objectives is to be able to have dependencies per router that could be run for anything within that router's path prefix, even if there's no matching route, e.g. for logging. An option would be to call them in the router, assuming the router is mounted and called as an ASGI app (contrary to the current behavior of only copying the router's routes).

### Problems

By default this wouldn't support extracting body parameters from those dependencies to add them to a single Pydantic body per router.

## Mount router, copy router and routes, add default router

Mount the router with standard Starlette mount, this way each router could have it's own middleware and dependencies and execute them.

But the dependencies are needed per route to extract any additional data for pydantic fields/models per route, so it would still add the dependencies to each route, and create a default route to run dependencies and middleware when there's no match.

Each router has a `.parent_router` attribute pointing to the parent router, also a method `.setup()` that reads the information that was set and the parent router's information to set its internal attributes. Then it also has a `.copy()` method that creates a new instance of the same class with the same attributes as the current router.

The copy allows setting a new parent router in the new instance and calling `.setup()` again, so that the new router can have the new parent dependencies, metadata, etc. but without overriding the old router.

Each route would also have a `.router` attribute, `.setup()` and `.copy()` method.

### Problems

* By default this wouldn't support routers with a prefix and a root route without trailing slash.
* It doesn't support multiple routers sharing the same path (prefix).

### Mount router, copy routers/routes, custom mount

This is everything above plus an internal new custom `APIMount` extending Starlette's one, with a bunch of extra logic to support root paths under a prefix without a trailing slash.

**This, plus everything above, is mostly what this PR implements**.

### Problems

* It still doesn't support redirecting in a router with a prefix ending without a trailing slash, with a root path without slash (e.g. prefix `"/items"` and path `""` to handle `/items`) and redirecting the path with trailing slash to non-trailing slash: `/items/` -> `/items`.
* It still doesn't support including multiple routers under the same prefix, once a router takes a request for a prefix there's no chance for other routers to handle it.

## Details of this PR

The way the current PR implements all this is by having the extra attributes `.router` (for routes) `.parent_router` (for rotuers) and the methods `.setup()`, `.copy()`, and having several new attributes like `._router_dependencies` to store the dependencies added on instantiation, apart from the ones including the dependencies of the parent router an the chain of parent routers that are set on the normal `.dependencies` attribute, and those are now set when calling `.setup()`.

This allows calling `.copy()` on a router or route to get a new instance with all the same attributes, and then set the new parent router, and then call `.setup()` again to set any new dependencies, metadata, etc.

### Details and Problems

This PR is too big, the code is too complex to my taste. There's too much logic setting the private attributes to then recompute the final ones with `.setup()`.

And because `.setup()` is called every time a router is included, in the end it will be called many times, although this is only on instantiation, at startup.

The whole point of storing private attributes and then setting up the rest right after/during the instantiation is to have those values cached instead of computing them live. But this means there's a lof of "caches" in attributes/memory in the code, and that's always cumbersome to handle. One of the main points of this refactor is to better support subclasses of routers and routes, but all the extra logic to store private attributes and then `.setup()` on instantiation or after including them in other routers and the logic to `.copy()` routers and routes, and then setting the new `.router` or `.parent_router` would make it more delicate and complex to implement those subclasses.

This PR also has a lot of custom code to support the corner cases. For example, there's a custom `APIMount` with quite some logic duplicated from Starlette just to support routers with a prefix and a root route without a prefix (e.g. `APIRouter(prefix="/items")` and a route inside that handles `/items`, instead of `/items/`).

## Questions

There's one existing feature that conflicts with a desired one (maybe more). Currently it is possible to include multiple routers under the same path (e.g. one with more restricted dependencies). But one of the objectives is to support middleware and dependencies per router that are executed even when there's no matching route (e.g. for logging, CORS). In this case, imagine there's:

```Python
protected_router = APIRouter(prefix="/items", dependencies=[Depends(new_auth)])
simple_router = APIRouter(prefix="/items")
```

Here we would probably want the routes in both routers to be included (as is currently the case).

But if there's a request for `/items` that doesn't match any path, should the dependency `new_auth` be executed?

If both routers define different CORS middlewares, which one should be applied?

---

I'm also wondering if I can make the implementation simpler, probably without all the private attributes and copies, maybe computing the data when needed is not that expensive or maybe I can do all that in a simpler and safer way, I'm gonna try that now.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -137,6 +137,10 @@ class FastAPI(Starlette):
         self.middleware_stack: ASGIApp = self.build_middleware_stack()
         self.setup()
 
+    @property
+    def routes(self) -> List[BaseRoute]:
+        return list(self.router.iter_all_routes())
+
     def build_middleware_stack(self) -> ASGIApp:
         # Duplicate/override from Starlette to add AsyncExitStackMiddleware
         # inside of ExceptionMiddleware, inside of custom user middlewares
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -152,7 +152,7 @@ def generate_operation_id(
     )
     if route.operation_id:
         return route.operation_id
-    path: str = route.path_format
+    path: str = route._route_full_path_format
     return generate_operation_id_for_path(name=route.name, path=path, method=method)
 
 
@@ -243,7 +243,7 @@ def get_openapi_path(
                             model_name_map=model_name_map,
                             operation_ids=operation_ids,
                         )
-                        callbacks[callback.name] = {callback.path: cb_path}
+                        callbacks[callback.name] = {callback._route_full_path: cb_path}
                 operation["callbacks"] = callbacks
             if route.status_code is not None:
                 status_code = str(route.status_code)
@@ -422,7 +422,7 @@ def get_openapi(
             if result:
                 path, security_schemes, path_definitions = result
                 if path:
-                    paths.setdefault(route.path_format, {}).update(path)
+                    paths.setdefault(route._route_full_path_format, {}).update(path)
                 if security_schemes:
                     components.setdefault("securitySchemes", {}).update(
                         security_schemes
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -311,6 +311,8 @@ class APIWebSocketRoute(routing.WebSocketRoute):
 
 
 class APIRoute(routing.Route):
+    _route_full_path_format: str  # only for mypy
+
     def __init__(
         self,
         path: str,
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -9,12 +9,14 @@ from typing import (
     Callable,
     Coroutine,
     Dict,
+    Iterator,
     List,
     Optional,
     Sequence,
     Set,
     Tuple,
     Type,
+    TypeVar,
     Union,
 )
 
@@ -56,6 +58,10 @@ from starlette.routing import (
 from starlette.status import WS_1008_POLICY_VIOLATION
 from starlette.types import ASGIApp, Scope
 from starlette.websockets import WebSocket
+
+APIRouteType = TypeVar("APIRouteType", bound="APIRoute")
+APIRouterType = TypeVar("APIRouterType", bound="APIRouter")
+APIMountType = TypeVar("APIMountType", bound="APIMount")
 
 
 def _prepare_response_content(
@@ -338,13 +344,13 @@ class APIRoute(routing.Route):
         generate_unique_id_function: Union[
             Callable[["APIRoute"], str], DefaultPlaceholder
         ] = Default(generate_unique_id),
+        router: Optional["APIRouter"] = None,
     ) -> None:
         self.path = path
         self.endpoint = endpoint
         self.response_model = response_model
         self.summary = summary
         self.response_description = response_description
-        self.deprecated = deprecated
         self.operation_id = operation_id
         self.response_model_include = response_model_include
         self.response_model_exclude = response_model_exclude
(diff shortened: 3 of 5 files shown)


ddanier:
I like this very much! 👍

Could this also reduce the impact the performance issue `created_cloned_field` imposes when using a lot of router includes? See https://github.com/tiangolo/fastapi/issues/4644

Onesco:
I love the detailed documentation of the PR.

Kludex:
> It would allow adding middleware per router, which currently isn't supported, not even by Starlette.

This was approved by Tom: https://github.com/encode/starlette/pull/1286. I didn't think we should follow that path, so I'm inclined to https://github.com/encode/starlette/pull/1464.

tiangolo:
Thanks Kludex! :rocket:

adriangb:
This is amazing, thank you for detailing your thought process @tiangolo !

One thing I'm thinking about is: can we push state management up to the root level (maybe the `FastAPI` class)?
We could push things like `dependency_overrides_provider` and the `AsyncExitStack` into an ASGI extension (`scope["extensions"]["fastapi"]["dependency_overrides_provider"]`).
This can then be managed by the `FastAPI` application.
I think this would help because routes would not have to modified when they're added to a router / app.
To get full decoupling of routes from where they are used though, we'd also need dependencies added to routers to execute with the router itself instead of executing when the route is called (which makes the route depend on where it was used instead of being self contained).

With regards to building the Pydantic model: it's not hard to introspect into routes and traverse the routing tree with a visitor-like pattern.
So you could do this once, at startup (e.g. triggered by the lifespan), and then store some sort of prepared thing (a Pydantic field I guess) at the application level. That's kind of what I'm doing in Xpresso: https://github.com/adriangb/xpresso/blob/b15dcc18b739a600523a47b69952fef035bdc2d6/xpresso/applications.py#L234-L284 (the implementation could use some cleanup though).

There's still some issues with this pattern though, namely:
1. Interaction of state management with ASIG lifespans and requests, particularly thorny if tests re-use the same app/route objects (see https://github.com/django/asgiref/issues/322)
2. Identifying a route that was used in multiple places. That is, a route can't just do `scope["extensions"]["fastapi"]["prepared_routes"][self]` because `self` would not be a unique key if it was used in multiple places _unless router dependencies were executed in the routers themselves_, in which case this works perfectly. Alternatively, each router could record itself into some `scope["extensions"]["fastapi"]["prepared_routes"]["traversed_routers"]` and then the lookup would be `scope["extensions"]["fastapi"]["prepared_routes"][(scope["extensions"]["fastapi"]["prepared_routes"]["traversed_routers"], self)]` but that's pretty ugly.

michaeloliverx:
Semi-related https://github.com/tiangolo/fastapi/issues/5343.

Calling `include_router` does a bunch of repeated work. It would be nice if there was a way to skip doing this work for the top most router.

Kludex:
https://github.com/encode/starlette/pull/1649 merged in Starlette, jfyk

JasperSui:
Looking forward to this refactor since it's a huge performance improvement for those projects using nested routers.

nidhishgajjar:

## Summary
This PR refactors FastAPI's `include_router` functionality to properly support mounting sub-routers with hierarchical path resolution and attribute inheritance. The implementation introduces a new attribute resolution pattern with `_route_*` and `_router_*` prefixes, adds `setup()` methods for deferred attribute computation, supports parent/child router relationships, and creates a new `APIMount` class. The refactor involves significant changes (606 additions, 159 deletions) across 5 files, fundamentally improving how routers compose and inherit attributes.

## Architecture
The PR introduces a comprehensive router composition architecture:

- **Attribute Resolution Pattern**: Routes and routers store local settings in `_route_*` and `_router_*` attributes, resolved in `setup()` methods
- **Hierarchical Router Support**: `parent_router` parameter enables nested router relationships
- **Path Computation**: Full path resolution for nested routers via `_route_full_path`
- **Attribute Merging**: Intelligent inheritance of dependencies, tags, responses, callbacks, etc.
- **Deferred Setup**: `setup()` methods enable proper attribute resolution after router hierarchy is established
- **Mount Points**: New `APIMount` class handles router mounting with proper path and attribute resolution
- **Backward Compatibility**: Maintains existing API while enabling advanced composition patterns

The refactor transforms FastAPI's router system from a flat composition model to a hierarchical one with proper attribute inheritance and resolution.

## Issues

### Critical

**No critical issues found** - The architecture is sound and the implementation is comprehensive.

### Warnings

1. **fastapi/routing.py:347** - **Complex initialization logic**: The `__init__` method now relies on `setup()` being called, which could lead to incomplete object state if `setup()` fails:

```python
def __init__(self, ...):
    # ... set _route_* attributes ...
    self.setup()  # If this fails, object is in incomplete state
```

   Consider using a factory pattern or two-phase initialization with explicit setup call.

2. **fastapi/routing.py:618** - **Similar issue in APIRouter**: 

```python
def __init__(self, ...):
    # ... set _router_* attributes ...
    self.setup()  # Setup in constructor
```

3. **fastapi/routing.py:1575** - **New APIMount class complexity**: The new class adds significant complexity:

```python
class APIMount(routing.Mount):
    def __init__(self, router: APIRouter, *, name: Optional[str] = None, parent_router: Optional[APIRouter] = None):
        self.router = router
        self.setup()  # Again, setup in constructor
```

### Suggestions

1. **fastapi/routing.py:360-430** - **Extract attribute resolution logic**: The `setup()` method in `APIRoute` is complex and could benefit from extraction:

```python
def _setup_dependencies(self) -> None:
    self.dependencies: List[params.Depends] = []
    if self.router:
        self.dependencies.extend(self.router.dependencies)
    self.dependencies.extend(self._route_dependencies)

def _setup_tags(self) -> None:
    self.tags: List[Union[str, Enum]] = []
    if self.router:
        self.tags.extend(self.router.tags)
    self.tags.extend(self._route_tags)

# Then in setup():
def setup(self) -> None:
    self._setup_full_path()
    self._setup_dependencies()
    self._setup_tags()
    # ... etc
```

2. **Missing validation**: Add validation for router hierarchy to prevent circular dependencies:

```python
def setup(self) -> None:
    if self.parent_router:
        # Check for circular dependencies
        current = self.parent_router
        while current:
            if current is self:
                raise ValueError("Circular router dependency detected")
            current = current.parent_router
    # ... rest of setup
```

3. **Performance considerations**: The `setup()` method is called frequently during router construction. Consider caching or memoization where appropriate:

```python
def setup(self) -> None:
    if self._setup_complete:
        return
    # ... setup logic ...
    self._setup_complete = True
```

4. **Documentation needs**: This significant architectural change requires comprehensive documentation:
   - How router hierarchy works
- Attribute inheritance rules
- When to use `include_router` vs direct mounting
- Migration guide for existing code
- Performance implications

5. **Test coverage gaps**: Consider adding tests for:
   - Deep router hierarchies (3+ levels)
- Circular dependency detection
- Attribute precedence conflicts
- Path resolution edge cases
- Performance with many nested routers

6. **Error handling improvements**: Add better error messages for setup failures:

```python
def setup(self) -> None:
    try:
        self._setup_full_path()
    except Exception as e:
        raise RuntimeError(f"Failed to setup full path for route {self.path}: {e}") from e
```

7. **Type safety improvements**: Some attributes could have more precise types:

```python
# Instead of:
self.dependencies: List[params.Depends] = []

# Consider:
self._resolved_dependencies: List[params.Depends] = []
```

8. **fastapi/routing.py:932-950** - **Complex include_router logic**: The `include_router` method has become quite complex. Consider extracting sub-methods:

```python
def _create_wrapping_router(self, router: "APIRouter", prefix: str, **kwargs) -> "APIRouter":
    """Create a wrapper router when needed for attribute override"""
    # ... existing logic ...

def _should_copy_flat_routes(self, prefix: str, copy_flat_routes: Optional[bool]) -> bool:
    """Determine if routes should be copied flat"""
    if copy_flat_routes is not None:
        return copy_flat_routes
    return not (prefix or self.prefix)
```

## Cross-file Impact

- **Breaking changes**: None (maintains backward compatibility)
- **Public API**: No breaking changes to public API
- **Internal architecture**: Significant refactoring of router system
- **Performance**: Potential performance impact from setup() calls (needs benchmarking)
- **Extensibility**: Greatly improved router composition capabilities
- **Maintenance**: Increased complexity requires good documentation and tests
- **Dependencies**: No new dependencies

## Assessment
✅ Approve

This is a well-executed architectural refactor that significantly improves FastAPI's router composition capabilities. The implementation introduces a sophisticated attribute resolution pattern that enables proper hierarchical router support while maintaining backward compatibility.

**Strengths:**
- Comprehensive architectural improvement
- Maintains backward compatibility
- Enables advanced router composition patterns
- Clean separation of local vs. resolved attributes
- Proper path resolution for nested routers
- Intelligent attribute merging and inheritance
- Follows FastAPI's existing patterns and conventions

**Areas for improvement:**
- Complexity of setup() methods could be reduced through extraction
- Need comprehensive documentation for the new architecture
- Performance characteristics should be benchmarked
- Test coverage could be expanded for edge cases
- Error handling could be more descriptive
- Validation for circular dependencies would be valuable

**Architecture considerations:**
- The two-phase initialization (construct + setup) is a reasonable pattern but should be well-documented
- The attribute resolution pattern is powerful but adds complexity that needs to be managed through good documentation
- The new APIMount class adds another layer of abstraction that users need to understand

This refactor represents a significant evolution of FastAPI's router system and will enable more complex application architectures. The implementation is solid and the architectural decisions are sound. The suggestions above are primarily about managing the added complexity through better documentation, testing, and code organization rather than fundamental issues with the approach.

The PR is ready to merge and will provide substantial benefits to users building complex FastAPI applications with nested router structures.

tiangolo:
Superseded by https://github.com/fastapi/fastapi/pull/15745 :tada: (finally, it's happeningggg!)

Issue #4402, opened 2022-01-09
Cannot use a pydantic `dataclass` nested in a `BaseModel` as a `model_response`


ypsah:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from fastapi import FastAPI
from pydantic import BaseModel
from pydantic.dataclasses import dataclass


app = FastAPI()


@dataclass
class Item:
    name: str


class Box(BaseModel):
    item: Item


@app.get("/item", response_model=Item)
async def item() -> Item:
    return Item(name="test")


@app.get("/box", response_model=Box)
async def box() -> Box:
    return Box(item=await item())
```

### Description

The `item` endpoint works fine, but the `box` one fails with a `ValidationError`:

```console
$ uvicorn main:app
INFO:     Started server process [13673]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     127.0.0.1:46020 - "GET /docs HTTP/1.1" 200 OK
INFO:     127.0.0.1:46020 - "GET /openapi.json HTTP/1.1" 200 OK
INFO:     127.0.0.1:46020 - "GET /item HTTP/1.1" 200 OK
INFO:     127.0.0.1:46020 - "GET /box HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/guest/.local/share/virtualenvs/tmp.KhplXQz0Am-XrtSmM4b/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 373, in run_asgi
    result = await app(self.scope, self.receive, self.send)
[...]

PR #4829, opened 2022-04-26
4402

databasedav:
addresses #4402 and adds `dict_factory` pass through for `dataclasses.asdict`

the alternative to this was to recursively iterate through the response dict to check for dataclasses **by default**, so since this part of the code manages all responses, i opted to be conservative and make the recursive dataclass reconcilation opt in with a new argument to the route decorator signatures

- [x] tests
- [ ] docs


databasedav:
this should actually be fixed at the pydantic level https://github.com/samuelcolvin/pydantic/issues/3764

Issue #3745, opened 2021-08-21
Generated schemas do not follow the OpenAPI specification (FastAPI v0.66+)


ml-evs:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import json
from pydantic import BaseModel, Field
from fastapi import FastAPI
from fastapi import __version__ as fastapi_version

class Model(BaseModel):
    field: str = Field("foo", const=True)


app = FastAPI(title=f"Created with FastAPI v{fastapi_version}")


@app.get("/", response_model=Model)
def root():
    return Model(field="foo")

print(json.dumps(Model.schema(), indent=2))
print(json.dumps(app.openapi(), indent=2))
```

### Description

With `fastapi==0.68.0` and `pydantic==1.8.2`, the above MWE produces an invalid OpenAPI schema due to the handling of extra keys introduced in the last release. The inclusion of the `const` key is not supported in OpenAPI 3 (see [Supported Keywords](https://swagger.io/docs/specification/data-models/keywords/)).

Here is a diff between the schema created for v0.65.2 vs v0.68 (also tested with 0.66, 0.67 with the same results).

```diff
// ...
  },
  "components": {
    "schemas": {
      "Model": {
        "title": "Model",
        "type": "object",
        "properties": {
          "field": {
            "title": "Field",
            "type": "string",
+           "const": "foo"
          }
// ...
```

This regression is caused by the addition of `extra = "allow"` to `fastapi.openapi.models.Schema` in (#1429) (see also the comment https://github.com/tiangolo/fastapi/pull/1429#issuecomment-889256569). This is something we ran into when updating FastAPI at https://github.com/Materials-Consortia/optimade-python-tools/pull/887.

[...]

PR #4846, opened 2022-05-01
📝 Add OpenAPI warning to "Body - Fields" docs with extra schema extensions

ml-evs:
This PR adds a warning in the docs about generating invalid OpenAPI schemas (see #3745) when using additional keys in pydantic `Field` specifications. This warning only effects versions `fastapi>=0.66`, which could be added to the note if helpful.

I would have liked to include a workaround for those who want to keep generating valid schemas, but is currently non-trivial and either involves overriding a whole chain of field info classes, or writing a custom OpenAPI schema method, to undo the effect of `extra = "allow"` in the `fastapi.openapi.models.Schema` class.
It is a bit of a shame this hasn't been acknowledged as an issue in #3745 but I hope that this warning will be helpful for those who are trying to use generic OpenAPI tools with FastAPI.

### DIFF
--- a/docs/en/docs/tutorial/body-fields.md
+++ b/docs/en/docs/tutorial/body-fields.md
@@ -55,6 +55,11 @@ You can then use `Field` with model attributes:
 
 You can declare extra information in `Field`, `Query`, `Body`, etc. And it will be included in the generated JSON Schema.
 
+!!! warning
+    Extra keys passed to `Field` will also be present in the resulting OpenAPI schema for your application.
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks @ml-evs! I updated the tip/warning a bit to avoid mentioning `const` as that would mean it would also have to explain how `const` is valid JSON Schema v6 but not valid OpenAPI 3.0.x, although as OpenAPI 3.1.x is based on JSON Schema v7 then it is compatible there.

PR #4851, opened 2022-05-02
Generate unique operation id with multiple methods (Fixes #4740)

### LINKED ISSUE #4740: Duplicated OperationID when adding route with multiple methods
### First Check

### Commit to Help

### Example Code

```python
router.add_api_route(
    "/clear",
    clear,
    methods=["POST", "DELETE"]
)
```

### Description

Since to be caused by #4650.

The new `generate_unique_id()` function uses `list(route.methods)[0].lower()` as suffix for the `operation_id`. Therefore, in my example, both post and delete endpoints get `_post` suffix for operation_id, causing it to no longer be unique.

It then issues a "UserWarning: Duplicate Operation ID"

### Operating System

Windows

### Operating System Details

_No response_

### FastAPI Version

0.75.0

### Python Version

3.10.2

### Additional Context

_No response_

bruchar1:
When providing multiple methods on a single route, the same operation id is generated on routes for each methods, using the first method in the list.

This PR adds a different route for each method, ensuring the proper method name is used for generating the operation id.


phillipuniverse:
@bruchar1 I pulled this in locally to work with an `@router.api_route` decorator and it works perfectly! Thanks for the fix!

tiangolo:
Thanks for the interest @bruchar1!

Using multiple methods for the same function is not actually supported. `@router.api_route()` is there as a semi-internal utility, and because Starlette supports multiple methods, the interface is there. But the official way to do it in FastAPI and have all the FastAPI features (including the automatic OpenAPI integration, with operation IDs) is by using the dedicated methods, like `app.post()`, `app.get()`, etc.

Given that I'll pass on this one, but thanks for the effort! :cake:

Ark-kun:
@tiangolo This bugs prevents us from creating a method that can work with arbitrary `fastapi.Request` requests. I've seen `@api_route` suggested in this repo when you need to work with generic Requests (e.g. you want to create a proxy).

PR #4869, opened 2022-05-06
📝 Add dark mode auto switch to docs based on OS preference

ComicShrimp:
Implement the [example](https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#system-preference) from Material MKdocs that make the docs auto switch from light to dark mode based on system preference.


tiangolo:
Great, thanks! :memo: :bat:

PR #4871, opened 2022-05-07
✨Add support for PEP-593 `Annotated` for specifying dependencies and parameters

### LINKED ISSUE #3323: Support PEP 593 `Annotated` for specifying dependencies and parameters
### First check

### Example

I propse to allow transforming:

```Python
from typing import Optional

app = FastAPI()


async def common_parameters(q: Optional[str] = None, skip: int = 0, limit: int = 100):
    return {"q": q, "skip": skip, "limit": limit}


@app.get("/items/")
async def read_items(commons: dict = Depends(common_parameters)):
    return commons
```
to 
```Python
from typing import Annotated, Optional

app = FastAPI()


async def common_parameters(q: Optional[str] = None, skip: int = 0, limit: int = 100):
    return {"q": q, "skip": skip, "limit": limit}


@app.get("/items/")
async def read_items(commons: Annotated[dict, Depends(common_parameters)]):
    return commons
```

### Discussion
[PEP 593](https://www.python.org/dev/peps/pep-0593/) Added `Annotated` for adding additional annotations beyond type annotations. I think FastAPI's `Depends`, `Query`, `Body` and the likes fit well with the kind of additional annotations this supports.

This would also make default values less awkward:
```python
@app.get("/items/")
async def read_items(q: Optional[str] = Query(None, max_length=50)):
    pass
```
Could become
```python
@app.get("/items/")
async def read_items(q: Annotated[Optional[str], Query(max_length=50)] = None):
    pass
```

This will also solve the issue mentioned [in the docs](https://fastapi.tiangolo.com/tutorial/path-params-numeric-validations/#order-the-parameters-as-you-need) of parameter ordering.

Finally, it is sometimes convenient to use the same function as both a FastAPI dependency and a regular function. In these cases, because `= Depends(...)` is a default parameter value, if you forget to pass a parameter the error is not caught by your IDE. Worse, it is not caught at runtime because Python will just pass along the `Depends` object. This will probably cause an error down the road, but may silently succeed in some cases.

I'm willing to implement this if you think it's a good idea.

nzig:
# What this does
This PR adds support for PEP-593 `Annotated` for specifying dependencies and parameters.
Fixes #3323.

This would allow you to write
```python
async def handler(dep: Annotated[MyClass, Depends(my_dependency)]):
    ...
```
instead of 
```python
async def handler(dep: MyClass = Depends(my_dependency)):
    ...
```

It would also allow you to define
```python
MyDependecy: TypeAlias = Annotated[MyClass, Depends(my_dependency)]
```

so that the handler could be written simply as
```python
async def handler(dep: MyDependecy):
    ...
```

Which saves you from having to write both the dependency and the type annotation, all while ensuring that your IDE and type-checker understand the type.

This will also solve the issue mentioned [in the docs](https://fastapi.tiangolo.com/tutorial/path-params-numeric-validations/#order-the-parameters-as-you-need) of parameter ordering. 

I am also quite fond of dual-use handlers, where the same function is used both as a FastAPI handler and also as a regular callable function. This currently works okay, except that because `Depends` is a defualt value,
you don't get an error if you forget to pass a parameter. Using `Annotated` avoids this problem as well.

## Other changes
This change required changing some of how FastAPI handles parameters.

- ~~`default` is no longer required for `Param`. This is because I think it should be possible to write a param as `my_header: Annotated[str, Header()]`  if it required, and `my_header: Annotated[str, Header()] = ''` for a param with a default value.~~ Update: This was already done by @tiangolo in #4906.
   - When used with `Annotated`, the default value may not be specified as `my_header: Annotated[str, Header('default')]` but only as `my_header: Annotated[str, Header()] = 'default'`.
   - ~~When used without `Annotated`, we check at runtime that `default` is set. This is like the previous behavior, except that it is now checked by FastAPI rather than Python erroring on a missing required parameter.~~
- Ambiguous annotations are now an error. Specifying something like `r: Request = Header(...)` would be ambiguous, in this case choosing to use the `Request` annotation and igonoring the `Header` annotation. Now this will be an error. I think this is preferable, as the ambiguity likely indicates a mistake by the developer.
- Specifying anything other than a `Path` annotation for a path parameter name is now an error. 
- A `default` value can not be specified for an explicit `Path` annotation. It is currently ignored anyway because a path parameter is always specified when the handler is called.

# Implementation
The main change, as expected, is to stop treating `param.default` as the FastAPI annotation. Instead, `analyze_param` will analyze the param's annotation and default value, handling both `Annotated` and default value FastAPI annotations. It returns the plain `type_annotation` (with `Annotated` stripped off), a `depends` if there is a `Depends` annotation, and a `field` like `get_param_field` used to.

## Docs
I took a conservative approach to the docs, and only added this feature as an advanced guide, without introducing it in any of the earlier tutorials.

If it is wanted, this can also be added to the "Python Types Intro", and also introduced earlier in the tutorial series.


yezz123 (approved):
LGTM! This PR Looks really Clean and with the fully support for PEP-593 ✨

docs/en/docs/advanced/annotated.md
```diff
@@ -0,0 +1,79 @@
+# Other annotations
+Python 3.9 introduced the `Annotated` syntax for type annotations. This may also be used in older versions of Python by using the `Annotated` from the `typing_extensions` package.
+
+In FastAPI, this allows us to specify annotations without using default values, and allows us to avoid repeating the type annotation for a dependency.
+
+Let's look at the example from the dependecies tutorial:
```
MrRawbin, line 6: "dependecies" typo

nzig: Fixed.

NowanIlfideme:
Honestly, with modern Python, this is the thing that fixes what feels "old" with FastAPI (and also the default value problem). I wonder why it's not yet merged? Docs?

tiangolo:
You did an amazing job here @nzig, it took me several days to properly review it all with all the attention, and it's an impeccable implementation. Thanks a lot! :bow: :cake: :rocket: 

This is the biggest feature in FastAPI in a long while. :sunglasses:

Also, you took the right approach by being conservative about the docs, that was perfect.

...in fact, I think `Annotated` is the **right approach**, and I'm making it the default recommendation from now on.

I'm rewriting ALL the documentation to use it, with examples for all Python versions, new docs in the respective sections, new tests, etc. That's why I removed the `Annotated`-specific docs in this PR, but again, that was the right approach.

I actually wanted to do this for a long while (even before I noticed your PR for the first time 😅) but I needed all the consecutive time dedicated full-time to this to get it done (I've been re-writing docs for 2 weeks 😂).

Here's the PR adding the new docs: https://github.com/tiangolo/fastapi/pull/9268

This will be available in the next release, in some hours. FastAPI `0.95.0`. :tada: 

---

Thanks for the reviews @MrRawbin and @yezz123 ! :coffee: :cake:

nzig:
Thanks @tiangolo ! I'm glad I'll be able to use this in FastAPI 🥳 .

JoelPendleton:
Hi guys, I experienced difficulties when using this new `Annotated` specification when following this section of the [docs](https://fastapi.tiangolo.com/tutorial/security/simple-oauth2/#__tabbed_5_3). In particular, I get the following error when using  `form_data: Annotated[OAuth2PasswordRequestForm, Depends()]`: ` raise fastapi.exceptions.FastAPIError(
fastapi.exceptions.FastAPIError: Invalid args for response field! Hint: check that typing.Annotated[fastapi.security.oauth2.OAuth2PasswordRequestForm, Depends(NoneType)] is a valid Pydantic field type. If you are using a return type annotation that is not a valid Pydantic field (e.g. Union[Response, dict, None]) you can disable generating the response model from the type annotation with the path operation decorator parameter response_model=None. Read more: https://fastapi.tiangolo.com/tutorial/response-model/`. Moreover, when checking my endpoints at http://127.0.0.1:8000/docs the padlock is not present on the "/users/me" endpoint and it treats the current_user argument as an input and not a dependency. I hope you can resolve this :)

srdjus:
@JoelPendleton 

I got the same error, Python 3.10.0,  FastAPI 0.94.1.

phillipuniverse:
@tiangolo @nzig just FYI this change had the unfortunate side effect of union response bodies no longer working in 0.95, see https://github.com/tiangolo/fastapi/discussions/9287#discussioncomment-5532496

9en9i:
@nzig 

Hi, i tried using 
```
T = TypeVar("T")

Inject = Annotated[T, Depends()]
``` 
and it doesn't seem to work. Should we expect it to ever work?

nzig:
@9en9i What are you expecting to happen? A more complete example might be helpful.
It's probably better to open it as a [discussion](https://github.com/tiangolo/fastapi/discussions) instead of discussing it on the PR.

9en9i:
@nzig I created a discussion #9450  and added a code example with a description of the problem there. I would be grateful if you could take a look.

PR #4887, opened 2022-05-11
🍻 Delete lots of tests to check if coverage badge points to broken PRs

tiangolo:
🍻 Delete lots of tests to check if coverage badge points to broken PRs


tiangolo:
It seems badges still point to `master` with 100% coverage and not to the latest PR, so we are good! Closing this now.

PR #4904, opened 2022-05-13
Add Spanish translation for mkdocs

JonanOribe:
Just something that was untranslated

### DIFF
--- a/docs/es/mkdocs.yml
+++ b/docs/es/mkdocs.yml
@@ -37,7 +37,7 @@ plugins:
     data: data
 nav:
 - FastAPI: index.md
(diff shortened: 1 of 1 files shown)


arnaldojvg (approved):
👍

tiangolo:
Thanks! But I prefer to keep that one in English, for all the languages, because if someone arrives at the documentation for another language they can probably see that in English and find the translation to their own language more easily.

Also, because of that, the scripts that generate the docs for each language also depend on this specific key.

Given that, I'll close this PR now, but thanks for the interest! :coffee:

PR #4905, opened 2022-05-13
📝 Update docs, compare enums with identity instead of equality

MicaelJarniac:
> Enumeration members are compared by identity
https://docs.python.org/3/library/enum.html#comparisons


yezz123 (approved):

jkklapp (approved):

tiangolo:
Cool, thanks! :nerd_face:

PR #4920, opened 2022-05-17
fixed missing argument in header

nycker:
- [x] Fixed TypeError: Header() missing 1 required positional argument: 'default'
- [x] Added request as paramerer in read_items so that Headers can be read and processed


tiangolo:
It seems you were using an older version of FastAPI. In recent versions the first parameter, `default`, is no longer required. If it's not passed, FastAPI will infer that the parameter is required.

Given that, I'll close this one, but thanks for the interest! :coffee:

PR #4933, opened 2022-05-20
🔧 Fix Type hint of `auto_error` which does not need to be `Optional[bool]`

DavidKimDY:
Default is True and If someone doesn't want auto_error option, then it should be False instead of None.  
I guess It is more Pythonic.


yezz123 (approved):

tiangolo:
Good catch, thanks @DavidKimDY! :rocket: :sunglasses:

PR #4940, opened 2022-05-23
Add possibility to limit doc description with `\f`

fxmarty:
It is already possible to limit the description of endpoints using `\f` in docstrings, see https://github.com/tiangolo/fastapi/issues/1358 and https://fastapi.tiangolo.com/advanced/path-operation-advanced-configuration/#advanced-description-from-docstring .

This PR adds the possibility as well in endpoints arguments description (typically if we want to not show part of a pydantic BaseModel class docstring).


tiangolo:
Thanks @fxmarty! 🍰 

Do you have a use case that needs this? Right now I can't think of a reason why it would be useful to limit the content of a field.

Let's start with a GitHub discussion question, please fill in the template, that way I can understand better your use case and see how this would solve it.

For now, I'll close this while we figure out the use case there, and then I can re-open it later. ☕

PR #4968, opened 2022-05-31
Customizable Request Handler in APIRoute

sihrc:
End goal was to provide customizable request decoding and response encoding and to forego `jsonable_encoder` in favor of one handled in my own custom serialization lib. 

This PR refactors the get_request_handler logic into an overridable class that can be passed into APIRoute which allows more options for modifying steps of request/response handling.

Typically flooded by github notifs, but I can by reached at the email on my profile if I forget to check for any updates here.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -146,7 +146,7 @@ class RequestHandler:
         else:
             self.actual_response_class = self.response_class
 
-    def get_decoder(self, request: Request) -> Optional[Callable[[bytes], Any]]:
+    def get_body_decoder(self, request: Request) -> Optional[Callable[[bytes], Any]]:
         content_type_value = request.headers.get("content-type")
         if not content_type_value:
             return json_loads
@@ -193,6 +193,13 @@ class RequestHandler:
         else:
             return jsonable_encoder(response_content)
 
+    async def create_response(
+        self, request: Request, raw_response: Any, **response_args: Any
+    ) -> Response:
+        response_data: Any = await self.encode_raw_response(request, raw_response)
+        response: Response = self.actual_response_class(response_data, **response_args)
+        return response
+
     async def __call__(self, request: Request) -> Response:
         body: Any = None
         try:
@@ -201,7 +208,7 @@ class RequestHandler:
             elif self.body_field:
                 body_bytes: bytes = await request.body()
                 if body_bytes:
-                    decoder: Optional[Callable[[bytes], Any]] = self.get_decoder(
+                    decoder: Optional[Callable[[bytes], Any]] = self.get_body_decoder(
                         request
                     )
                     if decoder:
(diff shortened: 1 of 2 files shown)


sihrc:
Realized I can run `./scripts/lint.sh` locally. Brb

Andrew-Chen-Wang:
+1 [I wanted to make a custom Request and WebSocket](https://stackoverflow.com/questions/75278075/fastapi-custom-websocket-object) class for faster attribute access and methods.

Ended up just monkeypatching the modules:

```
from typing import Callable

from fastapi import routing as fastapi_routing
from starlette._utils import is_async_callable
from starlette.concurrency import run_in_threadpool
from starlette.requests import Request as StarletteRequest
from starlette.websockets import WebSocket as StarletteWebSocket
from starlette.types import ASGIApp, Receive, Scope, Send

class Request(StarletteRequest):
    pass

class WebSocket(StarletteWebSocket):
    pass

def request_response(func: Callable) -> ASGIApp:
    """
    Takes a function or coroutine `func(request) -> response`,
    and returns an ASGI application.
    """
    is_coroutine = is_async_callable(func)

    async def app(scope: Scope, receive: Receive, send: Send) -> None:
        request = Request(scope, receive=receive, send=send)
        # Force all views to be a coroutine
        response = await func(request)
        if is_coroutine:
            response = await func(request)
        else:
            response = await run_in_threadpool(func, request)
        await response(scope, receive, send)

    return app


fastapi_routing.request_response = request_response

def websocket_session(func: Callable) -> ASGIApp:
    """
    Takes a coroutine `func(session)`, and returns an ASGI application.
    """
    # assert asyncio.iscoroutinefunction(func), "WebSocket endpoints must be async"

    async def app(scope: Scope, receive: Receive, send: Send) -> None:
        session = WebSocket(scope, receive=receive, send=send)
        await func(session)

    return app


fastapi_routing.websocket_session = websocket_session
```

fastapi/routing.py
```diff
@@ -447,13 +473,27 @@ def get_route_handler(self) -> Callable[[Request], Coroutine[Any, Any, Response]
             dependency_overrides_provider=self.dependency_overrides_provider,
         )
 
+        # Awaiting https://github.com/encode/starlette/pull/1444/files to remove this workaround
```
estan, line 476: I think that is merged and in starlette >= 0.27 which FastAPI master depends on, so this could be removed?

estan (commented):
I think the new `RequestHandler` class and the relevant methods now needs some docs, since it's designed to be subclassed by FastAPI users. Perhaps even a section in the Advanced User Guide?

But before that, maybe tag @tiangolo to comment on if he's in favour of this approach?

I think this would be a great addition, since so many people in https://github.com/tiangolo/fastapi/issues/1450 were interested in an official way of customizing the request/response serialization format while also avoiding FastAPI's internal JSON serialization. @sihrc if this fixes that issue, perhaps reference it here in the PR description? (the issue was later turned into a Discussion).

estan:
I see this was closed. I guess because original author disappeared? Is there some other official way of customizing request/response handling so that internal JSON serialization is avoided nowadays, that we could leave as a link here?

I think original author's motivation for the feature was same as mine back then, to be able to avoid library internal serialization to JSON of request/response, e.g. when using CBOR instead, in a convenient/documented way.

PR #4972, opened 2022-05-31
🐛 Fix `jsonable_encoder` alters `json_encoders` of Pydantic v1 objects

aboubacs:
This PR removes a side effect in the `jsonable_encoder` function, which was modifying the `json_encoders` configuration of pydantic models when any instance of that model was given to the function with a custom encoder, which fixes this issue: https://github.com/tiangolo/fastapi/issues/4962
Explanation: https://github.com/fastapi/fastapi/pull/4972#discussion_r892603230

**UPD YuriiMotov:** it also attempted to fix the following issue in tests, but I reverted this change since it should be done in separate PR and in consistent way:
- It changes the custom_encoder test to change the encoding from `o.isoformat()` to `o.strftime("%H:%M:%S")`, because `isoformat` is the already the [default pydantic encoder for datetime](https://github.com/samuelcolvin/pydantic/blob/master/pydantic/json.py#L48). The test wasn't properly testing the feature because it could pass even without the custom_encoder


nurettn (approved):

odiseo0:
Clean PR

fastapi/encoders.py
```diff
@@ -52,7 +52,7 @@ def jsonable_encoder(
     if isinstance(obj, BaseModel):
         encoder = getattr(obj.__config__, "json_encoders", {})
         if custom_encoder:
-            encoder.update(custom_encoder)
+            encoder = {**encoder, **custom_encoder}
```
odiseo0, line 55: Can you explain, how does this fix the `jsonable_encoder` side effect?

odiseo0: Thank you very much, easy to understand

odiseo0 (approved):

aboubacs:
What's the next step on this ? @odiseo0 @nurettinabaci

iudeen (approved):

aboubacs:
Any plan to include this fix in a future release ? The bug seems to still be here on the master branch, it happens as soon as you use multiple jsonable_encoders on the same model

YuriiMotov (approved):
LGTM!

@aboubacs, good catch! Thank you!

There are several more occurrences of `o.isoformat()` in tests. I think it should be addressed in separate PR and in consistent way (all occurrences of the problem)

tiangolo (approved):
Nice, thank you! 🙌 ☕

PR #5001, opened 2022-06-08
Sort methods to stabalize operation id generation

bneijt:
The methods attribute accessed by the `generate_unique_id` function is modeled as a python set, which is not ordered and between runs might change. This results in the same fastapi generating different openapi specs if started multiple times.

This also relates to the bigger issue https://github.com/tiangolo/fastapi/issues/4740 but is not a solution for that issue.


JarroVGIT (approved):
This would work and indeed solves a minor bug. You haven't supplied a test case for this though, might help in getting this PR approved by Tiangolo

bneijt:
Because it's runtime random behavior, I find it hard to come up with a proper test case that will work across different python versions _and_ will make sure we are actually doing the right thing instead of possibly passing because of shear luck. Consider the following output:
```python
Python 3.8.9 (default, Dec 16 2021, 12:39:57) 
[GCC 11.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information
>>> b=set(["2","1","3"])
>>> list(b)[0]
'1'
>>> b=set(["1","2","3"])
>>> list(b)[0]
'3'
>>> b=set(["1","2","3"])
>>> list(b)[0]
'3'
>>> b=set(["1","2","3"])
>>> list(b)[0]
'3'
>>> b=set(["1","2","3"])
>>> list(b)[0]
'3'
>>> b=set(["1","2","4"])
>>> list(b)[0]
'4'
```

JarroVGIT:
Yeah I was having a stroll outside right after I made that comment, and I had the exact same thought. You are right, I also couldn't find a way to properly test it. 

That leaves me with the thought that somehow (if not through a specific test case) we would want to make it specific that there was a reason this line of code uses `sorted`. I can think of two possibilities: 
1. Use a comment line in the code
2. Create a testcase with two lists (one reversed from the other) with explicit naming and pull that through the function. This doesn't test for sets, but it does make it explicit that despite different ordering in a `Iterable`, the outcome should be the same. 

Let me know your thoughts!

bneijt:
I've decided to add a comment.

I was writing the test case and came to the conclusion that there would be no way to write the test that would still make sense if APIRoute would be refactored to accept a single http method only.

Because I think that will probably end up being the outcome of #4740 I decided a comment would be best for now. Happy to hear from official reviewers on this.

tomtom103:
@JarroVGIT any updates on this? Recently came up in one of my projects, would love to see this getting merged eventually

tiangolo:

I'm not sure I understand the use case that this solves. :thinking: 

It would probably be better if we start in a GitHub Discussion, with a minimal self contained example that I can copy paste to understand the current issue, and then we can check if this would solve it, or maybe there's another way to handle it.

For now I'll close this one, but thanks for the effort! :cake:

bneijt:
The use case for me, when I filed the issue a fee years ago, was that I was deploying an AWS gateway configuration via Terraform based on the OpenAPI spec generated by FastAPI. This would randomly change and cause a reconfiguration without actual changes to the API implementation.

Like the developers of FastAPI, I also lost interest. Thank you for closing!

PR #5008, opened 2022-06-09
Update applications.py

tsv19su254052:
corrected indents and hyphens


fastapi/applications.py
```diff
@@ -196,18 +185,16 @@ def build_middleware_stack(self) -> ASGIApp:
...
-                openapi_version=self.openapi_version,
-                description=self.description,
-                terms_of_service=self.terms_of_service,
-                contact=self.contact,
-                license_info=self.license_info,
-                routes=self.routes,
-                tags=self.openapi_tags,
-                servers=self.servers,
-            )
+            self.openapi_schema = get_openapi(title=self.title,
```
JarroVGIT: I have left a remark here: https://github.com/tiangolo/fastapi/pull/5016#issuecomment-1152278151

JarroVGIT (changes requested):
Recommendation: retract as it makes the code not PEP-8 compliant. See https://github.com/tiangolo/fastapi/pull/5016#issuecomment-1152278151

tsv19su254052:
OK. It is a C++ Style.
I just wanted to show my style. If these changes do not comply with the PEP-8 standard, then you can refuse them

MarcusVinniciuss (approved):

tiangolo:

PR #5011, opened 2022-06-09
Update encoders.py

tsv19su254052:
corrected indents and hyphens


JarroVGIT (changes requested):

tsv19su254052:
OK. It is a **C++** Style.
I just wanted to show my style. If these changes do not comply with the **PEP-8** standard, then you can refuse them

tiangolo:
Thanks for the effort @tsv19su254052! :cake: 

It seems this was included in a recent PR, so I'll close this one.

And thanks for the interest @JarroVGIT, there's no need to manually check for PEP 8, I already enforce Black formatting for all the codebase, it's even done automatically with pre-commit. :sunglasses: :tada: 

Thanks again! :coffee:

PR #5012, opened 2022-06-09
Update exception_handlers.py

tsv19su254052:
corrected indents and hyphens


JarroVGIT (changes requested):

tsv19su254052:
OK. In my opinion, this is C++ Style.
I just wanted to offer my style.
If these changes do not comply with the **PEP-8** standard, then they can be discarded

tiangolo:

PR #5016, opened 2022-06-09
Update models.py

tsv19su:
tweaked a little

### DIFF
--- a/fastapi/openapi/models.py
+++ b/fastapi/openapi/models.py
@@ -18,8 +18,10 @@ except ImportError:  # pragma: no cover
 
         @classmethod
         def validate(cls, v: Any) -> str:
-            logger.warning("email-validator not installed, email fields will be treated as str.\n" 
-                           "To install, run: pip install email-validator")
+            logger.warning(
+                "email-validator not installed, email fields will be treated as str.\n"
+                "To install, run: pip install email-validator"
+            )
             return str(v)
 
 
--- a/fastapi/openapi/models.py
+++ b/fastapi/openapi/models.py
@@ -18,10 +18,8 @@ except ImportError:  # pragma: no cover
 
         @classmethod
         def validate(cls, v: Any) -> str:
-            logger.warning(
-                "email-validator not installed, email fields will be treated as str.\n"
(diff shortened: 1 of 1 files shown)


JarroVGIT:
Could you perhaps merge all these changes together? 

https://github.com/tiangolo/fastapi/pull/5016
https://github.com/tiangolo/fastapi/pull/5013
https://github.com/tiangolo/fastapi/pull/5012
https://github.com/tiangolo/fastapi/pull/5011
https://github.com/tiangolo/fastapi/pull/5010
https://github.com/tiangolo/fastapi/pull/5009
https://github.com/tiangolo/fastapi/pull/5008

They are all style based, so if accepted they can be accepted in one go. 

On the PR's itself: they are against PEP-8 in terms of indentation. See [here](https://peps.python.org/pep-0008/#indentation) for more information.

Example of correct indentation:
```python
def long_function_name(
        var_one, var_two, var_three,
        var_four):
    print(var_one)
```

Often in your cases, you outline everything to the first param, which you keep on the same line as the function call, for example (from applications.py):

```python
            self.openapi_schema = get_openapi(title=self.title, 
                                               version=self.version, 
                                               openapi_version=self.openapi_version, 
                                               description=self.description, 
                                               terms_of_service=self.terms_of_service, 
                                               contact=self.contact, 
                                               license_info=self.license_info, 
                                               routes=self.routes, 
                                               tags=self.openapi_tags, 
                                               servers=self.servers, )
```

In this case, PEP-8 compliant style would be:
```python
            self.openapi_schema = get_openapi(
                 title=self.title,
                 version=self.version,
                 openapi_version=self.openapi_version,
                 description=self.description,
                 terms_of_service=self.terms_of_service,
                 contact=self.contact,
                 license_info=self.license_info,
                 routes=self.routes,
                 tags=self.openapi_tags,
                 servers=self.servers,
             )
```
which, it already is. Would vote against all these PR's as not PEP-8 compliant as the rest of the source code.

JarroVGIT (changes requested):

tsv19su254052:
have no objection

tiangolo:

PR #5029, opened 2022-06-13
Replace regex expression

Peopl3s:
replaced the regular expression with a simpler one and remove unused variable


yezz123 (approved):

zoliknemet (approved):

tiangolo:
Thanks for the effort. I'm fine with the variable name there saying what is the data returned.

I guess this particular change is more about taste, but for now I'll pass on this. But thanks for the interest! :coffee:

Issue #5037, opened 2022-06-15
Add healthcheck instructions to example Dockerfile in the documentation


nikstuckenbrock:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
FROM python:3.9

WORKDIR /code

COPY ./requirements.txt /code/requirements.txt

RUN pip install --no-cache-dir --upgrade -r /code/requirements.txt

COPY ./app /code/app

CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "80"]
```

### Description

I recently read a recommendation to add the `HEALTHCHECK` command to a Dockerfile. I was wondering whether it would make sense to include a `HEALTHCHECK` in the FastAPI documentation. In a previous [discussion](https://github.com/tiangolo/fastapi/discussions/5035) I tried to verify if this issue makes sense. I've provided example code of how I would add a basic `HEALTHCHECK`.
Looking forward for some feedback!

### Wanted Solution

Add the `HEALTHCHECK` command to the `Dockerfile` in the FastAPI docs.

### Wanted Code

```python
FROM python:3.9

WORKDIR /code

COPY ./requirements.txt /code/requirements.txt

RUN pip install --no-cache-dir --upgrade -r /code/requirements.txt

COPY ./app /code/app

CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "80"]

HEALTHCHECK CMD http://localhost/
```

### Operating System

Linux

### FastAPI Version

0.75.2

### Python Version

3.9.2

iudeen:
It's a good practice to have health check for containers. A simple short example should suffice. May be we can link to a detailed doc about Docker health check in that for people to read further.

PR #5038, opened 2022-06-15
Add basic healthcheck operation to example Dockerfiles

nikstuckenbrock:
Fixes issue #5037.


iudeen (commented):
Can you also add what the user has to add in the route? By default localhost/ would result in 404 NOTFOUND. 

Also, we should mention if bind port is changed, that port has to be given accordingly.

nikstuckenbrock:
Hey @iudeen,

thanks for the feedback.

> Can you also add what the user has to add in the route? By default localhost/ would result in 404 NOTFOUND.

I've chosen the root route on purpose because every api needs one. I don't think it's a problem that the endpoint will result in a 404. The docker `HEALTHCHECK` command will only check the process result status of curl, which in case of a valid response, is 0, so the container is considered healthy.

> Also, we should mention if bind port is changed, that port has to be given accordingly.

Yes, but the example is build using port 80. Maybe I should just include the port in the URL so it's more intuitive?

iudeen:
> Yes, but the example is build using port 80. Maybe I should just include the port in the URL so it's more intuitive?

Yes! and may be a comment or small note in the doc about it.

nikstuckenbrock:
Hey @iudeen,

I've provided a new version. Is that what you had in mind?

iudeen (approved):

ComicShrimp (approved):

tiangolo:
Thanks @nikstuckenbrock! 🤓 

The consideration of what it means to be "healthy" would be quite dependent on the actual app, it could require a bunch of extra docs explaining all that... and there are also many other possible ideas or options that could be thought of as best practices, and it would not be feasible to include them all, or link to them all, also because it depends a lot on the specific team, etc.

So, for now, I'll pass on this one, but thanks for the effort! ☕

Issue #4975, opened 2022-06-02
Wrong HTTP Response codes in official docs


haasal:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
# I would have to copy the whole example from the doc which wouldn't help in any way so I will just outline the problematic parts:

# Should be 403 without the WWW-Authenticate Header

    for scope in security_scopes.scopes:
        if scope not in token_data.scopes:
            raise HTTPException(
                status_code=status.HTTP_401_UNAUTHORIZED,
                detail="Not enough permissions",
                headers={"WWW-Authenticate": authenticate_value},
            )

# Should't this be a 401?

user = authenticate_user(fake_users_db, form_data.username, form_data.password)
    if not user:
        raise HTTPException(status_code=400, detail="Incorrect username or password")

# Basically the two above but in different places in the docs. See below
```

### Description

I was following the docs and noticed that many of the HTTP error response code where wrong (Or so do I think...).

## Locations

- [Here](https://fastapi.tiangolo.com/advanced/security/oauth2-scopes/#global-view) and in all code "widgets" below I'm pretty sure the 401 should be a 403 like mentioned above. And I'm not sure but I believe the 400 should be a 401.
- 400 is probably fine for the inactive user.
- The errors are also inconsistent between doc pages as [here](https://fastapi.tiangolo.com/advanced/security/http-basic-auth/) 401 is used correctly.
- [Here](https://fastapi.tiangolo.com/tutorial/security/simple-oauth2/) we are using 400 for unauthenticated users again.
- I didn't look over everything but these where the ones I saw.

### 401 -> 403

401 might be called Unauthorized but this is the problem here. It's used for invalid credentials or invalid authentication method where the server doesn't yet recognize the client.
[...]

PR #5044, opened 2022-06-16
Fixing http error codes in security tutorials

haasal:
The docs still have to be updated so the yellow marker is at the correct places.
Before I do this: Can somebody review if everything is more or less fine like this.

And I just noticed that tests still have to be written due to the error code changes

This PR addresses #4975


haasal:
I'm sorry for the close and reopen. I wanted to revert this back to a draft PR

docs_src/security/tutorial003_py310.py
```diff
@@ -66,19 +66,30 @@ async def get_current_user(token: str = Depends(oauth2_scheme)):
 
 async def get_current_active_user(current_user: User = Depends(get_current_user)):
     if current_user.disabled:
-        raise HTTPException(status_code=400, detail="Inactive user")
+        raise HTTPException(
```
JarroVGIT, line 69: I would disagree here.

400 definition:
`The server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).` That is not the case here, it is not a client error.

I would reckon that 403 is more appropriate, especially because the authentication passed but the user is not allowed: `The client does not have access rights to the content; that is, it is unauthorized, so the server is refusing to give the requested resource. Unlike 401 Unauthorized, the client's identity is known to the server.`

docs_src/security/tutorial004.py
```diff
@@ -108,7 +108,9 @@ async def get_current_user(token: str = Depends(oauth2_scheme)):
 
 async def get_current_active_user(current_user: User = Depends(get_current_user)):
     if current_user.disabled:
-        raise HTTPException(status_code=400, detail="Inactive user")
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST, detail="Inactive user"
+        )
```
JarroVGIT, line 113: Same as previous comment

----
I would disagree here.

400 definition:
`The server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).` That is not the case here, it is not a client error.

I would reckon that 403 is more appropriate, especially because the authentication passed but the user is not allowed: `The client does not have access rights to the content; that is, it is unauthorized, so the server is refusing to give the requested resource. Unlike 401 Unauthorized, the client's identity is known to the server.`

docs_src/security/tutorial004.py
```diff
@@ -119,7 +121,7 @@ async def login_for_access_token(form_data: OAuth2PasswordRequestForm = Depends(
         raise HTTPException(
             status_code=status.HTTP_401_UNAUTHORIZED,
             detail="Incorrect username or password",
-            headers={"WWW-Authenticate": "Bearer"},
+            headers={"WWW-Authenticate": "Basic"},
```
JarroVGIT, line 124: This made me think for a bit, but I think I agree with you. Although we are using helper classes with `OAuth2` stuff in them, it is basically a basic authentication. On the other hand, the username/password are following the Oauth2 standards (e.g. with the form) and not with base64 encoding (as is required by the Basic Authentication standard). So, it can be both ways I guess?

docs_src/security/tutorial004_py310.py
```diff
@@ -107,7 +107,9 @@ async def get_current_user(token: str = Depends(oauth2_scheme)):
 
 async def get_current_active_user(current_user: User = Depends(get_current_user)):
     if current_user.disabled:
-        raise HTTPException(status_code=400, detail="Inactive user")
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST, detail="Inactive user"
+        )
```
JarroVGIT, line 112: See other comments, this should be 403.

docs_src/security/tutorial005.py
```diff
@@ -104,14 +104,10 @@ def create_access_token(data: dict, expires_delta: Union[timedelta, None] = None
...
-    if security_scopes.scopes:
-        authenticate_value = f'Bearer scope="{security_scopes.scope_str}"'
-    else:
-        authenticate_value = f"Bearer"
     credentials_exception = HTTPException(
         status_code=status.HTTP_401_UNAUTHORIZED,
         detail="Could not validate credentials",
-        headers={"WWW-Authenticate": authenticate_value},
+        headers={"WWW-Authenticate": "Bearer"},
     )
```
JarroVGIT, line 111: Please explain this, why disallowing `scope` in WWW-Authenticate header?

JarroVGIT (changes requested):
I am in agreement that some of the codes are  wrong codes used in the tutorials and that it should be fixed. However, I am not sure that referencing to an Enum value would be beneficial for in the docs. Using `int`, it is inline with the signature of the __init__ of `HTTPException` and might be easier to understand for people not experienced with FastAPI (which by definition are the majority of the people going through the documentation in the first place). 

Also made comments where I disagree with your suggested code.

haasal:
I think the error code constants make the code much more readable and show intent more clearly. Especially because new people are reading through the docs I think it's important to make constants the default. Also I'm not sure but I think in the docs the usage of constants and ints is rather inconsistent.

docs_src/security/tutorial003_py310.py
haasal: I agree with you

docs_src/security/tutorial004.py
haasal: sftm

docs_src/security/tutorial004.py
haasal: Hmm I don't know. Somehow it's neither of the two options. I found [this](https://stackoverflow.com/questions/1748374/http-401-whats-an-appropriate-www-authenticate-header-value) on this topic. I guess there is no simple soultion. Maybe just leave the WWW-Authenticate out for simplicities sake or leave as is?
Bearer is definitely false in my opinion.

haasal: When using basic auth, the browser prompts you with a username/passwd form. This might not even work, as the browser will probably put the credentials into the header instead of the body. I didn't test this out yet

docs_src/security/tutorial005.py
```diff
@@ -139,15 +134,21 @@ async def get_current_active_user(
     current_user: User = Security(get_current_user, scopes=["me"])
 ):
     if current_user.disabled:
-        raise HTTPException(status_code=400, detail="Inactive user")
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST, detail="Inactive user"
```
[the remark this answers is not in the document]
JarroVGIT: Yes, 403 would be more appropriate imo.

iudeen, line 138:
```suggestion
            status_code=status.HTTP_403_FORBIDDEN,, detail="Inactive user"
```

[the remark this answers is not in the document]
haasal: Ok

haasal:
I don't have time at this moment to write tests and apply all your suggestions. I will complete the PR once I have time again

PR #5052, opened 2022-06-18
Add support to Client Credentials Flow (OAuth2)

Riuzaky77:
issue : https://github.com/tiangolo/fastapi/issues/774


JarroVGIT (commented):
All for this! I do see you didn't provide any tests though, is that on the planning?

fastapi/security/oauth2.py
```diff
@@ -112,6 +117,53 @@ def __init__(
...
 
 
+class OAuth2ClientCredentialsRequestForm:
+    """
+    This is a dependency class, use it like:
+        token_scheme = HTTPBasicClientCredentials(
+            auto_error=False, scheme_name="oAuth2ClientCredentials"
+        )
+
+        @router.post("/login")
```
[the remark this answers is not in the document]
Riuzaky77: Good! will fix the comments and add some tests too!

Riuzaky77:
> All for this! I do see you didn't provide any tests though, is that on the planning?

Will add test yep, sorry!

Riuzaky77:
Ready to go!

kokolem:
Hey @JarroVGIT, just wondering, is there a chance to get this merged? I'm willing to pick this up if there's any additional work needed to be done before merging.

JarroVGIT:
@kokolem I am not the maintainer of this project; I just do some reviews now and then and help people out with issues. @tiangolo is the one that determines what is merged and not, but as you can see there is quite a backlog he is going through. Eventually he'll get to this, I'm sure :)

petri:
could some docs be added to the PR?

Ryandaydev:
Would it be appropriate to add some info about using this flow on the advanced/security/oauth2-scopes/ docs page?

nkukard:
@tiangolo is there any plans to implement client credentials flow?

aidanwardman:
Also looking for a good client credentials flow. Is this still going?

greigs:
Also here hoping for progress on the Client Credentials flow @tiangolo

fastapi/security/oauth2.py
```diff
@@ -214,6 +265,39 @@ async def __call__(self, request: Request) -> Optional[str]:
...
+        auto_error: bool = True,
+    ):
+        if not scopes:
+            scopes = {}
+        flows = OAuthFlowsModel(
+            clientCredentials={"tokenUrl": tokenUrl, "scopes": scopes}
+        )
+        super().__init__(
+            flows=flows,
+            scheme_name=SchemeName.OAUTH2_CLIENT_CREDENTIALS,
```
YuriiMotov, line 282: Why do we need this? Why don't we do it like in other classes (have `scheme_name` param and pass it to `super().__init__()`)?

fastapi/security/oauth2.py
```diff
@@ -112,6 +117,52 @@ def __init__(
...
+                client_id = form.client_id
+                client_secret = form.client_secret
+            elif basic_credentials:
+                client_id = basic_credentials.client_id
+                client_secret = basic_credentials.client_secret
+            else:
+                HTTPException(status_code=400, detail="Client credentials not provided")
+            pass
+
+    This will allow the client to send its credentials either via headers or body with the request for a token.
```
YuriiMotov, line 142: I think we shouldn't describe the _"allow the client to send its credentials either via headers or body with the request for a token"_ use case.
Just:
```py
        def create_access_token(form: Annotated[OAuth2ClientCredentialsRequestForm, Depends()]):
```

The docstring should be updated to follow the style of current implementation for other classes

YuriiMotov (changes requested):
@Riuzaky77, thanks for your interest and efforts!

I think this should be much simpler, without

> This will allow the client to send its credentials either via headers or body with the request for a token.

Currently the implementation is a bit outdated, needs to be updated to align with the current implementation of other security classes.

PR #5057, opened 2022-06-20
♻ Simplify internal RegEx in `fastapi/utils.py`

pylounge:
Replace regex for simple

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -133,14 +133,14 @@ def generate_operation_id_for_path(
         stacklevel=2,
     )
     operation_id = name + path
-    operation_id = re.sub("[^0-9a-zA-Z_]", "_", operation_id)
+    operation_id = re.sub("\W", "_", operation_id)
     operation_id = operation_id + "_" + method.lower()
     return operation_id
 
 
 def generate_unique_id(route: "APIRoute") -> str:
     operation_id = route.name + route.path_format
-    operation_id = re.sub("[^0-9a-zA-Z_]", "_", operation_id)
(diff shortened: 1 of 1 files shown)


JarroVGIT (approved):
Agree, `\W` is the equivalent of `[^a-zA-Z0-9_]` and is more elegant.

lucaswiman:
> Agree, `\W` is the equivalent of `[^a-zA-Z0-9_]` and is more elegant.

No, that's not true. It's equivalent to `[^\w]`, which is a wider range of characters that just those. For example:

```python
>>> import re
>>> re.match(r'\W', 'ü')
>>> re.match(r'\w', 'ü')
<re.Match object; span=(0, 1), match='ü'>
```

This change may lead to better support for languages with non-ASCII characters, but it's not precisely equivalent to the old code.

JarroVGIT:
> > Agree, `\W` is the equivalent of `[^a-zA-Z0-9_]` and is more elegant.
> 
> No, that's not true. It's equivalent to `[^\w]`, which is a wider range of characters that just those. For example:
> 
> ```python
> >>> import re
> >>> re.match(r'\W', 'ü')
> >>> re.match(r'\w', 'ü')
> <re.Match object; span=(0, 1), match='ü'>
> ```
> 
> This change may lead to better support for languages with non-ASCII characters, but it's not precisely equivalent to the old code.

I stand corrected. I took my information from regex101.com, and you are completely right. 

If this changes the behaviour, then I would expect a test case that would illustrate the exact intended behaviour.

tiangolo:
Great, thanks @pylounge! :cake:

PR #5075, opened 2022-06-24
Fix formats in fastapi/utils

drcode-rune:
Fix & improve formats in utils


fastapi/utils.py
```diff
@@ -169,8 +167,7 @@ def get_value_or_default(
 
     Otherwise, the first item (a `DefaultPlaceholder`) will be returned.
     """
-    items = (first_item,) + extra_items
-    for item in items:
+    for item in (first_item,) + extra_items:
```
JarroVGIT, line 170: It's a matter of taste of course, but I find the single line harder to read than the multi line code it replaces.

fastapi/utils.py
```diff
@@ -134,15 +132,15 @@ def generate_operation_id_for_path(
...
+    operation_id = f"{operation_id}_{method.lower()}"
     return operation_id
 
 
 def generate_unique_id(route: "APIRoute") -> str:
     operation_id = route.name + route.path_format
     operation_id = re.sub("[^0-9a-zA-Z_]", "_", operation_id)
     assert route.methods
-    operation_id = operation_id + "_" + list(route.methods)[0].lower()
+    operation_id = f"{operation_id}_{list(route.methods)[0].lower()}"
```
JarroVGIT, line 143: This is proper, it is in line with other LoC in this package so I would be supporting this.

fastapi/utils.py
```diff
@@ -134,15 +132,15 @@ def generate_operation_id_for_path(
     )
     operation_id = name + path
     operation_id = re.sub("[^0-9a-zA-Z_]", "_", operation_id)
-    operation_id = operation_id + "_" + method.lower()
+    operation_id = f"{operation_id}_{method.lower()}"
```
JarroVGIT, line 135: And therefor, I would also be supporting it here.

JarroVGIT (changes requested):
I like the changes to the string formatting (e.g using `f"blaat{var}"` as it is more widely used in other parts of FastAPI. There is one more instance I found you could add: 

https://github.com/tiangolo/fastapi/blob/1876ebc77949a9a254909ec61ea0c09365169ec2/fastapi/dependencies/utils.py#L724

However, the changes on line 54-56 are less explicit than what they are replacing. I find it better to set variables explicitly rather than inline when passing them as parameters in another method call (which you are proposing). The same goes for line 172/173; I like it better on two lines than in a single statement in a for-statement.

fastapi/utils.py
LeSZ0: Thanks @JarroVGIT for you review I appreciate it.
I thought this change will be more explicit, but I could agree with you, it might add some complexity to the code.
I'll roll back the changes using the previous multi line approach.

JarroVGIT (approved):

fastapi/utils.py
```diff
@@ -85,7 +84,7 @@ def create_model_field(
         kwargs.update(
             {
                 "type_": type_,
-                "class_validators": class_validators,
+                "class_validators": class_validators or {},
```
YuriiMotov, line 87: I think it was better before. I would revert this and restored `class_validators = class_validators or {}` in the beginning

YuriiMotov (commented):
@LeSZ0, thanks for your interest in FastAPI!

Some changes have already been applied in other PRs. So, only 2 changes remain.

I think changing string concatenation to f-strings should be done more consistently - across the entire codebase.

PR #5077, opened 2022-06-24
✨ Add support for wrapped functions (e.g. `@functools.wraps()`) used with forward references

### LINKED ISSUE #5065: FastAPI does not respect __wrapped__ attribute, causing NameError for forward references in decorated functions.
### First Check

### Commit to Help

### Example Code

```python
# This needs to be in two files because the nature of the bug involves different globalns
# entries for a decorated function versus the module of the original function.
# If you put the code here into two files (func.py and main.py) in a directory, then the
# bug can be reproduced by running:
# PYTHONPATH=.:$PYTHONPATH python ./main.py
# This should cause an exception `NameError: name 'MyModel' is not defined`

# func.py


def foo(obj: "MyModel") -> "MyModel":
    return obj


class MyModel(BaseModel):
    x: int

# main.py
import functools, typing

from fastapi import FastAPI
from func import foo

app = FastAPI()


def deco(func):
    @functools.wraps(func)
    def passthrough(*args, **kwargs):
        return func(*args, **kwargs)
    return passthrough

decorated = deco(foo)

assert issubclass(typing.get_type_hints(decorated)["obj"], BaseModel)
assert decorated.__annotations__ == foo.__annotations__
assert isinstance(decorated.__annotations__["obj"], str)

method = app.post('/endpoint')(decorated)
```

### Description

Using a decorator employing `functools.wraps` or a similar causes a `NameError` in `fastapi/dependencies/utils.py` in `get_typed_annotation`. The implementation there differs from the implementation in `typing`, which dereferences the `__wrapped__` attribute to find the namespace the original function was defined in. This allows forward references to be handled correctly:
https://github.com/python/cpython/blob/576dd901170af30fc50b0a7f07a388b38fd724a9/Lib/typing.py#L2314-L2315

In the "Example Code" section, my example includes assertions showing that `typing.get_type_hints` handles this situation correctly. I believe this can be fixed either by using the same implementation as typing uses, or by calling `typing.get_type_hints`, and only doing special FastAPI-specific logic if that fails.

### Operating System

macOS

### Operating System Details

I first saw this on Ubuntu Linux with python 3.8, then later reproduced it on my Macbook in python 3.9. The example code was tested on the Mac.

I do not believe the bug is related to operating systems. I think it's a bug in how pure python code handles type annotations.

### FastAPI Version

0.78.0

### Python Version

Python 3.9.1

### Additional Context

This bug is unlikely to come up in most ordinary code, but is likely to come up when using metaprogramming or decorators (especially in programmatically adding endpoints). The context I found it in was in a in-house workflow framework we are building at my job. The framework uses some decorators to alter the type signature of methods, and passes them to FastAPI to expose task endpoints. This leads to the above error.

I definitely understand that such an odd use case is probably not high priority for the maintainers, but I hope you'll be willing to review a PR from me fixing the bug.

Thank you for all your work on FastAPI! It is an amazingly well-made piece of software.

lucaswiman:
Fixes #5065 where the `globalns` computation in `get_typed_signature` misses an edge case handled by `typing.get_type_hints`. This PR makes the logic similar to `get_type_hints`.

## Background

`functools.wraps` updates the annotations and type signature to match the wrapped function. However, it copies the `__annotations__` dict verbatim, including forward references (strings). `get_type_hints` handles this by [dereferencing the `__wrapped__` attribute](https://github.com/python/cpython/blob/576dd901170af30fc50b0a7f07a388b38fd724a9/Lib/typing.py#L2314-L2315) until it gets to the original function, then uses the `__globals__` of _that_ function.

## Testing

I added a test case which reproduced the `NameError` seen in #5065. The test passes after my commit updating the implementation of `get_typed_signature`:

```
tests/test_wrapped_method_forward_reference.py:24:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
fastapi/routing.py:617: in decorator
    self.add_api_route(
fastapi/routing.py:556: in add_api_route
    route = route_class(
fastapi/routing.py:425: in __init__
    self.dependant = get_dependant(path=self.path_format, call=self.endpoint)
fastapi/dependencies/utils.py:279: in get_dependant
    endpoint_signature = get_typed_signature(call)
fastapi/dependencies/utils.py:249: in get_typed_signature
    typed_params = [
fastapi/dependencies/utils.py:254: in <listcomp>
    annotation=get_typed_annotation(param, globalns),
fastapi/dependencies/utils.py:266: in get_typed_annotation
    annotation = evaluate_forwardref(annotation, globalns, globalns)
pydantic/typing.py:76: in pydantic.typing.evaluate_forwardref
    ???
../../.pyenv/versions/3.9.1/lib/python3.9/typing.py:533: in _evaluate
    eval(self.__forward_code__, globalns, localns),
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   NameError: name 'ForwardRef' is not defined

<string>:1: NameError
============================================================== short test summary info ===============================================================
FAILED tests/test_wrapped_method_forward_reference.py::test_wrapped_method_type_inference - NameError: name 'ForwardRef' is not defined
================================================================= 1 failed in 0.59s ==================================================================
```


JarroVGIT (approved):
Very nice, agree with the solution and good work on providing a testcase with proper docstring explaining the issue at hand.

fastapi/dependencies/utils.py
```diff
@@ -245,7 +245,13 @@ def is_scalar_sequence_field(field: ModelField) -> bool:
 
 def get_typed_signature(call: Callable[..., Any]) -> inspect.Signature:
     signature = inspect.signature(call)
-    globalns = getattr(call, "__globals__", {})
+    nsobj = call
+    while hasattr(nsobj, "__wrapped__"):
+        # The __wrapped__ attribute is set by decorators, e.g. functools.wraps.
+        # This while loop allows rereferencing forward references on decorated
+        # methods.
+        nsobj = nsobj.__wrapped__  # type: ignore
```
uriyyo, line 253: It can be replaced with `inspect.unwrap`:
```suggestion
    nsobj = inspect.unwrap(call)
```

uriyyo (commented):
LGTM. I have one small suggestion

yezz123 (approved):
LGTM ⚡️

fastapi/dependencies/utils.py
lucaswiman: Looks like that was added in python 3.4 and the earliest supported version of python for FastAPI is 3.6, so sure. It also checks for cycles (which `typing.get_type_hints` does not).

uriyyo (approved):
LGTM

lucaswiman:
@tiangolo This PR has 3 approvals. Could you please merge this PR or comment about what additional steps I should take prior to merge? Thanks!

BilalAlpaslan (approved):

zachkirsch:
+1 I'd love for this to merge in!

dsinghvi:
@tiangolo what would it take to get this merged in?

dsinghvi:
@tiangolo following up again -- pr has 3 approvals, what can we do to get this in?

ashwin153:
@tiangolo Would it be possible to get this merged?

zachkirsch:
@tiangolo is there anything is particular that is blocking this PR from merging?

corps:
Also running into this and it's super subtle.
In our case we have a situation where the base abstract service's module has `Encounter` in its globals namespace, but the implementation (subclass) that gets registered has a /different/ model name `Encounter` its namespace.  The existing logic does an update, poluting namespaces by overriding one meaning of the word `Encounter` with the other module's, resulting in super subtle bugs downstream.


```
class AbstractEncountersV4Service(AbstractFernService):
    def some_method(self) -> Encounter:
        ...
```

```
from orm.models import Encounter

class MyEncounterService(AbstractEncountersV4Service):
     def some_method(self) -> fern.Encounter:
        ...  
```

In the end, the abstract class's `__globals__` gets polluted from the implementation's view, which /sometimes/ ends up being fine but /sometimes/ isn't, making for unreliable subtle behavior.

So this PR would majorly improve these edge cases.

tiangolo (approved):
Thank you for the fix @lucaswiman! 🙌 

And thanks everyone for the discussion. 🍰 

Sorry for the long delay getting back to this.

It will be available in FastAPI version 0.123.5 released in the next few hours. 🎉

PR #5082, opened 2022-06-26
websocket disconnect

objecthuman:
Previously in the `disconnect` method, the connection was not closed. The WebSocket object was removed from the list only.
I have called the `close` method before removing the object from the `list`.

### DIFF
--- a/docs_src/websockets/tutorial003.py
+++ b/docs_src/websockets/tutorial003.py
@@ -51,7 +51,8 @@ class ConnectionManager:
         await websocket.accept()
         self.active_connections.append(websocket)
 
-    def disconnect(self, websocket: WebSocket):
+    async def disconnect(self, websocket: WebSocket):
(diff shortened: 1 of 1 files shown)


docs_src/websockets/tutorial003.py
```diff
@@ -79,5 +80,5 @@ async def websocket_endpoint(websocket: WebSocket, client_id: int):
             await manager.send_personal_message(f"You wrote: {data}", websocket)
             await manager.broadcast(f"Client #{client_id} says: {data}")
     except WebSocketDisconnect:
-        manager.disconnect(websocket)
```
JarroVGIT, line 82: This would call `WebSocket.close()` on a already closed connection.

JarroVGIT (commented):
The disconnect method is called in the example, only when the client has disconnected.

PR #5095, opened 2022-06-30
memo: Fixes a typo in the documentation for numeric validations

frgfm:
Hello there :wave: 

Thanks so much for this awesome library, which I use quite a lot whenever some back-end is involved!
I was checking the documentation about numeric validations and found a weird rendering issue:
image

The documentation section is https://fastapi.tiangolo.com/tutorial/path-params-numeric-validations/#number-validations-greater-than-or-equal

I figured someone might have done an automatic replacement somewhere which caused this. This PR simply fixes that!

Any feedback is welcome!

### DIFF
--- a/docs/en/docs/tutorial/path-params-numeric-validations.md
+++ b/docs/en/docs/tutorial/path-params-numeric-validations.md
@@ -79,7 +79,7 @@ Python won't do anything with that `*`, but it will know that all the following 
 
 With `Query` and `Path` (and other's you'll see later) you can declare string constraints, but also number constraints.
 
-Here, with `ge=1`, `item_id` will need to be an integer number "greater than or equal" to `1`.
+Here, with `ge=1`, `item_id` will need to be an integer number "**g**reater than or **e**qual" to `1`.
 
 ```Python hl_lines="8"
 {!../../../docs_src/path_params_numeric_validations/tutorial004.py!}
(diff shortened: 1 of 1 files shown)


frgfm:
> memo Docs preview for commit [b661c22](https://github.com/tiangolo/fastapi/commit/b661c22535163df00fa7fadf57b41f505cdf71b2) at: https://62bded71abd38957507b41d6--fastapi.netlify.app

Impressive CI :relieved:

JarroVGIT:
I am not sure this is a mistake, but rather a reference to what the 'g' and 't' mean in `gt` or the 'l' and 'e' in `le`.

JarroVGIT (changes requested):
The change that you are proposing is from documentation that has been the same for over 3.5 years. I think it is quite elegant in how it brings the point across (of the meaning of 'g' and 'e').

frgfm:
Oh actually now I see it :sweat_smile: 
I would perhaps suggest making those letters bold rather than in code-format, what do you think?
I can easily update the PR accordingly

frgfm:
That looks more intuitive to me now :point_right: https://62c6aabd77d8786b6aa5dfdd--fastapi.netlify.app/tutorial/path-params-numeric-validations/

Happy to close the PR if you don't think it's an improvement :)

odiseo0 (approved):
LGTM

tiangolo:
Thanks @frgfm and everyone for the conversation! I actually think that the "code" formatting makes it even more obvious that it refers to those letters. So I prefer to keep it as-is. Given that, I'll close this one now. But thanks for the effort and interest! :cake:

PR #5114, opened 2022-07-08
Run sync context managers in a single thread

### LINKED ISSUE #3205: Using dependency injection to get SQLAlchemy session can lead to deadlock
### First check

- [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
- [x] After submitting this, I commit to one of:
  - Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
  - I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
  - Implement a Pull Request for a confirmed bug.

I've noticed that when using dependency injection with SQLAlchemy, a large number of concurrent requests can leave the app in a deadlocked state. This is especially noticeable with a small SQLAlchemy connection pool size (relative to the FastAPI thread pool size). Below is a self-contained example (you might have to tweak the pool size and the request body's sleep length but this should be a good starting point).

<details>
  <summary>View <code>app.py</code></summary>

```python
"""
Setup: pip install fastapi sqlalchemy uvicorn
Run: python app.py
"""
import time
import uvicorn

from sqlalchemy import create_engine
from sqlalchemy.orm import Session, sessionmaker
from sqlalchemy.pool import QueuePool


# SQLAlchemy setup
engine = create_engine(
    'sqlite:///test.db',
    connect_args={'check_same_thread': False},
    poolclass=QueuePool,
    pool_size=4,
    max_overflow=0,
    pool_timeout=None,  # Wait forever for a connection
)
SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)

# FastAPI
app = FastAPI()

def get_db(request: Request):
    db = SessionLocal()
    try:
        yield db
    finally:
        db.close()

@app.get('/')
def index(db: Session = Depends(get_db)):
    # Some blocking work
    _ = db.execute('select 1')
    time.sleep(1)
    return {'hello': 'world'}

# Run
if __name__ == '__main__':
    uvicorn.run('app:app', reload=True, host='0.0.0.0', port=80)
```

</details>

When running the above with 100 concurrent requests (I used [locust](https://locust.io/)), I noticed that only around 5 requests are served, and then the app freezes and is unable to serve any more requests. Below is the locustfile.

<details>
  <summary>View <code>locustfile.py</code></summary>

```python
"""
Setup: pip install locust
Run: Save as locustfile.py and run locust in terminal. Open http://localhost:8089 and run with 100, 100, http://localhost.
"""
from locust import HttpUser, task


class User(HttpUser):
    @task
    def index(self):
        self.client.get("/")
```

</details>

I suspect the following is happening. (Note that `SessionLocal()` is lazy, so `db = SessionLocal()` will return immediately even if no connections are available.)

1. The first `N` requests come in (where `N` >= thread pool size). Their `get_db` dependencies run and yield, and we start executing their path operation functions. At this point, the entire thread pool is full. Only `pool_size` (4) requests are able to get a connection, and the remaining requests wait (in their path operation functions).
2. The path operation functions that were able to get a connection return, opening up `pool_size` (4) spots in the thread pool. Because dependencies and requests run in separate threads, the `finally` blocks for these requests' `get_db` dependencies have not run yet, so the connections for these requests have not returned to the SQLAlchemy pool.
3. More requests come in, and like step 1, their `get_db` dependencies run, and we start executing their path operation functions. No connections have returned to the SQLAlchemy pool, so these requests wait. At this point, the entire thread pool is full, and every thread is waiting for a connection.
4. For the requests that finished in step 2, we try to schedule the `finally` blocks for their `get_db` dependencies in a new thread so we can free the connections, but all of the threads are busy, so we end up waiting.
5. None of the threads will ever finish because they are waiting for a connection, and no connections will be released because the thread pool is full, leaving the app in a deadlocked state.

This doesn't really seem like a bug in FastAPI or in SQLAlchemy, but it suggests that we should not use dependency injection like this when using synchronous database libraries. The only workaround I've found for this is to use a context manager to handle the session in the endpoint itself instead of injecting the database session directly.

Another thing I've noticed is that changing `get_db` to be an async function prevents deadlock (as does using the [middleware approach](https://fastapi.tiangolo.com/tutorial/sql-databases/#alternative-db-session-with-middleware)), but only if the endpoint does not have a `response_model`. If it has a `response_model` then the app will still lock up. I believe this is because if `response_model` is defined, then when we run [`serialize_response`](https://github.com/tiangolo/fastapi/blob/master/fastapi/routing.py#L106), `field` will be non-None, and we will attempt to run `field.validate` in a separate thread. If the thread pool is full with requests waiting for connections, we won't be able to serialize the response and won't be able to close the database connection. Maybe we could serialize the response in the same thread as the path operation function; I'm not sure what the benefit of serializing in a separate thread is.

There is similar discussion in https://github.com/tiangolo/full-stack-fastapi-postgresql/issues/104 and many others came to the conclusion that using a context manager is the right approach, but nothing really came of it. If others can validate that my suspicion is correct, then maybe we should change the docs to recommend using a context manager within the endpoint itself until a better solution is available.

### Environment

- OS: macOS Big Sur (11.3)
- FastAPI Version: 0.63.0
- SQLAlchemy Version: 1.4.13
- Python version: 3.9.4

adriangb:
This avoids a deadlock where `__exit__` can't be called because there are no thread tokens available. Partially resolves #3205 and other related issues.

### DIFF
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -45,10 +45,10 @@ async def contextmanager_in_threadpool(
         0, item_type=Optional[Exception]
     )
     async with AsyncExitStack() as stack:
-        await stack.enter_async_context(rcv_res)
-        await stack.enter_async_context(rcv_err)
-        await stack.enter_async_context(send_res)
-        await stack.enter_async_context(send_err)
+        stack.enter_context(rcv_res)
+        stack.enter_context(rcv_err)
+        stack.enter_context(send_res)
+        stack.enter_context(send_err)
         tg = await stack.enter_async_context(anyio.create_task_group())
         tg.start_soon(run_in_threadpool, _cm_thead_worker, cm, send_res, rcv_err)
         res = await rcv_res.receive()
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -37,9 +37,11 @@ async def contextmanager_in_threadpool(
     cm: ContextManager[_T],
 ) -> AsyncGenerator[_T, None]:
     # streams for the data
-    send_res, rcv_res = anyio.create_memory_object_stream(0, item_type=Any)
+    send_res, rcv_res = anyio.create_memory_object_stream(  # type: ignore
+        0, item_type=Any
+    )
     # streams for exceptions
-    send_err, rcv_err = anyio.create_memory_object_stream(
+    send_err, rcv_err = anyio.create_memory_object_stream(  # type: ignore
         0, item_type=Optional[Exception]
     )
     async with AsyncExitStack() as stack:
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -1,6 +1,8 @@
 import sys
-from typing import AsyncGenerator, ContextManager, TypeVar
+from typing import Any, AsyncGenerator, ContextManager, Optional, TypeVar
 
+import anyio
+from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from starlette.concurrency import iterate_in_threadpool as iterate_in_threadpool  # noqa
 from starlette.concurrency import run_in_threadpool as run_in_threadpool  # noqa
 from starlette.concurrency import (  # noqa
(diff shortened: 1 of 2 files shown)


yezz123 (approved):

adriangb:
Closing in favor of #5122

PR #5160, opened 2022-07-14
🔧 Make all badges point to img.shields.io

Jorricks:
Currently there is one badge that points to Github directly.
This causes that this badge might be displayed in a different color than the rests.


tiangolo:
Thanks for the interest! I actually like the badge for GitHub :sweat_smile: :shrug: 

So I'll close this one, but thanks for the interest! :cake:

Jorricks:
All good 😁 cheers

PR #5163, opened 2022-07-15
✨ Allow to set custom auto error detail

mahenzon:
Allow to set custom auto error detail for HTTP Security dependencies

Sometimes it's hard to find out which dependency returns a "Not authenticated" error. It would be handy if we could set custom error detail text to easily identify the error reason. For example "auth token not provided" for one HTTP Security object and "private id not provided" for another


JarroVGIT (changes requested):
LGTM, but would recommend to add some test case(s) to improve coverage.

mahenzon:
Hey @JarroVGIT, sorry for the delay! 😅 

I've added some tests. Could you please check? 
Many thanks!

Kludex (approved):

mahenzon:
Hey there! I've rebased my branch onto master and solved all the conflicts. 
Could you please check? Thank you!

svlandeg:
Apologies for the delay and thanks - we'll have a look!

fastapi/security/http.py
```diff
@@ -178,11 +178,24 @@ def __init__(
...
             ),
         ] = True,
+        auto_error_detail: Annotated[
+            str,
+            Doc(
+                """
+                The text to be returned to the client when `auto_error`
+                raises an HTTP exception.
+
+                It useful when you have multiple errors defined: set
```
svlandeg, line 188:
```suggestion
                It's useful when you have multiple errors defined: set
```

fastapi/security/http.py
```diff
@@ -293,10 +306,23 @@ def __init__(
...
```
svlandeg, line 316:
```suggestion
                It's useful when you have multiple errors defined: set
```

fastapi/security/http.py
```diff
@@ -395,10 +422,23 @@ def __init__(
...
```
svlandeg, line 432:
```suggestion
                It's useful when you have multiple errors defined: set
```

svlandeg (approved):
This mostly looks good to me, thanks again for your contribution!

I'll leave this up for Tiangolo for a final review. Should we consider adding a note in the documentation somewhere?

YuriiMotov (commented):
Although it might still be convenient to have such parameter, now there is quite simple way to implement this logic.

According to [docs](https://fastapi.tiangolo.com/how-to/authentication-error-status-code/) we can inherit from existing security class end override the `make_not_authenticated_error` method:

```py
class MyHTTPBasic(HTTPBasic):
    def __init__(self, *, auto_error: bool = True, auto_error_detail: str = "Not authorized"):
        super().__init__(auto_error=auto_error)
        self.auto_error_detail = auto_error_detail

    def make_not_authenticated_error(self) -> HTTPException:
        return HTTPException(
            status_code=status.HTTP_403_FORBIDDEN, detail=self.auto_error_detail
        )
```

<details><summary>Full code</summary>

```py
from typing import Optional

from fastapi import FastAPI, Security, HTTPException, status
from fastapi.security import HTTPBasic, HTTPBasicCredentials

class MyHTTPBasic(HTTPBasic):
    def __init__(self, *, auto_error: bool = True, auto_error_detail: str = "Not authorized"):
        super().__init__(auto_error=auto_error)
        self.auto_error_detail = auto_error_detail

    def make_not_authenticated_error(self) -> HTTPException:
        return HTTPException(
            status_code=status.HTTP_403_FORBIDDEN, detail=self.auto_error_detail
        )


app = FastAPI()

error_message = "not a 20 minute adventure"
security = MyHTTPBasic(auto_error=True, auto_error_detail=error_message)


@app.get("/users/me")
def read_current_user(credentials: Optional[HTTPBasicCredentials] = Security(security)):
    return {"username": credentials.username, "password": credentials.password}


client = TestClient(app)


def test_security_http_basic():
    response = client.get("/users/me", auth=("john", "secret"))
    assert response.json() == {"username": "john", "password": "secret"}


def test_security_http_basic_no_credentials():
    response = client.get("/users/me")
    assert response.status_code == 403, response.text
    assert response.json() == {"detail": error_message}
```

</details>

tiangolo:

I want to implement a few big refactors and features that might solve many of these use cases.

Thank you! 🍰

PR #5187, opened 2022-07-23
🐛 Fix support for strings in OpenAPI status codes: `default`, `1XX`, `2XX`, `3XX`, `4XX`, `5XX`

JarroVGIT:
Related to https://github.com/tiangolo/fastapi/issues/5166

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -19,7 +19,9 @@ if TYPE_CHECKING:  # pragma: nocover
 
 
 def is_body_allowed_for_status_code(status_code: Union[int, str, None]) -> bool:
-    if status_code in [None, 'default', '2XX', '2xx', '3XX', '3xx', '4XX', '4xx', '5XX', '5xx']:
+    if status_code is None:
+        return True
+    if status_code in ['default', '2XX', '2xx', '3XX', '3xx', '4XX', '4xx', '5XX', '5xx']:
         return True
     current_status_code = int(status_code)
     return not (current_status_code < 200 or current_status_code in {204, 304})
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -19,7 +19,7 @@ if TYPE_CHECKING:  # pragma: nocover
 
 
 def is_body_allowed_for_status_code(status_code: Union[int, str, None]) -> bool:
-    if status_code is None:
+    if status_code is None or status_code == "default":
         return True
     current_status_code = int(status_code)
     return not (current_status_code < 200 or current_status_code in {204, 304})
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -19,7 +19,7 @@ if TYPE_CHECKING:  # pragma: nocover
 
 
 def is_body_allowed_for_status_code(status_code: Union[int, str, None]) -> bool:
-    if status_code is None or status_code == "default":
+    if status_code in [None, 'default', '2XX', '2xx', '3XX', '3xx', '4XX', '4xx', '5XX', '5xx']:
         return True
     current_status_code = int(status_code)
     return not (current_status_code < 200 or current_status_code in {204, 304})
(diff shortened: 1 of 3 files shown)


fastapi/utils.py
```diff
@@ -19,7 +19,7 @@
 
 
 def is_body_allowed_for_status_code(status_code: Union[int, str, None]) -> bool:
-    if status_code is None:
+    if status_code is None or status_code == "default":
```
MarkusSintonen, line 22: This wont fix the OpenAPI status code range definitions, right?

See: https://swagger.io/docs/specification/describing-responses/#status-codes

JarroVGIT: That is correct, that is currently not supported. Was that working prior to 0.79? The fix would be easy though:

`if status_code is None or status_code == "default" or status_code in ["1XX", "2XX", "3XX", "4XX", "5XX"]:`

MarkusSintonen: > Was that working prior to 0.79?

Yes that was working previously. So eg defining:
```
FastAPI(..., responses={"5XX": {"model": ServerErrorResponse, "description": "Unknown Server Error"}})
```
Yields spec like in https://swagger.io/docs/specification/describing-responses/#status-codes for all of the routes.

Probably these unit tests should be extended to cover the regression: https://github.com/tiangolo/fastapi/blob/master/tests/test_additional_responses_router.py

JarroVGIT: i'd agree, I'll update the PR when i have time (propably tomorrow). Didn't know about the ranges, did you open an issue about those? I'll reference them in this PR for closure then.

JarroVGIT:
@tiangolo I know you are a busy man, but 0.79 unfortunately introduced a regression. This is fixed with this commit, and I've written the test case to prove it. Perhaps if you have the time, you could glance over it and produce a minor update?

fastapi/utils.py
```diff
@@ -21,6 +21,18 @@
 def is_body_allowed_for_status_code(status_code: Union[int, str, None]) -> bool:
     if status_code is None:
         return True
+    if status_code in [
```
iudeen, line 24: it'd be more efficient to have a set here instead of list?

MarkusSintonen: That would be also more consistent below where its also doing set-based check. So here using instead:
```python
    # https://swagger.io/docs/specification/describing-responses/#default
    # https://swagger.io/docs/specification/describing-responses/#status-codes
    if status_code in {
        "default",
        "2XX",
        "2xx",
        "3XX",
        "3xx",
        "4XX",
        "4xx",
        "5XX",
        "5xx",
    }:
        return True
```
Also adding a comment where these come from would be a good idea.

tests/test_additional_responses_router.py
```diff
@@ -33,6 +39,18 @@ async def c():
     return "c"
 
 
+@router.get(
+    "/d",
+    responses={
+        "400": {"description": "Error with str"},
+        "5xx": {"model": ResponseModel},
```
Kyle-sandeman-mrdfood, line 46: this is lowercase, is there a conversion hidden somewhere? Your edit to `is_body_allowed_for_status_code` does not `upper()` it

Kyle-sandeman-mrdfood:
Any status update on this PR? For our use case we have to keep the library pinned on an older version until this is merged. Or reduce the quality of our docs

yezz123 (approved):

BilalAlpaslan (approved):

dmitry-engineer (approved):

tiangolo:
Thanks for the contribution @JarroVGIT! 🍰 

And thanks for the input here everyone! ☕ 

I updated it a bit to include `1XX` and remove lower case versions to keep in line with the spec. This will be available in the next release, in the next hours. 🚀

fastapi/utils.py
```diff
@@ -21,6 +21,16 @@
 def is_body_allowed_for_status_code(status_code: Union[int, str, None]) -> bool:
     if status_code is None:
         return True
+    # Ref: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#patterned-fields-1
```
Kludex, line 24: FastAPI supports only 3.0.2, not 3.1.0.

Both versions are compliant with this change.

tiangolo: Right!

PR #5189, opened 2022-07-24
⬆ Upgrade Swagger UI - oauth2-redirect.html

mirusu400:
Pulled from : https://github.com/swagger-api/swagger-ui/blob/master/dist/oauth2-redirect.html

Current oauth redirection template is from commit [9253c0bf03ff83aac7b8a51dc83fe087502eef99](https://github.com/swagger-api/swagger-ui/blob/9253c0bf03ff83aac7b8a51dc83fe087502eef99/dist/oauth2-redirect.html), So I pulled from latest commit.

### DIFF
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -118,7 +118,7 @@ def get_swagger_ui_oauth2_redirect_html() -> HTMLResponse:
     html = """
     <!DOCTYPE html>
     <html lang="en-US">
-    <body onload="run()">
+    <body>
     </body>
     </html>
     <script>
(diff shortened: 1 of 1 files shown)


ypsah:
Hi, I opened almost the same PR to fix a bug with the "authorization_code with PKCE" authentication flow when `OpenIdConnect` is used. I only realized you created yours after the fact. :sweat_smile:

Normally, I'd close mine, but I feel like we'd miss some valuable context by not including the details I added to my commit message. Would you be open to merging your commit message and mine?

If so I'd be happy to close my PR and leave less work for the maintainers. :slightly_smiling_face:

tiangolo:
Thanks for the effort @mirusu400! :cake: 

This was recently solved in another PR, so I'll close this one. But thanks for the effort! :coffee:

PR #5195, opened 2022-07-25
Hotfix: redoc tables have to be fullwidth, but they aren't

kai3341:
The reason is tables have CSS property `{ display: block }`, that's why they renders wrong


tiangolo:
Could you please create an issue? Probably with screenshots to understand what's the problem. 

I think it's quite probable that if there's a problem, it's actually in ReDoc, and it should actually be solved there...

Thanks for the effort and interest! :coffee:

PR #5198, opened 2022-07-26
Refactoring

SunriseFading:


fastapi/dependencies/utils.py
```diff
@@ -100,22 +100,19 @@ def check_file_field(field: ModelField) -> None:
                 from multipart.multipart import parse_options_header  # type: ignore
 
                 assert parse_options_header
-            except ImportError:
+            except ImportError as e:
                 logger.error(multipart_incorrect_install_error)
-                raise RuntimeError(multipart_incorrect_install_error)
```
JarroVGIT, line 105: This is not refactoring but behavioural change. It will no longer show "while handling the Exception, another exception happened" but "this is cause by". I am not sure if it is clearer.

fastapi/encoders.py
```diff
@@ -109,22 +109,20 @@ def jsonable_encoder(
...
+            jsonable_encoder(
+                item,
+                include=include,
+                exclude=exclude,
+                by_alias=by_alias,
+                exclude_unset=exclude_unset,
+                exclude_defaults=exclude_defaults,
+                exclude_none=exclude_none,
+                custom_encoder=custom_encoder,
+                sqlalchemy_safe=sqlalchemy_safe,
```
JarroVGIT, line 122: Although it is inline with your other suggestions, this one might result in less readable code rather than more readable code.

fastapi/dependencies/utils.py
JarroVGIT: On second thought, I think I tend to oppose this change. The stack trace of the 'old' way mentions `During handling of the above exception, another exception occurred:` which is exactly what is happening. Meaning, `the above exception` refers to the ImportError, while the `another exception` refers to the RuntimeError. 

In your suggestion, this text will change to  `The above exception was the direct cause of the following exception:` where as this is not correct. The `above exception` is not a result from raising a RuntimeError. Hence, I think this might be confusing for users of FastAPI.

fastapi/encoders.py
```diff
@@ -141,7 +139,7 @@ def jsonable_encoder(
             data = vars(obj)
         except Exception as e:
             errors.append(e)
-            raise ValueError(errors)
+            raise ValueError(errors) from e
```
JarroVGIT, line 142: See my earlier comment, I don't think this is an improvement.

fastapi/middleware/asyncexitstack.py
```diff
@@ -18,7 +18,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
                     await self.app(scope, receive, send)
                 except Exception as e:
                     dependency_exception = e
-                    raise e
+                    raise dependency_exception from dependency_exception
```
JarroVGIT, line 21: This is more confusing than the 'old' way, the 'from' statement doesn't add anything (nor technically, nor in readability)

fastapi/security/http.py
```diff
@@ -89,8 +89,8 @@ async def __call__(  # type: ignore
                 return None
         try:
             data = b64decode(param).decode("ascii")
-        except (ValueError, UnicodeDecodeError, binascii.Error):
-            raise invalid_user_credentials_exc
```
JarroVGIT, line 93: This is behavioural change, why removing the UnicodeDecodeError? That type of exception can be thrown by the `encode()` call.

JarroVGIT (changes requested):
In general great work, better consistency and more pythonic code. I do have a comment about all the `raise SomeException from e`, see my comment about that bit. I think the stack trace is less correct.

SunriseFading:
Thanks for corrections

PR #5216, opened 2022-07-31
Allow return type of endpoint to be response model

### LINKED ISSUE #5215: Ability to define response_model with native typing
### First Check

### Commit to Help

### Example Code

```python
from fastapi import FastAPI

app = FastAPI()


class TypicalResponse(BaseModel):
    msg: str
    data: int


@app.get("/add/{x}/{y}")
async def get_storage(x: int, y: int) -> TypicalResponse:
    return TypicalResponse(msg="Done!", data=x + y)
```

### Description

Open the automatic doc on this handler
See no output schema defined
I would like to use typing "-> *Type*" for annotate response model.
This is good for simple handlers.

### Wanted Solution

I would like to autodoc parse this typing same as "response_model=TypicalResponse"

### Wanted Code

```python
from fastapi import FastAPI

app = FastAPI()

# Same
class TypicalResponse(BaseModel):
    msg: str
    data: int


@app.get("/add/{x}/{y}")
async def get_storage(x: int, y: int) -> TypicalResponse:
    return TypicalResponse(msg="Done!", data=x + y)
```

### Alternatives

I use kwarg "response_model" as single alternative
But typing already defines the output model, necessary duplicate as result.

### Operating System

Linux

### Operating System Details

_No response_

### FastAPI Version

0.79.0

### Python Version

Python 3.9.0

### Additional Context

Without patch:
изображение

With patch:
изображение

I think second is better by avoid information duplication

Meanwhile, solution is quite simple.
One of the ways to implement (don't judge too much, that's first idea):

1. Open APIRouter class
2. Modify `self.add_api_route` by adding
изображение

May be dirty, so welcome to search another ways.

JarroVGIT:
Based on the discussion here:

Fixes https://github.com/tiangolo/fastapi/issues/5215

This would allow the developer to choose between using something like 
```python
@app.get("/", response_model=MyModel)
def root():
    pass
```
and 
```python
@app.get("/")
def root() -> MyModel:
    pass
```

The latter is more idiomatic. In this PR, the `reponse_model=` method takes precedence, as all documentation is in that style. That means if both are set, then the return type will be ignored.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,0 +1,27 @@
+from fastapi import FastAPI
+from pydantic import BaseModel
+
+
+app = FastAPI()
+
+class ResponseModelA(BaseModel):
+    a: str
+
+class ResponseModelB(BaseModel):
+    b: str
+
+@app.get("/a", response_model=ResponseModelA)
+def get_a_with_response_model_a():
+    return ResponseModelA(a="a")
+
+@app.get("/b", response_model=ResponseModelA)
+def get_b_with_response_model_a_and_return_model_b() -> ResponseModelB:
+    return ResponseModelA(a="a")
+
+@app.get("/c")
+def get_c_with_return_type_b() -> ResponseModelB:
+    return ResponseModelB(b="b")
+
+if __name__ == "__main__":
+    import uvicorn
+    uvicorn.run(app, host="0.0.0.0", port=8000)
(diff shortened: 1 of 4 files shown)


JarroVGIT:
This has been discussed and rejected in #101 :(

PR #5240, opened 2022-08-07
🔧 Update package metadata, drop support for Python 3.6, move build internals from Flit to Hatch

ofek:
## Background

Hello there! The Python packaging ecosystem has standardized on the interface for build backends ([PEP 517](https://peps.python.org/pep-0517/)/[PEP 660](https://peps.python.org/pep-0660/)) and the format for metadata declaration ([PEP 621](https://peps.python.org/pep-0621/)/[PEP 631](https://peps.python.org/pep-0631/)).

So, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄

The is a follow-up to https://github.com/samuelcolvin/pydantic/pull/4282

### Metadata

This implements [PEP 621](https://peps.python.org/pep-0621/) and removes Flit as a development dependency, allowing `pip` to be used directly which simplifies the contributing workflow and CI/CD. The build backend `hatchling` (of which I am a maintainer in the [PyPA](https://github.com/pypa)) is now used as that is the default in the official [Python packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/). Hatchling is available on all the major distribution channels such as [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), [Arch Linux](https://archlinux.org/packages/community/any/python-hatchling/), [conda-forge](https://anaconda.org/conda-forge/hatchling), [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/hatchling/default.nix), [Alpine Linux](https://github.com/alpinelinux/aports/blob/master/community/py3-hatchling/APKBUILD), [FreeBSD](https://github.com/freebsd/freebsd-ports/blob/main/devel/py-hatchling/Makefile), [Gentoo Linux](https://github.com/gentoo/gentoo/tree/master/dev-python/hatchling), [MacPorts](https://github.com/macports/macports-ports/blob/master/python/py-hatchling/Portfile), [OpenEmbedded](https://github.com/openembedded/openembedded-core/commit/846e806181f1349be29cbce78c5041735dfd7e6f), [Spack](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-hatchling/package.py), etc.

### Python support

The earliest supported Python 3 version of Hatchling is 3.7, therefore I've also set that as the minimum here. Python 3.6 is EOL and other build backends like `flit-core` and `setuptools` also dropped support. Python 3.6 accounts for 3-4% of downloads in the last month: https://pypistats.org/packages/fastapi

If support is still desired for some reason, I can keep the existing cap and only require 3.7+ in CI/CD.

### Documentation

I updated the contributing docs, which accounts for most of the diff. Luckily, they were written in such a way that going from `flit` to `pip` didn't alter semantics for any language!

## Future

- We now have the ability to ship optional wheels compiled with Mypyc, as Black [recently did](https://ichard26.github.io/blog/2022/05/31/compiling-black-with-mypyc-part-1/), to be even faster in [benchmarks](https://github.com/TechEmpower/FrameworkBenchmarks/pull/7178). Though as @tiangolo mentioned offline, it may be best to wait for the great work @samuelcolvin is doing on Pydantic v2 to be over to be sure we don't cause any disruptions in their test suite.
- We could source the version from Git


ofek:
Ah looks like the 3.6 job is hard coded in this repo's GitHub settings as required.

pyproject.toml
```diff
@@ -1,12 +1,16 @@
...
-module = "fastapi"
-author = "Sebastián Ramírez"
-author-email = "tiangolo@gmail.com"
-home-page = "https://github.com/tiangolo/fastapi"
+[project]
+name = "fastapi"
+description = "FastAPI framework, high performance, easy to learn, fast to code, ready for production"
+readme = "README.md"
+requires-python = ">=3.7"
+license = "MIT"
```
tiangolo, line 10: I get an error in the editor telling me that this should be a table, the spec seems to say the same: https://peps.python.org/pep-0621/#license

Is there any other spec that overrides that and allows using the string description?

samuelcolvin: ye, I think it does.

see [here](https://github.com/samuelcolvin/dirty-equals/blob/5cf23ce233ec8804067bf8ac07272d16cd7cf1b9/pyproject.toml#L12) which as:
```toml
license = {file = 'LICENSE'}
```

ofek: PEP 639 is already supported by Hatchling https://peps.python.org/pep-0639/

tiangolo: Amazing, thanks!

.github/workflows/publish.yml
```diff
@@ -17,23 +17,21 @@ jobs:
...
+        run: pip install build
+      - name: Build distribution
+        run: python -m build
       - name: Publish
-        env:
-          FLIT_USERNAME: ${{ secrets.FLIT_USERNAME }}
-          FLIT_PASSWORD: ${{ secrets.FLIT_PASSWORD }}
-        run: bash scripts/publish.sh
+        uses: pypa/gh-action-pypi-publish@v1.5.1
+        with:
```
ofek, line 33:
```suggestion
        with:
          user: __token__
```

tiangolo: Is it needed? Then maybe the docs in the GitHub Action need an update too?

https://github.com/pypa/gh-action-pypi-publish#usage

tiangolo:
Awesome job, thanks a lot @ofek! FastAPI is now built with Hatch. :egg: :snake: :sunglasses: 

Also, Python 3.6 is now officially **NOT** supported. :fire:

.github/workflows/publish.yml
ofek: Ah, guess not!

PR #5271, opened 2022-08-14
Update tutorial005_py39.py

huncut2016:
In line 146 the `response_mode` is specified as `Token`.  When we return a `dict` instead of a `Token` object, pydantic raise a `Validation error`


tiangolo:
Hey there! The `response_model` is used to convert and validate the data sent from the function, so the function should be able to return a dict and it would be validated and converted with the `Token` model. The code example is intentional.

Is it not working for you? Could you write an example of what is not working?

PR #5273, opened 2022-08-16
Give the ability to set any route class when adding a third-party router

ikharyn:


iudeen:
Any context on why this is needed and what would be the impact?

ilyaTT:
> Any context on why this is needed and what would be the impact?

The routing of our project has a nested structure.
At the lower level, a router with an overridden ```__init__``` was used. However, when importing, we need to use a different ```route``` class.
Because everywhere in the project it is possible to pass the ```route_class_override``` parameter, then it would be right to allow it here as well.

YuriiMotov (changes requested):
@ilyaTT, thanks for contributing to FastAPI!

As I understand it, the idea is to provide the way to write
```python
router = APIRouter()
main_router.include_router(router=router, route_class_override=MyAPIRoute)
```

instead of

```python
router = APIRouter(route_class=MyAPIRoute)
main_router.include_router(router=router)
```

full code in details

<details>

```python
from typing import Callable

from fastapi import APIRouter, FastAPI, Request, Response

class MyAPIRoute(APIRoute):
    def get_route_handler(self) -> Callable:
        original_route_handler = super().get_route_handler()
        async def custom_route_handler(request: Request) -> Response:
            print("MyAPIRoute called")
            return await original_route_handler(request)
        return custom_route_handler

# router = APIRouter(route_class=MyAPIRoute)
router = APIRouter()

@router.get("/")
async def get_():
    pass

main_router = APIRouter()
# main_router.include_router(router=router)
main_router.include_router(router=router, route_class_override=MyAPIRoute)

app = FastAPI()
app.include_router(router=main_router)
```

</details>

Then, why don't we add the same `route_class_override` to `FastAPI` class?
Also, we need to add description for `route_class_override` parameter and tests.

PR #5307, opened 2022-08-26
📝 Add note about Python 3.10 `X | Y` operator in explanation about Response Models

MendyLanda:


tiangolo:
Cool, thank you @MendyLanda! :coffee:

PR #5310, opened 2022-08-27
Changing OAUTH2/OIDC dependencies response status code when Authorization header missing

felixilgatto:
Hello humans,

I recently have to use the OAuth and OpenID Connect dependencies and was not happy with the 403 status code while I was requesting my API without Header Authorization Header.
So I re-read the status code official documentation :

> The HyperText Transfer Protocol (HTTP) 401 Unauthorized response status code indicates that the client request has not been completed because **it lacks valid authentication credentials** for the requested resource. 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

> The HTTP 403 Forbidden response status code indicates that **the server understands the request** but refuses to authorize it.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403

My understanding of these definition, and correct me if I'm wrong, is that if the Authorization header with correct scheme but an invalid token is set will be resulting in a 403 response, otherwise if no Authorization header is set response status code will be 401.

Please let me know if I'm thinking right.

Thank you,


svlandeg:
Thanks for the contribution @felixilgatto  🙏. I'm closing this one in favour of https://github.com/fastapi/fastapi/pull/13786, which is more extensive and should (hopefully) address this issue once and for all.

PR #5313, opened 2022-08-29
Specificy a minimum version for flit

woutervb:
If we leave the flit version unbound and by change an 1.x or 2.x version
is already installed, failures will happen, as an attempt will be done
to import flit_core.inifile, which doesn't exist, as there is no
flit_core in onder versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit"]
+requires = ["flit >=3.2"]
 build-backend = "flit.buildapi"
 
 [tool.flit.metadata]


iudeen:
I think once FastAPI adheres to new build standard defined by PyPa, this PR https://github.com/tiangolo/fastapi/pull/5240, this won't be relevant.

woutervb:
> I think once FastAPI adheres to new build standard defined by PyPa, this PR #5240, this won't be relevant.

True, except this currently breaks packaging in an offline environment, any idea when PR #5240 will land in an release?

iudeen:
I don't see any blocker except for current support for Python36 at FastAPI side. Once the maintainers decide to drop 3.6 support, I hope the above mentioned PR will be merged.

tiangolo:
Thanks! As FastAPI is now built with Hatch, this is no longer necessary. So I'll close it. But thanks for the effort! :coffee:

PR #5323, opened 2022-08-30
Simplify code of `http_exception_handler`

PipeKnight:
Looking to `JSONResponse.__init__()` function, you can see that `headers` parameter is `None` by default, so it doesn't matter do we pass it here as `None` or do not pass at all. That's why we can safely remove this if-else expression and slightly simplify the code


yezz123 (approved):

BilalAlpaslan (approved):

iudeen (approved):
LGTM

mbroton (approved):

stlucasgarcia (approved):

ramnes:
LGTM but you probably can close it, it doesn't make sense anymore with the latest implementation of `http_exception_handler` from #5365:

https://github.com/tiangolo/fastapi/blob/97f04ab58c2c047d6d2ff7d803890a90190109c9/fastapi/exception_handlers.py#L10-L16

tiangolo:
Thanks @PipeKnight! 🍰 

And thanks everyone for the reviews and comments.

It seems this was covered in a recent change, so I'll close this one. But thanks for the effort! ☕

PR #5328, opened 2022-09-01
Use flit-core as build-system

sscherfke:
Build functionality has been moved into flit-core which makes repackaging (e.g. for Conda or Linux distros) easier.


iudeen:
The new standard of build system is using hatch. There is a PR for this already https://github.com/tiangolo/fastapi/pull/5240

PR #5332, opened 2022-09-01
Fix OIDC response code when authorization header is missing

skion:
The OIDC dependency currently returns a 403 response if an `Authorization` header is missing.

Using a 401 instead of 403 aligns with the HTTP standard when authentication is missing and with the [existing OAuth2 dependency](https://github.com/tiangolo/fastapi/blob/b9d7f86743eb9ba2805ae0c8b99a30a3a47a58bb/fastapi/security/oauth2.py#L207-L210).


YuriiMotov (approved):
The issue is still relevant to current version of FastAPI (0.110.0).

It was said in [this comment](https://github.com/tiangolo/fastapi/pull/2120#issuecomment-1236165134) to the similar issue, that it should be fixed, but we have to make sure that new solution follows the specification. 

OIDC follows the OAuth2 specification here. And in the current implementation of `OAuth2AuthorizationCodeBearer` and `OAuth2PasswordBearer` `status_code` and `headers` are exactly the same.

skion:
Thanks for the suggestion @YuriiMotov

Bladieblah:
Hello! Any idea when this will get merged?

skion:
This [seems](https://github.com/fastapi/fastapi/blob/master/fastapi/security/open_id_connect_url.py) still an issue in mainline as of today... cc @YuriiMotov

svlandeg:
Thanks for the contribution @skion 🙏. I'm closing this one in favour of https://github.com/fastapi/fastapi/pull/13786, which is more extensive and should (hopefully) address this issue once and for all.

PR #5350, opened 2022-09-05
♻️ Refactor code to support new HTTPX-based `TestClient`

tiangolo:
♻️ Refactor code to support new HTTPX-based `TestClient`

This was done locally using @Kludex's branch of Starlette migrating the `TestClient` to HTTPX.

These are the changes that will be needed in FastAPI after that is merged.

One import in `applications.py` will be needed either way after upgrading Starlette.

The rest of the changes all make sense and seem sensible. The code would probably make more sense after this. And although some users could need to update their test suite, their code would also look more sensible after the needed refactors.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -33,9 +33,10 @@ from fastapi.types import DecoratedCallable
 from fastapi.utils import generate_unique_id
 from starlette.applications import Starlette
 from starlette.datastructures import State
-from starlette.exceptions import ExceptionMiddleware, HTTPException
+from starlette.exceptions import HTTPException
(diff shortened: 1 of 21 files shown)


Kludex:
[bump-testclient](https://github.com/Kludex/bump-testclient) makes almost all the changes you did.

What it doesn't change still:
1. Move `cookies` to `TestClient()`.
2. Replace `data` as list of tuples to dictionary.

I'm working on the `2`. The `1` will not be supported by the transformer. JFYK.

tiangolo:
This is no longer needed, it was covered in another PR :tada: :fire:

PR #5360, opened 2022-09-07
🐛 Fix empty reponse body when default `status_code` is empty but the a `Response` parameter with `response.status_code` is set

tmeckel:
This PR contains a fix to a bug which cleared the reponse content if the status code is changed via a [`Resource` parameter described in the documentation](https://fastapi.tiangolo.com/advanced/response-change-status-code/#use-a-response-parameter). 

Sample code which creates an empty response body:

```python
@app.delete(
    "/{id}",
    responses={
        204: {"model": None, "description": "No Content"},
    },
    status_code=204,
)
async def delete_deployment(
    id: str = Path(None),
    response: Response = None,
) -> None:
    response.status_code = 400
    return {"msg": "Status overwritten"}
```


iudeen (commented):
Did you check if proper response code is reflecting in the api doc?

In Doc, it still shows `204`. This PR would be complete if that is handled as well, IMHO. 
image

tmeckel:
@iudeen can you elaborate what you mean by "...__proper response code is reflecting in the api doc?__" As you can see from the referenced link to documentation it's supported to override the response code by setting the code on a `Reponse` parameter object to the API method.

iudeen:
> can you elaborate what you mean by "...**proper response code is reflecting in the api doc?**" 

I was meaning to say if we change response codes, it should reflect in Swagger (or Redoc) docs as well. But I now realize this behavior was not there before as well and seems its out of scope for this PR. 

LGTM! 👍

tiangolo:
Awesome, thank you @tmeckel! :rocket: :cake: 

I updated the tests a bit to simplify them and cover OpenAPI (what can be supported).

@iudeen thanks for the interest, but the custom response code here is done only at runtime, there's no way to put that in OpenAPI, so that wouldn't apply here.

Thanks! :coffee:

PR #5365, opened 2022-09-08
 🐛Fix `RuntimeError` raised when `HTTPException` has a status code with no content

iudeen:
Fixes/related to https://github.com/tiangolo/fastapi/issues/4949#issuecomment-1241143806

Please advice if this approach is correct one.

Suggestions welcome!

### DIFF
--- a/tests/test_starlette_exception.py
+++ b/tests/test_starlette_exception.py
@@ -18,6 +18,18 @@ async def read_item(item_id: str):
     return {"item": items[item_id]}
 
 
+@app.get("/http-no-body-statuscode-exception")
+async def no_body_status_code_exception():
+    raise HTTPException(
+        status_code=204,
+    )
+
+
+@app.get("/http-no-body-statuscode-with-detail-exception")
+async def no_body_status_code_with_detail_exception():
+    raise HTTPException(status_code=204, detail="I should not make it!")
+
+
 @app.get("/starlette-items/{item_id}")
 async def read_starlette_item(item_id: str):
     if item_id not in items:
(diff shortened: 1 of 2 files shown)


tests/test_starlette_exception.py
```diff
@@ -18,6 +18,18 @@ async def read_item(item_id: str):
     return {"item": items[item_id]}
 
 
+@app.get("/http-no-body-statuscode-exception")
+async def no_body_status_code_exception():
+    raise HTTPException(
+        status_code=204,
+    )
```
Kludex, line 25:
```suggestion
    raise HTTPException(status_code=204)
```

Kludex (approved):
Yep, same idea as what we have in Starlette: https://github.com/encode/starlette/blob/bc61505faef15b673bf4bf65db4927daa354d6b8/starlette/middleware/exceptions.py#L99-L104

iudeen:
> Yep, same idea as what we have in Starlette: https://github.com/encode/starlette/blob/bc61505faef15b673bf4bf65db4927daa354d6b8/starlette/middleware/exceptions.py#L99-L104

The way I was able to quickly find the bug was adding debug points to ExceptionMiddleware while running the app using FastAPI and Starlette. 

And that code was from Starlette! (Sorry, should have mentioned)

BilalAlpaslan (approved):

JarroVGIT (approved):
LGTM

tiangolo:
Awesome, thanks for the fix @iudeen! :cake: :rocket: 

And thanks for the reviews and comments @Kludex, @JarroVGIT, @BilalAlpaslan :bow: :coffee:

PR #5371, opened 2022-09-11
Fix mount asgi app on router

BilalAlpaslan:
override mount method for empty path


BilalAlpaslan:
more clean version  https://github.com/tiangolo/fastapi/pull/5373

PR #5373, opened 2022-09-11
🐛 Fix mount subapp with APIRouter

BilalAlpaslan:


iudeen (approved):
LGTM

fastapi/routing.py
```diff
@@ -782,11 +782,19 @@ def include_router(
...
+                self.mount(prefix + route.path, route.app, name=route.name)
         for handler in router.on_startup:
             self.add_event_handler("startup", handler)
         for handler in router.on_shutdown:
             self.add_event_handler("shutdown", handler)
 
+    def mount(self, path: str, app: ASGIApp, name: Optional[str] = None) -> None:
+        route = routing.Mount(path, app=app, name=name)
+        if route.__getattribute__("path") == "":
+            route.__setattr__("path", "/")
```
mbroton, line 795: These are magic methods, IMO they shouldn't be called explicitly as the inbuilt functions that use the magic methods can perform additional checks.
If there's no particular reason to use magic method directly, It would be better to use `setattr`/`getattr`.

BilalAlpaslan: yes, you're right. I have e three solution for that and @Kludex i need your opinion
1- solve this issue in fastAPI:  override mount function and change the path
        if route.path == "":
            route.path = "/"
2-override Mount class and change this line: `self.path = path.rstrip("/")`
3-change this code on starlette
or maybe anyone has a better idea

Kludex: What is the problem here? 😅

BilalAlpaslan: if mount asgi app on APIRouter  Starlette router.Mount object return ' ' for path='/'. `self.path = path.rstrip("/")` cause this and fastapi raise `Prefix and path cannot be both empty`

Kludex: We should probably solve https://github.com/encode/starlette/issues/1336 before doing something in FastAPI 👀

Kludex:
I'm handling this on https://github.com/encode/starlette/pull/2352.

tiangolo:
It seems this was related to the ASGI `root_path` stuff, and handled in Starlette, right? Then I'm gonna close this one. :nerd_face: :coffee: 

If you still have issues, please create a discussion with a minimal self contained example that I can copy paste and run to understand the use case better. :nerd_face:

PR #5375, opened 2022-09-11
save the cookies

bearllm:


tiangolo:
Thanks for the interest! That's a very funny title. :sweat_smile: :cookie: 

I'm not fully sure what's the intention with this PR, let's start in a Discussion with a minimal self contained example that I can copy paste to understand your use case.

PR #5377, opened 2022-09-11
📝 Update `SECURITY.md`

Kludex:


tiangolo:
Yep, makes sense, thanks!

PR #3432, opened 2021-06-29
✨ Add allow disabling `redirect_slashes` at the FastAPI app level

cyberlis:
Right now `redirect_slashes` parameter in sub routers created with `APIRouter` ignored, because only main router's `redirect_slashes` inside FastAPI class is taken into account. FastAPI class creates new router without `redirect_slashes` argument and that's why router always uses default value `redirect_slashes=True` and always redirects on trailing slashes.
https://github.com/tiangolo/fastapi/blob/master/fastapi/applications.py#L69

Here is how to reproduce this problem
```python
from fastapi import FastAPI
from fastapi.routing import APIRouter

router = APIRouter(redirect_slashes=False)

@router.get('/hello/')
def hello_page() -> str:
    return 'Hello, World!'

app = FastAPI()
app.include_router(router)
```

And if you request page `http://127.0.0.1:8080/hello` (without trailing slash) you will be redirected even with `redirect_slashes=False`

```bash
uvicorn main:app
INFO:     Started server process [96046]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     127.0.0.1:57330 - "GET /hello HTTP/1.1" 307 Temporary Redirect
INFO:     127.0.0.1:57330 - "GET /hello/ HTTP/1.1" 200 OK
```

I added `redirect_slashes` to `__init__` method of FastAPI class and passed it to router's constructor.
That way I can change trailing slashes redirect behaviour globally for application instance.

PR #5392, opened 2022-09-15
Pass in redirect_slashes

lieryan:
Currently, `redirect_slashes` option that is used to turn off automatic redirection of trailing slash, which is available from Starlette but doesn't work in FastAPI.


yezz123 (commented):
for what this needs to be true, I guess that could be better if the user have the responsibility on what he want to pass.

> trailing slashes are purely conventional, and Google does not care whether you use them; as long as you're consistent.

BilalAlpaslan (approved):

lieryan:
@yezz123 generally, for websites it is often sensible to automatically redirect slashes and non-slash URL as identical since users might have copied or typed the URLs by hand and forgot the last slash; but API URLs are generally handled by machines, so that usability consideration doesn't apply. Moreoever, having two different endpoints for your API can silently cause issues if the API client used the wrong URL, and caching, ETag, etc becomes problematic. So, generally for APIs, I would always recommend just turning off slash redirect, so that API clients breaks hard and gets fixed rather than just silently doing the wrong thing.

JarroVGIT:
Interesting, I am curious to know how @tiangolo envisioned this working. It seems that even if you create your own `APIRouter` with the setting to false, it would be overridden by the setting in the main `APIRouter` when you call `include_router()`. That is strange. I would be in favour of this PR.

JarroVGIT (approved):
This is however a duplicate of #3432 (and that PR did provide additional tests).

tiangolo:
Thanks everyone for the discussion here! ☕ 

And thanks @JarroVGIT for the pointer to the other PR. 🤓 

Is this is already solved in that PR, I'll go ahead and close this one.

PR #5427, opened 2022-09-23
✨Add support for removing responses by passing empty dict

staffluck:
✨Add support for removing responses by passing empty dict

e.g
```
from fastapi import FastAPI

class Test(BaseModel):
    test_1: str
    test_2: int

app = FastAPI()

@app.post("/test", response_model=Test, responses={422: {}})
async def test_1(test_model: Test):
    return {"test_1": "123", "test_2": 123}

```
=> 
fastapi


yezz123 (approved):

YuriiMotov:
@staffluck, thanks for you interest in FastAPI!

Is this supposed to be used to disable automatically included `422` response only?
If yes, I think that `disable_validation_response` configuration parameter to `FastAPI()` would be better option.
Could anybody highlight if there are any other use-cases?

PR #5428, opened 2022-09-23
🐛 Set nullable = True on required Optional[T] OpenAPI properties

Shadow53:
Pydantic has an interesting behavior where the following field declaration is considered required but also allowed to be `None`:

```python
some_field: Optional[str] = Field(..., description="The ... means it has no default")
```

The current behavior of FastAPI is to treat the field as non-nullable required, i.e.

```json
{
    "properties": {
        "some_field": {
            "type": "string"
        }
    },
    "required": ["some_field"]
}
```

Clients generated from this will fail with an error if the service ever returns `some_field: null`.

This PR adds some code to check if a field is both required and nullable, and change it to the following:

```json
{
    "properties": {
        "some_field": {
            "nullable": true,
            "type": "string"
        }
    },
    "required": ["some_field"]
}
```

This is my first time contributing here, so please let me know if there is anything I should do differently or improve.


tiangolo:
Thanks @Shadow53! This would actually belong in Pydantic. And now with Pydantic v2, using OpenAPI 3.1.0 and the latest JSON Schema, it is handled properly.

You can probably just upgrade Pydantic to v2 and you'll get the new and more correct schema.

Given that, I'll close this one, but thanks for the effort! 🍰

Issue #3937, opened 2021-09-23
 Invalid args for response field! 


kenjoel:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
class PyObjectId(ObjectId):
    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    def validate(cls, v):
        if not ObjectId.is_valid(v):
            raise ValueError("Invalid objectid")
        return ObjectId(v)

    @classmethod
    def __modify_schema__(cls, field_schema):
        field_schema.update(type="string")


class Item(BaseModel):
    id: PyObjectId = Field(default_factory=PyObjectId, alias="_id")
    name: str = None
    quantity: str = None
    price: int = None
    available: bool = None

    class Config:
        BaseConfig.arbitrary_types_allowed = True
        allow_population_by_field_name = True
        json_encoders = {ObjectId: str}



def all_items():
    expected_items = db["items"].findAll()
    return expected_items


app = FastAPI()
BaseConfig.arbitrary_types_allowed = True

@app.get("get_item", response_model=Item)
async def get_all_items():
    data = await get_all_items()
    return data
```

### Description

Guys, I have really tried to find a solution to this problem and I have done all that was recommended in github issues to no avail. I am still getting the same error. Where am I going wrong?

### Operating System

Linux

### FastAPI Version

0.68.1

### Python Version

3.9

### Additional Context

Please help

Robin-Bonnin:
Hello! 

I haven't tested but it seems that you need to wait for the DB to get the response.
eg

    
      def all_items():
          return await db["items"].findAll()

kenjoel:
Hi, I am doing so on this method 

@app.get("get_item", response_model=Item)
async def get_all_items():
[...]

Issue #2382, opened 2020-11-20
Response model doesn't allow models with `allow_arbitrary_types`


fisheye36:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Example - not specifying `response_model`

Creating a `Pydantic` model that has a field with an arbitrary class will raise `RuntimeError` stating that there is no validator for class `ArbitraryClass`:

```Python
from fastapi import FastAPI
from pydantic import BaseModel

app = FastAPI()


class ArbitraryClass:

    def __init__(self, attr: str) -> None:
        self.attr = attr


class SomeModel(BaseModel):

    name: str = 'SomeModel'
    arbitrary_field: ArbitraryClass

    class Config:
        arbitrary_types_allowed = True


@app.get('/')
async def return_model_with_arbitrary_type():
    model = SomeModel(arbitrary_field=ArbitraryClass('attr_value'))
    return model
```

### Results of running example without `response_model` specified

This works fine, since `Pydantic` allows creation of models where fields have arbitrary types when we use `arbitrary_types_allowed` in `Config`. The returned JSON is also nicely formatted, as`attribute: value` pair:

```json
{
  "name": "SomeModel",
  "arbitrary_field": {
    "attr": "attr_value"
  }
}
```

### Example - specifying `response_model`

When we try to specify that the response model for this endpoint is `SomeModel`:

```Python
@app.get('/', response_model=SomeModel)
async def return_model_with_arbitrary_type():
    model = SomeModel(arbitrary_field=ArbitraryClass('attr_value'))
    return model
```

We get an error:

```
Traceback (most recent call last):
  File "/Users/fisheye36/Projects/fastapi-issue/.env/lib/python3.8/site-packages/fastapi/utils.py", line 65, in create_response_field
    return response_field(field_info=field_info)
[...]

Issue #1186, opened 2020-03-30
[QUESTION] Response model with pydantic's arbitrary_types_allowed not working


datasatanic:
I use pydantic model with custom type.  Set in config arbitrary_types_allowed=True. 
Without setting a parameter "response_model" it's work. But when I set this model in param:
```
fastapi.exceptions.FastAPIError: Invalid args for response field! Hint: check that <class 'Foo'> is a valid pydantic field type 
```
Pydantic class definition example:
```
import Foo

class Model(BaseModel):
    id: str
    custom: Optional[Foo] = Field(None)
    status: str = None
    fail_message: Optional[str] = None
    params: Dict[str, Any] = None

    @validator('custom')
    def deserialize_model(cls, v):
        return Foo.loads(v)

    class Config:
        arbitrary_types_allowed = True
        json_encoders = {Foo: lambda v: v.dumps()}
```

andreipopovici:
I see the same issue with an `Arrow` type returned from the SQLAlchemy model. `arbitrary_types_allowed` is not working when using the pydantic model in the `response_model`.

tiangolo:
Please create a small self-contained example with a minimal FastAPI app that shows the error and what you would expect to happen, that way I can understand what's happening and what you want to achieve.

bcb:
@tiangolo I think the config is getting lost in the `create_cloned_field` function.

```python
new_field = create_response_field(name=field.name, type_=use_type)
```

Shouldn't it also pass `model_config`?

bcb:
I ended up using a custom Pydantic field validator, instead of using `arbitrary_types_allowed`:

```python
class AssociationList(_AssociationList):
    """Sqlalchemy AssociationList"""

    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    def validate(cls, v):
        if not isinstance(v, _AssociationList):
            raise TypeError("AssociationList required")
        return [UserOut.from_orm(u) for u in v]

class TeamOut(TeamBase):
    members: AssociationList
    ...
```

eco-monk:
> Same problem here:
> 
> ```python
> from pydantic import BaseModel
> from fastapi import FastAPI
> 
> 
> class Pair:
>     def __init__(self, x: int, y: int):
>         self.x: int = x
>         self.y: int = y
> 
> 
> class M(BaseModel):
>     pair: Pair
>     other: int
> 
>     class Config:
>         arbitrary_types_allowed = True
> 
> 
[...]

PR #5429, opened 2022-09-24
🐛 Fix losing model config in `create_cloned_field`

daniil-berg:
## Summary

Until now, the function `utils.create_cloned_field` called `utils.create_response_field` [without passing along the model config](https://github.com/tiangolo/fastapi/blob/0.85.0/fastapi/utils.py#L107) (among other arguments), which caused an error in the latter function, whenever an [arbitrary type](https://pydantic-docs.helpmanual.io/usage/types/#arbitrary-types-allowed) was used for the field, even if the appropriate model configuration was done.

The workaround so far was to monkey-patch the `pydantic.BaseConfig`. With these small changes, the fields will just work as expected.

## Reproducible example

Previously, running the following piece of code passed the test-block without a problem, but threw an error at the route decorator:

<details>
<summary>Click here</summary>

```python
from typing import Union

from fastapi import FastAPI


class Foo:
    def __init__(self, value: int) -> None:
        self.value = value

    def __eq__(self, other: object) -> bool:
        if not isinstance(other, Foo):
            return NotImplemented
        return self.value == other.value

    def __repr__(self) -> str:
        return f"Foo(value={self.value})"

    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    def validate(cls, v: Union[int, Foo]) -> Foo:
        if isinstance(v, Foo):
            return v
        if isinstance(v, int):
            return Foo(v)
        raise TypeError("Invalid Foo type")

    @classmethod
    def __modify_schema__(cls, field_schema: dict) -> None:
        field_schema.update(
            type="number",
            examples=[-1, 0, 1024],
        )


class MyModel(BaseModel):
    foo: Foo

    class Config:
        arbitrary_types_allowed = True
        json_encoders = {
            Foo: lambda foo: foo.value,
        }


# Test:
_obj = MyModel(foo=Foo(123))
_parsed_obj = MyModel.parse_raw(_obj.json())
assert _obj == _parsed_obj
MyModel.schema()


api = FastAPI()


@api.get("/model/", response_model=MyModel)
def example_route():
    return MyModel(foo=Foo(123))
```

The error was documented in the issues listed below.

</details>

## Fixes

- #1186 
- #2382 
- #3937

## Bonus

That second commit is just to finally scratch that itch we all feel at that 99% coverage every time... Or is it just me? 😛


iudeen (approved):

yezz123 (approved):

dw-lucas (approved):

stlucasgarcia (approved):

tiangolo:
Thank you! It seems this was covered in another PR, so I'm gonna close this one now. But thanks for the effort! :cake:

Issue #2481, opened 2020-12-07
How to override global dependencies on a single endpoint?


PatrikTrelsmo-izettle:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to one of:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

Here's a self-contained, [minimal, reproducible, example](https://stackoverflow.com/help/minimal-reproducible-example) with my use case, copied from https://fastapi.tiangolo.com/tutorial/dependencies/global-dependencies/:

```Python
from fastapi import Depends, FastAPI, Header, HTTPException


async def verify_token(x_token: str = Header(...)):
    if x_token != "fake-super-secret-token":
        raise HTTPException(status_code=400, detail="X-Token header invalid")


async def verify_key(x_key: str = Header(...)):
    if x_key != "fake-super-secret-key":
        raise HTTPException(status_code=400, detail="X-Key header invalid")
    return x_key


app = FastAPI(dependencies=[Depends(verify_token), Depends(verify_key)])


@app.get("/items/")
async def read_items():
    return [{"item": "Portal Gun"}, {"item": "Plumbus"}]


@app.get("/users/")
async def read_users():
    return [{"username": "Rick"}, {"username": "Morty"}]

# This function shouldn't be authenticated because it is a health check endpoint
@app.get("/healthcheck")
async def healthcheck():
    return "ok"
```

### Description

Is there a way to "skip" or override global dependencies on a single endpoint? I really like the new global dependencies since it makes it easy to enforce authentication by design rather than remembering to add it to every single endpoint. 

[...]

PR #5433, opened 2022-09-26
Allow a _least privilege_ model of automatic dependencies by "disabling" dependencies for individulal routers or endpoints.

kristjanvalur:
Issue #2481 Mentions how it is difficult to remove a global dependency on a subset of endpoint.  The proposed solution in that defect is to add global depdencies on individual `APIRouter` instances and combine a set of those into the app.

This PR proposes a complimentary solution.  Anywhere where a `Depends(foo)` or `Security(bar)` is added as an item in a `dependencies` keyword argument, it can receive a `disable=True` keyword argument.

Once the final parameterless dependencies are gathered and added on a route, these are then processes and collapsed.  This allows an `APIRouter` to disable a global dependency which has already been added on an app object.

This allows a *least privilege* approach to API design, where nothing is allowed by default unless explicitly allowed on a case by case basis:  The `app` object can have a global dependency which requires **all** scopes, and then individual endpoints, for example those that deal with authentication, or other less privileged actions, can remove the more stringent scopes to allow access.

Additionally, multiple stacked calls to the same `Security` dependency are collapsed, merging all the provided *scopes* into a single list.  This allows the creation of virtual scopes, for example, prefixing a scope with a "-" do disable it, handling the processing of the all the scopes to a single call.  This is yet-another way disabling a scope requirement.

This is currently an undocumented feature request / proposal, containing tests.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -1,5 +1,6 @@
 import dataclasses
 import inspect
+from collections import defaultdict
 from contextlib import contextmanager
 from copy import deepcopy
 from typing import (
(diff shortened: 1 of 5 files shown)


tiangolo:
Thanks for the interest and effort @kristjanvalur ! ☕

This would add some significant complexity, in particular to explain it to new developers. And it would mean that adding a dependency somewhere would no longer guarantee that it would be executed everywhere down the line.

Given that I'll pass on this one, but thanks for the interest!

tiangolo:
@AdrianSacristanPB nope. If we think about it, a least privilege model would really mean that you would add all the things that block and restrict at the top level in your app. Even if they are only related to a small section of it. And then you would have to enable access to most of those for most of the other unrelated parts.

It would also mean that to know if a section in your code is covered by some protection, you would have to check the whole chain of included routers to make sure none is disabling it. That's kinda linear complexity (O(n)) to verify if something would apply or not. This itself could be a security risk, as people could assume that if something was applied somewhere up the chain, then it would also apply here, but that's no longer the case.

So I prefer to have related things and restrictions closer to where they apply, with the related code close together.

Now, you can probably hack around using scopes, and relax permissions with scopes. I'm pretty sure it would be a quite complicated implementation (at least mentally), but maybe you could make it work.

kristjanvalur:
The case I'm trying to solve is really:  "Here I want _almost_ the same restrictions as the rest of the app, but not quite".  Doing it currently is quite inconvenient, with multiple routers.

The _reverse_ solution is to set **no** rules at the top, and make sure that you apply them at **every** router/endpoint.  that is too an O(n) problem and you have no guarantees there either, because you have to make sure that you don't forget to apply the rules individually.

The sort of middle ground is to have multiple routers, each with a different set of rules applied to them and join them in a top app with no rules.  This works, but it can become messy and complex.   

A different way to achieve the same, or similar, results is by using `only Security` and make sure to `extend` the `scopes` when passing down the inheritance chain.  This PR can be modified to only join up Security scopes.  This way, the application itself can decide how to interpret added scopes.  An implementor can, for example, interpret scopes prefixed with '-' as disabling them in the security _dependency_.  Please also see the #5622 which merges parameterless `Securitiy` dependences with parameter dependencies.

An additional benefit of merging Security scopes is that it reduces calling overhead in the endpoint.  if you specify `Security(foo, scopes=["a"])` at the top, and `Security(foo, scopes=["b"])` lower down, the endpoint receives a single invocation of `foo` with `scopes==["a", "b"]` rather than two separate calls.   

I can submit an alternate PR which only does that.

PR #5450, opened 2022-10-02
Added Dockerfile and Docker Compose Support

Louden7:
- Created and added a Dockerfile to allow users to build a docker image from an officially supported python3-alpine base.

- Created and added a docker-compose.yml file to allow users to utilize docker compose to build or pull and run the FastAPI container with 1 command.

- Updated README.md to include docker/docker compose install instructions and how to build or pull the FastAPI image and run a container using docker and also docker-compose.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+# Create image from lightweight python3-alpine (https://hub.docker.com/_/python).
+FROM python:3-alpine
+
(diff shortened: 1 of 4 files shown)


tiangolo:
Thanks for the interest, but Docker is already covered here to the extent it's necessary for FastAPI: https://fastapi.tiangolo.com/deployment/docker/

Issue #5451, opened 2022-10-02
Dependecies overrides are not taken into account into the apenapi document


lmignon:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from fastapi import Depends, FastAPI, Header
from fastapi.security import OAuth2PasswordBearer
from fastapi.testclient import TestClient

app = FastAPI()


def get_user_id() -> int:
    pass


def get_user(user_id=Depends(get_user_id)) -> str:
    return f"{user_id}"


oauth2_scheme = OAuth2PasswordBearer(tokenUrl="token")


def get_user_id_from_auth_override(token: str = Depends(oauth2_scheme)) -> int:
    return 1


def get_user_id_from_header_override(user_id: int = Header()) -> int:
    return user_id


@app.get("/user")
def read_user(
    user: str = Depends(get_user),
):
    return user


client = TestClient(app)


override_with_security_schema = {
    "components": {
        "securitySchemes": {
            "OAuth2PasswordBearer": {
                "flows": {"password": {"scopes": {}, "tokenUrl": "token"}},
                "type": "oauth2",
            }
        }
    },
    "info": {"title": "FastAPI", "version": "0.1.0"},
    "openapi": "3.0.2",
    "paths": {
        "/user": {
            "get": {
                "operationId": "read_user_user_get",
                "responses": {
                    "200": {
                        "content": {"application/json": {"schema": {}}},
                        "description": "Successful " "Response",
                    }
                },
                "security": [{"OAuth2PasswordBearer": []}],
                "summary": "Read User",
            }
        }
    },
}

app.dependency_overrides[get_user_id] = get_user_id_from_auth_override
app.openapi_schema = None
response = client.get("/openapi.json")
assert response.status_code == 200, response.text
[...]

PR #5452, opened 2022-10-02
Fix openapi document with dependencies override (#5451)

lmignon:
refs #5451

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -161,7 +161,21 @@ if PYDANTIC_V2:
         def __hash__(self) -> int:
             # Each ModelField is unique for our purposes, to allow making a dict from
             # ModelField to its JSON Schema.
-            return id(self)
+            # build a hash from the field_info and name and mode
+            # This method is probably not safe enough... but how can
+            # we easily hash and compare ModelFields builds with the same data?
+            return hash(
+                (
+                    repr(self.field_info),
+                    self.name,
+                    self.mode,
+                )
+            )
+
+        def __eq__(self, other: Any) -> bool:
+            if not isinstance(other, ModelField):
+                return False
+            return self.__hash__() == other.__hash__()
 
     def get_annotation_from_field_info(
         annotation: Any, field_info: FieldInfo, field_name: str
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -161,21 +161,7 @@ if PYDANTIC_V2:
         def __hash__(self) -> int:
             # Each ModelField is unique for our purposes, to allow making a dict from
             # ModelField to its JSON Schema.
-            # build a hash from the field_info and name and mode
-            # This method is probably not safe enough... but how can
-            # we easily hash and compare ModelFields builds with the same data?
-            return hash(
-                (
-                    repr(self.field_info),
-                    self.name,
-                    self.mode,
-                )
-            )
-
-        def __eq__(self, other: Any) -> bool:
-            if not isinstance(other, ModelField):
-                return False
-            return self.__hash__() == other.__hash__()
+            return id(self)
 
     def get_annotation_from_field_info(
         annotation: Any, field_info: FieldInfo, field_name: str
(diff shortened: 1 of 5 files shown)


lmignon:
This PR also includes a commit to pin the version of trio since the last version 0.22.0 breaks tests. A fix is proposed here #5456

yezz123 (approved):
🚀

BilalAlpaslan (approved):

lmignon:
@tiangolo I rebased this branch on 0.96.0. I hope it will find a path to be included into a next release.

lmignon:
rebased on 0.97

tiangolo:
📝 Docs preview for commit cb8573b7b0e7fcb4d3dc3b37c98dec3f7e3e081c at: https://64940968de6ac3007e21b274--fastapi.netlify.app

lmignon:
I'll adapt the code to work on the 1.X version

lmignon:
rebased on 0.110.0

lmignon:
rebased  on 0.115.0

@tiangolo Any chance to get this reviewed or better merged? It's very useful. Thank you.

sebastienbeau:
@tiangolo We are interested too. Thanks @lmignon for your work

marielejeune:
@tiangolo I'm also interested! Could you please have a look? Thanks

YuriiMotov (commented):
Could anyone describe a real use case when you need to test openapi schema generated with overridden dependencies?

Regarding the use case described in https://github.com/fastapi/fastapi/issues/5451 (and discussion https://github.com/fastapi/fastapi/discussions/8392), I believe this is a misuse of dependencies_override and should be encouraged..

lmignon:
@YuriiMotov 

We use FastAPI's dependency override mechanism to create generic endpoint modules that declare what they need, but not how to get it. This allows us to build reusable API fragments, tailor behavior per deployment, and cleanly separate concerns. It's a lightweight and powerful way to introduce inversion of control in REST API design.

---

### 💡 Context

FastAPI currently provides `app.dependency_overrides` as a way to override dependencies — a feature that's primarily documented and discussed in the context of **testing**.

However,  We are using this mechanism in **production**, not for testing, but to enable **modular and context-adaptable API composition**. This usage pattern has proven extremely powerful in promoting clean architecture, reusable API fragments, and dependency inversion.

I would like to propose that this usage be recognized and supported in the documentation — and ideally improved to fully support production use, particularly in relation to OpenAPI schema generation.

---

### ✅ Use Case: Building Reusable API Fragments

We build endpoint modules that:
- Declare abstract dependencies using `Depends(...)`
- Do **not** hardcode how context-specific data (like the current user, tenant, or config) is obtained
- Can be reused across different apps or clients

At application startup, we use `app.dependency_overrides[...] = ...` to inject concrete implementations according to the current context:

```python
# api/user.py
from fastapi import APIRouter, Depends
from .schemas import User

def get_user() -> User:
    raise NotImplementedError("Override this in your app")

router = APIRouter()

@router.get("/me")
def get_me(user: User = Depends(get_user)):
    return {"username": user.username}

# main.py
from fastapi import FastAPI
from api.user import router as user_router, get_user
from .auth import get_current_user_from_oauth

app = FastAPI()

app.include_router(user_router)

# Runtime context binding
app.dependency_overrides[get_user] = get_current_user_from_oauth

```
---
### 💪 Benefits of This Pattern
1. **Modularity & Reusability**
  API logic becomes portable across projects. For example, the same user endpoints can be reused whether the context uses OAuth2, SSO, API keys, or a static user mock.
2. **Separation of Concerns**
  Endpoints focus only on **_what they need, not how it's obtained_**. Dependency resolution is centralized during app setup.
3. **Contextual Adaptation**
  At deployment time, we can easily swap in context-specific dependencies:
    * SaaS vs. on-premise
    * Auth method differences
    * Multi-tenant logic
    * A/B testing

In short:
* ✅ **Separation of concerns:** endpoint logic is decoupled from its environment
* ✅ **Reusability:** endpoint modules become portable and plug-and-play
* ✅ **Customizability:** each app or deployment can inject its own context logic
* ✅ **Testability**: remains fully compatible with test-time overrides
* ✅ **Clean architecture:** enables inversion of control with minimal boilerplate

This usage aligns closely with dependency injection best practices found in frameworks like Spring, Angular, or ASP.NET Core — and works extremely well within FastAPI's design philosophy.

---

### ⚠️ Limitation: OpenAPI Inaccuracy
The only real issue we've encountered is that the generated OpenAPI schema doesn't reflect overridden dependencies. For example:

* Security schemes introduced by the override (e.g., OAuth2) are not documented
* Request/response models from the override are not shown
* Parameter differences may be silently omitted

We understand that OpenAPI generation happens statically, while dependency_overrides is dynamic. However, this is the main blocker to using this pattern with full confidence in production.  Nevertheless, even if the current PR is not correctly written, it at least proves that it is possible to modify the current implementation to make it dynamic.

---

### 🙏 Thanks

FastAPI’s dependency injection system is one of its strongest features. We'd love to see this powerful and flexible mechanism promoted more broadly as a foundation for clean, modular API architecture — beyond testing.

Thanks for your amazing work!

sebastienbeau (approved):
Tested, and used in production !
Thanks for your work @lmignon

tiangolo:
Thanks for the interest and for the detailed explanation! :cake: 

The current intention of dependency overrides is mainly for tests. Maybe it would make sense to allow a more general way to allow defining dependency overrides in production, but in that case, I think it would probably need a different interface than what is currently available in FastAPI. This might be related to a large refactor I want to do later with how routers work.

Nevertheless, I would get nervous if the overriding dependency is so different that it would generate a different OpenAPI schema, I would think that then it would probably be doing too much, in a very different way, so I would prefer not to have that.

The other part is, any new change or feature I accept, I have to keep maintaining it going forward, and fix any and all the possible side effects and potential breaks it could cause.

I might revisit this idea in the future, but probably in a different way, after some other reactors and features I need to make.

For now, I prefer to pass on this one, but thanks for all the effort put into it! :coffee: :cake:

PR #5456, opened 2022-10-04
Fix PR checks

paulo-raca:
I noticed a couple of failures on my other PR that weren't related to my changes, so I decided to address them:

- On `test_websocket_invalid_path_doesnt_match`, the code inside the `with` block is not suppose to execute, so a `# pragma: no cover` was added
- [trio release 0.22](https://pypi.org/project/trio/0.22.0/) from a few days ago breaks anyio tests due to a DeprecationWarning. 
  It will be fixed in the next anyio release, but in me meantime I added a constraint for `trio < 0.22.0`

PR #5461, opened 2022-10-05
Resolve types correctly (PEP 563, postponed annotations)

### LINKED ISSUE #1654: Import of __future__ annotations breaks class dependency with NameError
Minimal working example:

```Python

import uvicorn
app = FastAPI()

class Test():
    def __call__(self, request: Request):
        return 'test'

@app.get("/test/")
def test_call(test: str = Depends(Test())):
    return {"test": test}

if __name__ == "__main__":
    uvicorn.run("main:app", host="127.0.0.1", port=8000, reload=True, debug=True)
```

### Description

* Run the file / start the server
* It fails to start with the following exception:

> Exception has occurred: NameError
> name 'Request' is not defined

image

### Environment

* OS: Windows:
* FastAPI Version 0.58.0:
* Python version: 3.7

### Additional context

The issue only happens when the `from __future__ import annotations` import is added. Without this import it works fine.
We depend on this import in our code. We don't want to use strings for type definitions.

FastAPI seems to have issues with `from __future__ import annotations`. Here is a similar issues from the past that got resolved: https://github.com/tiangolo/fastapi/issues/446

### LINKED ISSUE #4557: When using postponed annotations (from __future__ import annotations), class based dependencies fail to resolve their dependencies with NameErrorr

### LINKED ISSUE #4750: Forward return type of injected dependencies with Depends()
### First Check

### Commit to Help

### Example Code

```python
"""
I have a function that specifies a return type, but that type is erased 
when I inject it into an endpoint using `Depends()`
"""

app = FastAPI()

async def dependency() -> int:
    return 0

@app.get("/")
async def root(dep = Depends(dependency)):
    # Get mypy to emit the return type of the value returned by the dependency
    reveal_type(dep)  # Revealed type is "Any"
```

### Description

When injecting `dependency` it into my view function with `Depends(dependency)`, the return type (`int`) is erased and replaced with `Any`.

This can be verified by running `mypy` against my code sample above (`reveal_type()` should cause mypy to emit the type it infers)

### Wanted Solution

It would be great if the return type of `dependency` were preserved, i.e.

```python
@app.get("/")
async def root(dep = Depends(dependency)):
    reveal_type(dep)  # Revealed type is "int"
```

---

I see in the definition of [Depends](https://github.com/tiangolo/fastapi/blob/eddbae948f04e13fe412dc45a569d10e34b698a4/fastapi/param_functions.py#L278-L281) that the signature is as follows:

```python
def Depends(dependency: Optional[Callable[..., Any]] = None, *, use_cache: bool = True) -> Any:
```

I'm no expert on the Python type system or Mypy, but I think the following could get Mypy to correctly infer the return type of `dependency`:

```python
from typing import TypeVar, Awaitable, Union

T_ = TypeVar("T_")
MaybeAwaitable = Union[T_, Awaitable[T_]]

T = TypeVar("T")
def Depends(dependency: Optional[Callable[..., MaybeAwaitable[T]]] = None, *, use_cache: bool = True) -> T:
    return params.Depends(dependency=dependency, use_cache=use_cache)
```

### Wanted Code

```python
"""
The return type of my dependency should be preserved 
when injecting it with `Depends(dependency)`
"""

app = FastAPI()

async def dependency() -> int:
    return 0

@app.get("/")
async def root(dep = Depends(dependency)):
    reveal_type(dep)  # Revealed type is "int"
```

### Alternatives

_No response_

### Operating System

macOS

### Operating System Details

_No response_

### FastAPI Version

0.75.0

### Python Version

3.7.9

### Additional Context

My current approach is to explicitly annotate my injected dependencies, but this is prone to error since Mypy doesn't help catch these issues.

Consider:

```python

app = FastAPI()

# Returns a different type now
async def dict_dependency() -> dict:
    return {}

@app.get("/")
async def root(dict_dep: int = Depends(dict_dependency)):
	# But Mypy doesn't catch the issue
    reveal_type(dict_dep)  # Revealed type is "int"
```

### LINKED ISSUE #5065: FastAPI does not respect __wrapped__ attribute, causing NameError for forward references in decorated functions.

mastercoms:
This commit fixes various problems stemming from incorrect evaluation of types in function signatures. 

First, we use Python's own `typing.get_type_hints` method to ensure we can resolve type hints with the correct scope. Python handles classes, wrapped functions which are not considered in the current implementation. I also prefer this approach as it will most likely be the best at handling any future changes to Python.

However, unfortunately, `get_type_hints` resolves type hints for the class `__init__` function when passed a callable class, so I had to add some checks to see if the `__call__` function should be passed in rather than the Callable itself. If the callable is an actual function of some form, we can pass itself to `get_type_hints`. If the callable is a type annotation, we need to call the class type's `__init__` function, and thus we pass that into `get_type_hints`. Otherwise, we pass in the `__call__` function.

Finally, the return type with the old approach was not resolved, so it used the default `return_annotation=typing.Any`. `get_type_hints` is able to retrieve return type information as well, so I was able to pass in the return annotation and thus the return type is no longer erased. In the case that no return type annotation is found, we pass in the default of `Any` again to represent the unknown status of the type.

* Fixes #1654
* Fixes #4557
* Fixes #4750
* Fixes #5065


mastercoms:
Currently failing due to https://github.com/python/mypy/issues/5079

mastercoms:
Now building and passing tests, waiting on #5456 to fix some checks.

yezz123 (approved):

mastercoms:
I'll be looking to update this for 0.89.1 soon.

fastapi/dependencies/utils.py
```diff
@@ -247,36 +247,36 @@ def is_scalar_sequence_field(field: ModelField) -> bool:
 
 def get_typed_signature(call: Callable[..., Any]) -> inspect.Signature:
     signature = inspect.signature(call)
-    globalns = getattr(call, "__globals__", {})
+    is_true_function = (
```
Harry-Lees, line 250: It might look a bit nicer here if you pass a tuple of arguments to isinstance.

```python
if isinstance(call, (FunctionType, MethodType, BuiltinFunctionType))
```

mastercoms: Oh neat, thanks!

mastercoms:
The project seems to have taken a different path on this, unfortunately.

PR #5462, opened 2022-10-05
Improved Performance and Refactor of `jsonable_encoder`

Ben-Brady:
## Description

The current json_encoder function is slow, for several reasons, but the main reason being it runs the slow `isinstance` function for each potential type. I rearranged the comparsions to favour the most common types first as well as implement a direct type comparsion for the most common types. Additionally, in order to implement this it was neccessary to extract each type's parsing code into their own functions.

## Benchmark

| Old | New |
| ---- | ---- |
| image | image |

### DIFF
--- a/docs/en/docs/release-notes.md
+++ b/docs/en/docs/release-notes.md
@@ -23,6 +23,7 @@ hide:
 
 ### Internal
 
+* 🔧 Update Speakeasy sponsor graphic. PR [#13971](https://github.com/fastapi/fastapi/pull/13971) by [@chailandau](https://github.com/chailandau).
 * ⬆ [pre-commit.ci] pre-commit autoupdate. PR [#13969](https://github.com/fastapi/fastapi/pull/13969) by [@pre-commit-ci[bot]](https://github.com/apps/pre-commit-ci).
 * ⬆ Bump actions/download-artifact from 4 to 5. PR [#13975](https://github.com/fastapi/fastapi/pull/13975) by [@dependabot[bot]](https://github.com/apps/dependabot).
 * 👥 Update FastAPI People - Experts. PR [#13963](https://github.com/fastapi/fastapi/pull/13963) by [@tiangolo](https://github.com/tiangolo).


fastapi/encoders.py
```diff
@@ -142,6 +105,54 @@ def jsonable_encoder(
...
+    if type(obj) in (list, set, frozenset, GeneratorType, tuple):
+        return encode_array(obj)
+
+    if isinstance(obj, (str, int, float, type(None))):
+        return obj
+    if isinstance(obj, dict):
+        return encode_dict(obj)
+    if isinstance(obj, (list, set, frozenset, GeneratorType, tuple)):
+        return encode_array(obj)
+    if isinstance(obj, BaseModel):
```
sk-, line 146: Maybe this case should go before all the other `isinstance` checks, as it's quite common to have your endpoint return pydantic models.

Maybe we could move to the bottom all the instance checks for derived types of str, int, etc, as those cases should be rather rare.

How was the order of the checks defined?

sk- (commented):
Great to see this kind of PRs.

fastapi/encoders.py
```diff
@@ -142,6 +105,54 @@ def jsonable_encoder(
...
+        return jsonable_encoder(
+            obj_dict,
+            exclude_none=exclude_none,
+            exclude_defaults=exclude_defaults,
+            custom_encoder=encoder,
+            sqlalchemy_safe=sqlalchemy_safe,
+        )
+
+    # Use type comparsions on common types before expensive isinstance checks
+    if type(obj) in (str, int, float, type(None)):
```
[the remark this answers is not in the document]
Ben-Brady: Profiling with and without obj_type made a negligable difference, I imagine that's because getting an object's type is such a simple operation for the interpreter to perform.

fastapi/encoders.py
```diff
@@ -49,53 +49,15 @@ def jsonable_encoder(
...
-        )
-    if isinstance(obj, Enum):
-        return obj.value
-    if isinstance(obj, PurePath):
-        return str(obj)
-    if isinstance(obj, (str, int, float, type(None))):
-        return obj
-    if isinstance(obj, dict):
+
+    def encode_dict(obj: Any) -> Any:
```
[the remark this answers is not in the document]
Ben-Brady: The problem is that in order to extract out the encoding functions you'd have to pass the configuration down with each encoder which would require a lot more repetition. Although having the functions in this scope isn't ideal, it's better than the alternative for the time being.

fastapi/encoders.py
Ben-Brady: The order was kinda defined arbitrarily based on what I thought would be most common. I also originally thought that the `BaseModel`'s Fields would be counted as `str` instances, however since `BaseModel`s are getting converted to a `dict` they'll always be `str`. I've re-arranged the isinstnace checks to be more logical.

iudeen:
Can you share how you have benchmarked this?

Ben-Brady:
I lost the original benchmark I used, I recreated it but it's not exact. I took some schemas from my own project, applied the jsonable_encoder to them 10,000 times and used vprof to record the results. I felt the example data I used for the benchmarking was appropriate because it's from a real world example as well as making use of several features in the `jsonable_encoder` such as enums, custom encoders, and nested BaseModels.

> vprof -c p benchmark.py

> benchmark.py
```py
from schemas import Post

sample_data = {"id": 3377992785,"created_at": 1666691800.2500513,"uploader": 1552510247,"deleted": False,"source": "https://twitter.com/AdvosArt/status/1584520218507759616","rating": "unrated","full": {"url": "https://api.openbooru.org/media/pK5zTCOH30.webp","mimetype": "image/webp","height": 1024,"width": 1024,"type": "image"},"preview": {"url": "https://api.openbooru.org/media/35-uazKb9L.webp","mimetype": "image/webp","height": 750,"width": 750,"type": "image"},"thumbnail": {"url": "https://api.openbooru.org/media/GchBJhBbDU.webp","mimetype": "image/webp","height": 180,"width": 180,"type": "image"},"media_type": "image","hashes": {"md5s": ["56dcedb90a8fb6ad71a860eedc0ff45b","d23734ecbbcbdd188a7517dbed5cdbeb","7a2ad52ad62bb072deb581ffd47db3f6","48c965935f573a611b739688a7c18ccb"],"sha256s": ["0bbb7631d0c623f20c9aec346e65abe5272c03fa7ec68dd59771c86f21beed01","1dcd78fe0fb5c32f4caeb93c13773f3121acd15b1c74f53ee1f134c08a14870a","dae6dc9c6928c9f2cfb79a2f4652593e28922b98f186b817295945634312af24","8afff20da8b889ced57945505394db037b5f4505efbcd091323c57c2419cfbbe"],"phash": [ ]},"tags": ["image","advosart"],"comments": [ ],"edits": [ ],"upvotes": 0,"downvotes": 0}
post = Post.parse_obj(sample_data)

for _ in range(10_000):
    jsonable_encoder(post)
```

> schemas.py
```py
from pydantic import BaseModel as _PydanticBaseModel, Field
from enum import Enum
import time

class BaseModel(_PydanticBaseModel):
    class Config:
        smart_union = True
        json_encoders = { bytes: lambda _bytes: _bytes.hex() }


class MediaType(str, Enum):
    image = "image"
    animation = "animation"
    video = "video"


class Rating(str, Enum):
    unrated = "unrated"
    safe = "safe"
    sensitive = "sensitive"
    mature = "mature"
    explicit = "explicit"


class BaseMedia(BaseModel):
    url: str = Field(...)
    mimetype: str = Field(...)
    height: int = Field(...)
    width: int = Field(...)
    type: MediaType


class Image(BaseMedia):
    type: MediaType = Field(default=MediaType.image)


class Animation(BaseMedia):
    duration: float = Field(...)
    frame_count: int = Field(...)
    duration:float = Field(...)
    type:MediaType = Field(default=MediaType.animation)


class Video(BaseMedia):
    has_sound: bool = Field(...)
    duration: float = Field(...)
    fps: str = Field(...)
    type:MediaType = Field(default=MediaType.video)


class Hashes(BaseModel):
    md5s: list[bytes] = Field(default_factory=list, description="A list of MD5 Hashes")
    sha256s: list[bytes] = Field(default_factory=list, description="A list of SHA2 256bit Hashes")
    phashes: list[bytes] = Field(default_factory=list, description="A list of SHA2 256bit Hashes")


GenericMedia = Image | Animation | Video


class Post(BaseModel):
    id: int = Field(..., lt=2**64)
    created_at: float = Field(default_factory=time.time)
    uploader: int|None = Field(default=None)
    deleted: bool = Field(default=False)
    source: str = Field(default="")
    rating: Rating = Field(default=Rating.unrated)

    full: GenericMedia = Field(...)
    preview: GenericMedia|None = Field(default=None)
    thumbnail: Image = Field(...)

    media_type: MediaType = Field( ...)
    hashes: Hashes = Field(default_factory=Hashes)

    tags: list[str] = Field(default_factory=list, unique_items=True)
    comments: list[int] = Field(default_factory=list)

    upvotes: int = Field(default=0)
    downvotes: int = Field(default=0)
```

I re-ran the benchmarks after all the bug fixes: 2.2x Speedup

| New | Old |
| ---- | ---- |
| image| image|

fastapi/encoders.py
```diff
@@ -3,14 +3,17 @@
...
-from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Union
 
 from pydantic import BaseModel
 from pydantic.json import ENCODERS_BY_TYPE
 
 SetIntStr = Set[Union[int, str]]
 DictIntStrAny = Dict[Union[int, str], Any]
 
+PRIMATIVE_TYPES = (str, bool, int, float, type(None))
```
FrancescoLuzzi, line 14: I think this is a typo :)
Great work btw!

YuriiMotov:
@Ben-Brady, thanks for your efforts!

Are you still interested in working on it?
How would this PR look like if we try to achieve the performance increase with minimal changes?
I don't really like the idea of having nested functions.
As I understand it, this PR includes several tweaks. What if we measure the performance increase each of them gives us and accept only those that have significant performance increase but require less changes in code?

Ben-Brady:
Yeah I can still work on it, it's been a while but I'll try another crack at it and also fix merge issues from upstream. I can probably find some way to do it without the nested functions

Ben-Brady:
accientally closed the PR when trying to re-implement it on the up to date version
reopenned it here https://github.com/fastapi/fastapi/pull/13982

PR #5465, opened 2022-10-06
🐛 Close FormData (uploaded files) after the request is done

adriangb:


adriangb:
Looks like failures are unrelated to this change: https://github.com/tiangolo/fastapi/pull/5467

adriangb:
Failures resolved,  @tiangolo I think this is ready for review :)

yezz123 (approved):

iudeen (approved):

tiangolo:
Amazing, thank you @adriangb! :bow: 

Do you think you could add a test or two for this?

adriangb:
I think this might difficult to test without touching the internals of something. It does run for every test, so we know that it doesn’t crash at least. I’ll think a bit about how to test this.

adriangb:
So it seems like even Starlette is not testing that `UploadFile.close` gets called. I don't see any way to test this other than doing some monkey patching of stuff like that. @tiangolo would you accept this without a test or is that a no-go?

tiangolo:
I found a way to test it! :tada: I added the commit on top.

Thanks for the contribution! :rocket:

adriangb:
That was genius!

tiangolo:
Haha thank you! 😊🙈😬

PR #5468, opened 2022-10-06
make sure sub_value type is a function

utopictown:
to fix

fastapi/dependencies/utils.py", line 693, in request_body_to_args
    tg.start_soon(process_fn, sub_value.read)
AttributeError: 'str' object has no attribute 'read'


tiangolo:
Hey there, please create an issue with a minimal, self-contained example that shows your problem/use case and that I can copy paste to reproduce it. That way we can see what's happening and understand why you are interested in this change.

PR #5472, opened 2022-10-07
perf: improve solve_dependencies increasing requests/s by 18%

sk-:
Improve performance of function `solve_dependencies` by reducing accesses to headers, cookies and other request/response properties, which are slow in starlette. This yields and increase of requests / second of about 18%.

The PR include two changes:
The first were we use `Response(status_code=0)` to avoid creating the response with the `Content-Length` header, because accessing response.headers and mutating it is slow.

The second change checks whether there are going to be changes to the values. If not, then we don't access the request properties.

Note that the funciton `request_params_to_args` will return empty if the first argument is an empty list.

## Performance Tests
The change was tested using uvicorn with the following settings

```
uvicorn --no-access-log --no-proxy-headers test:app
```

and the following app

```py
from typing import Any

import fastapi
from fastapi import responses, routing

app = fastapi.FastAPI()

@app.get("/sync")
def sync():
    return "ok"

@app.get("/async")
async def asynch():
    return "ok"
```

### Baseline
```
$ wrk -d 30 http://localhost:8000/async
Running 30s test @ http://localhost:8000/async
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.22ms  204.74us   8.43ms   94.62%
    Req/Sec     4.06k   273.50     7.16k    92.18%
  243045 requests in 30.10s, 29.67MB read
Requests/sec:   8074.13
```

### With the change
```
$ wrk -d 30 http://localhost:8000/async
Running 30s test @ http://localhost:8000/async
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.04ms  307.42us  10.70ms   97.48%
    Req/Sec     4.80k   308.87     5.06k    94.67%
  286439 requests in 30.01s, 34.97MB read
Requests/sec:   9544.14
Transfer/sec:      1.17MB
```

Note that the original PR had also another change that allows it to squeeze some more performance gains, although minimal.

Instead of constructing the response as

```py
response = Response()
del response.headers["content-length"]
```

we changed it to

```py
response = Response(status_code=0)
```

which avoids creating the headers in the first place, which in turns also avoid creating a mutable copy of the headers.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -539,8 +539,9 @@ async def solve_dependencies(
     values: Dict[str, Any] = {}
     errors: List[Any] = []
     if response is None:
-        response = Response()
-        del response.headers["content-length"]
+        # HACK: avoid generating the content-length header by setting the status code
+        # to a number lower than 200.
+        response = Response(status_code=0)
         response.status_code = None  # type: ignore
     dependency_cache = dependency_cache or {}
     sub_dependant: Dependant
@@ -602,23 +603,30 @@ async def solve_dependencies(
             values[sub_dependant.name] = solved
         if sub_dependant.cache_key not in dependency_cache:
             dependency_cache[sub_dependant.cache_key] = solved
-    path_values, path_errors = request_params_to_args(
-        dependant.path_params, request.path_params
-    )
-    query_values, query_errors = request_params_to_args(
-        dependant.query_params, request.query_params
-    )
-    header_values, header_errors = request_params_to_args(
-        dependant.header_params, request.headers
-    )
-    cookie_values, cookie_errors = request_params_to_args(
-        dependant.cookie_params, request.cookies
-    )
-    values.update(path_values)
-    values.update(query_values)
-    values.update(header_values)
-    values.update(cookie_values)
-    errors += path_errors + query_errors + header_errors + cookie_errors
+    if dependant.path_params:
+        path_values, path_errors = request_params_to_args(
+            dependant.path_params, request.path_params
+        )
+        values.update(path_values)
+        errors.extend(path_errors)
+    if dependant.query_params:
+        query_values, query_errors = request_params_to_args(
+            dependant.query_params, request.query_params
+        )
+        values.update(query_values)
+        errors.extend(query_errors)
+    if dependant.header_params:
+        header_values, header_errors = request_params_to_args(
+            dependant.header_params, request.headers
+        )
+        values.update(header_values)
+        errors.extend(header_errors)
+    if dependant.cookie_params:
+        cookie_values, cookie_errors = request_params_to_args(
+            dependant.cookie_params, request.cookies
+        )
+        values.update(cookie_values)
+        errors.extend(cookie_errors)
     if dependant.body_params:
         (
             body_values,
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -539,9 +539,8 @@ async def solve_dependencies(
     values: Dict[str, Any] = {}
     errors: List[Any] = []
     if response is None:
-        # HACK: avoid generating the content-length header by setting the status code
-        # to a number lower than 200.
-        response = Response(status_code=0)
+        response = Response()
+        del response.headers["content-length"]
         response.status_code = None  # type: ignore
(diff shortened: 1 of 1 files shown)


sk-:
Note that the failing tests in python 3.7 and 3.9 are not related to this PR and are failing as well for other PRs.

sk-:
Tests are passing now, but the docs generation was cancelled and I have no way to retry the workflow without pushing unnecessary commits.

yezz123:
> Tests are passing now, but the docs generation was canceled and I have no way to retry the workflow without pushing unnecessary commits.

Lately some issues are related to the packages and containers we use Github actions that's why, but overall no problem if the PR doesn't have effect in docs 🚀

fastapi/dependencies/utils.py
```diff
@@ -473,8 +473,9 @@ async def solve_dependencies(
     values: Dict[str, Any] = {}
     errors: List[ErrorWrapper] = []
     if response is None:
-        response = Response()
-        del response.headers["content-length"]
+        # HACK: avoid generating the content-length header by setting the status code
+        # to a number lower than 200.
+        response = Response(status_code=0)
```
Kludex, line 478: This is just changing one hack by another 🤷‍♂️

sk-: But it's a more performant hack, because it does not require to create the mutable headers, just to remove the only entry.

I'm happy to remove it if that would mean the rest would be merged.

No idea though why it's creating an invalid response with a a `None` status code, which bypasses the type checks.

sk-: @Kludex any thoughts? I'm happy to remove the hack if that would make the pr easier to review.

adriangb: Yes, I think that without this hack, the other part should be pretty straightforward once Sebastián gets around to review it. Great work on that part!

sk-: Done, thanks!

Nikakto:
Does anybody know reasons why fastapi resolve dependencies one by one instead like asyncio.gather() ?

Code example:
```
@pytest.mark.asyncio
async def test_fastapi_is_slowapi():
    async def dep0():
        await asyncio.sleep(5)
        return 0

    async def dep1():
        await asyncio.sleep(5)
        return 0

    @app.get("/testslow")
    async def app_testslow(d0=Depends(dep0), d1=Depends(dep1)) -> int:
        return 10

    response = await client.get("/testslow")

    assert response.elapsed.seconds >= 10  # True
```

sk-:
I'm closing this PR because the fastapi team does not seem to be interested in this at all.

I took into consideration all requested changes, simplifying the PR, and commented on this issue and similar but no one from the core team has even commented on whether they are willing to merge it. At some point they marked it as investigating, but nothing was done. They didn't even requested further performance testing or benchmark.

Recently, they marked it as P4, without even mentioning why. This is surprising because it's not even a complicated PR.

So, in light of the team's (lack of) response, I'm closing this.

tiangolo:
Thanks for the interest @sk-, there are many PRs I need to review, and many depend on other things that are changing or are planned to change, so I have to make sure I'm doing everything in the right order. Putting labels allows me to organize them and see what I have to handle first.

Before deeply reviewing a PR I can't say if I'm gonna be willing to merge it or not, and until I can put the time to deeply review it I can't comment on that for each PR.

Mentioning in each PR why I put one or another label takes time, and that's time (e.g. at least 15 min just checking this and replying to this) I (and others) could have instead invested in reviewing it or any PR that needed reviewing first, which in turn would lead to a faster resolution in the end.

If you want things to move faster for your PR, something you could do is help me maintain the rest, you could answer questions, review other PRs, etc. That would definitely accelerate the process needed fro me to get to your specific PR. https://fastapi.tiangolo.com/help-fastapi/#help-maintain-fastapi

I think this PR is interesting, I have to review a couple of other things first to make sure it's alright. And I would potentially merge it, although I don't see an extreme hurry on this PR above other things (e.g. recently I was solving a security vulnerability, that always takes priority over other things), so I wouldn't promise when it would be merged. If you want, you could open it again, and as I continue reviewing things, I would get to this one. ☕

PR #5474, opened 2022-10-08
mm: allow custom redoc theme and custom stylesheet in redoc

matt-mercer:
support for redoc themes and an additional custom stylesheet in redoc custom html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ env3.*
 env
 docs_build
 venv
+.venv/
 docs.zip
 archive.zip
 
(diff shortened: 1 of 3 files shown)


tiangolo:
Thanks for the interest, but as there are many ways to customize and extend the docs, I perfect to explain how to completely override them: https://fastapi.tiangolo.com/advanced/extending-openapi/#include-the-custom-docs

Otherwise I would end up extending and adding parameters to FastAPI for every possible customization, with extra maintenance, and even worse, people would have to learn that, when in the end there's a high chance they would need to override the docs entirely either way.

PR #5489, opened 2022-10-12
Add tutorial for ContextVar based DB session access from a single place

ahmedsadman:
Database sessions can be accessed in FastAPI either through dependency or request state. Both of these methods require the db session to be passed from the controller layer to the target function. If there are more function layers in between, which is often the case for larger applications, the DB session needs to be passed through multiple layers. This can go against DRY principles. 

A similar issue regarding this was reported here: https://github.com/tiangolo/fastapi/issues/2894

There is a better solution with `contextvars` and middlewares in this case. This PR adds the tutorial in the official doc


ahmedsadman:
No one to review this PR? 😢

tiangolo:
Thanks for all the effort you put into this! ☕

ContextVars are a very delicate and nuanced subject/tool. It's very easy to end up with hard to debug problems, even more when interacting with threads, different async task contexts, etc. And you do all that with FastAPI, knowingly or unknowingly. For example, if you have any Middleware based on BaseHTTPMiddleware, as that creates a new task group, it creates a copy of the context vars context for the task, so, changes are not propagated upwards. The same way, you cannot set a context var in a sync function as that would be executed in a thread worker, it would have to be set in an async function that is run on the main thread running the event loop, etc. ...so, I recommend avoiding them when possible and only using them when strictly necessary.

And although I'm all in for explaining things and concepts, in FastAPI I'm normally explaining just the simplest and most straightforward way to solve something, and encouraging people to find the simplest solution.

So, in this case, I'll pass on this one. But if you want, I think you could want to write a blog post with your ideas. 🤓

Thanks again for the interest! ☕

ahmedsadman:
@tiangolo Is there any other better way to achieve this without using `ContextVars`?

tiangolo:
No, ContextVars is the way to go, but it's so delicate that I would try to avoid it when possible. I've seen so many problems, so many times, it always bites me, and I'm normally the last/only person that would be "able to debug it". ContextVars are great when they are necessary, but it's better to use them only when strictly necessary.

ahmedsadman:
@tiangolo So even for larger applications it's fine to pass down the database object to nested functions, or do you have some specific pattern for that? If so, can you please share?

tiangolo:
If possible, yes, ideally pass it down. If you are migrating an old code base that makes it unfeasible to pass it down, then maybe you have to deal with it and use context vars.

PR #5490, opened 2022-10-12
✏️ Update highlighted line in `docs/en/docs/tutorial/bigger-applications.md`

papb:


Ryandaydev:
@papb If I am reading the tutorial correctly, the existing doco incorrectly highlights row 4.

I think instead of this change (which highlights lines 4 & 5), I think it should be changed to highlight row 5.

Ryandaydev (changes requested):
@papb If I am reading the tutorial correctly, the existing doco incorrectly highlights row 4.

I think instead of this change (which highlights lines 4 & 5), I think it should be changed to highlight row 5.

papb:
@Ryandaydev Thank you, you're right. After re-reading I decided to also change another block to highlight one extra line, can you take a look?

iudeen (approved):
Looks good. Nice catch!

Ryandaydev (approved):
Requested changes have been applied.

alejsdev:
Good catch. Thanks @papb 🙂 

And thanks for the review @iudeen and @Ryandaydev ✌️

PR #5492, opened 2022-10-13
👷 Add GitHub Action gate/check

webknjaz:
This adds a GHA job that reliably determines if all the required dependencies have succeeded or not.

It also allows to reduce the list of required branch protection CI statuses to just one — `check`. This reduces the maintenance burden by a lot and have been battle-tested across a small bunch of projects in its action form and in-house implementations of other people.

It is now in use in aiohttp (and other aio-libs projects), CherryPy, attrs, coveragepy, some of the Ansible repositories, pip-tools, pydantic, spaceship-prompt, all of the jaraco's projects (like `setuptools`, `importlib_metadata`), some PyCQA, PyCA, PyPA and pytest projects, a few AWS Labs projects.

The story behind this is explained in more detail at https://github.com/marketplace/actions/alls-green#why.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,3 +75,19 @@ jobs:
         with:
           name: coverage-html
           path: htmlcov
+
+  # https://github.com/marketplace/actions/alls-green#why
+  check:  # This job does nothing and is only used for the branch protection
+
+    if: always()
+
+    needs:
+      - coverage-combine
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}


adriangb:
This is really smart!

iudeen (approved):
Amazing!

tiangolo:
Cool! I have a question, what happens with the smokeshow coverage job that is triggered after the main test workflow finishes?

webknjaz:
Nothing should change in that regard. Or do you have a specific concern on your mind?

tiangolo:
I understand this helps require all the "test" workflow to pass, then I would add this to the branch protection instead of the other individual rules, right?

Then, I also have "smokeshow / coverage" as required in branch protection. But that's not part of the same "test" workflow. So this wouldn't cover that and I would still need to add that one manually to branch protection, right?

webknjaz:
@tiangolo the answer to both questions is "yes".

By the way, I think you can now use reusable workflows in the same repo so you might be able to make them run in the same workflow run and make a common check gate. That's something someone could try experimenting with in the future.

webknjaz:
Rebased it.

webknjaz:
@tiangolo ready to give it a try? There's still no conflicts in the PR, so it's ready to be merged anytime.

tiangolo:
Very cool!

I have been trying it in other smaller projects. The only thing I miss is if it could output in the logs what matrix versions it checked. Currently, it only outputs the name of the job, but not the matrix combinations. E.g.

```
📝 test → ✓ success [required to succeed]
```

But it doesn't have info about the Python versions tested, to confirm that it's indeed checking all the combinations that I would put on the protection branch rules.

webknjaz:
> I have been trying it in other smaller projects. The only thing I miss is if it could output in the logs what matrix versions it checked. Currently, it only outputs the name of the job, but not the matrix combinations. E.g.

GitHub does not provide any such job details in the `${{ needs }}` context for the matrix jobs. It only exposes the names and the outcomes basically. Implementing heuristics seems fragile. You can control which matrix subjobs are allowed to fail via `continue-on-error` but other than that, there's really no way of lurking into the workflow.

tiangolo:
@webknjaz I'm checking this and I see that the current file doesn't check for smokeshow, and that's a triggered workflow because it needs access to tokens, so it lives in a different workflow file. How do I add smokeshow as a `need` in this action?

webknjaz:
You can't add cross-workflow deps.

webknjaz:
@tiangolo extending the explanation: you pass the `${{ needs }}` context which `alls-green` uses to analyze job outcomes. It is not only limited but also only exists in the job context of a workflow. Another workflow run will happen independently and won't have anything from that context.
Besides that workflow job's status does not even appear in the PR checks. Instead, it uses Checks API to report a custom check not matching the job name.

I'd recommend keeping that check in branch protection, just like the status from pre-commit.ci.

If you really want another aggregated check, you could try using https://github.com/WyriHaximus/github-action-wait-for-status which does status polling in a loop and could theoretically watch statuses reported by anything, not just GHA.
Note that you'd need to do this in a yet another separate workflow since it'd deadlock if you were to add this to the check job introduced here. I haven't tried doing this, though.

tiangolo:
Cool, thanks for the explanation @webknjaz! :rocket:

PR #2944, opened 2021-03-13
✨ Add support for `lifespan` async context managers (superseding `startup` and `shutdown` events)

uSpike:
This adds support for the `lifespan` argument to the application and router, which was added in starlette.

I tidied up some of the event tests as well to promote reuse.

Fixes #2943

Issue #2943, opened 2021-03-12
Support starlette "lifespan" context for application


uSpike:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Description

Starlette added support for the lifespan context function in 0.13.5.  It's not well documented, but the code does [suggest](https://github.com/encode/starlette/blob/ed73b97c9b8f4aa527eed88032c04ce9fe109a3c/starlette/applications.py#L51):
```python
        # The lifespan context function is a newer style that replaces
        # on_startup / on_shutdown handlers. Use one or the other, not both.
```

For my purposes, it's much more convenient to use a (async)contextmanager for the startup and shutdown events of my application.  It would be nice to have the option

### The solution you would like

I'd like an easier way, and one that's officially supported, to use the lifespan context function.

```Python
from fastapi import FastAPI

async def lifespan(app):
    print("startup")
    async with SomeResource():
        yield
    print("shutdown")

app = FastAPI(lifespan=lifespan)
```

or

```Python
from fastapi import FastAPI

app = FastAPI()

@app.lifespan
async def lifespan(app):
    print("startup")
    async with SomeResource():
        yield
    print("shutdown")
```

### Describe alternatives you've considered

I can already accomplish this simply by doing:
```python
from fastapi import FastAPI

async def lifespan(app):
    print("startup")
    async with SomeResource():
        yield
    print("shutdown")

app = FastAPI()
app.router.lifespan_context = lifespan
```
however this is not officially supported and would likely break if accidentally using `app.on_event` in addition.

[...]

PR #5503, opened 2022-10-16
Add lifespan context v3

JonathanPlasse:
rebased @uSpike's work from #2944 addressing #2943


yezz123 (approved):

fastapi/routing.py
```diff
@@ -41,6 +42,7 @@
 from pydantic.error_wrappers import ErrorWrapper, ValidationError
 from pydantic.fields import ModelField, Undefined
 from starlette import routing
+from starlette.applications import Starlette
```
JarroVGIT, line 45: You only use this import for Typing purposes, wouldn't it make more sense to make use of the more general `ASGIApp` type in `starlette.types`?

tests/test_router_events.py
```diff
@@ -85,3 +82,28 @@ def test_router_events():
     assert state.app_shutdown is True
     assert state.router_shutdown is True
     assert state.sub_router_shutdown is True
+
+
+def test_app_lifespan_state(state: State) -> None:
+    @asynccontextmanager
+    async def lifespan(app: Starlette) -> AsyncIterator[None]:
```
JarroVGIT, line 89: This would be a `AsynGenerator`, no🤔? Because of the `yield` in this function.

JarroVGIT (changes requested):
I think it's better to import specific type annotations rather than the whole of `Starlette` class, just for typing. 

Is there a reason you call this 'lifespan context v3'? What does the v3 refer to? This is quite established behaviour in Starlette (by the looks of it; around 3 years in or so).

tests/test_router_events.py
Kludex: ref in code: https://github.com/tiangolo/fastapi/blob/f67b19f0f73ebdca01775b8c7e531e51b9cecfae/fastapi/concurrency.py#L19

fastapi/routing.py
JonathanPlasse: In `starlette.routing` the type used is Any.

tests/test_router_events.py
JonathanPlasse: According to [StackOverflow](https://stackoverflow.com/questions/63125259/what-is-the-proper-way-to-type-hint-the-return-value-of-an-asynccontextmanager), `AsyncIterator` is also correct as a generator is an iterator, so it is correct and also less complex to type.

JonathanPlasse:
> I think it's better to import specific type annotations rather than the whole of `Starlette` class, just for typing.
> 
> Is there a reason you call this 'lifespan context v3'? What does the v3 refer to? This is quite established behaviour in Starlette (by the looks of it; around 3 years in or so).

This is the 3rd pull request made to add lifespan to FastAPI.

tests/test_router_events.py
JarroVGIT: Every generator is an iterator, but not every iterator is a generator. The point of typing in python is to provide information on the object type as precise as possible. If your take your argument to the extreme, everything could be typed as `object` and it would be technically correct, but not really useful. 

I don't agree with "less complex to type", it is literally one argument more (as generators have a 'send' type and a 'yield' type)?

fastapi/routing.py
```diff
@@ -480,6 +481,7 @@ def __init__(
         route_class: Type[APIRoute] = APIRoute,
         on_startup: Optional[Sequence[Callable[[], Any]]] = None,
         on_shutdown: Optional[Sequence[Callable[[], Any]]] = None,
+        lifespan: Optional[Callable[[Any], AsyncContextManager[Any]]] = None,
```
JarroVGIT, line 484: Now it is just inconsistent, no? It's typed `FastAPI` on L74 in applications.py, but here it's typed `Any`, while I would still argue the correct typing would be the `ASGIApp` from `starlette.types`. Because, if you have different orders of middleware (which would all be `ASGIApp`) you don't have to specifically have a `FastAPI` app object in the lifespan param.

JarroVGIT (changes requested):

fastapi/routing.py
JarroVGIT: Interestingly this is actually how it is typed in Starlette I just noticed. But my remark would remain the same, am I missing something maybe?

tests/test_router_events.py
JonathanPlasse: You are right, I made the change.

fastapi/routing.py
JonathanPlasse: `Any` could be replaced by `FastAPi` which would only be imported if `typing.TYPE_CHECKING` is True (during static analysis and not runtime).

JonathanPlasse:
All requested changes have been made.

tiangolo:
Thank you @JonathanPlasse! :cake: 

And thanks for the help @JarroVGIT, @Kludex, @yezz123 :bow: 

I included your commits in the original PR https://github.com/tiangolo/fastapi/pull/2944 and merged that one.

So I'll now close this one.

This will be available in FastAPI `0.93.0`, released in the next few hours. :tada: 

PR #5504, opened 2022-10-16
Update deta.md

mycaule:


tiangolo:
It seems this is actually not necessary, so I'll pass on that. But thanks for your interest! ☕

mycaule:
It IS necessary, the command doesn't work without specifying `--python`

mycaule:
When I tried to deploy using Python 3.10 and deta 1.1.0 it didn't work without the --python

iudeen:
Weird. It works just fine without the flag.

PR #5506, opened 2022-10-17
Add supporting for async callable object

sviat9440:


yezz123 (approved):

sviat9440:
What are the next steps?

fastapi/dependencies/utils.py
```diff
@@ -261,6 +263,15 @@ def get_typed_signature(call: Callable[..., Any]) -> inspect.Signature:
     return typed_signature
 
 
+def is_coroutine_function(obj: Any) -> bool:
+    while isinstance(obj, functools.partial):
+        obj = obj.func
+
+    return asyncio.iscoroutinefunction(obj) or (
+        callable(obj) and asyncio.iscoroutinefunction(obj.__call__)
+    )
```
Kludex, line 272: https://github.com/encode/starlette/blob/858629f5188bc79d452600b1eb90eaa0045f6454/starlette/_utils.py#L1-L12

sviat9440: image

sviat9440: @Kludex

Kludex:
The next step is to wait for Sebastián to review it. 🙏

iudeen (approved):

sviat9440:
@tiangolo @Kludex @iudeen any progress?

sviat9440:
...

sviat9440:
Is the project dead?

Kludex:
Nop.

Kludex:
I'll review tomorrow.

fastapi/dependencies/utils.py
Kludex: yes, ignore it.

Kludex (changes requested):

YuriiMotov (changes requested):
This gives wrong openapi schema if we use partial with named parameters:

```python
from functools import partial

app = FastAPI()

def dep1(n: int):
    return n + 1

@app.get("/")
def main(d1=Depends(partial(dep1, n=10))):  # but `partial(dep1, 10)` works well
    return d1
```
image

It uses `10` as a default value for parameter, but you can still override it passing `n` parameter:
image

Also, it doesn't work with `async` functions.

YuriiMotov:
There is an alternative PR #9753 that supports `async` functions, class methods and generators. But it also fails when named parameters are used with `partial`.

@sviat9440, thanks for your efforts!
Would you like to further elaborate on this?

sviat9440:
Guys, since I opened this pull request, I've changed two jobs. It's no longer relevant to me.

YuriiMotov:
> Guys, since I opened this pull request, I've changed two jobs. It's no longer relevant to me.

So, then let's close this in favor of #9753.
@sviat9440, thanks again!

PR #5512, opened 2022-10-18
Found a code omission in the tutorial

yuki1sntSnow:


yuki1sntSnow:
Found a code omission in the tutorial
Part [Query Parameters and String Validations] of the tutorial omits the [Ellipsis (...)]

JarroVGIT (changes requested):
This was specifically changed in PR 4906 when this feature was introduced. The `...` is not longer required in the Query param, and the omission is by design. 

I would recommend to close this PR.

yuki1sntSnow:
Thank you for providing that information.

tiangolo:
Thanks for the help @JarroVGIT ! 🍰

And thanks for coming back to close the issue @yuki1sntSnow ☕

PR #5517, opened 2022-10-18
Security Models Refactor

iamibi:
This Pull Request does the following:
- Refactors the usage of different models that were distributed over different files inside `fastapi/security/` module
- Moved their implementations to `fastapi/security/models.py` file to have their own namespace
- During the above process, other relevant files were touched and `fastapi/security/base.py` file was **removed**


iudeen:
Any benefits?

tiangolo:
I prefer to keep them as they are, with the models close to the code they are related to. So I'll pass on this one. But thanks for your interest! ☕

PR #5533, opened 2022-10-24
Use dict (ordered set) for APIRouter.tags 

aradipe:
It was possible to specify the same tags multiple times which caused issues with ReDoc Discovered while using fastapi-utils' CBV

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -499,7 +499,7 @@ class APIRouter(routing.Router):
                 "/"
             ), "A path prefix must not end with '/', as the routes will start with '/'"
         self.prefix = prefix
-        self.tags: List[Union[str, Enum]] = tags or []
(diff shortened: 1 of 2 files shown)


JarroVGIT:
If you provide the same tag twice, it ends up twice in the openapi.json. That doesn't seem like that is wrong behaviour at all, nor isn't it against the spec (it is [against the spec](https://swagger.io/specification/#openapi-object) to list the same tag twice in the openapi document `tags` key). That said, SwaggerUI plays nicely along (as it should) and only redocs seems to have an issue with it. I would argue this need to be solved in Redoc, not in FastAPI.

aradipe:
Closing as discussed in https://github.com/tiangolo/fastapi/issues/5534

tiangolo:
Thanks for the help @JarroVGIT! ☕ 

And thanks for closing it @aradipe 🍰

PR #5544, opened 2022-10-26
Make pyproject.toml PEP 621 compliant

simontegelid:
The issue is (had issues with building this with buildroot), which requires this to be correct:

```text
$ validate-pyproject pyproject.toml                          
Invalid file: pyproject.toml
[ERROR] `project.license` must be valid exactly by one definition (2 matches found):

    - keys:
        'file': {type: string}
      required: ['file']
    - keys:
        'text': {type: string}
      required: ['text']
```

For more info: https://peps.python.org/pep-0621/#license

After this change:

```text
$ validate-pyproject pyproject.toml
Valid file: pyproject.toml
```

For the record:

```text
$ validate-pyproject --version     
validate_pyproject 0.10.1
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "fastapi"
 description = "FastAPI framework, high performance, easy to learn, fast to code, ready for production"
 readme = "README.md"
 requires-python = ">=3.7"
-license = "MIT"
+license = {text = "MIT"}
 authors = [
     { name = "Sebastián Ramírez", email = "tiangolo@gmail.com" },
 ]


iudeen (commented):
As I understand, the newer PEP0639 supersedes 621. As per 639, subkeys are deprecated. 

Also we can note that PEP0639 is still in draft. I however leave this upto maintainer to temporarily merge this till PEP0639 is finalized. 

Refer: https://peps.python.org/pep-0639/#deprecate-license-key-table-subkeys

simontegelid:
Yes, you're right. I had an equivalent PR at uvicorn, which were closed due to PEP 639. I didn't know about that PEP until now. Lets close this one too and wait for PEP 639.

https://github.com/encode/uvicorn/pull/1734#issuecomment-1293549346

tiangolo:
Yep, indeed, it's not necessary, but thanks for the effort! :coffee: 

I'm not sure what's buildroot, I use Hatch here. You could also use plain `pip install ./`, with a modern/recent version of `pip`.

Given that, I'll close this one. but thanks! :cake:

PR #5545, opened 2022-10-27
📝 Fix typo in docs with examples for Python 3.10 instead of 3.9

ghpZ54K8ZRwU62zGVSePPs97yAv9swuAY0mVDR4:
`|` has to be in Python 3.10 version and above. Fix it.

PR #5547, opened 2022-10-27
🐛 Fix internal Trio test warnings

samuelcolvin:
Hi, without this fastapi is failing again due to trio warnings while testing. See #5545 and https://github.com/pydantic/pydantic/pull/4662.

There are two problems here (or one causing two issues):

There's a runtime warning from trio:

```
RuntimeWarning: You seem to already have a custom sys.excepthook handler installed. I'll skip installing Trio's custom handler, but this means MultiErrors will not show full tracebacks.
```

This was being raised by some tests, but also by the warning line `'ignore::trio.TrioDeprecationWarning',` - hence the new warning ignore has to come before that line.

I've no idea what change is causing this, would be great if every test dep could be frozen, e.g. with `pip-compile`, so these issues don't keep coming up.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,5 +137,6 @@ filterwarnings = [
     'ignore:The loop argument is deprecated since Python 3\.8, and scheduled for removal in Python 3\.10:DeprecationWarning:asyncio',
     'ignore:starlette.middleware.wsgi is deprecated and will be removed in a future release\..*:DeprecationWarning:starlette',
     # see https://trio.readthedocs.io/en/stable/history.html#trio-0-22-0-2022-09-28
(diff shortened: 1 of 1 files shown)


odiseo0 (approved):

iudeen (approved):
I think we need to freeze test dependencies as samuelcolvin mentioned.

samuelcolvin:
@tiangolo you probably want to remove pytest-cov, I understand it's no longer maintained and can be replaced by just

```bash
coverage run -m pytest
```

I haven't removed it here since you've set some configuration options that will need to move to `pyproject.toml`.

jujumilk3:
This PR needed to me too. thx.

Issue #5552, opened 2022-10-28
Duplicate tag names are possible on the OpenAPI object level


aradipe:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Union
from fastapi import FastAPI

app = FastAPI(openapi_tags=[{'name': 'items', 'description': 'items1'}, {'name': 'items', 'description': 'items2'}])

@app.get("/items", tags=["items"])
def read_item(q: Union[str, None] = None):
    return {"q": q}
```

### Description

The OpenAPI [specification](https://swagger.io/specification/#openapi-object) states that "Each tag name in the list MUST be unique." for the OpenAPI object.

```bash
$ curl -s http://localhost:8000/openapi.json | jq .tags
[
  {
    "name": "items",
    "description": "items1"
  },
  {
    "name": "items",
    "description": "items2"
  }
]
```
Also Swagger shows the items operation to be with `items1` tag, while ReDoc shows `items2`.

### Operating System

macOS

### FastAPI Version

0.85.1

### Python Version

Python 3.10.8

iudeen:
Nice catch! and thanks for the PR!!

PR #5553, opened 2022-10-28
✨ Do not allow duplicate tag names in the OpenAPI metadata (OpenAPI object)

aradipe:
Addresses issue #5552


aradipe:
Failing because of an issue already addressed in PR #5547

fastapi/openapi/models.py
```diff
@@ -400,6 +400,14 @@ class OpenAPI(BaseModel):
     class Config:
         extra = "allow"
 
+    @validator("tags")
+    def check_foo(cls, v):  # type: ignore
```
iudeen, line 404: let's add descriptive names? Ref: https://google.github.io/styleguide/pyguide.html#316-naming

iudeen (changes requested):

JarroVGIT:
Nice catch @aradipe !

fastapi/openapi/models.py
aradipe: Sorry about that, that was a lazy copy-paste error from the pydantic docs :) fixed

YuriiMotov:
@aradipe, thanks for your interest in FastAPI!

Tests fail with Pydantic V2. Would you like to fix the code?

PR #5554, opened 2022-10-28
Overhaul `solve_dependencies()` for performance

kristjanvalur:
This pr provides an overhaul of solve_dependencies().  It is aimed at fixing performance and increasing readability.

1. A hidden performance bug is removed, when an extra level of dependencies is traversed, even if the cache is hit.
2. A hidden performance bug is removed when a new Depends is created for every dependency if there are _any_ dependency overrides.
3. The evaluation of builtin-dependencies is moved into specialized callables/closures to avoid a lengthy series of `if` statements for every Dependency evaluation
4. the call-context of dependency solving is moved into its own `DependencySolverContext` dataclass avoiding a lot of boilerplate code in the main solve_dependencies() function

A new test is added to test deep `Depends` hierarchies and caching.

A specialized unittests is also added which uses the `timeit` module to measure performance of `Depends` processing.  We run an AsyncClient to reduce overhead, for complete requests.  Also direct evaluation of the endpoint dependencies, both for `Depends` and also `Request` . (The test does nothing by default using `if False:`.  Could also add a pytest switch to enable it.)

On a local PC, I measure, with the `pytest_cahce_perf` unittest enabled and `pytest --quiet -k test_deep_cache_perf` unittest:
 - before:
    ```
    tests/test_dependency_perf.py deep cache client requests
    did 1000 calls in 0.3157117000000653 seconds
    time per call: 0.32ms, rate: 3167.45/s
    deep cache direct solve
    did 2000 calls in 0.2325039000002107 seconds
    time per call: 0.12ms, rate: 8602.01/s
    request direct solve
    did 50000 calls in 0.3208132000004298 seconds
    time per call: 0.01ms, rate: 155853.94/s
    ```
 - after:
    ```
    tests/test_dependency_perf.py deep cache client requests
    did 1000 calls in 0.2511885999997503 seconds
    time per call: 0.25ms, rate: 3981.07/s
    deep cache direct solve
    did 5000 calls in 0.2859181999997418 seconds
    time per call: 0.06ms, rate: 17487.52/s
    request direct solve
    did 100000 calls in 0.3904711000000134 seconds
    time per call: 0.00ms, rate: 256100.90/s
    ```

Performance of deep Depends hierarchies is doubled, mostly due to the fixing of the cache mechanic.  But simple dependencies, like the Request, are also faster, since the whole depdendency evaluation loop is simpler, and we use dispatch functions to fill the parameters rather than switch statements.

Running the entire testsuite, with:
`pytest --quiet tests`
reports 28.4s before and 28.1 after the change.  The whole testsuite however typically creates an app and runs the endpoint once.  Also, there is tremendous overhead in using `TestClient` rather than `AsyncClient` for each endpoint call.  The optimizations are expected to trade app init time for endpoint execution time.


fastapi/dependencies/utils.py
```diff
@@ -454,77 +455,76 @@ async def solve_generator(
     return await stack.enter_async_context(cm)
 
 
+@dataclasses.dataclass
+class DependencySolverContext:
```
odiseo0, line 459: Why are you using `dataclases.dataclass` here instead of Pydantic?

I just want to know if there's a reason in particular

kristjanvalur: Dataclass is basically a quick way to declare a class with initializers.  I'm not looking for pydantic's validation.  This is supposed to be low overhead.
Could have used a plain class instead and written my own __init__().
Could also have made the solve_dependencies() an instance method, but I don't see much of that kind of programming style in FastAPI so I left it functional.

odiseo0: Indeed, very good PR. I really liked it.

fastapi/dependencies/utils.py
```diff
@@ -535,90 +535,188 @@ async def solve_dependencies(
...
+) -> Optional[List[DependencyGetter]]:
+    """
+    Process built-in dependencies into a list of getter functions
+    """
+    getters: List[DependencyGetter] = []
+    field: ModelField
+    for field in dependant.path_params:
+
+        def get_path_param_getter(field: ModelField) -> DependencyGetter:
+            def get_param(context: DependencySolverContext) -> None:
```
sk-, line 564: Why are all these getters defined inline? As they don't seem to be using any local state, could you move them out and declare them at the top level?

Note that this is specially necessary for getters defined in a loop as in each iteration it will create a new function, generating more memory garbage.

fastapi/dependencies/utils.py
```diff
@@ -535,90 +535,188 @@ async def solve_dependencies(
...
-    values.update(cookie_values)
-    errors += path_errors + query_errors + header_errors + cookie_errors
+        context.dependency_cache[cache_key] = solved
+
+    # solve built-in dependencies
+    if dependant.dependency_getters:
+        context.values = values
+        context.errors = errors
+        for dependency_getter in dependant.dependency_getters:
+            if inspect.iscoroutinefunction(dependency_getter):
```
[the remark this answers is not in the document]
kristjanvalur: Sure, we can do that.

fastapi/dependencies/utils.py
kristjanvalur: It's a style decision.  My original implementation attempted to not use the outer function at all (e.g. `get_path_param_getter()`, but crate the closure directly (e.g. `get_param()`, but that doesn't work in a loop neccessitating the outer helper function.  I kept it at the same place as the closure to keep all the logic in the same place.

You need not worry about "memory garbage".  The outer function is assigned to a local variable and gets destroyed when the "get_dependent_dependency_getters()" exits.  Each iteration just redefines the function.  But I can remove it out of the loop if you want.  It will just reduce the locality of the logic a bit.  Python is nice like that, local functions definitions are just fine.

This also happens during the setup of the endpoints so it is not critical performance-wise.

But I'm happy to move them out to the top level if you think that style is better.

fastapi/dependencies/utils.py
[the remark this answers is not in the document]
kristjanvalur: Ah, I'm looking at the code, and these are the "built-in" dependencies (Headers, params, etc)
There is only one type of `async` built-in dependency, and that tis the `get_body_params()`.
This is used for `Body()` type parameters.  FastAPI already schedules multipe FILE reads internally...  But I suppose we could use the `anyio` in this place..  Again, I'd like that to be a separate PR if someone thinks it is worthwhile.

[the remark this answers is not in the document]
kristjanvalur: There are so few comments in the code, it takes a lot of effort to reverse-engineer this :)

fastapi/dependencies/utils.py
kristjanvalur: I could also refactor this to use `functools.partial()` instead of closures.

kristjanvalur:
Is there any interest of reducing call overhead for FastAPI endpoint dependencies?

nonnibunk:
Could we get this PR merged, I'd love to get this into my project.

Tishka17:
If you are interested in faster dependency solving, I can suggest my project which can be used with fastapi together (or without it, if you need)

https://github.com/reagento/dishka/tree/develop

YuriiMotov:
@kristjanvalur, thank you for your efforts!

Do you think this is still relevant to current version of FastAPI?

kristjanvalur:
> Do you think this is still relevant to current version of FastAPI?

I haven't looked at this for over two years, so I don't know if it is relevant anymore. But there were low hanging fruits in my original PR, namely the traversal bugs I mention.  They might be still there.  Perhaps they should, in retrospect, have been put in as a separate PR.

archfz:
@tiangolo can be closed in favor of https://github.com/fastapi/fastapi/pull/13756

tiangolo:

Thank you! 🍰

ranjanprasad96:
@tiangolo We are seeing lot of perf problems when the dependency chain depth is huge. It seems like the solve_dependencies method is not as performant when the depth is huge. Is this planned to be fixed?

PR #5573, opened 2022-11-02
return authorization token in OpenIdConnect

LeoQuote:
In all oauth based security middlewares, the token is directly returned, IMO this should be the same in oidc , which is just oauth2 + a bunch of service discovery.

This could be a breaking change, or even wrong? In my case I would return the token directly because I didnt see any other authorization header other than bearer header


LeoQuote:
Checked the test suites found this is intentional, closing

PR #5584, opened 2022-11-03
✏ Fix typo in `docs/en/docs/tutorial/security/oauth2-jwt.md`

vivekashok1221:


tiangolo:
Good catch, thank you! :mag:

PR #5595, opened 2022-11-07
Validate Path duplication on added routes

### LINKED ISSUE #5543: Two functions with one path produce inconsistent results
### First Check

### Commit to Help

### Example Code

```python
from fastapi import FastAPI

app = FastAPI()


@api.get("/foo")
async def foo1() -> dict:
    """Foo #1"""
    return {"content": "First foo"}


@api.get("/foo")
async def foo2() -> dict:
    """Foo #2"""
    return {"content": "Second foo"}
```

### Description

* `GET /openapi.json` (and consequently Swagger/Redoc) refer to `/foo` with `Foo #2` documentation.
* `GET /foo` returns `First foo`.
* Expected `GET /foo` to return `Second foo` _or_ `GET /openapi.json` to return `Foo #1` documentation. As it stands, the results of overriding a path are inconsistent.

### Operating System

Linux

### Operating System Details

Debian 11.5

### FastAPI Version

0.85.1

### Python Version

3.10.8

### Additional Context

As for prior art, searching wider brings up https://stackoverflow.com/questions/72219039/how-to-overwrite-extend-fastapi-apirouter-route, though this doesn't address the inconsistency found.

iudeen:
Fixes: https://github.com/tiangolo/fastapi/issues/5543

- [x] Add basic tests to the PR. 
- [x] Add and verify websocket routes
- [x] Add prefix conditions to tests
- [x] Improve coverage to 100%

This pull request introduces a new error handling mechanism to prevent duplicate route definitions in FastAPI. The changes include adding a new custom exception, updating the routing logic to check for duplicate routes, and adding tests to ensure the new functionality works as expected.

### New Error Handling Mechanism:

* Added `RouteAlreadyExistsError` class to handle duplicate route definitions. (`fastapi/exceptions.py`)

### Routing Logic Updates:

* Introduced `get_path_hash_val` function to generate a unique hash for each route based on its path and methods. (`fastapi/dependencies/utils.py`)
* Updated `APIRouter` and `FastAPI` classes to store and check route hashes, raising `RouteAlreadyExistsError` if a duplicate route is detected. (`fastapi/routing.py`) [[1]](diffhunk://#diff-4d573079004a9f3d148baa4658e68e82b8a3d1a95d603fee8177daa92cf65c93R571) [[2]](diffhunk://#diff-4d573079004a9f3d148baa4658e68e82b8a3d1a95d603fee8177daa92cf65c93R864) [[3]](diffhunk://#diff-4d573079004a9f3d148baa4658e68e82b8a3d1a95d603fee8177daa92cf65c93R965-R967) [[4]](diffhunk://#diff-4d573079004a9f3d148baa4658e68e82b8a3d1a95d603fee8177daa92cf65c93R1051-R1054) [[5]](diffhunk://#diff-4d573079004a9f3d148baa4658e68e82b8a3d1a95d603fee8177daa92cf65c93R1345-R1348)

### Tests:

* Added comprehensive tests to verify that the application raises `RouteAlreadyExistsError` when duplicate routes are defined. (`tests/test_include_duplicate_path_route.py`)


iudeen:
I am not sure how to cover the dummy routes :/

iudeen:
I'm also in a dilemma if this would be ideal if it gets handled in Starlette. Any suggestions? 🤔

tiangolo:
📝 Docs preview for commit a3357a72fb4340f568ef69aa2d3abc3857c2bb22 at: https://6494522b1e30803212194d96--fastapi.netlify.app

iudeen:
Is this a good approach? 🤔

iudeen:
@tiangolo, can you let me know if this approach is right? If not I can take suggestions :)

fastapi/routing.py
```diff
@@ -968,6 +975,10 @@ def add_api_websocket_route(
             dependencies=current_dependencies,
             dependency_overrides_provider=self.dependency_overrides_provider,
         )
+        hash_val = get_path_hash_val(route.path)
+        if hash_val in self.added_routes:
+            raise RouteAlreadyExistsError(route.name)
```
andreygaag, line 980: We can do it without additional storage, something like this:
```python3
is_route_exist = [exist_route 
                  for exist_route in self.routes
                  if route.path == self.path and route.methods == self.methods]
if is_route_exist:
    raise RouteAlreadyExistsError(route.name)

andreygaag (changes requested):

fastapi/routing.py
iudeen: Not sure if this would work. `self.path` is not in the context of this function.

YuriiMotov (commented):
I also tried to approach this but in a different way. My idea was to create a function that would check for overlapped paths, so that it could be used in CI\CD.

I came to a conclusion that this is quite complex issue. There are so many things to consider. Like paths with `{item_id:str}` or `{path:path}`.
Also, in some cases people rely on this behavior: it's useful for API versioning, when you define only updated paths in V2 router and then include V1 router. This way V2 router will re-define several paths, but fall-back to versions from V1 router.

Anyway, I think the idea with path hash will not work here.
I suggest we close this for now.
You can probably implement this as a third-party tool first and invite people to test it.

@iudeen, thanks for your efforts!

tiangolo:
I agree that it's probably better to try this out externally first, I fear there could be many hidden corner cases that could break because they assume the current behavior.

On the other hand, I also have a planned refactor for the routers that should allow creating a custom router with this logic in a way that can be shared in an internal codebase without it having to be in FastAPI itself.

For now I'll pass on this one, but thanks a lot for the effort! :coffee:

PR #5598, opened 2022-11-08
fix: update middleware stack order allow the AsyncExitStackMiddleware to run before user_middleware

MrSalman333:
fixing https://github.com/tiangolo/fastapi/issues/5597


MrSalman333:
the fix will come from upstream starlette this PR is not needed

tiangolo:
Thanks for coming back to close the PR! ☕

PR #5608, opened 2022-11-10
Add missing return type hint

artofhuman:


tiangolo:
Thanks! But it's actually not necessary, see here: https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods

artofhuman:
Ok, thanks )

PR #5622, opened 2022-11-11
merge nonparam security dependency scopes into endpoint parameter security scopes

kristjanvalur:
This PR adds the possibility to declare `Security` `scopes` at the `FastAPI` or `APIRouter` level as *parameterless* dependencies, and then have the scopes **prepended** to any scopes on the same dependencies that are endpoint _parameters_.

This allows refining/adding scopes at the endpoint to already required parameterless scopes.

For example, the *app* can have a `dependencies=[Security(my_security, scopes=["live"])]` parameter, and the endpoint can specify
```
def my_endpoint(user=Security(my_security, scopes=["client"]): pass
```
causing the `my_security()` dependency to be called **once** with scopes `["live", "client']`.

This allows setting general scopes at higher levels and refine them at individual endpoints, alloweing better _factoring_ and DRY.

It also reduces the calling overhead of Scopes dependencies.  The above example, without this PR, would cause __two__ calls to the `my_security()` dependency.  This PR causes it to be called only __once__.

An added benefit of this is that it allows a **least privilege** approach to scopes.  if the top level scopes are, for example `["client, "admin"]`, requiring both scopes for all endpoints, an individual endpoint could then add the scope `["-admin"]`, causing the security dependency to be called with scopes `["client", "admin", "-admin"]`.  A suitably written dependency could then interpret this as _not_ requiring the `admin` scope on this endpoint.    See also #5433 which also merges scopes in the hierarchy of `dependencies` properties, and allows to _remove_ parametereless dependencies on individual endpoints.

As a bonus, parameterless dependencies which are also direct or indirect parameter dependencies are not applied to the endpoint, reducing invokation overhead.

### DIFF
--- a/docs/en/docs/advanced/security/oauth2-scopes.md
+++ b/docs/en/docs/advanced/security/oauth2-scopes.md
@@ -266,4 +266,15 @@ The most secure is the code flow, but is more complex to implement as it require
 
 ## `Security` in decorator `dependencies`
 
-The same way you can define a `list` of `Depends` in the decorator's `dependencies` parameter (as explained in [Dependencies in path operation decorators](../../tutorial/dependencies/dependencies-in-path-operation-decorators.md){.internal-link target=_blank}), you could also use `Security` with `scopes` there.
+The same way you can define a `list` of `Depends` in the decorator's `dependencies` parameter (as explained in [Dependencies in path operation decorators](../../tutorial/dependencies/dependencies-in-path-operation-decorators.md){.internal-link target=_blank}), you could also use `Security` with `scopes` there.  Similarly, they can be added
+as `dependencies` arguments to the `FastAPI` application object, and to `APIRouter` objects.
+
+### Merging of scopes from decorator `dependencies`
+
+If you add a `Security` into a decorator's `dependencies` parameter, and that same function is also present as a `Security` or `Depends` parameter for the endpoint, it results
+in the `scopes` lists being joined: The scopes from the decorator get **prepended** to any scopes of the endpoint dependencies (a `Depends` object is equivalent to a
+`Security` object with an empty list for its `scopes` argument).  This works recursively:  Wherever the same `Security` is seen in the dependency hierarchy
+of the endpoint argument dependencies, the scopes get joined.
+
+Using this mechanism, you can add a single scope requirement at the root of the **app** (using the `FastAPI` `dependencies` argument), or at individual `APIRouter` instances,
+and then refine the scopes required on the endpoint dependencies.
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -57,4 +57,7 @@ class Dependant:
         # Store the path to be able to re-generate a dependable from it in overrides
         self.path = path
         # Save the cache key at creation to optimize performance
+        self.set_cache_key()
+
+    def set_cache_key(self) -> None:
         self.cache_key = (self.call, tuple(sorted(set(self.security_scopes or []))))
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -32,6 +32,7 @@ class Dependant:
         background_tasks_param_name: Optional[str] = None,
         security_scopes_param_name: Optional[str] = None,
         security_scopes: Optional[List[str]] = None,
+        dependency_scopes: Optional[List[str]] = None,
         use_cache: bool = True,
         path: Optional[str] = None,
     ) -> None:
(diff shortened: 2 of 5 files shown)


YuriiMotov:
See my comment to #5667, it's also relevant for this PR: https://github.com/fastapi/fastapi/pull/5667#pullrequestreview-3310148297

As for adding scopes like `["-admin"]` - just a note: they all will appear in openapi schema

YuriiMotov:
I already commented on this topic in [this comment](https://github.com/fastapi/fastapi/pull/5667#pullrequestreview-3310148297), but will summarize my thoughts here.

## Idea of this PR

**There are 2 issues this PR aims to solve:**

1. **Reduce the number of times security dependencies with scopes are being solved.**

```python
async def get_user(scopes: SecurityScopes, token: Annotated[str, Depends(auth)]):
    # ... Expensive token decoding operation
    pass

app = FastAPI(dependencies=[Security(get_user, scopes=["a"])])

@app.get("/", dependencies=[Security(get_user, scopes=["b"])])
async def get_():
    pass
```

Currently (on `master`), `get_user` will be called twice - with `scopes=["a"]` and with `scopes=["b"]` separately.

Assuming we want to have “AND” logic, would be nice to call `get_user` only once with merged scopes.

2. **Provide the ability to override scopes, defined on router level, for specific path operation.**

(I personally don't really like this idea, but don't mind)

```python
async def get_user(scopes: SecurityScopes):
    remove_scopes = {sc for sc in scopes.scopes if sc.startswith("-")}
    real_scopes = {sc for sc in scopes.scopes if sc not in remove_scopes}
    # ... Expensive token decoding operation
    pass  # use real_scopes instead of scopes

app = FastAPI(dependencies=[Security(get_user, scopes=["a"])])

@app.get("/", dependencies=[Security(get_user, scopes=["-a", "b"])])
async def get_():
    pass
```

In the code example above, we add `"-a"` scope to indicate that `"a"` scope added on router level should be ignored.

Currently this will not work because scopes are not merged.

We can probably think about introducing the `scopes_exclude` parameter for this:

```python
@app.get("/", dependencies=[Security(get_user, scopes=["b"], scopes_exclude=["a"])])
    ...
```

(It will still not work without merging scopes)

## The problem of current implementation (this PR):

I personally like the idea of being able to make scopes merged and this way to reduce the number of dependency calls. But the implementation suggested in this PR has issue - it will break the use case when we want to use the dependency several times with different scopes with “OR” logic (e.g. several authentication options with `auto_error=False`).


<details><summary>See code in the details</summary>

```py
from typing import Annotated

from fastapi.security import OAuth2PasswordBearer, SecurityScopes

counter = {"cnt": 0, "calls": []}

auth = OAuth2PasswordBearer("token", scopes={"read": "...", "write": "..."}, auto_error=False)

async def get_user(scopes: SecurityScopes, token: Annotated[str, Depends(auth)]):
    # ... Expensive token decoding operation
    counter["cnt"] += 1
    counter["calls"].append(scopes.scope_str)

    token_scopes = token.split()  # Imagine we decoded token

    for scope_required in scopes.scopes:
        if scope_required not in token_scopes:
            return None
    return "user"


app = FastAPI()


@app.get("/dependencies-with-or-logic")
async def get_dependencies_with_or_logic(
    user_can_read: Annotated[str, Security(get_user, scopes=["read"])],
    user_can_write: Annotated[str, Security(get_user, scopes=["write"])],
):
    res = {}
    if user_can_read:
        res["read"] = "successfully read"
    if user_can_write:
        res["write"] = "successfully wrote"
    return res


client = TestClient(app)

def test_can_only_read():
    response = client.get("/dependencies-with-or-logic", headers={"Authorization": "Bearer read"})
    assert response.json() == {"read": "successfully read"}
```

</details>


## Idea

I would suggest considering another approach - introduce the **“singleton”** dependencies, that will be called only once per request no matter how many times they were added on different levels with different scopes. For such dependencies we can merge scopes:

```python
async def get_user(scopes: SecurityScopes, token: Annotated[str, Depends(auth)]):
    # ... Expensive token decoding operation
    pass

app = FastAPI(dependencies=[Security(get_user, scopes=["a"], singleton=True)])

@app.get("/", dependencies=[Security(get_user, scopes=["b"], singleton=True)])
async def get_():
    pass
```

It could also be `merge_scopes` parameter.

Or, we can create special function like `SecuritySgt()`:

```python
async def get_user(scopes: SecurityScopes, token: Annotated[str, Depends(auth)]):
    # ... Expensive token decoding operation
    pass

app = FastAPI(dependencies=[SecuritySgt(get_user, scopes=["a"])])

@app.get("/", dependencies=[SecuritySgt(get_user, scopes=["b"])])
async def get_():
    pass
```

Setting `use_cache=False` for such dependencies should be restricted (it doesn't make sense)

---

Let's wait for Sebastian to validate the idea and decide whether we need this or not

tiangolo:

Thank you! 🍰

PR #5629, opened 2022-11-13
✨ Re-export Starlette's `WebSocketException` and add it to docs

tiangolo:
✨ Re-export Starlette's `WebSocketException` and add it to docs


Kludex:
Gracias :)

PR #5634, opened 2022-11-13
⬆️ Upgrade and relax dependencies for extras "all"

tiangolo:
⬆️ Upgrade and relax dependencies for extras "all"

Because @ofek [convinced me](https://github.com/dependabot/dependabot-core/issues/3290#issuecomment-1289814612) with this link: https://iscinumpy.dev/post/bound-version-constraints/ ( by @henryiii)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,15 +92,15 @@ dev = [
     "pre-commit >=2.17.0,<3.0.0",
 ]
 all = [
-    "httpx >=0.23.0,<0.24.0",
-    "jinja2 >=2.11.2,<4.0.0",
-    "python-multipart >=0.0.5,<0.0.6",
-    "itsdangerous >=1.1.0,<3.0.0",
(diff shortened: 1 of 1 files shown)


sallner:
Thanks for relaxing some of the restrictions. May I ask, what convinced you to keep the other optional dependencies restricted in the old way? I can understand that you keep `pydantic` and `starlette` restricted, but now we have `pyyaml` some times with upper bounds and once without.

tiangolo:
Thanks for the feedback! I'm glad this was well received :sweat_smile: 

@sallner I didn't uncap the other extras because those shouldn't really be extras, they are development dependencies, for tests, docs, and local development. I should not have them exposed in the released versions/wheels.

I plan on removing those extras that shouldn't be public and replace them with something else, probably `requirements.txt` files with pinned dependencies that Dependabot can suggest to update later.

PR #5662, opened 2022-11-19
Update tutorial001_py310.py

subedikushal:
It is simpler to do it directly rather than via update method. This also eliminates the overhead of constructing another dictionary and calling a method, so should slightly improve performance. [ Read at Sourcery]


iudeen (approved):

tiangolo:
Thanks! In that case, it would also need an update in file for the Python 3.6+ version, not just the 3.10 one.

PR #5667, opened 2022-11-20
Combine inherited nonparam Security() objects

kristjanvalur:
A simplified version of the rejected #5433 , this one combines the list of inherited "nonparam" dependencies.  Multiple instances of a `Security` dependency are combined, their `scopes` joined into a single scope list.

This serves two purposes:
1) In case there are many different scope checks set up, combine the invokation into one dependency call for efficiency.
2) Allow a security dependency to interpret the combined list of scopes, for example, adding semantics where specially marked scopes can _disable_ a previously added scope.

Example:
```python
def my_security(scopes: SecurityScopes):
    active = set()
    for scope in scopes.scopes:
        if not scope.startswith("-"):
            active.add(scope)
        else:
            active.discard(scope[1:])
    apply_scopes(sorted(active))
```

see also #5622 which proposes to join nonparam dependencies with parameter dependencies in a similar manner.


yezz123 (approved):

YuriiMotov (commented):
@kristjanvalur, thank you for your contribution!

Although it might be interesting idea to merge scopes and reduce the number of security dependency calls, there are several issues with this implementation.

The idea of this PR is to make the `get_user` dependency being called only once with `scopes=["a", "b"]` in the following code example.

```py
async def get_user(scopes: SecurityScopes, token: Annotated[str, Depends(auth)]):
    # ... Expensive token decoding operation
    pass

app = FastAPI(dependencies=[Security(get_user, scopes=["a"])])

@app.get("/app-and-decorator", dependencies=[Security(get_user, scopes=["b"])])
async def get_app_and_decorator():
    pass
```


<details><summary>See test code in the details</summary>

```py
from typing import Annotated

from fastapi.security import OAuth2PasswordBearer, SecurityScopes

counter = {"cnt": 0, "calls": []}

auth = OAuth2PasswordBearer("token", scopes={"a": "a", "b": "b", "c": "c"})

async def get_user(scopes: SecurityScopes, token: Annotated[str, Depends(auth)]):
    # ... Expensive token decoding operation
    counter["cnt"] += 1
    counter["calls"].append(scopes.scope_str)
    return scopes.scope_str

app = FastAPI(dependencies=[Security(get_user, scopes=["a"])])

@app.get("/app-and-decorator", dependencies=[Security(get_user, scopes=["b"])])
async def get_app_and_decorator():
    pass

client = TestClient(app)
client.headers["Authorization"] = "bearer 123"

def test_app_and_decorator():
    counter["calls"] = []
    counter["cnt"] = 0
    client.get("/app-and-decorator")
    assert counter == {
        "calls": ["a b"],
        "cnt": 1,
    }
```

</details>

This test will pass on the branch of this PR, but fail on `master` (`get_user` will be called twice: with `scopes=["a"]` and with `scopes=["b"]`).
Looks great, we reduced the number times the expensive token decode operation is executed!

But this will not work with dependencies, specified as parameters of endpoint function, or as sub-dependencies (upd: I see this is intended, and there is #5622 that extends this behavior to "param-dependencies")..

<details><summary>See code in the details</summary>

```py
from typing import Annotated

from fastapi.security import OAuth2PasswordBearer, SecurityScopes

counter = {"cnt": 0, "calls": []}

auth = OAuth2PasswordBearer("token", scopes={"a": "...", "b": "...", "c": "...", "me": "..."})

async def get_user(scopes: SecurityScopes, token: Annotated[str, Depends(auth)]):
    # ... Expensive token decoding operation

    counter["cnt"] += 1
    counter["calls"].append(scopes.scope_str)
    return scopes.scope_str


async def get_active_user(user: Annotated[str, Security(get_user, scopes=["me"])]):
    pass


app = FastAPI(dependencies=[Security(get_user, scopes=["a"])])


@app.get(
    "/app-and-decorator-and-dependency",
    dependencies=[Security(get_user, scopes=["b"])],
)
async def get_app_and_decorator_and_dependency(
    dep_1: Annotated[str, Security(get_user, scopes=["c"])],
):
    pass


@app.get(
    "/app-and-sub-dependency-in-decorator",
    dependencies=[Depends(get_active_user)],
)
async def get_app_and_sub_dependency_in_decorator():
    pass

client = TestClient(app)
client.headers["Authorization"] = "bearer 123"

def test_app_and_decorator_and_dependency():
    counter["calls"] = []
    counter["cnt"] = 0
    client.get("/app-and-decorator-and-dependency")
    assert counter == {
        "calls": ["a b c"],
        "cnt": 1,
    }

def test_app_and_sub_dependency_in_decorator():
    counter["calls"] = []
    counter["cnt"] = 0
    client.get("/app-and-sub-dependency-in-decorator")
    assert counter == {
        "calls": ["a me"],
        "cnt": 1,
    }
```

</details>

Both of the tests above fail on the branch of this PR.

And, if we fix the tests above (make merging working for sub-dependencies and for dependencies in the parameters of endpoint function), we will have one more problem - what if we want to combine dependencies with OR logic instead of AND logic?

<details><summary>See code in the details</summary>

```py
from typing import Annotated

from fastapi.security import OAuth2PasswordBearer, SecurityScopes

counter = {"cnt": 0, "calls": []}

auth = OAuth2PasswordBearer("token", scopes={"read": "...", "write": "..."}, auto_error=False)

async def get_user(scopes: SecurityScopes, token: Annotated[str, Depends(auth)]):
    # ... Expensive token decoding operation
    counter["cnt"] += 1
    counter["calls"].append(scopes.scope_str)

    token_scopes = token.split()  # Imagine we decoded token

    for scope_required in scopes.scopes:
        if scope_required not in token_scopes:
            return None
    return "user"


app = FastAPI()


@app.get("/dependencies-with-or-logic")
async def get_dependencies_with_or_logic(
    user_can_read: Annotated[str, Security(get_user, scopes=["read"])],
    user_can_write: Annotated[str, Security(get_user, scopes=["write"])],
):
    res = {}
    if user_can_read:
        res["read"] = "successfully read"
    if user_can_write:
        res["write"] = "successfully wrote"
    return res


client = TestClient(app)

def test_can_only_read():
    response = client.get("/dependencies-with-or-logic", headers={"Authorization": "Bearer read"})
    assert response.json() == {"read": "successfully read"}
```

</details>

The code example above currently works on both, `master` branch and on the branch of this PR, but will fail if we enable merging scopes for sub-dependencies and dependencies in the parameters of endpoint function.

One more thing - will it work well with `use_cache` parameter specified?


I'm not sure if it's worth efforts to investigate more in this direction..

@kristjanvalur, what do you think?

kristjanvalur:
> @kristjanvalur, what do you think?

Hi there.

To be honest, it has been three years since this PR, and I have not been using FastAPI a lot since then.
The original idea was, IIRC, to help us wit complex security in an application, one where security scopes could be removed as well as added.

The 'collapsing' of the scopes serves to combine multiple instances security objects into one which gets all the scopes and can then make decisions on the set of scopes, rather than having each individual object working separately.

I think this is still a fundamentally good idea, since the sum of scopes evaluated at once does not have to be the same as the sum of the evaluation of individual scopes.

But as I said, its been so long and I am no longer actively using FastAPI so I am reluctant to personally pursue this further.

YuriiMotov:
@kristjanvalur, thank you for reply!

So, I think we should close this PR (as this way the approach will be inconsistent for "non-parameter" dependencies and dependencies specified as parameters in the endpoint functions and in sub-dependencies) and focus on #5622. If we can come up with the idea on how to resolve issues that I pointed in [this comment](https://github.com/fastapi/fastapi/pull/5667#pullrequestreview-3310148297), it will be interesting improvement

kristjanvalur:
Closing this PR as per discussion.

PR #5707, opened 2022-11-28
chore: refactor concurrency.py to run sync context managers in a single thread

adriangb:
I'm hoping that this (maybe) helps resolve issues like https://github.com/MagicStack/asyncpg/issues/978

This change makes sync context managers acquire and hold onto a thread for the entirety of their execution so that we don't split `__enter__` and `__exit__` calls across threads. I think this will resolve a whole class of intermittent bugs that might crop up for anything using thread locals or context variables.

The main "con" of this approach is that we won't "free" the thread in between `__enter__` and `__exit__` which means more contention for threads in the thread pool. In other words, despite the 40 thread default limit that anyio imposes it was previously possible to run more than 40 sync context managers as FastAPI dependencies since the 41st would just have to wait for any `__enter__` to finish before it could acquire a thread pool. With this change running more than 40 sync context managers will result in a deadlock. I see three options:
1. Do nothing. If users are running into capacity limiters they should probably be managing their own threads instead of relying on automagic behavior from their web framework.
2. Raise an error if we hit the capacity limit (instead of deadlocking)
3. Allow users to pass in their own capacity limit to each APIRoute (maybe with a global one on the FastAPI object as well)
4. Both 2 and 3

### DIFF
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -50,6 +50,8 @@ async def contextmanager_in_threadpool(
         stack.enter_context(send_err)
         tg = await stack.enter_async_context(anyio.create_task_group())
         target = partial(
+            # using a partial because start_soon does not accept kw args
+            # but run_sync(..., limiter=...) _must_ be a kw argument
             run_sync, _cm_thead_worker, cm, send_res, rcv_err, limiter=limiter
         )
         tg.start_soon(target)
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -52,7 +52,12 @@ async def contextmanager_in_threadpool(
         target = partial(
             # using a partial because start_soon does not accept kw args
             # but run_sync(..., limiter=...) _must_ be a kw argument
-            run_sync, _cm_thead_worker, cm, send_res, rcv_err, limiter=limiter
+            run_sync,
+            _cm_thead_worker,
+            cm,
+            send_res,
+            rcv_err,
+            limiter=limiter,
         )
         tg.start_soon(target)
         res = await rcv_res.receive()
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -1,13 +1,7 @@
 from contextlib import AsyncExitStack as AsyncExitStack  # noqa
 from contextlib import asynccontextmanager as asynccontextmanager
 from functools import partial
-from typing import (
-    Any,
-    AsyncGenerator,
-    ContextManager,
-    Optional,
-    TypeVar,
-)
+from typing import Any, AsyncGenerator, ContextManager, Optional, TypeVar
 
 import anyio
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -12,12 +12,12 @@ from starlette.concurrency import (  # noqa
     run_until_first_complete as run_until_first_complete,
 )
 
-T = TypeVar("T")
+_T = TypeVar("_T")
 
 
 def _cm_thead_worker(
-    cm: ContextManager[T],
-    res_stream: MemoryObjectSendStream[T],
+    cm: ContextManager[_T],
+    res_stream: MemoryObjectSendStream[_T],
     err_stream: MemoryObjectReceiveStream[Optional[Exception]],
 ) -> None:
     with cm as res:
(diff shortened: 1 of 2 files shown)


iudeen:
This means more resource optimised DB connections?🤔

(just wondering)

adriangb:
Not really. It just means that sync context manager's `__enter__` and `__exit__` run in the same thread. In fact this might make things _worse_ in some cases because the thread is checked out from the thread pool for the entirety of the context managers life. But I think it also eliminates a whole class of intermittent hard to reason about or figure out bugs.

iudeen:
Make sense, better house-keeping! 

Thanks for your explanation btw! 😅

adriangb:
marked as ready, cc @tiangolo

fastapi/concurrency.py
```diff
@@ -1,40 +1,62 @@
...
-                cm.__exit__, type(e), e, None, limiter=exit_limiter
-            )
-        )
-        if not ok:
-            raise e
-    else:
-        await anyio.to_thread.run_sync(
-            cm.__exit__, None, None, None, limiter=exit_limiter
+    cm: ContextManager[T],
+    limiter: Optional[anyio.CapacityLimiter] = None,
```
adriangb, line 36: This is currently unused but IMO exposing it makes sense for future customizability. Happy to remove it if not

adriangb:
@kglee0 or @alexted mind trying out this branch? It's a shot in the dark but worth at least confirming the null hypothesis.

adriangb:
Awesome! Did you get to test this branch?

yezz123 (approved):

jakekaplan:
Is this pr possible to merge sometime soon? Coming here from this thread: https://github.com/MagicStack/asyncpg/issues/978

adriangb:
For visibility: https://github.com/tiangolo/fastapi/discussions/8424#discussioncomment-6143037

YuriiMotov:
The test from this PR doesn't fail with current FastAPI version (0.116.0):

```python
import threading
from contextvars import ContextVar

ctx: ContextVar[str] = ContextVar("ctx")

async def context_same_thread():
    # test both a real world use case and a
    # simpler synthetic check
    ctx.set("foo")
    ident = threading.get_ident()
    yield
    assert ctx.get() == "foo"
    assert ident == threading.get_ident()

app = FastAPI()

@app.get("/check_cm_same_thread")
async def check_cm_same_thread(_=Depends(context_same_thread)):
    return None

def test_sync_runs_in_single_thread():
    client = TestClient(app)
    response = client.get("/check_cm_same_thread")
    response.raise_for_status()
```

@adriangb, could you please double-check? Seems that it has been fixed.

**UPD:** shouldn't `context_same_thread` be sync?

Kludex:
It's unfair to expect users to reply when review takes so long.

Isn't there a less aggressive strategy? Like pinging the user again a couple of weeks?

YuriiMotov:
> It's unfair to expect users to reply when review takes so long.
> Isn't there a less aggressive strategy? Like pinging the user again a couple of weeks?

@Kludex, I agree we should change this.
We will update the script to add more explanations of how we work on external PRs and why we do it this way, also we will add a reminder 1 week before closing PR.

What should we do with this PR now?
@adriangb, are you still interested in this PR?
Could you please take a look at [my comment](https://github.com/fastapi/fastapi/pull/5707#issuecomment-3052988824)?

tiangolo:
I'm realizing that this was related to a change done a while ago with dependencies with yield, this was solved in 0.74.0 (https://fastapi.tiangolo.com/release-notes/#0740)

The exit code after `yield` in a dependency was being executed in a different async context (contextvars context), so the value of contextvars was different than the original.

This was because custom middlewares would internally create an AnyIO task group, which creates a new context.

That logic was then refactored once again in 0.106.0 (https://fastapi.tiangolo.com/release-notes/#01060).

And is now refactored again here :sweat_smile: : https://github.com/fastapi/fastapi/pull/14099 , which will be released in version 0.118.0

---

But the most important point is, a test for this use case was added (in the first PR in the version described above) here: https://github.com/fastapi/fastapi/blob/master/tests/test_dependency_contextvars.py

So that should ensure this is handled/fixed and stays handled and fixed. :rocket: 

Given that, I'll now close this one, but if anyone faces a similar problem again, please create a new discussion with an example to replicate it so we can take a look. :nerd_face: 

Thanks all! :cake:

Issue #755, opened 2019-11-30
[BUG] asyncpg uuid implementation not recognized by fastapi because of too strict typechecking


RmStorm:
### Describe the bug

An alternative implementation of the `UUID` type is not serialized by fastapi but throws an error instead. This is because it is of type `asyncpg.pgproto.pgproto.UUID` instead of type `uuid.UUID`. However it is implemented in such a way that it does the exact same things as the regular UUID but just faster. Also,
```python
isinstance(asyncpg.pgproto.pgproto.UUID(), uuid.UUID) == True
```
this should make fastapi able to recognize it as a uuid.

### To Reproduce
1. Create a file called uuid_error.py with:

```Python
import uuid

from fastapi import FastAPI
from asyncpg.pgproto import pgproto

app = FastAPI()


@app.get("/normal_uuid")
def return_normal_uuid():
    regular_uuid = uuid.UUID("a10ff360-3b1e-4984-a26f-d3ab460bdb51")
    assert isinstance(regular_uuid, uuid.UUID)
    return {"normal_uuid": regular_uuid}


@app.get("/fast_uuid")
def return_fast_uuid():
    asyncpg_uuid = pgproto.UUID("a10ff360-3b1e-4984-a26f-d3ab460bdb51")
    assert isinstance(asyncpg_uuid, uuid.UUID)
    return {"fast_uuid": asyncpg_uuid}
```
2. Run it with uvicorn uuid_error:app
3. Open the browser and call the endpoint `/fast_uuid`.
4. It returns 'internal server error' .
5. But I expected it to return `{"fast_uuid":"a10ff360-3b1e-4984-a26f-d3ab460bdb51"}`.

### Expected behavior

I expected fastapi to jsonify the uuid and send, but instead it returns 'internal server error'. The precise error being thrown in the server is:
```
ValueError: [KeyError(<class 'asyncpg.pgproto.pgproto.UUID'>), TypeError("'asyncpg.pgproto.pgproto.UUID' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]
```

### Environment

- OS: Linux Mint 19 Tara
- FastAPI Version: 0.44.0
- Python version: 3.8.0

### Additional context

I initially made an issue in the asyncpg repo asking them if they could fix it: https://github.com/MagicStack/asyncpg/issues/512 I think it's usefull as additional context for this issue.
The problem can be solved here in the fastapi code by improving the code a little. I'm working on a PR, this line seems to be the issue:
[...]

PR #5718, opened 2022-11-30
♻️ Refactor `jsonable_encoder` to optimize subclass encoder resolution using MRO

ramnes:
The current implementation doesn't handle subclass instances when `fastapi.encoders.ENCODERS_BY_TYPE` is modified after `fastapi.encoders` import.

For example, in this example we could expect the `/` route to return an ObjectId string. But currently, it returns `{}` instead:

```python
import bson
import uvicorn
from fastapi import FastAPI
from fastapi.encoders import ENCODERS_BY_TYPE


ENCODERS_BY_TYPE |= {bson.ObjectId: str}

app = FastAPI()


class CustomObjectId(bson.ObjectId):
    pass


@app.get("/")
def get_child():
    return CustomObjectId()


uvicorn.run(app)
```

This PR resolves this (and simplifies the code, as well).

Continuation of #755 and #756.

cc @RmStorm

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -1,6 +1,6 @@
 import dataclasses
 import datetime
-from collections import defaultdict, deque
+from collections import deque
 from collections.abc import Callable
 from decimal import Decimal
 from enum import Enum
@@ -93,20 +93,6 @@ ENCODERS_BY_TYPE: dict[type[Any], Callable[[Any], Any]] = {
     Url: str,
     AnyUrl: str,
 }
-
-
-def generate_encoders_by_class_tuples(
-    type_encoder_map: dict[Any, Callable[[Any], Any]],
-) -> dict[Callable[[Any], Any], tuple[Any, ...]]:
-    encoders_by_class_tuples: dict[Callable[[Any], Any], tuple[Any, ...]] = defaultdict(
-        tuple
-    )
-    for type_, encoder in type_encoder_map.items():
-        encoders_by_class_tuples[encoder] += (type_,)
-    return encoders_by_class_tuples
-
-
-encoders_by_class_tuples = generate_encoders_by_class_tuples(ENCODERS_BY_TYPE)
 
 
 def jsonable_encoder(
@@ -314,11 +300,9 @@ def jsonable_encoder(
             )
         return encoded_list
 
-    if type(obj) in ENCODERS_BY_TYPE:
-        return ENCODERS_BY_TYPE[type(obj)](obj)
-    for encoder, classes_tuple in encoders_by_class_tuples.items():
-        if isinstance(obj, classes_tuple):
-            return encoder(obj)
+    for base in obj.__class__.__mro__[:-1]:
+        if base in ENCODERS_BY_TYPE:
+            return ENCODERS_BY_TYPE[base](obj)
     if is_pydantic_v1_model_instance(obj):
         raise PydanticV1NotSupportedError(
             "pydantic.v1 models are no longer supported by FastAPI."


ramnes:
@tiangolo While it wasn't my first purpose, it also seems to make `jsonable_encoder` ~40% faster on the benchmark you used in #756:

```bash
(.venv) $ # without #5718
(.venv) $ pip install -q -I https://github.com/tiangolo/fastapi/archive/83bcdc40d853e50752d4d4a30bbc8b23a8dbc4c1.zip
(.venv) $ hyperfine --warmup=3 'python performance.py'
Benchmark 1: python performance.py
  Time (mean ± σ):     27.843 s ±  0.308 s    [User: 27.651 s, System: 0.162 s]
  Range (min … max):   27.555 s … 28.495 s    10 runs

(.venv) $ # with #5718
(.venv) $ pip install -q -I https://github.com/Refty/fastapi/archive/1eecf8dda26e23c7f3b6dcd30f9a27171f5ae217.zip
(.venv) $ hyperfine --warmup=3 'python performance.py'
Benchmark 1: python performance.py
  Time (mean ± σ):     16.105 s ±  0.247 s    [User: 15.981 s, System: 0.110 s]
  Range (min … max):   15.907 s … 16.663 s    10 runs

```

If you want to replicate, I just changed the ranges to make it work better with hyperfine:

```python
import time
import uuid


class SubStr(str):
    pass


class MyUuid:
    def __init__(self, uuid_string: str):
        self.uuid = uuid_string

    def __str__(self):
        return self.uuid

    @property
    def __class__(self):
        return uuid.UUID

    @property
    def __dict__(self):
        """Spoof a missing __dict__ by raising TypeError, this is how
        asyncpg.pgroto.pgproto.UUID behaves"""
        raise TypeError("vars() argument must have __dict__ attribute")


test_input_3 = [MyUuid(str(uuid.uuid4())) for _ in range(1_000)]

for _ in range(10_000):
    jsonable_encoder(test_input_3)
```

PraveenNanda124 (commented):
Good work

ramnes:
I just updated the first message with a code example to make the problem clearer.

@tiangolo Any hint on the direction that this PR should take?

ramnes:
Also, FWIW, Pydantic uses the same approach: https://github.com/pydantic/pydantic/blob/f55515820a407c5d31c45a7b0307925eecf21f34/pydantic/json.py#L82-L90

dkatzan (approved):
LGTM
I was facing a similar issue, and wondered why did fastapi took a different approach on this from pydantic 
consider using the snippet u added to the PR as a base for a test for this fix

ramnes:
Thanks for the feedback @dkatzan! Yeah, I can definitely add a test or anything else that would help; I'd prefer a confirmation that it's the correct approach before putting more work into that PR though, especially since @tiangolo wrote most of the code I'm removing. 😅

ramnes:
rebased

ramnes:
rebased

ramnes:
rebased

ramnes:
rebased

fastapi/encoders.py
```diff
@@ -312,11 +298,9 @@ def jsonable_encoder(
...
         return encoded_list
 
-    if type(obj) in ENCODERS_BY_TYPE:
-        return ENCODERS_BY_TYPE[type(obj)](obj)
-    for encoder, classes_tuple in encoders_by_class_tuples.items():
-        if isinstance(obj, classes_tuple):
-            return encoder(obj)
+    for base in obj.__class__.__mro__[:-1]:
+        if base in ENCODERS_BY_TYPE:
+            return ENCODERS_BY_TYPE[base](obj)
```
dolfinus, line 303: Are you sure that `ENCODERS_BY_TYPE` always include the root class of the inheritance tree, and cannot include more nested classes?

ramnes: I'm not sure to understand, what problem do you have in mind? Can you give an example?

ramnes:
rebased

fastapi/encoders.py
dolfinus: Sorry, my bad

ramnes: No problem! Better safe than sorry. :)

YuriiMotov (approved):
LGTM

I can confirm positive performance impact:
Before changes:
```
15578851161
15829424601
15774882936
15638053304
15637803009
```
After changes:
```
12273862558
12398896355
12274255815
12209654364
12177034769
```
So, about 21% difference on [this code example](https://github.com/fastapi/fastapi/pull/5718#issuecomment-1340868065).

With these changes the algorithm changes a bit:
* Before changes, the order of items in `ENCODERS_BY_TYPE` matter.
* After changes, the order in object's `mro` matters.

See code example in details.

<details>

```python
import fastapi.encoders
from fastapi.encoders import ENCODERS_BY_TYPE, jsonable_encoder

class MyType1:
    pass

class MyType2:
    pass

ENCODERS_BY_TYPE[MyType2] = lambda x: f"Type2: {x}"
ENCODERS_BY_TYPE[MyType1] = lambda x: f"Type1: {x}"

# Condition is needed to make it run with new version
if hasattr(fastapi.encoders, "encoders_by_class_tuples"):
    fastapi.encoders.encoders_by_class_tuples = (
        fastapi.encoders.generate_encoders_by_class_tuples(ENCODERS_BY_TYPE)
    )

class MyType3(MyType1, MyType2):
    def __init__(self, x: int):
        self.x = x
    def __str__(self):
        return str(self.x)

def test_():
    res = jsonable_encoder(MyType3(x=123))
    assert res == "Type2: 123"  # AssertionError: assert 'Type1: 123' == 'Type2: 123'
```

</details>

I didn't manage to come up with real-life use-case where this may brake the logic.

YuriiMotov:
@ramnes, thanks for your efforts!

Could you please update the branch?

What if we rename this to something like `Refactor jsonable_encoder to use obj.__class__.__mro__`?

ramnes:
> @ramnes, thanks for your efforts!

thanks for the kind words :) 

> Could you please update the branch?

done, rebased

> What if we rename this to something like `Refactor jsonable_encoder to use obj.__class__.__mro__`?

I tend to explain *why* I do things rather than *what* I'm doing in my commit messages, but everything works for me if it gets this PR in the tree at the end of the day :)

YuriiMotov:
@ramnes, thanks!

> I tend to explain _why_ I do things rather than _what_ I'm doing in my commit messages

We have some [rules](https://fastapi.tiangolo.com/management-tasks/#edit-pr-titles) for PR titles. According to them we should first describe the goal, not the reason or actions.

I also think we should edit the description to focus on performance optimization rather than on fixing the issue described in the code example. I think the thing that is described in code example is not how it's supposed to work. You can pass the custom encoders using `custom_encoder` parameter, can't you?

ramnes:
> We have some [rules](https://fastapi.tiangolo.com/management-tasks/#edit-pr-titles) for PR titles. According to them we should first describe the goal, not the reason or actions.

LGTM!

> I also think we should edit the description to focus on performance optimization rather than on fixing the issue described in the code example.

that was not the original intention but again, happy do anything that would help getting this merged after three years :)

> I think the thing that is described in code example is not how it's supposed to work. You can pass the custom encoders using custom_encoder parameter, can't you?

IIRC that works with Pydantic in general, so I was expecting it to work with FastAPI as well? The goal was to profit from the standard handling of returned values without having to use `jsonable_encoder` in every route, to make the code cleaner and avoid boilerplate.

ramnes:
rebased

ramnes:
rebased

PR #5743, opened 2022-12-06
Revamp response serialization performance

kigawas:
Reference: https://github.com/tiangolo/fastapi/discussions/5669


kigawas:
@tiangolo Do you have time reviewing this?

tiangolo:
Thanks @kigawas! As this could hit corner cases not covered, I would rather not do it and instead do the current official way to convert dataclasses to dicts.

If you want to squeeze performance, you could serialize directly and return a response directly. Also, a future version of FastAPI will use Pydantic's own serialization, which is equivalent to orjson (Rust's SerDe), so that will probably help a lot. You would probably also get performance improvements by using Pydantic v2 with FastAPI in most places instead of dataclasses.

PR #5804, opened 2022-12-20
📝 Update docs on Pydantic using ujson internally

mvasilkov:
The docs currently suggest that Pydantic uses ujson for faster JSON parsing. This seems to not be the case since ca. 2020, so I think at this point it's safe to remove this line from the docs.


tiangolo:
Great, thanks @mvasilkov! ☕

PR #5811, opened 2022-12-27
🔥 Remove link to Pydantic's benchmark, as it was removed there

Kludex:
The link was removed.

Maybe there's a replacement with Pydantic V2? :shrug:


Ryandaydev (approved):
The broken link removal is good. @Kludex it appears that the section describing the benefits of Pydantic (lines 185-198) were from an old version of the pydantic home page. I think you could create another PR to update the rest of the bullets.

tiangolo:
Yep, makes sense, thanks @Kludex! :rocket: :fire:

tiangolo:
📝 Docs preview for commit cf0453211db5dd74b311497f362e7c5291e1d837 at: https://6484acdd2c003b7bcebeaad8--fastapi.netlify.app

PR #5812, opened 2022-12-27
Create main.html

roham96:
Hello Team 
All Persian translations with me :D

### DIFF
--- a/docs/fa/overrides/main.html
+++ b/docs/fa/overrides/main.html
@@ -1,0 +1,83 @@
+{% extends "base.html" %}
+
+{% block announce %}
+<div class="announce-wrapper">
+  <div id="announce-left">
+    <div class="item">
+      <a class="announce-link" href="https://twitter.com/fastapi" target="_blank">
+        <span class="twemoji twitter">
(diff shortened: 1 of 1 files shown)


yezz123 (commented):
Hello @roham96 I see you almost open 6 or 7 PRs for translation, you can just create one with all the files you translate, that's can be much easier to review and comment in 🤔

tiangolo:
Thanks @roham96 for the contribution! :rocket: 

The custom `main.html` overrides are currently not used in translations, as all the translations setup is quite custom for FastAPI. And as those change from time to time, I prefer to keep it just in the main version. Given that, I'll pass on this one, but thanks for the effort!

---

About @yezz123 's question, it's actually better to have multiple isolated PRs for each file. That way people can review a single file to request changes or approve it, instead of having to read and review all the files in a PR, which would make people not want to review the whole PR.

Also, if one file is ready and approved by 2 people, that one could be merged. But if that file is in the same PR as other files that are not ready yet, then that one would be blocked by the rest of the PR... so, in the end, it's better to have multiple PRs, each with a single file. :nerd_face:

PR #5816, opened 2022-12-27
Create sponsors_badge.yml

roham96:
add translation !


roham96:
File exists Error!

tiangolo:
Thanks! These files are actually not translated and are not used by the custom documentation system, only the Markdown files. Given that I'll pass on this, but thanks for the effort! :coffee:

PR #5852, opened 2023-01-07
Removed redundant update method for adding key-value pair in dict.

subedikushal:
It is simpler to do it directly rather than via update method. This also eliminates the overhead of constructing another dictionary and calling a method, so should slightly improve performance. [ Read at Sourcery]

### DIFF
--- a/docs_src/body_multiple_params/tutorial001.py
+++ b/docs_src/body_multiple_params/tutorial001.py
@@ -22,7 +22,7 @@ async def update_item(
 ):
     results = {"item_id": item_id}
     if q:
-        results["q"]= q
+        results["q"] = q
     if item:
-        results["item"]=item
+        results["item"] = item
     return results
--- a/docs_src/body_multiple_params/tutorial001.py
+++ b/docs_src/body_multiple_params/tutorial001.py
@@ -22,7 +22,7 @@ async def update_item(
 ):
     results = {"item_id": item_id}
     if q:
-        results.update({"q": q})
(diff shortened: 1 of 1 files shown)


docs_src/body_multiple_params/tutorial001.py
```diff
@@ -22,7 +22,7 @@ async def update_item(
 ):
     results = {"item_id": item_id}
     if q:
-        results.update({"q": q})
+        results.["q"]= q
```
PabloEmidio, line 25: Remove `.` before `[`

iudeen, line 25:
```suggestion
        results["q"]= q
```

iudeen (changes requested):
Please ensure you verify basic syntax before marking PRs as ready for review :)

r0b2g1t (approved):

PabloEmidio (approved):

Ryandaydev (approved):
All tests pass.

YuriiMotov:
@subedikushal, thanks for your interest in FastAPI!

This PR changes only one occurrence, but there are many other similar (search for `.update(` in `docs_src` gives me 139 results in 112 files).

If we want to correct this, we need to do it consistently across the whole code base.
But, I'm not sure we need it. This is just code examples in docs.

tiangolo:
Thanks for the interest, I think it's fine as it is, so I'll pass on this one. :coffee:

Issue #2296, opened 2020-11-04
Typehints  instead of response_model


uvicorn:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example
Now:
```Python
from fastapi import FastAPI
from typing import Dict, List
app = FastAPI()


@app.get("/", response_model=Dict[str, List[str]])
def read_root():
    return {"Hello": ["tom", "hesus"]}
```
### The solution you would like
I want to do it through typehints:
```Python
from fastapi import FastAPI
from typing import Dict, List
app = FastAPI()


@app.get("/")
def read_root() -> Dict[str, List[str]]:
    return {"Hello": ["tom", "hesus"]}
```

Kludex:
**I don't know** if it was only a decision or if there's an issue on the suggested approach.

My input here will be related to MyPy:

```python
from typing import Dict, List

from fastapi import FastAPI
from pydantic import BaseModel

app = FastAPI()

class Potato(BaseModel):
    name: str
    color: str

@app.get("/")
def get_potato() -> Potato:
    return {"name": "best potato", "color": "purple"}
```

Doing as above, MyPy will get sad:
```bash
test.py:14: error: Incompatible return value type (got "Dict[str, str]", expected "Potato")
Found 1 error in 1 file (checked 1 source file)
```

Maybe a "FastAPI MyPy plugin" can be created to allow it (in case the suggestion is accepted).

peach-lasagna:
No, i want use annotations instead of response_model. i want fastapi understand typehints

[...]

PR #5855, opened 2023-01-08
🐛 Ignore Response classes on return annotation

Kludex:
- Related to #2296

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -3,7 +3,6 @@ import dataclasses
 import email.message
 import inspect
 import json
-import sys
 from contextlib import AsyncExitStack
 from enum import Enum, IntEnum
 from typing import (
@@ -41,9 +40,9 @@ from fastapi.utils import (
     is_body_allowed_for_status_code,
 )
 from pydantic import BaseModel
+from pydantic.utils import lenient_issubclass
 from pydantic.error_wrappers import ErrorWrapper, ValidationError
 from pydantic.fields import ModelField, Undefined
-from pydantic.utils import lenient_issubclass as issubclass
 from starlette import routing
 from starlette.concurrency import run_in_threadpool
 from starlette.exceptions import HTTPException
@@ -60,11 +59,6 @@ from starlette.routing import (
 from starlette.status import WS_1008_POLICY_VIOLATION
 from starlette.types import ASGIApp, Scope
 from starlette.websockets import WebSocket
-
-if sys.version_info >= (3, 8):
-    from typing import get_args, get_origin
-else:
-    from typing_extensions import get_args, get_origin
 
 
 def _prepare_response_content(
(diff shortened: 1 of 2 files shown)


fastapi/routing.py
```diff
@@ -356,7 +356,13 @@ def __init__(
         self.path = path
         self.endpoint = endpoint
         if isinstance(response_model, DefaultPlaceholder):
-            response_model = get_typed_return_annotation(endpoint)
+            return_annotation = get_typed_return_annotation(endpoint)
+            if inspect.isclass(return_annotation) and issubclass(
+                return_annotation, Response
+            ):
```
uriyyo, line 362: It can be replaced with `lenient_issubclass`:
```suggestion
            if lenient_issubclass(return_annotation, Response):
```

adaamz (approved):

fastapi/routing.py
Kludex: Thanks :+1:

fastapi/routing.py
```diff
@@ -356,7 +359,20 @@ def __init__(
...
+                if len(args) > 1:
+                    response_model = Union[args]
+                elif len(args) == 1:
+                    response_model = args[0]
+                else:
+                    response_model = None
+            elif issubclass(return_annotation, Response):
+                response_model = None
+            else:
+                response_model = return_annotation
```
Kludex, line 375: Those are the cases we talked about:
1. When there's a single `Response` class `response_model` becomes `None`.
2. If it's an `Union`, we filter the `Response` classes. (I'm not sure if this is what we want ???)

Kludex (commented):
I think the best approach would be to not accept this, revert to the previous release, and with more time, release the feature again with more time to think about the corner cases. :eyes:

fastapi/routing.py
tiangolo: Yeah, this gets quite complicated. Also because a Union could contain internal Unions, etc.

I think just the `lenient_issubclass(return_annotation, Response)`, to cover the simplest use cases.

And then I'll add a bit to the docs explaining that if the user wants to return and annotate something different than a Pydantic field, it needs a `response_model=None`.

fastapi/routing.py
```diff
@@ -42,6 +43,7 @@
 from pydantic import BaseModel
 from pydantic.error_wrappers import ErrorWrapper, ValidationError
 from pydantic.fields import ModelField, Undefined
+from pydantic.utils import lenient_issubclass as issubclass
```
tiangolo, line 46:
```suggestion
from pydantic.utils import lenient_issubclass
```

Could you please not replace the global `issubclass`?

tiangolo (changes requested):

tiangolo (approved):
Great, thanks! :rocket: :cake: :coffee:

PR #5856, opened 2023-01-08
Add overloads to param functions

uriyyo:
Add overloads to param functions to catch typing errors on a dependency/route level declaration.

```py
def sync_function() -> int:
    return 1


async def async_function() -> int:
    return 1


def sync_generator_function() -> Iterator[int]:
    yield 1


async def async_generator_function() -> AsyncIterator[int]:
    yield 1

reveal_type(Depends(sync_function))
reveal_type(Depends(async_function))
reveal_type(Depends(sync_generator_function))
reveal_type(Depends(async_generator_function))
reveal_type(Depends())
```
```py
example.py:24: note: Revealed type is "builtins.int"
example.py:25: note: Revealed type is "builtins.int"
example.py:26: note: Revealed type is "builtins.int"
example.py:27: note: Revealed type is "builtins.int"
example.py:28: note: Revealed type is "Any"
```

Before this PR we weren't able to catch errors like this:
```py
from typing import Any

app = FastAPI()


async def get_params() -> dict[str, int | str]:
    return {}


@app.get("/")
def my_route(
    params: dict[str, str] = Depends(get_params),   # incorrect type here
) -> Any:
    pass
```


YuriiMotov (commented):
@uriyyo, thanks for contributing to FastAPI!

Although this might be useful in some cases, currently it's recommended to use the approach with `Annotated` and this PR will not help in this case.

Similar PR was closed with the [comment](https://github.com/fastapi/fastapi/pull/10053#issuecomment-2294632661).

Related discussions:
* https://github.com/fastapi/fastapi/discussions/12113
* https://github.com/fastapi/fastapi/discussions/9338
* https://github.com/fastapi/fastapi/discussions/6071
* https://github.com/fastapi/fastapi/discussions/8453

tiangolo:
Thanks for the proposal and work on this! For now I prefer to pass, as it might interact with several refactors planned. :coffee: 

I'll close this one for now to keep things clean.

Thank you! :cake:

PR #5866, opened 2023-01-09
✏ Update sponsors and python version

f1v3-dev:
Update sponsors and python version.


yezz123 (changes requested):
Isn't that hard to maintain, I mean having all of the sponsor's URLs in different indexes can take the fix or the update for @tiangolo much time and effort 🤔 

I see no reason for these changes, otherwise, the version update is remarkable in the files.

PR #5869, opened 2023-01-10
A small clarification for the launch of the site in `docs/en/docs/contributing.md`.

Xewus:
Sometimes I have seen that people, having made a translation, do not look at what the result looks like. Maybe it will help them a little.


Xewus:
It turns out that this line is somewhere at the bottom.

Issue #5878, opened 2023-01-12
Bug - Missing String representation of the fastapi.HTTPException when initializing with parameter names


CribberSix:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
import fastapi 
import pytest

# When creating the function with parameter names, the string representation does not work:
exc = fastapi.HTTPException(status_code=401, detail="Authentication error: Authorization header is missing.")
print(f"String representation of the exception: '{exc}'")

# When creating the function without parameter names, the string representation works:
exc = fastapi.HTTPException(401, "Authentication error: Authorization header is missing.")
print(f"String representation of the exception without params: '{exc}'")


# The missing string representation becomes a problem when using pytest to check for the raised
# exception as well as the exception message, as the match="..." parameter value is compared to the 
# string representation of the exception, which does not exist when the exception is initialized
# with parameter names. 
def test_for_invalid_something(): 
    with pytest.raises(fastapi.HTTPException, match=r"Authentication error: Authorization header is missing\."):        
        raise fastapi.HTTPException(status_code=401, detail="Authentication error: Authorization header is missing.")
```

### Description

As the code shows, when we initialize the exception with parameter names, the string representation is empty. I'm not sure what the cause could be... 

### Wanted Solution

When creating the function with parameter names, the string representation should also work-

### Wanted Code

```python
exc = fastapi.HTTPException(status_code=401, detail="Authentication error: Authorization header is missing.")
print(f"String representation of the exception: '{exc}'")
```

### Operating System

Windows

### FastAPI Version

0.89.1

### Python Version

3.8.10

PR #5879, opened 2023-01-12
add __str__ method to HTTPException

ebottos94:
As mentioned in Iusse #5878 there's a bug in HTTPException's string representation, it seems that \__repr__ method in starlette.exceptions.HTTPException doesn't works correctly without explicit call \__repr__ method. To avoid this I've implemented a \__str__ method in fastapi.exceptions.HTTPException

### Current situation :

image

### After my commit : 

image

### DIFF
--- a/fastapi/exceptions.py
+++ b/fastapi/exceptions.py
@@ -16,7 +16,7 @@ class HTTPException(StarletteHTTPException):
         super().__init__(status_code=status_code, detail=detail, headers=headers)
 
     def __str__(self) -> str:
-        return f"status_code : {self.status_code}, detail : {self.detail}"
+        return f"HTTPException(status_code: {self.status_code}, detail: {self.detail})"
 
 
 RequestErrorModel: Type[BaseModel] = create_model("Request")
--- a/fastapi/exceptions.py
+++ b/fastapi/exceptions.py
@@ -15,6 +15,9 @@ class HTTPException(StarletteHTTPException):
     ) -> None:
         super().__init__(status_code=status_code, detail=detail, headers=headers)
 
+    def __str__(self) -> str:
+        return f"status_code : {self.status_code}, detail : {self.detail}"
+
 
 RequestErrorModel: Type[BaseModel] = create_model("Request")
 WebSocketErrorModel: Type[BaseModel] = create_model("WebSocket")
--- a/tests/test_string_representation_exception.py
+++ b/tests/test_string_representation_exception.py
@@ -1,0 +1,5 @@
+from fastapi.exceptions import HTTPException
+
+def test_string_representation():
+    exception = HTTPException(status_code=400, detail='detail exception')
+    assert exception.__str__() == 'HTTPException(status_code: 400, detail: detail exception)'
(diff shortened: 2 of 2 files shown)


odiseo0 (changes requested):
You should add tests to pass the coverage requirement.

ebottos94:
> You should add tests to pass the coverage requirement.

Thank you, what kind of tests i have to add to pass coverage requirement?

ebottos94:
> You should add tests to pass the coverage requirement.

Thank you @odiseo0, I've add test and change string representation as you suggested.

yezz123 (approved):
LGTM 🚀

odiseo0 (approved):

iudeen:
Does it make sense to add this Starlette itself?🤔 Just wondering

r0b2g1t (approved):
That looks good for me.

LorhanSohaky (approved):

fastapi/exceptions.py
```diff
@@ -15,6 +15,9 @@ def __init__(
     ) -> None:
         super().__init__(status_code=status_code, detail=detail, headers=headers)
 
+    def __str__(self) -> str:
+        return f"HTTPException(status_code: {self.status_code}, detail: {self.detail})"
```
Harry-Lees, line 19: Should this use `self.__class__.__name__` like in starlette (https://github.com/encode/starlette/blob/ada845cc8ad0b531b16749b9bf3a57114c60c679/starlette/exceptions.py#L22)

I haven't actually pulled this PR, but I think if I subclassed this class I'd end up with mismatching `__repr__` and `__str__`

```python
class CustomHTTPException(HTTPException):
    pass

exc = CustomHTTPException(status_code=400, detail="something went wrong")

print(exc)
#> HTTPException(status_code=400, detail='something went wrong')

print(repr(exc))
#> CustomHTTPException(status_code=400, detail='Something went wrong')
```

Kyle-sandeman-mrdfood:
Honestly you could just do `__str__ = __repr__` (in definition, not patched like below)
```python
>>> from fastapi.exceptions import HTTPException
>>> h = HTTPException(status_code=200)
>>> str(h), repr(h)
('', "HTTPException(status_code=200, detail='OK')")
>>> HTTPException.__str__ = HTTPException.__repr__
>>> str(h), repr(h)
("HTTPException(status_code=200, detail='OK')", "HTTPException(status_code=200, detail='OK')")
>>> class Custom(HTTPException):
...   pass
... 
>>> c = Custom(status_code=201)
>>> str(c), repr(c)
("Custom(status_code=201, detail='Created')", "Custom(status_code=201, detail='Created')")
```

Kludex:
We'll implement this in Starlette.

Kludex:
https://github.com/encode/starlette/pull/2181 was merged.

tiangolo:
Thanks @ebottos94! 🍰 

And thanks everyone for the discussion. As @Kludex points out, as this is now in Starlette, I'll close this one here. But thanks for the effort! ☕

PR #5897, opened 2023-01-18
feat: add custom route on route level

### LINKED ISSUE #1174: How to use different middleware for different routes/path
### First check

### Description

Is it possible to use different middleware for different routes/path?

### Additional context

In my case my need comes from CORS. But, I am sure there is other cases than CORS requirements that someone would need different middlewares for different paths.

* I want `myapi.com/path1` to allow origins of calls from `myfrontend.com`
* I want `myapi.com/path2` to allow origins of calls from anywhere ('*') since it is a public facing api.

I checked if it was possible to add a middleware at a router level and did not find any documentation about it.

### Code example

```python
from fastapi import FastAPI
from fastapi.middleware.cors import CORSMiddleware

app = FastAPI()

app.add_middleware(
    CORSMiddleware,
    allow_origins=['myfrontend.com'],
    allow_credentials=True,
    allow_methods=["*"],
    allow_headers=["*"],
)

# I want /path1 to allow only calls from 'myfrontend.com'
@app.get("/path1")
async def path1():
    return {"status": "alive"}

# I want /path2 to be a public facing api that can be accessed from anywhere ['*']
@app.get("/path2")
async def path2():
    return {"status": "alive"}

```

arkadybag:
fixed: https://github.com/tiangolo/fastapi/issues/1174 https://github.com/tiangolo/fastapi/issues/4437
The problem:
Sometimes it is nice to have the ability to pass a custom route to a specific endpoint. For example, I need to check headers before processing with the body, [but FASTAPI forces me to read the body before.](https://github.com/tiangolo/fastapi/blob/fb7ea03bc728ea6528194b2a5a245d4ea9f3b0ff/fastapi/routing.py#L190)

Currently, I can't do it because `route_class_override` uses only internally, and need to reimplement a lot of things to solve it. So, my proposal is to let routers override a route class.

_I've tried to solve this issue through middlewares but it's not clear to check into middlewares if I should "filter" requests or not based on the path. Another solution is Mount, but it also not solving the problem because it requires creating sub-path for it (the same with `include_router`)._

PS:
I think my PR partially add support for that one(logically) https://github.com/encode/starlette/pull/1649 https://github.com/encode/starlette/pull/1464


yezz123:
This really looks cool 🚀 

But if you see it from a different perspective, is that it may lead to routing conflicts if the custom routes overlap with previously defined routes.

This can lead to unexpected behavior and make the API more difficult to understand and maintain. Additionally, if the custom routes are not well-designed or validated, they can introduce security vulnerabilities.

arkadybag:
@yezz123 That is the same as define custom route https://fastapi.tiangolo.com/advanced/custom-request-and-route/#create-a-custom-gziprequest-class.
The difference is only that you can use it on endpoint-based level. Your thoughts about 
> not well-designed or validated, they can introduce security vulnerabilities.

the same as for the custom route class and that is from a different scope. As a user I want to have an ability to overwrite it for endpoint, not for router.

arkadybag:
@tiangolo can you please take a look? Thanks!

nidhishgajjar:

## Summary

This PR exposes the existing `route_class_override` parameter from `APIRouter.add_api_route()` to the convenience decorators (`.get()`, `.post()`, `.put()`, `.delete()`, etc.) on both `FastAPI` and `APIRouter`. This allows users to specify a custom `APIRoute` subclass for individual endpoints without affecting the entire router.

## Architecture

The plumbing already existed in `add_api_route()` — this PR threads the parameter through the decorator methods that delegate to it. Clean and minimal.

## Issues

### 1. Incorrect imports in test file (Warning)

**File:** `tests/test_custom_route_class_for_endpoint.py`

The test imports `Request` from `urllib.request` and `Response` from `fastapi.openapi.models`. These should be:
- `Request` from `starlette.requests` (not `urllib.request`)
- `Response` from `starlette.responses` (not `fastapi.openapi.models`)

```python
# Current (incorrect):
from urllib.request import Request
from fastapi.openapi.models import Response

# Should be:
from starlette.responses import Response
```

While this doesn't affect runtime behavior (duck typing works), it gives wrong type information to type checkers and IDEs, and could confuse contributors reading the test.

### 2. Missing `Doc()` annotation (Suggestion)

**File:** `fastapi/routing.py`, `fastapi/applications.py`

The new `route_class_override` parameter lacks `Doc()` annotations in the decorator methods. Other parameters like `response_class`, `name`, and `callbacks` have `Doc()` annotations. Adding one would keep the API reference complete.

### 3. No documentation for users (Suggestion)

There's no documentation or example in `docs/` showing how to use this feature. A mention in the "Custom Route Class" documentation would help users discover it.

## Cross-file Impact

- **`fastapi/applications.py`**: Correctly adds the import of `APIRoute` and threads `route_class_override` through all route decorators.
- **`fastapi/routing.py`**: All decorator methods correctly pass through to `add_api_route`.
- Backward compatible — default is `None`, which preserves existing behavior.

## Assessment

**approve** — This is a clean, well-scoped PR that exposes existing functionality to the decorator API. The test imports should be fixed, but the feature itself is correct and useful. Good inclusion of parametrized tests.

tiangolo:

Thank you! 🍰

PR #5900, opened 2023-01-19
Add packaging status of fastapi for various linux distros to readme

rrrrrrmb:
This would provide an overview of the different versions of fastapi available in various linux distributions.
This would be helpful for people who would prefer to install software using package managers like apt/dnf/pacman.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,10 @@ FastAPI stands on the shoulders of giants:
 
 ## Installation
 
+<a href="https://repology.org/project/python:fastapi/versions">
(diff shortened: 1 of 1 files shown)


r0b2g1t (approved):

tiangolo:
Thanks for the interest! But I actually prefer to avoid suggesting to use distro-repackaged packages. People should pretty much never do that, it's not official, not as frequently updated, etc. So I'll pass on this one, but thanks for the effort! :coffee:

PR #5906, opened 2023-01-21
Add optional Swagger-related configurations to FastAPI constructs to …

pythonzaza:
…expose the js and css that swagger depends on to resolve the problem of not having access to the API documentation in China

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -81,6 +81,9 @@ class FastAPI(Starlette):
         deprecated: Optional[bool] = None,
         include_in_schema: bool = True,
         swagger_ui_parameters: Optional[Dict[str, Any]] = None,
+        swagger_js_url: str = "https://cdn.jsdelivr.net/npm/swagger-ui-dist@4/swagger-ui-bundle.js",
+        swagger_css_url: str = "https://cdn.jsdelivr.net/npm/swagger-ui-dist@4/swagger-ui.css",
+        swagger_favicon_url: str = "https://fastapi.tiangolo.com/img/favicon.png",
         generate_unique_id_function: Callable[[routing.APIRoute], str] = Default(
             generate_unique_id
         ),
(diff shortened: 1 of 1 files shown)


r0b2g1t (approved):

gtadamson:
Can someone fix this so it merges?

tiangolo:
Hello there, the best way to achieve this would be following the same ideas in the docs for self-hosting the JS and CSS assets: https://fastapi.tiangolo.com/advanced/extending-openapi/#self-hosting-javascript-and-css-for-docs

That way it's more flexible, it allows you to customize everything even further.

Given that, I'll pass on this one, but thanks for the effort! 🍰

gtadamson:
@tiangolo,

We currently have that recommendation (https://fastapi.tiangolo.com/advanced/extending-openapi/#self-hosting-javascript-and-css-for-docs) implemented, but with a recent release of FastAPI, the tool is re-writing the version of the openapi spec that we provide to a higher version upon deployment. The static files that are saved are incompatible with the new version of the spec and all of our swagger pages are broken when we upgrade.

One solution that you might propose is to upgrade all the static files for every server every time FastAPI changes the openapi spec version. This is what we're trying to avoid. The solution above allows us to have a single point to change the files. I think it also keeps our repositories cleaner (no copy of static files that are unrelated to the domain problem we're trying to solve) and simpler for more junior developers by abstracting these details into a single location.

I respectfully ask if you'll reconsider your decision.

Thank you for your time and effort, we have migrated to FastAPI for all of our new servers.

pythonzaza:
A relatively reasonable solution is provided in the documentation, and there seems to be no need to modify it.


| |
李先生
|
|
***@***.***
|
---- Replied Message ----
| From | ***@***.***> |
| Date | 8/10/2023 08:32 |
| To | ***@***.***> |
| Cc | ***@***.***> ,
***@***.***> |
| Subject | Re: [tiangolo/fastapi] Add optional Swagger-related configurations to FastAPI constructs to … (PR #5906) |

@tiangolo,

We currently have that recommendation (https://fastapi.tiangolo.com/advanced/extending-openapi/#self-hosting-javascript-and-css-for-docs) implemented, but with a recent release of FastAPI, the tool is re-writing the version of the openapi spec that we provide to a higher version upon deployment. The static files that are saved are incompatible with the new version of the spec and all of our swagger pages are broken when we upgrade.

One solution that you might propose is to upgrade all the static files for every server every time FastAPI changes the openapi spec version. This is what we're trying to avoid. The solution above allows us to have a single point to change the files. I think it also keeps our repositories cleaner (no copy of static files that are unrelated to the domain problem we're trying to solve) and simpler for more junior developers by abstracting these details into a single location.

I respectfully ask if you'll reconsider your decision.

Thank you for your time and effort, we have migrated to FastAPI for all of our new servers.

PR #5913, opened 2023-01-22
📝 Update code examples in docs for body, replace name `create_item` with `update_item` when appropriate

OttoAndrey:
Hi! I did some improvements for code exmples;
1. Renamed function because it is `put` endpoint and logicaly it should be `update` not `create`;
2. ~Removed redundant `if` statement because `q` param in those examples are required. Or I can add `None` to annotations;~


r0b2g1t (approved):
Looks good for me.

Ryandaydev (changes requested):
Tests pass.

The changes listed as "1" are good changes.

I understand the suggestion for "2", however I suggest not making those changes because that code is repeated in multiple places throughout the tutorials. As I understand the tutorials, they build on each other somewhat, and they focus on specific items in each section. It seems to me that making these changes in a few places adds inconsistencies that the learner has to investigate as they change. It may be more useful to keep it consistent even if the example isn't optimal?

OttoAndrey:
> I understand the suggestion for "2", however I suggest not making those changes because that code is repeated in multiple places throughout the tutorials. As I understand the tutorials, they build on each other somewhat, and they focus on specific items in each section. It seems to me that making these changes in a few places adds inconsistencies that the learner has to investigate as they change. It may be more useful to keep it consistent even if the example isn't optimal?

@Ryandaydev First time when learner see this `q` parameter in this article https://fastapi.tiangolo.com/tutorial/query-params/#optional-parameters . And here it looks logical because `q` is optional and code example has `if` condition for that parameter. For consisten I suggest add `None` for `q` param inside annotations. After this `if` statement make sense. And it looks like previous examples with `q` param:

```python
@app.get("/items/{item_id}")
async def read_items(q: str | None, item_id: int = Path(title="The ID of the item to get")):
    if q:
        results.update({"q": q})
        return results
```

Ryandaydev (approved):
I understand the reasoning on this change. Also, the tests still pass.

docs_src/path_params_numeric_validations/tutorial002.py
```diff
@@ -5,7 +5,5 @@
 
 @app.get("/items/{item_id}")
 async def read_items(q: str, item_id: int = Path(title="The ID of the item to get")):
-    results = {"item_id": item_id}
-    if q:
-        results.update({"q": q})
+    results = {"item_id": item_id, "q": q}
```
mjpieters, line 8: What if `q` is an empty string?

```
$ curl http://localhost:8000/items/42?q=
```

Nothing in the annotation sets a minimum length.

OttoAndrey: So it returns something like that
```python
{
    "item_id": 123,
    "q": "",
}
```
Why not? Just tutorial example

OttoAndrey:
Ok. I removed controversial changes and left only renamed functions

Kludex (approved):

tiangolo:
Nice, makes sense, thank you @OttoAndrey! 🍰 

And thanks everyone for the input! ☕

PR #5917, opened 2023-01-23
✨ Allow using `openapi.models.Server` model to specify `FastAPI(servers=[])`

aledenza:
# Motivation
As mentioned in original merge request(https://github.com/tiangolo/fastapi/pull/1547), that added feature to define multiple servers, it is possible to use `from fastapi.openapi.models import Server` as a server definition. But currently this usage cause an exception `'Server' object has no attribute 'get'`. So this mr fixes this issue

# Code to reproduce
```
from fastapi import FastAPI

server1 = Server(url="http://example.com", description="optional description")
server2 = Server(url="http://test.com")

app = FastAPI(servers=[server1, server2])
```


A2Data:
AttributeError: 'Server' object has no attribute 'get'

YuriiMotov (approved):
LGTM

This improves developer experience.
There is an [alternative approach with `TypedDict`](https://github.com/fastapi/fastapi/pull/3654), but it requires code duplication.

It would probably be nice to make this consistently for all parameters where we have Pydantic model, but let's first wait for Sebastián to approve the idea

tiangolo:

Thank you! 🍰

PR #5937, opened 2023-01-29
📝 Update template docs with more info about `url_for`

EzzEddin:
This adds more clarity on how to pass arguments to a URL route. It would save developers time figuring that out on their own, especially since the syntax of the `url_for` is a bit different from the Django's `{% url %}`


mjpieters:
The extra example looks out of place in the normal pattern of the FastAPI documentation. I'd update the `docs_src/templates/templates/item.html` example to always include a `<a href="{{ url_for('read_item', id=id) }}">` link in some way.

tiangolo:
📝 Docs preview for commit 83babfb4ac432880241cdf9688ba0afe3c0b2428 at: https://2f348dd8.fastapitiangolo.pages.dev

tiangolo:
Thanks @EzzEddin! 🍰 

And thanks for the idea @mjpieters! ☕ 

I updated the docs, the example code, and the tests a bit. 🤓

PR #5947, opened 2023-01-31
fix: display address in docs CLI live mode 📝 

ComicShrimp:
When using the `live` option in the CLI, provide feedback about the selected language and the mkdocs address


LorhanSohaky (approved):

tiangolo:
Thanks @ComicShrimp! I fixed the call to MkDocs in another PR at some point, so now it shows the default logs, including this. So this change is no longer necessary. Given that, I'll close this, but thanks for the help! 🍰

PR #5972, opened 2023-02-07
Use argon2 recommendation rather than bcrypt

pythonweb2:


waderoberts123:
The docs currently claim that "bcrypt" is the "recommended algorithm," however, in the current passlib docs, they push more towards using argon2 rather than legacy hashes.

From the passlib docs:

> While Argon2 is much younger than the others, it has seen heavy scrutiny, and was purpose-designed for password hashing. In the near future, it stands likely to become the recommended standard.

For this reason, it is proposed to change the docs to reflect this.

iudeen:
I read multiple places that Bcrypt is still the best available password hash for web apps and APIs.

tiangolo:
Thanks @pythonweb2! But I don't really think they push Argon2 as the recommendation. And reading the docs again with fresh eyes I am inclined to choose Bcypt: https://passlib.readthedocs.io/en/stable/narr/quickstart.html#making-a-decision

Agreed with @iudeen 🤓 

Given that, I'll close this one. But thanks for the interest! ☕

PR #5974, opened 2023-02-07
➕ [FEATURE] Mounting app creates API definition

aledenza:
# Feature
With this MR, if `subapp` is mounted to `mainapp`, you will see dropdown with ability to select API definition on the `mainapp` docs page.

# Description 
1. Mount the app as shown below
```
app = FastAPI(title="Main app", servers=[{"url": "http://localhost:8000"}])
@app.get("/app")
def read_main():
    return {"message": "Hello World from main app"}

subapi1 = FastAPI(title="Sub app 1")
@subapi1.get("/sub")
def read_sub():
    return {"message": "Hello World from sub API"}

app.mount("/subapi1", subapi1) # you can specify name of api definition. If not, path used as name
```
2. Go to `http://localhost:8000/docs` and you will see this
Снимок экрана 2023-02-08 в 00 08 58
3. You can select the mounted app API from dropdown
Снимок экрана 2023-02-08 в 00 09 08
4. Select mounted API and it will appear on the screen
Снимок экрана 2023-02-08 в 00 09 25
5. ReDoc now comes with same feature
Снимок экрана 2023-02-08 в 00 10 31

# Help wanted / Discussion 
Swagger UI comes with API definition feature almost out the box, unlike Redoc, that does not have this feature(or i did not found). So current solution is to use customized Swagger UI top bar with some JS on top of it.

So, if somebody know how to add this feature to redoc properly, I'll really appreciate your help.
Maybe Redoc does ned this feature if there's not native support.

### DIFF
--- a/docs_src/extending_openapi/tutorial002.py
+++ b/docs_src/extending_openapi/tutorial002.py
@@ -14,10 +14,11 @@ app.mount("/static", StaticFiles(directory="static"), name="static")
 @app.get("/docs", include_in_schema=False)
 async def custom_swagger_ui_html():
     return get_swagger_ui_html(
-        openapi_url=app.openapi_url,
+        openapi_urls=app.openapi_urls,
         title=app.title + " - Swagger UI",
         oauth2_redirect_url=app.swagger_ui_oauth2_redirect_url,
         swagger_js_url="/static/swagger-ui-bundle.js",
+        swagger_standalone_js_url="/static/swagger-ui-standalone-preset.js",
         swagger_css_url="/static/swagger-ui.css",
     )
 
@@ -30,9 +31,10 @@ async def swagger_ui_redirect():
 @app.get("/redoc", include_in_schema=False)
 async def redoc_html():
     return get_redoc_html(
-        openapi_url=app.openapi_url,
+        openapi_urls=app.openapi_urls,
         title=app.title + " - ReDoc",
         redoc_js_url="/static/redoc.standalone.js",
+        swagger_css_url="/static/swagger-ui.css",
     )
 
 
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -10,6 +10,7 @@ from typing import (
     Sequence,
     Type,
     Union,
+    cast,
 )
 
 from fastapi import routing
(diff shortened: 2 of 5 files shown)


r0b2g1t:
looks interesting.

BilalAlpaslan (approved):

Ryandaydev:
@ZakharovDenis I'd like to understand this PR a bit more. How much customization is required to add this functionality to the swagger Docs? And how does it affect the compatibility with the out-of-the-box swagger UI we include with fastapi? Since a major open source library like SwaggerUI will change at its own pace, we need to keep our use of it pretty vanilla. Otherwise we spend our time maintaining compatibility with external libraries rather than innovating fastapi.

ZakharovDenis:
> @ZakharovDenis I'd like to understand this PR a bit more. This looks like a nice feature, but I'd like to understand any overhead that comes with it. Does it require any customization of the SwaggerUI libraries? And how does it affect the compatibility with the out-of-the-box swagger UI we include with fastapi? Since a major open source library like SwaggerUI will change at its own pace, we need to keep our use of it pretty vanilla, IMHO. Otherwise we spend our time maintaining compatibility with external libraries rather than innovating fastapi.

@Ryandaydev, thanks for the question. 
If we are talking about swaggerUI, it does not require any customization at all. To make this feature available in swaggerUI, we have to: 
1. link one more official JavaScript file, which is `swagger-ui-standalone-preset.js`
2. change layout to `SwaggerUIStandalonePreset` 
3. use `urls` keyword instead of `url` in `SwaggerUIBundle` object initialization

All this steps is something, that, I think, we can call "SwaggerUI API". 
So, the main risk, that swaggerUI maintainers will drop this particular API for some reason.

Things become a little bit different if we talk about Redoc. Because Redoc does not have same feature out-of-the-box, i made a little customization to make this available.
So things, that I've done:
1. In HTML I created the topBar which is the same that SwaggerUI creates by itself. So it has same classes, ids and tags
2. Loaded `swagger-ui-standalone-preset.js`, which is loads CSS with styles required for the topBar.
3. Wrote a tiny JS function that changes OpenAPI file if dropdown value is changed. (Redoc has similar example here https://github.com/Redocly/redoc/blob/gh-pages/examples/multiple-apis/index.html)

The main risk here that the SwaggerUI maintainers will change topBar classes, so new classes will not correspond with classes, that i wrote for custom topBar, and styles will break.

In fact, I think, that the maintainers of SwaggerUI are more likely to change topBar classes than drop the API, but even class changes won't be frequent.

Also, as I mentioned in MR head, I think, that I probably did not found information how to enable same or similar feature for vanilla Redoc.

Also, It's worth mentioning, that if you do not mount any app and use single FastAPI app, the SwaggerUI and Redoc will look like before, without any topBars. So for single application, all behaviour stays same.

Ryandaydev:
Thanks for the explanation, sounds like the new functionality on the swaggerUI doesn't introduce unnecessary customization. On the redocs, like you said it would be better if we didn't have to customize quite so much. How does Swagger describe this functionality in their docs? (Maybe we could find or suggest something similar to the maintainers.)

tiangolo:
Thanks for the interest! But for now, I prefer to keep any submounted application as an independent app, with its own Swagger UI, etc.

I'm not sure what use case would be solved by this. :thinking: Maybe I'm missing something. But in that case, it would be better to start in a GitHub Discussion, filling the form and with a minimal self contained example that covers the use case and that I can copy paste and run to understand what you want to achieve. Maybe there's already another way to do it.

Issue #5935, opened 2023-01-28
Dataclasses get converted into dicts instead of being validated by pydantic directly


ImogenBits:
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from dataclasses import dataclass, field
from typing import Any
from pydantic import BaseModel, BaseConfig
from fastapi import FastAPI


class ThingBase:
    id: int
    circular_ref: "ThingBase"


class ThingID:
    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    def validate(cls, obj: Any) -> int:
        if isinstance(obj, int):
            return obj
        elif isinstance(obj, ThingBase):
            return obj.id
        else:
            raise TypeError


class Model(BaseModel):
    id: int
    circular_ref: ThingID

    class Config(BaseConfig):
        orm_mode = True


class ThingNormal(ThingBase):
    def __init__(self) -> None:
        super().__init__()
        self.id = 0
        self.circular_ref = self


@dataclass
class ThingDataclass(ThingBase):
    id: int = field(init=False)
    circular_ref: ThingBase = field(init=False)

    def __post_init__(self):
        self.id = 0
        self.circular_ref = self


app = FastAPI()


@app.get("/normal", response_model=Model)
def normal_route():
    return ThingNormal()


@app.get("/dataclass", response_model=Model)
def dataclass_route():
    return ThingDataclass()
```

### Description

[...]

PR #5978, opened 2023-02-08
Fix orm_mode enabled parsing of response content

ImogenBits:
This PR fixes the issue I described in #5935. Currently the response parsing functions check if the response content has the `read_with_orm_mode` flag set and then don't try to convert it into another data structure in order to be parsed and validated by pydantic. This is incorrect behaviour, the flag now is called simply `orm_mode` and it is placed on the model that is being constructed, not on the data we are constructing it from. This PR just changes that behaviour to check for the correct flag on the field class.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -71,15 +71,17 @@ def _prepare_response_content(
     exclude_unset: bool,
     exclude_defaults: bool = False,
     exclude_none: bool = False,
+    field: Optional[ModelField] = None,
 ) -> Any:
-    if isinstance(res, BaseModel):
-        read_with_orm_mode = getattr(res.__config__, "read_with_orm_mode", None)
+    if field is not None and issubclass(field.type_, BaseModel):
+        read_with_orm_mode = getattr(field.type_.__config__, "orm_mode", None)
         if read_with_orm_mode:
             # Let from_orm extract the data from this model instead of converting
             # it now to a dict.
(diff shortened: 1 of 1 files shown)


fastapi/routing.py
```diff
@@ -67,15 +67,17 @@ def _prepare_response_content(
     exclude_unset: bool,
     exclude_defaults: bool = False,
     exclude_none: bool = False,
+    field: Optional[ModelField] = None,
 ) -> Any:
-    if isinstance(res, BaseModel):
-        read_with_orm_mode = getattr(res.__config__, "read_with_orm_mode", None)
+    if field is not None and issubclass(field.type_, BaseModel):
```
odiseo0, line 72: I think you need to use `issubclass(type(field.type_), BaseModel)` or just use `isinstance`. Also, it would be nice if you add some tests

odiseo0:
Would be nice if you add some tests

fastapi/routing.py
ImogenBits: no, `field.type_` is the class passed via `response_model`, to check if it supports validation with orm mode we need to check if it is a subclass of `BaseModel`.

tiangolo:
Thanks @ImogenBits, but `read_with_orm_mode` is there as intended. It's used by SQLModel to tell FastAPI to read from attributes instead of pre-serializing, it's a different config from `orm_mode`, although it's related. Given that, I'll pass on this one. But thanks for the interest! ☕

PR #6038, opened 2023-02-23
✨ Add support for custom generic classes as dependencies

### LINKED ISSUE #3920: I hope fastapi can support TypeVar

YYKlimenko:
**Added support for injecting custom generic class objects via Depends().**
[how it worked before](https://ibb.co/23X6MtL)
[how it works now](https://ibb.co/Zg65g2F)

### The Issue:

Using custom generic classes via Depends() led to an unpredictable result. The fastAPI  could not parse the initializer signature correctly due to the way generic classes work in Python.

### Why is it important?

The ability to use generic classes via Depends() is expected behavior for a typed program. It allows you to make more stable program using DI and avoid writing duplicate code.

Using generics types makes the code more flexible, safer, and more concise.

### What's changed?

dependencies/utils.py module was changed.

- Added function 'is_generic_type':
      – to check whether the object is generic type.

- Added function ‘substitute_generic_type’:
      – to substitute a real annotate in the signature. 

 - Changed function ‘get_typed_annotation’:
       - added optional attribute ‘typevars’;
       - if 'typevars' is received, the function calls ‘substitute_generic_type’;

- Changed function ‘get_typed_signature:
      - it calls ‘is_generic_type ’ function. If the function returns False, it works as before. Otherwise, the typevars and types mapping dict is created. This dict pass to 'get_typed_annotation'.

### Testing:

The changes do not affect the existing implementation of the fastAPI. It only add new functionality.
All basic tests have been completed, and new ones have been added (**tests/test_dependency_generic_class.py**) that check work with custom generic types via Depends().
It has been tested with Python 3.7-3.10.

**UPD YuriiMotov:** closes #3920


caeser1996:
@YYKlimenko can you please review why the test cases are failing

YYKlimenko:
> @YYKlimenko can you please review why the test cases are failing

I found a silly mistake: two files were mixed up. I did a new commit.

YuriiMotov (commented):
I like the idea to support generic classes as dependencies.

For now if we try using generic class with `Depends`, it will give us openapi schema with `args` and `kwargs` as required Query parameters.
image


If the idea and implementation are accepted by Sebastián, we will need to add test for openapi schema

tiangolo:
Thanks for the interest @YYKlimenko.

I struggle to see a use case that would need this. As I was saying in https://github.com/fastapi/fastapi/issues/3920

> If you (or anyone) has a specific use case that needs this, please create a new discussion question with a minimal reproducible example that showcases what you want to achieve, and why it can only be achieved with this feature and not with regular dependencies.

For now, I'll pass on this one. Thanks!

PR #3032, opened 2021-04-01
✨ Support Python internal description on Pydantic model's docstring

Kludex:
## Problem

Currently, on FastAPI endpoint functions we can document for OpenAPI description/Swagger and for internal Python purposes:
```python
from fastapi import FastAPI
from pydantic import BaseModel

app = FastAPI()

@app.get("/")
def useful():
    """My description.
    \f
    Internal documentation.
    """
    ...
```
On Swagger we'll have:
image
 
When documenting models, it doesn't have the same behavior:
```python
from enum import Enum
from fastapi import FastAPI

class Banana(str, Enum):
    """My description.
    \f
    Internal documentation.
    """
    HAHA = "haha"
    HOHO = "hoho"

app = FastAPI()

@app.get("/")
def useful(banana: Banana):
    ...
```
On Swagger, we'll have:
image
    
## Solution

The proposed _two lines of code_ on this PR, makes it possible to have the same behavior as the endpoint description:
image

It follows the same idea as: https://github.com/tiangolo/fastapi/blob/10397ddc30ef9bb26e9247f04fe20ce3045081c4/fastapi/routing.py#L350

## Related issues
- https://github.com/tiangolo/fastapi/issues/3024

## Questions & Alternatives
- Should Pydantic handle this on the [model_process_schema()](https://github.com/samuelcolvin/pydantic/blob/e68c277afe6bf6e7966bfecfe9b228d33f2de44b/pydantic/schema.py#L519) function instead?

PR #6039, opened 2023-02-23
🐛 Fix inconsistent processing of model docstring formfeed char with Pydantic V1

MaxwellPayne:
## Problem

Generating openapi specs for models with a formfeed (`\f`) in the docstring yields inconsistent results. For example, in the case of:
```python
class Address(BaseModel):
    """
    This is a public description of an Address
    \f
    You can't see this part of the docstring, it's private!
    """
    ...
```
Sometimes the openapi `description` of `Address` correctly renders as `This is a public description of an Address\n`. But other times, you get the entire docstring even though the `\f` should be keeping the remainder of the docstring private. 

The bug seemed to be in the following function:
```python
def get_model_definitions(
    *,
    flat_models: Set[Union[Type[BaseModel], Type[Enum]]],
    model_name_map: Dict[Union[Type[BaseModel], Type[Enum]], str],
) -> Dict[str, Any]:
    definitions: Dict[str, Dict[str, Any]] = {}
    for model in flat_models:
        m_schema, m_definitions, m_nested_models = model_process_schema(
            model, model_name_map=model_name_map, ref_prefix=REF_PREFIX
        )
        definitions.update(m_definitions)
        model_name = model_name_map[model]
        if "description" in m_schema:
            m_schema["description"] = m_schema["description"].split("\f")[0]
        definitions[model_name] = m_schema
    return definitions
```
The issue is that the `definitions` dict has its entries updated in two different places:
1. `definitions.update(m_definitions)`
2. `definitions[model_name] = m_schema`

When a model definition is written in line (2), the proper formfeed escape of the docstring is persisted. But if the same model is subsequently written by line (1), the previous docstring fix is clobbered over and we see the entire docstring. 

This is a non-deterministic error, since `flat_models` is an unordered set. Depending on order of models in iteration, you could end up with either (1) or (2) as the final edit to a given model. 

## Solution

This fix works around the bug by handling the formfeed escape at the _end_ of model processing. By that point our `definitions` dict will be final, so any edits to the models are guaranteed to make it into the return value. 

## Related issues

- https://github.com/tiangolo/fastapi/pull/3032

## Questions & Alternatives
- Should Pydantic handle this on the [model_process_schema()](https://github.com/samuelcolvin/pydantic/blob/e68c277afe6bf6e7966bfecfe9b228d33f2de44b/pydantic/schema.py#L519) function instead? (also asked in #3032)
- Might there be other pre-existing, undiscovered bugs resulting from the fact that `definitions` is updated in two different ways?


YuriiMotov (approved):
**Preconditions:**
* Pydantic V1
* Nested model has description with formfeed character

**Symptoms:**
Generated openapi schema will be inconsistent (sometimes the part after `\f` will be truncated, sometimes not).
`test_openapi_schema` (from `test_get_model_definitions_formfeed_escape.py`) will fail sometimes.
`test_model_description_escaped_with_formfeed` will always fail for `sort_reversed=False`.

Tests in `test_get_model_definitions_formfeed_escape.py` prove that the result of `_compat.get_model_definitions` depends on the order of models passed in `flat_models` parameter.

**Explanation:**
* `Address` is nested into `Facility`.
* When `Address` model goes before `Facility` in `flat_models` (input parameter of `_compat.get_model_definitions`), it's added to `definitions` and its `description` is updated (truncated by formfeed charecter).
* But on the second iteration of for-loop, we have `Address` in `m_definitions` (since `Address` is nested into `Facility`), and `definitions.update(m_definitions)` overrides existing schema of `Address` in `definitions`.
* So, `definitions` now contain the `Address` schema with original `description` (not truncated).

**The idea of the fix** is to add all models to `definitions` first and only then update `description`'s of all models.

Looks good to me.

tests/test_get_model_definitions_formfeed_escape.py
```diff
@@ -0,0 +1,180 @@
...
+        super().__init__(seq)
+        self.sort_reversed = sort_reversed
+
+    def __iter__(self) -> Iterator[Type[Any]]:
+        members_sorted = sorted(
+            super().__iter__(),
+            key=lambda type_: type_.__name__,
+            reverse=self.sort_reversed,
+        )
+        yield from members_sorted
```
tiangolo, line 156: Very fancy! ✨ ✔️

tiangolo (approved):
Nice, sounds good, thank you! 🚀

PR #6041, opened 2023-02-24
✨ Add support for setting default values for `response_model` formatting

danielkatzan:
This PR is a follow up on this PR: https://github.com/tiangolo/fastapi/pull/2187 that was never merged

This would allow for setting the default value of `response_model_by_alias`, `response_model_exclude_unset`, `response_model_exclude_defaults` and `response_model_exclude_none` at an application or at a router level, rather than on each individual endpoint.

This is useful in cases where a consistent non-default behavior is desired across all endpoints and saves the need to declare so in each endpoint.


DanielSolomon (approved):

yuval-curv (approved):
Works in my env as well. looks good

pcorpet:
Anything blocking here? can we move forward with merging this PR?

Luiz-N:
any updates ?

nidhishgajjar:

## Summary

This PR adds the ability to set default values for `response_model_by_alias`, `response_model_exclude_unset`, `response_model_exclude_defaults`, and `response_model_exclude_none` at both the `FastAPI` app level and `APIRouter` level. Route-level settings take precedence over router-level, which takes precedence over app-level.

## Architecture

The implementation uses the existing `Default()` / `DefaultPlaceholder` pattern to distinguish between "not explicitly set" and "explicitly set to True/False". The cascade resolution is: route parameter → router default → app default. This is clean and consistent with how `response_class` and `generate_unique_id_function` already work.

Key additions:
- 4 new parameters on both `FastAPI.__init__()` and `APIRouter.__init__()`
- `get_value_or_default_value()` helper in `fastapi/utils.py`
- `include_router()` correctly resolves the cascade from included routes

## Issues

### 1. Type annotation mismatch in decorator methods (Warning)

**Files:** `fastapi/applications.py`, `fastapi/routing.py`

The route decorator methods (`.get()`, `.post()`, etc.) declare the parameter type as `bool` but use `Default(True)` as the default value:
```python
response_model_by_alias: bool = Default(True)  # type says bool, value is DefaultPlaceholder
```

Meanwhile, `add_api_route()` and `APIRoute.__init__()` correctly use:
```python
response_model_by_alias: Union[bool, DefaultPlaceholder] = Default(True)
```

This inconsistency will cause mypy/pyright errors. The decorator methods should also use `Union[bool, DefaultPlaceholder]`.

### 2. Documentation update is minimal (Suggestion)

**File:** `docs/en/docs/tutorial/response-model.md`

The doc update mentions the feature but doesn't show an example. Adding a short code snippet showing how to set these at the `FastAPI()` or `APIRouter()` level would be helpful:
```python
app = FastAPI(default_response_model_exclude_unset=True)
```

### 3. No `Doc()` annotations for new init parameters (Suggestion)

The new `default_response_model_by_alias` etc. parameters on `FastAPI.__init__()` and `APIRouter.__init__()` lack `Doc()` annotations, which means they won't appear in the generated API docs.

## Cross-file Impact

- **`fastapi/utils.py`**: New `get_value_or_default_value()` is clean and well-scoped.
- **`fastapi/routing.py`**: `APIRoute.__init__()` now unwraps `DefaultPlaceholder` values at route creation time, storing plain bools. This is correct and means `include_router()` can use `get_value_or_default()` for cascade resolution.
- **Tests**: Comprehensive parametrized tests covering all three levels of override (app → router → route). Well done.

## Assessment

**approve** — This is a well-implemented feature that follows existing patterns in the codebase. The type annotation inconsistency in decorator methods should be fixed for mypy compatibility, but the core logic is correct. Good test coverage demonstrates the cascade behavior clearly. This is a more complete solution than PR #9557 which only handles `response_model_by_alias`.

tiangolo:

Thank you! 🍰

PR #6083, opened 2023-02-27
✏ Fix wrong import from typing module in Persian translations for `docs/fa/docs/index.md`

Kimiaattaei:


tiangolo:
Good catch, thank you @Kimiaattaei! :cake:

PR #8222, opened 2023-02-28
Concurrency documentation consistency: Fixed first steps folder

francipvb:
Hello,

I saw that the documentation is not consistent according to the docs in the sync/async documentation page.

So I am sending a series of PRs to fix that little issue as my first contribution.

I looked at `doc_src` directory and there is the main issue, but I'll check tests folder later too.

The first folder to be fixed is the `first_steps` one and every PR will fix a folder.

Thank you for this library.


tiangolo:
Thanks for the interest, I would think this is actually as it's intended. If you read the docs for these files, it mentions that you can use `async def` or regular `def`, and it uses the different versions correspondingly. Check it here: https://fastapi.tiangolo.com/tutorial/first-steps/

Given that, I'll close this one. But if you think something's still wrong, please let me know. Maybe it would be better in a GitHub Discussion to share more information about why you think the current code is incorrect.

Thanks! :coffee:

PR #9245, opened 2023-03-12
🎨 Fix types for lifespan, upgrade Starlette to 0.26.1

tiangolo:
🎨 Fix types for lifespan, depends on https://github.com/encode/starlette/pull/2077

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -9,6 +9,7 @@ from typing import (
     Optional,
     Sequence,
     Type,
+    TypeVar,
     Union,
 )
 
@@ -31,7 +32,7 @@ from fastapi.openapi.utils import get_openapi
 from fastapi.params import Depends
 from fastapi.types import DecoratedCallable
 from fastapi.utils import generate_unique_id
-from starlette.applications import AppType, Starlette
+from starlette.applications import Starlette
 from starlette.datastructures import State
 from starlette.exceptions import HTTPException
 from starlette.middleware import Middleware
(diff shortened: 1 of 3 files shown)


adriangb:
You also need to add `self: AppType`

tiangolo:
Good point, thanks @adriangb! Trying it locally I realized I had to redefine `AppType` to be bound to `FastAPI` as well.

tiangolo:
I'm trying with this example:

```Python
from collections.abc import AsyncGenerator

from fastapi import FastAPI


@asynccontextmanager
async def lifespan(_: FastAPI) -> AsyncGenerator[None, None]:
    ...
    yield
    ...


app = FastAPI(lifespan=lifespan)
```

And I get this error:

```
mypy sand.py
fastapi/applications.py:112: error: "AppType" has no attribute "openapi_schema"
fastapi/applications.py:113: error: "AppType" has no attribute "openapi_url"
fastapi/applications.py:114: error: "AppType" has no attribute "title"
fastapi/applications.py:115: error: "AppType" has no attribute "version"
fastapi/applications.py:126: error: "AppType" has no attribute "dependency_overrides"
fastapi/applications.py:153: error: "AppType" has no attribute "setup"
Found 6 errors in 1 file (checked 1 source file)
```


...I guess it would be fine for external things to have `AppType` be bound to `Starlette`, but for the internals of `FastAPI` it seems to be a problem. But I guess it would be fine then to have the custom `AppType` here.

adriangb:
Weird, we even have a test in Starlette: https://github.com/encode/starlette/blob/8965f8849977b5752e415722a83aa00049f80ff3/tests/test_applications.py#L539-L550

tiangolo:
Ah, yeah, the error is not in the lifespan, but that the class doesn't have the custom attributes `FastAPI` defines.

PR #9262, opened 2023-03-15
add devcontainer definition

msander:
Hello,

I created a devcontainer definition for container-based development with vscode.
This helps to get started quickly and sets up an environment to be able to contribute to fastapi.

Regards Marcel


BilalAlpaslan:
idk this is the good way, adding to code directly. Maybe write a usage and adding contributing section. I think it can be more useful

tiangolo:
Thanks @msander! 🍰 

I'm currently not using devcontainers for FastAPI myself, so sadly I wouldn't really be able to maintain and make sure this file is up to date. I'm also not sure how many people are using devcontainers for working on FastAPI internals. For now, I'll pass on this, maybe at another point in the future.

Nevertheless, I think it would be a good idea for you to write a blog post about using devcontainers with FastAPI and then add the link in the external links section. 🤓

Ryandaydev:
Adding a comment for searchability. This request was intended to enable GitHub Codespaces configuration.

PR #9268, opened 2023-03-17
📝 Update all docs to use `Annotated` as the main recommendation, with new examples and tests

tiangolo:
📝 Update all docs to use `Annotated` as the main recommendation, with new examples and tests


nzig:
Hey @tiangolo, I'm not sure what you meant by the sentence
> Try to use the main, `Annotated` version better.

in all the examples. It sounds a bit weird to me. Maybe something like 
> Prefer to use the `Annotated` version if possible.

would be clearer?

tiangolo:
Sounds good! Just made another PR with that change here: https://github.com/tiangolo/fastapi/pull/9270

a-was:
What about database docs?
https://fastapi.tiangolo.com/tutorial/sql-databases/#create-your-fastapi-path-operations
Everywhere I can see `db: Session = Depends(get_db)`

PR #9273, opened 2023-03-19
✏ Fix typo/bug in inline code example in `docs/en/docs/tutorial/query-params-str-validations.md`

tim-x-y-z:
There was a missing comma in the docs that threw me off :)

Thanks for the great using `Annotated` !


tiangolo:
Oh! :scream: Good catch! Thank you! :coffee: :cake:

PR #9274, opened 2023-03-19
docs: restructure docs & source code links

neuodev:


tiangolo:
Thanks, I'm not really sure what was the intention with this change, but I don't think it's necessary, so I'll close this one. But thanks for the interest. ☕

PR #9280, opened 2023-03-19
swagger-ui 5

raphaelauv:


Ryandaydev (commented):
@raphaelauv SwaggerUI 5 is in pre-release stage. Is this a proposal for early testing?

PR #9289, opened 2023-03-21
add OpenAPIAnnotation params to add info to simple Annotated type

vincentsarago:
this is rather a complex solution to fix the issue we have with FastAPI 0.95 (see https://github.com/tiangolo/fastapi/discussions/9278). 

With `>=0.95`, a param of type `Path`(defined in the route) cannot be use as a param of type `Query` (defined in the endpoint function). This was something not officially documented but mentioned couple times (https://github.com/tiangolo/fastapi/issues/945#issuecomment-585993501) 

Enabling a `mix` of Query/Path, is pretty useful when you have a multi routes endpoint with optional parameters
```python

app = FastAPI()

@app.get("/test/{val}")
@app.get("/test")
def endpoint(val: str = Query("yo", description="a value")):
    return val
```

☝️ This code won't work with >=0.95 and give
```
File fastapi/dependencies/utils.py:461, in analyze_param(param_name, annotation, value, is_path_param)
    458 if field_info is not None:
    459     if is_path_param:
    460         assert isinstance(field_info, params.Path), (
--> 461             f"Cannot use `{field_info.__class__.__name__}` for path param"
    462             f" {param_name!r}"
    463         )
    464     elif (
    465         isinstance(field_info, params.Param)
    466         and getattr(field_info, "in_", None) is None
    467     ):
    468         field_info.in_ = params.ParamTypes.query

AssertionError: Cannot use `Query` for path param 'val'
```

### 2 Solutions:

- The first solution as proposed in https://github.com/tiangolo/fastapi/discussions/9278 is to not use any FieldInfo class (Path or Query), because FastAPI will be smart enough to assign either a `param.Query` or `param.Path` [here](https://github.com/tiangolo/fastapi/blob/d666ccb62216e45ca78643b52c235ba0d2c53986/fastapi/dependencies/utils.py#L429-L437) 

```python

app = FastAPI()

@app.get("/test/{val}")
@app.get("/test")
def endpoint(val: str = "yo"):
    return val
```

But then we can't have anything forwarded to the OpenAPI doc (title, description, ...) which is one of the nice feature of the FastAPI Query/Path param 😭  

- The second solution (as introduced by this PR) is to hypothetically use a neutral class (has no direct influence on how the code is run) which only purpose is to forward information to the OpenAPI doc:

```python
from fastapi import FastAPI
from fastapi.params import OpenAPIAnnotation

@app.get("/description/{foo}")
@app.get("/description")
async def description(
    foo: Annotated[
        str, 
        OpenAPIAnnotation(title="get a description", description="description")
    ]
) = None:
    return {"foo": foo}
```

I totally agree that adding a new class might not be the best solution but it's the only thing I could think off 😅 

I really appreciate any help/comment 🙏

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -367,6 +367,7 @@ def analyze_param(
     used_default_field_info = False
     depends = None
     type_annotation: Any = Any
+    metadata = None
     if (
         annotation is not inspect.Signature.empty
         and get_origin(annotation) is Annotated  # type: ignore[comparison-overlap]
@@ -376,7 +377,7 @@ def analyze_param(
         fastapi_annotations = [
             arg
             for arg in annotated_args[1:]
-            if isinstance(arg, (FieldInfo, params.Depends))
+            if isinstance(arg, (FieldInfo, params.Depends, params.OpenAPIAnnotation))
         ]
         assert (
             len(fastapi_annotations) <= 1
@@ -395,6 +396,11 @@ def analyze_param(
                 field_info.default = Required
         elif isinstance(fastapi_annotation, params.Depends):
             depends = fastapi_annotation
+        elif isinstance(fastapi_annotation, params.OpenAPIAnnotation):
+            # If of type `string` the fastapi_annotation
+            # is considered as the field description
+            metadata = fastapi_annotation
+
     elif annotation is not inspect.Signature.empty:
         type_annotation = annotation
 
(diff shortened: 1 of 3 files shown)


kirisakow:
>```py
>from fastapi import FastAPI
>from fastapi.params import OpenAPIAnnotation
>
>@app.get("/description/{foo}")
>@app.get("/description")
>async def description(
>    foo: Annotated[
>        str, 
>        OpenAPIAnnotation(title="get a description", description="description")
>    ]
>) = None:
>    return {"foo": foo}
>```

@vincentsarago missing `from typing import Annotated`

PR #9298, opened 2023-03-22
📝 Add notification message warning about old versions of FastAPI not supporting `Annotated`

grdworkin:
The new annotated dependency feature creates some to unexpected errors on old versions:

https://github.com/tiangolo/fastapi/discussions/9277
https://github.com/tiangolo/fastapi/discussions/9297

This adds in a quick warning to the dependencies tutorial for anyone about to integrate the new functionality.

Feel free to close this request without discussion if you don't think it will be beneficial. It won't bother me


tiangolo:
Great, thanks! I updated the message a bit and added it in the first two places in the docs where it is used. :nerd_face: :cake:

Issue #2077, opened 2020-09-19
Add support for query parameter serialization styles


jakub-bacic:
While implementing filtering API recently and I came across a problem with parsing list of comma-separated ids from query string:

**Example:**
```
GET /api/items?id=1,3,5
```

I know this type of parsing can be easily done by declaring `id` parameter as `str` and parsing it inside the function but then we're missing awesome OpenAPI doc features.

Please find the PR with proposed solution.

**OpenAPI Docs:** https://swagger.io/docs/specification/serialization/

**Comments regarding code changes:**
- all changes should be backward compatible
- there's no support for `style=deepObject` as it would require more changes (because query params cannot be non-scalar values at the moment)
- in case multiple values are present (e.g. `?id=1&id=2,3`) and `explode=False`, only last value is parsed (in this case `2,3`); it's consistent with the currently implemented behaviour in case of declaring parameter as non-sequence (e.g. `str`) and receiving more than one value in query string

codyflynnNC:
Any movement on this?

ycd:
@codyflynnNC OpenAPI specification allows this as Jakub mentioned, and he already opened a PR for this, so for now we are just waiting for @tiangolo's feedback.

adriangb:
This would be a great addition. There is obviously no singular way to do things, especially when you want to support filtering like `GET /api/items?id=gt:1,lt:3`. But it would be nice to have _a_ way to do this.

shufflebits:
It seems that the reticence on allowing for encoding styles for `deepObject` might be that there is no _standard_ way of encoding it? 
How about allowing for pluggable parameter encoding styles, so we can implement whatever is needed. As it stands, those of us trying to implement endpoints for populating tables that support filtering find it impossible to create a nice documented API that allows for `lt`, `gt` and so on comparison operators.

PR #9299, opened 2023-03-23
Add FromLifespan to extract dependencies from the lifespan state

adriangb:
This adds a `FromLifespan` annotation that extracts dependencies from lifespans, which I believe satisfies the requests in #617, #9215, part of #3641 and probably several other issues.

It also adds what is essentially an extractor system that allows dependencies to get a view into their parameter; this could effectively allow replacing a lot of `fastapi/dependencies/utils.py` with implementations for each type of parameter, making things like #2077 a lot easier to implement and opening up the door for 3rd party extensions to FastAPI's dependency / extractor system. This does the rest of #3641.

I purposefully added as single docs-style test to demonstrate how this is used and in-lieu of writing docs; I'll do that once / if there's some initial feedback and buy in.

### DIFF
--- a/fastapi/dependencies/_aliases.py
+++ b/fastapi/dependencies/_aliases.py
@@ -1,7 +1,7 @@
 from inspect import Parameter
 from typing import Any, Dict, Optional, Type, TypeVar, cast
 
-from fastapi.params import Deferred, Depends
+from fastapi.params import Deferred, Depends, DependsContext
 from fastapi.requests import Request
 from typing_extensions import Annotated, get_args, get_origin
 
@@ -19,8 +19,8 @@ def get_type(param: Parameter) -> Type[Any]:
 
 
 class _FromLifespan:
-    def __init__(self, param: Parameter) -> None:
-        self._type = get_type(param)
+    def __init__(self, ctx: DependsContext) -> None:
+        self._type = get_type(ctx.param)
         self._key: Optional[str] = None
 
     def __call__(self, request: Request) -> Any:
@@ -35,7 +35,7 @@ class _FromLifespan:
         return state[self._key]
 
 
-def _from_lifespan(param: Parameter) -> Depends:
+def _from_lifespan(param: DependsContext) -> Depends:
     return Depends(_FromLifespan(param))
 
 
--- a/fastapi/dependencies/_aliases.py
+++ b/fastapi/dependencies/_aliases.py
@@ -1,5 +1,5 @@
 from inspect import Parameter
-from typing import Any, Dict, Optional, TypeVar, cast
+from typing import Any, Dict, Optional, Type, TypeVar, cast
 
 from fastapi.params import Deferred, Depends
 from fastapi.requests import Request
@@ -11,11 +11,11 @@ T = TypeVar("T")
 LifespanState = Dict[str, Any]
 
 
-def get_type(param: Parameter) -> type[Any]:
+def get_type(param: Parameter) -> Type[Any]:
     annotation = param.annotation
     while get_origin(annotation) is Annotated:
         annotation = get_args(annotation)[0]
-    return annotation
+    return annotation  # type: ignore[no-any-return]
 
 
 class _FromLifespan:
--- a/fastapi/dependencies/_aliases.py
+++ b/fastapi/dependencies/_aliases.py
@@ -1,0 +1,42 @@
+from inspect import Parameter
+from typing import Any, Dict, Optional, TypeVar, cast
+
+from fastapi.params import Deferred, Depends
+from fastapi.requests import Request
+from typing_extensions import Annotated, get_args, get_origin
+
+T = TypeVar("T")
+
+
+LifespanState = Dict[str, Any]
+
+
+def get_type(param: Parameter) -> type[Any]:
+    annotation = param.annotation
+    while get_origin(annotation) is Annotated:
+        annotation = get_args(annotation)[0]
+    return annotation
+
+
+class _FromLifespan:
+    def __init__(self, param: Parameter) -> None:
+        self._type = get_type(param)
+        self._key: Optional[str] = None
+
+    def __call__(self, request: Request) -> Any:
+        state = cast(LifespanState, request["state"])
+        if self._key is None:
+            for key, value in state.items():
+                if isinstance(value, self._type):
+                    self._key = key
+                    break
+        if self._key is None:
+            raise RuntimeError(f"{self._type} not found in lifespan state")
+        return state[self._key]
+
+
+def _from_lifespan(param: Parameter) -> Depends:
+    return Depends(_FromLifespan(param))
+
+
+FromLifespan = Annotated[T, Deferred(_from_lifespan)]
--- a/fastapi/params.py
+++ b/fastapi/params.py
@@ -1,4 +1,5 @@
 from enum import Enum
+from inspect import Parameter
 from typing import Any, Callable, Dict, Optional, Sequence
 
 from pydantic.fields import FieldInfo, Undefined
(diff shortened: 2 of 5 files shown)


fastapi/dependencies/_aliases.py
```diff
@@ -0,0 +1,42 @@
+from inspect import Parameter
```
adriangb, line 1: I wasn't sure where to put this, I leave that up to you

fastapi/dependencies/_aliases.py
```diff
@@ -0,0 +1,42 @@
...
+    annotation = param.annotation
+    while get_origin(annotation) is Annotated:
+        annotation = get_args(annotation)[0]
+    return annotation  # type: ignore[no-any-return]
+
+
+class _FromLifespan:
+    def __init__(self, param: Parameter) -> None:
+        self._type = get_type(param)
+        self._key: Optional[str] = None
```
adriangb, line 24: One interesting thing about this API is that there might be some type information available from the lifepsan itself. Should FastAPI introspect that? Should it force users to return something typed? Would that help prevent mistakes at startup, e.g. where the user requests a type that is not available or is maybe ambiguous?

fastapi/dependencies/_aliases.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+LifespanState = Dict[str, Any]
+
+
+def get_type(param: Parameter) -> Type[Any]:
+    annotation = param.annotation
+    while get_origin(annotation) is Annotated:
+        annotation = get_args(annotation)[0]
+    return annotation  # type: ignore[no-any-return]
```
adriangb, line 18: There probably is (or maybe could be) a utility function somewhere else that does this and other manipulation of `Annotated`.

fastapi/params.py
```diff
@@ -379,3 +380,16 @@ def __init__(
     ):
         super().__init__(dependency=dependency, use_cache=use_cache)
         self.scopes = scopes or []
+
+
+class DependsContext:
+    def __init__(self, param: Parameter) -> None:
+        self.param = param
```
adriangb, line 387: I'm wrapping the single item in a custom class so that in the future we can add more stuff here in a backwards compatible manner. E.g. I can imagine a world where a dependency wants to know what path it is registered under, what dependencies depend on it, etc.

Kludex:
Isn't the following the same thing?

```python
```

Kludex:
I'm failing to see how all those issues are related to this. :thinking: 

Don't we want a dependency that can be accessed on the lifespan itself? Like the "app" scope that you have on xpresso.

If I do the following, isn't it the same as what this PR proposes?
```python
from contextlib import asynccontexmanager
from typing import Annotation

app = FastAPI()

@asynccontextmanager
async def lifespan(app: FastAPI) -> dict:
    async with connect() as db:
        yield {"db": db}

async def get_connection(request: Request):
    yield request.state.db

@app.get("/")
async def homepage(db: Annotation[Session, Depends(get_connection)]):
    ...
```

adriangb:
Yes, it's the same thing. This is just making it more seamless.

tests/test_dependency_from_lifespan.py
```diff
@@ -0,0 +1,72 @@
...
+            yield DatabaseConnection(self.expected)
+
+    @asynccontextmanager
+    async def connect(expected: int) -> AsyncIterator[DatabaseConnectionPool]:
+        yield DatabaseConnectionPool(expected)
+
+    @asynccontextmanager
+    async def lifespan(app: FastAPI) -> AsyncIterator[Dict[str, Any]]:
+        async with connect(42) as db:
+            yield {"db": db}
```
Chippiewill, line 34: I'm not a big fan of this design as a solution to lifespan dependency, it's really just syntactic sugar over the lifespan `state` like this.

The value of dependencies is the ability to inject dependencies in a more distributed way. For example I might have a lifespan dependency that's only applicable to one group of API routes so I want to define it as a lifespan dependency in that context rather than having to define it on a lifespan passed directly into the main FastAPI application.

adriangb: I think there’s actually quite a bit of value in having users manually wire the dependencies in the lifespan. It mirrors the “composition root” concept in most standalone dependency injection systems. It’s a singular well identified place where you tell the dependency injection system how to build your dependencies.

This keeps the execution model clear and simple. I’ve implemented the more “distributed” version in Xpresso and it can get messy.

YuriiMotov (commented):
As I understand it, this PR aims to provide the ability to simplify:
```python
    async def get_connection(
        request: Request,
    ) -> AsyncIterator[DatabaseConnection]:
        state = cast(dict[str, Any], request["state"])
        db = cast(DatabaseConnectionPool, state["db"])
        async with db.acquire() as conn:
            yield conn
```
to
```python
    async def get_connection(
        db: FromLifespan[DatabaseConnectionPool],
    ) -> AsyncIterator[DatabaseConnection]:
        async with db.acquire() as conn:
            yield conn
```

**Advantages I can see in this approach:**
* user code is shorter
* improved typing (no need to cast in user code)

**Drawbacks of this implementation:**
* you can't have multiple instances of dependencies of the same type
* _(not really important)_ less control over the error handling in case of misconfiguration (error is hardcoded)

To solve the problem with multiple instances, we can add optional `key` parameter (extract by `key` if provided, otherwise search by type)

**UPD:** 
As an alternative, we could add the ability to inject typed request state:
```python
    class MyState(TypedDict):
        db: DatabaseConnectionPool

    async def get_connection(
        state: AppState[MyState],
    ) -> AsyncIterator[DatabaseConnection]:
        async with state["db"].acquire() as conn:
            yield conn
```

YuriiMotov:
@adriangb, thanks for your efforts!
Are you still interested in working on this?
I think we need to ask Sebastián to confirm the idea of this feature first

adriangb:
No I won't have time to work on this. I've packaged it up as an external package fwiw: https://github.com/adriangb/asapi

nidhishgajjar:

## Review of PR #9299: Add FromLifespan to extract dependencies from the lifespan state

This is an elegant feature design that solves a long-standing request. The `Deferred` mechanism is a particularly nice abstraction that opens up possibilities for other parameter-aware dependency patterns.

### Design observations 💡

1. **Type-based lookup ambiguity**: The `_FromLifespan.__call__` method searches for values by type using `isinstance()`:
   ```python
   for key, value in state.items():
       if isinstance(value, self._type):
           self._key = key
           break
   ```
   If the lifespan state contains multiple values of the same type, this picks the first match (dict order). Consider supporting an explicit key override, e.g. `FromLifespan[DatabaseConnection, "db"]` or a separate parameter, to avoid ambiguity in larger applications.

2. **Subclass matching**: `isinstance()` will match subclasses of the annotated type. This could be surprising if someone has `class ReadOnlyDB(DB): ...` and `class ReadWriteDB(DB): ...` in state — requesting `FromLifespan[DB]` would get whichever is iterated first.

3. **`Deferred` is a great abstraction**: The parameter-aware dependency factory pattern is clean and extensible. This could enable many similar patterns (e.g., extracting from request state, headers, etc.) without further changes to `analyze_param`.

### What looks good ✅

- **Clean type alias**: `FromLifespan[T]` reads naturally as a type annotation
- **Key caching**: First lookup is O(n), subsequent lookups are O(1) via `_key` cache
- **Proper error handling**: `RuntimeError` with a descriptive message for missing types
- **`get_type` helper**: Correctly unwraps nested `Annotated` layers
- **Test coverage**: Both success and failure cases are tested
- **Internal API refactoring**: Changing `analyze_param` to accept `param` instead of separate values is cleaner

### Minor notes

- Consider a more specific exception type than `RuntimeError` (e.g., a `DependencyError` or `LifespanStateError`) for better error handling by users
- The `_key` attribute on `_FromLifespan` starts as `None` and is set on first call — since FastAPI caches dependency results by default, this is only called once per dependency instance, so thread safety is not a concern in practice

### Summary

Well-designed feature with a clean, extensible architecture. The `Deferred` mechanism is a particularly thoughtful addition that enables more than just lifespan state extraction. The type-based lookup is a reasonable default, but an explicit key option would make it more robust for complex applications.

**Assessment: approve**

PR #9312, opened 2023-03-25
fix: give correct unauthorized response on OpenID Connect Auth

rhuanbarreto:
According to https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.2 the correct response in this case of missing value is 401.


tomarsnap (approved):

tiangolo:
Thanks @rhuanbarreto, but with a 401:

> MUST send a [WWW-Authenticate](https://www.rfc-editor.org/rfc/rfc9110.html#field.www-authenticate) header field

And as it's not clear which `WWW-Authenticate` header field should be used, not providing it is actually invalid, and although 403 is not the best error code, it's a valid option. So, for now, I'll leave it as is.

Later I'll do some more research to see if there's a simple way to provide the correct `WWW-Authenticate` values and then allow using 401.

But thanks for the interest! ☕

PR #9313, opened 2023-03-25
📝 Update usage of Token model in security docs

piotrszacilowski:
In the documentation regarding security in the code were created Token model, while it was not used. I have fixed it.


piotrszacilowski:
anyone willing to review?

tiangolo:
📝 Docs preview for commit 4cd6cc93f505cc4cad53944781033b3b341aa355 at: https://fdcbc765.fastapitiangolo.pages.dev

alejsdev:
Thanks for your help! @piotrszacilowski 🙂 

The Token was actually used in the `response_model` but indeed it was better to return an instance of it. 

I also updated the response model to use the Token as a return type and updated the highlighted lines in the docs.

Thanks 😸

PR #9314, opened 2023-03-25
Raise RequestValidationError on ValueError from a Depedency

1337vida:
Adds the same functionality as @validator from PyDantic to Headers, Cookies, Path and Query params.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -642,10 +642,16 @@ async def solve_dependencies(
         except ValueError as exc:
             if sub_dependant.name is not None:
                 field = field_map[sub_dependant.name]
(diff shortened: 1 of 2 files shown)


90:
LGTM

YuriiMotov (commented):
@1337vida, thanks for your interest and efforts!

I agree that it would be nice to provide an easy way to raise `RequestValidationError` from inside the dependency. But I think it should be done in different way.

Sometimes you may have dependencies that use Pydantic models internally (e.g. to extract data from DB and validate it) and they may also raise `ValueError`.
Such errors aren't user-errors and we shouldn't return 422 error response for them, but rather 500 server error response.

PR #9319, opened 2023-03-27
✨ Allow duplicated variable names in dependencies

### LINKED ISSUE #4610: Body() fields are ignored if they are mapped to same variable name in any dependency
### First Check

### Commit to Help

### Example Code

```python
from typing import Awaitable, Callable

from fastapi import APIRouter, Body, Depends


api_router = APIRouter()


def _make_field(name: str) -> Callable[..., Awaitable[str]]:
    async def _inner(
        value: str = Body(..., alias=name),  # <<<<<<<<<<<<<< HERE
        # ^^^ HERE
    ) -> str:
        return value

    return _inner


@api_router.post("/test1")
async def test(
    _field_0: str = Body(..., alias="field_0"),
    _field_1: str = Depends(_make_field("field_1")),
    _field_2: str = Depends(_make_field("field_2")),
) -> None:
    pass
```

### Description

When two ore more `Body()` fields are added indirectly and are bound to the same variable name, only the last one will be visible in docs. Aliasing them will not make any difference.

The example above will show only field_0 and field_2 in the docs, while field_1 is missing.
When trying to execute the endpoint it will operate correctly.

If there are two different dependencies using internally the same variable name for different body values, it will also not work.

### Operating System

Other

### Operating System Details

_No response_

### FastAPI Version

0.74.1

### Python Version

3.10.1

### Additional Context

_No response_

kkosyl-rtbhouse:
Fixes https://github.com/tiangolo/fastapi/issues/4610


Ryandaydev (approved):
Tests pass

peku33:
Hi, is there any chance to have this merged anytime soon?

jkbak:
@Kludex I see that you added an `investigate` label. Is that still blocking this merge?

peku33:
@Ryandaydev, maybe you have a chance to merge it after 2 months since it was initially created?

Ryandaydev:
> @Ryandaydev, maybe you have a chance to merge it after 2 months since it was initially created?

@tiangolo merges the PRs.

peku33:
Is there anything else we can do, to move this forward? I believe we did everything to have it ready for merge. Three months later there is no single sign of decision in any direction and the bug still exists. @tiangolo @Ryandaydev

YuriiMotov (dismissed):
LGTM

`test_openapi_schema` test fails before changes in both Pydantic V1 and V2.
This PR fixes it.

`test_valid` and `test_missing` pass in current version (0.116.0).

I would rename this PR to `Fix openapi schema for single Body parameters with aliases`

YuriiMotov (changes requested):
There is a problem here.
It only works if we have multiple body parameters (with different variable names) or set `embed=True`.
Otherwise FastAPI fails to understand that it should create body model for the endpoint.

```python
from fastapi import Body, Depends, FastAPI

app = FastAPI()

# Without embed=True

def dep_1(param: str = Body(..., alias="param_1")):
    return param


def dep_2(param: str = Body(..., alias="param_2")):
    return param


@app.post("/example")
def example(
    field_1: str = Depends(dep_1),
    field_2: str = Depends(dep_2),
) -> list[str]:
    return [field_1, field_2]

# With embed=True

def dep_1_emb(param: str = Body(..., alias="param_1", embed=True)):
    return param


def dep_2_emb(param: str = Body(..., alias="param_2", embed=True)):
    return param


@app.post("/example-emb")
def example_emb(
    field_1: str = Depends(dep_1_emb),
    field_2: str = Depends(dep_2_emb),
) -> list[str]:
    return [field_1, field_2]
```
This will show incorrect schema for `/example` endpoint
image

That's because FastAPI treats those parameters as one and doesn't force embedding them.

If we set `embed=True` then it works as we expect it (as it's shown in `/example-emb` endpoint):
image

@kkosyl-rtbhouse, could you please take a look?

Also, I think tests for `Form` and `File` parameters should also be added. What do you think?

PR #9334, opened 2023-03-30
Makes HTTPException picklable

dkuryakin:
Following simple code

```python
import pickle

from fastapi.exceptions import HTTPException

x = pickle.dumps(HTTPException(status_code=415, detail={'foo': 'bar'}))
xx = pickle.loads(x)
```

leads to an error:

```
  File "...", line 6, in <module>
    xx = pickle.loads(x)
TypeError: __init__() missing 1 required positional argument: 'status_code'
```

This behavior can breaks serialization/deserialization mechanics in case of complex systems (rpc/queues/etc).

This PR makes HTTPException unpicklable.

### DIFF
--- a/fastapi/exceptions.py
+++ b/fastapi/exceptions.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Optional, Sequence, Type, Tuple
+from typing import Any, Dict, Optional, Sequence, Tuple, Type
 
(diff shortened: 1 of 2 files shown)


Kludex:
Ideally, this should be added in Starlette, but it was already discussed over there: https://github.com/encode/starlette/pull/1670

On the link, you can find an alternative on how to pickle the `HTTPException` class.

PR #9340, opened 2023-04-02
Added `response_dict_by_alias` parameter.

awtb:
Hello, FastAPI people!

At work, I have a small package for iteracting with billing-system.
So, package uses pydantic-models for its schemas, but FastAPI converts pydantic-model to dict with `by_alias=True` flag.
I added `response_dict_by_alias` flag for disabling this stuff, why?

Let's write a mock-service 

```python
# external_service.py

class ExternalSchema(BaseModel):
    field_name: str = Field(alias="FLD_NAME")


class ExternalService:
    """
    Some external service
    IMPORTANT: This service uses pydantic for its models.
    """
    def __init__(self):
        self.initialized = True

    def get_external_data(self) -> ExternalSchema:
        # Just imitation

        data = {
            "FLD_NAME": "Data!"
        }

        return ExternalSchema.parse_obj(data)
```

Okay, now let's write a `main.py` with our FastAPI app.

```python
# main.py

from .external_service import ExternalService

class ResponseSchema(BaseModel):
    field_name: str = Field(alias="fieldName")
    
    class Config:
        allow_population_by_field_name = True    


app = FastAPI()

def get_client() -> ExternalService:
    # Just dependency, initializes service & returns it.
    return ExternalService()

@app.get("/", response_model=ResponseSchema)
def index(client: ExternalService = Depends(get_client)):
    data = client.get_external_data()

    return data

```

Now, run it!

```bash
uvicorn main:app
```

Make request `GET localhost:8000/` and get this error.

```bash
pydantic.error_wrappers.ValidationError: 1 validation error for ResponseSchema
response -> fieldName
  field required (type=value_error.missing)
```

Why? Because `_prepare_response_content` generates dictionary with `by_alias=True`, so instead of `field_name` `FLD_NAME` will be passed, adding a flag will fix this.

Let's use `response_dict_by_alias=False` flag

```python
@app.get("/", response_model=ResponseSchema, response_dict_by_alias=False)
def index(client: ExternalService = Depends(get_client)):
    data = client.get_external_data()

    return data
```

Make request again!

```json
{
  "fieldName": "Data!"
}
```

Problem solved!


tests/test_response_dict_by_alias.py
```diff
@@ -0,0 +1,58 @@
...
+def by_field_name(service: MockService = Depends(get_mock_service)):
+    data = service.get_data()
+
+    return data
+
+
+client = TestClient(app)
+
+
+def test_response_dict_by_alias():
```
CiaronHowell, line 49: I'd recommend adding a test that you know will fail e.g. asserting the status code to be 422 for a validation error. The example you give in your summary would be good to replicate as this would cover any changes to the underlying functionality e.g., if, without setting `response_dict_by_alias` to true, we can set by dict name rather than alias then we should be notified as some changes might have made this bool redundant or some underlying logic has been changed which has some unwanted side effects.

CiaronHowell (changes requested):
Other than my recommendation, everything looks good. Not sure if or where you should document this new flag unfortunately so hopefully someone else can comment on that. 

Thank you for your clear summary of the issue, I can clearly see scenarios where this would be useful!

YuriiMotov:
@awtb, thanks for your interest in FastAPI!

Seems that this solves quite niche problem (usually you will probably use consistent approach to give names for fields) and the number of people who are interested in this PR being merged approves this.

Even if you can't modify `ExternalSchema` model to make naming consistent, it's still possible to solve this using `validation_alias` in Pydantic V2 or `validator` in Pydantic V1.

So, I would close this.
Feel free to bring additional arguments if you think I'm missing something

tiangolo:
I would think what you need is Pydantic's v2 independent serialization and validation alias: https://docs.pydantic.dev/latest/concepts/alias/#Alias

I would prefer not to add another parameter that would alter the response content but violating the schema in OpenAPI (there are a couple there that I would like to deprecate as they end up creating more issues than helping).

For now I'll pass on this one but thanks for the interest! :coffee:

PR #9348, opened 2023-04-04
Decorators parameter for APIRouter and app.include_router to wrap all endpoints in router inside decorator(s)

richardnm-2:
After a thorough search on `middlewares` and not being able to solve [this scrambling issue with middlewares](https://stackoverflow.com/questions/75706159/async-warnings-generation-and-catching-inside-starlette-middleware-using-fastapi), I thought it would be nice to be able to wrap every endpoint in a router in a decorator (or multiple decorators). #9244

First I tested looping through all the routes as below, but to get it working, I had to update the dependant as well, which felt a little too hacky:

```
for route in router.routes:
    route.endpoint = decorator(route.endpoint)
    route.dependendant = get_dependant(
        path=route.path_format, call=route.endpoint)
```

So I came up with the `decorators` parameter for the `APIRouter` and for the `app.include_router`, passing a list of tuples where the first element of the tuple should be the decorator function and the other elements, the decorator arguments, if any.

`router_decorated = APIRouter(decorators=[(decorator_1, ), (decorator_2, DECORATOR_ARG)])`

In branch `richardnm2/fastapi/tree/dummy_decorator_test` I forced all routes to be wrapped around a dummy decorator, and all tests still passed.

Looking forward for your thoughts on this one, would really help for my use case, described in the SO thread.


tiangolo:
Thanks! About your specific need with warnings, I replied in your question. In short, you shouldn't be using `warning.catch_warnings` for production stuff in concurrent code (web, etc.) as it's not thread/concurrent safe.

About this specific PR, I don't think this approach would solve the current situation well, and it would add some logic that would need to be explained and understood, but that is not necessarily useful for most people. Also, as what it receives are functions that should wrap other functions but are not really decorating them (they aren't really put on top with `@`), I wouldn't think the complexity added would justify the feature.

So, for now, I'll pass on this one, but thanks for the interest. ☕

PR #9367, opened 2023-04-08
support for configuring CSS, JS and favicon for Swagger UI html.

lucas-six:


tiangolo:
Hello there, the best way to achieve this would be following the same ideas in the docs for self-hosting the JS and CSS assets: https://fastapi.tiangolo.com/advanced/extending-openapi/#self-hosting-javascript-and-css-for-docs

That way it's more flexible, it allows you to customize everything even further.

Given that, I'll pass on this one, but thanks for the effort! 🍰

PR #9383, opened 2023-04-12
🐛 Fix handling of empty value for `byte` file list parameter

Viicos:
Fixes https://github.com/tiangolo/fastapi/discussions/9364

It appears that the value list extracted from the `FormData` is not necessarily composed of only `UploadFile` instances.

OP said that it was happening when using the Swagger UI, I haven't tested without. ~~Let me know if you'd like a test for this issue.~~ seems to be required for coverage, I'll take a deeper look


YuriiMotov:
@Viicos, hope you don't mind I pushed some updates.

I think the most correct approach would be to return error 422 as for `UploadFile`, but this would require significant changes..

I suggest we mark this PR as ready and ask Sebastian to review and decide which way to solve this we should choose

PR #9315, opened 2023-03-26
🐛 Fix using `Annotated` in routers or path operations decorated multiple times

sharonyogev:
We need to deep clone the field_info to prevent ourself from mutating it.  This allows multiple path or nested routers ,etc.

Should solve those discussions:
https://github.com/tiangolo/fastapi/discussions/9279
https://github.com/tiangolo/fastapi/discussions/9309
https://github.com/tiangolo/fastapi/discussions/9306

PR #9389, opened 2023-04-13
✨ Allow Annotated FieldInfo with default values

sharonyogev:
This allows creating annotated arguments (Body/Path/Query) with default value definition as part of the FieldInfo instance.

Which allows:
1) setting up default value for the FastAPI endpoint that will not be part of the function definition if called directly (see test). 
2) setting the default value in a shared annotated type e.g. `Verbose=Annotated[bool, Query(True)]`.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -344,14 +344,16 @@ def analyze_param(
             field_info = copy_field_info(
                 field_info=fastapi_annotation, annotation=annotation
             )
-            assert field_info.default is Undefined or field_info.default is Required, (
-                f"`{field_info.__class__.__name__}` default value cannot be set in"
-                f" `Annotated` for {param_name!r}. Set the default value with `=` instead."
-            )
             if value is not inspect.Signature.empty:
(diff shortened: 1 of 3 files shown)


sharonyogev:
@tiangolo Thanks for merging #9315 🙏 
I needed it to open this PR as it was based on it...
Also, I wonder how do you feel with this suggested feature?
I find it fairly simple and helpful, you might be able to achieve it using an extra Depends, but this I think this is way shorter and also fits well to the FastAPI & Pydantic ecosystem

wu-clan:
Need it badly

tiangolo:
Thanks @sharonyogev!

But I think that the default value should be on the default value of the parameter, not having a default value in the function but having it in `Annotated` would mean that the code would behave in one way in some situations (e.g. with FastAPI) and a different way in others (calling it directly). And avoiding that was one of the main reasons to support `Annotated`, that the code should behave as consistently as possible to make it easier to use, reuse, and think about.

So I'll pass on this, but thanks for the interest! ☕

PR #9442, opened 2023-04-25
🧪 Pydantic v2 migration, investigation, debug

tiangolo:
🧪 Pydantic v2 migration, investigation, debug


tiangolo:
I'll continue this in https://github.com/tiangolo/fastapi/pull/9500, with a branch on this repo to be able to push to it.

PR #9443, opened 2023-04-26
🌐 improve text for benchmark explanation on pt and en documentation

felipythondev:
The goal of the PR is to update the version in English and in Portuguese of the documentation.


docs/en/docs/benchmarks.md
```diff
@@ -1,6 +1,6 @@
 # Benchmarks
 
-Independent TechEmpower benchmarks show **FastAPI** applications running under Uvicorn as <a href="https://www.techempower.com/benchmarks/#section=test&runid=7464e520-0dc2-473d-bd34-dbdfd7e85911&hw=ph&test=query&l=zijzen-7" class="external-link" target="_blank">one of the fastest Python frameworks available</a>, only below Starlette and Uvicorn themselves (used internally by FastAPI). (*)
+Independent TechEmpower benchmarks show **FastAPI** applications running among Uvicorn and Starlette (used internally by FastAPI) as <a href="https://www.techempower.com/benchmarks/#section=test&runid=7464e520-0dc2-473d-bd34-dbdfd7e85911&hw=ph&test=query&l=zijzen-7" class="external-link" target="_blank">one of the fastest Python frameworks available</a>.
```
tiangolo, line 3: Thanks! Why do you think "among" would make more sense than "under Uvicorn"?

docs/pt/docs/benchmarks.md
```diff
@@ -1,6 +1,6 @@
 # Comparações
 
-As comparações independentes da TechEmpower mostram as aplicações **FastAPI** rodando com Uvicorn como <a href="https://www.techempower.com/benchmarks/#section=test&runid=7464e520-0dc2-473d-bd34-dbdfd7e85911&hw=ph&test=query&l=zijzen-7" class="external-link" target="_blank">um dos _frameworks_ Python mais rápidos disponíveis</a>, somente atrás dos próprios Starlette e Uvicorn (utilizados internamente pelo FastAPI). (*)
+As comparações independentes da TechEmpower mostram as aplicações **FastAPI** rodando com Uvicorn como <a href="https://www.techempower.com/benchmarks/#section=test&runid=7464e520-0dc2-473d-bd34-dbdfd7e85911&hw=ph&test=query&l=zijzen-7" class="external-link" target="_blank">um dos _frameworks_ Python mais rápidos disponíveis</a>, junto com os próprios Starlette (utilizado internamente pelo FastAPI).
```
tiangolo, line 3: It would be better to have the update in the original docs as one PR and then the update for the Portuguese version as a secondary PR that can be reviewed independently, and after the first one is concluded.

tiangolo (commented):
Thanks! I have a couple of questions/comments.

docs/en/docs/benchmarks.md
lipemorais: Hi, @tiangolo ! Thanks for the questions. 

We have talked about it in person at PyCon. The problem is that when you click on the link, FastAPI doesn't appear, and when it appears with a few clicks like [this](https://www.techempower.com/benchmarks/#section=test&runid=7464e520-0dc2-473d-bd34-dbdfd7e85911&hw=cl&test=query&l=zijzen-7) it's above Starlette and Uvicorn. That's why I mentioned "among" instead of "under".
image

docs/pt/docs/benchmarks.md
Kludex: Where did I say that?

Kludex: Ah, well... I don't remember.

But since we were on the same place, I might have said "...but check with him first", pointing at Sebastian hahaha didn't I?

tiangolo:
Thanks @lipemorais! The word "among" is not what I meant here, it was what is written there, as the internal implementation of those tests are all using Uvicorn, so all of them, FastAPI, Starlette (and pure Uvicorn) use Uvicorn.

So I'll pass on this one, but thanks for the interest! ☕

PR #9453, opened 2023-04-27
✏ Allow integer constraints on openapi schema

alesanfra:
Hi everyone,

I'm opening this PR because I've recently found a weird behavior when dealing with constrained integers. Consider the following example:

```python
from typing import Annotated

app = FastAPI()


@app.get("/item")
async def get_items(limit: Annotated[int, Query(ge=1, le=1000)] = 100):
    return {"message": f"Got items with limit {limit}"}
```

It will produce the following schema (not relevant sections elided):
```json
{
  "paths": {
    "/item": {
      "get": {
        "summary": "Get Items",
        "operationId": "get_items_item_get",
        "parameters": [
          {
            "required": false,
            "schema": {
              "title": "Limit",
              "maximum": 1000.0,
              "minimum": 1.0,
              "type": "integer",
              "default": 100
            },
            "name": "limit",
            "in": "query"
          }
        ],
  },
}
```

As you can see even though `ge=1` and `le=1000` were defined as integers, the internal pydantic model coherces them to floats (`"maximum": 1000.0` and `"minimum": 1.0`). This is a problem when using some security tools like [42 Crunch](https://42crunch.com/) because the mismatch between type and values is reported as a possible vulnerability.

With this PR fastapi will keep the constraints as defined in the source code leveraging the [smart union feature of pydantic](https://docs.pydantic.dev/usage/model_config/#smart-union), therefore in example above fastapi will produce the following schema:

```json
{
  "paths": {
    "/item": {
      "get": {
        "summary": "Get Items",
        "operationId": "get_items_item_get",
        "parameters": [
          {
            "required": false,
            "schema": {
              "title": "Limit",
              "maximum": 1000,
              "minimum": 1,
              "type": "integer",
              "default": 100
            },
            "name": "limit",
            "in": "query"
          }
        ],
  },
}
```

### DIFF
--- a/fastapi/openapi/models.py
+++ b/fastapi/openapi/models.py
@@ -103,11 +103,11 @@ class ExternalDocumentation(BaseModel):
 class Schema(BaseModel):
     ref: Optional[str] = Field(default=None, alias="$ref")
     title: Optional[str] = None
-    multipleOf: Optional[float] = None
-    maximum: Optional[float] = None
-    exclusiveMaximum: Optional[float] = None
-    minimum: Optional[float] = None
-    exclusiveMinimum: Optional[float] = None
+    multipleOf: Optional[Union[float, int]] = None
+    maximum: Optional[Union[float, int]] = None
+    exclusiveMaximum: Optional[Union[float, int]] = None
+    minimum: Optional[Union[float, int]] = None
+    exclusiveMinimum: Optional[Union[float, int]] = None
     maxLength: Optional[int] = Field(default=None, gte=0)
     minLength: Optional[int] = Field(default=None, gte=0)
     pattern: Optional[str] = None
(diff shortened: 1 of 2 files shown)


leotac (approved):
LGTM, much needed.

matheushenrique98 (changes requested):
You are changing 5 `fields` but:
1 - Not all fields you are changing have test (multipleOf)
2 - The change occurs in 5 fields, but you have not adjusted all the tests who use these fields

About point 2: the type `number` could be more representative, because can be (`int` or `float`), and is a pattern used in other tests for the same context

YuriiMotov:
Duplicate: https://github.com/fastapi/fastapi/pull/9517

This will convert values `100.0` to just `0` in schema.

tiangolo:
I would think the behavior from that tool is a bit strange. In JSON, there's no distinction between floats and integers, there's only one type "number". It's weird (and probably wrong) that a tool would flag and interpret different types of numbers. :thinking: 

Do you have a specific use case that needs this other than complying with what the tool is telling you?

PR #9495, opened 2023-05-07
fix schema from docs_src/sql_app_py310

dfiallo35:
BaseModel does not support fields of type "<class 'types.UnionType'>"


tiangolo:
Looking at the change, I would suspect that you were trying an example for Python 3.10 in a lower version of Python (like Python 3.9).

Although I'm not sure with the description of the PR. 🤔 

It would be better to start with a question, filling all the template, that would help me/us understand your use case better:  https://github.com/tiangolo/fastapi/discussions/new?category=questions

Given that, I'll close this one, but thanks for the interest! ☕

PR #9498, opened 2023-05-08
Remove websockets package

dennis-rall:
The installation of the 'websockets' package is no longer required since the websockets package is part of the uvicorn[standard] installation.

In fact, I had some problems when I installed this package manually. Specifically, a SecurityError 'line too long' was thrown during the websocket handshake when there were too many cookies on the page. Removing the installation of the package fixed this.

That being said, it is a little confusing to install a package in a tutorial and not use it. I think without this it is clearer and easier.


tiangolo:
Thanks for the interest! But this would only apply if people installed `uvicorn[standard]`, if they didn't they would need to ensure they have it. :thinking: 

If you had problems installing it, it would be better to create a new discussion with your use case so we can check it.

For now, I'll pass on this one, but thanks for the effort!

dennisrall:
Yeah, but `uvicorn[standard]` is a dependecy of `fastapi[standard]`, so if `fastapi` is installed (which you assume in the tutorial), `uvicorn` and thereby `websockets` are installed. So I don't see the need for the explicit install, indeed I find it more confusing than helpful.

PR #9499, opened 2023-05-08
✨ Add possibility to customize the response class for `openapi.json` endpoint

flyaroundme:
Hi! I found out that in FastAPI it is not possible to configure a response class for openapi.json. In case of a very large openapi.json response, faster json encoders like orjson would come in handy. In this pull request, I suggest parameterizing the openapi.json response class in the init method of the FastAPI.


YuriiMotov (commented):
@flyaroundme, thanks for your interest and for your efforts!

Not sure it's really important to specify response class for `openapi.json` endpoint, but it seems that there is no simple way to do it for now. The only way I know is to disable docs and add your own implementation for `openapi.json` and `/docs` endpoints.

Don't see a problem in adding such option.
We will need to add a docstring to `openapi_response_class` parameter

tiangolo:

Thank you! 🍰

PR #9503, opened 2023-05-08
✅ Refactor OpenAPI tests, prepare for Pydantic v2

tiangolo:
✅ Refactor OpenAPI tests, prepare for Pydantic v2

During the migration, it will be very useful to check the tests with `insert_assert` from https://github.com/samuelcolvin/python-devtools

This prepares the code for that.

---

Before this change, there was normally a variable with `openapi_schema`, containing all the OpenAPI schema for each app tested. Then that was checked in a test for that.

But this doesn't allow easily inlining the result with `insert_assert`. By moving that schema to inside of the test, I'll be able to explore with `insert_assert` and verify in the diff what is wrong (later during the migration to Pydantic v2).

This needed to be done here, before that, because it only moves code/data around. But the diff is gigantic, so it shouldn't be mixed with code changing actual logic.

I also wanted to do it, to be able to keep using `insert_assert`, but that needed many consecutive hours of work, and a reason to go through all that work. But now this is a good reason to do it. :nerd_face:

### DIFF
--- a/tests/test_additional_properties.py
+++ b/tests/test_additional_properties.py
@@ -19,92 +19,91 @@ def foo(items: Items):
 client = TestClient(app)
 
 
-openapi_schema = {
-    "openapi": "3.0.2",
-    "info": {"title": "FastAPI", "version": "0.1.0"},
-    "paths": {
-        "/foo": {
-            "post": {
-                "responses": {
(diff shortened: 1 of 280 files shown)


GITHub8803:
lol

PR #9505, opened 2023-05-08
Traceback propagation to allow better logging- concurrency.py

zilinjak:
This way we keep the original traceback which can be queried later.

For example when ValidationError happens we can log traceback which is accurate in our exception handlerers.


arjwilliams:
I believe that this can be fixed by just passing the original traceback object into the `__exit__` call, rather than waiting for it to be unset then setting it back again:

```diff
- cm.__exit__, type(e), e, None, limiter=exit_limiter
+ cm.__exit__, type(e), e, e.__traceback__, limiter=exit_limiter
```

PR #9509, opened 2023-05-09
Fix duplicated section in menu if a translated file is not mentioned in mkdocs.yaml

flpm:
**Context**

In the Portuguese translation, there is a translated file (`tutorial/security/first-steps.md`) that is not in `mkdocs.yaml`.

When building the documentation, the file is still used but ends up with an English section name, which causes a duplicated section to appear in the docs menu.

image

**What was done**

- Added a check to `./scripts/docs.py lang-build` to confirm that translated files are explicitly mentioned in `mkdocs.yml` and if not treat the case as if the file did not exist, falling back to the English version.
- Included the missing translated file in the PT `mkdocs.yaml`

**How was it tested**

1) Built the docs using `scripts/docs.py build-lang` and verified that the existing translated (`first-steps.md`) that is not mentioned in mkdocs is ignored and replaced by the English version under the appropriate translated section names.

image

2) Added the missing reference to `tutorial/security/first-steps.md` in the PT `mkdocs.yam`l and rebuild the doc to verify that it works as before

image

### DIFF
--- a/docs/pt/mkdocs.yml
+++ b/docs/pt/mkdocs.yml
@@ -88,6 +88,7 @@ nav:
   - tutorial/encoder.md
   - Segurança:
     - tutorial/security/index.md
+    - tutorial/security/first-steps.md
   - tutorial/background-tasks.md
   - tutorial/static-files.md
   - Guia de Usuário Avançado:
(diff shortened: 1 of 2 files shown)


tiangolo:
Thank you @flpm! ☕ 

I recently restructured how all these docs and translations systems work and now the duplicated `mkdocs.yml` files are no longer necessary, the duplication and synchronization is not necessary either, and the order of the sections is always preserved! 🎉 

So I'll pass on this one, but thanks for the catch and for the effort! 🍰

PR #9517, opened 2023-05-10
🐛 Fix `int` values being serialized as `float` in the OpenAPI schema

alexturcea:
Example:

```
app = FastAPI()

class Foo(BaseModel):
    bar: int = Field(..., ge=0, le=100, multiple_of=3)
    baz: float = Field(..., gt=0.0, lt=100.0, multiple_of=3.14)

@app.get("/foo")
def foo() -> Foo:
    return Foo(bar=3, baz=3.14)
```

Schema with the current version:

```
{
...
    "components": {
        "schemas": {
            "Foo": {
                "title": "Foo",
                "required": [
                    "bar",
                    "baz"
                ],
                "type": "object",
                "properties": {
                    "bar": {
                        "title": "Bar",
                        "multipleOf": 3.0,   <-- HERE
                        "maximum": 100.0,    <-- HERE
                        "minimum": 0.0,      <-- HERE
                        "type": "integer"
                    },
                    "baz": {
                        "title": "Baz",
                        "multipleOf": 3.14,
                        "exclusiveMaximum": 100.0,
                        "exclusiveMinimum": 0.0,
                        "type": "number"
                    }
                }
            }
        }
    }
}
```

Schema with the proposed fix:

```
{
...
    "components": {
        "schemas": {
            "Foo": {
                "title": "Foo",
                "required": [
                    "bar",
                    "baz"
                ],
                "type": "object",
                "properties": {
                    "bar": {
                        "title": "Bar",
                        "multipleOf": 3,   <-- FIXED
                        "maximum": 100,    <-- FIXED
                        "minimum": 0,      <-- FIXED
                        "type": "integer"
                    },
                    "baz": {
                        "title": "Baz",
                        "multipleOf": 3.14,           <-- SAME AS BEFORE
                        "exclusiveMaximum": 100.0,    <-- SAME AS BEFORE
                        "exclusiveMinimum": 0.0,      <-- SAME AS BEFORE
                        "type": "number"
                    }
                }
            }
        }
    }
}
```


alexturcea:
Hi @tiangolo, please review this PR 😄

Ryandaydev (approved):
Tests pass.

alexturcea:
Thanks for taking a look @Ryandaydev - is there a process that needs to be followed to get this PR approved?

YuriiMotov (changes requested):
1) Tests are needed
2) These changes don't work with Pydantic 1.10.14 (`3.0` turns into `3`)

Test example:

**test_application.py**
```
def test_openapi_schema_num_constraints_int_float():
    response_json = response.json()
    assert isinstance(
        response_json["paths"]["/path/param-gt/{item_id}"]["get"]["parameters"][0][
            "schema"
        ]["exclusiveMinimum"],
        float,
    )

    assert isinstance(
        response_json["paths"]["/path/param-gt-int/{item_id}"]["get"]["parameters"][0][
            "schema"
        ]["exclusiveMinimum"],
        int,
    )
```
Test above works with Pydantic 2.6.1, but fails with Pydantic 1.10.14

YuriiMotov:
This solution now converts `0.0` to just `0`.

```python
from fastapi import FastAPI

app = FastAPI()


class Foo(BaseModel):
    zero_float: float = Field(ge=0.0)
    non_zero_float: float = Field(ge=0.0)

@app.get("/foo")
def foo() -> Foo:
    pass
```

Results in:

```json
      "Foo": {
        "properties": {
          "zero_float": {
            "type": "number",
            "minimum": 0,                       <-- SHOULD BE 0.0
            "title": "Zero Float"
          },
          "non_zero_float": {
            "type": "number",
            "minimum": 0.1,                       <-- this is Ok
            "title": "Non Zero Float"
          }
        },
```

Not sure this behavior is better than converting all values to `float`s.

tiangolo:
Interesting, is there a specific use case you would need this for? It would probably be better to start in a discussion with your specific use case and a self contained example.

In JSON, there's no notion or difference between int and float, so having a float would be in theory valid. Are you consuming OpenAPI from something that interprets JSON with some difference between ints and floats?

For now I'll close this one, but if you want to discuss it more, it would be better in a GitHub Discussion. :coffee:

PR #9534, opened 2023-05-14
Update README.md

manojduraisamy7991:
i added 1st step for fastapi install


manojduraisamy7991:
i added 1st step for fastapi install

manojduraisamy7991:
i added 1st step for fastapi install

tiangolo:
Thanks for the interest! Nevertheless, handling environments depends a lot on each project, some people could use Poetry, PDM, Docker, centralize the location of the environments, etc.

So I'll pass on this one, but thanks for the effort! ☕

PR #9552, opened 2023-05-19
alias import

adarshpunj:
minor change in import statement (removes alias import with same name)


tiangolo:
Hey there! As you can see from the tests, this was necessary. So I'll go ahead and close this one, but thanks for the interest! ☕

PR #9557, opened 2023-05-21
feat: supported globally setting response_model_by_alias

MarkShawn2020:
It enables user to set `response_model_by_alias` in `APIRouter`, and the inherience effect is tested as expected.

```python
root_router = APIRouter(prefix='/', response_model_by_alias=False)
```


MarkShawn2020:
Since I changed the param type, and added extra param for APIRouter, the doc should be updated as necessary.

samredway:
Would be great to get this in!

nidhishgajjar:

## Summary

This PR allows setting `response_model_by_alias` at the `APIRouter` level as a default, so individual routes inherit it without specifying it each time. The parameter default is changed from `True` to `None` across all route decorators, with `None` resolving to the router-level setting (or `True` as the final fallback).

## Architecture

The approach follows the existing pattern used by other router-level settings (e.g., `responses`, `dependencies`). A new `response_model_by_alias` attribute on `APIRouter` is checked as a fallback when the per-route value is `None`.

## Issues

### 1. No `FastAPI` app-level support (Warning)

**File:** `fastapi/applications.py`

The `APIRouter.__init__()` now accepts `response_model_by_alias`, but the `FastAPI` class (which wraps an internal `APIRouter`) does **not** expose this parameter in its `__init__()`. Users who want a global default must create a separate `APIRouter` and include it, rather than setting it on the `FastAPI()` constructor directly.

This limits the feature's usability. Consider adding `response_model_by_alias: Optional[bool] = None` to `FastAPI.__init__()` and passing it through to `self.router`.

### 2. No test coverage (Warning)

The PR adds no tests. At minimum, tests should verify:
- Setting `response_model_by_alias=False` on an `APIRouter` and confirming routes without explicit `response_model_by_alias` inherit the setting
- Setting it to `None` (default) still produces `True` behavior (backward compatibility)
- Per-route override still works when the router default is set

### 3. Missing `Doc()` annotation (Suggestion)

**File:** `fastapi/routing.py`, `APIRouter.__init__()`

The new `response_model_by_alias` parameter lacks a `Doc()` annotation, unlike other parameters in the same constructor. This means it won't appear in the generated API reference documentation.

### 4. Backward compatibility note (Observation)

The behavioral contract is preserved — existing code that doesn't set `response_model_by_alias` will still get `True` as the effective value. The `None → True` fallback in `get_request_handler()` ensures this. Good.

## Cross-file Impact

- **`fastapi/params.py`** / **`fastapi/dependencies`**: Not affected — this only touches routing and response serialization.
- **Docs**: No documentation updates are included. A "how-to" or tutorial mention would help users discover this feature.

## Assessment

**request-changes** — The feature is a good addition but needs test coverage and ideally `FastAPI`-level support to be truly useful. The missing tests make it hard to verify correctness, and without `FastAPI.__init__()` support, the "global" setting is only available for sub-routers.

tiangolo:

For the specific use case, the best would be to document it in a new GitHub Discussion with the specific problem to solve (without details of the implementation) to track the feature request.

Thank you! 🍰

PR #9581, opened 2023-05-28
✏ Rewording in `docs/en/docs/tutorial/debugging.md`

ivan-abc:
A suggestion to change some words in `__name__ == "__main__"` section of the `debugging.md` to simplify some expressions.

And also replaced the word "automatic" in order not to mix up with [automatic variable](https://en.wikipedia.org/wiki/Automatic_variable) definition 🤔 (as I understand it has another meaning in this context).


tiangolo:
Thanks!

I actually prefer the original words as the variables are not being set by the user, or by FastAPI or anything else, they just "have the value". Of course, this becomes a lot dependent on personal preference, but anyway, I prefer the previous wording.

I had never heard the term "automatic variable", I guess that comes from C++ and Java, maybe. But anyway, I updated it to not use that exact term.

Thanks! ☕

PR #9588, opened 2023-05-29
✏️ Fix tooltips for light/dark theme toggler in docs

pankaj1707k:
This is a minor fix 😅

When already in dark mode, hovering over the switcher shows "Switch to dark mode" in the tool-tip. Similarly, it shows "Switch to light mode" when its already in light mode.

I have fixed this by swapping the tool-tip texts in `docs/en/mkdocs.yml`.

PS: Please clarify if this needs to be done for every language separately or is it sourced from the *en* directory itself 😅

### DIFF
--- a/docs/en/mkdocs.yml
+++ b/docs/en/mkdocs.yml
@@ -11,14 +11,14 @@ theme:
     accent: amber
     toggle:
       icon: material/lightbulb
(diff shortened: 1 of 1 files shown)


tiangolo:
Yep, good catch, thanks! 🚀

PR #9594, opened 2023-05-31
Rework is_body_allowed_for_status_code function

pylounge:
Added an enumeration for OpenAPI status codes ranges.
Simplified the expression to check the condition.
Added exception handling if status code cannot be converted to int.


tiangolo:
Thanks for your interest! But I don't really think this is currently necessary, so I'll pass on this one. ☕ 

If you have a particular use case that needs this, please create a new GitHub Discussion question following the form, that will help me/us understand it better: https://github.com/tiangolo/fastapi/discussions/new?category=questions

PR #9620, opened 2023-06-05
📝 Update `Annotated` notes in `docs/en/docs/tutorial/schema-extra-example.md`

Alexandrhub:
When working on the translation, I noticed that this page is missing the annotated tips that can be found in the rest of the documentation (I checked, and it's the only page where they're missing).


ivan-abc (approved):

tiangolo:
Good catch, thank you @Alexandrhub! 🙇 ☕

PR #9630, opened 2023-06-06
🐛 Ensure that `app.include_router` merges nested lifespans

Lancetnik:
I am working on my own [Propan](https://github.com/Lancetnik/Propan) library and as a part of it I implemented a custom **FastAPI** router with the following behavior.

```python
from fastapi import FastAPI
from propan.fastapi import RabbitRouter

router = RabbitRouter("amqp://guest:guest@localhost:5672")
app = FastAPI()
app.include_router(router)
```

To connect RabbitMQ at **FastAPI** startup I used `router.on_startup` events and all worked fine, but now it is deprecated. So I migrate to `lifespan` and now **FastAPI** doesn't run router lifespan at the code above.

I suppose it was missed cuz the original `include_router` method includes all nested `on_startup` and `on_shutdown` deprecated events [here](https://github.com/tiangolo/fastapi/blob/master/fastapi/routing.py#L1307), but doesn't includes a router lifespan.

So, the following code doesn't work

```python

@asynccontextmanager
async def router_lifespan(app):
    print('router start')
    yield
    print('router shutdown')

router = APIRouter(lifespan=router_lifespan)
app = FastAPI()
app.include_router(router)
```

The current PR fixes the original framework miss.
All original test and lint.sh checks works fine.


tiangolo:
📝 Docs preview for commit 8f26ed5c37a03b41b1ecd445f0894cca9e54bf89 at: https://6484cca7bf187a105b276a5e--fastapi.netlify.app

Lancetnik:
@tiangolo can you take a look here, please? It's a minor bugfix, but it's important for my project.

tiangolo:
📝 Docs preview for commit f322c99f8ba981502511e58f815bd744c40d02ee at: https://648c2ec07c8ac9541405829c--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit 3e2fd98791e46c6dec0481e482b76e9e63ad0d2b at: https://648e16b538df29410a152328--fastapi.netlify.app

Kludex:
> @tiangolo can you take a look here, please? It's a minor bugfix, but it's important for my project.

This is not minor, and ideally we'd like this to live in Starlette.

Lancetnik:
@Kludex maybe, but how we can move it to Starlette if `include_router` is the **FastAPI**-level method? Do you suggest move the method to Starlette?

Lancetnik:
This will require a lot of work in both **FastAPI** and **Starlette**, since none of the **FastAPI** methods uses `super().add_route()` from  **Starlette**. Therefore, it is necessary to rewrite all the methods `add_api_route`, `add_websocket_route`, `add_api_websocket_route` to use the parent methods of Starlette. It will also require a consistent update of these two packages. It seems to me that this is too difficult way, but I will do it. Perhaps it makes sense to merge the  current PR for now before we can make a more elegant solution.

Lancetnik:
@Kludex, so can you check a **Starlette** [PR](https://github.com/encode/starlette/pull/2189)?

Lancetnik:
> This is not minor, and ideally we'd like this to live in Starlette.

@Kludex this bug doesn't on **Starlette** side and not connected with this [Issue](https://github.com/tiangolo/fastapi/issues/811) or this **Starlette** [PR](https://github.com/encode/starlette/pull/1988) or [Issue](https://github.com/encode/starlette/issues/649).

It should be fixed in **FastAPI** cuz we have no **Starlette** calls inside `include_router` method. It doesn't use **Starlette** `Host` or `Mount` that can be nested at **Starlette** side, but copies all included router attributes to `app.router` directly. So, we should merge the incuded lifespan here too.

So can you review and approve this PR?

Kludex:
Starlette should support running the lifespan of other routers... It shouldn't be here.

Lancetnik:
> Starlette should support running the lifespan of other routers... It shouldn't be here.

But there no *other routers* we have the original `app.router` and copy all routes to it. What do you want to do with it?

neeps:
Would love this to get merged, recently upgraded fastapi code that made use of @app.on_event("startup") inside a router. But the switch to lifespans have broken this functionality, when the router is included into a FastAPI() object.

Lancetnik:
@neeps yes, lifespan functional is a part of Starlite, but 'incude_router' is a FastAPI custom method and the bug is here.

neeps:
Thank you for clarifying, I've downgraded to pre-lifespan for the time being.

Lancetnik:
@Kludex sorry for being annoying. I know, that 

> Refactor routers, to re-use router instances insteaad of cloning path operations

Is in roadmap and It will fix the problem

But why we can't merge this PR for now? It just a few lines of code and it fixes the problem already without any waiting for abstract "roadmap". Working solution is better than perfect, but in the future, isn't it?

logankaser:
If we can't fix this, could, would you accept a PR to raise a `NotImplementedError`
when some poor fool like me tries to use the `lifespan` parameter on `APIRouter`
only to tear his application apart trying to figure out why it was not running?

Tishka17:
Any updates on this PR? Is it going to be merged?

@tiangolo

abdullahsych:
Any updates on this PR? @tiangolo

Praneethvvs:
> Thank you for clarifying, I've downgraded to pre-lifespan for the time being.

Hi, I'm trying to use lifespan with ApiRouter() insted of fastapi instance. It is still not working. Can you pin the latest stable  version where the startup/shutdown was functional.

logankaser:
> Hi, I'm trying to use lifespan with ApiRouter() instead of FastApi instance. It is still not working. Can you pin the latest stable version where the startup/shutdown was functional.
There is no functional version of this outside this PR right now.
You can go back to using the old event system until this works though.

Tishka17:
Can we talk to a maintainer, please?

Kludex:
> If we can't fix this now, would you accept a PR to raise a `NotImplementedError` when some poor fool like me tries to use the `lifespan` parameter on `APIRouter` only to tear his application apart trying to figure out why it was not running?

PR welcome on Starlette for this.

Kludex:
> Can we talk to a maintainer, please?

I said above that this idea should live in Starlette, not in FastAPI.

Lancetnik:
> I said above that this idea should live in Starlette, not in FastAPI.

For sure, but now `include_router` exists on **FastAPI** side only. If you want to move this logic to **Starlette**, you should remove/rewrite this method anyway. Why we can't  just patch the current implementation, if it will be refactored, when Starlette will be ready? People are waiting for fix a year already

Tishka17:
I see we stuck here. What about merging this PR,  releasing the version with bugfix and then starting a proper refactoring in case it is still necessary. 

If there is another fix for this bug I guess we can close this PR, but afaik there isn't.

coryphoenixxx:
+1

IvanKirpichnikov:
+1
still waiting 4 the fix

Maclovi:
Тианголо заеbал, merge this PR

KrySeyt:
+1

PlzTrustMe:
+1
Wait 4 fix

krau5:
+1

madnoberson:
+1

lubaskinc0de:
+1

PlzTrustMe:
+1

ovflw:
+1

Flosckow:
+1

FraterCRC:
+1)

ApostolFet:
+1

iNerV:
Wait for fix

bomzheg:
+1

jekel:
+1

Tishka17 (approved):

Tishka17:
@tiangolo can we have a minute of your time here?

Tishka17:
I am sorry to annoy you, but it is 14 months since PR is opened. Can we, as a community, have a decision or plan?

Lancetnik:
@svlandeg @estebanx64 @alejsdev Hello, team! Sorry for such annoyiment, but we are still waiting for any decision about this PR.

**FastAPI** now has an obvious bug /  weaknesses. And we have the PR fixing it.

I know, that this thing should be fixed in the future according to Roadmap, but we are waiting year already for it. Have we any reason to not merge 10 lines of code if they fixing the problem already and we can remove them in the future as a part of Roamap feature scope?

Dark04072006:
+1

Alkalit:
+1

Cub11k:
+1

chirizxc:
+1

prostomarkeloff:
+

dmitryleafall:
+1

fastapi/routing.py
```diff
@@ -119,6 +121,18 @@ def _prepare_response_content(
...
 
 
+def _merge_lifespan_context(
+    original_context: Lifespan[Any], nested_context: Lifespan[Any]
+) -> Lifespan[Any]:
+    @asynccontextmanager
+    async def merged_lifespan(app: AppType) -> AsyncIterator[Mapping[str, Any]]:
+        async with original_context(app) as maybe_self_context:
+            async with nested_context(app) as maybe_nested_context:
+                yield {**(maybe_self_context or {}), **(maybe_nested_context or {})}
```
Kludex, line 131: The _yielded_ value is actually called `maybe_state` on Starlette's code. For consistency, let's keep the same name.
 
```suggestion
        async with original_context(app) as maybe_original_state:
            async with nested_context(app) as maybe_nested_state:
                yield {**(maybe_original_state or {}), **(maybe_nested_state or {})}
```

Kludex (approved):

tiangolo:
Thank you for your contribution @Lancetnik! :cake: 

Thanks everyone for the input. :coffee: 

Thanks a lot for the help, as always, @Kludex :pray: :bow: 

This will be available in FastAPI version `0.112.2`, released in the next few hours. :tada: 

---

Thanks everyone for the patience. As this involved some internal changes, that means there's a bit of a higher chance it could break something else elsewhere (something we didn't expect and didn't account for), so I wanted to handle some other lower-impact issues and releases first.

Now I was finally able to sit a few hours dedicated just to this, read several parts of the ASGI spec again, ensure this would be the right way to handle things (e.g. what nested lifespan states should override which), etc.

Now, time to enjoy your new nested lifespans. :sunglasses: :coffee:

PR #9633, opened 2023-06-07
⬆️ Bump starlette to 0.28.0

Kludex:


pyproject.toml
```diff
@@ -41,7 +41,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
-    "starlette>=0.27.0,<0.28.0",
+    "starlette>=0.27.0,<0.29.0",
```
Kludex, line 44: We didn't add any breaking change, but change the range as you wish. 👀

Kludex: nevermind, it looks like starlette breaks fastapi test suite.

huaiguoguo:
when will "starlette" be updated to latest version 
latest version of starlette is  0.28.0

Kludex:
Subscribe to https://github.com/tiangolo/fastapi/pull/9636.

PR #9640, opened 2023-06-08
Add support for new Starlette request.form() parameters max_files

will-franceys:
Context:

- Starlette added `max_files` to `request.form()` parameters and a default limit to limiting max files. See the changes [here](https://github.com/encode/starlette/commit/8c74c2c8dba7030154f8af18e016136bea1938fa) due to [this](https://github.com/advisories/GHSA-74m5-2c7w-9w3x) advisory
- This means when sending multipart-form requests to FastAPI with more than 1000 files will break, see the reported issue [here](https://github.com/tiangolo/fastapi/discussions/9634)
- This PR adds an option to configure this setting through FastAPI

Changes:

- Added an option `max_files` to the route decorators to set this

```python
@app.post("/", max_files=2000)
def files(files: list[UploadFile]):
    return files
```

Tests:

- Added a test to check the default behaviour remains the same and also checks that the new setting works

Questions and TODOs:

- [ ] Is this the right approach? Is there a standard FastAPI pattern for setting this type of configuration?
- [ ] Update docs if this is the right approach
- [ ] Include max_fields option
- [ ] Make test assert the error detail as well


tests/test_startlette_request_form.py
```diff
@@ -0,0 +1,34 @@
...
+
+client = TestClient(app)
+
+
+def test_max_files_setting_default():
+    # Test Starlette default remains in place
+    files = [("files", ("foo.txt", "foo"))] * 1500
+    response = client.post("/route-with-default-limit", files=files)
+    # Fails with {'detail': 'Too many files. Maximum number of files is 1000.'}
+    assert response.status_code == 400
```
goranmrd, line 27: Why don't you assert against error details ? Like this is too broad.

adriangb:
How about using `annotated-types`' `MaxLen`: https://github.com/annotated-types/annotated-types/blob/a31912d536cd2a4e4e05d243bdbc941515633293/annotated_types/__init__.py#L243-L250.

Usage would be `async def endoint(files: Annotated[list[UploadFile], MaxLen(10)])` which is nicely analogous to `async def endpoint(users: Annotated[list[User], MaxLen(10)])`

tests/test_startlette_request_form.py
wfranceys: Sound sensible will include! It is annoying to couple against the text in starlette

wfranceys:
> How about using `annotated-types`' `MaxLen`: https://github.com/annotated-types/annotated-types/#minlen-maxlen-len.
> 
> Usage would be `async def endoint(files: Annotated[list[UploadFile], MaxLen(10)])` which is nicely analogous to `async def endpoint(users: Annotated[list[User], MaxLen(10)])`

@adriangb Interesting! Will take a look at that approach 

From what I can see I believe this would be the first usage of the [annotated-types](https://github.com/annotated-types/annotated-types) library in FastAPI - I would feel hesitant to put forward such a dependency without it being the recommended approach

adriangb:
It’s going to be a dependency anyway via Pydantic V2

wfranceys:
> It’s going to be a dependency anyway via Pydantic V2

@adriangb Ah great, that's good to know thanks! 

Do you know at all if FastAPI will start using them as a common pattern?

adriangb:
Yes it will be, Sebastián is working on but not sure when he’ll be merging the PR with Pydantic V2, but I’m guessing before this PR

tiangolo:
📝 Docs preview for commit 6212e1f3b3923bd608188caa33208f3277ac5197 at: https://6486235a6038ce493d4f630f--fastapi.netlify.app

gungjodi:
hi @tiangolo @adriangb any updates on this?

cardoso-neto:
Having the ability to do this would be useful for many usecases. Looking forward for updates.

Themis3000:
For my specific use case this is pretty important to me. Anyone aware of any workarounds? Any updates on this?

aaoobd3:
any updates on this

gungjodi:
Hi,
I found a workaround on this issue. As pointed on this [StackOverflow thread ](https://stackoverflow.com/questions/76970173/how-to-get-files-and-form-data-using-the-request-object-in-fastapi)

Basically, you need to set the request.form on Starlette's Request object as below : 

Set max files / max fields
```
async def get_body(request: Request):
    try:
        ## adjust max_fields & max_files based on your needs
        return await request.form(max_fields=1000000, max_files=100000)
    except Exception as exc:
        raise HTTPException(status_code=400, detail='Invalid Form data') from exc
```

then on your route, use above function as Depend on body
```
@app.post("/send-files")
async def send_results_endpoint(body = Depends(get_body)):
    ## get your files using getlist('<files field name>')
    if isinstance(body, FormData):
        files = body.getlist('files')

    ## 'files' variable will be instance of List[UploadFile]
    ....process your files here
    
    return {
        'totalFiles': len(files)
    }
```

Hope this helps

Yrij-Zhavoronkov:
What's the outcome? It's already 2025, but it's still not possible to conveniently set the maximum number of files?

Yrij-Zhavoronkov:
> Hi, I found a workaround on this issue. As pointed on this [StackOverflow thread ](https://stackoverflow.com/questions/76970173/how-to-get-files-and-form-data-using-the-request-object-in-fastapi)
> 
> Basically, you need to set the max_fields & max_files in request.form on Starlette's Request object as below :
> 
> Set max files / max fields
> 
> ```
> async def get_body(request: Request):
>     try:
>         ## adjust max_fields & max_files based on your needs
>         return await request.form(max_fields=1000000, max_files=100000)
>     except Exception as exc:
>         raise HTTPException(status_code=400, detail='Invalid Form data') from exc
> ```
> 
> then on your route, use above function as Depend on body
> 
> ```
> @app.post("/send-files")
> async def send_files(body = Depends(get_body)):
>     ## get your files using getlist('<files field name>')
>     if isinstance(body, FormData):
>         files = body.getlist('files')
> 
>     ## 'files' variable will be instance of List[UploadFile]
>     ....process your files here
>     
>     return {
>         'totalFiles': len(files)
>     }
> ```
> 
> image
> 
> Hope this helps

Your solution looks to work, but it’s not clear how to combine it with this:
```
@app.post("/images")
async def upload_images(
   field_0: Annotated[str, Body(...)],
   ...
   field_k: Annotated[str, Body(...)],
   images: Annotated[List[UploadFile], File()]):
```

without losing the default behavior (pydantic type validations, auto documentation, etc..). So far the only option that comes to mind is: parse the form, create a pydantic model, perform validation, pass the fields directly to the method `upload_images`.

PR #9670, opened 2023-06-13
use annotated description as Field description

vincentsarago:
ref: https://github.com/tiangolo/fastapi/discussions/9278

## History
Since FastAPI 0.95 we cannot have multiple endpoint routes with optional `Path` parameter (because we cannot mix the Query/Path for a variable:  https://github.com/tiangolo/fastapi/commit/375513f11494bc3499050ad2a0d378fb6e37ca98#r105161665), while it seems this `breaking` change is assumed, I personally still need this feature. 

To solve this, I have to remove the usage of `Query/Path` class and use simple `Annotated` form, but this means that I loose the nice `OpenAPI` kwargs provided by FieldInfo.

## Proposed change

With this PR I'm proposing that we use the first metadata annotation as Field's description 
```python
from typing import Annotated
from fastapi import FastAPI

app = FastAPI()

@app.get("/test")
async def variable(foo: Annotated[str,  "get a description"] = "bar"):
   return {"foo": foo}
```
in ☝️ `"get a description"` will be forwarded to the OpenAPI schema as field's description 👇 

Screenshot 2023-06-13 at 6 17 58 PM

I originally designed a more complex solution over https://github.com/tiangolo/fastapi/pull/9289, but I think what I'm proposing with this PR is much more simple (thus acceptable ? 😅).

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -367,6 +367,7 @@ def analyze_param(
     used_default_field_info = False
     depends = None
     type_annotation: Any = Any
+    description: Optional[str] = None
     if (
         annotation is not inspect.Signature.empty
         and get_origin(annotation) is Annotated  # type: ignore[comparison-overlap]
@@ -376,7 +377,7 @@ def analyze_param(
         fastapi_annotations = [
             arg
             for arg in annotated_args[1:]
-            if isinstance(arg, (FieldInfo, params.Depends))
+            if isinstance(arg, (FieldInfo, params.Depends, str))
         ]
         assert (
             len(fastapi_annotations) <= 1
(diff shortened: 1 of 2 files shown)


tiangolo:
📝 Docs preview for commit 60d163dce555d898c526503914a56c12508195f4 at: https://648898509e9ccd2141608062--fastapi.netlify.app

vincentsarago:
Is there anything I need to do here? 🙏

tiangolo:
This could cause conflicts with a bunch of other behaviors. I'll need to handle this later in a different way. For now, I'll close this one. ☕

tiangolo:
This could cause conflicts with a bunch of other behaviors. I'll need to handle this later in a different way. For now, I'll close this one. ☕

PR #9722, opened 2023-06-22
🔧 Run CI only once on push

mateusoliveira43:
## Description

I noticed the project CI runs "twice" in each commit to the repository (one for the real commit and one for the `latest-changes` job).
Ex.: https://github.com/tiangolo/fastapi/actions/runs/5348732342 and https://github.com/tiangolo/fastapi/actions/runs/5348736904

This PR assures it only runs after `latest-changes` job.

## What has been done to achieve the goal

Run `grep -Iinr 'push' .github/workflows` in the project root to check that all jobs that are run `on push` were updated.

## Testing Instructions

I think there is not a easy to test out github actions (what I usually do before merging them, if I want to test prior to merging, is running it from my fork).

But I do this in my personal projects, so I think it safe :smile: Ex.: If you check the [commits in this repo](https://github.com/mateusoliveira43/cly/commits/main), the CI only run after the `latest-changes` job.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - master
+    paths:
+      - 'docs/en/docs/release-notes.md'
   pull_request:
     types:
       - opened
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    paths:
+      - 'docs/en/docs/release-notes.md'
   pull_request:
     types:
       - opened


tiangolo:
📝 Docs preview for commit 5956d1a203f986fcb3cdbd760c00b7487f78f347 at: https://6494a91e47ca610b1aab5880--fastapi.netlify.app

yezz123 (approved):

codespearhead (approved):

tiangolo:
Thanks! But this would mean that tests on push would strongly depend on the GitHub Action `latest-changes`. If for some reason that action is not triggered (e.g. GitHub Actions glitch, expired token, etc.) then these tests wouldn't run. So I prefer to keep it as it is. But thanks for the interest! :coffee:

PR #9741, opened 2023-06-25
Adding the option to add an external websocket route class

roysagi:
This will enable users to use their own websocket route class with their dependencies and restrictions without overriding the `add_api_websocket_route` function


Kludex:
Can you show me a use-case this would be useful for?

roysagi:
@Kludex - Can't show you but can explain.
In my company, we have an internal mechanism to check permissions for API/WebSocket calls done by external users and our FE. This mechanism is done by checking a specific kind of dependency - we wanted to make sure each of our routes is covered with a permission check and will not let a route be set without it - thus we needed to implement our own WebSocket Route class.

We feel this use case can be expanded, letting users control what checks are done on the route and maybe even how the WebSocket connection is handled by letting them implement their own Route classes with minimal change to the existing code - which led to this PR.

tiangolo:
Thanks for the interest @roysagi!

I'm not sure I understand the use case well. :thinking: 

Let's start in a GitHub Discussion, filling the form with all the data, and including a minimal self contained example that we can copy-paste and run and that shows what you intend to achieve but you can't yet. Maybe there's another way to achieve it.

Then, after that, we can consider a PR for the use case, with some tests that cover the use case and don't pass on `master`, but pass with the fix.

For now I'll close this one, let's start in a Discussion. :coffee:

PR #9753, opened 2023-06-27
✨ Allow using dependables with `functools.partial()`

### LINKED ISSUE #10720: functools.partial() does not work on an async dependable
### Discussed in https://github.com/tiangolo/fastapi/discussions/9744

<sup>Originally posted by **lieryan** June 26, 2023</sup>
### First Check

### Commit to Help

### Example Code

```python
from functools import partial

app = FastAPI()

def the_answer():
    return 42


async def dep1(
    n: int = Depends(the_answer),
):
    return n + 1

@app.get("/")
def main(
    d1=Depends(partial(dep1, n=10)),
):
    print(d1)   # prints "<coroutine object dep1 at 0x7facc996fb50>"
    return {}


if __name__ == '__main__':
    import uvicorn
    uvicorn.run(app, port=1122)
```

### Description

I would've expected `d1` above to be the return value of the dependable, but it seems like fastapi didn't identify the partial as an coroutine function and instead passed the unawaited coroutine into the view function.

`inspect.iscoroutinefunction()` and `inspect.signature()` were able to deconstruct the partial object to correctly identify whether a partial contains a coroutine or a regular function, so it seems like it should've been possible for FastAPI to use handle partial correctly:

```
In [5]: def regular(a, b): pass

In [6]: async def coro(a, b): pass

In [7]: inspect.iscoroutinefunction(partial(regular, 1))
Out[7]: False

In [8]: inspect.iscoroutinefunction(partial(coro, 1))
Out[8]: True

In [20]: inspect.signature(partial(coro, 1))
Out[20]: <Signature (b)>
```
 

### Operating System

Linux

### Operating System Details

_No response_

### FastAPI Version

0.98.0

### Python Version

Python 3.10.6

### Additional Context

_No response_</div>

lieryan:
Discussion: https://github.com/tiangolo/fastapi/discussions/9744

Using `functools.partial()` with dependables feels like a very obvious idiom. `partial()` already works for dependables that are plain function, but it doesn't work consistently, partial of async and callable classes didn't work consistently.

This PR makes partial() work consistently in all combinations.

test_dependency_class.py have been renamed to test_dependency_types.py to reflect its bigger scope now.

There's quite a bit of repetition in the tests, part of me wants to just refactor those; but another part of me also wants to keep the tests simple.


Lemminkyinen:
This would be nice to have

jtfidje:
Yeah - I also just hit this! Please merge 😅

jtfidje:
@tiangolo Are we any closer to a merge on this? 🙂

pachewise:
@lieryan could you resolve conflicts for this to be ready to merge?

YuriiMotov:
@lieryan, thanks for contributing to FastAPI!

Are you still interested in getting this PR merged?
I think it would be better to move new tests to separate file (e.g. `test_dependencies_partial.py`) to make it easier to review.
~~Also, I would add second parameters to all initial functions, so that partial would reduce the number of parameters from 2 to 1.~~
Would you like to make these changes? Or, don't you mind if I make these changes?

fastapi/dependencies/utils.py
```diff
@@ -528,24 +529,28 @@ def add_param_to_fields(*, field: ModelField, dependant: Dependant) -> None:
...
 
 def is_coroutine_callable(call: Callable[..., Any]) -> bool:
-    if inspect.isroutine(call):
-        return inspect.iscoroutinefunction(call)
-    if inspect.isclass(call):
-        return False
+    if inspect.iscoroutinefunction(call):
+        return True
+    if isinstance(call, partial):
+        return is_coroutine_callable(call.func)
```
```suggestion
    if inspect.isclass(call):
        return False
    if isinstance(call, partial):
        return is_coroutine_callable(call.func)
```
Why do we need to remove this condition (`isclass`)?
I just tried running tests with these lines and it still works.
I think if it's not necessary for making partial work, we shouldn't remove this lines in this PR

YuriiMotov (changes requested):
There is a problem when we use named parameters to `prtial`. **UPD:** resolved [here](https://github.com/fastapi/fastapi/pull/9753#issuecomment-3116660599)

Currently it produces wrong schema:
```python
from functools import partial

app = FastAPI()

async def dep1(n: int):
    return n + 1

@app.get("/")
def main(d1=Depends(partial(dep1, n=10))):  # But `partial(dep1, 10)` works fine
    return d1
```

image

`n` Query parameter still appears in schema and even works (you can pass value and it will be used as input parameter).
I expected it shouldn't appear in the schema and wouldn't be used as input

YuriiMotov:
@lieryan, thanks for your work on this!

Do you have any ideas how to fix the problem I mentioned above?

ruitcatarino:
> There is a problem when we use named parameters to `prtial`.
> 
> 
> 
> Currently it produces wrong schema:
> 
> ```python
> 
> from functools import partial
> 
> 
> 
> from fastapi import Depends, FastAPI
> 
> 
> 
> app = FastAPI()
> 
> 
> 
> async def dep1(n: int):
> 
>     return n + 1
> 
> 
> 
> @app.get("/")
> 
> def main(d1=Depends(partial(dep1, n=10))):  # But `partial(dep1, 10)` works fine
> 
>     return d1
> 
> ```
> 
> 
> 
> image
> 
> 
> 
> `n` Query parameter still appears in schema and even works (you can pass value and it will be used as input parameter).
> 
> I expected it shouldn't appear in the schema and wouldn't be used as input


Hi, today I ran into the same issue that resulted in this pull request, not being able to use `functools.partial` as a dependency. I eventually found this pull request and this issue. I did a little digging and the conclusion that I reached is that this has to do with how `functools.partial` works and positional vs keyword arguments.

```python
import inspect

from functools import partial

def add_one(num: int) -> int:
    return num + 1

two = partial(add_one, 1)
two_kwarg = partial(add_one, num=1)


inspect.signature(add_one) # <Signature (num: int) -> int>
inspect.signature(add_one).parameters.values() # odict_values([<Parameter "num: int">])

inspect.signature(two) # <Signature () -> int>
inspect.signature(two).parameters.values() # odict_values([])

inspect.signature(two_kwarg) # <Signature (*, num: int = 1) -> int>
inspect.signature(two_kwarg).parameters.values() # odict_values([<Parameter "num: int = 1">])
```

For this reason when `get_typed_signature` (from fastapi code) is called we will get diferente results depending on how we used `functools.partial`.

> `n` Query parameter still appears in schema and even works (you can pass value and it will be used as input parameter).

As for this, this works when using keyword arguments with `functools.partial` because:

> Return a new partial object which when called will behave like func called with the positional arguments args and keyword arguments keywords. If more arguments are supplied to the call, they are appended to args. If additional keyword arguments are supplied, they extend and override keywords.

https://docs.python.org/3/library/functools.html#functools.partial

Thinking on the example above calling `two_kwarg(num=3)` would still return `4` since the keyword argument `num` would get overridden.

Hope this helps out

YuriiMotov:
> As for this, this works when using keyword arguments with `functools.partial` because:
> 
> > Return a new partial object which when called will behave like func called with the positional arguments args and keyword arguments keywords. If more arguments are supplied to the call, they are appended to args. If additional keyword arguments are supplied, they extend and override keywords.
> 
> https://docs.python.org/3/library/functools.html#functools.partial

@ruitcatarino, thanks for explaining this!

So, then we can (we should) just mention this pitfall somewhere in docs.
We can probably skip this for now and add a section in docs named "What you can use inside `Depends`" later on

YuriiMotov (approved):
LGTM in general

There are 2 things to think about:
* I don't understand why the `if inspect.isclass(call):` condition was removed [here](https://github.com/fastapi/fastapi/pull/9753/files#r2192922090). Although, tests pass regardless of this condition.
* There is an [issue](https://github.com/fastapi/fastapi/pull/9753#pullrequestreview-2998359406) when `partial` is used with named parameters, but this is just how `partial` works. I think we can later explain this in docs

@lieryan, @ruitcatarino, thanks a lot!

ruitcatarino:
You're welcome, glad it helped @YuriiMotov!

I would also suggest adding some tests using keyword arguments and partial since they have different behaviours.

Even maybe validate that if you're using partial and positional sending the query param won't affect the output but using keyword arguments and partial will.

If you do find this valuable and no one else wants to do it I would be glad to help out (I could also add this "issue" to the docs if needed).

YuriiMotov:
@ruitcatarino, I personally don't think we need more tests here.
Let's wait for Sebastian to review this and if this is accepted, we will add docs (probably as a separate PR).

fastapi/dependencies/models.py
```diff
@@ -53,25 +53,34 @@ def cache_key(self) -> DependencyCacheKey:
...
-        if inspect.isroutine(self.call):
-            return iscoroutinefunction(self.call)
-        if inspect.isclass(self.call):
+        use_call: Any = self.call
+        if isinstance(self.call, partial):
+            use_call = self.call.func
+        if inspect.isroutine(use_call):
+            return iscoroutinefunction(use_call)
+        if inspect.isclass(use_call):
             return False
```
YuriiMotov: In the initial version of this PR it was suggested to change these lines to something like:
```diff
-        if inspect.isroutine(use_call):
-            return inspect.iscoroutinefunction(use_call)
+        if inspect.iscoroutinefunction(use_call):
+            return True
-        if inspect.isclass(use_call):
-            return False
```

See original commit: https://github.com/fastapi/fastapi/pull/9753/commits/f16dd38b478c0a88ae4ae79fd73094fdb3ffa425

I updated it to be closer to current implementation (don't simplify first condition and don't remove `if inspect.isclass(use_call):`)

tiangolo (approved):
Thank you @lieryan! 🚀 

Sorry for the delay getting back to this.

Thanks all for the discussion, thanks @YuriiMotov for the help. ☕ 

This will be available in FastAPI 0.123.5, released in the next few hours. 🎉

wu-clan:
> > As for this, this works when using keyword arguments with `functools.partial` because:
> > > Return a new partial object which when called will behave like func called with the positional arguments args and keyword arguments keywords. If more arguments are supplied to the call, they are appended to args. If additional keyword arguments are supplied, they extend and override keywords.
> > 
> > 
> > https://docs.python.org/3/library/functools.html#functools.partial
> 
> @ruitcatarino, thanks for explaining this!
> 
> So, then we can (we should) just mention this pitfall somewhere in docs. We can probably skip this for now and add a section in docs named "What you can use inside `Depends`" later on

I used this feature in the new feature today and discovered this issue. Can we find a way to exclude it from the schema? I think it's a distraction for the schema because Swagger display them.

PR #9783, opened 2023-07-02
🐛 Fix `OAuth2PasswordRequestForm` and `OAuth2PasswordRequestFormStrict` fixed `grant_type` "password" RegEx

skarfie123:
When using the OAuth2 password request form, the regex for the `grant_type` is not strict enough. It currently allows any string that contains the word "password" in it.

The docstring says:
```
grant_type: the OAuth2 spec says it is required and MUST be the fixed string "password"
```

This PR changes the regex to ensure the exact string.

I also added extra test cases to cover this.

Minimal example:
```python
from typing import Annotated
from fastapi.security import OAuth2PasswordRequestForm

app = FastAPI()

@app.post("/login")
async def login(form_data: Annotated[OAuth2PasswordRequestForm, Depends()]):
    pass
```

Using the Swagger UI docs to test, with `grant_type=passwordblah` it returns 200 before but 422 after the change, as expected.


iudeen:
I wonder if we should let the user decide the pattern, rather than forcing a stricter requirement by default.

skarfie123:
My understanding is that this class is specifically for the OAuth2 password flow
https://www.oauth.com/oauth2-servers/access-tokens/password-grant/
That specifies that the `grant_type` must be `password`

If the user wants a different flow they can implement it themselves, or a more general class should be provided.
In the current state I think it is misleading.

YuriiMotov (approved):
Reviewed this PR. Everything is good!

1. Tests fail before changes (returns response=`200` instead of `422` when passing `grant_type` 'prefix password' or 'password suffix').
2. Tests are Ok after applying changes.
3. Documentation is Ok (there is no need to change anything)

skarfie123:
I've resolved the merge conflicts.

alejsdev:
Hi @skarfie123, thanks for your interest in contributing to FastAPI. We appreciate your effort and the time you have taken to submit this PR. We have a high volume of PRs and we're working hard to review and classify them.  We appreciate your patience as we managed the queue, we'll review your PR soon. Thanks! 🙇‍♀️

skarfie123:
No worries @alejsdev ! I've resolved the merge conflicts again.

alejsdev:
I already checked it and it's ok, from my point of view it's ready to be merged, I'll leave the final decision to @tiangolo :rocket: 
I appreciate your patience and help! Thanks @skarfie123 :bowing_woman:

svlandeg:
I merged in `master` and resolved the conflicts - thanks to the recently simplied test suite, there are much less test files that need correction now 😎

tiangolo (approved):
Nice, thank you! :rocket: :bug: 

And thanks for the reviews and help everyone! :coffee:

PR #9786, opened 2023-07-03
♻️ Deprecate parameter `regex`, use `pattern` instead

tiangolo:
♻️ Deprecate parameter `regex`, use `pattern` instead


tiangolo:
📝 Docs preview for commit b070fe449e658677de66b16668b77865d35bf47f at: https://64a321276c68bc2805e49574--fastapi.netlify.app

Issue #1543, opened 2020-06-09
Allow free-form queries within FastAPI


yeralin:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Description
There is a notion of "free-form query parameters" that allows a user to accept multiple non-predefined query params:
![image](https://user-images.githubusercontent.com/8620461/84204081-9329a980-aa78-11ea-9618-18d4f9dd5387.png)

This is in fact can be described by Open-API 3.0:
https://stackoverflow.com/questions/49582559/how-to-document-dynamic-query-parameter-names-in-openapi-swagger

Unfortunately, as of now you cannot define "free-form query parameters" within FastAPI.

### Example

This is how it would look like:

```Python
from fastapi import FastAPI

app = FastAPI()


@app.get("/")
def read_root(arguments: Dict[str, str] = Query({})):
    ...
```

But unfortunately, this raises an exception:
```
AssertionError: Param: arguments can only be a request body, using Body(...)
```

While at it, I hacked FastAPI a bit, and managed to make it render docs page with the valid entries:
![image](https://user-images.githubusercontent.com/8620461/84204398-1fd46780-aa79-11ea-9e27-1cffc0f86206.png)
By modifying: `sequence_shapes` set (I understand that it is not how it should be done, I just wanted to verify that it would generate a valid `openapi.json`)
```
from fastapi.dependencies.utils import sequence_shapes
from pydantic.fields import SHAPE_MAPPING

sequence_shapes.add(SHAPE_MAPPING)
```

### Describe alternatives you've considered

[...]

PR #9837, opened 2023-07-08
✨ Support Free From query parameters

mikeedjones:
Originally a fork of https://github.com/tiangolo/fastapi/pull/1885

This pull request adds support for ["free form" query parameters](https://swagger.io/specification/#:~:text=A%20free%2Dform%20query%20parameter%2C%20allowing%20undefined%20parameters%20of%20a%20specific%20type%3A) in FastAPI, allowing endpoint functions to receive arbitrary query parameters as mappings (dictionaries), addressing #1543.

- Added support for endpoints to accept arbitrary query parameters as `Dict[str, str]`, `Dict[str, int]`, or `Dict[str, List[int]]` using the `Query` dependency, enabling endpoints to flexibly capture unknown or dynamic query parameters. 
- Introduced new utility functions (`is_scalar_mapping_field`, `is_scalar_sequence_mapping_field`) in `fastapi/_compat` and its submodules to detect mapping types in both Pydantic v1 and v2, and integrated these into the dependency analysis and value extraction logic.
- Updated `fastapi/dependencies/utils.py` to extract mapping values from query parameters, handle validation errors by removing invalid keys, and ensure that keys captured by mapping fields do not conflict with explicitly defined parameters. 
- Added a new section "Free Form Query Parameters" to the tutorial, with code examples and a screenshot, explaining how to use the new feature.
- Added new endpoint and test cases in `tests/main.py` to cover various combinations of mapping and sequence mapping query parameters. 

This PR adds quite a few operations onto the critical path, but does add support for a useful feature of restful interfaces.


tiangolo:
📝 Docs preview for commit 49725969a32232cfc9b2f71cd908663e3b76c9e6 at: https://64a941c351861f6ed23a20c7--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit 58ed8150e1198f59a87f54d3889ef3c5486ca19a at: https://64a94fa69090837342ade1e4--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit 9beb1a75f7812dbd7ccd6f605cf6cd0024b949fe at: https://64a951486ffd10777eee213e--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit a42d91288092ceb37748e531390a55bc948649c5 at: https://64a952fdbcf4a17648b0dae6--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit ef7ca71c23b0e6318539f7fbaecaae8cb1354467 at: https://64aa953d7ed34f19262eb319--fastapi.netlify.app

docs/en/docs/tutorial/query-params.md
```diff
@@ -223,3 +223,45 @@ In this case, there are 3 query parameters:
 
 !!! tip
     You could also use `Enum`s the same way as with [Path Parameters](path-params.md#predefined-values){.internal-link target=_blank}.
+
+## Free Form Query Parameters
+
+Sometimes you want to receive some query parameters, but you don't know in advance what they are called. **FastAPI** supports provides support for this use case.
```
Viicos, line 229:
```suggestion
Sometimes you want to receive some query parameters, but you don't know in advance what they are called. **FastAPI** provides support for this use case.
```

tiangolo:
📝 Docs preview for commit 12c98cd4068f6e33c0bc97c82621527591ca14c2 at: https://64acdedae293c434bbe27227--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit 25a269832cc5e511b236f6eb80cfb4ae56875554 at: https://64ad02052acb1a493f461359--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit cec06df8643a4b93de6c3e5fcbb549bf747dda23 at: https://f9486749.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit 182638ba70a9b7211615418d15b0e68b1258c156 at: https://48491e52.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit a5a9e6d52116ed5bb15c4967edffb4d5b1a12018 at: https://fe3406fc.fastapitiangolo.pages.dev

tests/test_query.py
```diff
@@ -408,3 +408,36 @@ def test_query_frozenset_query_1_query_1_query_2():
...
+    assert response.status_code == 200
+    assert response.json() == "foo bar fuzz baz baz buzz fizz"
+
+
+def test_mapping_with_non_mapping_query_mixed_types():
+    response = client.get("/query/mixed-type-params/?foo=fuzz&foo=baz&bar=buzz&query=1")
+    assert response.status_code == 200
+    assert (
+        response.json()
+        == "foo bar 1 {'foo': 'baz', 'bar': 'buzz', 'query': '1'}  {'query': 1} {'foo': [], 'bar': [], 'query': [1]}"
```
YuriiMotov: I would expect explicitly specified `query` parameter not be included in `sequence_mapping_queries`, `mapping_query_str` and `mapping_query_int`.
But maybe this is not a problem if documented

tests/main.py
```diff
@@ -184,6 +184,38 @@ def get_query_param_required_type(query: int = Query()):
     return f"foo bar {query}"
 
 
+@app.get("/query/mapping-params")
+def get_mapping_query_params(queries: Dict[str, str] = Query({})):
+    return f"foo bar {queries['foo']} {queries['bar']}"
```
YuriiMotov:
```suggestion
    return {
        "queries": queries,
    }
```

And then test it in `` as:
```py
assert response.json() == {"foo": "fuzz", "bar":  "buzz"}
```

And the same for all other enptoints

YuriiMotov (commented):
@mikeedjones, thank you for working on this!

I think this PR is interesting.
Are you still interested in working on it? Will you have time to resolve conflicts?

mikeedjones:
@YuriiMotov yes I can do. I will find some time today to resolve the conflicts and the review. Thank you!

tests/test_query.py
mikeedjones: added logic in to cover this and removed the section from the docs.

tests/main.py
mikeedjones: updated them

mikeedjones:
@YuriiMotov that was a little more involved then I wanted it to be, but with pydantic >2.6 there's not much added on the critical path and the validation is done neatly by pydantic - for <2.6 and 1.10 its not great because the `OnErrorOmit` feature had not yet been added to pydantic. 

I think the currently implemented behaviour would be expected by users - but changing the annotation is an anti-pattern in the libray, AFAICT. We could also require that users set the `OnErrorOmit` themselves and call it out as a pydanticV2.6+ feature in the docs?

Maybe worth waiting to after fastapi deprecates pydantic-v1 support? Or adding some warnings to the docs. What do you think?

fastapi/dependencies/utils.py
```diff
@@ -819,6 +850,14 @@ def request_params_to_args(
...
         else:
             values[field.name] = v_
+    # remove keys which were captured by a mapping query field but were
+    # specified as individual fields
+    for field in fields:
+        if isinstance(values.get(field.name), dict) and (
+            is_scalar_mapping_field(field) or is_scalar_sequence_mapping_field(field)
+        ):
+            for f_ in fields:
+                values[field.name].pop(f_.alias, None)
```
mikeedjones, line 904: This could also be moved to inside `_get_multidict_value` so the field types are only checked once if we don't mind passing `fields` into that function

YuriiMotov:
@mikeedjones, thank you! I see that was a lot of work..

> Maybe worth waiting to after fastapi deprecates pydantic-v1 support?

I haven't reviewed the changes in-depth yet, but I trust your opinion.
There are also other reasons for this - we'd like to focus on important bug fixes first.
So, let's put this on pause until we have a clear understanding of when FastAPI will drop the support of Pydantic V1.

Thanks again!

mikeedjones:
> @mikeedjones, thank you! I see that was a lot of work..
> 
> > Maybe worth waiting to after fastapi deprecates pydantic-v1 support?
> 
> I haven't reviewed the changes in-depth yet, but I trust your opinion. There are also other reasons for this - we'd like to focus on important bug fixes first. So, let's put this on pause until we have a clear understanding of when FastAPI will drop the support of Pydantic V1.
> 
> Thanks again!

@YuriiMotov Took another look at this now that pydantic v1 is no longer supported and added some more tests. Thank you!

mikeedjones:
@YuriiMotov Is there anything else this needs do you think? Appreciate OSS maintainers are under a lot of load at the moment - thank you for the hard work!

YuriiMotov:
> @YuriiMotov Is there anything else this needs do you think? Appreciate OSS maintainers are under a lot of load at the moment - thank you for the hard work!

Sorry for the long time not replying..
I think for now we should wait for Sebastian to take a look and decide whether it's right time to add this support and if the approach is correct.
I would recommend we just wait. I wouldn't put more efforts into it until he confirms we are going to implement this support.
Thank you!

tiangolo:

I want to implement a few big refactors and features that might solve many of these use cases.

For the specific use case, the best would be to document it in a new GitHub Discussion (or :+1: any existing one) with the specific problem to solve (without details of the implementation) to track the feature request.

Thank you! 🍰

PR #9852, opened 2023-07-10
🐛 Replace `MultHostUrl` to `AnyUrl` for compatibility with older versions of Pydantic v1

Kludex:


tiangolo:
Awesome, thank you @Kludex! 🙇 🚀

PR #9770, opened 2023-06-30
✨ Add support for OpenAPI 3.1.0

tiangolo:
✨ Add support for OpenAPI 3.1.0

PR #9866, opened 2023-07-12
Multiple Body Examples regression in v0.100

austinorr:
OpenAPI 3.1 ✨now strictly implements the 2020-12 JSON Schema for their [Schema Object](https://spec.openapis.org/oas/v3.1.0#schema-object) which defines the 'examples' keyword as type `List[Any]`. This is great, but when we define multiple examples meant to be included in the Swagger UI docs for routes which accept a `Body` we often want to use the nice dropdown picker, like in this screengrab from the docs, which in v0.100 with examples passed strictly as lists, this UI is no longer achievable (#9770, working in #1267):

image

This functionality was achieved in `fastapi>=0.64, <0.100` by adding the examples to the requestBody that conforms to the 
[Media Type Object](https://spec.openapis.org/oas/v3.1.0#media-type-object) specification, which takes a _map_ of [example objects](https://spec.openapis.org/oas/v3.1.0#example-object) rather than a list of schemas.

The legacy `fastapi` implementation grabbed the examples passed to the `Body` kwarg and attached them to the OpenAPI MediaType so that the dict keys and values can be parsed by the Swagger UI into the picker on the /docs page. This means that the call to `Body` does double duty, it defines the 'schema' with strict conformance to the JSON Schema format _and_ its attributes are used to build the 'content[application/json]' (i.e., media type) spec, which uses a different 'examples' convention than schemas so that Swagger UI can parse it into the picker. 

To restore the examples picker, we need a convention for passing along the examples dict so that it gets correctly attached to the Media Type in Open API spec. When we pass the examples as a list, it's constructed correctly in the 'schema' but the only way to view the examples is to read the openapi.json -- Swagger UI renders only the first example from the list.

This PR proposes to restore this useful functionality, and I've explored two implementations that seem to work well.

### 1) Add a new `media_type_extra` kwarg to `Body`
The changes in this PR implement a minimal diff that adds the desired functionality. This implementation has the advantage that it continues the nice semantic similarity with `Body` kwargs & JSON Schema fields & types while also allowing for the dict of examples to be elevated into the Media Type portion so that Swagger UI can find it. Here's an example based on schema_extra_example/tutorial0004.py:

```python
from typing import Union

app = FastAPI()


class Item(BaseModel):
    name: str
    description: Union[str, None] = None
    price: float
    tax: Union[float, None] = None


item_examples = {
    "Example Item": {
        "value": {
            "name": "Foo",
            "description": "A very nice Item",
            "price": 35.4,
            "tax": 3.2,
        }
    },
    "Example Item; coerce string to float": {
        "value": {
            "name": "Bar",
            "price": "35.4",
        }
    },
    "Raise validation error for 'price'": {
        "value": {
            "name": "Baz",
            "price": "thirty five point four",
        }
    },
}


@app.put("/items/{item_id}")
async def update_item(
    *,
    item_id: int,
    item: Item = Body(
        examples=[dct["value"] for dct in item_examples.values()],
        media_type_extra={"examples": item_examples},
    ),
):
    results = {"item_id": item_id, "item": item}
    return results
```

This implementation allows users to explicitly elevate their examples into the Media Type if they wish _and_ define examples on the schema that conform with JSON Schema.

The main drawback of this option is that it adds a new kwarg, resulting in a breaking change (if you consider loss of the UI picker a breaking change).

### 2) Fancy handling of Body examples depending on type
The other way I got this to work in a backward compatible way is to check the type of the 'examples' passed to `get_openapi_operation_request_body` and if it's a dict, also attach it to the Media Type.

This method is backward compatible, but it requires us to loosen the type narrowing on the `Body` examples kwarg to something like:
```python
...
examples: Optional[Union[List[Any], Dict[str, Any]]] = None,
...
```

## Options for moving forward
It's entirely possible that #9770 deliberately removed the functionality that enabled the multiple examples picker in Swagger UI, and if that's the case, let's just update the docs to remove the image that shows the dropdown picker. 

It's also possible that we simply forgot that the `Body` is used to build both the schema _and_ the media type, and we didn't have time to think about cute example pickers during the herculean effort of integrating both the OpenAPI updates and compat with `pydantic` v2. I hope this is the case, and that this PR helps start the discussion on how to best implement support for the Swagger UI example picker.

This PR doesn't (yet) include any changes to the docs, but I'm happy to contribute to those if one of the above approaches is preferred/accepted. I also volunteer to help add to & revise the tests as well.

### DIFF
--- a/docs_src/schema_extra_example/tutorial004.py
+++ b/docs_src/schema_extra_example/tutorial004.py
@@ -13,27 +13,37 @@ class Item(BaseModel):
     tax: Union[float, None] = None
 
 
+item_examples = {
+    "Example Item": {
+        "value": {
+            "name": "Foo",
+            "description": "A very nice Item",
+            "price": 35.4,
+            "tax": 3.2,
+        }
+    },
+    "Example Item; coerce string to float": {
+        "value": {
+            "name": "Bar",
+            "price": "35.4",
+        }
+    },
+    "Raise validation error for 'price'": {
+        "value": {
+            "name": "Baz",
+            "price": "thirty five point four",
+        }
+    },
+}
+
+
 @app.put("/items/{item_id}")
 async def update_item(
     *,
     item_id: int,
     item: Item = Body(
-        examples=[
-            {
-                "name": "Foo",
-                "description": "A very nice Item",
-                "price": 35.4,
-                "tax": 3.2,
-            },
-            {
-                "name": "Bar",
-                "price": "35.4",
-            },
-            {
-                "name": "Baz",
-                "price": "thirty five point four",
-            },
-        ],
+        examples=[dct["value"] for dct in item_examples.values()],
+        media_type_extra={"examples": item_examples},
     ),
 ):
     results = {"item_id": item_id, "item": item}
--- a/docs_src/schema_extra_example/tutorial004.py
+++ b/docs_src/schema_extra_example/tutorial004.py
@@ -1,16 +1,9 @@
 from typing import Union
 
 from fastapi import Body, FastAPI
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 app = FastAPI()
-
-
-class Item(BaseModel):
-    name: str
-    description: Union[str, None] = None
-    price: float
-    tax: Union[float, None] = None
 
 
 item_examples = {
@@ -36,14 +29,34 @@ item_examples = {
     },
 }
 
+item_examples_list = [dct["value"] for dct in item_examples.values() if "value" in dct]
+media_type_extra = {"examples": item_examples}
+
+
+class Item(BaseModel):
+    name: str = Field(examples=["Foo", "Bar"])
+    description: str | None = Field(default=None, examples=["A very nice Item"])
+    price: float = Field(examples=[35.4])
+    tax: float | None = Field(default=None, examples=[3.2, None])
+
+    model_config = {
+        "json_schema_extra": {
+            "example": {
+                "name": "Bar",
+                "price": "35.4",
+            },
+            "examples": item_examples_list[:1],
+        }
+    }
+
 
 @app.put("/items/{item_id}")
 async def update_item(
     *,
     item_id: int,
     item: Item = Body(
-        examples=[dct["value"] for dct in item_examples.values()],
-        media_type_extra={"examples": item_examples},
+        examples=item_examples,
+        media_type_extra=media_type_extra,
     ),
 ):
     results = {"item_id": item_id, "item": item}
--- a/docs_src/schema_extra_example/tutorial004.py
+++ b/docs_src/schema_extra_example/tutorial004.py
@@ -1,53 +1,16 @@
 from typing import Union
 
 from fastapi import Body, FastAPI
-from pydantic import BaseModel, Field
+from pydantic import BaseModel
 
 app = FastAPI()
 
 
-item_examples = {
-    "Example Item": {
-        "value": {
-            "name": "Foo",
-            "description": "A very nice Item",
-            "price": 35.4,
-            "tax": 3.2,
-        }
-    },
-    "Example Item; coerce string to float": {
-        "value": {
-            "name": "Bar",
-            "price": "35.4",
-        }
-    },
-    "Raise validation error for 'price'": {
-        "value": {
-            "name": "Baz",
-            "price": "thirty five point four",
-        }
-    },
-}
-
-item_examples_list = [dct["value"] for dct in item_examples.values() if "value" in dct]
-media_type_extra = {"examples": item_examples}
-
-
 class Item(BaseModel):
-    name: str = Field(examples=["Foo", "Bar"])
-    description: str | None = Field(default=None, examples=["A very nice Item"])
-    price: float = Field(examples=[35.4])
-    tax: float | None = Field(default=None, examples=[3.2, None])
-
-    model_config = {
-        "json_schema_extra": {
-            "example": {
-                "name": "Bar",
-                "price": "35.4",
-            },
-            "examples": item_examples_list[:1],
-        }
-    }
+    name: str
+    description: Union[str, None] = None
+    price: float
+    tax: Union[float, None] = None
 
 
 @app.put("/items/{item_id}")
@@ -55,8 +18,22 @@ async def update_item(
     *,
     item_id: int,
     item: Item = Body(
-        examples=item_examples,
-        media_type_extra=media_type_extra,
+        examples=[
+            {
+                "name": "Foo",
+                "description": "A very nice Item",
+                "price": 35.4,
+                "tax": 3.2,
+            },
+            {
+                "name": "Bar",
+                "price": "35.4",
+            },
+            {
+                "name": "Baz",
+                "price": "thirty five point four",
+            },
+        ],
     ),
 ):
     results = {"item_id": item_id, "item": item}
--- a/docs_src/schema_extra_example/tutorial005.py
+++ b/docs_src/schema_extra_example/tutorial005.py
@@ -1,4 +1,4 @@
-from fastapi import Body, FastAPI
+from fastapi import Body, FastAPI, Path
 from pydantic import BaseModel, Field
 
 app = FastAPI()
@@ -28,7 +28,6 @@ item_examples = {
 }
 
 item_examples_list = [dct["value"] for dct in item_examples.values() if "value" in dct]
-media_type_extra = {"examples": item_examples}
 
 
 class Item(BaseModel):
@@ -43,7 +42,7 @@ class Item(BaseModel):
                 "name": "Bar",
                 "price": "35.4",
             },
-            "examples": item_examples_list[:1],
+            "examples": item_examples_list,
         }
     }
 
@@ -51,11 +50,15 @@ class Item(BaseModel):
 @app.put("/items/{item_id}")
 async def update_item(
     *,
-    item_id: int,
-    item: Item = Body(
-        examples=item_examples,
-        media_type_extra=media_type_extra,
+    item_id: int = Path(
+        ...,
+        examples={
+            "id as int": {"value": 5},
+            "id as string": {"value": "5"},
+            "invalid id": {"value": "anything else"},
+        },
     ),
+    item: Item = Body(examples=item_examples),
 ):
     results = {"item_id": item_id, "item": item}
     return results
--- a/docs_src/schema_extra_example/tutorial005.py
+++ b/docs_src/schema_extra_example/tutorial005.py
@@ -1,5 +1,7 @@
+from typing import Union
+
 from fastapi import Body, FastAPI, Path
-from pydantic import BaseModel, Field
+from pydantic import BaseModel
 
 app = FastAPI()
 
@@ -31,10 +33,10 @@ item_examples_list = [dct["value"] for dct in item_examples.values() if "value" 
 
 
 class Item(BaseModel):
-    name: str = Field(examples=["Foo", "Bar"])
-    description: str | None = Field(default=None, examples=["A very nice Item"])
-    price: float = Field(examples=[35.4])
-    tax: float | None = Field(default=None, examples=[3.2, None])
+    name: str
+    description: Union[str, None] = None
+    price: float
+    tax: Union[float, None] = None
 
     model_config = {
         "json_schema_extra": {
--- a/docs_src/schema_extra_example/tutorial005.py
+++ b/docs_src/schema_extra_example/tutorial005.py
@@ -1,5 +1,3 @@
-from typing import Union
-
 from fastapi import Body, FastAPI
 from pydantic import BaseModel, Field
 
--- a/docs_src/schema_extra_example/tutorial005.py
+++ b/docs_src/schema_extra_example/tutorial005.py
@@ -1,0 +1,63 @@
+from typing import Union
+
+from fastapi import Body, FastAPI
+from pydantic import BaseModel, Field
+
+app = FastAPI()
+
+
+item_examples = {
+    "Example Item": {
+        "value": {
+            "name": "Foo",
+            "description": "A very nice Item",
+            "price": 35.4,
+            "tax": 3.2,
+        }
+    },
+    "Example Item; coerce string to float": {
+        "value": {
+            "name": "Bar",
+            "price": "35.4",
+        }
+    },
+    "Raise validation error for 'price'": {
+        "value": {
+            "name": "Baz",
+            "price": "thirty five point four",
+        }
+    },
+}
+
+item_examples_list = [dct["value"] for dct in item_examples.values() if "value" in dct]
+media_type_extra = {"examples": item_examples}
+
+
+class Item(BaseModel):
+    name: str = Field(examples=["Foo", "Bar"])
+    description: str | None = Field(default=None, examples=["A very nice Item"])
+    price: float = Field(examples=[35.4])
+    tax: float | None = Field(default=None, examples=[3.2, None])
+
+    model_config = {
+        "json_schema_extra": {
+            "example": {
+                "name": "Bar",
+                "price": "35.4",
+            },
+            "examples": item_examples_list[:1],
+        }
+    }
+
+
+@app.put("/items/{item_id}")
+async def update_item(
+    *,
+    item_id: int,
+    item: Item = Body(
+        examples=item_examples,
+        media_type_extra=media_type_extra,
+    ),
+):
+    results = {"item_id": item_id, "item": item}
+    return results
--- a/docs_src/schema_extra_example/tutorial006.py
+++ b/docs_src/schema_extra_example/tutorial006.py
@@ -1,0 +1,166 @@
+"""Creating examples that are useful in Swagger UI were removed in v0.100.0
+for all parameter types.
+I recommend trying this example app in both to see how it changes.
+"""
+
+from enum import Enum
+from typing import Dict, Union
+
+from fastapi import Body, FastAPI, Header, HTTPException, Path, Query, Request, status
+from pydantic import BaseModel
+
+app = FastAPI()
+
+
+class Item(BaseModel):
+    name: str
+    description: Union[str, None] = None
+    price: float
+    tax: Union[float, None] = None
+
+
+ITEM_STORE = {}
+
+item_examples = {
+    "normal": {
+        "summary": "A normal example",
+        "description": "A **normal** item works correctly.",
+        "value": {
+            "name": "Foo",
+            "description": "A very nice Item",
+            "price": 35.4,
+            "tax": 3.2,
+        },
+    },
+    "converted": {
+        "summary": "An example with converted data",
+        "description": "FastAPI can convert price `strings` to actual `numbers` automatically",
+        "value": {
+            "name": "Bar",
+            "price": "35.4",
+        },
+    },
+    "invalid": {
+        "summary": "Invalid data is rejected with an error",
+        "value": {
+            "name": "Baz",
+            "price": "thirty five point four",
+        },
+    },
+}
+
+
+class ErrorModel(BaseModel):
+    detail: Union[str, Dict[str, str]]
+
+
+class ErrorCode(str, Enum):
+    UNSUPPORTED_FORMAT = "UNSUPPORTED_FORMAT"
+    ANYTHING_BUT_GEOJSON = "ANYTHING_BUT_GEOJSON"
+    BAD_HEADER = "BAD_HEADER"
+
+
+@app.post(
+    "/items/{item_id}",
+    responses={
+        status.HTTP_400_BAD_REQUEST: {
+            "model": ErrorModel,
+            "content": {
+                "application/json": {
+                    "examples": {
+                        ErrorCode.UNSUPPORTED_FORMAT: {
+                            "summary": "The format is not recognized.",
+                            "value": {"detail": ErrorCode.UNSUPPORTED_FORMAT},
+                        },
+                        ErrorCode.ANYTHING_BUT_GEOJSON: {
+                            "summary": (
+                                "Amazingly, this format is supported on AGOL"
+                                "but not on any desktop software made by ESRI."
+                            ),
+                            "value": {
+                                "detail": {
+                                    "code": ErrorCode.ANYTHING_BUT_GEOJSON,
+                                    "reason": "Likely on purpose.",
+                                }
+                            },
+                        },
+                    }
+                }
+            },
+        },
+        status.HTTP_500_INTERNAL_SERVER_ERROR: {
+            "model": ErrorModel,
+            "content": {
+                "application/json": {
+                    "examples": {
+                        ErrorCode.BAD_HEADER: {
+                            "summary": ("This header crashes the server. It's bad."),
+                            "value": {
+                                "detail": {
+                                    "code": ErrorCode.BAD_HEADER,
+                                    "reason": "Likely on purpose.",
+                                }
+                            },
+                        },
+                    }
+                }
+            },
+        },
+    },
+)
+async def post_item(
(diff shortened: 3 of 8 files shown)


desilinguist:
Thank you so much for this @austinorr! Just ran into this issue myself! 🙇🏽

austinorr:
> Thank you so much for this @austinorr! Just ran into this issue myself! 🙇🏽

Does this PR (Option 1 n the discussion above) fix it for you?

desilinguist:
Hmm, actually it doesn't seem to work for me :( Here's how I have defined my requests: 

```python
example_requests = {
    "text": {
        "basic": {
            "value": {"request_id": "basic", "text": "..."},
        },
        "nogarbage": {
            "value": {"request_id": "no_garbage", "detect_garbage": False, "text": "..."},
        },
        "prompt": {
            "value": {
                "request_id": "prompt_extra_time",
                "promptid": "example_prompt_id",
                "text": "...",
                "timeout": 200,
            },
        },
        "score_grade": {
            "value": {
                "request_id": "score_text",
                "scoring_model": "writing_quality",
                "grade": 8,
                "text": "...",
            },
        },
        "extras": {
            "value": {
                "request_id": "extras_text",
                "text": "...",
                "extras": {
                    "prompttext": "...",
                    "promptxml": "...",
                    "specified_topics": {"budget": []},
                },
            },
        },
    },
    "audio": { ... }
```

And here's how I am using them in my endpoint definition:

```python
@router.post("/api/v1/features/text", tags=["features"], responses={**sync_responses})
def submit_text_request(
    background_tasks: BackgroundTasks,
    rfs_request: RFSTextRequest = Body(
        ..., media_type_extra=example_requests["text"]
    ),
    commons: dict = Depends(common_sync_parameters),
    user: User = Depends(get_current_active_user),
    logging_config: S3LoggingConfig = Depends(),
):
...
```

And the UI still looks like this:

CleanShot 2023-07-12 at 16 09 24@2x

desilinguist:
Never mind @austinorr, it does works! Just realized that I had a bug in how I was using `media_type_extra`. It should be:

```python
media_type_extra={"examples": example_requests["text"]},
```

and not just:
```python
media_type_extra= example_requests["text"],
```

hochstibe:
It worked for me for a body parameter. Should it work as well on other parameter types? I tested with Header, but it was not included in openapi.json.

In my opinion, the `media_type_extra` is good, if you want to give additional context to the examples (like the "summary"). Just for adding one or more examples, it seems a bit clumsy.

austinorr:
@hochstibe I've only added it to the `Body` so far, thought I'd put this patch out there, gauge the need, and propose an approach like adding  that new kwarg. It's the function I most often need to build up multiple complex examples for and it's the one presented in the docs, so that's where I started. Depending on the consensus, I imagine the other parameter types could be supported too.

> In my opinion, the media_type_extra is good, if you want to give additional context to the examples (like the "summary"). Just for adding one or more examples, it seems a bit clumsy.

I agree with you that for adding examples the new kwarg is clumsy. If you're attaching just one example then you can use the 'example' or 'examples' kwargs, but if you have multiple examples and you want the picker in Swagger UI docs, we need to either add the kwarg that accepts the examples as a dict (not a list), or we need to agree to loosen the type expectations to allow users to pass the mapping directly to the 'examples'. 

I think the 'media_type_extra' kwarg is ugly, but it clearly shows where the data is going to be placed, and it allows the other `Body` kwargs to be aligned with job of building the content schema, even though `Body` technically also builds the  media_type (where the 'examples' as a dict go for the UI examples picker) as well.

austinorr:
@hochstibe if you want to try to get this working on `Header`, you can look at the changes I made to `Body` [here](https://github.com/tiangolo/fastapi/pull/9866/files#diff-d9b3d8d061f78332319cf3ada9cf9d8517d130f8971c2ee1ba5589a34d300036), and [here](https://github.com/tiangolo/fastapi/pull/9866/files#diff-e764e3d0b1f181634ee4377f83085e82eb19f121009a4f8dd31ad6e8090632a3) and do something similar to the others you'd like to try. I think those are the only places you'd need to make changes to support all the other parameter types.

desilinguist:
I agree with @austinorr. I also use mostly examples in the Body as the most common use case. I think this solution, while a bit hacks, is fairly reasonable and quite necessary as showing multiple examples is quite important in many scenarios.

UltimateLobster:
I''m not sure about the "media_type_extra" implementation. I think it would go against the ease that FastAPI supposed to achieve. I think your alternate suggestion would be better (where the "examples" field can be a dict). The avarage user would simply go by what the IDE suggests them. If they see the "examples" field, they would expect it to "just work" without extra work.

I think the FastAPI interface should go based on "what would be the most intuitive usage for the avarage developer" and then do whatever we need behind the scenes to adapt it to any specification we wish to support. The examples picker is a useful option and therefore we should strive to have it "when possible" i.e. when the user specified a dictionary of examples.

austinorr:
@UltimateLobster I’m traveling today and tomorrow, but early this week I’ll add a commit here that has a draft implementation of the ‘if dict then that’ approach. It’ll re-use the same ‘examples’ kwarg to build both valid JSON schema examples (list) and valid media_type OpenAPI examples (dict) that works with Swagger UI. 

If this is the way we want to go, I’ll also clean this up to remove the new kwarg after this is green-lit by a maintainer review.

tiangolo:
📝 Docs preview for commit 1d7c2f823c8fcb9df64e3ca2e35de10d703af717 at: https://64b5c5b7e6ac89131eb4e067--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit fbc4adb80f72cd92d6bb9a1cdae38a2012c6f2c9 at: https://64b5d2038221c81b82c3b499--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit d950857413fcff535b34bb9f99590c2e15c22182 at: https://64b6e31e7f45ee1bf4f7da3a--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit 43d3880c833d8efa97ab2080546a2d16b4c55dd7 at: https://64b7294a68eede254dccdd19--fastapi.netlify.app

austinorr:
@UltimateLobster @hochstibe I've finished this update that adds the 'examples' as dict back to the correct spot on the OpenAPI schema and it works on all parameters. Y'all care to give it a shot to see if it works for you?

I've added a new file `docs_src/schema_extra_example/tutorial006.py` that demonstrates how to add examples to all the params, and I hope there's a path toward re-enabling support for Swagger UI's interactive examples.

Since this piece of OpenAPI 3.1 didn't change at all with their update (Swagger UI examples have always been siblings of schemas, rather than children, and that's why they don't work now.), this tutorial works just as well with `fastapi`<0.100.0. This patch just re-enables all this nice UI. See below for a screen shot of an API with working multi-examples for each of Body, Path, Query, Header, and Responses. Cookie should work too, I just ran out of time to add a demo one to the app.

image

The thinking here is that I can always go _backwards_ from the dict to a list to pull out the examples that the user passes. I can then add the full dict to the correct location on the OpenAPI schema as a sibling of 'schema' rather than as a child, and I can re-add the list to the 'schema' as a child. I do this only if the examples is passed as a dict, so that all the current tests pass. This works great, and appears to me to achieve both goals. If we like it, I'll add tests so it's fully covered.

There is a single test failing on all platforms currently, which is checking that the entire OpenAPI schema is identical to the reference in the source code (tests/test_schema_extra_examples.py::test_openapi_schema). I can't make this PR conform to the old ref, so getting this to pass tests may require changing the test. I'm happy to invest more time in that if asked to by a maintainer.

For posterity/completeness, here's the schema that generated the above ui in `docs_src/schema_extra_example/tutorial006.py`
```
{
  "openapi": "3.1.0",
  "info": {
    "title": "FastAPI",
    "version": "0.1.0"
  },
  "paths": {
    "/items/{item_id}": {
      "post": {
        "summary": "Post Item",
        "operationId": "post_item_items__item_id__post",
        "parameters": [
          {
            "name": "item_id",
            "in": "path",
            "required": true,
            "schema": {
              "type": "integer",
              "examples": [5, "5", "anything else"],
              "title": "Item Id"
            },
            "examples": {
              "id as int": {
                "value": 5
              },
              "id as string": {
                "value": "5"
              },
              "invalid id": {
                "value": "anything else"
              }
            }
          },
          {
            "name": "f",
            "in": "query",
            "required": false,
            "schema": {
              "anyOf": [
                {
                  "type": "string"
                },
                {
                  "type": "null"
                }
              ],
              "examples": ["wkt", "json", "geojson"],
              "default": "wkt",
              "title": "F"
            },
            "examples": {
              "text (default)": {
                "value": "wkt"
              },
              "response as json": {
                "value": "json"
              },
              "format 3": {
                "summary": "response as geojson",
                "value": "geojson"
              }
            }
          },
          {
            "name": "strange_header",
            "in": "header",
            "required": false,
            "schema": {
              "anyOf": [
                {
                  "type": "string"
                },
                {
                  "type": "null"
                }
              ],
              "examples": ["some long string", "bad header"],
              "title": "Strange Header"
            },
            "examples": {
              "good": {
                "value": "some long string"
              },
              "bad": {
                "summary": "this one's bad",
                "value": "bad header"
              }
            }
          }
        ],
        "requestBody": {
          "required": true,
          "content": {
            "application/json": {
              "examples": {
                "normal": {
                  "summary": "A normal example",
                  "description": "A **normal** item works correctly.",
                  "value": {
                    "name": "Foo",
                    "description": "A very nice Item",
                    "price": 35.4,
                    "tax": 3.2
                  }
                },
                "converted": {
                  "summary": "An example with converted data",
                  "description": "FastAPI can convert price `strings` to actual `numbers` automatically",
                  "value": {
                    "name": "Bar",
                    "price": "35.4"
                  }
                },
                "invalid": {
                  "summary": "Invalid data is rejected with an error",
                  "value": {
                    "name": "Baz",
                    "price": "thirty five point four"
                  }
                }
              },
              "schema": {
                "allOf": [
                  {
                    "$ref": "#/components/schemas/Item"
                  }
                ],
                "examples": [
                  {
                    "name": "Foo",
                    "description": "A very nice Item",
                    "price": 35.4,
                    "tax": 3.2
                  },
                  {
                    "name": "Bar",
                    "price": "35.4"
                  },
                  {
                    "name": "Baz",
                    "price": "thirty five point four"
                  }
                ],
                "title": "Item"
              }
            }
          }
        },
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {}
              }
            }
          },
          "400": {
            "content": {
              "application/json": {
                "examples": {
                  "UNSUPPORTED_FORMAT": {
                    "summary": "The format is not recognized.",
                    "value": {
                      "detail": "UNSUPPORTED_FORMAT"
                    }
                  },
                  "ANYTHING_BUT_GEOJSON": {
                    "summary": "Amazingly, this format is supported on AGOL but not on any desktop software made by ESRI.",
                    "value": {
                      "detail": {
                        "code": "ANYTHING_BUT_GEOJSON",
                        "reason": "Likely on purpose."
                      }
                    }
                  }
                },
                "schema": {
                  "$ref": "#/components/schemas/ErrorModel"
                }
              }
            },
            "description": "Bad Request"
          },
          "422": {
            "content": {
              "application/json": {
                "schema": {
                }
              }
            }
          },
          "500": {
            "content": {
              "application/json": {
                "examples": {
                  "BAD_HEADER": {
                    "summary": "This header crashes the server. It's bad.",
                    "value": {
                      "detail": {
                        "code": "BAD_HEADER",
                        "reason": "Likely on purpose."
                      }
                    }
                  }
                },
                "schema": {
                  "$ref": "#/components/schemas/ErrorModel"
                }
              }
            },
            "description": "Internal Server Error"
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "ErrorModel": {
        "properties": {
          "detail": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "additionalProperties": {
                  "type": "string"
                },
                "type": "object"
              }
            ],
            "title": "Detail"
          }
        },
        "type": "object",
        "required": ["detail"],
        "title": "ErrorModel"
      },
      "HTTPValidationError": {
        "properties": {
          "detail": {
            "items": {
              "$ref": "#/components/schemas/ValidationError"
            },
            "type": "array",
            "title": "Detail"
          }
        },
        "type": "object",
        "title": "HTTPValidationError"
      },
      "Item": {
        "properties": {
          "name": {
            "type": "string",
            "title": "Name"
          },
          "description": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
            "title": "Description"
          },
          "price": {
            "type": "number",
            "title": "Price"
          },
          "tax": {
            "anyOf": [
              {
                "type": "number"
              },
              {
                "type": "null"
              }
            ],
            "title": "Tax"
          }
        },
        "type": "object",
        "required": ["name", "price"],
        "title": "Item"
      },
      "ValidationError": {
        "properties": {
          "loc": {
            "items": {
              "anyOf": [
                {
                  "type": "string"
                },
                {
                  "type": "integer"
                }
              ]
            },
            "type": "array",
            "title": "Location"
          },
          "msg": {
            "type": "string",
            "title": "Message"
          },
          "type": {
            "type": "string",
            "title": "Error Type"
          }
        },
        "type": "object",
        "title": "ValidationError"
      }
    }
  }
}

```

I love using `fastapi` because it helps me build tricky cases right into my docs and communicate with my team. Adding interactive examples to the UI so they don't have to read the actual openapi.json to see what examples I've prepared used to save me time and be seamless. I recognize that there are unsupported edge cases here, like if the user passes multiple Body's to the route, FastAPI will embed them into a "super body", and in that case, this PR actually throws a very clear `pydantic` validation error that 'examples' must be passed as list. 

This first one does error, since the contents of the super body are schemas, and their examples must be lists. The error makes it easy to express my thought a different way, either make my own super body and add examples that work with the picker, or convert my examples to a list so they're embedded in the schema and move on.
```python

item_examples = {
    "normal": {
        "summary": "A normal example",
        "description": "A **normal** item works correctly.",
        "value": {
            "name": "Foo",
            "description": "A very nice Item",
            "price": 35.4,
            "tax": 3.2,
        },
    },
    "converted": {
        "summary": "An example with converted data",
        "description": "FastAPI can convert price `strings` to actual `numbers` automatically",
        "value": {
            "name": "Bar",
            "price": "35.4",
        },
    },
    "invalid": {
        "summary": "Invalid data is rejected with an error",
        "value": {
            "name": "Baz",
            "price": "thirty five point four",
        },
    },
}

@app.post("/items/{item_id}")
async def post_item(
    request: Request,
    *,
    item_id: int = Path(
        examples={
            "id as int": {"value": 5},
            "id as string": {"value": "5"},
            "invalid id": {"value": "anything else"},
        }
    ),
    user: User,
    item: Item = Body(examples=item_examples)
):
    ...
    return
```

This doesn't error, and super body is made/used flawlessly.
```python

@app.post("/items/{item_id}")
async def post_item(
    request: Request,
    *,
    item_id: int = Path(
        examples={
            "id as int": {"value": 5},
            "id as string": {"value": "5"},
            "invalid id": {"value": "anything else"},
        }
    ),
    user: User,
    item: Item = Body(examples=[dct["value"] for dct in item_examples.values()])
):
    ...
    return
```

Swagger UI examples are a feature I use often, and from what I've seen in the discussion many others feel the same. I think the interactive docs are a major selling point for `fastapi` and if new users cannot build the full Swagger UI compliant openapi.json, including the examples pickers, then I'll bet they'll keep asking this question on SO, and on the discussion boards, and keep opening issues, either here on `fastapi`, or even on the `openapi` github ([eg](https://github.com/Rhosys/openapi-explorer/issues/163)).

thomasaarholt:
Hello and thank you very much for dealing with this! I noticed this same regression on migration to pydantic 2.

I think a) this sort of example-list functionality is extremely desirable, but b) that it is starting to get a little confusing as to where one should add examples. It would be nice if there was only one (or maybe two) places to put examples, but with the model_config, this makes three different ways, and its not super clear to me when one wishes to use which one.

A challenge: Is there any reason that the examples keyword has to conform to the OpenAPI `examples` spec? It doesn't seem sensible for this option to _not_ be a dict. I don't understand how an `examples` of type `list` is currently used by fastapi 0.100 today, other than just showing the first example in the list.

austinorr:
> b) that it is starting to get a little confusing as to where one should add examples

You're right. Best case is we might have to accept some different functionality for 'examples' declared in our BaseModels model_config (I agree that these should be enforced as lists, since our models only ever build _schemas_) and examples declared in `fastapi` Params (like Body, Query, etc) which build the schema wrapper objects that have different requirements for 'examples' in the OpenAPI spec (they must be dicts). I think the 'examples' kwarg on the Params should be the one spot to add examples that work with the docs. This PR (re)enables that.

In my opinion, `fastapi`'s sweet spot is helping developers build a high performance API with best in class docs with a library that 'does the right thing'. In this case, I feel that the right thing to do when I'm adding examples to my _route_ by calling Body or any other Param directly, is not to _modify a data model I already declared with `pydantic`_, especially when doing so hobbles the docs and leaves very useful portions of the OpenAPI spec inaccessible.

To be fair, the docs will render the first example in the schema for a RequestBody, but it does not do this for any of the parameter objects like query, cookie, header, or path (MRE below). Examples that are declared as a dict and as a _sibling_ to the schema is still the right way to add examples to all of these -- this did not change at all with the 3.0->3.1 update or with Swagger v5.

MRE for how swagger ignores examples within the schema for any param. Copy the json below into (SwaggerEditor)[https://editor-next.swagger.io/]
In v0.100.0 if I add this to a route
```python
f: Union[str, None] = Query(None, examples=["json", "geojson", "wkt"]),
```
I get this OpenAPI spec with the 'schema' member being a correct JSON Schema, but Swagger docs don't see the examples at all.
```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "FastAPI",
    "version": "0.1.0"
  },
  "paths": {
    "/items/{item_id}": {
      "post": {
        "summary": "Post Item",
        "operationId": "post_item_items__item_id__post",
        "parameters": [
          {
            "name": "f",
            "in": "query",
            "required": false,
            "schema": {
              "anyOf": [
                {
                  "type": "string"
                },
                {
                  "type": "null"
                }
              ],
              "examples": [
                "json",
                "geojson",
                "wkt"
              ],
              "title": "F"
            }
          }
        ],
        "responses": {}
      }
    }
  },
  "components": {}
}
```

The only way to get an example on a param is with the dict. This is not a bug in Swagger, they are correctly interpreting the OpenAPI 3.1 spec for [how to declare examples on parameters](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#parameter-object)

The below json fixes the MRE above, and the 'schema' member still fully complies with JSON Schema. This output is achievable with the changes included in this PR.
```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "FastAPI",
    "version": "0.1.0"
  },
  "paths": {
    "/items/{item_id}": {
      "post": {
        "summary": "Post Item",
        "operationId": "post_item_items__item_id__post",
        "parameters": [
          {
            "name": "f",
            "in": "query",
            "required": false,
            "schema": {
              "anyOf": [
                {
                  "type": "string"
                },
                {
                  "type": "null"
                }
              ],
              "examples": [
                "json",
                "geojson",
                "wkt"
              ],
              "title": "F"
            },
            "examples": {
              "well known text": {"value": "wkt"},
              "json": {"value": "json"}
            }
          }
        ],
        "responses": {}
      }
    }
  },
  "components": {}
}
```

tiangolo:
📝 Docs preview for commit 45cdef504d9075fc4dd2f354bd417d71cc403180 at: https://64bad21d2f20a1007c9dea3c--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit f8c9537d2d5a489e54e75387422a3ef513fb4192 at: https://64c03f3525931d0675b466a6--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit 56ed2248f82dd6b59a75c1362a22203d7ecf5c48 at: https://64c0410d84547c0ea4f80287--fastapi.netlify.app

ezeparziale:
Any news about this point? or alternatives?

austinorr:
I've been working on cleaning this up and getting it tested. Now we wait until a maintainer either reviews and agrees, or decides that OpenAPI examples are not a feature that `fastapi` supports anymore. If I'm being honest, based on the 'answer' on the discussion I don't think this PR stands much of a chance, but it's nice to know that others relied on OpenAPI/Swagger UI docs examples too.

I'm hoping this is just a very low priority for the maintainers, since that might mean once the major transition to v0.100+ & pydantic v2 is stable enough they'll have bandwidth for this feature. When that time comes, hopefully having this already written and tested helps them.

tiangolo:
📝 Docs preview for commit 04846ac9422ae56f3bb9dc09b8db87374cc86613 at: https://64c3e1c273d55419c5455a63--fastapi.netlify.app

crvillanueva:
I really liked this feature too, I hope this gets merged.

v-raja:
+1 for this feature. It's super important to be able to describe examples for endpoints as close as possible to the code. I can't image handling examples for my FastAPI endpoints in another place (or even worse, not having examples because it requires additional overhead to maintain!).

b-rad-c:
Also leaving a +1 on this PR, I think it's important to be able to put multiple examples in the swagger page and describe them.

I created a [pull request](https://github.com/austinorr/fastapi/pull/1) to your pull request which makes it even easier to add multiple examples in the swagger page. I think it gives us the best of both worlds: simplicity and complexity if you need it. You can pass in a dictionary of examples, or you can supply a list and it will generate names based on list index for the to populate the dropdown.

Now it's very fastapi/pythonic in that you can get a list of examples easily and you can also easily add descriptive examples if you want to take the time. The following code works in my [PR](https://github.com/austinorr/fastapi/pull/1), it includes a list of examples and a dictionary of descriptive examples, you can test that both work by swapping which variable is provided to the `Body` object.

```
from typing import Union


class Item(BaseModel):
    name: str
    description: Union[str, None] = None
    price: float
    tax: Union[float, None] = None


named_examples = {
    "normal": {
        "summary": "A normal example",
        "description": "A **normal** item works correctly.",
        "value": {
            "name": "Foo",
            "description": "A very nice Item",
            "price": 35.4,
            "tax": 3.2,
        },
    },
    "converted": {
        "summary": "An example with converted data",
        "description": "FastAPI can convert price `strings` to actual `numbers` automatically",
        "value": {
            "name": "Bar",
            "price": "35.4",
        },
    },
    "invalid": {
        "summary": "Invalid data is rejected with an error",
        "value": {
            "name": "Baz",
            "price": "thirty five point four",
        },
    },
}


list_examples = [
    {
        "name": "Foo",
        "description": "A very nice Item!",
        "price": 35.4,
        "tax": 3.2,
    },
    {
        "name": "Bar",
        "price": "35.4",
    },
    {
        "name": "Baz",
        "price": "thirty five point four",
    },
]

#
# app
#

app = FastAPI()

@app.post("/items")
async def post_item(item: Item = Body(examples=named_examples)) -> Item:
    return item
```

tiangolo:
📝 Docs preview for commit f3313ef004ebff41e38dce46be1e3ae165b904f4 at: https://e28a769a.fastapitiangolo.pages.dev

tiangolo:
Hello all! Thanks a lot for all the effort @austinorr and everyone for the discussion. ☕ 

I thought it would be best to have an independent parameter `openapi_examples` for the OpenAPI-specific examples, and to annotate it with a `TypedDict` to help editors and tools verify the contents (I didn't have this before). This is probably closer to the first idea of another argument, although just for the examples. I added that here https://github.com/tiangolo/fastapi/pull/10152 with all the docs and tests.

It's available now in FastAPI 0.103.0, the new docs are here: https://fastapi.tiangolo.com/tutorial/schema-extra-example/#openapi-specific-examples

Given that I'll close this one, thanks again! 🍰

PR #9909, opened 2023-07-19
Classifiers pyproject.toml

kostyaten:
It seems to me that the classifiers do not match the current state of the project!


iudeen:
I believe it’s beta till 1.0

tiangolo:
Thanks @kostyaten! This would be very subjective to what people would consider "beta" or not. Indeed, FastAPI is ready for production, and used in lots of companies and organizations already, but as I haven't released a version 1.x.x and there could be more (small) "breaking changes" before that, I prefer to delay changing this for now. So I'll pass on this one, but thanks! ☕

PR #9910, opened 2023-07-20
Add TaskScheduler

achaayb:
New Feature: Task Scheduler for Function Execution

This pull request introduces a new functionality called "Task Scheduler" by introducing a completely new class called TaskScheduler. This class empowers users to schedule the execution of functions, seamlessly handling both synchronous and asynchronous tasks.

Key Changes:

- TaskScheduler Class: The new TaskScheduler class has been created to manage task scheduling and execution.

- add_task Method: The TaskScheduler class now includes a add_task method. This method allows users to schedule the execution of a function by providing a func parameter to represent the function to be executed, and execute_at datetime parameter for future execution, and optional args and kwargs parameters for passing function arguments and keyword arguments.

Example Usage:

python

# Importing the required dependency
from fastapi import get_scheduler

# Creating an async and a sync function to be used in scheduling
```python
async def sample_async_task(name: str):
    print(f"Executing task for {name} at {datetime.now()}")
    await asyncio.sleep(2)
    print("Execution Done")

def sample_sync_task(name: str):
    print(f"Executing task for {name} at {datetime.now()}")
    time.sleep(2)
    print("Execution Done")
```
# Example Usage
```python
@app.get("/")
async def schedule_async_task_endpoint(scheduler: TaskScheduler = Depends(get_scheduler)):
    
    # Get the current time
    current_time = datetime.now()

    # Calculate the time 3 seconds from now
    execute_at = current_time + timedelta(seconds=10)

    scheduler.add_task(sample_async_task, execute_at, args=("ali",))
    return {"message": "Task scheduled."}

@app.get("/2")
async def schedule_sync_task_endpoint(scheduler: TaskScheduler = Depends(get_scheduler)):

    # Get the current time
    current_time = datetime.now()

    # Calculate the time 3 seconds from now
    execute_at = current_time + timedelta(seconds=10)

    scheduler.add_task(sample_sync_task, execute_at, args=("ali",))
    return {"message": "Task scheduled."}
```

# scheduler.add_task arguments
```
add_task(
function callable sync or async,
execution time,
args as tuple
kwargs as dict
)
```

This new Task Scheduler functionality is introduced through the TaskScheduler class, offering users a robust solution for scheduling and executing functions in a variety of contexts, whether they are synchronous or asynchronous. This PR adds significant value to the project, enhancing its capabilities and enabling efficient task automation and concurrency management.


yezz123:
## IMO

if your application is relatively simple, without any time-sensitive or background processing requirements, and doesn't involve complex workflows, a task scheduler might not be necessary (especially at the framework level).

You can manage simple periodic tasks using basic Python functionalities like the `time.sleep()` function or a simple cron job (if your application is deployed on Unix-like systems).

## What the community use

In the Python ecosystem, there are several libraries available for task scheduling, such as `schedule`, `Celery`. These libraries can be integrated into your app easily, (IMO if you include `TaskScheduler` you will somehow move the architecture of Django, Batteries 🥲).

IMO, the decision to add a `Scheduler` back to Sebastian, but I can't see the value of creating something on the top of the framework level, for an integration that can be easy to add if we use some celery.

tiangolo:
Thanks @achaayb! 🍰 

I actually agree with @yezz123, this would probably belong outside of FastAPI.

There are a lot of things to have in mind for use cases like this, and it's probably better to have a tool like Celery, RQ, ARQ... or maybe Prefect, Dagster, etc.

Some alternatives you could try working on would be tutorials to integrate one of those tools. Or if you definitely want to use this approach, you could write a blog post about how to write this code as part of the app. Or you could even create a library to expose it if you want.

Given that, I'll close this one, but thanks for the effort! ☕

PR #9916, opened 2023-07-20
Fix ValidationExceptions missing type annotations after v0.100.0

MarkusSintonen:
Type information got lost from `ValidationException`s after v0.100.0. This PR brings them back. `ErrorDetails` is used as is from Pydantic V2. With V1 its compatible with Pydantic V1 `.errors()`.

Related https://github.com/tiangolo/fastapi/discussions/9907

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -169,7 +169,7 @@ if PYDANTIC_V2:
             if isinstance(error, dict):
                 use_errors.append(error)
             else:
-                raise NotImplementedError(f"Unexpected ErrorWrapper: {error}")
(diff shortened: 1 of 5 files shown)


tiangolo:
Checking the conversation in the discussion to understand the use case, it seems the use case would be solved by using a custom exception, that would be much simpler and independent of FastAPI internals. :nerd_face: 

Given that, I'll pass on this one, but  thanks for the interest! :coffee:

MarkusSintonen:
@tiangolo why the `errors()` can not be properly typed when the class is documented https://fastapi.tiangolo.com/tutorial/handling-errors/#override-request-validation-exceptions ?

PR #9925, opened 2023-07-22
Fix validation error Optional[list[...]], Union[list[...], ...], Optional[set[...]] and etc in Form

dotX12:
## Fix issues with `Optional` and `Union` for `list, set, tuple, etc...` in `Form`:

```python3
a: Optional[list[int]] = Form(None)
b: Union[list[int], list[float]] = Form(None)
```

which lead to a validation error: `"Input should be a valid list"`

## Reproduce the issue:

```python3
from typing import Optional

app = FastAPI()


@app.post("/form")
async def test_form(
    list_int: Optional[list[int]] = Form(None),
):
    return {"elements": list_int}


client = TestClient(app=app)


def test_send_optional_one_element():
    data = {"list_int": "100"}
    response = client.post("/form", data=data)
    assert response.json() == {"elements": [100]}

    # {'detail': [{'input': '100',
    #              'loc': ['body', 'list_int'],
    #              'msg': 'Input should be a valid list',
    #              'type': 'list_type',
    #              'url': 'https://errors.pydantic.dev/2.0.3/v/list_type'}]} != {'elements': [100]}


def test_send_optional_two_list_elements():
    data = {"list_int": ["10", "20"]}
    response = client.post("/form", data=data)
    assert response.json() == {"elements": [10, 20]}
    # {'detail': [{'input': '20',
    #              'loc': ['body', 'list_int'],
    #              'msg': 'Input should be a valid list',
    #              'type': 'list_type',
    #              'url': 'https://errors.pydantic.dev/2.0.3/v/list_type'}]} != {'elements': [10, 20]}


def test_send_optional_list_elements():
    response = client.post("/form")
    assert response.json() == {"elements": []}
    # {'elements': None} != {'elements': []}


```

Previously, the ModelField had a `shape` attribute that said what kind of field it was (list, set, tuple, etc.).
So then the `Optional[list[int]]` shape had `SHAPE_LIST`

In the new version, this field (shape) no longer exists, and for the Form to parse data, FastAPI itself tries to collect values depending on the type.
for `list[int]` iterates through the collection `value=12&value=20` and does `value=[12, 20]`

In the new version, `Optional[list[int]]` does not count as a list when tested with is_sequence_field , which causes the value to be fetched as one element by :
`value = received_body.get(field.alias)`

If we made an annotation like `Optional[list[int]]` then pydantic expects us to pass `[1, 2, 3]` or `[1]`, but we are passing the value directly, which causes a validation error
`"input should be a valid list"`

In simple terms, what is happening:
```python3
class Foo(BaseModel):
      a: Optional[list[int]] = Field(None)

Foo(a="1")  # raise ValidationError
```
And it should happen:
```python3
Foo(a=["1"])  # Ok
```

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -228,6 +228,19 @@ if PYDANTIC_V2:
     def is_scalar_sequence_field(field: ModelField) -> bool:
         return field_annotation_is_scalar_sequence(field.field_info.annotation)
 
+
+    def field_annotation_is_optional_sequence(field: ModelField) -> bool:
+        if get_origin(field.field_info.annotation) == Union:
+            args = get_args(field.field_info.annotation)
+            if not args:
+                return False
+            else:
+                first_argument = args[0]
+                if hasattr(first_argument, "__origin__"):
+                    if first_argument.__origin__ in sequence_types:
+                        return True
+        return False
+
     def is_bytes_field(field: ModelField) -> bool:
         return is_bytes_or_nonable_bytes_annotation(field.type_)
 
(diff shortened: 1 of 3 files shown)


dotX12:
Initially, I decided to add field_annotation_is_optional_sequence for both V1 and V2, but after analyzing in more detail, V1 works correctly because of the shape, so field_annotation_is_optional_sequence is needed only for V2.

Danipulok:
Nice! Waiting for when this is approved. @tiangolo, this PR seems really nice and ready to merge

dotX12:
@tiangolo, I would be grateful for a review and making these changes :)

dotX12:
Fixed commit naming and opened a new PR: https://github.com/tiangolo/fastapi/pull/9928

PR #9930, opened 2023-07-23
📝 More explicitly explain dependencies in websockets doc page

martinky24:
The note at the top of [the websockets doc page](https://fastapi.tiangolo.com/advanced/websockets/) implies FastAPI has an implicit dependency on the `websockets` package. As a result, I mistakenly spent quite a bit of time reading the `websockets` docs, thinking it would help me work with the `fastapi.websocket` functionality.

FastAPI itself instead relies solely on the Starlette WebSockets implementation, which has an entirely different API. 

This MR attempts to update the documentation to make it more explicit WHY `websockets` is required. It's not a fastapi dependency by any means, but a uvicorn dependency (and the tutorial makes use of uvicorn)


tiangolo:
📝 Docs preview for commit 9944f99f33394a7d9094b983ee367da860b9b1bd at: https://64bc7cd6a11600468968067f--fastapi.netlify.app

tiangolo:
📝 Docs preview for commit 9df46e204c76817c91b76425b39b17821c75cf2c at: https://64be82424e1061008bd78812--fastapi.netlify.app

codespearhead (approved):
Good catch!

Uvicorn is even invoked a few paragraphs down in the same tutorial despite not being installed.

https://github.com/tiangolo/fastapi/blob/62705820d60365cdb4b05010c103c07020190e99/docs/en/docs/advanced/websockets.md?plain=1#L75

YuriiMotov (commented):
The text of this section was updated since the time this PR was opened.
But it still looks a bit incomplete: it's said that we need to create environment and install `websockets`, but it's said nothing about installing FastAPI itself.

How about the following?
---

````
## Install `uvicorn` and `websockets`

Make sure you create a [virtual environment](../virtual-environments.md){.internal-link target=_blank}, activate it, and install `fastapi`, `uvicorn` and `websockets`:

<div class="termy">

```console
$ pip install fastapi uvicorn websockets

---> 100%
```

</div>

/// tip 

If you install FastAPI using `pip install fastapi[standard]`, you don't need to install websockets separately - it will be installed by default.

///

/// note

You will not interact with `websockets` package directly, it's required for Uvicorn to handle the websocket protocol.

///

````

PR #9941, opened 2023-07-25
Shorter param definitions

vitalik:
Introduction of `Annotated` params made function params more logical, but on the other hand longer/more verbose:

```Python
@api.post("/form")
def some_form(
    x: Annotated[int, Form()],
    y: Annotated[int, Form()],
    z: Annotated[LaxInt, Form()],
):
   ...
```

when you look at that you mostly focus on the "annotated" word, while the actual marker is harder to spot (just cognitively)

This PR add a little bit magic to params  by making them act like Annotated without the "annotated" word:

```Python
@api.post("/form")
def some_form(
    x: Form[int],
    y: Form[int],
    z: Form[LaxInt],
):
   ...
```

this look way more clear and backward compatible with `y: Annotated[xxx, Form]`  and  `y: xxx = Form(...)`

To add openapi extras the following code can be used:
```Python
y: Form[int, dict(title="Horizontal axis")]
```

so far all works fine except one little caveat - mypy, pylance do not like this in users and marks as error (Annotated[T] does not allow extra param)

to overcome this the extra params can be introduced like FormEx, BodyEx, HeaderEx = that will be just aliased to Annotated 

```Python
if TYPE_CHECKING:
    ...
    from typing import Annotated as FormEx
    from typing import Annotated as QueryEx
    ...
else:
     ...
     QueryEx = Query
     FormEx = Form
     ...
```

`dict(title="Horizontal axis")` can be enhanced with:

```Python
class Param(TypedDict):
    ge: int
    title: str
    ...
```
Looking forward for feedback


vitalik:
СС: @tiangolo @Kludex @dmontagu

adriangb:
> To add openapi extras the following code can be used:
> 
> ```python
> y: Form[int, dict(title="Horizontal axis")]
> ```

I don't think working around mypy like this makes sense for FastAPI.

> Hi @adriangb - what do you think about this PR ? Is this correct that FastAPI inspired Annotated approach from your [xpresso](https://github.com/adriangb/xpresso) framework ? I think it was cognitively more cleaner and readable... Currently when I look at some FastAPI projects - the actual endpoints are very hard to read because of too many "annotated" word

I've talked with @tiangolo about this a while ago when he first added Annotated support. If I remember his reaction correctly it was that he's not opposed to adding type aliases like in Xpresso but wanted to wait to do it (I don't remember why). I think if you simplify this PR to just that there's a good chance it will eventually make it into FastAPI.

vitalik:
>  think if you simplify this PR...
where do you see simplification ? 

the whole functionality is actually this simple wrapper class:
```Python
class ParamShortcut:
    ...

    def __getitem__(self, args: Any) -> Any:
        if isinstance(args, tuple):
            return Annotated[args[0], self._base_func(**args[1])]
        return Annotated[args, self._base_func()]
```

the rest is basically boilerplate to make mypy happy


> `y: Form[int, dict(title="Horizontal axis")]`
> I don't think working around mypy like this makes sense for FastAPI.


yeah this part is not perfect.. maybe it should be a Pydantic `Field` or custom `Typed dict`


```Python
y: FormEx[int, Field(title="Horizontal axis")]
# or
y: FormEx[int, P(title="Horizontal axis")]  # P is a typed dict with all defined param attrs
```

adriangb:
Sorry let me clarify: type checkers will never be happy with something like `SomeClass[AType, SomeExtraData]` if `SomeClass` is anything other than `Annotated` (I believe even if you do `from typing import Annotated as SomeClass` they won't like it!).

I didn't even mean to comment on the `dict("Horizontal axis")` part of the proposal. I would highly recommend not throwing random dicts into `Annotated`. Making it a class makes it much easier for tooling to recognize what's in there and avoid confusion or clobbering between multiple users of Annotated.

vitalik:
> Sorry let me clarify: type checkers will never be happy with something like `SomeClass[AType, SomeExtraData]` if `SomeClass` is anything other than `Annotated` (I believe even if you do `from typing import Annotated as SomeClass` they won't like it!).

yeah..  that's what my research concluded - and for that each parameter have and 'Ex` copanion:

```Python
Form[str]
FormEx[str, P(title="foo")]

Body[str]
BodyEx[str, P(title="foo")]

Query
QueryEx
...
```
this fully satisfies mypy

adriangb:
I haven't pulled this branch and tested it but I'm surprised that you got `FormEx[str, P(title="foo")]` to work with type checkers (I say type checkers but in particular I mean mypy *and* pyright), I'll have to look at how you did it!

tiangolo:
Thanks @vitalik!

I would prefer to teach people to use `Annotated` and to have a single way of doing things, if I added this, there would be a lot of things to teach and explain, test, explain the caveats, and explain to people when it would work and when it wouldn't, etc.

At some point it should be possible to use more than one thing inside of `Annotated`, for example Pydantic's `Field()`, and `typer.Option()`, etc. And this would also risk having people just thinking that there's some black magic and not knowing how to use those other things.

About the verbosity, I want to add support for Pydantic models to be used with `Form`, so it would be one single line.

So, I'll pass on this one, but thank for the interest and effort! 🍰 ☕

PR #9943, opened 2023-07-25
✅ Fix tests for compatibility with pydantic 2.1.1

dmontagu:
Pydantic 2.1.1 is not yet released (needs a minor tweak on top of 2.1.0 for fastapi compatibility), but when it is, these tests will fail due to schema generation changes.

I think we really should (on the pydantic side) fix these schemas, specifically the error locs for decimal, but anyway I'm opening this to spare others the hassle of fixing up core schemas that I went through while trying to resolve the incompatibility between fastapi 0.100.0 and pydantic 2.1.0.

@tiangolo feel free to take over this PR, or just steal the commit/changes into a future PR that makes it compatible with the next version of pydantic.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
         id: cache
         with:
           path: ${{ env.pythonLocation }}
-          key: ${{ runner.os }}-python-${{ env.pythonLocation }}-pydantic-v2-${{ hashFiles('pyproject.toml', 'requirements-tests.txt') }}-test-v03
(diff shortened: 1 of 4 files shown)


tiangolo:
Nice, thank you @dmontagu! 🙇 

I updated the pin to allow using 2.1.0 as the only changes are in the error schemas, mainly relevant for the tests, but no internals changed, so I think it's fine to keep supporting 2.1.0 as well, just in case someone needs that.

tiangolo:
Ah! Get it! I just saw another thread on the other discussion. Thanks! I'll revert my change. 🤓

PR #9972, opened 2023-07-31
fix(doc): remove deprecated X- Header recommendation

saroh:
Hi :wave:,

Currently the doc states 
> Have in mind that custom proprietary headers can be added [using the 'X-' prefix](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers).

the MDN doc referred to states 
> Custom proprietary headers have historically been used with an X- prefix, but this convention was deprecated in June 2012 because of the inconveniences it caused when nonstandard fields became standard in [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648)

So in alignment with the RFC I think it's probably better not to state that custom headers should be prefixed by `X-`.

I can "fix" the python example as well. `headers = {"X-Cat-Dog": "alone in the world", "Content-Language": "en-US"}`

### DIFF
--- a/docs/em/docs/advanced/response-headers.md
+++ b/docs/em/docs/advanced/response-headers.md
@@ -37,6 +37,4 @@
 
 ## 🛃 🎚
 
-✔️ 🤯 👈 🛃 © 🎚 💪 🚮 <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers" class="external-link" target="_blank">⚙️ '✖-' 🔡</a>.
-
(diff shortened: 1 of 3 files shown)


tiangolo:
📝 Docs preview for commit 6cccb1ea8315a403e0ab169ca0b03fc3fccc33b6 at: https://64c77bcf03fc330bb0a9b2d9--fastapi.netlify.app

tiangolo:
Thank you! This text was updated in another PR and no longer refers to `X-` headers, so for now I'll pass on this one.

But thanks for the effort! :coffee:

PR #10011, opened 2023-08-04
✨ Enable Pydantic's serialization mode for responses, add support for Pydantic's `computed_field`, better OpenAPI for response models, proper required attributes, better generated clients

tiangolo:
✨ Enable Pydantic's serialization mode for responses, add support for Pydantic's `computed_field`.

Ref: https://docs.pydantic.dev/latest/usage/computed_fields/

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -448,9 +448,7 @@ class APIRoute(routing.Route):
             self.response_field = create_response_field(
                 name=response_name,
                 type_=self.response_model,
-                # TODO: This should actually set mode='serialization', just, that changes the schemas
-                # mode="serialization",
-                mode="validation",
+                mode="serialization",
             )
             # Create a clone of the field, so that a Pydantic submodel is not returned
             # as is just because it's an instance of a subclass of a more limited class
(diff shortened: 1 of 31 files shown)


acnebs:
@tiangolo This change seems to break parts of OpenAPI / JSONSchema generation.

Specifically, fields that should not be "required" in the schema now are.

Simple repro:

```python
from fastapi import FastAPI

app = FastAPI()


class Item(BaseModel):
    name: str
    price: float
    is_offer: None | bool = None


@app.get("/item", response_model=Item)
async def get_item():
    return {
        "name": "Foo",
        "price": 42.0,
        "is_offer": True,
    }
```

In the above, the generated schema for the `Item` model has `is_offer` listed in the `required` field of the JSON schema, even though the [docs for Pydantic 2](https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields) indicate that this should *not* be required, and it was working as expected in FastAPI 0.100.x

Can make a separate issue for this, but thought I'd post here first.

goda-kazuki:
same problem。

PR #10036, opened 2023-08-07
Use field names instead of alias

rbuffat:
We want to create a HTTP endpoint, where the model fields represent the names exposed in the endpoint, whereas we use a model alias that corresponds to identifiers used in our data source. Following is a small example of such a use case:
```
from fastapi import FastAPI
from pydantic import BaseModel, ConfigDict, PrivateAttr


class Test(BaseModel):
    model_config = ConfigDict(
        alias_generator=lambda x: f"db_{x}", populate_by_name=True
    )

    _hidden: str = PrivateAttr()
    field1: str
    field2: str


app = FastAPI()


@app.get("/test/")
async def create_test(item: Test):
    print(item.model_dump())
```

However, this code currently expose the endpoint with the following schema: 

```
{
  "db_field1": "string",
  "db_field2": "string"
}
```

while we intend/expect that the schema would look like this:

```
{
  "field1": "string",
  "field2": "string"
}
```

This PR allows to control if for the schema the alias should be used or not.  With this PR, we could modify our code so that it works as we expect: 

```
from fastapi import FastAPI
from pydantic import BaseModel, ConfigDict, PrivateAttr


class Test(BaseModel):
    model_config = ConfigDict(
        alias_generator=lambda x: f"db_{x}", populate_by_name=True
    )

    _hidden: str = PrivateAttr()
    field1: str
    field2: str


app = FastAPI(by_alias=False)


@app.get("/test/")
async def create_test(item: Test):
    print(item.model_dump())
```


pohmelie:
Faced this too. I think fastapi should work pretty the same way as pydantic `model_dump` method.

YuriiMotov:
Similar to https://github.com/fastapi/fastapi/pull/10621

This way the openapi schema will be invalid

tiangolo:
I suspect what you need is probably Pydantic aliases. In Pydantic v2 there's support for independent validation and serialization aliases.

For now I'll pass on this one but thanks for the interest! :coffee:

PR #10053, opened 2023-08-09
Add Type to Depends return value

noxy19:
Now whenever you use the injection dependency mechanism of fast api you're are getting an Any. This is error prompt.

So instead of that, you can infer the return type given the return type of the provider function. That way the injection mechanism will be more robust.

### DIFF
--- a/fastapi/param_functions.py
+++ b/fastapi/param_functions.py
@@ -549,11 +549,13 @@ def File(  # noqa: N802
     )
 
 
-DependsReturnType = TypeVar('DependsReturnType')
+DependsReturnType = TypeVar("DependsReturnType")
 
 
 def Depends(  # noqa: N802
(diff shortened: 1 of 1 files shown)


uriyyo:
Hi @noxy19,

I guess your PR is similar to mine #5856

noxy19:
@uriyyo yes, why didnt your PR got accepted?

iudeen:
You can do the same like this right?
```
from typing import Annotated

from fastapi import Depends


CommonsDep = Annotated[dict, Depends(common_parameters)]
```

tiangolo:
Thanks for the interest! Now you can (and probably should) use `Annotated`, that should simplify a lot of it.

Also, this would probably have a conflict with dependency classes :thinking: 

Anyway, it's probably better to start in a GitHub Discussion, filling the form, so we can get all the information and the minimal self contained example to understand your use case and be able to help you better.

For now, I'll pass on this one, but thanks for the effort! :coffee:

PR #10058, opened 2023-08-10
Fixed responses model return format to also use serialization mode

zakstucke:
As initially mentioned in https://github.com/tiangolo/fastapi/discussions/10003. 

I discovered even with `0.101.0` this still didn't work when `responses` was used. 
E.g. my example in that discussion still fails if modified such as: 

```Python
import typing as tp
from fastapi import FastAPI

app = FastAPI()

class Foo(BaseModel):
    bar: int
    handler: tp.Callable[[], None] = Field(exclude=True)

@app.post(
    "/check",
    responses={
        200: {"model": Foo},
    } 
)
def validator_check():
    return Foo(handler=lambda: None, bar=1)

app.openapi()
```

Added `mode="serialization"` to the applicable `create_response_field()` and is now working.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -478,7 +478,11 @@ class APIRoute(routing.Route):
                     additional_status_code
                 ), f"Status code {additional_status_code} must not have a response body"
                 response_name = f"Response_{additional_status_code}_{self.unique_id}"
-                response_field = create_response_field(name=response_name, type_=model)
+                response_field = create_response_field(
+                    name=response_name,
+                    type_=model,
+                    mode="serialization",
+                )
                 response_fields[additional_status_code] = response_field
         if response_fields:
             self.response_fields: Dict[Union[int, str], ModelField] = response_fields


yurihartmann:
I have the same problem:

```
from fastapi import FastAPI

app = FastAPI()


class Foo(BaseModel):
    status_code: int = Field(serialization_alias="StatusCode")


@app.get(
    "/",
    response_model=Foo,
    responses={
        400: {"model": Foo},
    },
)
def index():
    return Foo(status_code=1)
```

And the docs appears:
```
200: {
  "StatusCode": 0
}

400: {
  "status_code": 0
}

```

YuriiMotov:
Seems that it works well in current version (without this fix).
@zakstucke, could you check it please?

zakstucke:
@YuriiMotov I opened this PR just under 2 years ago, I worked around this issue a long time ago and I don't have a way to test anymore 😅

YuriiMotov:
Ok. I checked it one more time.
I think we can close this PR as the issue has been resolved.

* Works perfectly fine with Pydantic V2
* With Pydantic V1 it shows 2 warnings, but the produced schema is equal to the schema with Pydantic V2

Schema:
```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "FastAPI",
    "version": "0.1.0"
  },
  "paths": {
    "/check": {
      "post": {
        "summary": "Validator Check",
        "operationId": "validator_check_check_post",
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Foo"
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Foo": {
        "properties": {
          "bar": {
            "type": "integer",
            "title": "Bar"
          }
        },
        "type": "object",
        "required": [
          "bar"
        ],
        "title": "Foo"
      }
    }
  }
}
```

Warnings (with Pydantic V1):
```
fastapi/_compat.py:386: UserWarning: Callable handler was excluded from schema since JSON schema has no equivalent type.
  m_schema, m_definitions, m_nested_models = model_process_schema(
fastapi/_compat.py:466: UserWarning: Callable handler was excluded from schema since JSON schema has no equivalent type.
  return field_schema(  # type: ignore[no-any-return]
```

Looking at code I see that `mode="serialization"` parameter has already been added to the call (at the same time, `create_response_field` was changed to `create_model_field`).

@zakstucke, thank you for your efforts!

PR #10075, opened 2023-08-13
replaced some replaced parameters

bocchi810:


tiangolo:
Hey there @hm9245, I see that the tests are failing, and I don't really see what would be the need for this, or what would it solve, so I'll pass on this one. ☕ 

But if you have questions or have a use case that you are wanting to handle, the best place to start would be a GitHub Discussion question, that will allow me/us to understand better your use case and help you better: https://github.com/tiangolo/fastapi/discussions/new?category=questions

PR #10078, opened 2023-08-14
✨ Add `ResponseValidationError` printable details, to show up in server error logs

tiangolo:
✨ Add `ResponseValidationError` printable details, to show up in server error logs


JonasKs:
Thanks - we just experienced this and it's already fixed 🙏

PR #10109, opened 2023-08-19
Fix handling of custom Pydantic V2 types in parameters

erosennin:
Pydantic V2 allows to [add support for third-party types via `typing.Annotated`](https://docs.pydantic.dev/2.2/usage/types/custom/#handling-third-party-types). Unfortunately, this does not work with FastAPI because [FastAPI removes third-party `typing.Annotated` metadata from parameter type hints](https://github.com/tiangolo/fastapi/blob/a6ae5af7d6c9e7e33490307cccff66a49671433b/fastapi/dependencies/utils.py#L332). 

Example:
```python
# example.py

from typing import Any, Callable, Annotated

from pydantic_core import core_schema
from pydantic import GetJsonSchemaHandler
from pydantic.json_schema import JsonSchemaValue

class ThirdPartyType:
    x: int

    def __init__(self, x: int):
        self.x = x


class _ThirdPartyTypePydanticAnnotation:
    @classmethod
    def __get_pydantic_core_schema__(
        cls, _source_type: Any, _handler: Callable[[Any], core_schema.CoreSchema]
    ) -> core_schema.CoreSchema:
        from_int_schema = core_schema.chain_schema(
            [
                core_schema.int_schema(),
                core_schema.no_info_plain_validator_function(ThirdPartyType),
            ]
        )
        return core_schema.json_or_python_schema(
            json_schema=from_int_schema,
            python_schema=core_schema.union_schema(
                [
                    core_schema.is_instance_schema(ThirdPartyType),
                    from_int_schema,
                ]
            ),
            serialization=core_schema.plain_serializer_function_ser_schema(
                lambda instance: instance.x
            ),
        )

    @classmethod
    def __get_pydantic_json_schema__(
        cls, _core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
    ) -> JsonSchemaValue:
        # Use the same schema that would be used for `int`
        return handler(core_schema.int_schema())


PydanticThirdPartyType = Annotated[ThirdPartyType, _ThirdPartyTypePydanticAnnotation]

app = FastAPI()


@app.get("/")
def echo(x: PydanticThirdPartyType = Query()) -> PydanticThirdPartyType:
    return x
```

```shell
$ uvicorn example:app
...
fastapi.exceptions.FastAPIError: Invalid args for response field! Hint: check that <class 'example.ThirdPartyType'> is a valid Pydantic field type. If you are using a return type annotation that is not a valid Pydantic field (e.g. Union[Response, dict, None]) you can disable generating the response model from the type annotation with the path operation decorator parameter response_model=None. Read more: https://fastapi.tiangolo.com/tutorial/response-model/
```

This PR modifies FastAPI to preserve third-party `typing.Annotated` metadata, so everything works.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -329,12 +329,14 @@ def analyze_param(
         and get_origin(annotation) is Annotated
     ):
         annotated_args = get_args(annotation)
-        type_annotation = annotated_args[0]
-        fastapi_annotations = [
-            arg
-            for arg in annotated_args[1:]
-            if isinstance(arg, (FieldInfo, params.Depends))
-        ]
+        fastapi_annotations = []
+        other_annotations = []
+        for arg in annotated_args[1:]:
+            if isinstance(arg, (FieldInfo, params.Depends)):
+                fastapi_annotations.append(arg)
+            else:
+                other_annotations.append(arg)
+
         assert (
             len(fastapi_annotations) <= 1
         ), f"Cannot specify multiple `Annotated` FastAPI arguments for {param_name!r}"
@@ -355,6 +357,11 @@ def analyze_param(
                 field_info.default = Required
         elif isinstance(fastapi_annotation, params.Depends):
             depends = fastapi_annotation
+
+        if other_annotations:
+            type_annotation = Annotated[(annotated_args[0], *other_annotations)]
+        else:
+            type_annotation = annotated_args[0]
     elif annotation is not inspect.Signature.empty:
         type_annotation = annotation
 
--- a/tests/test_custom_types.py
+++ b/tests/test_custom_types.py
@@ -1,0 +1,161 @@
+import pytest
+from fastapi.testclient import TestClient
+
+from .utils import needs_pydanticv2
+
+
+@pytest.fixture(name="client", params=["default", "annotated"])
+def get_client(request) -> TestClient:
+    from typing import Any, Callable
+
+    from fastapi import FastAPI, Query
+    from pydantic import GetJsonSchemaHandler
+    from pydantic.json_schema import JsonSchemaValue
+    from pydantic_core import core_schema
+    from typing_extensions import Annotated
+
+    class WrappedInt:
+        x: int
+
+        def __init__(self, x: int):
+            self.x = x
+
+    class _WrappedIntPydanticAnnotation:
+        @classmethod
+        def __get_pydantic_core_schema__(
+            cls, _source_type: Any, _handler: Callable[[Any], core_schema.CoreSchema]
+        ) -> core_schema.CoreSchema:
+            from_int_schema = core_schema.chain_schema(
+                [
+                    core_schema.int_schema(),
+                    core_schema.no_info_plain_validator_function(WrappedInt),
+                ]
+            )
+            return core_schema.json_or_python_schema(
+                json_schema=from_int_schema,
+                python_schema=core_schema.union_schema(
+                    [
+                        core_schema.is_instance_schema(WrappedInt),
+                        from_int_schema,
+                    ]
+                ),
+                serialization=core_schema.plain_serializer_function_ser_schema(
+                    lambda instance: instance.x
+                ),
+            )
+
+        @classmethod
+        def __get_pydantic_json_schema__(
+            cls, _core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
+        ) -> JsonSchemaValue:
+            return handler(core_schema.int_schema())
+
+    PydanticWrappedInt = Annotated[WrappedInt, _WrappedIntPydanticAnnotation]
+
+    app = FastAPI()
+
+    param_style = request.param
+    if param_style == "default":
+
+        @app.get("/echo")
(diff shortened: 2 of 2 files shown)


yezz123 (approved):
LGTM!!

alex-pobeditel-2004:
**UPD**: No, looks like this error is not fixed by your patch (tested it locally), so I created a discussion for it: https://github.com/tiangolo/fastapi/discussions/10300

**Original problem**:
I have similar problem with custom type usage, but error is different. Can you please tell if this is the same issue or I need to create a new one? Or I just do something wrong?
  
Here is an example (`fastapi==0.103.1`, `pydantic==2.3.0`):

```
from typing import Annotated

from fastapi import FastAPI
from pydantic import Field


NonNegativeInt32 = Annotated[int, Field(ge=0, le=2_147_483_647)]
app = FastAPI()


@app.get("/{some_number}")
def echo(
        some_number: NonNegativeInt32
) -> NonNegativeInt32:
    return some_number
```

Execution gives this error:  
```
  File "C:\Users\user\test.py", line 13, in <module>
    def echo(                                                                                             
  File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\routing.py", line 704, in decorator
    self.add_api_route(
  File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\routing.py", line 643, in add_api_route
    route = route_class(
  File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\routing.py", line 489, in __init__
    self.dependant = get_dependant(path=self.path_format, call=self.endpoint)
  File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\dependencies\utils.py", line 261, in get_dependant
    type_annotation, depends, param_field = analyze_param(
  File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\dependencies\utils.py", line 410, in analyze_param
    assert isinstance(field_info, params.Path), (
AssertionError: Cannot use `FieldInfo` for path param 'some_number'
```

robo-corg:
> **UPD**: No, looks like this error is not fixed by your patch (tested it locally), so I created a discussion for it: #10300
> 
> **Original problem**: I have similar problem with custom type usage, but error is different. Can you please tell if this is the same issue or I need to create a new one? Or I just do something wrong?
> 
> Here is an example (`fastapi==0.103.1`, `pydantic==2.3.0`):
> 
> ```
> from typing import Annotated
> 
> from fastapi import FastAPI
> from pydantic import Field
> 
> 
> NonNegativeInt32 = Annotated[int, Field(ge=0, le=2_147_483_647)]
> app = FastAPI()
> 
> 
> @app.get("/{some_number}")
> def echo(
>         some_number: NonNegativeInt32
> ) -> NonNegativeInt32:
>     return some_number
> ```
> 
> Execution gives this error:
> 
> ```
> Traceback (most recent call last):
>   File "C:\Users\user\test.py", line 13, in <module>
>     def echo(                                                                                             
>   File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\routing.py", line 704, in decorator
>     self.add_api_route(
>   File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\routing.py", line 643, in add_api_route
>     route = route_class(
>   File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\routing.py", line 489, in __init__
>     self.dependant = get_dependant(path=self.path_format, call=self.endpoint)
>   File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\dependencies\utils.py", line 261, in get_dependant
>     type_annotation, depends, param_field = analyze_param(
>   File "C:\Users\user\.virtualenvs\test-vvPeaeos\lib\site-packages\fastapi\dependencies\utils.py", line 410, in analyze_param
>     assert isinstance(field_info, params.Path), (
> AssertionError: Cannot use `FieldInfo` for path param 'some_number'
> ```

I am still learning pydantic and fastapi but reading the code while debugging this same exact issue (which lead me to this PR) I get the impression that Fields are not allowed as Path parameters. My assumption from reading the code is that some of the options on a Field/FieldInfo would clash with how Path parameters are expected to behave. What exactly doesn't work is not documented in the code and is not clear.

Seems like in addition to the use case in the description of this PR there is a bigger/different issue with Fields + Path being incompatible. Maybe this worked better with pydantic 1 or something?

It would be cool to land this PR to get third party types working since I think without this support you can't have properly type annotated third party types.

alex-pobeditel-2004:
@robo-corg yes, Field annotations can't be used as path parameters now, found it here: https://github.com/tiangolo/fastapi/discussions/10105#discussioncomment-6779426  
So temporary (or persistent) fix is to declare pairs of annotations - one as Field and one as Path, to use them in code separately.

YuriiMotov:
@erosennin, thanks for your interest in FastAPI!

Seems that the situation has been changed since the time this PR was updated last time.
Tests with `Annotated` pass in current version (without this fix), but tests with `=Query()` still fail.
Would you like to review and update it?

PR #10143, opened 2023-08-25
Ability to set `openapi_version` from the constructor.

michaelhenry:
So it can be editable from a configuration upon creating the object rather than setting them afterward.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -59,6 +59,7 @@ class FastAPI(Starlette):
         version: str = "0.1.0",
         openapi_url: Optional[str] = "/openapi.json",
         openapi_tags: Optional[List[Dict[str, Any]]] = None,
+        openapi_version: str = "3.1.0",
         servers: Optional[List[Dict[str, Union[str, Any]]]] = None,
         dependencies: Optional[Sequence[Depends]] = None,
         default_response_class: Type[Response] = Default(JSONResponse),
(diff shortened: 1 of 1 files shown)


iudeen (approved):
LGTM. I think this is cleaner

parker02311 (approved):

yezz123 (approved):
LGTM

Needed!!!

Kludex:
This will be misleading.

Changing the version here will not change the generated schema.

parker02311:
> This will be misleading.
> 
> Changing the version here will not change the generated schema.

Sorry, I didn't look much into this. Is this changing the actual schema version and not the visible version?

Kludex:
I didn't get what you said. The generated schema will be 3.1 compliant, and version on the schema will be the version you set by `openapi_version`. But it doesn't matter the version you select, it will always be compliant with 3.1.

Christian-Schmid:
> I didn't get what you said. The generated schema will be 3.1 compliant, and version on the schema will be the version you set by `openapi_version`. But it doesn't matter the version you select, it will always be compliant with 3.1.

Hi, 

we just stumbled across some issue within the schema generation. I'm a bit confused what the current expected behaviour of fastapi should be when setting this property.

On the [documentation page](https://fastapi.tiangolo.com/how-to/extending-openapi/#the-normal-process) of the property it states the following:

> `openapi_version`: The version of the OpenAPI specification used. By default, the latest: 3.1.0.

This somehow raises my expectation that if I configure here `3.0.3`, that the schema output of the generation would be compliant with that particular open api version. 

But following the discussion here (and also by looking at the actual behaviour of fastapi) It looks for me the only thing this property does is setting the value of the version field in the resulting output. 

Is this correct? 

This would mean fastapi ONLY supports OpenApi 3.1.0 as of today, correct? 

Short answer would be really appreciated :-) 

Thanks
Chris

PR #10158, opened 2023-08-27
 ✨[WIP-Proto] Add ApiEnum

ManiMozaffar:
Please check the design, let me know if it's good enough so that I would finish it + add docs and etc.

I described here that when would we need it
- https://github.com/tiangolo/fastapi/discussions/10156

P.S:
A catchable metadata for endpoints is great! Would be nice if it also support type annotation in some way.
like this

```python
from fastapi import Metadata
from my_module.config import CustomMetadata

def create_something(metadata: CustomMetadata = Metadata()):
     ...
```

### DIFF
--- a/docs_src/dependencies/tutorial005.py
+++ b/docs_src/dependencies/tutorial005.py
@@ -1,6 +1,7 @@
-from typing import Union
 from enum import Enum, auto
-from fastapi import Cookie, Depends, FastAPI, EndPoint
+from typing import Union
+
+from fastapi import Cookie, Depends, EndPoint, FastAPI
 
 app = FastAPI()
 
--- a/docs_src/dependencies/tutorial005.py
+++ b/docs_src/dependencies/tutorial005.py
@@ -1,6 +1,6 @@
 from typing import Union
-
-from fastapi import Cookie, Depends, FastAPI
+from enum import Enum, auto
+from fastapi import Cookie, Depends, FastAPI, EndPoint
 
 app = FastAPI()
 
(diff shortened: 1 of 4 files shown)


tiangolo:
📝 Docs preview for commit 9c72a80253ad5bc7b1e108cff9c14a4dbfccf644 at: https://dc227669.fastapitiangolo.pages.dev

MatsiukMykola:
> > Have similar idea: put to route something like Enum: Create, Read... etc....
> > i use this for quick autotest all endpoints at the test stage to Read endpoints (params plan get from examples) Anyway, think need just like the meta_information param to route for flexible
> 
> Indeed, a catchable metadata for endpoints is great! Would be nice if it also support type annotation in some way.

type: Any ?)

ManiMozaffar:
> > > Have similar idea: put to route something like Enum: Create, Read... etc....
> > > i use this for quick autotest all endpoints at the test stage to Read endpoints (params plan get from examples) Anyway, think need just like the meta_information param to route for flexible
> > 
> > 
> > Indeed, a catchable metadata for endpoints is great! Would be nice if it also support type annotation in some way.
> 
> type: Any ?)

edited my message, any wouldn't give you type annotation.

wu-clan:
This is not possible with fastapi unless Starlette supports it.

ManiMozaffar:
> This is not possible with fastapi unless Starlette supports it.

Not really, I almost implemented it when I was doing it on my project, it almost worked but i was having some weird pydantic erros from somewhere inside the internal of frameworks.
I didn't try to debug it, but at least the ApiStr didn't need any modification to Starlette. I was handling it with caching in router, but I was having a bit of problem when trying to catch it on my dependant creator since it was defined per all api routes, not single

MatsiukMykola:
```python
@router.post(
    path=f"/get_books/{CrudEnum.list.value}",
    name=CrudEnum.list,
    description="Get books",
    response_model=BookListResponseSchema,
)
```

frontend guys, don't worry see path
name - for my internal tasks :(

this UGLY, but  don't see any variant to add some metadata to route
and right now don't understand where to read link (route + function) definition, to extract all examples and run in autotest

MatsiukMykola:
https://github.com/encode/starlette/discussions/2259#discussion-5567579

wu-clan:
This description does make it less clear, could it be simply translated into: how do you add extras to the interface parameters?

Then you can get them via the request object, and then you can implement your permission validation via middleware or Depends

tiangolo:
Thanks for the effort!

Let's start in the GitHub discussion to understand better the use case, it's not clear yet to me, later we can see if there's a need to change something, or maybe there's a way to achieve what you want in some other way.

For now I'll close this one. Let's continue in the discussion. :coffee:

PR #10162, opened 2023-08-28
✏️ Fix typos in `typing_doc` page

pawamoy:


tiangolo:
Good catch, thank you @pawamoy! 🙇

PR #10250, opened 2023-09-15
🐛 Allow async class methods as dependencies not decorated

danielfcollier:
Found this problem: https://github.com/tiangolo/fastapi/discussions/10238

Which happens to be related with this issue: https://github.com/tiangolo/fastapi/issues/10236

I could find the solution for async class methods in not decorated API routes.

Added tests to coverage the cases.

p.s.: still, missing to work with generators.


Kludex:
I think there's an utility function in starlette.utils for this.

danielfcollier:
I've checked **Starlette**, but the test cases for the `add_route` method are much simpler in scope and cases.

This changes I've checked going down with the debugger, I've seen only issues in the **FastAPI** layer. But, later I could check more to see if there is any relation between Starlette and the Generators part.

ordinary-jamie:
Thanks @danielfcollier , can you enumerate the tests that failed / send a link of the CI build -- I will take a look at this and see what I can do.

tiangolo:
Thanks for the interest @danielfcollier! :coffee: 

It seems that the discussion is about some code that is not really supported by Pydantic nor FastAPI (Pydantic models as attributes of another class).

Let's wait to see if there's a reason why that code was like that, or what was the original intention. For now, I'll pass on this one, but thanks for the effort! :cake:

PR #10257, opened 2023-09-17
🐛 Fix schema for `Body` \ `Form` \ `File` fields with `alias`

StefansM:
Addresses #10024.

Without this change, the `alias` parameter of the form fields is disregarded when generating the OpenAPI documentation, though request bodies are still deserialized correctly. The change is copied from [pydantic](https://github.com/pydantic/pydantic/blob/v2.3.0/pydantic/fields.py#L808-L812).


harol97:
check [#10319](https://github.com/tiangolo/fastapi/pull/10319)

fastapi/params.py
```diff
@@ -571,6 +571,12 @@ def __init__(
         else:
             kwargs["deprecated"] = deprecated
         if PYDANTIC_V2:
+            if serialization_alias in (_Unset, None) and isinstance(alias, str):
+                serialization_alias = alias
+
+            if validation_alias in (_Unset, None):
```
YuriiMotov, line 577:
```suggestion
            if validation_alias in (_Unset, None) and isinstance(alias, str):
```

YuriiMotov (changes requested):
@StefansM, thank you for your interest and efforts!

I believe the approach is correct, but I think we need to work on tests a bit more.

This PR actually affects not only `Form` fields, but also single `Body` fields and `File` fields.

```py
@app.post("/body")
def route_with_body(body_param: str = Body(alias="aliased-field", embed=True)):
    return {}
```

PR #10240, opened 2023-09-12
🎨 Fix typing annotation for semi-internal `FastAPI.add_api_route()`

ordinary-jamie:
Closes #10236

Make typing of endpoint parameter consistent with downstream calls and other utilities.

Justification for picking `Callable[..., Any]` over original:

- `FastAPI.add_api_route` makes a direct call to `APIRouter.add_api_route` anyways 
- `fastapi.dependencies.utils.get_typed_return_annotation` has signature `def get_typed_return_annotation(call: Callable[..., Any]) -> Any:`
- A lot of other functions with similar parameters use the same typing (e.g. `FastAPI.add_api_websocket_route`)
- The api router instantiates `route_class: Type[APIRoute]` of which `fastapi.routing.APIRoute` subclasses [`starlette.routing.Route` which also types `endpoint` with `Callable[..., typing.Any]`](https://github.com/encode/starlette/blob/master/starlette/routing.py#L211-L215)

Issue #10236, opened 2023-09-12
Inconsistent add_api_route types


Kludex:
### Discussed in https://github.com/tiangolo/fastapi/discussions/10235

<div type='discussions-op-text'>

<sup>Originally posted by **sidekick-eimantas** September 12, 2023</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/pydantic/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
### Direct mounting of routes
import pathlib

from fastapi import FastAPI, APIRouter
import uvicorn
import pydantic

class HelloRequest(pydantic.BaseModel):
    id: str


class HelloResponse(pydantic.BaseModel):
    hello: str


class Hello:
    def __init__(self, name: str) -> None:
        self.name = name

    async def handle(self, request: HelloRequest) -> HelloResponse:
        return HelloResponse(hello=self.name)

app = FastAPI()
hello_handler = Hello(name="test")
app.add_api_route("/hello", hello_handler.handle, methods=["POST"])






### Mounting of paths via Router
import pathlib

from fastapi import FastAPI, APIRouter
import uvicorn
import pydantic

class HelloRequest(pydantic.BaseModel):
    id: str


class HelloResponse(pydantic.BaseModel):
    hello: str


class Hello:
    def __init__(self, name: str) -> None:
        self.name = name

    async def handle(self, request: HelloRequest) -> HelloResponse:
        return HelloResponse(hello=self.name)


class RootRouter:
    def __init__(self, name: str):
        self.router = APIRouter()
        hello_handler = Hello(name=name)
        self.router.add_api_route("/hello", hello_handler.handle, methods=["POST"])


app = FastAPI()
root_router = RootRouter(name="test")
app.include_router(root_router.router)
```

### Description

First example produces a mypy error:
```
[...]

PR #10258, opened 2023-09-18
Add tests for #10240

danielfcollier:
Hi Jamie from #10240, here is a sample branch with your commit and the minimum required tests for issue #10236.

I've also added some tests for the generator's case, but they are commented to separate the problems.

### DIFF
--- a/tests/test_dependency_class.py
+++ b/tests/test_dependency_class.py
@@ -53,9 +53,19 @@ async def get_callable_dependency(value: str = Depends(callable_dependency)):
     return value
 
 
+app.add_api_route(
+    "/callable-dependency-not-decorated", callable_dependency, methods=["GET"]
+)
+
+
 @app.get("/callable-gen-dependency")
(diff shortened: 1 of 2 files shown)


tests/test_dependency_class.py
```diff
@@ -53,67 +53,148 @@ async def get_callable_dependency(value: str = Depends(callable_dependency)):
...
+
+
 @app.get("/synchronous-method-gen-dependency")
 async def get_synchronous_method_gen_dependency(
     value: str = Depends(methods_dependency.synchronous_gen),
 ):
     return value
 
 
+# app.add_api_route(
```
GDemay, line 120: Is it still usefull to keep it?

tiangolo:
Thanks @danielfcollier :coffee: 

I'm not sure how these changes would test the updated types. :thinking: 

For now, I'll pass on this one, but thanks for the interest. :cake:

PR #10278, opened 2023-09-19
📝 Add OIDC - JWT Access Token validation tutorial

laveolus:
I tried starting a discussion in #10175, but ended up writing a tutorial for OIDC using jose.

It does require some setup with an (external) OIDC provider. I tested with Okta, but I'm not using vendor specific quirks.


davidhuser:
Without knowing this PR or any official docs or implementation guide we created a library that works with a specific Auth Provider (zitadel), using some of the patterns described here. LMK what you think! 

https://github.com/cleanenergyexchange/fastapi-zitadel-auth

laveolus:
> LMK what you think!
It looks well built/tested! Looking back, I also had a hard time provisioning the username/subject/email into the access log context for auditing purposes (not covered in this PR).

Taking a step back, from a FastAPI user perspective, I think they will be having a hard time understanding the Oauth2/OIDC concepts, finding the right vendor-specific library and/or developing things themselves.

YuriiMotov (changes requested):
@laveolus, thanks for your interest and efforts!

This is a good starting point, but I'm afraid it's still a bit rough in its current state.

Here are a few things I think should be improved:
* The style and formatting should be aligned with the rest of the documentation.
* We should focus more on how to set things up on the provider's side.
* Rather than keeping it provider-agnostic, I’d suggest using GitHub as a concrete example (it's just my personal opinion. We should probably ask Sebastian for confirmation).

Would you like to work on this?
If yes, I will discuss this with Sebastian to clarify the right direction before you proceed

dracon80:
I'm a novice, but following these instructions I was able to get OIDC working with Auth0 as the provider.
The problem I'm having is that in the swagger-ui. There are 2 possible OpenIdConnect authentication options. Both OpenIdConnect (OAuth2, authorization_code with PKCE), and OpenIdConnect (OAuth2, implicit). Do I need to manually hide the implicit flow, or should it not be showing since I have specified PKCE is true?

laveolus:
> Do I need to manually hide the implicit flow, or should it not be showing since I have specified PKCE is true?
I remember 2 years ago I could not find a way to filter the auth method at all in the swagger UI. It makes no sense to list them all (and there are many!), because your auth provider configuration controls what's allowed anyway, and it confuses my swagger users (devs, testers).

laveolus:
@YuriiMotov, I'm happy to work on it if I'm working in the right direction  

I'm not sure I'm following the current best practices of the fastapi.security modules. Also, should PyJWT now be used or jose?

tiangolo:
Thanks for the effort @laveolus! :coffee: :rocket: 

As this is a complex topic, I want to give it a thorough review and make sure the right way to approach it. Maybe there's a way to abstract a bit of the complexity too. :thinking: But I have to start by checking it all properly. :nerd_face:

nidhishgajjar:

## Review of PR #10278: Add OIDC - JWT Access Token validation tutorial

This is a well-structured tutorial for a commonly requested topic — OIDC integration with FastAPI. The step-by-step setup guide and working code example are valuable. However, as a **security tutorial**, it needs to demonstrate best practices, and I have some concerns.

### Security Concerns 🔴

1. **`python-jose` is unmaintained**: The `jose` library has known issues and is no longer actively maintained. For a security-focused tutorial, I'd strongly recommend using `PyJWT` instead (or at minimum, calling this out). The FastAPI ecosystem has been moving toward PyJWT. This is especially important because a tutorial sets the example that many developers will copy.

2. **No explicit algorithm specification in `jwt.decode()`**: The `jwt.decode()` call doesn't specify `algorithms=`, which is a security best practice to prevent algorithm confusion attacks. Even if python-jose infers the algorithm from the key, you should explicitly whitelist allowed algorithms (e.g., `algorithms=["RS256"]`):
   ```python
   verified_token = jwt.decode(
       token=access_token,
       key=keyset,
       algorithms=["RS256"],  # Explicitly specify allowed algorithms
       audience=self.audience,
       issuer=self.issuer,
   )
   ```

3. **JWKS URL should come from the well-known configuration**: The `jwks_url` is manually constructed as `f"{settings.issuer}/v1/keys"`, but different OIDC providers may use different paths. Since `oidc_url` already points to `/.well-known/openid-configuration`, the correct approach is to fetch the `jwks_uri` from that well-known config. This is the OIDC-standard way to discover the JWKS endpoint:
   ```python
   # Fetch from well-known config instead of hardcoding
   async with httpx.AsyncClient() as client:
       config = await client.get(oidc_url)
       jwks_url = config.json()["jwks_uri"]
   ```

### Bug / Robustness 🟡

4. **Missing error handling for JWKS fetch**: `get_jwt_keyset()` doesn't handle HTTP errors from the authorization server. If the server is unavailable or returns an error, this will raise an unhandled exception. Consider wrapping in a try/except:
   ```python
   response = await client.get(self.uri)
   response.raise_for_status()
   ```

5. **`# type: ignore` on `__call__`**: The `__call__` method changes the return type from the parent class, hence the `# type: ignore`. This works but could confuse users. Consider adding a brief comment explaining why the type override is intentional.

### Documentation Nits 📝

6. Minor typos in the tutorial markdown:
   - "provder" → "provider" (Step: Setup OpenID provider)
   - "Encryprion" → "Encryption" (Step 4 description)
   - "Write down user/password" → "Write down username/password"

### What looks good ✅

- Clean separation of concerns with `AccessTokenValidator` as a reusable class
- Proper use of `SecurityScopes` for role-based authorization
- Correct HTTP status codes (400 for invalid tokens, 403 for insufficient permissions)
- Good use of `pydantic_settings` for configuration
- TTLCache for JWKS caching is a smart optimization
- Generic error messages that don't leak token validation details
- Well-organized step-by-step setup guide

### Summary

The tutorial structure and code organization are good, but as a security tutorial it should follow current best practices: consider switching from `python-jose` to `PyJWT`, explicitly specify JWT algorithms, and fetch the JWKS URL from the well-known OIDC configuration rather than hardcoding the path.

**Assessment: request-changes** (security concerns around algorithm specification and JWKS discovery)

tiangolo:
I'll pass on this for now, I want to have docs around this, but I would prefer to do it in a different way, and figure out a way to simplify the logic needed by users.

For now, if you want to publish this content, I would suggest making it a blog post that can be shared more easily.

Thanks for understanding and for the efforts here! :coffee:

PR #10294, opened 2023-09-20
docs: update type in oauth2 JWT example

jd-solanki:
When you use `dict.get` you can also get `None` hence updated type from `str` to `str | None`.

### DIFF
--- a/docs_src/security/tutorial004.py
+++ b/docs_src/security/tutorial004.py
@@ -94,7 +94,7 @@ async def get_current_user(token: str = Depends(oauth2_scheme)):
     )
     try:
         payload = jwt.decode(token, SECRET_KEY, algorithms=[ALGORITHM])
-        username: str = payload.get("sub")
+        username: str | None = payload.get("sub")
         if username is None:
             raise credentials_exception
         token_data = TokenData(username=username)
--- a/docs_src/security/tutorial004_an_py39.py
+++ b/docs_src/security/tutorial004_an_py39.py
@@ -94,7 +94,7 @@ async def get_current_user(token: Annotated[str, Depends(oauth2_scheme)]):
     )
     try:
         payload = jwt.decode(token, SECRET_KEY, algorithms=[ALGORITHM])
-        username: str = payload.get("sub")
+        username: str | None = payload.get("sub")
         if username is None:
             raise credentials_exception
         token_data = TokenData(username=username)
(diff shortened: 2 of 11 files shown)


tiangolo:
📝 Docs preview for commit 303d27f8f498307604fea0b6869ca93232f838de at: https://74db433d.fastapitiangolo.pages.dev

docs_src/security/tutorial004.py
```diff
@@ -94,7 +94,7 @@ async def get_current_user(token: str = Depends(oauth2_scheme)):
     )
     try:
         payload = jwt.decode(token, SECRET_KEY, algorithms=[ALGORITHM])
-        username: str = payload.get("sub")
+        username: str | None = payload.get("sub")
```
tiangolo, line 97: This example is for Python 3.8, so it should be `Union[str, None]`

docs_src/security/tutorial004_an_py39.py
```diff
@@ -94,7 +94,7 @@ async def get_current_user(token: Annotated[str, Depends(oauth2_scheme)]):
     )
     try:
         payload = jwt.decode(token, SECRET_KEY, algorithms=[ALGORITHM])
-        username: str = payload.get("sub")
+        username: str | None = payload.get("sub")
```
tiangolo, line 97: This example is for Python 3.9, so it should be `Union[str, None]`

docs_src/security/tutorial004_py310.py
```diff
@@ -93,7 +93,7 @@ async def get_current_user(token: str = Depends(oauth2_scheme)):
     )
     try:
         payload = jwt.decode(token, SECRET_KEY, algorithms=[ALGORITHM])
-        username: str = payload.get("sub")
+        username: str | None = payload.get("sub")
```
tiangolo, line 96: This is fine, as this example is for Python 3.10 🚀

tiangolo (changes requested):
Thanks! 🍰 

The example files are for Python 3.8, 3.9, and 3.10. The syntax of `str | None` is only supported in Python 3.10, for all the others, it should be `Union[str, None]`. Could you change it please? 🤓

tiangolo:
As this PR had requested changes to be applied but has been inactive for a while, it's now going to be closed. But if there's anyone interested, feel free to create a new PR.

Issue #10286, opened 2023-09-20
`Body` related parameters don't take `alias` in consideration


Kludex:
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

### Issue Content

The `Body`, `File`, and `Form` parameters don't support `alias`.

The following snippet demonstrates the issue: 

```py
from typing import Annotated

from fastapi import Body, Cookie, FastAPI, Form, UploadFile, File, Query, Header, Path

app = FastAPI()


@app.post("/{path}")
def endpoint(
    path: Annotated[int, Path(alias="PathAlias")],
    cookie: Annotated[int, Cookie(alias="CookieAlias")],
    header: Annotated[int, Header(alias="HeaderAlias")],
    query: Annotated[int, Query(alias="QueryAlias")],
    body: Annotated[int, Body(alias="BodyAlias")],
    form: Annotated[int, Form(alias="FormAlias")],
    file: Annotated[UploadFile, File(alias="FileAlias")],
):
    ...
```

If you look at the generated Swagger you see that `alias` is only being used for `params.Param` related fields i.e. `Path`, `Cookie`, etc.

<img width="1426" alt="Screenshot 2023-09-20 at 10 04 21" src="https://github.com/tiangolo/fastapi/assets/7353520/f5303787-615b-496b-8406-14f72c3e245e">

People can overcome this right now using `validation_alias` as follows:

```py
from typing import Annotated

from fastapi import Body, Cookie, FastAPI, Form, UploadFile, File, Query, Header, Path

app = FastAPI()


@app.post("/{path}")
def endpoint(
    path: Annotated[int, Path(alias="PathAlias")],
    cookie: Annotated[int, Cookie(alias="CookieAlias")],
    header: Annotated[int, Header(alias="HeaderAlias")],
    query: Annotated[int, Query(alias="QueryAlias")],
    body: Annotated[int, Body(validation_alias="BodyAlias")],
    form: Annotated[int, Form(validation_alias="FormAlias")],
    file: Annotated[UploadFile, File(validation_alias="FileAlias")],
):
    ...
```

In any case, this should be fixed in FastAPI.

harol1997:
**When I use only validation_alias in Form I can see alias in Swagger, but the validation is not working well.
For the moment I'm using both alias and validation_alias, It works.**

I think issue is as FastAPI pass validation_alias to FieldInfo of Pydantic v2 in super().__init__ () inside class Body

[...]

PR #10319, opened 2023-09-25
Fix consider alias in form

harol97:
Addresses [#10024](https://github.com/tiangolo/fastapi/discussions/10024) and [#10286](https://github.com/tiangolo/fastapi/issues/10286)
Now, alias is considered in swagger and redoc docs


Ryandaydev (changes requested):
@harol1997 I verified the behavior before and after this change, and the PR corrects it in both redoc and swagger.

However, I couldn't get the test coverage script to run successfully, and your branch appears to be out of date with master. Do the tests run successfully on your end?

harol1997:
> @harol1997 I verified the behavior before and after this change, and the PR corrects it in both redoc and swagger.
> 
> However, I couldn't get the test coverage script to run successfully, and your branch appears to be out of date with master. Do the tests run successfully on your end?

@Ryandaydev  I already updated repository from master fastapi (original branch)  and I was able to  run test succesfully.

Ryandaydev (approved):
All tests pass now that branch is updated, including new test.
 
I verified the behavior before and after this change, and the PR corrects it in both redoc and swagger.

hruzeda:
any updates on this guys?

harol97:
> any updates on this guys?

For the moment You could use validation_alias and alias at same time .

I hope @tiangolo check it P.R soon

hruzeda:
@harol97, the workaround didn't work for me, am I doing something wrong?

```...
    for route in app.routes:
            # TODO: body params aliases won't work in Swagger - https://github.com/tiangolo/fastapi/issues/10286
            _transform_snake_case(route.dependant.body_params)
            _transform_snake_case(route.dependant.query_params)
            _transform_snake_case(route.dependant.header_params)```

def _transform_snake_case(params: list[ModelField]) -> None:
    for param in params:
        camelized = humps.camelize(param.name)
        param.field_info.alias = camelized
        param.field_info.validation_alias = camelized
```

image

YuriiMotov (changes requested):
I've just run the `test_get_alias` on current FastAPI version (0.109.2) and it works without errors.
The issue isn't covered by test.

The alias actually works, but the problem is that it is not displayed correctly in the json schema (at least in the current version of FastAPI).

We need to add test case that checks JSON schema. I would also remove `test_without_alias` (I bet this case is already covered in other places).

I propose next code:
```

app: FastAPI = FastAPI()


@app.post("/testing_alias")
async def check_alias(id_test: int = Form(alias="otherId")):
    return {"other_id": id_test}


client = TestClient(app)


def test_get_alias():
    response = client.post("/testing_alias", data={"otherId": "1"})
    assert response.json() == {"other_id": 1}


def test_file_alias_schema():
    schema = response.json()

    assert (
        "otherId"
        in schema["components"]["schemas"]["Body_check_alias_testing_alias_post"][
            "properties"
        ]
    )
```
@harol97, can you check it?

harol97:
> I've just run the `test_get_alias` on current FastAPI version (0.109.2) and it works without errors. The issue isn't covered by test.
> 
> The alias actually works, but the problem is that it is not displayed correctly in the json schema (at least in the current version of FastAPI).
> 
> We need to add test case that checks JSON schema. I would also remove `test_without_alias` (I bet this case is already covered in other places).
> 
> I propose next code:
> 
> ```
> from starlette.testclient import TestClient
> 
> from fastapi import FastAPI, Form
> 
> app: FastAPI = FastAPI()
> 
> 
> @app.post("/testing_alias")
> async def check_alias(id_test: int = Form(alias="otherId")):
>     return {"other_id": id_test}
> 
> 
> client = TestClient(app)
> 
> 
> def test_get_alias():
>     response = client.post("/testing_alias", data={"otherId": "1"})
>     assert response.status_code == 200
>     assert response.json() == {"other_id": 1}
> 
> 
> def test_file_alias_schema():
>     response = client.get("/openapi.json")
>     assert response.status_code == 200, response.text
>     schema = response.json()
> 
>     assert (
>         "otherId"
>         in schema["components"]["schemas"]["Body_check_alias_testing_alias_post"][
>             "properties"
>         ]
>     )
> ```
> 
> @harol97, can you check it?

You're right. I changed the test file and I also add support to validation_alias when it's instance of str. validation_alias is priority above alias.

@YuriiMotov  check new changes

Thanks @YuriiMotov

.gitignore
```diff
@@ -28,3 +28,4 @@ archive.zip
 
 # macOS
 .DS_Store
+.python-version
```
YuriiMotov, line 31:
```suggestion
```
Avoid adding unnecessary changes

fastapi/_compat.py
```diff
@@ -282,7 +285,9 @@ def get_missing_field_error(loc: Tuple[str, ...]) -> Dict[str, Any]:
     def create_body_model(
         *, fields: Sequence[ModelField], model_name: str
     ) -> Type[BaseModel]:
-        field_params = {f.name: (f.field_info.annotation, f.field_info) for f in fields}
+        field_params = {
```
YuriiMotov, line 288: I think this is incorrect. We should solve this the way it's done in #10257 (by fixing `serialization_alias` in `field_info`)

fastapi/_compat.py
```diff
@@ -93,7 +93,10 @@ class ModelField:
 
         @property
         def alias(self) -> str:
-            a = self.field_info.alias
+            a = (
+                isinstance(self.field_info.validation_alias, str)
+                and self.field_info.validation_alias
+            ) or self.field_info.alias
```
YuriiMotov, line 99:
```suggestion
        va = self.field_info.validation_alias
        if isinstance(va, str) and va:
            return va
        a = self.field_info.alias
```
I think this way it's easier to read

YuriiMotov (changes requested):
@harol97, thank your for your interest and efforts!

The changes from this PR actually affect much more than just `Form` fields (all kinds of parameters?). So, we should add more tests.

fastapi/_compat.py
harol97: Hello @YuriiMotov . I would have to review what has changed in the latest versions of FastAPI and verify that the error persists. I will close this PR since I believe it has been too long since it was opened, and if I find that this error persists, I will open a new one.

harol97:
Hello @YuriiMotov  . I would have to review what has changed in the latest versions of FastAPI and verify that the error persists. I will close this PR since I believe it has been too long since it was opened, and if I find that this error persists, I will open a new one.

PR #10353, opened 2023-09-29
✨ Support using context managers as a dependency

notpushkin:
## Background

Currently, you can't pass a context manager to `Depends()`, as it would treat it as a function. This means that if you want to use a function both as a dependency in FastAPI and a context manager (e. g. in scripts or Alembic migrations etc), you need to define both versions:

```py
# For FastAPI:
async def get_session():
    async with AsyncSession(engine) as session:
        yield session

# For scripts:
get_session_mgr = asynccontextmanager(get_session)
```

---

Unfortunately, as @a-recknagel pointed out in https://github.com/tiangolo/fastapi/discussions/8955#discussioncomment-5666195, it's impossible to check whether given function is a context manager (defined with a decorator from contextlib) or not:

```py
@asynccontextmanager
async def get_session():
    yield 42

get_session.__aenter__  # => AttributeError: 'function' object has no attribute '__aenter__'
```

It is however possible to check that after we call the function:

```py
called = get_session()  # => <contextlib._AsyncGeneratorContextManager at ...>

if hasattr(called, "__aenter__"):  # => True
    solved = await stack.enter_async_context(called)
else:
    solved = await called
```

This PR implements that check (for async context managers only for now – sync version would require doing some threadpool magic I'm still looking into :-)

# Related discussions
* https://github.com/fastapi/fastapi/discussions/11122
* https://github.com/fastapi/fastapi/discussions/8955
* https://github.com/fastapi/fastapi/discussions/9054


iudeen:
Interesting...

notpushkin:
I think I've figured it out.

Turns out, both sync and async context manager functions are called synchronously, so both of them can be resolved with `run_in_threadpool`.

We can then check if the result has either `__aenter__` or `__enter__` method, and either pass it to `enter_async_context` directly or wrap it with `contextmanager_in_threadpool` first. If it has neither of those attributes, we can pass it down as a value.

I've updated my branch with the above logic implemented. Here's an example app: https://gist.github.com/notpushkin/cfa23e4ab8a73ef9505e2b6e6b2dca5c

Still need to figure out parameters and caching, and add some tests.

notpushkin:
I'm back! Let's get this sorted out. I've rebased on `master`, looking into tests now.

notpushkin:
I've [added a couple basic tests](https://github.com/tiangolo/fastapi/pull/10353/files#diff-b03775a3ac272f60ef13f98a9591fe966225464243ea22ab52a3b27f70d99400) and updated docs to mention that [you can now use context managers directly](https://github.com/tiangolo/fastapi/pull/10353/commits/0ef770c7df6bda6a505e485319bb3c4a73989adb).

I'm not sure how to run [doc tests](https://fastapi.tiangolo.com/contributing/#docs-for-tests) though: when I run `python ./scripts/docs.py live`, I don't see any errors. How do I proceed from here?

tiangolo:
📝 Docs preview for commit d93c701720515479e9e568d9bd663e2804041013 at: https://683c36c4.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit 736f5447080d2610ea62d08846c916204ef83b02 at: https://87d81350.fastapitiangolo.pages.dev

notpushkin:
I've added an ignore rule for the docs, since `DBSession` is an example here and not something we actually use.

Coverage is now failing at 99%, but the lines that aren't covered are in a completely different part of the codebase: https://smokeshow.helpmanual.io/4z4c2y1i1r2j495l4a0i/d_6de190ce15502a91_models_py.html

notpushkin:
@iudeen Would you happen to know what are next steps to get this reviewed and (hopefully!) merged? Sorry if you're not the right person to ask, I just see that you're a Contributor on this repo so I was hoping that you can point me in the right direction :-)

tiangolo:
📝 Docs preview for commit b9bb157988d2aa6f1aaed1efdf3496dbd1683cf2 at: https://e99a602f.fastapitiangolo.pages.dev

notpushkin:
@tiangolo Would love to hear your thoughts about this! Let me know if you'd like any changes made (apart from resolving the conflicts). If this isn't something you think we should merge in at this time, please do let me know, too. 😊

Anton-Aracor:
Conflicts:
image

gsiddhant159:
Any update on this? Would like to see this merged

wssf812:
Any update on this? Would like to see this merged

parikls:
+1 for this.

docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
```diff
@@ -23,6 +23,8 @@ In fact, FastAPI uses those two decorators internally.
 
 ///
 
+    And if you have a context manager already, you can use it as a dependency, too.
```
YuriiMotov:
```suggestion
And if you have a context manager already, you can use it as a dependency, too.
```

YuriiMotov: At least the indentation should be removed.
But I would actually remove this line at all. This is covered in "Context Managers as dependencies" later on

docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
```diff
@@ -240,36 +242,41 @@ When the `with` block finishes, it makes sure to close the file, even if there w
 
 When you create a dependency with `yield`, **FastAPI** will internally create a context manager for it, and combine it with some other related tools.
 
-### Using context managers in dependencies with `yield`
+## Context Managers as dependencies
```
YuriiMotov:
```suggestion
### Context Managers as dependencies
```
I think this should be inside "## Context Managers", right?

docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
```diff
@@ -240,36 +242,41 @@ When the `with` block finishes, it makes sure to close the file, even if there w
...
-/// warning
+You don’t have to create a Context Manager to use FastAPI dependencies.
 
-This is, more or less, an "advanced" idea.
+But sometimes you might want to use a dependency both inside and outside FastAPI. For example, you might want to connect to a database in a DB migration script. You can create a Context Manager manually then:
 
-If you are just starting with **FastAPI** you might want to skip it for now.
+```Python
+{!../../../docs_src/dependencies/tutorial010_ctx.py!}
+```
```
YuriiMotov:
```suggestion
{* ../../docs_src/dependencies/tutorial010_ctx.py *}
```

docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
```diff
@@ -240,36 +242,41 @@ When the `with` block finishes, it makes sure to close the file, even if there w
...
 
-///
+This way you can use it with `Depends()` in your app, and as a regular Context Manager everywhere else:
 
-In Python, you can create Context Managers by <a href="https://docs.python.org/3/reference/datamodel.html#context-managers" class="external-link" target="_blank">creating a class with two methods: `__enter__()` and `__exit__()`</a>.
+```Python
+{!../../../docs_src/dependencies/tutorial010_ctx_usage.py!}
+```
```
YuriiMotov:
```suggestion
{* ../../docs_src/dependencies/tutorial010_ctx_usage.py *}
```

YuriiMotov (changes requested):
A few fixes suggested.

Also (optionally), I would remove the "### Defining Context Managers as classes" section at all. It's not necessary and just increases the complexity of docs.

Lastly, I think we should assemble context manager code example into one file and cover by tests. But I suggest we skip this part for now and do after the general approach is approved by Sebastián.

docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
notpushkin: Yeah, the admonition text indented in the previous version of Mkdocs I think. Good catch!

YuriiMotov:
**Note:** we probably need to add a remark that this is supported since XXX version and for previous versions you should wrap context managers by function with `with` inside

nidhishgajjar:

## Review of PR #10353: Support using context managers as a dependency

This is a great feature addition — allowing context managers to be used directly as dependencies is a natural ergonomics improvement that eliminates boilerplate wrapper functions. The implementation approach is sound and the test coverage is solid.

### Bug in documentation example 🔴

In `docs_src/dependencies/tutorial010_ctx_usage.py`, the `get_db()` function is decorated with `@asynccontextmanager`, but the usage example uses `with` instead of `async with`:

```python
# tutorial010_ctx.py defines:
@asynccontextmanager
async def get_db():
    ...

# tutorial010_ctx_usage.py uses:
async def migrate():
    with get_db() as db:  # ← Bug: should be `async with`
        ...
```

This will raise `RuntimeError: no context manager` at runtime. It should be:
```python
async def migrate():
    async with get_db() as db:
        ...
```

Or alternatively, change `tutorial010_ctx.py` to use `@contextmanager` (sync) if you want to demonstrate a sync context manager usage.

### Implementation observations 🟡

1. **Scope of change**: The PR modifies the `else` branch in `solve_dependencies` (sync, non-generator, non-coroutine callables). This correctly handles sync functions that return context managers. However, async non-generator functions that return context managers (e.g., `async def get_db(): return SomeAsyncCM()`) are not handled — they go through the `is_coroutine_callable` branch and the returned CM is not entered. This is a known limitation and can be addressed in a follow-up, but it's worth documenting.

2. **`contextmanager_in_threadpool` compatibility**: The sync context manager path wraps the result with `contextmanager_in_threadpool()`. This works for generator-based context managers (from `@contextmanager`), but may not work correctly for class-based context managers (with `__enter__`/`__exit__` methods) since `contextmanager_in_threadpool` expects a generator/iterator interface. This is an edge case but worth noting.

3. **Check ordering**: Checking `__aenter__` before `__enter__` is correct — async context managers should take priority.

### What looks good ✅

- **Clean implementation**: The three-way check (async CM → sync CM → plain value) is straightforward
- **Proper exit stack integration**: Using `async_exit_stack.enter_async_context()` ensures cleanup happens correctly
- **Good test coverage**: Tests for both sync and async context managers, verifying both setup and teardown
- **Documentation restructure**: The reorganized docs flow better, introducing the simple case first
- **Technical Details note**: Explaining the internal `__enter__`/`__aenter__` check is helpful

### Summary

Solid feature with good tests and documentation. The one blocking issue is the `with` vs `async with` bug in the documentation example, which would confuse users trying to follow the tutorial.

**Assessment: request-changes** (doc example bug needs fixing)

tiangolo:

Thank you! 🍰

PR #10356, opened 2023-09-30
fix: Annotated's second args not work

Dragon-GCS:
Annotated not work when it annotate a pydantic annotated_types

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -355,6 +355,11 @@ def analyze_param(
                 field_info.default = Required
         elif isinstance(fastapi_annotation, params.Depends):
             depends = fastapi_annotation
+        # When there is no `FieldInfo` or `Depends` in annotated_args,
+        # like Annotated[str, StringConstraints(...)]
+        # we need use the `annotation` as the `type_annotation` to keep the metadata.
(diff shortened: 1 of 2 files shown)


Kludex:
What's the problem this solves?

Dragon-GCS:
Yes, when using `Annoated[type_, ...]` as a type annotation, if the second parameter is not `FieldInfo` or `params.Depends`, but instead is a `param.Path` or other `param.XXX`, it will result in the second parameter being ignored as a metadata attribute.

The following code(`fastapi.dependencies.utils:333`) snippet causes the type annotation to be ignored.
```python
fastapi_annotations = [
    arg
    for arg in annotated_args[1:]
    if isinstance(arg, (FieldInfo, params.Depends))
]
```

And here(`fastapi.dependencies.utils:398`) create a new `param.Path` with `type_annotation` that is the first argument of annotion. 
```python
if is_path_param:
    field_info = params.Path(annotation=type_annotation)
````
So I fix it by replacing `type_annotation` with `annotion` to construct new param when it's second argument is not `FieldInfo` or `params.Depends`

tiangolo:
It seems this was solved at some point, so I'll now close this one, thanks for the effort and interest! :coffee:

PR #10359, opened 2023-09-30
Handle RecursionError in jsonable_encoder when trying to encode an object with circular reference

Abdallah-Darwish:
Hey,

This is my attempt to address [RecursionError in jsonable_encoder when trying to encode an object with circular reference](https://github.com/tiangolo/fastapi/discussions/10358).

Please let me know if I didn't follow any of the repo's conventions or guidelines.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -14,7 +14,7 @@ from ipaddress import (
 from pathlib import Path, PurePath
 from re import Pattern
 from types import GeneratorType
-from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union, Set
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Type, Union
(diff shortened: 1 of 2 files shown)


YuriiMotov (commented):
@Abdallah-Darwish, thanks for your interest and efforts!

Sorry, but I think we don't need this feature.
I haven't seen anybody else requested it. And I don't think it's worth adding additional complexity to codebase.

Also, I looked at the implementation and it seems that it could be optimized (in terms of number of function created).

Let's wait for Sebastián to review this and decide whether we need it or not.

tiangolo:
Thanks for the interest! I think this would add some significant complexity, and that always has the potential to break some other corner cases. And the end result is mainly switching one exception for another, both resulting in a server error. So I would prefer not to take it. But thanks for the interest! :cake:

PR #10372, opened 2023-10-03
New param to take a token from headers or cookies

taskevich:
It is not very convenient to contain the token only in the headers. I thought why not get a token from cookies


iudeen:
I might be wrong, but I think having Authorization and Bearer in header is the standard. 

You can always extend these classes and modify it to meet your needs.

PR #10376, opened 2023-10-03
📝 Reword in docs, from "have in mind" to "keep in mind"

malicious:
"Keep in mind" is how you would express this in an imperative form.

First pass was done on macOS 10.15 with:

    LC_ALL=C find docs -type f -exec sed -i '' 's/Have in mind/Keep in mind/g' '{}' +
    LC_ALL=C find docs -type f -exec sed -i '' 's/have in mind/keep in mind/g' '{}' +
    git restore docs/en/docs/img


tiangolo:
📝 Docs preview for commit 5805b1051464b2b9a0824243535fe43b251141d8 at: https://75293bc4.fastapitiangolo.pages.dev

gregmuellegger:
As a non-native speaker I wondered if `Have in mind` is wrong, or what the difference is between the two.

The definition of [have in mind](https://idioms.thefreedictionary.com/have+in+mind):
> To be considering someone or something as a candidate or possibility for something. In this usage, a noun or pronoun can be used between "have" and "in."

The definition of [keep in mind](https://idioms.thefreedictionary.com/keep+in+mind):
> Used to introduce something that one wants to emphasize or be remembered.

So I think both versions are valid, but the later fits the situations it is used in the documentation better.

Another random [discussion in a language forum](https://forum.wordreference.com/threads/have-keep-bear-in-mind-that.3152130/):
> I think there is no difference in meaning between the three: "While doing AAA, (have/keep/bear) in mind BBB." All of them say "for the period of time you are doing AAA, **also think about** BBB, because it affects your understanding of AAA." All are good AE usage.  
> 
> **bear in mind** is the most common. Here **bear** literally means **carry**. It is easy to imagine **carrying** BBB while doing AAA.
> 
> **keep in mind** is also common in this use. It has the concept of **holding** BBB while doing AAA.
> 
> **have in mind** is okay, but it could be a little bit confusing, only because **to have something in mind** (meaning "to already have your own ideas about the current topic") is such a common phrase.

Based on that information, I support that change.

gregmuellegger (changes requested):
All proposed changes make sense.

However there are places left in the code that the regex you used didn't catch. I found them by searching for `in mind`:

- https://github.com/malicious/fastapi/blob/e09646441d67ea59e2daece896d027dc5a067815/docs/en/docs/benchmarks.md#L5
- https://github.com/malicious/fastapi/blob/5805b1051464b2b9a0824243535fe43b251141d8/docs/en/docs/deployment/manually.md#L17
- https://github.com/malicious/fastapi/blob/0976185af96ab2ee39c949c0456be616b01f8669/docs/en/docs/tutorial/sql-databases.md#L304

tiangolo (changes requested):
Thanks @malicious! 🤓 

And thanks for the review @gregmuellegger 🔍 

@malicious, would you please implement the suggestions from @gregmuellegger?

tiangolo:
📝 Docs preview for commit d9f8b017eabc2c4a7f854c9e765e664c52a591bc at: https://6975e493.fastapitiangolo.pages.dev

alejsdev:
Thank you! @malicious 🤓 

And thanks for the help too @gregmuellegger 😸

PR #10392, opened 2023-10-05
✨ Add reference (code API) docs with PEP 727, add subclass with custom docstrings for `BackgroundTasks`, refactor docs structure

tiangolo:
✨ Add reference (code API) docs with PEP 727, add subclass with custom docstrings for `BackgroundTasks`, refactor docs structure.


tiangolo:
📝 Docs preview for commit df4c501136c76a2ef83e3c7e8330c15b5f84491b at: https://5419df85.fastapitiangolo.pages.dev

pawamoy:
If you install Black alongside mkdocstrings-python, it will use it to format long signatures (such as `fastapi.FastAPI`).
Also, tables are a bit too large and have an horizontal scrollbar, maybe the `list` or `spacy` formats for docstring sections would be more readable? Happy to work with you on improving rendering :slightly_smiling_face:

pawamoy:
There's no such option, but I'll see how we can make it happen.

tiangolo:
Thanks @pawamoy!

> Also I suppose you cannot use `from __future__ import annotations` (and then use PEP 604's | syntax) for Pydantic/runtime reasons?

I'm not fully sure, I might be able to do that, at least in some places. It's funny and ironic that I would recommend people not to use `from __future__ import annotations` to not break/complicate their FastAPI and Pydantic apps... but I might actually be able to use it at least in some places (Pydantic already does that as well). 😅  I will try that as well.

pawamoy:
As long as you know what you're doing :wink: And it's only for 1 year and 11 months, until 3.9 is EOL :shushing_face:

fastapi/applications.py
```diff
@@ -43,59 +43,608 @@
...
+                """
+                Extra keyword arguments to be stored by FastAPI, not used anywhere.
+                """
+            ),
+        ],
     ) -> None:
-        self.debug = debug
+        """
+        ---
+        """
```
tiangolo, line 646: @pawamoy I'm noticing that to render the `__init__()` parameters I need to have some non-empty docstring in the `__init__()` method, even if there's already a class docstring...

Am I doing something wrong in the configs or is it expected?

pawamoy: I noticed that in the PR yes. That's probably an oversight, I'll take a look :)

tiangolo:
📝 Docs preview for commit f55ce131307dc1605a945604d55b8467aadd2536 at: https://bfaa233b.fastapitiangolo.pages.dev

tiangolo:
@pawamoy am I using the filters correctly? I have tried several settings, but I still see that when I add things from Starlette, it always shows everything, including private attributes, e.g.: https://bfaa233b.fastapitiangolo.pages.dev/reference/testclient/#fastapi.testclient.TestClient._portal_factory

Is there some additional config I have to add? Or something I should add to those specific references?

tiangolo:
📝 Docs preview for commit c9a2fb35ca46d53fc82d0f2c91ae437a18c05cbd at: https://7adbf1fe.fastapitiangolo.pages.dev

fastapi/applications.py
```diff
@@ -994,68 +1322,2064 @@ def decorator(func: DecoratedCallable) -> DecoratedCallable:
...
+                Read more about it in the
+                [FastAPI docs for Dependencies in path operation decorators](https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-in-path-operation-decorators/).
+                """
+            ),
+        ] = None,
+        summary: Annotated[
+            Optional[str],
+            Doc(
+                """
+                An summary for the *path operation*.
```
pawamoy, line 1990:
```suggestion
                A summary for the *path operation*.
```

More occurrences below, I'll let you regex replace or something :slightly_smiling_face:

fastapi/applications.py
```diff
@@ -994,68 +1322,2064 @@ def decorator(func: DecoratedCallable) -> DecoratedCallable:
...
+                Read more about it in the
+                [FastAPI docs for Path Operation Configuration](https://fastapi.tiangolo.com/tutorial/path-operation-configuration/).
+                """
+            ),
+        ] = None,
+        description: Annotated[
+            Optional[str],
+            Doc(
+                """
+                An description for the *path operation*.
```
pawamoy, line 2003:
```suggestion
                A description for the *path operation*.
```

More occurrences below, I'll let you regex replace or something :slightly_smiling_face:

fastapi/param_functions.py
```diff
@@ -323,51 +1277,302 @@ def Cookie(  # noqa: N802
...
+            """
+        ),
+    ] = None,
     regex: Annotated[
         Optional[str],
+        Doc(
+            """
+            RegEx pattern for strings.
+
+            "Deprecated in FastAPI 0.100.0 and Pydantic v2, use `pattern` instead."
```
pawamoy, line 1445: Extraneous quotes probably? More occurrences below.

pawamoy (commented):
Noticed a few typo

pawamoy:
> @pawamoy am I using the filters correctly?

Hmm that's curious indeed. I'm able to reproduce locally, will continue to investigate.

pawamoy:
Ah, it's because of the global `members: true` which takes precedence over filters. But then without it, nothing shows up under `TestClient` :thinking: Continuing investigation!

pawamoy:
OK, everything works as expected:

- `members: true` was taking precedence over `filters: ["!^_"]`, showing every member of the test client
- without `members: true`, filters are applied, as well as the docstring filter: any object without a docstring is filtered out (see https://mkdocstrings.github.io/python/usage/configuration/docstrings/#show_if_no_docstring)

The result is: no members shown.

But if we look at the test client class, it inherits from `httpx.Client` which does have public objects with docstrings: to select those, you must enable [inherited members](https://mkdocstrings.github.io/python/usage/configuration/members/#inherited_members), and also [preload](https://mkdocstrings.github.io/python/usage/configuration/general/#preload_modules) the `httpx` and `starlette` modules. Example:

```
::: fastapi.testclient.TestClient
    options:
        preload_modules: [httpx, starlette]
        inherited_members: true
```

These two options can also be defined globally, in mkdocs.yml. Inherited members will then affect everything, it might not be what you want. Preloading modules, while being both a global and local option, really only affects the global context, so it's just a matter of preference: whether you find it more explicit to define it in mkdocs.yml or under the objects that need it.

Don't forget to remove `members: true` from mkdocs.yml :slightly_smiling_face:

pawamoy:
Regarding deprecated objects: I see that you sometimes duplicate the message of `deprecated` in the docstring. It's not strictly necessary since `griffe-typingdoc` understands `deprecated` too in `Annotated` and will insert the message in the docstring, leading to it being duplicated in the rendered docs:

depr

Instead you could write the message in `deprecated` only, with proper markup (Markdown), and let it be inserted at the top of the docstring. I do think we have to improve the rendering though (in the screenshot: no newline after deprecation message, no styling for the message).

For objects deprecated with the `@deprecated` decorator, you could either use the `griffe-typing-deprecated` extension (which is an Insiders project), or create a custom Griffe extension, for example:

```python
# griffe_extensions.py
"""Griffe extensions."""

import ast
from typing import TYPE_CHECKING
import textwrap

from griffe import Docstring, Extension
from griffe.docstrings.dataclasses import DocstringSectionAdmonition, DocstringSectionText

if TYPE_CHECKING:
    from griffe import Class, Function, ObjectNode


def _deprecated(obj: Class | Function) -> str | None:
    for decorator in obj.decorators:
        if decorator.callable_path in {"typing.deprecated", "typing_extensions.deprecated"}:
            return ast.literal_eval(str(decorator.value.arguments[0]))
    return None


class DeprecatedExtension(Extension):
    """Griffe extension for `@typing.deprecated` decorator support."""

    def _insert_message(self, obj: Function | Class, message: str) -> None:
        if not obj.docstring:
            obj.docstring = Docstring("", parent=obj)
        sections = obj.docstring.parsed
        sections.insert(0, DocstringSectionAdmonition(kind="warning", text=message, title="Deprecated"))

    def on_class_instance(self, node: ast.AST | ObjectNode, cls: Class) -> None:
        """Add section to docstrings of deprecated classes."""
        if message := _deprecated(cls):
            self._insert_message(cls, message)
            cls.labels.add("deprecated")

    def on_function_instance(self, node: ast.AST | ObjectNode, func: Function) -> None:
        """Add section to docstrings of deprecated functions."""
        if message := _deprecated(func):
            self._insert_message(func, message)
            func.labels.add("deprecated")
```

```yaml
plugins:
- mkdocstrings:
    handlers:
      python:
        options:
          extensions:
          - path/to/griffe_extensions.py
```

fastapi/applications.py
pawamoy: This is now fixed by latest versions of Griffe and griffe-typingdoc :smile:

fastapi/applications.py
tiangolo: Awesome, thanks!

fastapi/applications.py
tiangolo: Good catch, thank you!

fastapi/param_functions.py
tiangolo: Yep! Thanks! 🔍

tiangolo:
Amazing, thank you @pawamoy! 🙇 

I updated the configs with your suggestions! 🚀  Also thanks for the typos.

> I see that you sometimes duplicate the message of deprecated in the docstring. It's not strictly necessary since griffe-typingdoc understands deprecated too in Annotated and will insert the message in the docstring

Awesome! I was thinking that I could duplicate it there too for editors, but editors still don't support `Doc()` so it's the same... 😅  I'll remove those duplications!

> For objects deprecated with the @deprecated decorator, you could either use the griffe-typing-deprecated extension (which is an Insiders project)

Nice! I was trying to find how to install insiders the other day but I didn't find it... did I miss the docs somewhere?

tiangolo:
📝 Docs preview for commit d7b1e1b49d58bc0f7387c5bbff6a475f9ac76b37 at: https://dca072ea.fastapitiangolo.pages.dev

wu-clan:
Another amazing job!👍

fastapi/applications.py
tiangolo: Thank you! ...although I can't seem to make it work yet 🤔

tiangolo:
📝 Docs preview for commit a6b92eac714011535790617a454199ca2f6ed37b at: https://5797c3d1.fastapitiangolo.pages.dev

fastapi/applications.py
pawamoy: Can you tell me what is not working? I tried locally to remove the docstring from `fastapi.FastAPI.__init__` and the parameters are correctly inserted into the class docs.

pawamoy:
Found another bug: `Source code in` not displaying the actual file path. A property was missing in Griffe, pushing a bugfix release right now.

tiangolo:
Thanks @pawamoy! Let me know, I think this is pretty much ready. 😎

tiangolo:
📝 Docs preview for commit ba112c39c33742a24306987d5877dc08dfd31583 at: https://3dd93525.fastapitiangolo.pages.dev

pawamoy:
> I was trying to find how to install insiders the other day but I didn't find it... did I miss the docs somewhere?

It's documented here: https://mkdocstrings.github.io/griffe-typing-deprecated/insiders/installation/. I wonder if this extension still makes sense and if we shouldn't just merge it in the typing-doc one.

pawamoy:
Griffe 0.36.7 is out :slightly_smiling_face:

tiangolo:
Awesome, thanks @pawamoy! 🚀 

For some reason, I hadn't found the insiders installation, and now I see that it's quite open there... I was probably too tired and didn't see it 😅 🙈

tiangolo:
📝 Docs preview for commit e2c36729e1a7712a5d1678c5e2faa4acf1670307 at: https://690c34e1.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit c34b8c9b8277f105eba63ff21aee75e84b5453f8 at: https://eca9a21b.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit 77cb0c7d6ec09ad2899e148231af5bd531662af3 at: https://ecb1abc9.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit ac82b9ed29894a2955ce723b91c8da8dc76b1125 at: https://2124dde5.fastapitiangolo.pages.dev

pawamoy:
I'll tell you when griffe-typing-deprecated is up-to-date with latest Griffe so you can list it in mkdocs.yml :slightly_smiling_face:

tiangolo:
Thanks @pawamoy! For now, I'll merge and release this. 😎 🚀

pawamoy:
Congrats! I'm amazed by your efficiency here: in just a few days you managed to document the (almost?) entire API :exploding_head:

pawamoy:
Oh and for `signature_crossrefs` to take effect, you need to enable `show_signature_annotations` too!

PR #10393, opened 2023-10-06
Fix _compat.is_uploadfile_sequence_annotation

rigens:
We are using `BaseCollectionModel` class from `pydantic-collection` package as route parameters.
After updating to the latest FastAPI version it is not automatically recognized as body parameters.
We have determined that the reason is the incorrect implementation of the `_compat.is_uploadfile_sequence_annotation` function.

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -623,7 +623,8 @@ def is_uploadfile_sequence_annotation(annotation: Any) -> bool:
                 at_least_one = True
                 continue
         return at_least_one
-    return field_annotation_is_sequence(annotation) and all(
+    args = get_args(annotation)
+    return field_annotation_is_sequence(annotation) and args and all(
         is_uploadfile_or_nonable_uploadfile_annotation(sub_annotation)
-        for sub_annotation in get_args(annotation)
+        for sub_annotation in args
     )
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -613,7 +613,6 @@ def is_bytes_sequence_annotation(annotation: Any) -> bool:
         for sub_annotation in get_args(annotation)
     )
 
-
 def is_uploadfile_sequence_annotation(annotation: Any) -> bool:
     origin = get_origin(annotation)
     if origin is Union or origin is UnionType:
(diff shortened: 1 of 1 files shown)


rigens:
Just now `is_uploadfile_sequence_annotation` returns True for builtin `list` type (not generic). This is an obvious bug.

romis2012:
We have the same problem, please fix the mentioned issue

tiangolo (changes requested):
Thanks @rigens! I'm actually not away of the package you mention `pydantic-collection`, not sure why it would be needed. :thinking: 

It would probably be better to start the conversation in a Discussion, that way you can share the exact problem, with code that we can copy paste to understand and replicate your problem.

And then, the PR should have a test that covers your use case and breaks with master, but that passes with your change.

Still, probably better to start in a Discussion. :coffee:

rigens:
@tiangolo, the issue is described above.

code example

```python
import uvicorn
from pydantic_collections import BaseCollectionModel


class SomeItem(BaseModel):
    some_attr: str


class SomeCollection(BaseCollectionModel[SomeItem]):
    pass


app = FastAPI()


@app.post('/post-items')
async def post_items(items: SomeCollection):
    return items


uvicorn.run(
    app,  # type: ignore
    host='127.0.0.1',
    port=8088,
)
```

_Note that `SomeCollection` is a subclass of `pydantic.BaseModel` and should be treated as `Body` by default._ But FastAPI treats it as `Form`.

Request:
```
POST http://localhost:8088/post-items

[{"some_attr": "some_val1"}, {"some_attr": "some_val2"}]
```

Response:
```
HTTP/1.1 422 Unprocessable Entity
...
{
  "detail": [
    {
      "type": "missing",
      "loc": [
        "body",
        "items"
      ],
      "msg": "Field required",
      "input": null
    }
  ]
}
```

This PR fixes the issue (and other potential issues)

rigens:
```python
from fastapi._compat import is_uploadfile_sequence_annotation
assert not is_uploadfile_sequence_annotation(list), "built-in list type is not a collection of a File"
```

```pytb
  File "/home/rigens/projects/test_scripts/fast_api_pydantic_collection.py", line 8, in <module>
    assert not is_uploadfile_sequence_annotation(list), "built-in list type is not a collection of a File"
AssertionError: built-in list type is not a collection of a File
```

@tiangolo, just fix this error please

rigens:
Discussion created: https://github.com/fastapi/fastapi/discussions/12038

tiangolo:
Thanks for the feedback and for creating the discussion!

The library you mention isn't officially supported by FastAPI.

In FastAPI you can use regular Pydantic type annotations for fields, you are not restricted to Pydantic models, so you can do things like:

```Python
from fastapi import FastAPI

app = FastAPI()


@app.post("/post-strings")
async def post_strings(strings: list[str]):
    return strings
```

And if you need to validate and/or customize things further, you can use Pydantic validators: https://docs.pydantic.dev/latest/concepts/validators/

`RootModels` are not really necessary, that's why they are not officially supported or documented.

But let me know if there's a reason why regular type annotations with validators wouldn't work in your use case.

Also, if this PR solves any other issue, please create a test for it using the types that are supported (documented) in FastAPI.

PR #10406, opened 2023-10-07
add jwt signature expiry exception

Sheldenburg:
Hi, 

I was reading this tutorial 
https://fastapi.tiangolo.com/tutorial/security/oauth2-jwt/#__tabbed_3_1

But was a bit confused whether jose jwt handles the token expiry itself. But I think the source code confirms this itself. 
https://github.com/mpdavis/python-jose/blob/96474ecfb6ad3ce16f41b0814ab5126d58725e2a/jose/jwt.py#L82

And this probably needs to be caught in a separate exception.

### DIFF
--- a/docs_src/security/tutorial004.py
+++ b/docs_src/security/tutorial004.py
@@ -98,7 +98,7 @@ async def get_current_user(token: str = Depends(oauth2_scheme)):
         if username is None:
             raise credentials_exception
         token_data = TokenData(username=username)
-    except ExpiredSignatureError: # Check signature expiry
+    except ExpiredSignatureError:  # Check signature expiry
         raise credentials_exception
     except JWTError:
         raise credentials_exception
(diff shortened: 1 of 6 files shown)


tiangolo:
📝 Docs preview for commit fe216eb467dcc27ef2759641e4439ba2e1a85304 at: https://7c260a54.fastapitiangolo.pages.dev

giladfuchs:
Hi @Sheldenburg I just wondering 
why not separate the detail for proper message instead or

credentials_exception -> Could not validate credentials
signature_expired_exception  -> signature has expired

so the client gets a more specific reason.

docs_src/security/tutorial004.py
```diff
@@ -98,6 +98,8 @@ async def get_current_user(token: str = Depends(oauth2_scheme)):
         if username is None:
             raise credentials_exception
         token_data = TokenData(username=username)
+    except ExpiredSignatureError:  # Check signature expiry
+        raise credentials_exception
```
Ale-Cas, line 102: I would modify to:
```python
except ExpiredSignatureError as expired_signature:  # Check signature expiry
        raise credentials_exception from expired_signature
```
and/or follow @giladfuchs [suggestion](https://github.com/tiangolo/fastapi/pull/10406#issuecomment-1766126432) and raise a different exception in this case

tiangolo:
Thank you! The tutorials now use PyJWT, so this probably no longer applies. Given that I'll close this one, but thanks for the effort! :coffee:

PR #10417, opened 2023-10-10
Documentation uses deprecated pydantic.utils.GetterDict

fredrikaverpil:
### Why this change?

When following [the current documentation on making the Peewee ORM work with FastAPI](https://fastapi.tiangolo.com/how-to/sql-databases-peewee/), the instructions are using the deprecated `pydantic.utils.GetterDict`.

This leads to this kind of deprecation warning in end-user apps that uses Pydantic v2:

```
/opt/venv/lib/python3.11/site-packages/pydantic/_migration.py:288: UserWarning: `pydantic.utils:GetterDict` has been removed. We are importing from `pydantic.v1.utils:GetterDict` instead.See the migration guide for more details: https://docs.pydantic.dev/latest/migration/
```

The migration guide mentioned: https://docs.pydantic.dev/latest/migration/

### What was changed?

Encourage use of `pydantic.v1.utils.GetterDict` instead, as this does not produce the deprecation warning.


tiangolo:
📝 Docs preview for commit 73e6e3f51a4fd514946a2c1334b3052f1fabc817 at: https://b16bee40.fastapitiangolo.pages.dev

Ale-Cas (approved):

tiangolo:
Thanks! That tutorial has a mark at the top making it explicit that it only supports Pydantic v1.

In fact, I'm actually gonna remove it soon, it doesn't really make much sense to maintain it given there are much better alternatives, so it doesn't make much sense to keep it around either. 😅 

PR #10494, opened 2023-10-21
✨ Allow empty FormData fields

owjs3901:
Regarding FormData, during development.
I saw that FormData without a key and a key present but empty data were treated the same.

However, these two are completely different, and related options exist in swagger as well.
So, This problem needs to be resolved.

The two cases below should be different, but are currently the same.

image
image

[FormData swagger option]

image

image


dtleal (approved):

dtleal (approved):

owjs3901:
@tiangolo 
I don't have any major complaints, but I'm curious as to why the merge isn't working.

YuriiMotov:
This PR is outdated.

The behavior changed since the time this PR was updated last time.
Now FastAPI treats absence of parameter as `None` and empty value as empty string

owjs3901:
would I close this pr?

YuriiMotov:
> would I close this pr?

Please, check if I got it right. If you confirm that this PR is not relevant anymore, then yes, we should close it to reduce the amount of PRs to review.

Thanks for your efforts!

PR #10509, opened 2023-10-24
Update _compat.py

bmoore:
This is to fix an issue where having list[Model] with response_model_include doesn't return an object.


bmoore:
Closing this according to deep documentation in 

https://docs.pydantic.dev/2.4/concepts/serialization/#advanced-include-and-exclude

# To exclude a field from all members of a nested list or tuple, use "__all__":
print(user.model_dump(exclude={'hobbies': {'__all__': {'info'}}}))

PR #10553, opened 2023-10-30
Use typing.Optional instead of typing.Union in docs

droctothorpe:
This is a tiny, admittedly pedantic PR to use the more succinct `Optional[<type>] = None` over `Union[<type>, None] = None` syntax in the docs, since:

> Optional[X] is equivalent to X | None (or Union[X, None]).
- [Source](https://docs.python.org/3/library/typing.html#typing.Optional)

I thought that maybe `Union` was a deliberate choice but saw that `Optional` was used liberally throughout the codebase. I made the change in all language-specific files since the code blocks are identical.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ $ pip install "uvicorn[standard]"
 * Create a file `main.py` with:
 
 ```Python
-from typing import Union
+from typing import Optional
 
 from fastapi import FastAPI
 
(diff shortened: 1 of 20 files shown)


nilslindemann:
Hello @droctothorpe, check this out: [Using Union or Optional](https://fastapi.tiangolo.com/python-types/#using-union-or-optional)

tiangolo:
Thanks for the help @nilslindemann! It's very much appreciated. 🙇 

Yep, exactly @droctothorpe, it's intentional. 😅  ...I actually spent a long time replacing `Optional` with `Union` in the docs a while ago.

droctothorpe:
Thanks, ya'll. Agree that `Optional` overloads the common-sense definition of the term in a way that's counter-intuitive.

PR #10546, opened 2023-10-29
⚡ Decrease import time with deferred OpenAPI model building

sam-mosleh:
This pull request introduces a performance improvement for importing FastAPI, with the potential for up to a 40% reduction in import times. This optimization is particularly crucial for environments like AWS Lambda, where OpenAPI schemas are typically not required to get built during cold starts. The key enhancement in this PR is the utilization of the `defer_build` option in Pydantic, which has been available since version 2.3.0.

### Performance Difference:
To quantify the impact of these changes, performance measurements were taken:

- **Baseline Version:**
  - Command: `python -X importtime -c "import fastapi"`
  - Import Time: 608,727 microseconds
- **Optimized Version:**
  - Command: `FASTAPI_OPENAPI_DEFER_BUILD=true python -X importtime -c "import fastapi"`
  - Import Time: 383,697 microseconds

PR #10589, opened 2023-11-04
⚡️ Ensure faster start-up when using many nested routers

jvllmr:
First of all: Thanks a lot for your on this project! I never thought that Python Web Development (especially for back-ends) could be this much more fun.

Now to the main part: In my FastAPI project I have a lot of nested routers. Because the amount of nested routers I soon reached a FastAPI start-up time of 30 seconds or more. This hurt my developer experience a lot and I soon began digging in to why FastAPI takes so long to load. 

After some research I found out that each router calculates all relevant infos on initialization. A router, which includes another router, will re-calculate these values without re-using information from the router it includes. When having a flat project structure this is not a problem. 

So the solution was simple: Defer the calculation of values until they are actually needed. By doing that the start-up time was shortened significantly. 

One caveat remains: Deferring the calculation results to Exceptions not being raised at start-up and they are only raised at the first call of a route. So they might not be caught if a route is not tested at least once.

This pull request is my proposal to integrate this performance optimization into FastAPI without breaking user space and having to modify tests. I mitigated the Exception's issue by introducing an argument to the APIRouter and APIRoute which decides if values are deferred or not and never deferring values in the top-level (app-integrated) router.

I'm sure there might be a better way. So let me hear if you have suggestions for improvement!

You can find my research and testing environment for this here: https://github.com/jvllmr/fastapi-deferred-init


Jamim:
Thanks you, @jvllmr!
This PR and #10546 might perfectly complement each other 🚀

jvllmr:
I refactored the pull-request according to the latest changes.

In the meantime the performance improved already (probably because of changes in Pydantic). In my benchmark the median improvement was around 8 to 10x. My benchmark is very simple and there might be more improvement in real world applications.

The tests are failing as of now. I will look into it as soon as I can.

svlandeg:
Hi @jvllmr!

> The tests are failing as of now. I will look into it as soon as I can.

There was an unrelated issue with our CI that is now fixed, so you shouldn't see so many failing tests anymore.

The only test still failing here is the one measuring "coverage" - this basically ensures that all code paths in the code base are tested in our test suite. So, to fix it, it would be good to add a test with this PR that checks the correct behaviour of the new functionality introduced here 🚀 

Specifically, when I look at the detailed coverage report, it's the `init_routes` function that isn't getting properly tested: https://smokeshow.helpmanual.io/6t005p5s2v3j6u064g0q/z_35ba7d295f00afa4_routing_py.html#t1160

jvllmr:
Thanks for the hint. I will see if I can come up with some good tests over the weekend.

jvllmr:
I updated the PR with tests for all the new logic I introduced, but I don't know why 99.0% coverage are still shown here.
It was green at first but then went back to red. When looking at the detailed report it still says 100%.
I'm also sorry for the messed up commit history (squash to the rescue).

Anyway, I  added more FastAPI features to my benchmark and the improvement became even more significant: 
https://jvllmr.github.io/fastapi-deferred-init/dev/bench/

jvllmr:
I actually found the offending lines. For some reason the summary at the bottom showed 100%, but I found the lines immediately when checking "Hide covered". All green now ✅ 🏁

jvllmr:
I dropped the unintentional commits just to be sure

svlandeg:
Thanks! We've got a bit of a reviewing queue at the moment, but we've added this one and will get back to you once we've been able to review it in more detail 🙏

svlandeg:
(FYI - no need to merge in `master` when there are no conflicts)

YuriiMotov (commented):
LGTM in general.

Let's wait for Sebastian to validate the idea, then I think we should work a bit more on tests (split by test cases, simplify a bit, maybe add more tests) and probably mention this somewhere in docs.

### The idea of this PR:
When we include the router into another, it creates copies of each `APIRoute` recursively, triggering `__init__` methods each time.
But there are some "expensive" attributes that are initialized in `__init__` and there is no point to calculate them each time we include the chain of routers (it will create a copy and trigger `__init__` again).

It makes sense to create the tree of routers first and then init those attributes once per each instance in the tree.
This PR turns those attributes into cached properties, so they are only calculated on first access (by default, but it's configurable using `defer_init` parameter).


### Performance test

| parameter | master | this pr |
|---|---|---|
| App init time | 0.0118 | 0.0052 |
| First request time | 0.0027 | 0.0024 |
| Second request time | 0.0017 | 0.0016 |
| Total time | 0.0162 | 0.0091 |

<details><summary>Test code</summary>

```py
import time

class MyModel(BaseModel):
    id: int
    name: str


def get_app():

    router9 = APIRouter(prefix="/level9")
    @router9.post("/")
    async def read_root9(data: MyModel):
        return {"message": "Hello from Level 9"}

    router8 = APIRouter(prefix="/level8")
    router8.include_router(router9)
    @router8.post("/")
    async def read_root8(data: MyModel):
        return {"message": "Hello from Level 8"}

    router7 = APIRouter(prefix="/level7")
    router7.include_router(router8)
    @router7.post("/")
    async def read_root7(data: MyModel):
        return {"message": "Hello from Level 7"}

    router6 = APIRouter(prefix="/level6")
    router6.include_router(router7)
    @router6.post("/")
    async def read_root6(data: MyModel):
        return {"message": "Hello from Level 6"}

    router5 = APIRouter(prefix="/level5")
    router5.include_router(router6)
    @router5.post("/")
    async def read_root5(data: MyModel):
        return {"message": "Hello from Level 5"}

    router4 = APIRouter(prefix="/level4")
    router4.include_router(router5)
    @router4.post("/")
    async def read_root4(data: MyModel):
        return {"message": "Hello from Level 4"}

    router3 = APIRouter(prefix="/level3")
    router3.include_router(router4)
    @router3.post("/")
    async def read_root3(data: MyModel):
        return {"message": "Hello from Level 3"}

    router2 = APIRouter(prefix="/level2")
    router2.include_router(router3)
    @router2.post("/")
    async def read_root2(data: MyModel):
        return {"message": "Hello from Level 2"}

    router1 = APIRouter(prefix="/level1")
    router1.include_router(router2)
    @router1.post("/")
    async def read_root1(data: MyModel):
        return {"message": "Hello from Level 1"}

    app = FastAPI(title="Deeply Nested Routers Example", version="1.0")

    app.include_router(prefix="/api_1",router=router1)
    app.include_router(prefix="/api_2",router=router1)

    return app

def test():
    app_init_results = []
    first_request_results = []
    second_request_results = []
    total_time_results = []

    for _i in range(10):
        start_time = time.time()
        app = get_app()
        app_init_completion_time = time.time()
        app_init_results.append(app_init_completion_time - start_time)
        client = TestClient(app)

        response = client.post(
            "/api_1/level1/level2/level3/level4/level5/level6/level7/level8/level9",
            json={"id": 1, "name": "Test"}
        )
        first_request_completion_time = time.time()
        first_request_results.append(first_request_completion_time - app_init_completion_time)

        response = client.post(
            "/api_2/level1/level2/level3/level4/level5/level6/level7/level8/level9",
            json={"id": 1, "name": "Test"}
        )
        second_request_completion_time = time.time()
        second_request_results.append(second_request_completion_time - first_request_completion_time)

        end_time = time.time()
        duration = end_time - start_time
        total_time_results.append(duration)

    print(f"App init time: {sum(app_init_results)/len(app_init_results)}")
    print(f"First request time: {sum(first_request_results)/len(first_request_results)}")
    print(f"Second request time: {sum(second_request_results)/len(second_request_results)}")
    print(f"Total time: {sum(total_time_results)/len(total_time_results)}")


if __name__ == "__main__":
    test()
```

</details>


@jvllmr, thanks!

fastapi/routing.py
```diff
@@ -1002,7 +1001,7 @@ def body_field(self) -> ModelField | None:
         )
 
     @cached_property
-    def app(self) -> ASGIApp:  # type: ignore
+    def app(self) -> ASGIApp:
```
Jamim: _Hello @jvllmr,_

It's not just a comment.
This removal causes [an error on CI](https://github.com/fastapi/fastapi/actions/runs/25688976393/job/75420059622?pr=10589#step:8:27), so it's worth reverting.

Jamim (changes requested):

tiangolo:

Thank you! 🍰

PR #10621, opened 2023-11-10
Add option to generate OpenAPI specification by either Pydantic field name or alias

YYYasin19:
Hi!

Following the issue raised [here](https://github.com/tiangolo/fastapi/issues/771) and the discussion, I wanted to submit a simple option to the schema generation that I believe solves this problem (thanks @rbuffat for the right code places!).

TL;DR: Some people want to generate their openapi schemas based on the field names in their Models and not the aliases. This was already supported by Pydantic's Schema generation process but not enabled by the FastAPI interface.

I have updated the method overview in the "How To"-Guide but don't think that it's important enough to update the code examples as well. I can update those as well.

**Questions**: 

- [ ] Are there any side-effects to consider (e.g. other types of schema generation)?
- [ ] Does it make more sense to couple this with `response_model_by_alias: bool` in the route handler? That way a user can pass this option for specific routes only and his API spec and return types always match!

### DIFF
--- a/docs/en/docs/how-to/extending-openapi.md
+++ b/docs/en/docs/how-to/extending-openapi.md
@@ -26,6 +26,7 @@ And that function `get_openapi()` receives as parameters:
 * `summary`: A short summary of the API.
 * `description`: The description of your API, this can include markdown and will be shown in the docs.
 * `routes`: A list of routes, these are each of the registered *path operations*. They are taken from `app.routes`.
+* `field_names_by_alias`: A boolean indicating whether to use the field names or the alias names for the models.
 
 !!! info
     The parameter `summary` is available in OpenAPI 3.1.0 and above, supported by FastAPI 0.99.0 and above.
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -347,6 +347,7 @@ else:
     @dataclass
     class GenerateJsonSchema:  # type: ignore[no-redef]
         ref_template: str
+        by_alias: bool = True
 
     class PydanticSchemaGenerationError(Exception):  # type: ignore[no-redef]
         pass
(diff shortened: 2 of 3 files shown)


tiangolo:
📝 Docs preview for commit 217bf8f2b99bc1a07b0e0c2dad98dcb6eab3ad8b at: https://be7cffd2.fastapitiangolo.pages.dev

janickrohrbach:
This is great and would be very useful for us! In my opinion, it makes sense to couple this with `response_model_by_alias` such that the schema always shows the correct response_model.

YYYasin19:
I don't think that's easily available on the FastAPI end since, afaik, the Schema Generator is instantiated once and generates the schema for all routes.

One could
(1) instantiate a new schema generator for each route -- maybe a bit inefficient with a lot of routes
(2) configure the pydantic response model to use the field name instead of the alias during `model_dump_json_schema()`

YuriiMotov:
This generates invalid openapi schema.

```python
from fastapi import FastAPI

app = FastAPI()

class Item(BaseModel):
    id_: str = Field(alias="id")

@app.post("/")
async def add_item(item: Item) -> Item:
    return item

```

Object schema is:
```
      "Item": {
        "properties": {
          "id_": {
            "type": "string",
            "title": "Id"
          }
        },
        "type": "object",
        "required": [
          "id_"
        ],
        "title": "Item"
      },
```

But if you try sending request according to this schema (`{"id_": "string"}`) you will get error response 422.

If you want to change alias behavior for either request or response, you can use `validation_alias` and `serialization_alias` parameters:
```python
class Item(BaseModel):
    id_: str = Field(validation_alias="id")
```
This will create separate schemas for input and output models and validation will not be broken.

tiangolo:
Thanks for the review @YuriiMotov! :rocket: 

The way to customize alias should be with Pydantic alias support, separated for serialization and validation: https://docs.pydantic.dev/latest/concepts/alias/

But then, I would first need a use case that actually needs this, I can't come up with something that makes having a model for an API have an alias defined but not use it. :thinking: 

It would probably be best to start a new GitHub Discussion with the specific use case you have in mind, how it is different from the previous ones, and why do you think it would justify having support for this in the framework directly.

For now, I'll pass on this one, but thanks for the effort and interest! :coffee:

PR #10631, opened 2023-11-12
feat: MsgSpecJSONResponse 

rapidrabbit76:
discussions: #10616 

update MsgSpecJSONResponse using "[msgspec](https://github.com/jcrist/msgspec)" like ORJSON, UJSON.


peidrao (commented):
Hey @rapidrabbit76 great PR

I would like some test cases to ensure the consistency of your updates.

rapidrabbit76:
i used tests with a similar structure to existing ["ORJSON"](https://github.com/tiangolo/fastapi/blob/master/tests/test_orjson_response_class.py) test cases.

i will add tests for several JSON schemas and use them through "response_class".

Would it be okay?

tiangolo:
Thanks @rapidrabbit76!

For these use cases is precisely what I created this section in the docs for: https://fastapi.tiangolo.com/advanced/custom-response/#custom-response-class

And the solution is indeed what you did in the question, to create the small custom response right as you need it (3 lines) and use it:

```Python
class MsgSpecJSONResponse(JSONResponse):
        return msgspec.json.encode(content)
```

It would be very difficult to take each possible type of response class (e.g. someone had proposed simplejson, it would be understandable to then also propose YAML, XML, etc.). So I prefer to explain clearly how to do it on the developer side instead of including more responses here.

PR #10638, opened 2023-11-13
✨ Allow custom `FastApi.openapi` to be async

bkis:
As explained in #10637, I argue there are good reasons to allow custom implementations of `FastApi.openapi` to be `async`. This PR would do just that.

I am aware that in terms of code style, there are numerous ways to do this. I went for *readability-before-elegance*. 

The test case that is part of this PR closes the hole in coverage (of 1 single line of code) but is also a bit clunky. I see no other way write a meaningful test for this, though – and I'm happy if anyone comes up with a better idea.


YuriiMotov (changes requested):
This feature seems to cover very niche use case.
But at the same time I don't see any problems if this is added.

I would only move test to separate module (like `test_openapi_async.py`). **UPD:** done
The `test_application.py` is not right place to put this test in

tests/test_openapi_async.py
```diff
@@ -0,0 +1,14 @@
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+
+
+def test_allow_async_openapi():
+    async def async_openapi():
+        return {"foo": "bar"}
+
+    mod_app = FastAPI()  # use fresh instance to not affect other tests
```
YuriiMotov:
```suggestion
    mod_app = FastAPI()
```

This comment is not needed anymore

YuriiMotov (approved):

YuriiMotov:
Just realized that mypy will argue on the following line
```python
app.openapi = async_openapi
```
with `Incompatible types in assignment (expression has type "Callable[[], Coroutine[Any, Any, dict[str, Any]]]", variable has type "Callable[[], dict[str, Any]]")`.

Although, it argues on the code from Docs example either with `Cannot assign to a method`.

Let's think if we can do something with it

bkis:
> Just realized that mypy will argue on the following line
> 
> ```python
> app.openapi = async_openapi
> ```
> 
> with `Incompatible types in assignment (expression has type "Callable[[], Coroutine[Any, Any, dict[str, Any]]]", variable has type "Callable[[], dict[str, Any]]")`.

I see... so I guess as overriding this method is [documented](https://fastapi.tiangolo.com/how-to/extending-openapi/#override-the-method) and mypy should be able to understand what's going on, we'll have to prepare it for this new case? But I really don't know how, because the `Callable[...]` part is an inferred type and we can't really make it explicit anywhere, right?

YuriiMotov:
> I see... so I guess as overriding this method is [documented](https://fastapi.tiangolo.com/how-to/extending-openapi/#override-the-method) and mypy should be able to understand what's going on, we'll have to prepare it for this new case? But I really don't know how, because the `Callable[...]` part is an inferred type and we can't really make it explicit anywhere, right?

Yes, seems that fixing this will require quite a lot of changes. Now I'm not sure it's worth the efforts.
Introducing badly-typed code just to address very rare use-case seems to be not a good idea either.
I would suggest dropping this and think about more important things

bkis:
I agree. And no hard feelings from my side. It's not a very critical feature, I can see that :)
Thank you for your time and effort anyway!

YuriiMotov:
Thanks for understanding! And for your efforts!

PR #10647, opened 2023-11-14
fix: respect call localns at get_typed_signature

Lancetnik:
Original [`get_typed_signature`](https://github.com/tiangolo/fastapi/blob/master/fastapi/dependencies/utils.py#L227) doesn't respect a function local scope, that leads to pydantic error in this case:

Probably, related to #10007

```python

from fastapi import FastAPI


def main():
    app = FastAPI()

    class M(BaseModel):
        m: str

    @app.post("/")
    async def handler(m: M):
        print(m)

        r = client.post("/", json={"m": "Hi!"})
        print(r.json())

main()
```

Pydantic error:

```bash
pydantic.errors.PydanticUndefinedAnnotation: name 'M' is not defined
```

I made this function `localns` respect and write a test, broken in the original **FastAPI** and fixed now.

Probably, it is not the best way to find `outer_frame`, we can find smth better, I think. But it still the only one way to fix the Issue.


YuriiMotov:
Probably related PR: https://github.com/fastapi/fastapi/pull/11355

PR #10694, opened 2023-11-22
✨ Add support for using `@app.api_route()` in user code with multiple `methods`

emikil:
See details in discussion: https://github.com/tiangolo/fastapi/discussions/8449

When creating an API endpoint with more than one method, the unique id is being generated for only the first method.
The proposed solution keeps the current functionality for regular endpoints (with only one method) and regenerates the unique id for each method separately.

Example:

```python
@app.api_route("/multiple-methods", methods=["GET", "POST"])
def multiple_methods():
```

will generate 2 unique IDs:
- `multiple_methods_multiple_methods_get`
- `multiple_methods_multiple_methods_post`

instead of generating `multiple_methods_multiple_methods_post` twice.


hasansezertasan (approved):
LGTM

dagadbm:
can someone please merge this?

I am trying to source control openapi-schema.json and do a git diff when it changes but I can't since it always changes the operationId randomly.

Is there something that can be done user side to prevent this?

YuriiMotov (approved):
The problem still exists (FastAPI v.0.110.0).
This PR fixes this issue.
I'm not sure that it's the best way to fix it, but another way I see (removing `operation_id = f"{operation_id}_{list(route.methods)[0].lower()}"` from `generate_unique_id` and removing `if len(route.methods) > 1:` condition from `get_openapi_operation_metadata`) causes many tests to fail and requires much more changes..
So, I think this PR should be merged

dagadbm:
You know what i ended up doing?

Just removing all methods and writing each one separately.

Also had to sort the json schem api dump and now I finally have an idempotent openapi schema generator function which we now source control which we can also use to generate typescript types

parfaire (approved):

iudeen (approved):
NEAT

Acervans:
Any reviewer to resolve the PR please? It's been well over a year since it was requested

YuriiMotov:
This PR also fixes one more thing.

The following test fails **sometimes** in current version (0.115.12) because schema name may be either `Body_update_item_items__item_id__put` or `Body_update_item_items__item_id__patch`:

```python


def test_():
    app = FastAPI()

    class Item(BaseModel):
        name: str

    async def update_item(item_id: str, item_data: Item = Body(embed=True)):
        pass

    app.add_api_route("/items/{item_id}", update_item, methods=["PATCH", "PUT"])

    openapi = app.openapi()

    assert openapi["paths"]["/items/{item_id}"]["put"]["requestBody"] == {
                    "required": True,
                    "content": {
                        "application/json": {
                            "schema": {
                                "$ref": "#/components/schemas/Body_update_item_items__item_id__put"
                            }
                        }
                    },
                }
```

This must be annoying for those who are using code generators.

With this PR schema name is always `Body_update_item_items__item_id_` for path operations with multiple methods.

YuriiMotov:
The thing I don't really like in this approach is that this way we spread the logic of `operationID` generation across two modules (it's generated in `fastapi.utils.generate_unique_id` and then being modified in `fastapi.openapi.utils.get_openapi_operation_metadata` for paths that have multiple methods).
As a side effect, people can't have full control over generating `operationID` for paths with multiple methods (they can override first part, but the suffix will be added by `get_openapi_operation_metadata`).

IMO, it would be better to add optional `method` parameter to `fastapi.utils.generate_unique_id` function and call it with `method=None` to generate unique id for path object (it's then being used to generate request body schema name and the title of response schema), and with `method=OPERATION_METHOD` to generate unique id for operation object.

I attempted to implement this approach (in [this branch](https://github.com/YuriiMotov/fastapi/tree/fix-duplicated-operationID-for-routes-with-multiple-methods)). It works. It took some time to make mypy happy.
If anybody is up to take it and create an alternative PR out of it, feel free to do so!
To review it you mainly need to look at [this](https://github.com/YuriiMotov/fastapi/blob/781e7b593d3c8432e2f4f0249864864d52b8a708/fastapi/utils.py#L179-L190), [this](https://github.com/YuriiMotov/fastapi/blob/781e7b593d3c8432e2f4f0249864864d52b8a708/fastapi/openapi/utils.py#L237) and [this](https://github.com/YuriiMotov/fastapi/blob/781e7b593d3c8432e2f4f0249864864d52b8a708/fastapi/routing.py#L516-L533). Or, look at all changes, they are in [one commit](https://github.com/fastapi/fastapi/compare/master...YuriiMotov:fastapi:fix-duplicated-operationID-for-routes-with-multiple-methods).

@emikil, @iudeen, @hasansezertasan, could you take a look?

YuriiMotov:
I just found PR #4851 that also addressed the problem of using `api_route` with multiple methods . It was rejected with the reason:

> Using multiple methods for the same function is not actually supported. @router.api_route() is there as a semi-internal utility, and because Starlette supports multiple methods, the interface is there. But the official way to do it in FastAPI and have all the FastAPI features (including the automatic OpenAPI integration, with operation IDs) is by using the dedicated methods, like app.post(), app.get(), etc.

So, the proposed solution is to avoid using `.api_route("/", methods=["PUT", "POST"])`. Use multiple decorators with single methods instead:

```python
@app.put("/")
@app.post("/")
async def po() -> MyModel:
    return {}
```

**PS:** we should probably add the admonition about this in the docs and the docstring

bitjerry:
I hope FastAPI will merge this PR. 
If  router.api_route  is a public method, then specifying  methods  should be available; otherwise, this method should be set to  router._api_route . 
For me, I care about this PR because writing  [get, post]  directly is more concise than using  router.get  and  router.post .

tiangolo:
Thanks for the interest, `@app.api_route("/multiple-methods", methods=["GET", "POST"])` is not supported / documented in FastAPI.

Supporting it would soft-encourage doing that, and I consider it an anti-pattern that I wanted to avoid since the original design, so I prefer to not take it. Thanks for the effort! :coffee:

PR #10698, opened 2023-11-22
fix: Reading List Type Query Parameter with Annotated Type Hints

nao23:
This patch fixes the issue where, when using specific type hints (when `Annotated` type appears within nested type) with list type query parameters, parameter values cannot be read as list.

When I execute the following code, FastAPI returns a 422 response.
(Environment: python version: `3.9.6`, fastapi version: `0.104.1`, pydantic version: `2.4.2`)

```python
from pydantic import conlist  
from typing import Optional  
  
  
app = FastAPI()
  
@app.get("/test")  
async def test(  
    tag: Optional[conlist(str, min_length=1, max_length=5)] = Query(None, description="tag")  
):  
    return {"tag": tag}  
  
  
if __name__ == "__main__":  
        res = client.get("/test", params={"tag": ["a", "b"]})  
        print(res.json())
```

log:
```
{'detail': [{'type': 'list_type', 'loc': ['query', 'tag'], 'msg': 'Input should be a valid list', 'input': 'b', 'url': 'https://errors.pydantic.dev/2.4/v/list_type'}]}
```

As I read through the code, I found that in `request_params_to_args()`, the result of whether a parameter is a scalar sequence is done with `is_scalar_sequence_field() -> field_annotation_is_scalar_sequence()` and depending on the result, how to read parameter value is changed.
https://github.com/tiangolo/fastapi/blob/0.104.1/fastapi/dependencies/utils.py#L651
https://github.com/tiangolo/fastapi/blob/0.104.1/fastapi/_compat.py#L242
https://github.com/tiangolo/fastapi/blob/0.104.1/fastapi/_compat.py#L563

In the given code execution, type annotation passed to `field_annotation_is_scalar_sequence()` is  `Optional[Annotated[List[str], Len(min_length=1, max_length=5)]]`. But, because there is no implementation to interpret `Annotated` , so it is unable to recognize it as a scalar sequence and returns False.

In this patch, I modified to `field_annotation_is_scalar_sequence()` to be enabled to parse `Annotated`.

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -571,6 +571,8 @@ def field_annotation_is_scalar_sequence(annotation: Union[Type[Any], None]) -> b
             elif not field_annotation_is_scalar(arg):
                 return False
         return at_least_one_scalar_sequence
+    elif origin is Annotated:
+        return field_annotation_is_scalar_sequence(get_args(annotation)[0])
     return field_annotation_is_sequence(annotation) and all(
         field_annotation_is_scalar(sub_annotation)
         for sub_annotation in get_args(annotation)
--- a/tests/test_annotated.py
+++ b/tests/test_annotated.py
@@ -1,3 +1,5 @@
+from typing import Union
+
 import pytest
 from dirty_equals import IsDict
 from fastapi import APIRouter, FastAPI, Query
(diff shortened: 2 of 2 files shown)


Lancetnik:
Looks like your example

```python
Union[Annotated[str, Query()], None] = None
```

Has no sence cuz it should be refactored as

```python
Optional[Annotated[str, Query()]] = None
```

And, finally
```python
Annotated[Optional[str], Query()] = None
```

I think, `Annotated` can't be nested and can't contains more than 2 items: the first item is a variable type, the second item is a variable resolving way (source, depends, etc)

That is the semantic, but you are trying to breake it. Probably, you should find another example to proof your idea.

tiangolo:
Thanks for the interest @nao23! As @Lancetnik mentions, you should put `Annotated` around the rest.

If you still have questions, please create a new GitHub Discussion, it would make it easier to understand your use case and help you out: https://github.com/tiangolo/fastapi/discussions/new?category=questions

For now, I'll pass on this one. ☕

PR #10723, opened 2023-11-27
Ignore .python-version file

0xsirsaif:


dtleal (approved):

dtleal (approved):

tiangolo:
Thanks for the interest! But I'm not really sure what this is about, what does it help with or what tool would use or generate this file, so I'll pass on this one.

I think it would probably be better to start with a GitHub Discussion question, filling all the template data with all the details, that would help me/us understand your use case and help you: https://github.com/tiangolo/fastapi/discussions/new?category=questions

PR #10778, opened 2023-12-14
fix params description error

hubertshelley:


fastapi/dependencies/utils.py
```diff
@@ -109,20 +109,44 @@ def get_param_sub_dependant(
...
+    for query_param in dependant.query_params:
+        query_param_field = depends.dependency.model_fields.get(query_param.name)
+        if query_param_field:
+            query_param.field_info.description = (
+                query_param_field.description or query_param_field.title or ""
+            )
+        else:
+            query_param.field_info.description = (
+                query_param_field.description or query_param_field.title or ""
+            )
```
andreygaag, line 128: Looks like the same actions are performed regardless of the condition. Here's a more concise version:
```python
for query_param in dependant.query_params:
    query_param.field_info.description = query_param_field.description or query_param_field.title or ""
```
This should yield the same result.

Do you perhaps mean the following?
```python
for query_param in dependant.query_params:
    if not query_param.field_info.description:
        query_param.field_info.description = query_param_field.description or query_param_field.title or ""
```
to fill description if it empty.

andreygaag (changes requested):

fastapi/dependencies/utils.py
hubertshelley: Okay, thank you for the suggestion

PR #10052, opened 2023-08-09
Change .dict() to .model_dump()

sulavmhrzn:
Use .model_dump() method because .dict() method is deprecated

PR #10811, opened 2023-12-22
updating sql-docs

xnpltn:
1. Change "item.dict()" to "item.model_dump()" . Because dict() method is deprecated and primitive. It is better practice to use updated version. 
2. Removed hard coded status codes and used "status" object.

### DIFF
--- a/docs/en/docs/tutorial/sql-databases.md
+++ b/docs/en/docs/tutorial/sql-databases.md
@@ -470,15 +470,15 @@ The steps are:
 !!! tip
     Instead of passing each of the keyword arguments to `Item` and reading each one of them from the Pydantic *model*, we are generating a `dict` with the Pydantic *model*'s data with:
 
-    `item.dict()`
+    `item.model_dump()`
 
     and then we are passing the `dict`'s key-value pairs as the keyword arguments to the SQLAlchemy `Item`, with:
 
-    `Item(**item.dict())`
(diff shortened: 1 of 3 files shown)


tiangolo:
📝 Docs preview for commit b55817cf81354b1b6066ac30de9cd42f11f3e0d5 at: https://b9f484d9.fastapitiangolo.pages.dev

Xnpltn:
@tiangolo @Kludex @tokusumi can you review these changes?

xnpltn:
@tiangolo @Kludex  can you review those changes?

nilslindemann (approved):
This is good in the context of `sql-databases.md`. However, there are many more appearances of `.dict` in the docs. #10052 could do that, as that PR´s title implies "whole docs" :-)

tiangolo:
Thanks @xnpltn! ☕ 

And thanks for the help here @nilslindemann! 🙇 

The problem is that FastAPI is still compatible with Pydantic v1, and the examples are currently compatible with both Pydantic v1 and v2, updating them to use `.model_dump()` makes them no longer compatible with Pydantic v1.

To alleviate the problem a bit, I added notes to the docs suggesting people to use `.model_dump()`, while keeping the examples compatible with Pydantic v1 here: https://github.com/tiangolo/fastapi/pull/10929

Given that, I'll pass on this, but thanks for the effort! 🍰

PR #10831, opened 2023-12-25
✨ Add support for raising exceptions (including `HTTPException`) in dependencies with `yield` in the exit code, do not support them in background tasks

tiangolo:
✨ Add support for raising exceptions (including `HTTPException`) in dependencies with `yield` in the exit code, do not support them in background tasks.

This moves the `AsyncExitStack` from a middleware that would run the exit code after the response is sent and background tasks are executed to inside of the dependencies logic, after creating the response but before sending it.

So, now dependencies with yield are nicely wrapped by exception handlers, middleware, and the request-response cycle.

Before they would be executed after background tasks finished. The intention was to allow sharing the same database connection, but this would mean holding a resource (e.g. DB connection) until after the request travels through the network, it would prevent supporting raising exceptions after `yield`, and allowed dangerously coupling logic related to resources from dependencies with `yield` with background tasks (e.g. hard to test).

This will also enable several new features and changes, including upgrades to the recent versions of Starlette, and/or making background tasks independent of the request-response lifecycle, etc.


tiangolo:
📝 Docs preview for commit 01b1fe0273282c6232bf4768331ac95a1f29affa at: https://48e522da.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit c239211d43fe03be373f16529c8d3a56dcbab8b0 at: https://775033af.fastapitiangolo.pages.dev

Issue #2008, opened 2020-09-03
Automatic OPTIONS request with route schema


samuelcolvin:
### First check

Yes to all (great list of checks btw, I might copy them to pydantic 👍 )

### Thank you

@tiangolo FastAPI is great. This is the first time I've used it in production and I'm really impressed by both the range of features and particularly the documentation. I've written and deployed web apps using django, flask, aiohttp, starlette and now fastapi. FastAPI is the only python web framework I've seen to come close to django in terms of quality of documentation, it's definitely now my go-to web framework.

**Thank you for the amazing project**, I feel really proud to have contributed to it indirectly.

### Feature Request

The only thing I've seen missing from fastpi is support for informative OPTIONS requests for all/most/some endpoints, automatically included based in the route's definition.

I effectively want the following:

``py
class LoginForm(BaseModel):
    email: EmailStr
    password: SecretStr

@app.post('/login/')
async def login(login_form: LoginForm):
    logger.info('login attempt from %s', login_form.email)
    m = 'invalid username or password'
    raise exceptions.Http470(m, details=[{'loc': ['password'], 'msg': m}])

@app.options('/login/', include_in_schema=False)
async def login_options():
    return LoginForm.schema(by_alias=True)
```

Without having to define the second endpoint. This is so my react frontend can build forms without having to duplicate the field definitions.

I guess it might have to be slightly more complex to include parameters, query args and body model(s).

I know I can get the information I need from `/openapi.json`, but that involves more processing in js and more data downloaded.

In the past when I've implemented a feature like this myself, I had to deal with `OPTIONS` requests associated with preflight checks separately, but that's relatively easy by looking for the `Access-Control-Request-Method` header.

I guess if this feature was to be approved it would have to be optional, default off. I guess through a keyword arg on `FastAPI`, `include_router` and each route.

Vlatterran:
According to the [specifications of OPTIONS method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS), i'm not sure if it is supposed to do what you want

If I understands correctly, you want to have a shortcut to the schema of some endpoints, while OPTIONS describing the whole path. 

I think it'll be better to  have such a shortcut by <METHOD> http:\\example.net\docs\<path> for the <METHOD> <path> endpoint

samuelcolvin:
No, I want the scheme for that URL, which is exactly what OPTIONS requests are for.

gwerbin:
[...]

PR #10861, opened 2023-12-30
✨ Add automatic OPTIONS request with route schema

whysage:


fastapi/applications.py
```diff
@@ -1788,6 +1792,14 @@ def get(
                 """
             ),
         ] = Default(generate_unique_id),
+        add_auto_options_route: Annotated[
+            Optional[bool],
+            Doc(
+                """
+                Auto create options route.
```
gwerbin, line 1799: How about something like this?

```suggestion
                Automatically create a route to handle the OPTIONS HTTP verb.
```

whysage: thanks, updated

kaze (approved):

whysage:
@kaze thank you for the request approval!

lyksdu:
@whysage @kaze  Sorry to disturb. This merge request is a very meaningful merge request. Is there a way for us to rerun the two tasks that did not get results so that this request can be merged smoothly?

fastapi/routing.py
```diff
@@ -921,6 +924,37 @@ def add_api_route(
...
+
+            async def options_route() -> Response:
+                return Response(headers={"Allow": ", ".join(allowed_methods)})
+
+            self.routes.append(
+                APIRoute(
+                    self.prefix + path,
+                    endpoint=options_route,
+                    methods=["OPTIONS"],
+                    include_in_schema=True,
```
RickardCardell, line 954: Does this mean that it will be visible in the docs-UI? I.e. will each endpoint also have the corresponding endpoint with the options-method visible in the UI?

whysage:
> @whysage @kaze Sorry to disturb. This merge request is a very meaningful merge request. Is there a way for us to rerun the two tasks that did not get results so that this request can be merged smoothly?

Hi @lyksdu, I tried to get review and approve for this pull request from Aug 10, 2023 to Mar 19, 2024 (7 month) and after all, was blocked from discussion (https://github.com/fastapi/fastapi/issues/2008#issuecomment-2007639569). Maybe it's fair due to many messages and tagging others.

If this pull request has any chance to be merged, I'll continue work on it.

YuriiMotov:
@whysage, thanks for your interest! And thank you for your patience🙏

Could you please explain why we may need the support for `OPTIONS` method if we already have openapi schema that contains the information about the supported methods for each path? Please, add this information (and, probably, reasons to have this feature) to the description of this PR to simplify the review process

whysage:
@YuriiMotov I could only quote the original issue text 
"I know I can get the information I need from /openapi.json, but that involves more processing in js and more data downloaded."
https://github.com/fastapi/fastapi/issues/2008#:~:text=body%20model(s).-,I%20know%20I%20can%20get%20the%20information%20I%20need%20from%20/openapi.json%2C%20but%20that%20involves%20more%20processing%20in%20js%20and%20more%20data%20downloaded.,-In%20the%20past

YuriiMotov:
> @YuriiMotov I could only quote the original issue text "I know I can get the information I need from /openapi.json, but that involves more processing in js and more data downloaded." #2008

Yes, but:
* This PR doesn't implement the feature that is requested in that issue (it's requested to return the schema, not the list of supported methods)
* Say we need this to avoid loading and parsing openapi.json, how the client will get the list of supported paths?

Could you please explain it? I don't quite understand how it works.

I thought you can just [generate the client SDK](https://fastapi.tiangolo.com/advanced/generate-clients/) automatically from openapi.json and then just use that SDK in your client code.

whysage:
@YuriiMotov, thanks for the reply!
You are absolutely right. Currently, the PR introduces a standard OPTIONS response with "Allow" data containing a list of allowed HTTP methods.
I'm also not sure:
1) if it brings value now
2) if OPTIONS is designed to return URL schema.

Maybe we can tune the PR in the direction of handling preflighted requests in CORS https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/OPTIONS#preflighted_requests_in_cors if this brings value and has a use case for it.

Thanks for the review!

YuriiMotov:
> Maybe we can tune the PR in the direction of handling preflighted requests in CORS

We already have `CORSMiddleware` for this: https://fastapi.tiangolo.com/tutorial/cors/#use-corsmiddleware

whysage:
So what is the resume? 
No added value found.
Let's close the PR?

YuriiMotov:
> So what is the resume? No added value found. Let's close the PR?

Yes, I think we should close this PR, and investigate if the feature requested in #2008 makes sense

tiangolo:
I'll pass on this one for now, thanks for the interest! :coffee:

PR #10886, opened 2024-01-05
♻️ Simplify Pydantic configs in OpenAPI models in `fastapi/openapi/models.py`

JoeTanto2:
Abstracted conditional class config to parent class to avoid repetitive checks.


tiangolo:
Good idea, thanks! :rocket:

PR #10893, opened 2024-01-05
🔧 Fix Ruff configuration unintentionally enabling and re-disabling mccabe complexity check

jiridanek:
Enabling `"C"` turns on complexity checks (C90, mccabe), which is unintended. Instead, enable `"C4"` to get flake8-comprehensions checks

See docs at https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4


tiangolo:
Nice, thanks @jiridanek! 🚀

PR #10896, opened 2024-01-06
🐛 Fix File field cannot be defined after Form field

hungtsetse:
🐛 If File field is defined after Form field, it will cause a validation error while processing a valid request.

I've added a test case in PR which can test against the issue.

```
# Good
@app.post("/file/")
def create_file(
    file: UploadFile = File(), token: str = Form()
):

# Validation error
@app.post("/files2/")
def create_file_2(
   token: str = Form(), file: UploadFile = File()
):
```

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -713,6 +713,7 @@ async def request_body_to_args(
                     except AttributeError:
                         errors.append(get_missing_field_error(loc))
                         continue
+            field_info = field.field_info
             if (
                 value is None
                 or (isinstance(field_info, params.Form) and value == "")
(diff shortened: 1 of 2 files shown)


igoose1:
I don't know if it's related but I'd like to point out about a regression somewhere. I upgraded to FastAPI 0.108.0. An endpoint with `Annotated[UploadFile | None, File()]` currently doesn't generate a valid OpenAPI schema. Swagger asks for a string to be inserted, however before upgrade (I guess before fastapi's pydantic being migrated to v2) I could upload a file.

YuriiMotov:
It seems like this problem is not longer active.
I tried running test `test_form_is_declared_before_file` from this PR on current FastAPI version and it doesn't show errors.

But I didn't manage to find the test that covers this case

YuriiMotov (commented):
The problem is no longer relevant in current FastAPI version (0.109.2).
I tried running test `test_form_is_declared_before_file` from this PR on current FastAPI version and it doesn't show errors.

PR #10898, opened 2024-01-06
✨ Add Service() annotation

dolfinus:
Currently dependency injection mechanism based on `Depends()` can be used for 2 different cases:

1. Move query/body fields from endpoint function annotation to separate class, but avoid making these fields nested in API schema.
2. Create objects of some internal classes (e.g. services) used for business logic, like data validation/creation/bypassing to external applications and so on.

Example application:

```python

from sqlalchemy import select
from sqlalchemy import Column, Integer, String
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy.ext.asyncio import AsyncSession


@dataclass
class Item:
    id: int
    name: str


Base = declarative_base()

class DBItem(Base):
    __tablename__ = "items"
    id = Column(Integer, primary_key=True)
    name =  Column(String(50))


class ItemsRepository:
    def __init__(self, session: AsyncSession = Depends()):
        self.session = session

    async def get_items(self, limit: int, offset: int) -> list[DBItem]:
        query = select(DBItem).limit(limit).offset(offset)
        result = await self.session.scalars(query)
        return result.all()


async def session_factory(url: str):
    ...


app = FastAPI()
app.dependency_overrides[AsyncSession] = session_factory(settings.db_url)



@dataclass
class ItemsQuery:
  limit: int = 10
  offset: int = 0

@app.get("/items/")
async def get_items(
  query: ItemsQuery = Depends(),
  repository: ItemsRepository = Depends(),
) -> list[Item]:
    return await repository.get_items(limit=query.limit, offset=query.offset)
```

Here `Depends()` is used to create objects:
1. `query = ItemsQuery(limit=10, offset=0)`
2. `repository = ItemsRepository(session=AsyncSession(...))`

And then pass both of them as `get_items` function arguments.

If both cases `Depends()` traverses annotated class fields to generate dependency tree and so on. But later cases are very different:
1. For `ItemsQuery = Depends()` FastAPI should include `limit` and `offset` as query options in OpenAPI schema, and validate input query against this schema.
4. For `ItemsRepository = Depends()` FastAPI should construct object from values like database URL, engine options and so on, not from user input. And these options are not a part of OpenAPI schema, but a part of some kind of internal documentation (e.g. for DevOps or developer who is going to deploy this service).

So `Depends()` should simultaneously include some fields into OpenAPI schema and perform validation, but exclude others. And this should work recursively on the whole tree of dependencies in application.

This was already mentioned in #8668 - in some cases internal class fields (e.g. repository, auth provider configuration, etc) were included to OpenAPI schema because `Depends()` traverses dependency signature. But it *must* traverse signature to create proper dependency tree.

Some users created wrappers which allowed to make `Depends()` think that class `__init__(...)` method signature does not accept any arguments:
https://github.com/Tishka17/fastapi-template/blob/c5ef8b62d69f04c37ccce40a1cc3f103e9eacdd9/src/app/api/depends_stub.py

But this fails on Protocols and sometimes painful to use.

------------------------

The proposed change is to split `Depends()` functionality to 2 different classes with different behavior:
1. `Depends()` - traverse dependency signatures, build dependency tree, **add mandatory fields** as query/body parameters for schema and validation. During endpoint call substitute all the fields using dependency tree + `app.dependency_overrides`. If some field value is missing, return 422 status code with body marking the exact location of this field.
2. `Service()` - traverse dependency signatures, build dependency tree, but **do nothing with mandatory fields**. During endpoint call substitute all the fields using dependency tree + `app.dependency_overrides`. If some field value is missing, interpreter will raise TypeError exception in some dependency constructor. This means that something wrong with application setup or source code, not the user input, so 500 status code is returned.

<details>
<summary>Why adding one more class?</summary>

Previously I thought about just adding an option `Depends(expose=False)` which disabled handling mandatory fields as body/query params. But `Security()` is subclass of `Depends()`, and it meant to be used to mark some dependencies as adding even more information to OpenAPI schema. So I decided to move this to a separated class.
</details>

Currently annotations look like this:
1. `ItemsQuery = Depends()` - passing valid `limit` or `offset` is up to API user, wrong value == 422. Same as before, nothing is changed.
2. `ItemsRepository = Service()` - passing valid `AsyncSession` is up to application developer, wrong value == 500. Should be used only for internal classes, not user input.

Added tests for new class, including cases like:
```python

# this can be Protocol or abstract class
class BaseItemsRepository(ABC):
    # abstract class with just methods signatures. should be inherited by all implementations of ItemsRepository

    @abstractmethod
    async def get_items(self, limit: int, offset: int) -> list[Item]:
        ...

class InternalItemsRepository(BaseItemsRepository):
    # implementation which stores items in internal database using SQLAlchemy

    def __init__(self, session: AsyncSession = Service()):
        self.session = session

    async def get_items(self, limit: int, offset: int) -> list[Item]:
        # make query to DB here

class ExternalItemsRepository(BaseItemsRepository):
    # implementation which stores items in external service using API

    def __init__(self, client: ExternalAPIClient = Service()):
        self.client = client

    async def get_items(self, limit: int, offset: int) -> list[Item]:
        # make query to external REST API here


def application_factory(settings) -> FastAPI:
    app = FastAPI()
    # change chosen implementation based on application settings
    if settings.items_provider.type == "db":
        app.dependency_overrides[BaseItemsRepository] = InternalItemsRepository
        app.dependency_overrides[AsyncSession] = async_session_factory(settings.items_provider.db_url)

    elif settings.items_provider.type == "api":
        app.dependency_overrides[BaseItemsRepository] = ExternalItemsRepository
        app.dependency_overrides[ExternalAPIClient] = external_client_factory(settings.items_provider.api_url)

    return app
```

I haven't changed documentation much, lets discuss the proposal first.

### DIFF
--- a/docs/en/docs/reference/dependencies.md
+++ b/docs/en/docs/reference/dependencies.md
@@ -1,4 +1,4 @@
-# Dependencies - `Depends()` and `Security()`
+# Dependencies - `Depends()`, `Security()` and `Service()`
 
 ## `Depends()`
 
@@ -27,3 +27,18 @@ from fastapi import Security
 ```
 
 ::: fastapi.Security
+
+## `Service()`
+
+`Depends()` is designed to include matching fields of annotated class as body/query params.
+To avoid this, you can annotate class using `Service()` instead of `Depends()`.
+
+Here is the reference for it and its parameters.
+
+You can import it directly from `fastapi`:
+
+```python
+from fastapi import Service
+```
+
+::: fastapi.Service
--- a/fastapi/__init__.py
+++ b/fastapi/__init__.py
@@ -18,6 +18,7 @@ from .param_functions import Header as Header
 from .param_functions import Path as Path
 from .param_functions import Query as Query
 from .param_functions import Security as Security
+from .param_functions import Service as Service
 from .requests import Request as Request
 from .responses import Response as Response
 from .routing import APIRouter as APIRouter
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -32,6 +32,7 @@ class Dependant:
         background_tasks_param_name: Optional[str] = None,
         security_scopes_param_name: Optional[str] = None,
         security_scopes: Optional[List[str]] = None,
+        expose: bool = True,
         use_cache: bool = True,
         path: Optional[str] = None,
     ) -> None:
@@ -51,6 +52,7 @@ class Dependant:
         self.security_scopes_param_name = security_scopes_param_name
         self.name = name
         self.call = call
+        self.expose = expose
         self.use_cache = use_cache
         # Store the path to be able to re-generate a dependable from it in overrides
         self.path = path
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -134,6 +134,9 @@ def get_sub_dependant(
 ) -> Dependant:
     security_requirement = None
     security_scopes = security_scopes or []
+    expose = True
+    if isinstance(depends, params.Service):
+        expose = False
     if isinstance(depends, params.Security):
         dependency_scopes = depends.scopes
         security_scopes.extend(dependency_scopes)
@@ -149,6 +152,7 @@ def get_sub_dependant(
         call=dependency,
         name=name,
         security_scopes=security_scopes,
+        expose=expose,
         use_cache=depends.use_cache,
     )
     if security_requirement:
@@ -244,6 +248,7 @@ def get_dependant(
     call: Callable[..., Any],
     name: Optional[str] = None,
     security_scopes: Optional[List[str]] = None,
+    expose: bool = True,
     use_cache: bool = True,
 ) -> Dependant:
     path_param_names = get_path_param_names(path)
@@ -254,6 +259,7 @@ def get_dependant(
         name=name,
         path=path,
         security_scopes=security_scopes,
+        expose=expose,
         use_cache=use_cache,
     )
     for param_name, param in signature_params.items():
@@ -283,7 +289,9 @@ def get_dependant(
             ), f"Cannot specify multiple FastAPI annotations for {param_name!r}"
             continue
         assert param_field is not None
-        if is_body_param(param_field=param_field, is_path_param=is_path_param):
+        if not expose:
+            continue
+        elif is_body_param(param_field=param_field, is_path_param=is_path_param):
             dependant.body_params.append(param_field)
         else:
             add_param_to_fields(field=param_field, dependant=dependant)
@@ -567,6 +575,7 @@ async def solve_dependencies(
                 call=call,
                 name=sub_dependant.name,
                 security_scopes=sub_dependant.security_scopes,
+                expose=sub_dependant.expose,
             )
 
         solved_result = await solve_dependencies(
--- a/fastapi/param_functions.py
+++ b/fastapi/param_functions.py
@@ -2246,9 +2246,15 @@ def Depends(  # noqa: N802
     ] = True,
 ) -> Any:
     """
-    Declare a FastAPI dependency.
+    Declare a FastAPI Field dependency.
 
-    It takes a single "dependable" callable (like a function).
+    Objects of annotated class are automatically created and filled up by FastAPI
+    dependency injection mechanism (they should be annotated with `Body()`/`Query()`/etc).
+
+    Fields are automatically exposed to OpenAPI schema.
+
+    It takes a single "dependable" callable (like a function) which is factory creating objects of
+    annotated class. If "dependable" is omitted, FastAPI will use class constructor.
 
     Don't call it directly, FastAPI will call it for you.
 
@@ -2266,6 +2272,7 @@ def Depends(  # noqa: N802
 
 
     async def common_parameters(q: str | None = None, skip: int = 0, limit: int = 100):
+        # Query params
         return {"q": q, "skip": skip, "limit": limit}
 
 
@@ -2298,7 +2305,7 @@ def Security(  # noqa: N802
             dependency.
 
             The term "scope" comes from the OAuth2 specification, it seems to be
-            intentionaly vague and interpretable. It normally refers to permissions,
+            intentionally vague and interpretable. It normally refers to permissions,
             in cases to roles.
 
             These scopes are integrated with OpenAPI (and the API docs at `/docs`).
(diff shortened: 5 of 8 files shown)


tiangolo:
📝 Docs preview for commit 5af1052ea70ebb189febe499fd7b552b1b23cee5 at: https://e6b42526.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit db1db09242c3ea55e64d220e2058a6e1aabd5711 at: https://5cde3e4a.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit 3a58d2b10ed7024a03894a7afad525c2c6c41585 at: https://12752979.fastapitiangolo.pages.dev

purepani:
This is exactly what I was looking for in the dependency injection feature!

tiangolo:
📝 Docs preview for commit 870c884a26067c2ff339ca4372ed8026bc798304 at: https://d115dd74.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit 6ffc10399b636aba858f1c90496fbe4a065f0f46 at: https://191459c5.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit 9a18adbc609d0ea8acabd1b4237de2d96170532a at: https://c221152a.fastapitiangolo.pages.dev

AFutureD:
I am very excited that the PR is going to be merged.
The `Service()` is critical for my project and others using `FastApi`.

YuriiMotov (commented):
@dolfinus, thanks for your interest and efforts!

I'll express my opinion here, we can discuss it and then we will ask Sebastián to review the idea and take the decision.

As far as I understand it, the main idea is to address the following use case:
```Python

app = FastAPI()

class SomeService:
    def __init__(self, **kwargs):
        pass

@app.get("/")
async def index(service: SomeService = Depends()):
    pass
```
Running the code above you will see the required `kwargs` query parameter.
image

With the approach proposed in this PR you will use `Service` instead of `Depends` and it will hide that parameter from openapi schema.

I think we already have everything we need to solve it quickly. And this way is clear and less error-prone:
```python

app = FastAPI()


class SomeService:
    def __init__(self, **kwargs):
        pass


async def get_service():
    return SomeService(some_param="value")

@app.get("/")
async def index(service: SomeService = Depends(get_service)):
    pass
```

Regarding the last code example where you inject dependencies via `app.dependency_overrides` in production, I think this is a misuse of `app.dependency_overrides` and you shouldn't do this. Look at [this](https://github.com/fastapi/fastapi/pull/5452#issuecomment-2979998948) and previous comments.

So, I don't think we need this feature

dolfinus:
> Regarding the last code example where you inject dependencies via app.dependency_overrides in production, I think this is a misuse of app.dependency_overrides and you shouldn't do this.

Example above using `Depends(some_generator)` maybe okay for proof of concept applications, but things are more complicated in larger applications. For example, we need to create SQLAlchemy engine object (=connection pool) with custom settings (pool size, timeouts and so on), and then get connections from this configured pool. 

The approach used in documentation looks like this:
```python

engine: AsyncEngine = create_engine(
  os.environ["DB_URL"],
  pool_size=os.environ["DB_POOL_SIZE"],
  ...
)
session_factory: SessionFactory = create_session_factory(engine)

async def get_session() -> AsyncSession:
  async with session_factory() as session:
    yield session

@router.get("/users")
async def get_users(session: AsyncSession = Depends(get_session)):
  ...
```

This leads to creating a global Egine object, not during application start, but during module import. This is bad for testing - users cannot import application code without running database, it is impossible to run tests in parallel because configuration is performed using one environment variable, and not from test code.

Compare this with:
```python
@router.get("/usere")
async def get_users(session: AsyncSession):
  ...

def create_app(settings: AppSettings):
  app = FastAPI(...)
  engine: AsyncEngine = create_engine(settings.db.url, pool_size=settings.db.pool_size)
  app.dependency_override["AsyncSession"] = create_session_factory()
  return app
```

Here we can create new application with custom setting at any time, with no global objects. Great for testing, great for production as well (as user mey configure other factories for declared dependencies).

Without dependency_override, building production grade applications become very tricky. Imagine using code like this:
```python
def get_session_factory(settings: Settings) -> SessionFactory:
  engine = create_engine(settings.db.url, pool_size=settings.db.pool_size)
  return create_session_factory()

async def get_session(session_factory: SessionFactory = Depends(get_session_factory):
  async with session_factory():
    yield session

@router.get("/users")
async def get_users(session: AsyncSession = Depends(get_session)):
  ...
```
Engine here is created on _every_ `GET /users` request. There is no reusing of the same connection pool for the entire application, as FastAPI doesn't have DI scopes like *per-request* dependency or *per-application* dependency. Only `dependency_override` allows to implement per-app scope, and it does it well enough. It's not something the docs recommend, but also it's very far from "unusable".

YuriiMotov:
@dolfinus, thank you for reply!

> Engine here is created on _every_ `GET /users` request. There is no reusing of the same connection pool for the entire application, as FastAPI doesn't have DI scopes like _per-request_ dependency or _per-application_ dependency.

I'm probably missing something, but it seems to me that you can easily fix your last code example by decorating the `get_session_factory` with `@lru_cache` and creating the dependency function to get settings.
Something like:
```python
@lru_cache()
def get_settings() -> Settings:
    return Settings()

@lru_cache()
def get_session_factory(settings: Annotated[Settings, Depends(get_settings)]) -> SessionFactory:
    ...
```
This will also allow you to override it in tests and configure in production (add the conditions in `get_session_factory`).

Another option is to create the session pull in `lifespan` event. Also configurable and testable.

If you really need the feature like `dependency_override` for production (not for tests), I would recommend you open a discussion, describe use-cases and gather some feedback. If the idea is supported by community, I think FastAPI will accept such feature. But I believe it should be a dedicated mechanism, not `dependency_override`.

dolfinus:
> I'm probably missing something, but it seems to me that you can easily fix your last code example by decorating the get_session_factory with @lru_cache and creating the dependency function to get settings.

This is still a global object which cannot be altered for specific application instance. For example, I have an app which can be configured with different auth providers.

Using `Depends(generator)` approach:
```python
def get_auth_provider() -> AuthProvider:
  class_name = os.environ["AUTH_PROVIDER_CLASS"]
  if class_name == "KeycloakProvider":
    return KeycloakProvider()
  if class_name == "LDAPAuthProvider":
    return LDAPAuthProvider()
  raise ConfigurationError()

@router.get("/users")
async def get_users(auth: AuthProvider = Depends(get_auth_provider)):
  ...
```

requires using Mock or monkey patching to override `get_auth_provider` method result. It is much simpler to pass specific implementation via `app.dependency_override` during application setup:

```python
# configured during application creation
def create_app():
  app = FastAPI()

  if settings.auth_provider.class_name == "KeycloakAuthProvider":
    app.dependency_override["AuthProvider"] = KeycloakAuthProvider
  elif settings.auth_provider.class_name == "LDAPAuthProvider":
    app.dependency_override["AuthProvider"] = LDAPAuthProvider
  else:
    raise ConfigurationError()

  return app

@router.get("/users")
async def get_users(auth: AuthProvider = Depends()):
  ...
```

> Another option is to create the session pull in lifespan event. Also configurable and testable.

This still requires using `app.dependency_override`, so nothing really changes. The only difference is that lifespan is async but application creation is not.

> If you really need the feature like dependency_override for production (not for tests), I would recommend you open a discussion, describe use-cases and gather some feedback.

Using `app.dependency_override` for production application is already supported by FastAPI. The main question here is how tell FastAPI that if class generator is passed via `app.dependency_override`, it should not treat this Dependency as regular Body/Form/Header/Query param. Currently I use a small hark to hide class constructor from FastAPI.

YuriiMotov:
> Using app.dependency_override for production application is already supported by FastAPI.

I don't think so. The fact that it works doesn't mean that it's officially supported.
For me it looks like a misuse of `dependency_override`. So, you can use it on your risk.

Anyway, if you are Ok with using `dependency_override` in production, why don't you just create the dependency functions for every possible provider and use them instead of classes?

```python
def get_keycloak_provider(keycloak_settings: Annotated[KeycloakSettings, Depends(get_keycloak_settings)]):
    return KeycloakAuthProvider()

def create_app():
  ...
  if settings.auth_provider.class_name == "KeycloakAuthProvider":
    app.dependency_override["AuthProvider"] = get_keycloak_provider
```
You don't need `Service` feature for this, right?

It seems to me that what we are discussing here is more about `dependency_override` than the `Service` feature introduced by this PR.
I suggest you create a separate discussion where you can explain the use-case that you failed to solve without `dependency_override` and we will try to find an alternative solution.
For now it's hard for me to answer questions from your comment, because it seems to me there is no problem there. But maybe I just miss some requirements.
If you could describe your task with all requirements and probably provide an MRE with `dependency_override`, I would try to come up with the alternative solution.

And, I suggest we hide the unrelated comments here so that it would be easier for Sebastián to review this PR

dolfinus:
> Anyway, if you are Ok with using dependency_override in production, why don't you just create the dependency functions for every possible provider and use them instead of classes?

That looks like a lot of boilerplate, which can be easily avoided with just one annotation change. I thought that the purpose of FastAPI is to speed up writing applications bu reducing boilerplate, not by introducing it.


I'm closing this PR. It took 2 years for maintainers to start this discussion, and it'll took 2 more years to go to final conclusion that this should not be implemented.

YuriiMotov:
> That looks like a lot of boilerplate,

It's just 2-3 lines of code per provider class.
You can probably even reduce it by creating a factory function (in case you have a lot of providers with unified set of input parameters).

> I thought that the purpose of FastAPI is to speed up writing applications

I think we must also keep FastAPI being fast to learn. And for this reason we must carefully choose the features we accept.
That's why I asked to provide clarifications. We need to clearly understand why this feature is requested by developers and make sure that the issue can't be solved easily without this feature.

I can understand you frustration about how long it took for this PR to be reviewed. Sorry for that.
Hope for your understanding

dolfinus:
Created #13913. Maybe someone will have an insight

dolfinus:
> Regarding the last code example where you inject dependencies via app.dependency_overrides in production, I think this is a misuse of app.dependency_overrides and you shouldn't do this. Look at https://github.com/fastapi/fastapi/pull/5452#issuecomment-2979998948 and previous comments.

The mentioned comment says:
> Nevertheless, I would get nervous if the overriding dependency is so different that it would generate a different OpenAPI schema, I would think that then it would probably be doing too much, in a very different way, so I would prefer not to have that.

Current solution doesn't change OpenAPI schema on the fly, as all dependencies annotated with `Service()` are completely excluded from it.

PR #10903, opened 2024-01-08
Allow setting schema generator instance

mvdbeek:
This means one can follow an approach similar to
https://github.com/pydantic/pydantic/issues/7376#issuecomment-1763462208 when dealing with generics.


YuriiMotov:
Duplicate: #11357

tiangolo:
As the linked comment would actually not solve the problem in the original issue, this would probably be dangerous/not help.

I think the best would be to start with a discussion following the template form, including the use case you would like to solve, and then we can take it from there. That way there would be a clear objective for the changes, something to document, something to test, etc.

For now, I'll pass on this one, but thanks for the interest! :coffee:

PR #10914, opened 2024-01-09
updated examples due to pydantic deprecated method .dict(), to .model_dump()

jpizquierdo:
updated examples due to pydantic deprecated method .dict(), to .model_dump() due to Changes to pydantic.BaseModel

https://docs.pydantic.dev/latest/migration/

### DIFF
--- a/docs_src/async_sql_databases/tutorial001.py
+++ b/docs_src/async_sql_databases/tutorial001.py
@@ -62,4 +62,4 @@ async def read_notes():
 async def create_note(note: NoteIn):
     query = notes.insert().values(text=note.text, completed=note.completed)
     last_record_id = await database.execute(query)
-    return {**note.dict(), "id": last_record_id}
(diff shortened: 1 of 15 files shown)


tiangolo:
📝 Docs preview for commit 19383f724b15ba46e540d30340f454f6437f1baa at: https://e0b69fa5.fastapitiangolo.pages.dev

tiangolo:
Thanks @jpizquierdo! ☕ 

The problem is that FastAPI is still compatible with Pydantic v1, and the examples are currently compatible with both Pydantic v1 and v2, updating them to use `.model_dump()` makes them no longer compatible with Pydantic v1.

To alleviate the problem a bit, I added notes to the docs suggesting people to use `.model_dump()`, while keeping the examples compatible with Pydantic v1 here: https://github.com/tiangolo/fastapi/pull/10929

Given that, I'll pass on this, but thanks for the effort! 🍰

PR #10941, opened 2024-01-11
Distinction between function arguments and function parameters.

s111d:
I'm not super sure about this one, but it seems to me that, by definition, argument is something being passed in during the execution time, while the parameter is a part of function's signature. If so, I think the argument would be more appropriate in this case. Any thoughts?

### DIFF
--- a/docs/en/docs/alternatives.md
+++ b/docs/en/docs/alternatives.md
@@ -242,7 +242,7 @@ It was one of the first extremely fast Python frameworks based on `asyncio`. It 
 
 Falcon is another high performance Python framework, it is designed to be minimal, and work as the foundation of other frameworks like Hug.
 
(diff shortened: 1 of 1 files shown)


tiangolo:
📝 Docs preview for commit 1cc72fa9f0e63ecaf1e1461c3cd5fd2c956b0867 at: https://0b8f3f3e.fastapitiangolo.pages.dev

aanchlia:
> I'm not super sure about this one, but it seems to me that, by definition, argument is something being passed in during the execution time, while the parameter is a part of function's signature. If so, I think the argument would be more appropriate in this case. Any thoughts?

Yep, agreed with you on distinction between argument and parameter, and `arguments` is more appropriate in this context.

tiangolo:
Thanks! I wouldn't think it's necessarily too strictly speaking about *calling* a function vs *defining* a function, so it seems to me the word "parameters" is fine here too, so I'll leave it as is. But thanks! ☕

PR #10947, opened 2024-01-12
Printing object failing to encode to json

leonarf:
I was using FastAPI on a project an I had to see which object was making fail the json encoding to understand what to change in my code. So I suggest to add an Exception in this case that allow dev to read what object make encoding fails.
In my case it was an object of type numpy.int64, but I couldn't find which one before printing it.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -324,6 +324,7 @@ def jsonable_encoder(
         data = dict(obj)
     except Exception as e:
         errors: List[Exception] = []
+        errors.append(Exception(f"Exception while encoding {obj}"))
         errors.append(e)
         try:
             data = vars(obj)


codespearhead (approved):

tiangolo (changes requested):
Thanks! Could you please add a test that shows the use case and how the current logic is not enough, and that would fail on `master` and would be solved by this?

leonarf:
I'm sorry, but my usecase is too old, and I can't remember what object I was encoding to json, and why it was failing.
But my PR just adds some logging for debugging purpose. Although, if you think it's dangerous to merge it, you can close the PR and next time I face this usecase, I'll try to write a test.

tiangolo:
Thanks for the response! I'll then close it for now. :nerd_face: :coffee:

PR #10951, opened 2024-01-12
Allow multiple swagger css and js

ajatkj:
Raised this PR based on discussion [here](https://github.com/tiangolo/fastapi/discussions/10927#discussioncomment-8090100).

Basically, this PR allows users to add multiple swagger `css` and `js` urls in `get_swagger_ui_html` function.

This is specially useful if you want to override things on swagger UI. Check out my comment linked above for example.

### DIFF
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -1,5 +1,5 @@
 import json
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union
 
 from fastapi.encoders import jsonable_encoder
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks! But I prefer to show people how to fully customize their docs UI any way they want as in here: https://fastapi.tiangolo.com/how-to/custom-docs-ui-assets/#self-hosting-javascript-and-css-for-docs

If someone, for example you, wanted to have a more customized docs UI, it's possible to entirely override the whole HTML, which is more flexible than just the list of CSS and JS files to include.

PR #10952, opened 2024-01-12
🔧  Group dependencies on dependabot updates

Kludex:
https://til.simonwillison.net/github/dependabot-python-setup

Single PR on dependabot.


tiangolo:
Great, thanks! 🚀

PR #11050, opened 2024-01-30
♻️ Flexible router args

k0t3n:
I'm implementing a cache for FastAPI and found the most suitable way to do this is to create a custom router. Unfortunately, routers are not designed for inheritance and extension and I had to copy-paste lots of code to create custom params for a router. 

[Demo](https://github.com/k0t3n/cachepot/blob/3fbc12f85c476b9d7dc2dec23a64528776f952ea/cachepot/routing.py#L848)

Adding support for custom params will make the design cleaner and avoid copying lots of code.

#9627


k0t3n:
@Kludex I found yours #1917. The resolution was
> Then, if someone later has a clear use case that couldn't be solved with the available tools and ideas (like dependencies), I would be happy to review this again (or an alternative).

The cache I showed in the demo cannot be implemented with middleware or dependency:
* _**Dependency**_: runs only before route execution while cache should be set after response.
```python
@app.get('/', dependencies=[CacheDependency()])
def cached_route(cache_dependency):
    response = do_some_stuff()
    cache_dependency.cache(response)  # always call this?
    return response
```

* _**Middlewares**_: can be executed before and after the route, but it's a Starlette's mechanism and I found **no way to get any context about the FastAPI route** or even resolve the associated route. Only a raw request, while we have to know the route it will be addressed to understand if we need to cache it or not.
* _**Decorator**_: besides tests complexity, it's a too tricky way to inject request/response objects into the context if they're missing ([example](https://github.com/long2ice/fastapi-cache/blob/main/fastapi_cache/decorator.py#L121C9-L121C22))

So, I think all of us want to make FastAPI flexible and extendable. The changes are non-breaking, backward-compatible and extendable. Finally, I wrote a lib for the demo and I would be happy to refactor to be nice and clean after this fix is merged.

k0t3n:
Changes in the core are mostly breaking without args/kwargs. For example, [this commit](https://github.com/tiangolo/fastapi/commit/a4aa79e0b4cacc6b428d415d04d234a8c77af9d5#diff-4d573079004a9f3d148baa4658e68e82b8a3d1a95d603fee8177daa92cf65c93R280) broken backward compatibility because of new arg in `solve_dependencies()`

k0t3n:
@Kludex is there any chance to come back to it? I'm feeling inspired by FastAPI, but unextendable Router and inability to override the request processing in routes create a big problem if you want to use the framework in large projects. I know this can be quite annoying, but Django is a great example of expandability and flexibility, I believe FastAPI would only benefit from following the same way.

tiangolo:

I want to implement a big refactor in how routers work that will solve many of these use cases. Precisely to allow better / proper inheritance.

Similar to my comment here: https://github.com/fastapi/fastapi/pull/11053 :point_down: 

---

I have not wanted to close PRs that target the same or similar problems before I have an alternative, but then I have left many PRs open for too long, which is not useful or clear either.

Thank you! 🍰

PR #11053, opened 2024-01-30
♻️ Add `router_class` param

k0t3n:
I'm facing a problem when using a custom router becomes a real hack. The `router_class` param allows to pass custom router class.
The current solution:

```python
import redis.asyncio as redis

from cachepot.constants import CachePolicy
from cachepot.routing import CachedAPIRouter
from cachepot.storages.redis import RedisStorage

app = FastAPI()

app.router = CachedAPIRouter(
    dependency_overrides_provider=app,
    # define dependencies inside overridden router only
    dependencies=(Depends(...),),
)

# hack to finish setting up custom router
app.setup()

client = redis.from_url('redis://127.0.0.1:6379')
storage = RedisStorage(client)

cache_policy = CachePolicy(
    storage=storage,
    key='cached_hello_world',
    ttl=30,
)


# Do not use @app.route() due to incapability
@app.router.get(path='', cache_policy=cache_policy)
async def cached_hello_world():
    return {'result': 'hello, world!'}

```

As you can see, I have to patch app with a new router using
```python
app.router = CachedAPIRouter()
```
and then call 
```python
app.setup()
```
to be able to use it.
Also, direct
```python
@app.router.get()
```
call is required because router doesn't support custom params. I would appreciate it if you would review [this fix](https://github.com/tiangolo/fastapi/pull/11050) too.


k0t3n:
I can also add some docs if there are no objections to the solution

tiangolo:
Thanks for the interest and work on this! :coffee: 

I have not wanted to close PRs that target the same or similar problems before I have an alternative, but then I have left many PRs open for too long, which is not useful or clear either.

For the specific use case, the best would be to document it in a new GitHub Discussion with the specific problem to solve (without details of the implementation) to track the feature request.

Thank you! :cake:

PR #11062, opened 2024-01-31
♻️ Update pyproject.toml with separate optional dependencies

Zlyden-1:
I met an issue that `poetry add fastapi` and `poetry add 'fastapi[all]'` works correctly, but not `poetry add 'fastapi[uvicorn]'`. So I propose to fix it by adding following lines into the `[project.optional-dependencies]` section of `pyproject.toml` in order to let users add only needed dependencies in poetry:
```toml
httpx = [
    "httpx >=0.23.0",
]
jinja = [
    "jinja2 >=2.11.2",
]
python-multipart = [
    "python-multipart >=0.0.5",
]
# and so on for other extra dependencies, full code see in committed changes
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,39 @@ all = [
     "pydantic-settings >=2.0.0",
     "pydantic-extra-types >=2.0.0",
 ]
+httpx = [
+    "httpx >=0.23.0",
+]
+jinja = [
+    "jinja2 >=2.11.2",
+]
+python-multipart = [
+    "python-multipart >=0.0.5",
+]
+itsdangerous = [
+    "itsdangerous >=1.1.0",
+]
+pyyaml = [
+    "pyyaml >=5.3.1",
+]
+ujson = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -66,6 +66,39 @@ all = [
...
     "pydantic-extra-types >=2.0.0",
 ]
+httpx = [
+    "httpx >=0.23.0",
+]
+jinja = [
+    "jinja2 >=2.11.2",
+]
+python-multipart = [
+    "python-multipart >=0.0.5",
```
dekkers, line 76: Python-multipart has been bumped to `>=0.0.7` so this needs to be updated here too.

Zlyden-1: Updated, thanks for reminding!

tiangolo:
Thanks for the interest @Zlyden-1! But I prefer not to add all the extras and handle and maintain those version independently. If you only need FastAPI and Uvicorn, I think it would be better to add Uvicorn to your Poetry project.

I also plan on simplifying how you declare the dependencies and which you would have by default.

PR #11086, opened 2024-02-03
⬆️ Upgrade version of Starlette to `>= 0.36.3`

tiangolo:
⬆️ Upgrade version of Starlette to `>= 0.36.3`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
-    "starlette>=0.35.0,<0.36.0",
+    "starlette>=0.36.2,<0.37.0",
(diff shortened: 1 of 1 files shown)


tvasenin:
@tiangolo There was an unintentionally breaking change that affected Starlette from `0.36.0` to `0.37.0`.

Fixed in v0.37.1: https://github.com/encode/starlette/pull/2485

Could you pls bump requiremens to `starlette>=0.31.7` and release FastAPI `v0.109.2`?

alimcmaster1:
I agree - https://github.com/tiangolo/fastapi/pull/11146

@tvasenin - what is the Starlette API deprecation policy. Wondering if we need a max version

tvasenin:
> I agree - #11146
> 
> @tvasenin - what is the Starlette API deprecation policy. Wondering if we need a max version

I'm not aware of Starlette deprecation policy -- though, given the v0.x version scheme, according to SemVer, anything could happen in any release :)

tvasenin:
Newly released `v0.110.0` still has the old dependency -- here's PR #11320 to finally fix this.

PR #11093, opened 2024-02-05
Fix body hiding in openapi schema

istrebitel-1:
- Fixed condition to hide Body from openapi

Now if we set `data: Any= Body(..., include_in_schema=False)` we still can see "body" in schema

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -138,7 +138,11 @@ def get_openapi_operation_request_body(
     ],
     separate_input_output_schemas: bool = True,
 ) -> Optional[Dict[str, Any]]:
-    if not body_field:
+    if (
+        not body_field
+        or not isinstance(body_field.field_info, Body)
+        or not body_field.field_info.include_in_schema
+    ):
         return None
     assert isinstance(body_field, ModelField)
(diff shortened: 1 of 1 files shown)


lacop11:
Ran into this as well. @tiangolo is there something blocking this PR? @istrebitel-1 are you still interested in seeing this merged, or no longer a priority for you? If not I could take a look.

istrebitel-1:
> Ran into this as well. @tiangolo is there something blocking this PR? @istrebitel-1 are you still interested in seeing this merged, or no longer a priority for you? If not I could take a look.

Still need to be merged but not so critical as it affected only DEV envs… @tiangolo waiting for your reply too

YuriiMotov:
@istrebitel-1, thanks for the interest in FastAPI!

Seems that this doesn't work with multiple `Body` parameters:
```python

app = FastAPI()

class Item(BaseModel):
    name: str

class Item2(BaseModel):
    name: str

@app.post("/")
async def post_items(
    param1: Item = Body(include_in_schema=False),
    param2: Item2 = Body(),
):
    pass
```

And with `embed=True` neither (`Body(embed=True, include_in_schema=False)`)

PR #11111, opened 2024-02-08
Sketch of adding response_model_config to routes

lukemetz:
It seems that when creating a route fastAPI drops the top level Config in the pydantic base model for which the response is being created for [at least with pydantic v1.10.x]

This seems to be caused by the ModelField wrapping that fastAPI does.

FastAPI partially exposes the ability to set this config in `create_response_field`, but this is not threaded through up to route definitions.

Would you all be open to threading through this value into all the routes? This code is not tested, but should give the idea -- I wanted to get your thoughts first.

This seems to be a common problem -- https://github.com/tiangolo/fastapi/issues/1186 for example -- and the workaround suggested is to globally set `BaseConfig. arbitrary_types_allowed=True` which is a bit to strong of a fix for my usecase.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -52,7 +52,7 @@ from fastapi.utils import (
     get_value_or_default,
     is_body_allowed_for_status_code,
 )
-from pydantic import BaseModel, BaseConfig
+from pydantic import BaseConfig, BaseModel
 from starlette import routing
 from starlette.concurrency import run_in_threadpool
 from starlette.exceptions import HTTPException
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks @lukemetz for the interest! It would be better to start with a discussion, sharing what you currently want to achieve and what is not already working. That way it's gonna be easier to see the problem and understand the solution. There might be another way to solve it too.

Also, please try with Pydantic v2, if it's already handled by v2, then there's no need to change anything else, as upgrading to v2 would solve it. 🤓

tiangolo:
As this PR had requested changes to be applied but has been inactive for a while, it's now going to be closed. But if there's anyone interested, feel free to create a new PR.

PR #11146, opened 2024-02-14
Update Starlette Min Version

alimcmaster1:
I don't think we need a max version here, I imagine builds against starlette main would help to diagnose any issues prior to release. 

Not sure what Starlette deprecation policy is with respect to breaking changes but we could add a max 0.38 if required.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
-    "starlette>=0.31.7",
+    "starlette>=0.37.1",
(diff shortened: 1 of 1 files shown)


Kludex:
Starlette broke something on FastAPI at some point.

pyproject.toml
```diff
@@ -41,7 +41,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
-    "starlette>=0.36.3,<0.37.0",
+    "starlette>=0.31.7",
```
alex-anthropic, line 44: I think this wants to be 0.37.1, rather than 0.31.7

alimcmaster1: Right you are thanks!

Issue #10998, opened 2024-01-21
Raw docstring (leading `r`) defeats form feed `\f` truncation


Kludex:
### Discussed in https://github.com/tiangolo/fastapi/discussions/10531

<div type='discussions-op-text'>

<sup>Originally posted by **jamesbraza** October 26, 2023</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/pydantic/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from fastapi import FastAPI

app = FastAPI()


@app.post("/lof")
def foo(arg: int = 5) -> int:
    """
    Some function.
    
    \f
    
    Args:
        arg: Some argument

    Returns:
        Some integer.
    """
    return arg
```

### Description

Running `ruff==0.1.3` on this, [`D301`](https://docs.astral.sh/ruff/rules/escape-sequence-in-docstring/) will autofix to make the docstring be a raw string (lead by `r`).

```python
@app.post("/lof")
def foo(arg: int = 5) -> int:
    r"""
    Some function.
    
    \f
    
    Args:
        arg: Some argument

    Returns:
        Some integer.
    """
    return arg
```

However, this breaks the form feed `\f` truncation from working, per https://fastapi.tiangolo.com/advanced/path-operation-advanced-configuration/#advanced-description-from-docstring

I think form feed `\f` truncation should still work even if the docstring is a raw string.

### Operating System

macOS

### Operating System Details

n/a

### FastAPI Version

0.104.0

### Pydantic Version

2.4.2

### Python Version

3.11.5

### Additional Context

_No response_</div>

KaniKim:
I little bit handled this by adding some code.
I don't know what you actually want, but you want like this code situation?

[...]

PR #11149, opened 2024-02-15
Raw docstring #10998

ShreySinha02:
Raw docstring (leading r) defeats form feed \f truncation


codespearhead:
@ShreySinha02 @KaniKim @Kludex , I'm afraid it's not clear to me which use case #10998 is meant to address, because changing the behavior of how a raw string is parsed would mean it's no longer being treated as a raw string.

Besides, as @dolfinus pointed out in that issue, which I encourage you to [link to this PR btw](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), both the proposed fix and this one will act out if you mention `C:\files\data` in the description.

Either way, here's some data to help make you guys' explanations more clear.

## Minimal Reproducible Example

1. Make sure you have Git, Python 3, PIP 3 and virtualenv installed on your system:

```bash
git --version &&
python3 -V &&
pip3 -V &&
pip3 show virtualenv
```

2. Create a Python virtual environment and activate it:

```bash
python3 -m venv .venv &&
source .venv/bin/activate
```

3. Set up the demo:

```bash
pip install fastapi &&
git init &&
curl https://raw.githubusercontent.com/github/gitignore/main/Python.gitignore -o .gitignore
```

4. Add the demo entrypoint and run it:

```bash
cat > main.py <<EOF && python main.py && git add .
from fastapi import FastAPI
from json import dump

app = FastAPI()

@app.post("/foo")
def foo(arg: int = 5) -> int:
    r"""
    Some function.
    
    \f
    
    Args:
        arg: Some argument

    Returns:
        Some integer.
    """
    return arg

@app.post("/bar")
def foo(arg: int = 5) -> int:
    r"""
    Some function. C:\files\data C:/files/data
    
    \f
    
    Args:
        arg: Some argument

    Returns:
        Some integer.
    """
    return arg

openapi_schema = app.openapi()
with open("openapi.json", "w") as file:
    dump(openapi_schema, file, indent=2)
EOF
```

## Outputs

[v0.110.1.json](https://github.com/tiangolo/fastapi/files/14842260/v0.110.1.json)
[pr-11149.json](https://github.com/tiangolo/fastapi/files/14842261/pr-11149.json)

## Git Diff

```diff
# git diff v0.110.1.json pr-11149.json

diff --git a/v0.110.1.json b/pr-11149.json
index f8d2a52..249b92b 100644
--- a/v0.110.1.json
+++ b/pr-11149.json
@@ -8,7 +8,7 @@
     "/foo": {
       "post": {
         "summary": "Foo",
-        "description": "Some function.\n\n\\f\n\nArgs:\n    arg: Some argument\n\nReturns:\n    Some integer.",
+        "description": "Some function.",
         "operationId": "foo_foo_post",
         "parameters": [
           {
@@ -50,7 +50,7 @@
     "/bar": {
       "post": {
         "summary": "Foo",
-        "description": "Some function. C:\\files\\data C:/files/data\n\n\\f\n\nArgs:\n    arg: Some argument\n\nReturns:\n    Some integer.",
+        "description": "Some function. C:",
         "operationId": "foo_bar_post",
         "parameters": [
           {
```

YuriiMotov:
Agree with @codespearhead and @dolfinus.

I think it's incorrect to treat r"\f" as form feed character. It's not.

I think the best way is to add a note in docs that r"\f" is not equal to "\f" and if one wants to use truncating by form feed character they shouldn't use raw string and should disable ruff rule for this line.

tiangolo:
Yep, agreed with @YuriiMotov, `r"\f"` and `"\f"` are two different things.

The linter rule is assuming that you are *not* intending to use a form feed, but in this case, you are. So the rule here is wrong, if you want to use the form feed character, you should disable that rule for that line.

---

Additionally, the idea of the `\f` form feed character is already a bit weird to me, I never use it.

I have never seen a need to separate the docs of a path operation function and some additional docstring part of it that is in some way not part of the docstring.

I would actually suggest not using that, but anyway, some people want it, so we'll keep the functionality there, but it wouldn't make sense to have some extra custom behavior when the rule is the one that is being a bit weird here.

Given that, I'll pass on this one, but thanks for the effort! :coffee:

PR #11152, opened 2024-02-16
🔧 Add changelog URL to `pyproject.toml`, shows in PyPI

Pierre-VF:
The URL of the changelog was added to the project URLs in order to make it easier to see latest changes when coming from PyPi. The english version was used as a target, given that all other links are to english materials.


Kludex (approved):

Pierre-VF:
> That's a good idea. Will it show up here?
> image
> 
> 

Yes, that's exactly the point 🙂

alejsdev (approved):

alejsdev:
I updated the URL to go to the Changelog in the official docs :smile: :rocket:

pyproject.toml
```diff
@@ -49,6 +49,7 @@ dependencies = [
 [project.urls]
 Homepage = "https://github.com/fastapi/fastapi"
 Documentation = "https://fastapi.tiangolo.com/"
+Changelog = "https://fastapi.tiangolo.com/release-notes/"
 Repository = "https://github.com/fastapi/fastapi"
```
frost-nzcr4, line 53: I also suggest adding an `Issues = "https://github.com/fastapi/fastapi/issues"` to not return to this a second time.

frost-nzcr4 (changes requested):

pyproject.toml
Pierre-VF: Good idea ! I just added the suggested link

frost-nzcr4 (approved):

tiangolo:
Nice, thank you! :rocket:

PR #11160, opened 2024-02-18
🐛 Fix root_path behavior

richin13:
**Description**

I came across a bug with setting the `root_path` value directly into the `FastAPI` app. The docs state:

> Passing the `root_path` to `FastAPI` would be the equivalent of passing the `--root-path` command line option to Uvicorn or Hypercorn.

But that wasn't the case since providing the `--root-path` option to Uvicorn or Hypercorn makes it so that the scope's `path` and `raw_path` contain the `root_path`. The behavior in FastAPI was to patch the scope's `root_path` value only, leaving the `path` and `raw_path` unchanged.

When running behind a proxy, this would cause the 307 redirect on missing trailing slashes to set the `Location` header to the path without the `root_path` value, resulting in a 404 error.


insistence:
@tiangolo Excuse me, would it be convenient for you to prioritize this PR?

jmmk:
I ran into the same thing after upgrading FastAPI.

Setup:

Reverse Proxy at `/api` that strips the `/api` prefix
FastAPI with `root_path="/api"`
Router with prefix like `/my-router` and a handler at `/`, creating a route like `/my-router/`

Symptoms:

A request to `/api/my-router` returns a 307 redirect to `/my-router/` which will hit our reverse proxy with `/my-router/` and match nothing or whatever happens to be configured at that path
A request to `/api/api/my-router` (with an extra `/api` in there) returns a 307 redirect to `/api/my-router/` which matches
A request to `/api/my-router/` will match the handler and return the expected response

dolfinus:
> Reverse Proxy at /api that strips the /api prefix

You can setup reverse proxy to not strip that prefix, and run FastAPI with `--root-path=/api`

wynneevans:
This is issue is also affecting my team. We wanted to update from pydantic v1 to the latest v2 version. When updating pydantic from v1 to v2 and using FastAPI in the same project, pip updates FastAPI accordingly. In our cases that was 0.95.1 to 0.112.0. This issue was a horrible gotcha because we had to update pydantic model code during this upgrade, so it was natural to suspect pydantic not fastapi was causing the error somehow. Now we have isolated the issue to updating fastapi from 0.108.0 to 0.109.0. We have pydantic 2.8.2 installed.

def create_app() -> FastAPI:
    """Create service app (FastAPI object)"""
 
    logger.info("Initializing web application...")
    get_configs(Path(__file__).parent.parent / 'config_env.ini', config_type='env')
    app = FastAPI(root_path='/index-analytics/profgen')   #  this line only works with fastapi up to v0.108.0
    #app = FastAPI(root_path='')   #  this line when uncommented works with fastapi v0.109.0
    app.mount('/v0', v0.create_app())  #  v0.create_app() involves the proxy
 
    @app.get('/echo')
    def echo() -> dict:
        return {'message': 'success'}
 
    return app

Would this pull request allow the line containing "root_path='index-analytics/profgen'" to work with fastapi v0.109.0+ ?

Soufian20:
Any updates on that?

alligatorbait (commented):
# Review

## Overall Comments

**The good**: This change allows the path proxy to function properly for the expected endpoints, both locally where the proxy is not used, and when a path proxy is used.

**The potential problem**: The reason it works in both environments is through a duplication of endpoints where the the path to all endpoints is created with just the mount path and then another endpoint created with `root_path` + mount path.
**example:**
- An application has an endpoint `/myendpoint`
- Proxy uses a `root_path` set as `/myendpoint`
- Mount path is `/`

**Results:**
- `/myendpoint` is created as an endpoint on `/` mount path
- `/myendpoint/` is created as the base path using `root_path`, there could be conflicts with the endpoint created prior vs this combined `root_path` + mount path.

## Local Test

### Local Test Conditions:

- `root_path` set as `/my/path` via FastAPI keyword argument `root_path`
- Sub-Application mount of `/webapi`

### Local Test Results

- Accessed openapi docs endpoint via `localhost:/webapi/docs`
- openapi.json was found at two endpoints `/my/path/webapi/openapi.json` and `/webapi/openapi.json`
- All openapi endpoints appear to be available on the combined `root_path` and mount path and just mount path endpoints. i.e. `/my/path/webapi/docs` and `/webapi/docs`
- Displayed paths are the `root_path` + mount path `/my/path/webapi` on the openapi docs pages. Executed REST calls display the `root_path` + mount path and make calls to the combined path endpoints.

## Nginx Kubernetes Proxied Test

### Nginx Test Conditions

- `root_path` set as `/my/path` via FastAPI keyword argument `root_path`
- Sub-Application mount of `/webapi`

### Nginx Test Results

- Accessed openapi docs endpoint via `remotehost:/my/path/webapi/docs` and at `remotehost:/my/path/my/path/webapi/docs`.
- openapi.json was found at two endpoints `remotehost:/my/path/webapi/openapi.json` and `/my/path/my/path/webapi/openapi.json`
- All openapi endpoints appear to be available on the combined `root_path` and mount path as well as `root_path` + `root_path` + mount path . i.e. `/my/path/webapi/docs` and `/my/path/my/path/webapi/docs`
- Displayed paths are the `root_path` + mount path `/my/path/webapi` on the openapi docs pages. Executed REST calls display the `root_path` + mount path and make calls to the combined path endpoints.

kunesj:
I can confirm that the bug this fixes still exists in FastAPI 0.115.12.
In my case, it is causing 404 errors when accessing mounted `StaticFiles`. I hope it can be merged soon.

fastapi/applications.py
```diff
@@ -1050,7 +1050,21 @@ async def redoc_html(req: Request) -> HTMLResponse:
...
+                logger.warning(
+                    f"The ASGI server is using a different root path than the one "
+                    f"configured in FastAPI. The configured root path is: "
+                    f"{self.root_path}, the ASGI server root path is: {root_path}. "
+                    f"The former will be used."
+                )
             scope["root_path"] = self.root_path
+            path = scope.get("path")
+            if path and not path.startswith(self.root_path):
+                scope["path"] = self.root_path + path
```
YuriiMotov, line 1064: One more condition is needed here to make sure there is "/" character or end of line after `root_path` prefix:

For now the following will fail because "/apify" starts with "/api":
```python
from fastapi import FastAPI

app = FastAPI(root_path="/api")

@app.get("/apify/")
async def apify():
    return "Ok"


def test_apify():
    client = TestClient(app, base_url="http://testserver")
    response = client.get("/apify", follow_redirects=False)
    assert response.status_code == 307
    assert response.headers["location"] == "http://testserver/api/apify/"

```

fastapi/applications.py
```diff
@@ -1050,7 +1050,21 @@ async def redoc_html(req: Request) -> HTMLResponse:
...
+            raw_path: bytes | None = scope.get("raw_path")
+            if raw_path and not raw_path.startswith(self.root_path.encode()):
+                scope["raw_path"] = self.root_path.encode() + raw_path
```
YuriiMotov, line 1067: Do we really need to change this?

> raw_path (byte string) – The original HTTP path component, excluding any query string, unmodified from the bytes that were received by the web server. 

https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope

YuriiMotov (commented):
I think this looks a bit like a workaround, but it solves long-lasting issues (see in details).

<details>

* https://github.com/fastapi/fastapi/discussions/11153
* https://github.com/fastapi/fastapi/discussions/11148
* https://github.com/fastapi/fastapi/discussions/11147
* https://github.com/encode/starlette/discussions/2928
* https://github.com/encode/starlette/discussions/2557
* https://stackoverflow.com/questions/79599624/fastapi-application-with-nginx-staticfiles-not-working
* https://github.com/fastapi/fastapi/discussions/11033
* https://github.com/fastapi/fastapi/discussions/13590
* https://github.com/fastapi/fastapi/discussions/13421
* https://github.com/fastapi/fastapi/discussions/11959

</details>

Beside problem with redirects, it also solves problem with mounted apps with `root_path`. So, I would also add this test case to avoid breaking it again in the future:

```python
def test_mounted_app_with_root_path():
    app = FastAPI(root_path="/prefix")

    sub_app = FastAPI()
    @sub_app.get("/index")
    async def get_index():
        return {"status": "Ok"}

    app.mount("/sub", sub_app)

    client = TestClient(app, base_url="http://testserver")
    response = client.get("/sub/index")
    assert response.json() == {"status": "Ok"}
```

richin13:
It took them 18 months to review this and the bot closed it in less than 1 month 😂

YuriiMotov:
> It took them 18 months to review this and the bot closed it in less than 1 month 😂

Hello @richin13,

I understand your frustration - waiting 18 months and then having the PR closed by a bot doesn’t feel great.

Right now, we have a lot of open PRs, and to manage them better, we’ve set time limits for replies. If there’s no update from the author for a while, we close the PR to keep things moving. But that doesn’t mean your work is lost - if the idea is good, we add it to our internal plans and come back to it later when we can.

We really appreciate the effort you put into this, and we hope for your understanding.
If you’re still interested in working on this, please feel free to open a new PR - we’d be happy to work with you!

Thanks again for contributing!

**UPD:** it seems we can reopen this PR. Should I reopen it?

lukehsiao:
We ran into this as well. Our workaround for now is to use the `--root-path` command line option.

It seems https://fastapi.tiangolo.com/advanced/behind-a-proxy/?h=behind#setting-the-root-path-in-the-fastapi-app:

> Passing the root_path to FastAPI would be the equivalent of passing the --root-path command line option to Uvicorn or Hypercorn.

is not true.

PR #11168, opened 2024-02-20
✏️ Fix import typo in reference example for `Security`

0shah0:
#### Description: 
  Seems the example does not import `Security` from fastapi instead points to `Depends`


YuriiMotov (approved):

tiangolo:
Good catch, thank you! :mag:

PR #11190, opened 2024-02-24
🔧 Expose examples at swagger

jujumilk3:
Expose `examples` attributes at swagger docs.

If I use example at param, it warns `'example' has been deprecated, please use 'examples' instead`.
But `Query param's examples attributes` are couldn't see at swagger docs.

This PR exposes `Query param's examples attributes` at swagger docs.


tiangolo:
Thanks for the interest! You can do it now with the `openapi_examples` parameter: https://fastapi.tiangolo.com/tutorial/schema-extra-example/#using-the-openapi_examples-parameter

Given that, I'll close this one. :cake:

PR #11191, opened 2024-02-24
🐛 Fix unhandled growing memory for internal server errors, refactor dependencies with `yield` and `except` to require raising again as in regular Python

tiangolo:
🐛 Fix unhandled growing memory for internal server errors, refactor dependencies with `yield` and `except` to require raising again as in regular Python.

This fixes a bug of growing unhandled memory/memory leak reported internally by @rushilsrivastava. :nerd_face:


rushilsrivastava:
Thanks for the patch, sorry I didn't get a chance to open the discussion! 🙌

Issue #4625, opened 2022-02-28
multipart List[float] 422 bugged?


tamis-laan:
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
javascript
@router.post('/add',status_code=200)
async def add(values: List[float] = Form(...), file: UploadFile = File(...)):
    ...
```
```


### Description

Given a post route with a file and array of floats when using the swagger interface '/docs' and posting a file and an array with a SINGLE item everything works. Posting an file and an array with MORE then one value fastAPI returns:

```console
422 Unprocessable Entity
```

Working CURL request:
``` console
curl -X 'POST' \
  'http://localhost:8000/add' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'file=@image.jpg;type=image/jpeg' \
  -F 'value=1'
```

Failing CURL request:
``` console
curl -X 'POST' \
  'http://localhost:8000/add' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'file=@image.jpg;type=image/jpeg' \
  -F 'value=1,2'
```

### Operating System

Linux

### FastAPI Version

0.74.1

### Python Version

3.9.10

tamis-laan:
If I change `List[float] = Form(...)` to `List = Form(...)` and use the swagger interface with: 
``` console
curl -X 'POST' \
  'http://localhost:8000/add' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'values=1,2,3' \
  -F 'file=@file.jpg;type=image/jpeg'
```
I get the following:
``` javascript
> print(values)
['1,2,3']
```
Clearly wrong...

How do I send a file and an array using fastAPI?

[...]

PR #11194, opened 2024-02-25
🐛 Fix validation error when `File` is declared after `Form` parameter

thomasleveil:
This PR updates the fix (6185a92) from @jsg921019 so that tests pass with the latest FastAPI code (0.110.0).
It also adds a test to cover the case with multiple files, and another to ensure the order of parameters in an endpoint definition does not affect the way FastAPI handles the request.

**Problem**
When declaring an endpoint which accepts a `File` but also a `Form` parameter, the `File` parameter must be declared first or the endpoint would respond with error 422.

**Context**
- See discussion #9116
- And probably also #4625


YuriiMotov (changes requested):
Nothing critical (except import from typing).

I think it would be better to put all these related tests into one file and get rid of some of them (like `test_post_form_no_file` (and so on) for multiple files)

thomasleveil:
Thank you for the feed back. I pushed force with your suggested changes. Let me know if anything else can be improved.

yezz123 (approved):

YuriiMotov (approved):
The solution works.
Tests cover the issue. Tests fail before changes and pass after

cfreksen:
I am wondering if this PR has been forgotten?

I am not that familiar with the workflow in this repository, so excuse me if these things just takes time or if this has low priority and there have been more important things to spend your time on.

However, I was affected by #9116 which led me to this PR, and since it has some approvals and it hasn't been touched in 9 months, I thought it would be ok to "bump" it.

carl-sauter:
Hey what about this pr? At least there should be a better error message. Wasted my time only for finding this.

YuriiMotov:
@thomasleveil, thanks for contributing to FastAPI!
Would you like to continue working on this PR?
There is a conflict with `master` that needs to be resolved

thomasleveil:
Here you go @YuriiMotov, fix re-written

YuriiMotov:
@thomasleveil, great job!

As the branch that handles list of files modified, I think it should also be covered by tests.
What do you think?
As I see it, it could be just adding 2 more endpoints (`file_list_before_form`, `file_list_after_form`) and adding them to `parametrize`

thomasleveil:
@YuriiMotov 
> As the branch that handles list of files modified

I don't get what you mean or refer to

YuriiMotov:
> I don't get what you mean or refer to

I'm talking about `elif` block:
```python
        elif (
            is_bytes_sequence_field(field)
            and isinstance(field_info, params.File)
            and value_is_sequence(value)
        ):
```
It handles list of files and it was also modified. So, I think it should be covered by tests

thomasleveil:
@YuriiMotov here you go.

YuriiMotov (approved):
LGTM!

tiangolo (approved):
Great, thank you! 🚀 🙌

PR #11200, opened 2024-02-26
✨ Allow decorating `async` routes with class based decorator

zzzzlzzzz:
Hi, I can't use class based decorator with async __call__, so proposes this changes for review.


YuriiMotov:
Alternative solution: #11508

tests/test_endpoint_with_async_class_decorator.py
```diff
@@ -0,0 +1,32 @@
...
+data = {"working": True}
+
+
+@SomeDecorator
+async def route():
+    return data
+
+
+app = FastAPI()
+app.get("/")(route)
```
YuriiMotov:
```suggestion
@app.get("/")
@SomeDecorator
async def route1():
    return {"working": True}
```

tests/test_endpoint_with_async_class_decorator.py
```diff
@@ -0,0 +1,32 @@
...
+        update_wrapper(wrapper=self, wrapped=original_route)
+        self.route = original_route
+
+    async def __call__(self, *args, **kwargs):
+        return await self.route(*args, **kwargs)
+
+
+data = {"working": True}
+
+
```
YuriiMotov:
```suggestion
```

tests/test_endpoint_with_async_class_decorator.py
```diff
@@ -0,0 +1,32 @@
...
+app = FastAPI()
+app.get("/")(route)
+
+client = TestClient(app)
+
+
+def test_endpoint_with_async_class_decorator():
+    response = client.get("/")
+    assert data == response.json()
```
YuriiMotov:
```suggestion
    assert response.json() == {"working": True}
```

YuriiMotov (approved):
LGTM

The idea of PR is that for now we can use function-based decorators to decorate route function:
```python
def noop_wrap(func):
    @wraps(func)
    async def wrapper(*args, **kwargs):
        return await func(*args, **kwargs)

    return wrapper

@app.get("/function-decorator")
@noop_wrap  # This works
async def route2():
    return {"working": True}
```
But we can not use class-based decorators:
```python
class SomeDecorator:
    def __init__(self, original_route):
        update_wrapper(wrapper=self, wrapped=original_route)
        self.route = original_route

    async def __call__(self, *args, **kwargs):
        return await self.route(*args, **kwargs)

@app.get("/class-decorator")
@SomeDecorator  # This doesn't work for now, but this PR fixes this
async def route1():
    return {"working": True}
```

This PR makes it possible to use class-based decorators to decorate route functions.

Implementation looks nice to me (I simplified the test a bit).

Full tests code in details:

<details>


```python
from functools import wraps

from fastapi import FastAPI

app = FastAPI()

def noop_wrap(func):
    @wraps(func)
    async def wrapper(*args, **kwargs):
        return await func(*args, **kwargs)

    return wrapper

@app.get("/function-decorator")
@noop_wrap
async def route2():
    return {"working": True}

client = TestClient(app)

def test_endpoint_with_async_function_decorator():
    response = client.get("/function-decorator")
    assert response.json() == {"working": True}

```

```python
from functools import update_wrapper

from fastapi import FastAPI

app = FastAPI()

class SomeDecorator:
    def __init__(self, original_route):
        update_wrapper(wrapper=self, wrapped=original_route)
        self.route = original_route

    async def __call__(self, *args, **kwargs):
        return await self.route(*args, **kwargs)

@app.get("/class-decorator")
@SomeDecorator
async def route1():
    return {"working": True}

client = TestClient(app)

def test_endpoint_with_async_class_decorator():
    response = client.get("/class-decorator")
    # ValueError: [TypeError("'coroutine' object is not iterable ...

    assert response.json() == {"working": True}
```

</details>

YuriiMotov:
This was fixed by https://github.com/fastapi/fastapi/pull/9555.

Test passes on `master`.

Will be available in FastAPI 0.123.5

PR #11232, opened 2024-03-01
✨ Allow superimposing an additional response's media type in openapi schema

vinicius-oa:
I've seen the following discussion: https://github.com/tiangolo/fastapi/discussions/11217 ; and I thought that it would be handy to superimpose an additional response's media type over the default route's response class's media type that is displayed in _/docs_.

Example:
```python
@app.get(
    "/a",
    responses={
        HTTP_200_OK: {"superimpose": True, "content": {"text/event-stream": {}}},
        HTTP_500_INTERNAL_SERVER_ERROR: {"description": "Error", "model": Error}
    },
)
def a():
    pass
```
with the key _superimpose_, _fastapi's openAPI_ generator will ignore the default media type object set by the route's response class (in this case _JSONResponse_), so in _/docs_ is displayed in the _responses section 200 status code_ only one media type (_text/event-stream_) in the media type dropdown

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -347,7 +347,9 @@ def get_openapi_path(
                         response_schema = {}
 
                 route_responses = {str(k): v for k, v in route.responses.items()}
-                if status_code not in route_responses or not route_responses.get(status_code, {}).get('superimpose'):
+                if status_code not in route_responses or not route_responses.get(
+                    status_code, {}
+                ).get("superimpose"):
                     operation.setdefault("responses", {}).setdefault(
                         status_code, {}
                     ).setdefault("content", {}).setdefault(
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -347,12 +347,14 @@ def get_openapi_path(
                         response_schema = {}
 
                 route_responses = dict((str(k), v) for k, v in route.responses.items())
-                if status_code not in route_responses or not route_responses.get(status_code).get('superimpose'):
+                if status_code not in route_responses or not route_responses.get(
+                    status_code
+                ).get("superimpose"):
                     operation.setdefault("responses", {}).setdefault(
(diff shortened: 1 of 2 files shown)


vinicius-oa:
ping review @tiangolo

YuriiMotov:
I like the proposed approach.

For now you can remove default media type using `"application/json": None`:
```python
@app.get(
    "/a",
    responses={
        HTTP_200_OK: {"content": {"text/event-stream": {}, "application/json": None}}
    },
)
def a():
    pass
```

**But** I don't like the approach with setting it to `None` to remove since the current behavior of removing `null` values from openapi schema causes some other problems (e.g. [null values are discarded from OpenAPI examples](https://github.com/fastapi/fastapi/discussions/8401)) and will probably be changed one day.

Alternative solution could be to introduce special value (e.g. `REMOVE_KEY`) to indicate that this item should be removed:
```python
@app.get(
    "/a",
    responses={
        HTTP_200_OK: {"content": {"text/event-stream": {}, "application/json": REMOVE_KEY}}
    },
)
def a():
    pass
```

YuriiMotov:
What if we call it **override** or **replace** instead of **superimpose**? I think **superimpose** is a bit misleading.

I also think that code should be simplified (IMO, chains of operations harder to read)..

And, there should be probably more tests added. For example with `response_model=SomeModel`

vinicius-oa:
> What if we call it **override** or **replace** instead of **superimpose**? I think **superimpose** is a bit misleading.
> 
> I also think that code should be simplified (IMO, chains of operations harder to read)..
> 
> And, there should be probably more tests added. For example with `response_model=SomeModel`

Yeah, I think _override_/_replace_ suits better. The code just followed surrounding patterns, but that could be changed. For the tests, fair enough, but for now I don't see any willingness to accept third-party pull requests (_306_ open...), so let's wait for @tiangolo's review...

PR #11238, opened 2024-03-04
Resolve the mismatch of 'item' type.

liuyiyangwang:
```
The error message indicates that an object of type "Item" cannot be assigned to a parameter named "_value" which expects a dictionary with string keys and string values in the function "setitem". The message specifies that "Item" is incompatible with "dict[str, str]".
```
The line fake_db[item.id] = item in your code is trying to assign an Item type object to an element of fake_db. However, fake_db is a dictionary, and its values should be of dictionary type, not Item type. When you try to add the item object (which is of Item type) to fake_db, a type mismatch problem occurs.

### DIFF
--- a/docs_src/app_testing/app_b/main.py
+++ b/docs_src/app_testing/app_b/main.py
@@ -34,5 +34,5 @@ async def create_item(item: Item, x_token: str = Header()):
         raise HTTPException(status_code=400, detail="Invalid X-Token header")
     if item.id in fake_db:
         raise HTTPException(status_code=409, detail="Item already exists")
-    fake_db[item.id] = item
+    fake_db[item.id] = item.model_dump()
     return item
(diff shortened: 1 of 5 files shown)


tiangolo:
📝 Docs preview for commit ddf342d60b95e8703c0a0f2394751addf658cc7e at: https://3ae9e3e2.fastapitiangolo.pages.dev

liuyiyangwang:
> 📝 Docs preview for commit [ddf342d](https://github.com/tiangolo/fastapi/commit/ddf342d60b95e8703c0a0f2394751addf658cc7e) at: https://3ae9e3e2.fastapitiangolo.pages.dev

item.model_dump() is supported in pydantic.v2.
item.dict() is supported in pydantic.v1, but it will prompt a deprecation warning in pydantic.v2.
For compatibility, should we still change it to item.dict()?

tiangolo (changes requested):
The problem with this change is that the method `item.model_dump()` is only available in Pydantic v2. And these examples need to run in Pydantic v1 as well.

Another way to solve it would be to update the original `fake_db` object to contain `Item` objects. :nerd_face:

PR #11243, opened 2024-03-04
Validation of pydantic models in query params

mylius:
If a pydantic model is supplied as params and auto cast via fastapi.Depends() the models config is ignored:

Take a look at this example:

```python
app = FastAPI()

class NoExtraFields(BaseModel):
    model_config = ConfigDict(extra="forbid")
    a: int
    b: int

@app.get("/")
def foo(foo: NoExtraFields = Depends()):
    return foo

client = TestClient(app)

response = client.get("/", params={"a": 1, "b": 2, "c": 3})
```
This assert will pass, since the config is ignored. This is a result of all the params field getting validated separately in `solve_dependencies` instead of together as is done with `body_params`. 

While there are workarounds as [discussed here](https://github.com/tiangolo/fastapi/issues/2859), this is not a satisfactory state considering the otherwise great compatibility with pydantic.

This PR addresses this by doing a validation for the model as a whole for query params as well.

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -21,7 +21,7 @@ from fastapi.exceptions import RequestErrorModel
 from fastapi.types import IncEx, ModelNameMap, UnionType
 from pydantic import BaseModel, create_model
 from pydantic.version import VERSION as PYDANTIC_VERSION
-from starlette.datastructures import UploadFile
+from starlette.datastructures import Headers, QueryParams, UploadFile
 from typing_extensions import Annotated, Literal, get_args, get_origin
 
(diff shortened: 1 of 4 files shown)


tiangolo:
Thanks @mylius!

Using `Depends()` with Pydantic models was never officially supported, it was a hack that happened to work, discovered by chance by someone. 😅

But now there's official support for Pydantic models with `Query`, `Cookie`, `Header` (https://github.com/fastapi/fastapi/pull/12199), so that should cover this use case.

Given that, I'll pass on this one, but thanks for the effort! ☕🍰

Issue #10978, opened 2024-01-16
"Behind a Proxy" changed behaviour in the new version 0.109.0 (in Uvicorn)


tiangolo:
### Discussed in https://github.com/tiangolo/fastapi/discussions/10948

This is related to a change in Starlette and Uvicorn, making them more formally follow the ASGI spec, Starlette already does since version 0.35.0 (recently), Uvicorn needs an equivalent change, in process on: https://github.com/encode/uvicorn/pull/2213

More details in the Starlette PR: https://github.com/encode/starlette/pull/2400

<div type='discussions-op-text'>

<sup>Originally posted by **neugeeug** January 12, 2024</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/pydantic/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
uvicorn rest.main:server --host 0.0.0.0 --port 8080 --reload --root-path /proxy

server = FastAPI()

# separate spec at /v1/docs
server.mount("/v1", app_v1)
# separate spec at /v2/docs
server.mount("/v2", app_v2)

#where app_v1, app_v2 is like below
app_v1 = FastAPI(openapi_tags=tags_metadata, title="Pet store")
```

### Description

When I run the uvicorn with the `--root-path` option for the above setup, the server is expecting that I include the defined root path in every request e.g:
 http://localhost:8080/v2/docs is not working (404)
 http://localhost:8080/proxy/v2/docs is working

IN version 0.108.0 It works as described here https://fastapi.tiangolo.com/advanced/behind-a-proxy/

### Operating System

Linux

### FastAPI Version

0.109.0

### Pydantic Version

2.5.3

### Python Version

3.11.5

### Additional Context

_No response_</div>

Kludex:
[...]

PR #11253, opened 2024-03-06
Add test for self-hosted docs behind a proxy

bertomaniac:
Related to #10978, #11033, #11160 and #11622

No code changes here, but adds a test for static assets when using `root_path` (behind a proxy)

Note, this test passes in `<=0.108.0` and fails beginning at `0.109.0`.

### DIFF
--- a/tests/test_offline_docs/static/swagger.css
+++ b/tests/test_offline_docs/static/swagger.css
@@ -1,0 +1,1 @@
+.swagger-ui{color:#3b4151;}
--- a/tests/test_offline_docs/static/swagger.js
+++ b/tests/test_offline_docs/static/swagger.js
@@ -1,0 +1,3 @@
+function foo() {
+    return "foo"
+}
--- a/tests/test_offline_docs/test_root_path_with_static_mount.py
+++ b/tests/test_offline_docs/test_root_path_with_static_mount.py
@@ -1,6 +1,6 @@
 from fastapi import FastAPI
-from fastapi.staticfiles import StaticFiles
 from fastapi.openapi.docs import get_swagger_ui_html
+from fastapi.staticfiles import StaticFiles
 from fastapi.testclient import TestClient
 
 root_path = "/api"
@@ -12,7 +12,10 @@ app = FastAPI(
     redoc_url=None,
 )
 
-app.mount("/static", StaticFiles(directory="tests/test_offline_docs/static"), name="static")
+app.mount(
+    "/static", StaticFiles(directory="tests/test_offline_docs/static"), name="static"
+)
+
 
 @app.get("/")
 async def custom_swagger_ui_html():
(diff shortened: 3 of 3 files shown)


YuriiMotov:
```python

    response = client.get("/static/swagger.js")

    response = client.get("/static/swagger.css")

    assert "/static/swagger.js" in swagger_html
    assert f"{root_path}/static/swagger.js" not in swagger_html
    assert "/static/swagger.css" in swagger_html
    assert f"{root_path}/static/swagger.css" not in swagger_html
```
The lines above are actually wrong. Since our app is behind the proxy with prefix '/api', all assets' paths should start with `root_path` ('/api')

There is an issue (IMO) in Starlette that makes static files inaccessible via path without prefix (those that don't start with `root_path`).
Read more: https://github.com/encode/starlette/discussions/2931

Although, there is still a minor issue here.
The code of `custom_swagger_ui_html` should actually include adding `root_path` prefix to all assets.
So,  I think the [related code example in docs](https://fastapi.tiangolo.com/how-to/custom-docs-ui-assets/#serve-the-static-files) should be updated.

```python
async def custom_swagger_ui_html(request: Request):
    """
    Sets up a localized version of the Swagger (OpenAPI) docs that can be run without assets from the Internet.
    """

    root_path = request.scope.get("root_path", "").rstrip("/")
    return get_swagger_ui_html(
        openapi_url=root_path + "/openapi.json",
        title=app.title,
        swagger_js_url=root_path + "/static/swagger.js",
        swagger_css_url=root_path + "/static/swagger.css",
    )

```

In tests, all paths should be updated to start with `root_path` prefix:
```python
    response = client.get(f"{root_path}/openapi.json")

    response = client.get(f"{root_path}/static/swagger.js")

    response = client.get(f"{root_path}/static/swagger.css")

    assert f"{root_path}/static/swagger.js" in swagger_html
    assert f"{root_path}/static/swagger.css" in swagger_html
    
```

@bertomaniac, would you like to work on this?

bertomaniac:
@YuriiMotov Sure. I could probably look at starting on this in the next couple of weeks.

Just to be clear: You're talking about updating these tests and then updating the documentation as well?

YuriiMotov:
IMO, ideally we just need to update docs example (add `root_path` prefixes to URLs of assets as it's done in the [original endpoint function](https://github.com/fastapi/fastapi/blob/28038f19cfae040c826ad7a89f511fc854d9d44f/fastapi/applications.py#L1014-L1026)) and parameterize test for that code example from docs to use app _with_ and _without_ `root_path` (using `client = TestClient(app=app, root_path=..)`?).

Hope I was clear enough 🙂

Actually, I would close this PR and start new one from scratch to keep it clean and easy to review.

But don't trust me blindly, double-check please

bertomaniac:
@YuriiMotov

I think I got things in order. New PR is here: https://github.com/fastapi/fastapi/pull/13825

YuriiMotov:
Closing this in favor of new #13825

PR #11255, opened 2024-03-07
🧑‍💻 Define strict return value typing for `Depends` and `Security`

JaniM:
Depends can be used with a direct form, which although not recommended, is still supported. The following code is problematic wrt. type checking.

```
def dependency() -> str:
    return "dependency"

@router.get("/")
def endpoint(value: int = Depends(dependency)):
    pass
```

Currently, `Depends(dependency)` returns Any, which then happily gets assigned to the int. This patch changes it to return a type matching what the dependency returns, making the above code fail type checking with mypy  as it should.

### DIFF
--- a/fastapi/param_functions.py
+++ b/fastapi/param_functions.py
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Dict, List, Optional, Sequence, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, TypeVar, Union, cast
 
 from fastapi import params
 from fastapi._compat import Undefined
@@ -2217,9 +2217,12 @@ def File(  # noqa: N802
     )
 
 
+DependencyResult = TypeVar("DependencyResult")
+
+
 def Depends(  # noqa: N802
     dependency: Annotated[
-        Optional[Callable[..., Any]],
+        Optional[Callable[..., DependencyResult]],
         Doc(
             """
             A "dependable" callable (like a function).
@@ -2244,7 +2247,7 @@ def Depends(  # noqa: N802
             """
         ),
     ] = True,
-) -> Any:
+) -> DependencyResult:
     """
     Declare a FastAPI dependency.
 
@@ -2274,12 +2277,14 @@ def Depends(  # noqa: N802
         return commons
     ```
     """
-    return params.Depends(dependency=dependency, use_cache=use_cache)
+    return cast(
+        DependencyResult, params.Depends(dependency=dependency, use_cache=use_cache)
+    )
 
 
 def Security(  # noqa: N802
     dependency: Annotated[
-        Optional[Callable[..., Any]],
+        Optional[Callable[..., DependencyResult]],
         Doc(
             """
             A "dependable" callable (like a function).
@@ -2321,7 +2326,7 @@ def Security(  # noqa: N802
             """
         ),
     ] = True,
-) -> Any:
+) -> DependencyResult:
     """
     Declare a FastAPI Security dependency.
 
@@ -2357,4 +2362,7 @@ def Security(  # noqa: N802
         return [{"item_id": "Foo", "owner": current_user.username}]
     ```
     """
-    return params.Security(dependency=dependency, scopes=scopes, use_cache=use_cache)
+    return cast(
+        DependencyResult,
+        params.Security(dependency=dependency, scopes=scopes, use_cache=use_cache),
+    )


YuriiMotov (approved):
I think it's a good fix!

I would only change the title and description to make things clearer.
In fact, this fix does not cause FastAPI to show any errors if the returned dependency type does not match the parameter type. But this fix allows you to find this error using `mypy`.

With this fix, 'mypy' will show `error: Incompatible default for argument "value" (default has type "str", argument has type "int")  [assignment]`.

UPD: why don't you add the same to the `Security` as well?

JaniM:
The checks seem to be stuck, can someone give them a nudge?

fastapi/param_functions.py
```diff
@@ -2274,12 +2277,14 @@ async def read_items(commons: Annotated[dict, Depends(common_parameters)]):
         return commons
     ```
     """
-    return params.Depends(dependency=dependency, use_cache=use_cache)
+    depends_any: Any = params.Depends(dependency=dependency, use_cache=use_cache)
+    depends: DependencyResult = depends_any
+    return depends
```
Zaczero, line 2282: Why can't we keep the old return login unchanged? I believe the type annotation on the method itself should be enough.

JaniM: return params.Depends(dependency=dependency, use_cache=use_cache)
Fails mypy type checking, because `Depends` is not assignable to the arbitrary `DependencyResult`.

    depends_any: Any = params.Depends(dependency=dependency, use_cache=use_cache)
    return depends_any
Would pass mypy, but not the stronger lints the CI has, failing due to trying to return an `Any` when an `Any` isn't expected.

This whole thing is a bit of a hack, but so is the way `Depend` is used.

Zaczero:
```py
    depends: Any = params.Depends(dependency=dependency, use_cache=use_cache)
    depends: DependencyResult
    return depends
```

Do you know if that works? It follows that idea but skips the assignment.

JaniM: Easy enough to test.

```
fastapi/param_functions.py:2281: error: Name "depends" already defined on line 2280  [no-redef]
fastapi/param_functions.py:2282: error: Returning Any from function declared to return "DependencyResult"  [no-any-return]
```

Zaczero: Thank you. I have one last thought.

> Would pass mypy, but not the stronger lints the CI has, failing due to trying to return an Any when an Any isn't expected.

But do linters check external libraries and complain on them? This project does not have strict type requirements (I think). So is there actually any benefit to fixing the behavior here? Or does it only affect FastAPI contributors?

JaniM: `The CI mentioned is FastAPI's own lint check. Although I was a bit misguided on that comment - it does fail mypy when run with FastAPI's configuration. I just had laxer configuration loaded somehow.

YuriiMotov: You can do as followed:
```
    return cast(
        DependencyResult, params.Depends(dependency=dependency, use_cache=use_cache)
    )
```

JaniM: Good idea, I made the change.

YuriiMotov (approved):
Good useful fix!

jsimonlane:
Good fix! Hoping this gets merged soon...

YuriiMotov:
@JaniM, could you please update the branch?

YuriiMotov:
Similar [PR](https://github.com/fastapi/fastapi/pull/10053) was closed with the [comment](https://github.com/fastapi/fastapi/pull/10053#issuecomment-2294632661):
> Now you can (and probably should) use Annotated, that should simplify a lot of it.
> Also, this would probably have a conflict with dependency classes 🤔

And, there is one more similar open PR #5856 (different approach)

Related discussions:
* https://github.com/fastapi/fastapi/discussions/12113
* https://github.com/fastapi/fastapi/discussions/9338
* https://github.com/fastapi/fastapi/discussions/6071
* https://github.com/fastapi/fastapi/discussions/8453

tiangolo:

As @YuriiMotov says, the recommended approach now is to use `Annotated`, this solves this and other issues, so it should be better.

There's also other case of class dependencies, where `Depends()` is empty, and having these stricter type annotations probably conflicts with that, so I'll pass on this one for now. :cake:

PR #11256, opened 2024-03-07
Performance improvements for jsonable_encoder

Zaczero:
Hello! This is my first contribution to FastAPI so don't hesitate to point out any issues so I can improve. Today I have focused on improving the performance of **jsonable_encoder** as it seems to be a hot path in many applications.

Given this [test file](https://openaedmap.org/api/v1/countries/PL.geojson), there seems to be about ~14% performance gain. Most of which come from dict/list comprehensions and reduced set operations.

**Before**
```sh
python -m timeit -n 15 -s 'from fastapi.encoders import jsonable_encoder' -s 'from pathlib import Path' -s 'import json' -s 'data = json.loads(Path("PL.geojson").read_bytes())' 'jsonable_encoder(data)'
```
> 15 loops, best of 5: 145 msec per loop

**After**
```sh
python -m timeit -n 15 -s 'from fastapi.encoders import jsonable_encoder' -s 'from pathlib import Path' -s 'import json' -s 'data = json.loads(Path("PL.geojson").read_bytes())' 'jsonable_encoder(data)'
```
> 15 loops, best of 5: 125 msec per loop

Obviously, your experience may vary depending on the environment and the data. Having said that, I believe those are safe and widely applicable optimizations.

Feel free to ask questions - I am very active on GitHub and I tend to respond quickly.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -262,11 +262,6 @@ def jsonable_encoder(
         return obj
     if isinstance(obj, dict):
         encoded_dict = {}
-        allowed_keys = set(obj.keys())
-        if include is not None:
-            allowed_keys &= set(include)
-        if exclude is not None:
-            allowed_keys -= set(exclude)
         for key, value in obj.items():
             if (
                 (
@@ -275,7 +270,8 @@ def jsonable_encoder(
                     or (not key.startswith("_sa"))
                 )
                 and (value is not None or not exclude_none)
-                and key in allowed_keys
+                and (include is None or key in include)
+                and (exclude is None or key not in exclude)
             ):
                 encoded_key = jsonable_encoder(
                     key,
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for the interest! But I don't feel comfortable with the code changes and the stability we could get from that... also, we would lose the type annotations and won't be able to detect errors from that. So I prefer to pass on this one.

I want to optimize the jsonable_enoder later after some profiling, but for now, I'll pass on this one.

But thanks for the interest! :cake:

Zaczero:
> we would lose the type annotations and won't be able to detect errors from that

Hey, changes to ENCODERS_BY_TYPE can be reverted as they don't impact most of the performance gains in this PR. Improvements to jsonable_encoder are the key - and they are pretty well typed. They don't change the flow of the application in any way, but rather use a more optimized syntax. Let me know if you change your mind :+1:

ramnes:
@tiangolo, PR #5718 makes `jsonable_encoder` 40% faster with minimal code changes.

PR #11301, opened 2024-03-16
Secure Source of Randomness

gjskha:
Replaces the random.random() function with the more secure function secrets.SystemRandom().random()

### DIFF
--- a/.github/actions/notify-translations/app/main.py
+++ b/.github/actions/notify-translations/app/main.py
@@ -1,5 +1,4 @@
 import logging
-import random
 import sys
 import time
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


dolfinus:
Why? This is a part of Github automatization related to translations, not a part of FastAPI source code

Zaczero:
> Why? This is a part of Github automatization related to translations, not a part of FastAPI source code

Also I think in this case, faster random is a slightly better choice. There seems to be nothing cryptographic about this sleep_time.

tiangolo:
Thanks for the interest @gjskha, but yeah, as others point out, this is not needed here. So I'll close this one. But thanks for the interest! :coffee:

PR #11306, opened 2024-03-17
Support Pydantic root model as query parameter

### LINKED ISSUE #11134: Cannot use simple models derived from str (with Pydantic RootModel) as query params
### Discussed in https://github.com/tiangolo/fastapi/discussions/11101

<sup>Originally posted by **WarpedPixel** February  6, 2024</sup>
### First Check

### Commit to Help

### Example Code

```python
# fail.py
# uvicorn fail:app --reload

from fastapi import FastAPI
from pydantic import RootModel

SimpleID = RootModel[str]

app = FastAPI()

@app.get("/test1")
async def t1(param1: SimpleID):     # BUG: expected to be query param, but it is body
    return {"message": f"Hello {param1}"}

# @app.get("/test2")
# async def t2(param1: Annotated[SimpleID, Query()]): # this workaround does NOT work
#     return {"message": f"Hello {param1}"}

@app.get("/test3")
async def t3(param1: SimpleID, param2: str):  # param2 is query param as expected, param1 is NOT
    return {"message": f"Hello {param1} {param2}"}
```

### Description

Simple types should be usable as query params. Pydantic provides `RootModel` (instead of `BaseModel`) precisely to implement types with rich validation based on native types like `list` or `str` (instead of the more typical object/dict). Types built with `RootModel` do not seem to work as query params, *even if it is functionally equivalent* to a simple type like `str`.
Furthermore, we cannot force them to work as query params with `Query()` annotations either.
**NOTE:** the example is overly simplified, my RootModel-derived types do much more in terms of validation.

### Operating System

macOS

### Operating System Details

_No response_

### FastAPI Version

0.108.0

### Pydantic Version

2.5.3

### Python Version

Python 3.11.4

### Additional Context

I am not sure this is possible, but I would like for FastAPI to understand RootModel derived types, if the root type is one of the types that can be used as query param, then it should work as such.
The workaround in my case is to use `str` parameters in my FastAPI calls then manually convert to the rich Pydantic type that does validation. Which kind of defeats the purpose of FastAPI rich type annotations.</div>

MarkusSintonen:
Fixes https://github.com/tiangolo/fastapi/issues/11134

Adds support for Pydantic root model in query/path parameters. Supports PydaticV2 `RootModel[X]` and PydanticV1 `BaseModel.__root__`. This is a non breaking change as previously root models usage threw errors on route initialization when they were not used as bodies.

Adds plenty of root model related tests.

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -50,7 +50,7 @@ Url: Type[Any]
 
 if PYDANTIC_V2:
     from pydantic import PydanticSchemaGenerationError as PydanticSchemaGenerationError
-    from pydantic import TypeAdapter
+    from pydantic import RootModel, TypeAdapter
     from pydantic import ValidationError as ValidationError
     from pydantic._internal._schema_generation_shared import (  # type: ignore[attr-defined]
         GetJsonSchemaHandler as GetJsonSchemaHandler,
@@ -292,6 +292,12 @@ if PYDANTIC_V2:
             for name, field_info in model.model_fields.items()
         ]
 
+    def root_model_inner_type(annotation: Any) -> Any:
+        if lenient_issubclass(annotation, RootModel):
+            inner = annotation.model_fields["root"].annotation
+            return root_model_inner_type(inner) or inner
+        return None
+
 else:
     from fastapi.openapi.constants import REF_PREFIX as REF_PREFIX
     from pydantic import AnyUrl as Url  # noqa: F401
@@ -453,7 +459,10 @@ else:
     def _model_dump(
         model: BaseModel, mode: Literal["json", "python"] = "json", **kwargs: Any
     ) -> Any:
-        return model.dict(**kwargs)
+        serialized = model.dict(**kwargs)
+        if isinstance(serialized, dict) and "__root__" in serialized:
+            return serialized["__root__"]
+        return serialized
 
     def _get_model_config(model: BaseModel) -> Any:
         return model.__config__  # type: ignore[attr-defined]
@@ -495,12 +504,21 @@ else:
         )
 
     def is_scalar_field(field: ModelField) -> bool:
+        if (inner := root_model_inner_type(field.type_)) is not None:
+            return field_annotation_is_scalar(inner)
+
         return is_pv1_scalar_field(field)
 
     def is_sequence_field(field: ModelField) -> bool:
+        if (inner := root_model_inner_type(field.type_)) is not None:
+            return field_annotation_is_sequence(inner)
+
         return field.shape in sequence_shapes or _annotation_is_sequence(field.type_)  # type: ignore[attr-defined]
 
     def is_scalar_sequence_field(field: ModelField) -> bool:
+        if (inner := root_model_inner_type(field.type_)) is not None:
+            return field_annotation_is_scalar_sequence(inner)
+
         return is_pv1_scalar_sequence_field(field)
 
     def is_bytes_field(field: ModelField) -> bool:
@@ -531,6 +549,15 @@ else:
     def get_model_fields(model: Type[BaseModel]) -> List[ModelField]:
         return list(model.__fields__.values())  # type: ignore[attr-defined]
 
+    def root_model_inner_type(annotation: Any) -> Any:
+        if (
+            lenient_issubclass(annotation, BaseModel)
+            and "__root__" in annotation.__fields__
+        ):
+            inner = annotation.__fields__["__root__"].annotation
+            return root_model_inner_type(inner) or inner
+        return None
+
 
 def _regenerate_error_with_loc(
     *, errors: Sequence[Any], loc_prefix: Tuple[Union[str, int], ...]
@@ -550,6 +577,9 @@ def _annotation_is_sequence(annotation: Union[Type[Any], None]) -> bool:
 
 
 def field_annotation_is_sequence(annotation: Union[Type[Any], None]) -> bool:
+    if (inner := root_model_inner_type(annotation)) is not None:
+        return field_annotation_is_sequence(inner)
+
     origin = get_origin(annotation)
     if origin is Union or origin is UnionType:
         for arg in get_args(annotation):
@@ -573,7 +603,25 @@ def _annotation_is_complex(annotation: Union[Type[Any], None]) -> bool:
     )
 
 
+def field_annotation_is_root_model(annotation: Union[Type[Any], None]) -> bool:
+    origin = get_origin(annotation)
+    if origin is Union or origin is UnionType:
+        return any(field_annotation_is_root_model(arg) for arg in get_args(annotation))
+
+    return root_model_inner_type(annotation) is not None
+
+
+def resolves_to_basemodel(annotation: Union[Type[Any], None]) -> bool:
+    if (inner := root_model_inner_type(annotation)) is not None:
+        return resolves_to_basemodel(inner)
+
+    return lenient_issubclass(annotation, BaseModel)
+
+
 def field_annotation_is_complex(annotation: Union[Type[Any], None]) -> bool:
+    if (inner := root_model_inner_type(annotation)) is not None:
+        return field_annotation_is_complex(inner)
+
     origin = get_origin(annotation)
     if origin is Union or origin is UnionType:
         return any(field_annotation_is_complex(arg) for arg in get_args(annotation))
@@ -587,11 +635,17 @@ def field_annotation_is_complex(annotation: Union[Type[Any], None]) -> bool:
 
 
 def field_annotation_is_scalar(annotation: Any) -> bool:
+    if (inner := root_model_inner_type(annotation)) is not None:
+        return field_annotation_is_scalar(inner)
+
     # handle Ellipsis here to make tuple[int, ...] work nicely
     return annotation is Ellipsis or not field_annotation_is_complex(annotation)
 
 
 def field_annotation_is_scalar_sequence(annotation: Union[Type[Any], None]) -> bool:
+    if (inner := root_model_inner_type(annotation)) is not None:
+        return field_annotation_is_scalar_sequence(inner)
+
     origin = get_origin(annotation)
     if origin is Union or origin is UnionType:
         at_least_one_scalar_sequence = False
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -31,6 +31,7 @@ from fastapi._compat import (
     copy_field_info,
     create_body_model,
     evaluate_forwardref,
+    field_annotation_is_root_model,
     field_annotation_is_scalar,
     get_annotation_from_field_info,
     get_cached_model_fields,
@@ -43,6 +44,7 @@ from fastapi._compat import (
     is_uploadfile_or_nonable_uploadfile_annotation,
     is_uploadfile_sequence_annotation,
     lenient_issubclass,
+    resolves_to_basemodel,
     sequence_types,
     serialize_sequence_value,
     value_is_sequence,
@@ -219,7 +221,7 @@ def _get_flat_fields_from_params(fields: List[ModelField]) -> List[ModelField]:
     if not fields:
         return fields
     first_field = fields[0]
-    if len(fields) == 1 and lenient_issubclass(first_field.type_, BaseModel):
+    if len(fields) == 1 and resolves_to_basemodel(first_field.type_):
         fields_to_extract = get_cached_model_fields(first_field.type_)
         return fields_to_extract
     return fields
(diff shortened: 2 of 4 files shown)


fastapi/_compat.py
```diff
@@ -478,7 +489,11 @@ def get_definitions(
         )
 
     def is_scalar_field(field: ModelField) -> bool:
-        return is_pv1_scalar_field(field)
+        from fastapi import params
+
+        return is_pv1_scalar_field(field) and not isinstance(
+            field.field_info, params.Body
```
MarkusSintonen, line 495: This additional `params.Body` is same as in PydanticV2 `def is_scalar_field` side

fastapi/_compat.py
```diff
@@ -230,11 +230,7 @@ def get_definitions(
         return field_mapping, definitions  # type: ignore[return-value]
 
     def is_scalar_field(field: ModelField) -> bool:
-        from fastapi import params
-
-        return field_annotation_is_scalar(
-            field.field_info.annotation
-        ) and not isinstance(field.field_info, params.Body)
```
MarkusSintonen, line 237: This check was missing from v1 side. Lifted the `params.Body` check out from this "scalar field check" into `is_body_param` check which is more appropriate for checking the `params.Body` case.

fastapi/dependencies/utils.py
```diff
@@ -412,7 +413,11 @@ def analyze_param(
             type_annotation
         ) or is_uploadfile_sequence_annotation(type_annotation):
             field_info = params.File(annotation=use_annotation, default=default_value)
-        elif not field_annotation_is_scalar(annotation=type_annotation):
+        elif (
+            not field_annotation_is_scalar(type_annotation)
+            # Root models by default regarded as bodies
```
MarkusSintonen, line 418: Backward compatibility. Root models have been regarded as body params by default eventhough they were scalar.

MarkusSintonen: Another option would be to explicitly require annotating by `Body`/`Query`,etc

MarkusSintonen:
Can I have a review @tiangolo / @alejsdev 🙏

MarkusSintonen:
@tiangolo is there a chance for getting a review for this PR? 🙏

0xSecure:
UP

atatulyan:
Please consider this MR

uliantsev-a:
I wanted this changes! @tiangolo, please do review it.

YuriiMotov:
@MarkusSintonen, thanks for your interest in FastAPI, and sorry for long time waiting for review..

Could you please take a look and resolve the conflict with `master`?

MarkusSintonen:
> @MarkusSintonen, thanks for your interest in FastAPI, and sorry for long time waiting for review..
> 
> Could you please take a look and resolve the conflict with `master`?

Hi! I can if there is actually a chance on getting this reviewed/merged? But I don't want to work on it if there is no activity from maintainer(s). As this has been dead for so long.

YuriiMotov:
> Hi! I can if there is actually a chance on getting this reviewed/merged? But I don't want to work on it if there is no activity from maintainer(s). As this has been dead for so long.

I understand your concerns.
I can't guarantee that it will be merged, but I will review it and if everything is Ok I will pass it to Sebastián for review.
Since this PR addresses open issue, I think it has all chances to be merged.

fastapi/_compat.py
```diff
@@ -494,12 +499,21 @@ def get_definitions(
         )
 
     def is_scalar_field(field: ModelField) -> bool:
+        if (inner := root_model_inner_type(field.type_)) is not None:
```
MarkusSintonen, line 507: _compat checks in V1/V2 work quite differently and I'm not sure is it worth trying to support legacy Pydantic. Could just remove these and let it break with legacy `__root__` models like previously

MarkusSintonen:
Hi @YuriiMotov This is now rebased

fastapi/dependencies/utils.py
```diff
@@ -454,7 +456,11 @@ def analyze_param(
...
```
```suggestion
            # Treat root models as Body parameters by default
```

fastapi/_compat.py
```diff
@@ -572,7 +597,25 @@ def _annotation_is_complex(annotation: Union[Type[Any], None]) -> bool:
...
 
+def field_annotation_is_root_model(annotation: Union[Type[Any], None]) -> bool:
+    origin = get_origin(annotation)
+    if origin is Union or origin is UnionType:
+        return any(field_annotation_is_root_model(arg) for arg in get_args(annotation))
+
+    return root_model_inner_type(annotation) is not None
+
+
+def field_annotation_has_submodel_fields(annotation: Union[Type[Any], None]) -> bool:
```
YuriiMotov: `field_annotation_has_submodel_fields` looks unclear to me..
How about?
* unwraps_to_model
* unwraps_to_basemodel
* resolves_to_basemodel
* is_or_unwraps_to_basemodel

fastapi/dependencies/utils.py
```diff
@@ -752,8 +758,9 @@ def request_params_to_args(
     single_not_embedded_field = False
     default_convert_underscores = True
     if len(fields) == 1 and lenient_issubclass(first_field.type_, BaseModel):
-        fields_to_extract = get_cached_model_fields(first_field.type_)
-        single_not_embedded_field = True
+        if field_annotation_has_submodel_fields(first_field.type_):
+            fields_to_extract = get_cached_model_fields(first_field.type_)
+            single_not_embedded_field = True
```
YuriiMotov: I think we can simplify this part to just:
```py
    if len(fields) == 1 and field_annotation_has_submodel_fields(first_field.type_):
        fields_to_extract = get_cached_model_fields(first_field.type_)
        single_not_embedded_field = True
```
No need to handle case with `RootModel` that doesn't wrap other model (doesn't have sub-fields) here.
If we have only one field and we don't change `fields_to_extract`, then `field.field_info` will be the same that `first_field.field_info` here:

```py
            convert_underscores = getattr(
                field.field_info, "convert_underscores", default_convert_underscores
            )
```

fastapi/encoders.py
```diff
@@ -230,10 +230,16 @@ def jsonable_encoder(
...
-        if "__root__" in obj_dict:
-            obj_dict = obj_dict["__root__"]
+
+        if (
+            not PYDANTIC_V2
+            and isinstance(serialized, dict)
+            and "__root__" in serialized
+        ):
+            serialized = serialized["__root__"]
+
```
YuriiMotov: Wouldn't it be better to move this to `_model_dump` implementation?

fastapi/_compat.py
```diff
@@ -292,6 +292,11 @@ def get_model_fields(model: Type[BaseModel]) -> List[ModelField]:
             for name, field_info in model.model_fields.items()
         ]
 
+    def root_model_inner_type(annotation: Any) -> Any:
+        if lenient_issubclass(annotation, RootModel):
+            return annotation.model_fields["root"].annotation
```
YuriiMotov: Do we need to handle cases with root model that wraps root model?

YuriiMotov (commented):
Great job!
Thanks, @MarkusSintonen !

Just a few cosmetic suggestions in the comments.

Leave it here just not to forget:
* we also need to measure the performance impact (will do it a bit later)
* would be nice to create several use cases where `RootModels` as non-Body parameters shine

fastapi/dependencies/utils.py
MarkusSintonen: Done

fastapi/_compat.py
MarkusSintonen: Done, went with `resolves_to_basemodel`

fastapi/dependencies/utils.py
MarkusSintonen: Done, simplified

fastapi/encoders.py
MarkusSintonen: Done, moved to the V1 handling side

fastapi/_compat.py
MarkusSintonen: Done, added handling for that, also added a test for RootModel inside RootModel case. (It also worked previously but better to handle it already in `root_model_inner_type`)

fastapi/dependencies/utils.py
```diff
@@ -925,7 +931,7 @@ async def request_body_to_args(
 
     if (
         single_not_embedded_field
-        and lenient_issubclass(first_field.type_, BaseModel)
+        and resolves_to_basemodel(first_field.type_)
```
YuriiMotov: This is not covered by tests.
As this PR aims to support Root models for non-body parameters (Query and Path?), do we need to update this here (it's related to body parameters)? I suggest we do it in a separate PR and add some tests there.
Or, we can update the Title and description to make this PR broader and then cover this case by tests

fastapi/_compat.py
```diff
@@ -573,7 +603,25 @@ def _annotation_is_complex(annotation: Union[Type[Any], None]) -> bool:
     )
 
 
+def field_annotation_is_root_model(annotation: Union[Type[Any], None]) -> bool:
+    origin = get_origin(annotation)
+    if origin is Union or origin is UnionType:
+        return any(field_annotation_is_root_model(arg) for arg in get_args(annotation))
```
YuriiMotov, line 609: Side note: this condition is not covered by tests. ToDo: think about use cases that it enables

tests/test_root_model.py
```diff
@@ -0,0 +1,564 @@
...
+
+
+if PYDANTIC_V2:
+    from pydantic import ConfigDict, Field, field_validator, model_serializer
+
+    class FieldWrap(RootModel[str]):
+        model_config = ConfigDict(
+            json_schema_extra={"description": "parameter starts with bar_"}
+        )
+        root: str = Field(pattern=r"^bar_.*$")
```
YuriiMotov, line 45: I did some experiments with `description` and `pattern` parameters and I find it a bit confusing..
We can specify description in 3 places:
* `json_schema_extra={"description": "..."}` - will be displayed by Swagger in components->schemas
* `root: str = Field(description="...")` - will be displayed by Swagger in components->schemas, but description from from json_schema_extra overrides it
* `def query_fieldwrap(q: Annotated[FieldWrap, Query(description="...")]):` - will be displayed by Swagger in parameters

Not sure there is an issue here, but we should make sure why it works this way and decide how we should recommend people specify it.

---

For `pattern` parameter we have 2 options:
* `root: str = Field(pattern=r"^bar_.*$")` - will work (displayed in parameters and in components->schemas, and will be used for validation)
* `async def read_root(name: Param = Query(pattern="^[a-z]+[0-9]$")):` - will be used to generate the schema name (`Param_a-z___0-9____`). FastAPI will raise `TypeError: Unable to apply constraint 'pattern' to supplied value root='eee' for schema of type 'model'` on request

This doesn't look intuitively for me.. We should probably handle this and raise an error on startup. Or support specifying this parameter this way.

<details><summary>Code in the details</summary>

```py
from typing import Annotated

from pydantic import ConfigDict, Field, RootModel

app = FastAPI()


class Param(RootModel[str]):
    root: Annotated[
        str,
        Field(
            description="Description from RootModel root annotation",  # Displayed in Components, overriden by json_schema_extra
            pattern="^[a-z]+$",  # This works (displayed in parameters and validated)
        ),
    ]

    model_config = ConfigDict(
        json_schema_extra={
            "description": "Description from model_config"  # Displayed in Components
        }
    )


@app.get("/")
async def read_root(
    name: Annotated[
        Param,
        Query(
            description="Description from function parameter annotation",  # Works (displayed in parameters)
            # pattern="^[a-z]+[0-9]$",  # Raises "Unable to apply constraint 'pattern' to supplied value" on request
        ),
    ],
):
    return {"Hello": name}
```

</details>

YuriiMotov (commented):
I tested it a bit more and there are probably some issues with how `FieldInfo` parameters are handled\combined. Anyway, it requires more efforts to test it carefully.

Before we proceed, I suggest we wait for Sebastian to review current state and validate the idea.

As for use cases for RootModels as non-body parameters, this one is the best I could find so far: https://github.com/fastapi/fastapi/issues/11134#issuecomment-1986977600

P.S.: I still haven't tested the performance impact of these changes

YuriiMotov:
To move this forward we need to come up with strong arguments on why we need to support this.
Some real-life examples where `RootModel` can solve the problem that is hard to be solved without it.

fastapi/dependencies/utils.py
MarkusSintonen: I reverted that back to using the original check

fastapi/_compat.py
MarkusSintonen: Added a test case [here](https://github.com/fastapi/fastapi/pull/11306/commits/6b5fefadab788cc13808f84a255898485d571de3)

MartinBernstorff:
> To move this forward we need to come up with strong arguments on why we need to support this. Some real-life examples where `RootModel` can solve the problem that is hard to be solved without it.

Isn't this basically the same as the argument for a NewType? Avoiding primitive obsession? https://refactoring.guru/smells/primitive-obsession

Would love to see this merged, and thanks so much for working on this!

tiangolo:
Thanks for the interest here!

Let me copy my response in https://github.com/fastapi/fastapi/issues/11134

> I have never been a fan of Pydantic's `RootModel`, as it adds impedance on what the type actually is. There was no way to achieve it in Pydantic v1, so I understand its existence. And Pydantic v2 need to have some type of compatibility with v1, so I understand why it's available in Pydantic v2.

> I think now that Pydantic v2 supports `AfterValidator` and others, things can be achieved in ways that don't interfere with how types work, without using `RootModel`.

> For now I'll pass on this request, but if some people can come up with compelling use cases of why this is needed and can't be achieved with the other regular Pydantic tools (like `AfterValidator`), please create a discussion question with the specific use case and I'll review it again.

For now, I'll close this one. Thanks! ☕

PR #11313, opened 2024-03-19
Empty string("") as a path parameter default value.

osangu:
Hello! 

This PR is about set empty string(`""`) as a default value of path.

I think `@app.get()` is more elegance than `@app.get("")`,
beacause of below reasones. 

- if `()`and `("")` do same works, may be "" is not important information. 

1. we don't have to do same job of typing "".
2. When using with other parameters, you can focus on other parameters.

**before**
```python
from http import HTTPStatus

app = FastAPI()
router = APIRouter(prefix="/users")

@router.get("")
def foo():
    pass

@router.post(
    "", # or path=""
    response_model=Bar,
    status_code=HTTPStatus.CREATED
)
def bar():
    return Bar()

app.include_router(router)
```

**after**
```python
from http import HTTPStatus

app = FastAPI()
router = APIRouter(prefix="/users")

@router.get()
def foo():
    pass

@router.post(
    response_model=Bar,
    status_code=HTTPStatus.OK
)
def bar():
    return Bar()

app.include_router(router)
```

And i think about problems after this change and how to prevent.

1. backward compatibility: It is just a **default**, so people who using with "" or "/" will not trap in any problem.
2. Not expected path by default:  show warning log or option about want to use default value.

So i made this changes : )

----
P. S
this is just my opinion and i also want to listen about other people thinking!

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -1468,7 +1468,7 @@ class FastAPI(Starlette):
                 For example, in `http://example.com/items`, the path is `/items`.
                 """
             ),
-        ],
+        ] = "",
         *,
         response_model: Annotated[
             Any,
@@ -1841,7 +1841,7 @@ class FastAPI(Starlette):
                 For example, in `http://example.com/items`, the path is `/items`.
                 """
             ),
-        ],
+        ] = "",
         *,
         response_model: Annotated[
             Any,
@@ -2219,7 +2219,7 @@ class FastAPI(Starlette):
                 For example, in `http://example.com/items`, the path is `/items`.
                 """
             ),
-        ],
+        ] = "",
         *,
         response_model: Annotated[
             Any,
@@ -2597,7 +2597,7 @@ class FastAPI(Starlette):
                 For example, in `http://example.com/items`, the path is `/items`.
                 """
             ),
-        ],
+        ] = "",
         *,
         response_model: Annotated[
             Any,
@@ -2970,7 +2970,7 @@ class FastAPI(Starlette):
                 For example, in `http://example.com/items`, the path is `/items`.
                 """
             ),
-        ],
+        ] = "",
         *,
         response_model: Annotated[
             Any,
@@ -3343,7 +3343,7 @@ class FastAPI(Starlette):
                 For example, in `http://example.com/items`, the path is `/items`.
                 """
             ),
-        ],
+        ] = "",
         *,
         response_model: Annotated[
             Any,
@@ -3716,7 +3716,7 @@ class FastAPI(Starlette):
                 For example, in `http://example.com/items`, the path is `/items`.
                 """
             ),
-        ],
+        ] = "",
         *,
         response_model: Annotated[
             Any,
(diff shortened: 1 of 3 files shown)


codespearhead:
I'd argue we should be careful with this PR as is because it'd be a slippery slope to [this](https://github.com/spring-projects/spring-framework/issues/28552#issuecomment-1469740759) and [this](https://www.mastertheboss.com/soa-cloud/quarkus/addressing-cve-2023-4853-in-quarkus/).

The default value should be a trailing slash in my opinion.

Also, there should be a mechanism to prevent users from creating endpoints with file extensions to prevent [this](https://ahrefs.com/blog/trailing-slash/#:~:text=Files%20shouldn%E2%80%99t%20end%20in%20a%20slash).

osangu:
First, thkas for your comment. 
I'm really happy to talking about this PR.

I read about your links, and summarize like below.
> please check about summaries tight, cause i'm not good at english.  :cry:
> I hope they will correct with link context.
- This change can cause many cost and problems.
- CVE-2023-4853 Security Problem: access to secure endpoint by using multi slash.
- Files shouldn’t end in a slash.

But, i guess those problems caused when using trailing forward slash as default.
So i want to know why you recommend slash as default.
By using empty string, can't we prevent problems?

----
P. S
This PR is for a *default path*, that make developers can skip writing none path.
I also known and thought about side effects.

And this is my conclusion.

The latest version and previous version, `path` was necessary parameter.
So `path` must exist in there code, and this change will not bring effects to them.

It only works on new colde.

...

I hope my thking is share to you kindly.
Please, ask me if you have any questions!

codespearhead:
The use of a trailing slash in URIs, or lack thereof, is [largely inconsistent](https://github.com/slorber/trailing-slash-guide?tab=readme-ov-file#summary) across service providers, because some will add it when it's missing, whereas others will just ignore it when it's there. This results in a brittle deployment since that behavior is out of our control.

Besides, this lets the consumers of your API choose either flavor and run the risk of being affected without warrant, as illustrated in that [Spring Boot thread](https://github.com/spring-projects/spring-framework/issues/28552#issuecomment-1469740759). Imagine the disaster if that change was in the [gateway of your cloud services provider](https://www.reddit.com/r/aws/comments/bg46g5/cloudfront_s3_returning_404_when_no_trailing_slash/) instead of in the framework you're using?

Adding a trailing slash will spare us the hassle, [as concluded here](https://github.com/facebook/docusaurus/issues/2654#issuecomment-639427489), because [adding it in URIs if it wasn't there used to be a common practice](https://wordpress.org/support/topic/trailing-slash-added-to-user-defined-url-404-on-file-links/), and changing the behavior of legacy codebases is something that's generally out of our control.

tiangolo:
So, the usage of an empty string for a route/path operation is already a strange corner case. I prefer to keep it explicit that it is taking an empty path than remove it altogether. I'm also gonna refactor how routers work, and that change could affect this, so for now I'll rather pass on this.

But thanks for the interest! :coffee:

osangu:
After reading an article about open source, I think I got too excited and my hasty PR could have potentially harmed the wonderful culture of open source. I apologize for this. Next time, I will consider the side effects more carefully before opening an issue or PR. Thank you and i'm really sorry 🥲

PR #11323, opened 2024-03-20
⚡️ Skip doing any work when sub dependant is already cached

xkabylgSICKAG:
Hi, we noticed very large performance gains after making this change for our app that heavily uses the dependency injection system. One of our slower endpoints went from taking ~1100ms to ~400ms.


YuriiMotov (approved):
It looks really reasonable to check whether the sub_dependency is already solved and cached before solving it!

Related PR, that uses almost the same approach: https://github.com/tiangolo/fastapi/pull/4089 (closed)

xkabylgSICKAG:
@YuriiMotov 
> Related PR, that uses almost the same approach: #4089

That PR also has a test! :)

I think that other PR, with your suggested changes, is a better change than this PR. Do you think it makes more sense to close this in favour of the other one?

fastapi/dependencies/utils.py
```diff
@@ -549,6 +549,16 @@ async def solve_dependencies(
...
         )
+
+        # If the dependant is already cached, skip doing any work here
+        if (
+            sub_dependant.use_cache
+            and sub_dependant.cache_key in dependency_cache
+            and sub_dependant.name is not None
+        ):
+            values[sub_dependant.name] = dependency_cache[sub_dependant.cache_key]
+            continue
```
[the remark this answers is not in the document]
xkabylgSICKAG: That makes sense 👍 

In which cases will a sub_dependant not have a name?

xkabylgSICKAG:
I've rebased and updated the PR with the suggested changes from @YuriiMotov and with the change to dependency_cache empty dict checking from #4089.

svlandeg:
@xkabylgSICKAG and @YuriiMotov: thanks for your work on this! Would it make sense to have a similar test as in #4089, effectively merging both PRs? Then we can go ahead and close the old PR and try to finish up this one.

xkabylgSICKAG:
@svlandeg Sure! I can spend some time on that later tonight. I would be very happy to get this merged so we can stop running a forked version of fastapi 🙂

xkabylgSICKAG:
@svlandeg I updated the code to add that test from #4089. I would have liked to create a similar test that didn't use include pydantic, but I wasn't able to figure out how to do that and still test the same thing.

fastapi/dependencies/utils.py
```diff
@@ -587,13 +587,25 @@ async def solve_dependencies(
...
+
+        if sub_dependant.use_cache:
+            # Use a unique object to compare against in case the cached value is None
+            cache_miss = object()
+            cached_value = dependency_cache.get(sub_dependant.cache_key, cache_miss)
+            # If the sub dependant is already cached, skip doing any more work
+            if cached_value is not cache_miss:
+                if sub_dependant.name is not None:
+                    values[sub_dependant.name] = cached_value
+                continue
```
svlandeg, line 607: If I understand this right, nothing will happen in the end if `sub_dependant.name is None`, right? So maybe when that is the case we shouldn't even check the cache? I'm thinking we could change the `if` on L599 to include the `if` from L605, if you see what I mean:

```suggestion
        if sub_dependant.use_cache and sub_dependant.name is not None:
            # Use a unique object to compare against in case the cached value is None
            cache_miss = object()
            cached_value = dependency_cache.get(sub_dependant.cache_key, cache_miss)
            # If the sub dependant is already cached, skip doing any more work
            if cached_value is not cache_miss:
                values[sub_dependant.name] = cached_value
                continue
```

xkabylgSICKAG: Hm, not exactly _nothing_, since it will still hit that `continue` and skip doing any work here the value was cached. I'm not sure if this is a real thing that can happen though. Do you know of any cases where there would not be a `sub_dependant.name`?

xkabylgSICKAG: So from reading a bit of the code it seems like dependencies that are added to an api route directly won't have a name. But I think we should still skip getting them again if they are in the cache.

https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-in-path-operation-decorators/#add-dependencies-to-the-path-operation-decorator

YuriiMotov: @xkabylgSICKAG right, we should leave this block of code as it is. We don't store the value. but we still need to skip solving this dependency.

YuriiMotov:
@xkabylgSICKAG , great job!
Could you please add the problem description to initial message (you can copy it from #4089) and hide the outdated comments (that have been already resolved) to make it easier to review by Sebastián?

fastapi/dependencies/utils.py
YuriiMotov, line 607: Are there any reasons not to simplify it this way?

```suggestion
            if sub_dependant.cache_key in dependency_cache:
                cached_value = dependency_cache[sub_dependant.cache_key]
                if sub_dependant.name is not None:
                    values[sub_dependant.name] = cached_value
                continue
```

dolfinus: `dict.get(...)` is atomic operation, and thus thread safe, while consequent `key in dict; dict[key]` is not

YuriiMotov: Good point!
But currently dependency solving process is not parallel.
Also, it's highly unlikely (useless?) that somebody decide to delete existing key from `dependency_cache`.

I would vote for clarity in this case if nobody brings other arguments

xkabylgSICKAG: We have been running this fork in our production application for quite a while now, and before we changed to an atomic `.get()` we got hard to reproduce KeyErrors.

YuriiMotov (approved):
LGTM

I still believe we can safely simplify [this part](https://github.com/fastapi/fastapi/pull/11323/files#r2088209107), but it's not critical

xkabylgSICKAG:
> I still believe we can safely simplify [this part](https://github.com/fastapi/fastapi/pull/11323/files#r2088209107), but it's not critical

I would have thought it could be simplified too, but we _did_ get `KeyError`s before introducing that change to an atomic operation. It only happened a couple of times a week in total, but that's still too many times. 🤷‍♀️ 

It might be that we are doing something unusually weird with the dependency injection system in our app, but I couldn't find a clean way to fix the issue without this change in FastAPI.

Any idea if this has a chance to get merged soon? I would love to not have an internal fork of FastAPI anymore 😄

xkabylgSICKAG:
Rebased to resolve the conflicts

tests/test_dependency_cache.py
```diff
@@ -1,9 +1,27 @@
...
+counter_holder = {"counter": 0, "parsing_counter": 0}
+
+if PYDANTIC_V2:
+    from pydantic import model_validator
+
+    decorator = model_validator(mode="before")
+else:
+    from pydantic import root_validator
+
+    decorator = root_validator
```
YuriiMotov:
```suggestion
```

ranjanprasad96:
Looking forward for this fix since this improves the performance of our apps in production

ranjanprasad96:
@tiangolo Any plan on when this patch will be in? I feel this patch is really making a difference and would love to see as part of the official release.
Thank you so much.

fastapi/dependencies/utils.py
```diff
@@ -658,6 +658,21 @@ async def solve_dependencies(
...
     for sub_dependant in dependant.dependencies:
         sub_dependant.call = cast(Callable[..., Any], sub_dependant.call)
+        sub_dependant_cache_key = _get_cache_key(
+            dependant=sub_dependant,
+            uses_scopes_cache=_uses_scopes_cache,
+        )
+
+        if sub_dependant.use_cache:
+            # Use a unique object to compare against in case the cached value is None
+            cache_miss = object()
```
dolfinus: sentinel object can be created outside for-loop, or even outside the function itself

fastapi/dependencies/utils.py
dolfinus, line 690:
```suggestion
        if sub_dependant.use_cache:
```

YuriiMotov: Maybe makes sense (not sure actually, we need to think if it can break some specific use cases), but I think this is out of scope for this PR

dolfinus: This PR moves the check for `key in dependency_cache` into block above with `continue`, so check in this particular line will always be true.
Also there is no need to update dependency cache if dependency does not require caching in the first place.

xkabylgSICKAG: If there is another call of the same dependency that would use the cache, then it wouldn't find the cached value from this call? (I'm not sure if that is what I would _expect_ to happen, but it is the current behavior.)

Imagine the following:
```py
async def needy_dependency(
    fresh_value: Annotated[str, Depends(get_value, use_cache=False)],
    cached_value: Annotated[str, Depends(get_value, use_cache=True)],
):
```

Previously `fresh_value` would be resolved and then added to the cache, and `cached_value` would be grabbed from the cache. With the suggested change it would not add the result from `fresh_value` to the cache and have to resolve `cached_value`.

fastapi/dependencies/utils.py
xkabylgSICKAG: Moved it outside the function.

xkabylgSICKAG:
Rebased on latest upstream and moved the cache_miss sentinel value outside the function as suggested by @dolfinus

PR #11335, opened 2024-03-22
♻️ Simplify string format with f-strings in `fastapi/applications.py`

igeni:
changed string sum to f-strings to avoid case of type's mismatch and for improving readability


codespearhead (approved):

alejsdev:
Nice, thanks for the contribution! @igeni @codespearhead :partying_face: :rocket:

PR #11342, opened 2024-03-25
Example: native api client (without generating code) - DO NOT MERGE

LamaAni:
The following PR is an example for the issue: https://github.com/tiangolo/fastapi/discussions/8242 and  https://github.com/tiangolo/fastapi/discussions/11343

In short, an FastApiClient object was added that allows the auto capture of calls to the decorated method
in which the call is either directed to the function itself or is redirected to the an api client, depending on the context.

E.g. we can create an api client in the following manner:
```python
from fastapi import FastAPI, FastAPIClient


class MyApiClient(FastAPIClient):
    # Client paramters to be added here.
    # But this dose nothing, We could implement a default api client.
    def __init__(self, host: str = "localhost:8080") -> None:
        super().__init__()

    def send(self, route: APIRoute, args: list, kwargs: dict):
        # Send the request given the route.
        # Dummy just returns the route and input values
        return route, args, kwargs

    async def send_async(self, route: APIRoute, args: list, kwargs: dict):
        # Send the aysnc request given the route.
        # Dummy just returns the route and input values
        return route, args, kwargs


api = FastAPI()


@api.get("/my_func")
def my_func(a, b):
    return a + b


client = MyApiClient()
# a call will be made to my_func.
print(my_func(2, 3))
with client:
    # A call will be made to the server, returning
    # the proper response.
    print(my_func(2, 3))
# a call will be made to my_func.
print(my_func(2, 3))
```

### DIFF
--- a/fastapi/__init__.py
+++ b/fastapi/__init__.py
@@ -20,7 +20,7 @@ from .param_functions import Query as Query
 from .param_functions import Security as Security
 from .requests import Request as Request
 from .responses import Response as Response
-from .routing import APIRouter as APIRouter
 from .routing import APIClient as FastAPIClient
+from .routing import APIRouter as APIRouter
 from .websockets import WebSocket as WebSocket
 from .websockets import WebSocketDisconnect as WebSocketDisconnect
--- a/fastapi/__init__.py
+++ b/fastapi/__init__.py
@@ -21,5 +21,6 @@ from .param_functions import Security as Security
 from .requests import Request as Request
 from .responses import Response as Response
 from .routing import APIRouter as APIRouter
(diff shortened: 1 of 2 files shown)


tiangolo:

This would belong in an independent package, not in FastAPI itself. It would probably use the OpenAPI generated by FastAPI, it would probably use JSON Schema to create all the models, etc.

As this would be out of scope for FastAPI, I'll go ahead and close this one, but thanks for the interest! :coffee:

LamaAni:
Hi, I have created a client independent package [here](https://github.com/LamaAni/fastapi_client)
It dose not use the python schema :)

PR #11368, opened 2024-03-29
✏️ Fix typo in `fastapi/security/oauth2.py`

shandongbinzhou:


alejsdev:
Good catch, thanks! @shandongbinzhou :nerd_face: :mag:

PR #11377, opened 2024-03-31
👷 Add cron to run test once a week on monday

estebanx64:


codespearhead:
What's the point of having that workflow run once in a week when it already runs on every push?

tiangolo:
Great, thanks @estebanx64! :rocket:

---

@codespearhead I asked for that. To ensure things are working, even if there are changes in the dependencies, etc.

PR #11382, opened 2024-04-01
Update docs_src/body_fields with more Annotated in one place where it should be

Forelyl:
Tutorial, File: 001_an_py310.py

We can also use Annotated for fields in BaseModel 

I am not going to add this to other python versions, because I haven't checked if everything is working nicely for older versions

### DIFF
--- a/docs_src/body_fields/tutorial001_an_py310.py
+++ b/docs_src/body_fields/tutorial001_an_py310.py
@@ -8,8 +8,12 @@ app = FastAPI()
 
 class Item(BaseModel):
     name: str
-    description: Annotated[str | None, Field(title="The description of the item", max_length=300)] = None
-    price: Annotated[float, Field(gt=0, description="The price must be greater than zero")]
+    description: Annotated[
+        str | None, Field(title="The description of the item", max_length=300)
+    ] = None
+    price: Annotated[
+        float, Field(gt=0, description="The price must be greater than zero")
+    ]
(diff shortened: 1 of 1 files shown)


tiangolo:
📝 Docs preview for commit 8bdcf2ea5acdf0c27eebad2a75e4edb57031b5ea at: https://af00241c.fastapitiangolo.pages.dev

Forelyl:
There also should be a change body-fields.md
(it is seen in raw mode, the hl_lines is been changed)

Before:
```Python hl_lines="11-14"
{!> ../../../docs_src/body_fields/tutorial001_an_py310.py!}
```

After
```Python hl_lines="11-16"
{!> ../../../docs_src/body_fields/tutorial001_an_py310.py!}
```

(sorry don't see a way to add this change to current pull request)

Forelyl:
This Annotation could be also added to the lessons:

-  [Declare Request Example Data](https://fastapi.tiangolo.com/tutorial/schema-extra-example/)

tiangolo:

Yeah, I know this would be viable, but it's less obvious when it's done in Pydantic models as they have the custom dataclass-like behavior. They are very similar. And dataclasses don't support `Annotated` (yet). Probably for that reason, Pydantic uses the old syntax in most of their docs. So, for the Pydantic models, I'll leave them as is for now. But thanks for the interest! :coffee:

PR #11398, opened 2024-04-02
Add documentation and fix lint

koogle:


codespearhead (approved):

koogle:
@Kludex  before I go to deep on debugging it appears the middleware is not actually executed from the test failures. I have looked at the interactions between `fastapi.routing` and the underlying starlette classes but I am not quite sure where the handover is missing

tiangolo:
Thanks for the interest @koogle! :nerd_face: 

Nevertheless, I want to improve the API to interact with and add middlewares, that's why I haven't added any of this to the param docs.

Given that, I'll pass on this one, but thanks for the interest! :coffee:

PR #11425, opened 2024-04-11
Update routing.py to handle circular references

JavierSanchezCastro:
From #11424


fastapi/routing.py
```diff
@@ -1204,6 +1204,7 @@ def read_users():
         app.include_router(internal_router)
         ```
         """
+        assert id(self) != id(router), "Router cannot be the same as parent"
```
killjoy1221, line 1207: would `self is not router` not be sufficient?

JavierSanchezCastro: That's completely true, updated!

mattmess1221 (approved):

eljoserass (approved):
👍

connesy:
Hi @JavierSanchezCastro 

Sorry for pinging you on an old PR, but was this PR ever merged into FastAPI master? I ran into this exact issue today, and couldn't find your `assert self is not router` in the source code.

JavierSanchezCastro:
@connesy Hey, I think it got closed automatically because I made changes in my repo, I'll create another one!

PR #11436, opened 2024-04-14
♻️ Refactor getting first item from list in `fastapi/utils.py`

jd-solanki:
Changes according to ruff linting [rule](https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/)


Jaza:
Looks good to me. Per the ruff rule description, doing it this way is less expensive, and basically has the same behaviour. Tests / docs shouldn't need updating.

YuriiMotov:
I'm not sure this is important here.
It's said in the description of the rule that this construction can be expensive for large collections. This is not the case here.

> Calling e.g. list(...) will create a new list of the entire collection, which can be very expensive for large collections. 

At the same time `next(iter())` is less readable (IMO).
I would suggest to keep it as it is.

Also, if we decide to follow this rule, shouldn't we enable it in config then?

jd-solanki:
@alejsdev I didn't get any notification regarding you were waiting for something?

Did we decide to discard it?

YuriiMotov:
@jd-solanki,

Considering this PR was created more than an year ago, we assume that if author doesn't reply to comments for a month, they are probably not interested in it anymore.

You didn't reply to https://github.com/fastapi/fastapi/pull/11436#issuecomment-2957778966 and it was automatically closed.
If you think we should reopen this, please address the comment above.

Thanks for contributing to FastAPI!

jd-solanki:
Hi @YuriiMotov 

Thanks for your response. I believe this is minor performance improvement just like how everyone knows about FastAPI.

> _next(iter()) is less readable (IMO)_

I totally agree with you but for performance this is the only way and you might have used it someone in your life.


> _if we decide to follow this rule, shouldn't we enable it in config then_

Yes, It's good to add it to config. If you allow and willing to merge, I can update the PR.


> _P.S. It's quite surprising to me that PR with minor performance boost got rejected and we don't care about making FastAPI a bit faster_ 🤔

YuriiMotov:
> _P.S. It's quite surprising to me that PR with minor performance boost got rejected and we don't care about making FastAPI a bit faster_ 🤔

We always try to find a balance between speed, code clarity and the time we spend on reviewing PR.
This part of code is not frequently executed (it's only invoked once per route during the app startup). I think nobody will notice the performance increase.

Also, if we are talking about performance increase, we should prove it. I recently saw a PR that aimed to improve performance, and looking at changes it seemed that it should give that improvement. I tested it and in fact we got a decrease in performance.

If you still think that we should bring it to Sebastián's attention, I will re-open this PR.

Hope for your understanding.

jd-solanki:
If it's for initial startup only, then we can close this. I'm fine with this.

Thanks for your time.

PR #11440, opened 2024-04-15
Allow streamable response as response_class to work

92dev:
Basically classes such as `StreamingResponse` in `starlette` cannot work unless specified directly in the response-type due to this serialization concept (which they always fallback to use JSON)

There is the alternative which I opted out of for now, which is pass a `response_serialize: bool` parameter throughout the entire flow, it might be better but I wanted to avoid API changes (even through defaults) until any maintainer gives his opinion on the change

example usage:
```

# Here `response_class` is only used for documentation
@app.get('/test_before', response_class=StreamingResponse)
async def before_fix(msg: str):
    async def txt_generator():
        for i in range(10):
            yield f'event: something\ndata: {i}\n\n'
    return StreamingResponse(txt_generator())

# Here `response_class` is actually working
@app.get('/test_after', response_class=StreamingResponse)
async def after_fix(msg: str):
    async def txt_generator():
        for i in range(10):
            yield f'event: something\ndata: {i}\n\n'
    return txt_generator()
```

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -294,6 +294,7 @@ def get_request_handler(
                         if sub_response.status_code:
                             response_args["status_code"] = sub_response.status_code
                         # Some response classes handle serialization themselves via iteration
+                        content = raw_response
                         if not hasattr(actual_response_class, 'body_iterator'):
                             content = await serialize_response(
                                 field=response_field,
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -293,17 +293,19 @@ def get_request_handler(
                             response_args["status_code"] = current_status_code
                         if sub_response.status_code:
                             response_args["status_code"] = sub_response.status_code
-                        content = await serialize_response(
-                            field=response_field,
-                            response_content=raw_response,
-                            include=response_model_include,
-                            exclude=response_model_exclude,
-                            by_alias=response_model_by_alias,
-                            exclude_unset=response_model_exclude_unset,
-                            exclude_defaults=response_model_exclude_defaults,
-                            exclude_none=response_model_exclude_none,
-                            is_coroutine=is_coroutine,
-                        )
+                        # Some response classes handle serialization themselves via iteration
+                        if not hasattr(actual_response_class, 'body_iterator'):
+                            content = await serialize_response(
+                                field=response_field,
+                                response_content=raw_response,
+                                include=response_model_include,
+                                exclude=response_model_exclude,
+                                by_alias=response_model_by_alias,
+                                exclude_unset=response_model_exclude_unset,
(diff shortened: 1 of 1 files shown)


nidhishgajjar:

## Review of PR #11440: Allow streamable response as response_class to work

The PR addresses a real pain point — when using `response_class=StreamingResponse`, FastAPI's serialization still tries to JSON-serialize the raw content, which breaks streaming responses. The fix is on the right track but has some concerns worth addressing.

### Concerns 🟡

1. **Reliance on Starlette internal attribute**: The check `hasattr(actual_response_class, 'body_iterator')` depends on `StreamingResponse` having a `body_iterator` class-level attribute. This is an implementation detail of Starlette that could change between versions. A more robust approach would be to check the actual content type:

   ```python
   import types
   
   if isinstance(raw_response, (types.AsyncGeneratorType, types.GeneratorType)):
       content = raw_response
   else:
       content = await serialize_response(...)
   ```

   This checks whether the returned content is inherently un-serializable (a generator) rather than checking the response class.

2. **No tests included**: For a behavioral change in the request/response pipeline, test coverage is essential. At minimum, tests should cover:
   - Sync generator with `response_class=StreamingResponse`
   - Async generator with `response_class=StreamingResponse`
   - Regular JSON response still works with `response_class=JSONResponse` (regression test)

3. **Edge case — iterable but not generator**: Some responses might return a list or other iterable that should still be serialized, not streamed. The current approach (checking the response class) handles this correctly since only `StreamingResponse`-like classes have `body_iterator`. But if the content-based approach is used, it should also check for `collections.abc.AsyncIterable` / `collections.abc.Iterable` with appropriate exclusions for lists/dicts.

### What looks good ✅

- Correctly identifies the problem: serialization breaking streaming responses
- The comment explaining the rationale is helpful
- Minimal code change — only touches the necessary code path
- The approach of checking the response class rather than adding API parameters is reasonable for a first pass

### Summary

The fix correctly identifies and addresses a real issue with streaming responses. However, relying on `body_iterator` as a detection mechanism ties the implementation to Starlette internals, and the lack of tests makes it hard to verify correctness and prevent regressions. Consider checking the content type instead of the response class attribute.

**Assessment: request-changes** (add tests, consider more robust detection mechanism)

tiangolo:

This was something I wanted to solve for a while, but I wanted to have a specific interface for it, being able to yield directly from the function.

It is implemented now: https://fastapi.tiangolo.com/tutorial/stream-json-lines/

I felt weird about closing a PR before having an alternative, but then I ended up leaving this one open for very long, sorry for that.

92dev:
@tiangolo no worries, IMO best to not close before we have better options, thanks for the update !

PR #11449, opened 2024-04-17
there is issue while creating the binary file when workers are added in the the main function   so I solved these error by making change in file name. 

100rab100:
with the help of pyinstaller you can make binary  file of the  fastapi application  i suggest here that you can make binary file with addition of the workers which are configurable and  you can change it after the making binary file also.
please check my  code


tiangolo:
Hey @100rab100 :wave: 

This content would be better in a blog post showing your use case and solution, so I'll pass on this one. But thanks for the interest! :coffee:

PR #11452, opened 2024-04-17
Adjust a test for error URL added in Pydantic 2.7.0

musicinmybrain:
One test, `tests/test_tutorial/test_path_params/test_tutorial005.py::test_get_enums_invalid`, has started failing with Pydantic 2.7.0

The failure is easily reproduced by checking out the current master, 3425c834cc113060d00b8a295d3456eb306a109d, and running the tests as described in https://fastapi.tiangolo.com/contributing/#tests.

This PR adapts that test for the new error URL included in the response to the validation failure.

### DIFF
--- a/tests/test_tutorial/test_path_params/test_tutorial005.py
+++ b/tests/test_tutorial/test_path_params/test_tutorial005.py
@@ -38,7 +38,7 @@ def test_get_enums_invalid():
                     "input": "foo",
                     "ctx": {"expected": "'alexnet', 'resnet' or 'lenet'"},
                     # e.g. https://errors.pydantic.dev/2.7/v/enum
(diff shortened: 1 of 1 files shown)


musicinmybrain:
It seems like the lint failure is not related to this PR:

```
fastapi/openapi/utils.py:126: error: Incompatible types in assignment (expression has type "deprecated | str | Literal[True]", target has type "dict[str, Any] | bool | str")  [assignment]
```

tiangolo:
Thank you! I just updated those tests and also how these things work. Including the removal of the URL from errors. So this is now handled. But thanks for the work! :coffee:

musicinmybrain:
> Thank you! I just updated those tests and also how these things work. Including the removal of the URL from errors. So this is now handled. But thanks for the work! ☕

Wonderful; that is much better. Thank you for the update!

PR #11493, opened 2024-04-27
Add easy "debug-on-exception" debugger middleware functionality (and remote debugging)

izo0x90:
### Intent:

- Quicker to replicate issues, less cycles to fix
  - Often webdev workflow involve dealing with mistakes that require some set of step to replicate, having the a debugger start up when the issue occurs initially can really mitigate the number of iterations on figuring out and fixing issues.

- Easy remote/ docker-ized debugging
  - Having easy access to a remote debugger such as webdev also makes debugging docker-ized (and/ or dev. envs. running on a remote machine) straightforward.

- Additionally having a debugger right in the browser can make the "ergonomics" of this cycle less cumbersome.

This PR attempts to achieve the above by providing an easy way to add debug-on-exception feature to Fastapi along side with some sensible default debugger options.
 
### Overview:
- Adds middleware class to allow intercepting exceptions in endpoints and starting a debugger in those stack frames

- Debugger is customizable by passing in a custom callable that can run debugger of choice

- Default debugger if not customized is PDB

- Adds support to use web-PDB by including pre made callable that can be passed in as a middleware arg

### DIFF
--- a/fastapi/middleware/debugger.py
+++ b/fastapi/middleware/debugger.py
@@ -1,0 +1,104 @@
+import typing
+
+from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
+from starlette.requests import Request
+from starlette.responses import Response
+from starlette.types import ASGIApp
+
+
+async def catch_exceptions_middleware(
+    request: Request, call_next: RequestResponseEndpoint
+) -> Response:
+    import pdb
+    import sys
+
+    try:
+        return await call_next(request)
+    except Exception as e:
+        sys.last_traceback = e.__traceback__
+        pdb.pm()
+        raise
+
+
+async def webpdb_catch_exceptions_middleware(
+    request: Request, call_next: RequestResponseEndpoint
+) -> Response:
+    """
+    Requires web-pdb package.
+
+    ```bash
+    pip install web-pdb
+    ```
+    """
+    import web_pdb  # type: ignore
+
+    with web_pdb.catch_post_mortem():
+        return await call_next(request)
+
+
+class DebuggerMiddleware(BaseHTTPMiddleware):
+    """
+    Middleware to intercept exceptions and start a debugger in that stack frame.
+
+    The default configured debugger is PDB since it comes packaged with Python.
+
+    Additionally web UI can be added by add web-pdb a dep., or a custom callable
+    be passed through the `start_debugger_func` middleware parameter to run
+    a debugger of choice.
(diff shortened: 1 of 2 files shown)


Jaza:
I guess this is nice-to-have. But:

- I don't really see why this needs to live directly in FastAPI or any of its dependencies
- Even if you ignore my first point, this could live in Starlette rather than in FastAPI, Starlette already includes a number of middlewares, and AFAICT this doesn't depend on anything in FastAPI, it only depends on Starlette
- Personally I'm happy enough just with stack traces showing in the browser / console in my local environment, and when I want to step it up a notch, temporarily chucking a `pdb.set_trace()` manually at one spot in my code is enough for me (but maybe other folks feel the need for a richer debugging experience to be included out-of-the-box?)
- My preference would be for this to live in in its own little library / repo (or even just as a gist or similar)

Do other people have thoughts on the matter?

izo0x90:
> I guess this is nice-to-have. But:
> 
> * I don't really see why this needs to live directly in FastAPI or any of its dependencies
> * Even if you ignore my first point, this could live in Starlette rather than in FastAPI, Starlette already includes a number of middlewares, and AFAICT this doesn't depend on anything in FastAPI, it only depends on Starlette
> * Personally I'm happy enough just with stack traces showing in the browser / console in my local environment, and when I want to step it up a notch, temporarily chucking a `pdb.set_trace()` manually at one spot in my code is enough for me (but maybe other folks feel the need for a richer debugging experience to be included out-of-the-box?)
> * My preference would be for this to live in in its own little library / repo (or even just as a gist or similar)
> 
> Do other people have thoughts on the matter?

Thank @Jaza appreciate the feedback.Very fair points. I do still think having something like this out of the box can be very convenient, esp. as ppl spin up new projects. If the consensus ends up being that it doesn't make sense to house in Fastapi I will move it over to Starlette and see how that goes. 

But curious to see if there is any other input/ opinions here?

tiangolo:
Thanks for the interest @izo0x90! :nerd_face: 

I prefer to encourage and use the integrated debugger in an editor. Or if investigating errors, to see them on the terminal in the logs.

Nevertheless, as this is something that would solve a use case for you, there's a chance other people have the same type of use case, so maybe it would be worth it creating a new external package, that's an approach I would suggest if you would like to have this publicly available. :nerd_face: 

izo0x90:
@tiangolo Makes sense, thank you!

PR #11496, opened 2024-04-29
📝 Update references to Python version, FastAPI supports all the current versions, no need to make the version explicit

tiangolo:
📝 Update references to Python version, FastAPI supports all the current versions, no need to make the version explicit


ranvier2d2:
Im sorry, this may be a stupid question but does that mean it supports 3.12 now? Just want to make sure before i mess things up. I knew it was compatible up to 3.10

PR #11503, opened 2024-04-30
🔧 Add configs and setup for `fastapi-slim` including optional extras `fastapi-slim[standard]`, and `fastapi` including by default the same `standard` extras

tiangolo:
🔧 Add configs and setup for `fastapi-slim` including optional extras `fastapi-slim[standard]`, and `fastapi` including by default the same `standard` extras

The `fastapi` package now includes the `standard` default recommended packages, and `fastapi-slim` comes without those recommended standard packages and with a group of optional dependencies `fastapi-slim[standard]`, equivalent to `fastapi`, for those that want to opt out of those default `standard` packages.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,10 +4,6 @@ on:
   release:
     types:
       - created
-  # TODO: remove this
-  push:
-    branches:
-      - slim
 
 jobs:
   publish:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,10 @@ on:
   release:
     types:
       - created
+  # TODO: remove this
+  push:
+    branches:
+      - slim
 
 jobs:
   publish:
@@ -12,6 +16,7 @@ jobs:
       matrix:
         package:
           - fastapi
+          - fastapi-slim
     permissions:
       id-token: write
     steps:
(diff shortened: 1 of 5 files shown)


tiangolo:
📝 Docs preview for commit 79621a3acefddaadbc6a9933338f52e0cdde1a82 at: https://28bb498b.fastapitiangolo.pages.dev

noklam:
Hi! https://discuss.python.org/t/optional-dependency-groups-omitting-package-requirements/20833 brought me to here as I am researching how to split out some optional dependencies in a nice way.

I see that you have taken a custom approach, here is what I understand:
- `fastapi-slim` share the same codebase as `fastapi`
- `pip install fastapi` installs the extra optional dependencies while `pip install fastapi-slim` doesn't
- `pip install fastapi[standard]` is equivalent to `pip install fastapi[standard]` in terms of dependencies

I have a few questions:
- What lead you to decide to implement a custom solution?
- How does this actually work? I see that you are using `pdm`, but I am surprised when I clone the repository and do `pip install -e .` it also install the [standard] dependencies automatically, how does this work? I expect this is some "magic" that you add during build time only.

edouardlp:
This PR causes issues for library maintainers. See https://github.com/tiangolo/fastapi/discussions/11733

PR #11517, opened 2024-05-02
🐛 Fix custom response input output schema bug

slafs:
This PR addresses the issue (point 1.) from https://github.com/tiangolo/fastapi/discussions/10697

Add missing `mode` argument when calling `create_model_field`
for "custom" response models defined in the `responses` option of a route.
The same way we do that for the "main" response model (see the other call of `create_model_field`).

Otherwise that custom response model schema (if separate i/o models are enabled)
is generated with the "-Input" suffix (instead of `-Output`).

NOTE this PR doesn't address docs issues that are mentioned in https://github.com/tiangolo/fastapi/discussions/10697 (point 2.)

PR #11518, opened 2024-05-02
⚙️ Record and show test coverage contexts (what test covers which line)

slafs:
As documented in https://coverage.readthedocs.io/en/7.5.0/contexts.html#dynamic-contexts

While working on #11517 I wanted to see which test is best to modify for the fix.
One way to easily find this out is to see which line is covered by which test.

This PR adds a dynamic context setting
and makes them show on the HTML coverage report.
The contexts are test function names.

NOTE pytest-cov plugin has even better support
for pytest test names as coverage contexts,
but I think this PR is already an imporovement.


pyproject.toml
```diff
@@ -206,7 +206,7 @@ source = [
     "tests",
     "fastapi"
 ]
-context = '${CONTEXT}'
```
slafs, line 209: Ah, not sure what this piece was for TBH 😅. Decided to remove it though.

tiangolo:
📝 Docs preview for commit 84bc7c6f90dda05fa8c93cbf711623edecfc1764 at: https://1ba1b8ae.fastapitiangolo.pages.dev

pyproject.toml
tiangolo: This env var is used by/with GitHub Actions so that coverage can know where to store the coverage files, that way they can be combined later. So I think doing this trick wouldn't work. :thinking:

tiangolo:
I was checking the docs, and it seems both contexts can be combined, so we can know which Python version and which tests ran each function. :tada: :rocket: 

https://coverage.readthedocs.io/en/7.6.1/contexts.html#dynamic-contexts

I updated that bit to get that. :nerd_face:

PR #11574, opened 2024-05-13
🔨 Make it easy to change docs asset urls

waketzheng:
Support change docs CDN urls by this way:
```py
from fastapi import FastAPI
from fastapi.openapi.docs import AssetUrls

AssetUrls.swagger_js = 'https://unpkg.com/swagger-ui-dist@5.9.0/swagger-ui-bundle.js'
AssetUrls.swagger_css = 'https://unpkg.com/swagger-ui-dist@5.9.0/swagger-ui.css'
AssetUrls.redoc_js = 'https://unpkg.com/redoc@next/bundles/redoc.standalone.js'

app = FastAPI()
```

### DIFF
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -23,6 +23,16 @@ swagger_ui_default_parameters: Annotated[
 }
 
 
+class AssetUrls:
+    swagger_favicon = "https://fastapi.tiangolo.com/img/favicon.png"
+    swagger_js = (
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for the interest. :coffee: 

But this would mean that overriding a global variable (class attribute) would affect any other FastAPI apps, so it wouldn't work well. I also think it's better to document how to override and customize everything if people need more than the defaults: 

https://fastapi.tiangolo.com/how-to/configure-swagger-ui/
https://fastapi.tiangolo.com/how-to/custom-docs-ui-assets/

Given that I'll pass on this, but thanks for the effort! :cake:

PR #11582, opened 2024-05-15
Create CODE_OF_CONDUCT.md

vlasvlasvlas:
just adding code of conduct i think this will help :)


tiangolo:

What I don't like about these things is that they are read and used mainly (only?) by the people that don't need them. And the ones that are not behaving well wouldn't go to read it.

Maybe it makes sense to have something to point out when there's a need to ban someone, I hope that never happens. But either way, I would prefer to research it and write/choose the code of conduct myself. So I'll pass on this one for now. But thanks for the interest! :coffee:

PR #11627, opened 2024-05-23
📝 Add monthly downloads badge to `README.md`

ayushjain01:
Added monthly downloads badge. View more at - https://piptrends.com/widgets/fastapi


tiangolo:
Thanks! I don't think it's necessary to have it on the docs, so I'll pass for now, but thank you for the interest! 🤓

PR #11642, opened 2024-05-28
README Update to Account for Keyboard Interrupt Error with Current Uvicorn Version

pollyandhiscat:
When using the current version of Uvicorn shipping with FastAPI, a specific error will occur when running the server in dev mode. When you make a change to the 'main.py' file, the console throws a long error string, which ends up being related to a keyboard interrupt issue. Below are some screenshots:

image

image

image

image

In short, the workaround is simple: roll back to a prior version of Uvicorn. This README update simply lets FastAPI users know there is an issue with Uvicorn and that this small step can be taken to avoid a lot of confusion.

Command to roll back:

image

Then after saving file changes, you get the nice clean output we all like to see:

image

I know this is technically an issue with an external library, but it is an external library referenced directly by the current version of the README and this change might save some people headaches.

Note: my images here are using PowerShell, but the change to the README is not picky.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,8 +133,6 @@ FastAPI stands on the shoulders of giants:
 
 ```console
 $ pip install fastapi
-
----> 100%
 ```
 
 </div>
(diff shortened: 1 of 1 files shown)


Kludex:
I'll make a release on uvicorn's side.

pollyandhiscat:
@Kludex Thank you! Most appreciated! 🙌

PR #11670, opened 2024-06-01
✨ Provide serialize context to response models

ilyatotl:


estebanx64:
Hi @ilyatotl 👋

Thank you for your time and interest in helping us with your contribution. 🙇

Before we can continue reviewing this PR. Can you add some tests for the context you added and fix the lint? 🤓
Thanks again in advance!🚀

estebanx64:
@ilyatotl  it seems someone else created a PR before you to add this feature #11634.

Since there is another PR older with the same contribution to add their feature, We are going to keep the oldest one and continue there. 🤓

Thank you again for your time and for taking care of this feature! 🚀

PR #11676, opened 2024-06-03
Add to the documentation the functionality of generating raw HTML from a template

tanrax:


tiangolo:
Thanks for the interest! I don't think it's necessary to add this here, the full documentation around Starlette templates is here: https://www.starlette.io/templates/

People might also want to check the official JInja docs. I'll pass on this one but thanks for the effort! :coffee:

PR #11684, opened 2024-06-05
✨ Add CookieToAuth2 middleware to convert cookies to Authorization header

LainezDev:
Allows developers to use cookies and authorization headers without changing authentication settings. OAuth2PasswordBearer is designed to look for the token in the authorization header of the request. The middleware that converts the token cookie into an Authorization header ensuring that when OAuth2PasswordBearer inspects the request, it finds the necessary header in the correct format.  It is a simple proposal to several discussions about the use of cookies for authorization of web pages with FastAPI.

I'm looking forward to your feedback

```py
from fastapi import FastAPI
from fastapi.middleware.cookietoauth2 import CookieToAuth2Middleware

app = FastAPI()

app.add_middleware(CookieToAuth2Middleware, cookie_name="cookie_name")
```


tiangolo:
Thanks for the interest! The way to do this would probably be with dependencies, and not autoconverting from one source to another.

If you have a use case that you need to solve and have questions, the best place to start would be with a GitHub discussion. :nerd_face: 

For now I'll pass on this one. :coffee:

PR #11695, opened 2024-06-07
Use python-jose instead of pyjwt

utopianf:
Since FastAPI suggests using python-jose we should use `from jose import jwt`


tiangolo:
📝 Docs preview for commit 7d2dc5a7d56132e65f78340263685b264adab042 at: https://aa058787.fastapitiangolo.pages.dev

utopianf:
I found #11589 that FastAPI is suggesting to use PyJWT? Then I think this change was not needed so far.

PR #11722, opened 2024-06-14
Update README to Allow Install and Run Commands to Be Copied As Is

pollyandhiscat:
The current version of the README contains extra text that make commands fail.

Example:

image

But if we use the command as is, we get an error:

image

image

My humble proposal is to align the README with other parts of the README, which adopts a similar format (without the dollar sign and output as part of the command text).

Here is an example currently in the README, at the bottom:

image

Additionally, this format is good as it aligns with other documentation formats, such as that used by GitHub. Here is an example from the GitHub docs on generating new SSH keys:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,8 +134,6 @@ FastAPI stands on the shoulders of giants:
 
 ```console
 $ pip install fastapi
-
----> 100%
 ```
 
 </div>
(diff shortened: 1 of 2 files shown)


tiangolo:
📝 Docs preview for commit ea2b0b1bde8bd2a2df3e9ed9e3e56c80d1718b20 at: https://5e1c09ff.fastapitiangolo.pages.dev

tiangolo:
Thanks for the interest! This is because the README is more or less a copy of the markdown in the docs: https://fastapi.tiangolo.com/

Those commands are shown in a mini-animated terminal there. :nerd_face: 

So, I'll pass on this one, but thanks for the effort. :coffee:

PR #11782, opened 2024-07-01
📝 Add documentation on how to use the Django ORM inside FastAPI

patrick91:
- **Add initial documentation for using the Django ORM**
- **Add async example**

PR #11807, opened 2024-07-08
✨ Add Django middleware

patrick91:
This PR adds a Django middleware that allows to run the Django Request lifecycle 😊

This is still draft as I want to make sure it works in various cases, I also want to do the following:

- [ ] Only run the middleware if we have requested `django_request` 
      @tiangolo is there an easy way to check the requested dependencies from a middleware?)
- [ ] Allow to run Django views (for example the admin)
      Not sure if we should try/check if there's a Django route or have a list of paths that we should delegate to django :)
- [ ] Add a dependency for getting the current user
- [ ] More tests
- [ ] Add docs

### DIFF
--- a/fastapi/django/__init__.py
+++ b/fastapi/django/__init__.py
@@ -1,0 +1,82 @@
+from contextvars import ContextVar
+from typing import Annotated
+
+from django.core.asgi import ASGIHandler
+from django.http import HttpRequest, HttpResponse, StreamingHttpResponse
+from fastapi import Depends, Request, Response
+from fastapi.responses import StreamingResponse
+from starlette.middleware.base import BaseHTTPMiddleware
+
+_django_request = ContextVar[HttpRequest | None]("fastapi_django_request", default=None)
+
+
+def get_django_request():
+    django_request = _django_request.get()
+
+    if not django_request:
+        raise ValueError(
+            "Django Request not found, did you forget to add the Django Middleware?"
+        )
+
+    return django_request
+
+
+DjangoRequestDep = Annotated[HttpRequest, Depends(get_django_request)]
+
+
+class DjangoMiddleware(BaseHTTPMiddleware, ASGIHandler):
+    """A FastAPI Middleware that runs the Django HTTP Request lifecycle.
+
+    This middleware is responsible for running the Django HTTP Request lifecycle
+    in the FastAPI application. It is useful when you want to use Django's
+    authentication system, or any other Django feature that requires the
+    Django Request object to be available."""
+
+    def __init__(self, *args, **kwargs):
+        ASGIHandler.__init__(self)
+
+        super().__init__(*args, **kwargs)
+
+    async def _get_response_async(self, request):
+        fastapi_response = await self._call_next(request)
+
+        assert isinstance(fastapi_response, StreamingResponse)
+
+        return StreamingHttpResponse(
+            streaming_content=fastapi_response.body_iterator,
+            headers=fastapi_response.headers,
+            status=fastapi_response.status_code,
+        )
+
+    async def __call__(self, scope, receive, send):
+        self._django_request, _ = self.create_request(scope, "")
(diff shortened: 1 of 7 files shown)


tests/django/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+    yield
+
+    connection.creation.destroy_test_db("default", verbosity=0)
+
+
+@pytest.fixture(autouse=True)
+def flush_db():
+    sql_list = sql_flush(no_style(), connection, allow_cascade=False)
+
+    connection.ops.execute_sql_flush(sql_list)
```
patrick91, line 44: Same as https://github.com/tiangolo/fastapi/pull/11782

tests/django/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+    ],
+    DATABASES={
+        "default": {
+            "ENGINE": "django.db.backends.sqlite3",
+            "NAME": ":memory:",
+        }
+    },
+)
+
+django.setup()
```
patrick91, line 28: Django needs settings to be configured, this is the simplest way to configure django (we also add the required apps to get auth and sessions working)

tests/django/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+
+@pytest.fixture
+def authenticated_session_id():
+    from django.contrib.auth.models import User
+
+    from tests.django.utils import create_authenticated_session
+
+    user = User.objects.create_user(username="test", password="test")
+
+    return create_authenticated_session(user)
```
patrick91, line 55: this creates an authenticated session, so we can test if the user is logged in or not.

The reason why the import is here is mostly because we need Django to be configured before importing any model 😊

tests/django/test_django_request_no_middleware.py
```diff
@@ -0,0 +1,24 @@
...
+
+    return {"username": user.username}
+
+
+client = TestClient(app)
+
+
+def test_returns_an_error():
+    with pytest.raises(ValueError, match="Django Request not found"):
+        client.get("/")
```
patrick91, line 24: I wanted to give a hint if the user forgot to add the middleware, not sure if this is the best way though 😊

fastapi/django/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+            async def streaming():
+                async for chunk in django_response.streaming_content:
+                    yield chunk
+
+            return StreamingResponse(
+                status_code=django_response.status_code,
+                content=streaming(),
+                headers=django_response.headers,
+            )
```
patrick91, line 80: Some of this code can be abstracted, I'll do so in future 😊

YuriiMotov: Hi!
Sorry for my criticism, but I think the chance that this PR will be merged is very low. Repository owner tends not to increase the complexity of the code base.
This middleware seems to have quite specific use case. I think it would be better to implement this middleware as a separate project.

Tishka17:
I do not think it is a good idea to make hacks in fastapi to reuse badly designed code from another framework.
For example, in tests there is `aget_user` call. In better world you do not need HTTP-related things to get user. You need HTTP-related things to retrieve tokens/session id, which is different.

rroblf01:
Would it be possible to use the Django ORM with these changes?

patrick91:
@rroblf01 sorry for the late, but you can already use the Django ORM with FastAPI, see this PR: https://github.com/fastapi/fastapi/pull/11782#event-14043647967

eliottreich:
There's a $30 bounty posted for FastAPI middleware improvements on [Code.TaskBounty](https://code.task-bounty.com/task/add-fastapi-feature-middleware).

The goal is to get competing implementations so maintainers can evaluate different approaches before merging. Might help move this forward — it's been sitting for a while.

Bounty is funded and live. Anyone can submit a solution.

PR #11811, opened 2024-07-09
📝 Update `docs/en/docs/tutorial/first-steps.md`

matt2000:
According to [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-post), the current HTTP specification says:

"The POST method requests that the [target resource](https://datatracker.ietf.org/doc/html/rfc9110#target.resource) process the representation enclosed in the request according to the resource's own specific semantics."

... and:

"The PUT method requests that the state of the [target resource](https://datatracker.ietf.org/doc/html/rfc9110#target.resource) be created or replaced with the state defined by the representation enclosed in the request message content."

So, the tutorial is patently incorrect about PUT (updating has never been it's function, as far as I remember), and, I think, it unnecessarily over-simplifies POST.

This patch increases the accuracy of the doc, while only trivially increasing the complexity. 

If this patch is controversial, I suggest simply removing PUT from this tutorial, since the new definition of POST makes PUT unnecessary for very many web applications. (Which have been using POST and ignoring PUT for many years before the new standard anyway...)

### DIFF
--- a/docs/en/docs/tutorial/first-steps.md
+++ b/docs/en/docs/tutorial/first-steps.md
@@ -227,9 +227,9 @@ When building APIs, you normally use these specific HTTP methods to perform a sp
 
 Normally you use:
 
-* `POST`: to create data.
+* `POST`: to create, append, or update data in a data record.
 * `GET`: to read data.
-* `PUT`: to update data.
+* `PUT`: to create or replace a data record.
 * `DELETE`: to delete data.
 
 So, in OpenAPI, each of the HTTP methods is called an "operation".


tiangolo:
📝 Docs preview for commit 10b84710cbeabe57b64caa87a0abb0aa1747f59a at: https://dce198f3.fastapitiangolo.pages.dev

ceb10n:
I completely agree that the current statement simplifies what POST or PUT are.

But we need to remember it is the **First Steps** doc. We need to think that someone that is reading this doc is potentially a beginner, and despite being a simplified description, an introduction document **needs** to be simple. What we normally see in tutorials for beginners is precisely this simplification to make it easier to understand the basic operations of a CRUD when we are building an API.

So following this thought, we cannot say that the document is wrong.

But what was added didn't complicate the concept much.

matt2000:
I would argue that it is worse to give a beginner false information, and the doc's current definition of PUT is patently false, with respect to the RFC. If the language in this patch isn't sufficiently simple for the expected reader, then please just remove the reference to PUT. A beginner who needs more than POST can then look up better guidance elsewhere.

tiangolo:
Yep, as @ceb10n says, this is introductory content, I don't intend to copy the RFC as is, just to give "good enough" guidelines for newcomers.

That's also why I say "Normally you use", instead of "As per the spec" or something similar.

Under my point of view, the docs are fine as they are, and are a "good enough" tradeoff for the objective I have for them, so I'll pass on this one, but thanks for the interest. :coffee:

PR #11818, opened 2024-07-10
📝 Fix inconsistent response code when item already exists in docs for testing

lokomilo:
In https://github.com/tiangolo/fastapi/pull/4482, the response status code was changed to return 409 when the item already exists in non-Annotated examples, but it is 400 in the other examples created after that PR.


tiangolo:
📝 Docs preview for commit f3c1a2d5926458db870d9c749f78ba62559e6d2b at: https://7d76d610.fastapitiangolo.pages.dev

GDemay (approved):

Kludex (approved):

tiangolo:
Good catch, thank you @lokomilo! :mag: :rocket:

PR #11837, opened 2024-07-13
📝Add docstring for SecurityRequirement and Dependant class.

ashvin-a:


svlandeg (commented):
Hi, thanks for the contribution!

The `__init__` methods in FastAPI don't usually have docstrings, and to be honest I'm not sure that this particular docstring adds much information, as it's more of a generic comment (`__init__` always initializes an instance of the class). If we do want these, we'd probably want them consistenly everywhere, but again I'm not sure it really adds anything valuable to the code base.

tiangolo:
Yep, we have some other docs in the reference for the things I want to expose. These are only used internally so I prefer to leave them as they are. I'll close this one but thanks for the interest! :coffee:

PR #11839, opened 2024-07-14
📝 Update terminal displays to use more compatible fonts

logan2d5:
Before:
image
After:
image

### DIFF
--- a/docs/en/docs/css/termynal.css
+++ b/docs/en/docs/css/termynal.css
@@ -20,7 +20,8 @@
     /* font-size: 18px; */
     font-size: 15px;
     /* font-family: 'Fira Mono', Consolas, Menlo, Monaco, 'Courier New', Courier, monospace; */
-    font-family: 'Roboto Mono', 'Fira Mono', Consolas, Menlo, Monaco, 'Courier New', Courier, monospace;
+    /* font-family: 'Roboto Mono', 'Fira Mono', Consolas, Menlo, Monaco, 'Courier New', Courier, monospace; */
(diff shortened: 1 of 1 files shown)


tiangolo:
📝 Docs preview for commit 6788a4e6096a36f78b1634df9a200fcf05097d9b at: https://eeb70891.fastapitiangolo.pages.dev

Ryandaydev (commented):
Appears to be a change for cosmetic reasons, is that right? Regarding "incorrect" is that a project standard?

logan2d5:
> Appears to be a change for cosmetic reasons, is that right? Regarding "incorrect" is that a project standard?

I think using monospaced fonts as terminal fonts is a project standard, rather than merely for cosmetic reasons.

Compare the two lines of code below. The second line provides better compatibility across different platforms without relying on external fonts.

```
font-family: 'Roboto Mono', 'Fira Mono', Consolas, Menlo, Monaco, 'Courier New', Courier, monospace;

font-family: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
```

Ryandaydev:
Makes sense, and I agree it's more compatible. Searching in the repo there are quite a few references to Roboto Mono so I suspect to apply that as standard for all terminal displays across the project would require a bit more review.

It's possible the title of the pull request would then be something like "Update terminal displays to use more compatible fonts".

tiangolo:
Thanks for the interest @logan2d5! :coffee: 

Roboto Mono is a monospace font. It seems to work fine in all the browsers I've seen. I would suspect it's something particular to your environment. Have you tried in a different computer, different browser, mobile?

For now, I will pass on this, but if you still have problems, please create a new discussion. :cake:

PR #11853, opened 2024-07-16
🥅 Catch `PydanticUndefinedAnnotation` in `create_response_field` and outline a clearer error message

mchataigner:
This exception may be raised when having some unresolvable forward refs. Most probably when trying to return a complex SQLAlchemy object.


YuriiMotov:
@mchataigner, thanks for you interest in FastAPI!

Could you please add a bit more context information (update initial description)?

Like:
* What problem does it solve?
* Why do you think this is important?
* Are there any links to related discussions\issues?
* Were thee any previous attempts to solve this?

YuriiMotov:
@mchataigner, are you still interested in this PR?

I tried one more time and still can't reproduce the issue.
I believe we need to add test for this

PR #11871, opened 2024-07-22
📝 Update Speakeasy URL

ndimares:
Speakeasy is now found at speakeasy.com


tiangolo:
Thanks! I updated an extra config file that was missing, this is now ready. Thank you! :nerd_face: :coffee:

PR #11874, opened 2024-07-23
📝 Warning added wrt ORJSONResponse usage at custom response docs

FedericoMassaioli:
Hi, I've lost enough time today ;-) to convince me that other people's time may be spared by adding this (maybe obvious) warning.

### DIFF
--- a/docs/en/docs/advanced/custom-response.md
+++ b/docs/en/docs/advanced/custom-response.md
@@ -32,7 +32,7 @@ But if you are certain that the content that you are returning is **serializable
 ```
 
 !!! warning
-    If you use `ORJSONResponse` and declared a Pydantic model for the response, the API documentation will still be correct, but neither validation not data filtering will be applied to response content.
+    If you use `ORJSONResponse` and declared a Pydantic model for the response, the API documentation will still follow the Pydantic model, but neither validation nor data filtering will be applied to response content.
 
 !!! info
     The parameter `response_class` will also be used to define the "media type" of the response.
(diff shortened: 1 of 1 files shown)


tiangolo:
📝 Docs preview for commit fbd46afb3bcbe0cbb5d6b976b92690e145656681 at: https://268e1b07.fastapitiangolo.pages.dev

FedericoMassaioli:
Spotted a typo and reworded a phrase.

tiangolo:
📝 Docs preview for commit d319d5f4fb835318e52df0186510cf3799b65391 at: https://cba9a7dd.fastapitiangolo.pages.dev

estebanx64:
Hey @FedericoMassaioli.

There is a phrase at the beginning of the document that explicitly says. 🤓

**But if you return a Response directly, the data won't be automatically converted, and the documentation won't be automatically generated (for example, including the specific "media type", in the HTTP header Content-Type as part of the generated OpenAPI).**

We might find this warning a little redundant.

Thanks so much for your time and interest in helping us with the documentation. 🙇

FedericoMassaioli:
Hi @estebanx64 , I saw the phrase you quoted, but I misread it as applying to `Response` only, not to `*JSONResponse`.

Moreover, I just spotted, at the end of the page introduction, the statement:

**And if that Response has a JSON media type (application/json), like is the case with the JSONResponse and UJSONResponse, the data you return will be automatically converted (and filtered) with any Pydantic response_model that you declared in the path operation decorator.**

which did not happen to be true, in my experience, which moved me to add the warning

tiangolo:
Thanks @FedericoMassaioli, this was clarified here in this PR: https://github.com/fastapi/fastapi/pull/12067, to cover not only `ORJSONResponse` but also the other classes, as the behavior applies to all.

I'll now close this one, but thanks for the interest! :coffee:

PR #11880, opened 2024-07-24
♻️ Change methods not using its bound instance to staticmethods

gitworkflows:
The method doesn't use its bound instance. Decorate this method with `@staticmethod` decorator, so that Python does not have to instantiate a bound method for every instance of this class thereby saving memory and computation. Read more about staticmethods [here](https://docs.python.org/3/library/functions.html#staticmethod).

### DIFF
--- a/docs_src/websockets/tutorial003.py
+++ b/docs_src/websockets/tutorial003.py
@@ -54,7 +54,8 @@ class ConnectionManager:
     def disconnect(self, websocket: WebSocket):
         self.active_connections.remove(websocket)
 
-    async def send_personal_message(self, message: str, websocket: WebSocket):
+    @staticmethod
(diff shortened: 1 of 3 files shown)


tiangolo:
📝 Docs preview for commit d3071952b6ee4105d3e6a3d8f9f066d6d3aa0c2f at: https://9cb8dfe1.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit 81541735ab657ce636b88ffd6e93fa97c2eea6aa at: https://77c3a885.fastapitiangolo.pages.dev

tiangolo:
Thanks for the interest. The intention of these docs is to explain the concepts, only about dependencies. This would be an optimization specific to how to handle classes that requires its own long explanation and wouldn't fit in this section, so I'll pass on this one. :coffee:

PR #11881, opened 2024-07-24
:recycle: Change default array argument in functions at query parameters docs

gitworkflows:


tiangolo:
📝 Docs preview for commit 8f968ea6146b74841df4dd4b3508441795e5e931 at: https://d474171f.fastapitiangolo.pages.dev

siavashyj:
Can you please explain what makes it dangerous?

Havsalt:
Usually, a mutable default argument can be bad if the value is mutated inside the function body or other places in the code. let's say you do `.append(42)` to list `my_list`, and it is a default value, then the default value for the next time would already be populated with 42, because it is pointing to the same list. Using `None` and then checking for `None` and setting it to a **new** list is a great way to prevent this. If the codebase of FastAPI does not mutate this value, or a unique instance is created in the decorators/other internal code (using the clever type system baked into this wonderous library) - then this would not be crucial to fix, however, it is usually a safer choice to check for `None`. A downside to this is that the parameter type annotation has to be `Annotated[T, x] | None = None` instead of simply `Annotated[T, x] = T` (where T is a mutable default value).

Hope this was of some help

tiangolo:
📝 Docs preview for commit 4077d780f6c356d40ecf247fc6c8629e6a7e0007 at: https://66f0ff2e.fastapitiangolo.pages.dev

tiangolo:
Thanks for your interest, but this doesn't apply to Pydantic, and FastAPI is built on top of Pydantic :sunglasses: https://docs.pydantic.dev/latest/concepts/models/#fields-with-non-hashable-default-values

Given that I'll close this one, but thanks for the interest. :coffee:

PR #11895, opened 2024-07-26
✨ Allow using different CapacityLimiters for Dependencies

leo-b:
Allow using separate per Dependency CapacityLimiter to avoid requests or other synchronous code being blocked.
See https://github.com/tiangolo/fastapi/discussions/11886.


patrick91:
Would it be possible to add a test for this? 😊

leo-b:
> Would it be possible to add a test for this? 😊

I added a test to check if the correct limiters are used.
If anyio >= 4.2.0 would be required, initializing the `CapacityLimiter`s could be done cleaner.

tests/test_depends_limiter.py
```diff
@@ -0,0 +1,130 @@
...
+
+# The following is a temporary workaround for anyio < 4.2.0:
+try:
+    limiters = init_limiters()
+except sniffio.AsyncLibraryNotFoundError:
+    # not in an async context yet
+    async def _init_limiters():
+        return init_limiters()
+
+    limiters = anyio.run(_init_limiters)
```
YuriiMotov, line 41: Is it possible to move this to lifespan event? I think this should work and we will be able to simplify the code

YuriiMotov (commented):
@leo-b, Thanks for your interest and efforts!

I think we should first wait for Sebastián to validate the idea, then we will need to work more on tests and docs

tiangolo:

Thank you! 🍰

PR #11898, opened 2024-07-27
📝 Tell the readers not to use `|` in `extra_models/tutorial003_py310.py`

wxh06:
`extra_models/tutorial003_py310.py` was exactly the same as `extra_models/tutorial003.py`
This may confuse first-time readers

https://github.com/tiangolo/fastapi/blob/0976185af96ab2ee39c949c0456be616b01f8669/docs_src/extra_models/tutorial003.py#L33

https://github.com/tiangolo/fastapi/blob/0976185af96ab2ee39c949c0456be616b01f8669/docs_src/extra_models/tutorial003_py310.py#L33


tiangolo:
📝 Docs preview for commit d5f23097dc6159f40e325ae877b0baebbdd847b4 at: https://fb0867bb.fastapitiangolo.pages.dev

tiangolo:
📝 Docs preview for commit 85f5d8489deee2c1789c28f1884aba9285eb1fc9 at: https://d2e43bf2.fastapitiangolo.pages.dev

svlandeg (commented):
Hi, thanks for the PR!

I can see how this is confusing at first sight. I'm not 100% sure whether adding a comment to the code snippet is the best way forward though. In the end, the paragraph right below this code does clearly mention:

```
Because we are passing it as a value to an argument instead of putting it in a type annotation, we have to use Union even in Python 3.10.
```

As such, I would personally not change the code snippet as the surrounding text offers the explanation in more detail. 

If we do change it however, we need to update the highlight of L33 to L35 to capture the line where `Union[PlaneItem, CarItem]` is mentioned.

tiangolo:
Thanks for the interest @wxh06. :nerd_face: 

Yep, as @svlandeg says, this would be explained in the text there. So I'll pass on this one, but thanks for the effort! :coffee:

PR #11522, opened 2024-05-02
✨ Add FastAPI CLI, the new `fastapi` command

tiangolo:
✨ Add FastAPI CLI, the new `fastapi` command

PR #11935, opened 2024-08-02
♻️ Add support for `pip install "fastapi[standard]"` with standard dependencies and `python -m fastapi`

tiangolo:
♻️ Add support for `pip install "fastapi[standard]"` including the standard dependencies, the CLI, Uvicorn, etc. and make `fastapi` not include the optional standard dependencies. `fastapi` and `fastapi-slim` are now the same.

This also adds support for calling it with:

```bash
python -m fastapi
```

And if someone doesn't install `fastapi[standard]` and tries to call the `fastapi` command, it will show them an error telling them to install `fastapi[standard]`.

## Summary

Install with:

```bash
pip install "fastapi[standard]"
```

This also upgrades `fastapi-cli[standard] >=0.0.5`.

#### Technical Details

Before this, `fastapi` would include the standard dependencies, with Uvicorn and the `fastapi-cli`, etc.

And `fastapi-slim` would not include those standard dependencies.

Now `fastapi` doesn't include those standard dependencies unless you install with `pip install "fastapi[standard]"`.

Before, you would install `pip install fastapi`, now you should include the `standard` optional dependencies (unless you want to exclude one of those): `pip install "fastapi[standard]"`.

This change is because having the standard optional dependencies installed by default was being inconvenient to several users, and having to install instead `fastapi-slim` was not being a feasible solution.

Discussed here: https://github.com/fastapi/fastapi/pull/11522 and here: https://github.com/fastapi/fastapi/discussions/11525

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ FastAPI stands on the shoulders of giants:
 <div class="termy">
 
 ```console
-$ pip install fastapi
+$ pip install "fastapi[standard]"
 
 ---> 100%
 ```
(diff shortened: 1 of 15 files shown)


hamadrehman:
image
is this a typo ? 
`python -m fastapi` instead of `python -m python`

PR #11940, opened 2024-08-02
🙈 Add .coverage* to `.gitignore`

gitworkflows:


Kludex:
Is the project generating anything other than `.coverage`?

svlandeg (commented):
Hi @gitworkflows, thanks for the PR! 

<s>The change in this PR doesn't seem necessary to me though, so I'm not actually sure we should merge it.</s>

killjoy1221:
`.coverage*` files are being created when running coverage locally. It's because coverage is configured with
https://github.com/fastapi/fastapi/blob/75705617a66300847436e39ba703af1b8d109963/pyproject.toml#L177-L178

The proper way to handle these files is to run `coverage combine` after running the tests.

tiangolo:
Thanks! :cake: :coffee:

PR #11945, opened 2024-08-03
Fix typos

AmeerArsala:


alejsdev:
Hi @AmeerArsala, it's not a typo, it's written as intended, so this one will be closed but thanks for the effort :v:

PR #11949, opened 2024-08-04
🔇 Ignore warning from attrs in Trio

tiangolo:
🔇 Ignore warning from attrs in Trio

A recent release of a attrs marked a parameter `hash` as deprecated. It is used by Trio (fixed here: https://github.com/python-trio/trio/pull/3054, but still unreleased). Trio is used by AnyIO's pytest plugin.

This currently shows a deprecation warning that is breaking tests in FastAPI. This PR adds an ignore for that warning so tests here can keep running.


hynek:
I'm sowwy, if I knew it would've even reach the far shores of FastAPI, I would've only soft-deprecated it. 🥺

PR #11957, opened 2024-08-06
📝 Tweak middleware code sample `time.time()` to `time.perf_counter()`

domdent:
Change `time.time()` → `time.perf_counter()`

Given the performance of FastAPI, [time.perf_counter](https://docs.python.org/3/library/time.html#time.perf_counter) offers a more accurate way to time short durations: "clock with the highest available resolution to measure a short duration".


estebanx64:
Hi @domdent 👋

Thanks for your interest in helping us to improve the docs. 🙇

It seems the perf_counter is better for this scenario since it provides a high-resolution time counter for measuring short durations.

It's gonna be merged now. 🚀

PR #11960, opened 2024-08-06
📝 Add note in Docker docs about ensuring graceful shutdowns and lifespan events with `CMD` exec form

GPla:
There is a problem when using `docker compose` and the `shell form`  of the `CMD` command (`CMD fastapi run app/main.py --port 80`) to run `fastapi`/`uvicorn`.  

Because the `shell form` implicitly invokes a shell (`sh -c "fastapi run ..."`) when starting the container, the `uvicorn` process is not running as PID 1, the shell `sh` is. Using the `exec form` (`CMD ["fastapi", "run", "app/main.py", "--port", "80"]`), `uvicorn` is running as PID 1.  When the user attempts to stop the container by pressing <kbd>Ctrl</kbd> + <kbd>C</kbd>, `docker compose` will send a `SIGTERM` to the process with PID 1. In the `shell form`, the `SIGTERM` will not reach `uvicorn`/`fastapi`. This prevents a graceful shutdown and lifespan events from being triggered / finished. Everything works as expected when using the `exec form`. 

There have been questions on Stack Overflow of [lifespan events not getting triggered](https://stackoverflow.com/questions/77041522/fastapi-lifespan-event-in-docker-container-running-not-executing-the-shutdown/78841238#78841238) and user often recommend using the `shell form` in answers without being aware of this issue. Also, the first result that comes up in Google when searching for `graceful shutdown fastapi` is #6912 which does not help much.


tiangolo:
Great, thank you @GPla! :rocket: I tweaked it a bit and made it its own section so we can link to it more easily. :nerd_face:

PR #11968, opened 2024-08-08
🌐 Add Portuguese translation for `docs/pt/docs/reference/security/index.md`

marcelomarkus:
- Portuguese translation
docs/pt/docs/reference/security/index.md


nazarepiedady (approved):
I think that we should translate the headings of sections too, but if they are not supposed to be translated so we have a well-done contribution.

ceb10n (approved):
❤️

cassiobotaro (approved):

nazarepiedady (approved):

tiangolo:
Thank you for your contribution! :bow: 

I recently realized that there are some sections where the benefit of translating them has a tradeoff with the content... for example, the content of the reference is injected from the source files, and most of the content in that section ends up being in English either way... and then there's a chance that if the reference is updated including other objects to be injected, the translation would be obsolete. :pensive: 

So I decided not to translate some sections, including the reference. https://fastapi.tiangolo.com/contributing/#dont-translate-these-pages

So, I'm gonna close this one, but thanks for the effort! :bow:

PR #11978, opened 2024-08-09
📝 Highlight correct line in tutorial `docs/en/docs/tutorial/body-multiple-params.md`

svlandeg:
The tutorial section https://fastapi.tiangolo.com/tutorial/body-multiple-params/#multiple-body-params-and-query is about adding query parameters on top of (several) body parameters. In the sections before it, the tutorial had introduced 3 body parameters: `item`, `user` and `importance`. This section then focuses on adding the query parameter `q`, so it seems to me like the next line should be highlighted in the code snippets - the line defining `q` instead of the one defining `importance`. 

Or, alternatively, we could highlight all 4 lines (the 3 body parameters + the additional query parameter)


tiangolo:
Great, thank you! :bow:

PR #11988, opened 2024-08-10
📝 Improve documentation on using dataclasses in FastAPI

AdamMusa:
I was reading the FastAPI documentation this morning and came across the following sentence: 'FastAPI is built on top of Pydantic, and I have been showing you how to use Pydantic models to declare requests and responses. But FastAPI also supports using [dataclasses](https://docs.python.org/3/library/dataclasses.html) the same way.' While I don't disagree with the content, it is widely known that FastAPI was initially maintained by a single person. Therefore, I decided to rephrase the sentence to remove the personal tone.

Revised the documentation to provide a clearer explanation of how FastAPI integrates with Pydantic and supports dataclasses. The previous language was instructional and personal, which may not be appropriate for all readers. The updated text maintains a neutral tone and emphasizes the functionality of FastAPI in handling dataclasses similarly to Pydantic models.

### DIFF
--- a/docs/en/docs/advanced/dataclasses.md
+++ b/docs/en/docs/advanced/dataclasses.md
@@ -1,6 +1,6 @@
 # Using Dataclasses
 
-FastAPI is built on top of **Pydantic**, and allows the use of Pydantic models to declare requests and responses. 
+FastAPI is built on top of **Pydantic**, and allows the use of Pydantic models to declare requests and responses.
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for the interest @AdamMusa, but I think it's fine as it is. So I'll pass on this one. But thanks! :coffee:

PR #11996, opened 2024-08-12
📝 Edit the link to the OpenAPI "Responses Object" and "Response Object" sections in the "Additional Responses in OpenAPI" section

VaitoSoi:
Because the link to the "Responses Object" and "Response Object" section of OpenAPI in the "Additional Responses in OpenAPI" section did not lead to the correct section. So I corrected the link to lead to that section :D


waketzheng (approved):

svlandeg (approved):
Thanks for the PR! I double checked and you're right, the old link was not leading to the correct section. Thanks for contributing the fix! 🙏

tiangolo:
Great, thanks @VaitoSoi! :cake: 

And thanks @svlandeg for the review :bow:

PR #12023, opened 2024-08-15
✏️ Fix typo for dependencies docs in script example tutorial006

Roshan310:
verify_token() function is missing the 'return x_token' value. The function is missing a return statement after the 'if' statement. Now, I've added the return statement and am proposing this change.

### DIFF
--- a/docs_src/dependencies/tutorial006.py
+++ b/docs_src/dependencies/tutorial006.py
@@ -6,6 +6,7 @@ app = FastAPI()
 async def verify_token(x_token: str = Header()):
     if x_token != "fake-super-secret-token":
         raise HTTPException(status_code=400, detail="X-Token header invalid")
+    return x_token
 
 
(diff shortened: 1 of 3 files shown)


Roshan310:
the file where I've made the change is located here: docs_src/dependencies/tutorial006_an_py39.py

estebanx64:
Thanks for your time helping us! 🙇

We'll be merging this little fix in the docs scripts soon! 🤓

estebanx64:
Thanks for helping us fix this. 🙇

We'll be reviewing it and merging it soon. 🤓

estebanx64:
Hi @Roshan310.

The return here is not mandatory as the context in the docs says you can or cannot return a value. It means you can use dependencies to validate some logic and it isn't mandatory to return something.

I'll be closing this PR since it is not necessary any changes.

Thank you again for taking care of the details in the documentation. 🙇

PR #12028, opened 2024-08-16
🌐 Update Chinese translation for `docs/zh/docs/tutorial/dependencies/dependencies-with-yield.md`

xuvjso:
https://github.com/fastapi/fastapi/pull/10831 ensures that the code following `yield` executes after the response is create but before it is sent to the client. The statement in https://github.com/fastapi/fastapi/issues/3620#issuecomment-888081460 was outdated, it is now safe to commit the session after `yield` as the code is guaranteed to execute after the response is created but before it is sent.
```python
def get_db() -> Session:
    session = SessionLocal()
    try:
        yield session
        session.commit()
    except:
        session.rollback()
        raise
    finally:
        session.close()
```


estebanx64:
Thanks for your time and interest in helping us! 🙇

We are going to review these little changes 🤓

xuvjso:
I noticed that the chinese version of the "dependencies with yield" section of the documentation is outdated, so I provided an updated translation to prevent users from being misled by outdated information and to align with the latest English documentation.

Aruelius (approved):

alejsdev:
Hi @xuvjso 
Thanks for your interest in contributing to FastAPI :bowing_woman: 
To avoid problems with the labels. Could you better separate the EN from the ZH docs in another PR? Thanks in advance! :pray:

alejsdev:
Great! Thanks for your contribution :partying_face: @xuvjso 
And thanks for the review @Aruelius :100:

xuvjso:
> Great! Thanks for your contribution 🥳 @xuvjso And thanks for the review 💯 @Aruelius

It's my pleasure. 😊

PR #12031, opened 2024-08-16
🔥 Remove Sentry link from Advanced Middleware docs

alejsdev:
🔥 Remove Sentry link from Advanced Middleware docs

As this is in the section about middlewares and the new integration is not really a middleware, better to remove it from this page.

We should probably add it later to another section. :thinking:


tiangolo:
Great, thanks! :nerd_face: :cake:

PR #12033, opened 2024-08-17
👷 Update FastAPI People to use local script and shared dependencies

tiangolo:
👷 Update FastAPI People to use local script and shared dependencies


tiangolo:
This was covered in another PR. :fire:

PR #12045, opened 2024-08-19
📝 Update `docs/en/docs/tutorial/dependencies/dependencies-with-yield.md`

xuvjso:
After FastAPI 0.106.0, the code after yield now runs after the response is created but before it is sent to the client.


svlandeg (commented):
Hi, thanks for your contribution! 🙏 

I'm not sure I fully understand the suggestion here, and what "but before sending it" actually means. You mention that this has changed after FastAPI 0.106.0, can you share a link to the relevant commit/PR/change log?

[UPDATE]: Ah, found it, this is probably related to https://github.com/fastapi/fastapi/pull/10831?

svlandeg (approved):
Yep I think this looks good, thanks again! 🙏

tiangolo (approved):
Yep, good catch @xuvjso. thanks! :rocket: :cake:

PR #12062, opened 2024-08-23
📝 Update comma in `docs/en/docs/async.md`

Alec-Gillis:
Add commas to a series in a sentence under the Coroutines section to enhance readability.

### DIFF
--- a/docs/en/docs/async.md
+++ b/docs/en/docs/async.md
@@ -387,7 +387,7 @@ In previous versions of NodeJS / Browser JavaScript, you would have used "callba
 
 ## Coroutines
 
-**Coroutine** is just the very fancy term for the thing returned by an `async def` function. Python knows that it is something like a function, that it can start, and that it will end at some point, but that it might be paused ⏸ internally too, whenever there is an `await` inside of it.
+**Coroutine** is just the very fancy term for the thing returned by an `async def` function. Python knows that it is something like a function, that it can start and that it will end at some point, but that it might be paused ⏸ internally too, whenever there is an `await` inside of it.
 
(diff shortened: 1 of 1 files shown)


docs/en/docs/async.md
```diff
@@ -387,7 +387,7 @@ In previous versions of NodeJS / Browser JavaScript, you would have used "callba
 
 ## Coroutines
 
-**Coroutine** is just the very fancy term for the thing returned by an `async def` function. Python knows that it is something like a function that it can start and that it will end at some point, but that it might be paused ⏸ internally too, whenever there is an `await` inside of it.
+**Coroutine** is just the very fancy term for the thing returned by an `async def` function. Python knows that it is something like a function, that it can start, and that it will end at some point, but that it might be paused ⏸ internally too, whenever there is an `await` inside of it.
```
svlandeg, line 390:
```suggestion
**Coroutine** is just the very fancy term for the thing returned by an `async def` function. Python knows that it is something like a function, that it can start and that it will end at some point, but that it might be paused ⏸ internally too, whenever there is an `await` inside of it.
```

svlandeg (commented):
Thanks for the contribution! I agree that we could make this sentence more readable. I would however suggest to only insert the first comma, so that we don't have a comma followed immediately by "and". The "it can start and it wil end" belongs together in a subsentence, IMO.

docs/en/docs/async.md
Alec-Gillis: Uh oh, we have an Oxford comma dissenter in our midst 😆 Totally works for me - it definitely makes the sentence flow better!

svlandeg (approved):

tiangolo:
Cool, thanks! :nerd_face:

PR #12066, opened 2024-08-24
🐛 Fix deadlock that can occur when closing dependencies during response model validation

noahklein00:
Related to #5122

In my project, I had noticed I was getting a sqlalchemy `QueuePool` limit error like others had noted in https://github.com/fastapi/fastapi/discussions/6628. This was marked as probably fixed by https://github.com/fastapi/fastapi/pull/5122.

Even on fastapi commit `e4727ed2` (> v0.112.2), you can still run into this issue when running the tests created by the author of #5122 - https://github.com/harsh8398/fastapi-dependency-issue. I also notice others running into similar deadlocks after picking up the `0.82.0` patch: https://github.com/vipyrsec/dragonfly-mainframe/issues/235.

The author of the issue/discussion #6628 noted that when adding a `response_model` to the endpoint function, they noticed that the suggested quick fixes (i.e. using a contextmanager, async session manager, etc.) didn't seem to work either. It seems like the response model validation runs into the same issue that when using `Depends`, a deadlock can occur when closing sessions.

I believe this pr fixes this `Depends` issue all the way.

I tested this pr against https://github.com/harsh8398/fastapi-dependency-issue and notice that all test cases pass equally as fast! This patch also fixes the deadlocks/timeouts I was seeing in my own codebase.

**UPD YuriiMotov:** related discussion on SQLModel repo (https://github.com/fastapi/sqlmodel/discussions/1417)


svlandeg:
Thanks for the PR! Really appreciate the details and links to related discussions & issues as well 🙏 

We'll get back to you after the team has had a chance to review.

In the meantime, I was wondering whether the tests from https://github.com/harsh8398/fastapi-dependency-issue could/should be included here as unit tests?

noahklein00:
I've added a unit test. Without the fix, it will deadlock. I tried to make it as fast as possible and also not use any dependencies like sqlalchemy that it seemed like they were unsure of in the previous pr #5122.

If this looks good, I could expand it to cover the multiple cases that are tested in the test repo.

goccert25:
> The author of the issue/discussion https://github.com/fastapi/fastapi/discussions/6628 noted that when adding a response_model to the endpoint function, they noticed that the suggested quick fixes (i.e. using a contextmanager, async session manager, etc.) didn't seem to work either. It seems like the response model validation runs into the same issue that when using Depends, a deadlock can occur when closing sessions.

Adding a comment here just to say that I too am running into this issue (and this caused an outage in production recently). Also can confirm that this PR eliminates the deadlocking and significantly improves the response times

Lego3:
I have also encountered this issue. Is there a chance it will go into a not too distant release? Who might know?

olgeni:
We have almost daily catastrophic crashes in production because of this, and the amount of work to remove the db-related Depends is basically the same as rewriting the product, so it would be quite nice if this somehow could end up in some release 😕

olgeni:
We still had database-related deadlocks after patching the virtualenv manually with this fix 👀

After another “stern warning” from the client's lawyers, we switched to pgbouncer and NullPool, and now all is good and well.

Apparently the model thing was not directly involved in the locks (we still have the same models!) \o/

YuriiMotov:
Looks like this is still an issue in 0.118.0.

To reproduce:

<details><summary>Code of the app in the details</summary>

**main.py**
```py
from typing import Union

from sqlalchemy.orm import Session

from sqlalchemy.orm import Mapped, mapped_column

from sqlalchemy import create_engine, select
from sqlalchemy.orm import sessionmaker, DeclarativeBase


engine = create_engine("postgresql://user:mysecretpassword@localhost:5432/some_db")
SessionLocal = sessionmaker(bind=engine)


class Base(DeclarativeBase):
    pass


def get_db():
    db = SessionLocal()
    try:
        yield db
    finally:
        db.close()


class ItemModel(Base):
    __tablename__ = "items"

    id: Mapped[int] = mapped_column(primary_key=True, index=True)
    title: Mapped[str] = mapped_column(index=True)
    description: Mapped[str] = mapped_column(index=True)


class ItemSchema(BaseModel):
    id: int
    title: str
    description: Union[str, None] = None

app = FastAPI()


@app.on_event("startup")
def startup():
    Base.metadata.create_all(bind=engine)


@app.get("/items", response_model=list[ItemSchema])
def read_items(db: Session = Depends(get_db)):
    return db.scalars(select(ItemModel)).all()
```

</details>

1) Adjust the DB connection string in the script

2) Run PostgreSQL (create DB)

3) Run app with `uvicorn main:app`

4) run test with `ab -n 200 -c 200 http://localhost:8000/items`

It will freeze for 30 seconds and then you will see errors like `sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00` in app's log.

Remove `response_model` and it works without errors

---

As for suggested solution, I think we should try to find alternative solution (without using one-time capacity limiter for response validation)

olgeni:
Meanwhile the NullPool solution is holding up for us, in case anybody needs a quick fix 😅

olgeni:
Is this still on the radar? Is it fixed? Is hope lost? I lost track of it 😅

svlandeg:
> Is this still on the radar? Is it fixed? Is hope lost? I lost track of it 😅

The status of this PR is such that Yurii reviewed it in October 2025, confirmed the issue, and left some comments:

> As for suggested solution, I think we should try to find alternative solution (without using one-time capacity limiter for response validation

After this, we've marked it internally for Tiangolo to review, but as you can imagine, he's quite a busy man 🥲  I've now fixed the merge conflict to make his job slightly easier.

Thanks for your patience 🙏

elemoine:
This patch fixes a real bug / deadlock in FastAPI, that will trigger for people using FastAPI at scale.  A [similar fix](https://github.com/fastapi/fastapi/pull/5122) was applied to `concurrency.py` 4 years ago.  So I am not quite sure why this PR doesn’t get the attention it deserves.

olliemath:
> As for suggested solution, I think we should try to find alternative solution (without using one-time capacity limiter for response validation

In general, instead of creating one-time limiters how about some sort of limiter-with-overflow pattern. I notice that there are only two locations in the codebase that seem to require an escape-hatch to acquire and run code when the main threadpool is full.

The pattern looks like this:

- maintain a global limiter with capacity of Limit + Overflow
- maintain a separate limiter with capacity of Limit

By default execute using the default limiter (capacity = Limit), and for certain high priority tasks (this and the one in concurrency.py) do not use the lower capacity limiter. Ideally you wrap in a helper function like `run_in_threadpool(..., with_overflow=False)`.

I'm planning to actually build this out to check it really works under high load. The only issue I can see is with messing with the limiter in use - anybody already directly modifying anyio's default token limit would need to change their setup (a helper function could also be provided to set limit+overflow directly).

YuriiMotov:
I think we need to reserve one thread for priority tasks (response validation and maybe some others) and use it if main pool is exhausted

olliemath:
Yeah, the hard part here is using the extra thread "only if the main pool is exhausted".

This naive code has a race condition between checking a token is available and actually executing the thread:
```python
if limiter.available_tokens > 0:
    await anyio.to_thread.run_sync(fn)
else:
    # run with overflow thread
```

Flipping the other way around and wrapping almost all non-critical `run_in_threadpool` calls in an `async with _anti_deadlock_limiter():` (with capacity 39) you can reserve an extra thread. This is what I've tested locally and it works well for performance under high load. We would need to provide a distinct `fastapi.run_in_threadpool` for users to use, as using `starlette.run_in_threadpool` could still consume all the threads (since it's not wrapped in the "default_limit - 1" capacity limiter).

olliemath:
I have enough now to write and release a small 1-file library that monkey-patches both fastapi and starlette's `run_in_threadpool`, which can ease some pressure in getting the final solution thought through and merged.

olliemath:
Created: https://pypi.org/project/fastapi-overflow/ you can now bypass the deadlocking via

```python
import fastapi_overflow
fastapi_overflow.patch()
```

olliemath:
I added my multiple-capacity-limiters (1 extra thread reserved) changes over here: https://github.com/fastapi/fastapi/pull/15372 using the tests in this PR to evaluate them. The required changes are unfortunately more extensive than the diff on this MR. Let me know if you'd like to merge(?) them with this PR, use them, or if it's not a good direction to go in.

Apologies for the large amount of mail anyone subscribed to this PR will be getting :mailbox: :envelope: :envelope:

olliemath:
So I spoke to the maintainer of anyio and they very kindly convinced me that the overflow idea is overcomplicating things. In general, the advice is just: maintain one limiter for acquiring resources and a separate one for releasing them.

In which case the only recommended modification needed to this PR would be:

- a single extra global limiter and helper function for releasing resources in concurrency.py
- replacing both uses of the "unlimited" one-time limiters with the helper function
- likely with a default capacity of ~5 (since we cannot control how expensive the user's teardown in a dependency is, 1 seems too low)
- along with a helper method to modify global capacity limits (since we cannot control how expensive the user's teardown in a dependency is, any number could be "wrong" for somebody's personal use case)

I'm happy to make these changes if you're happy with them @YuriiMotov or if @noahklein00 has not moved on after all this time he could make them and I can help review?

olliemath:
> Yeah, the hard part here is using the extra thread "only if the main pool is exhausted".
> This naive code has a race condition between checking a token is available and actually executing the thread:

Quick update. I added code to [demo the deadlock issues with "only if the main pool is exhausted"](https://github.com/olliemath/fastapi/pull/2). I don't think they can be overcome without changes to anyio itself - which is not impossible, but deeper that one might think.

I also wrote up the simple one-time limiter version of this PR over here: https://github.com/fastapi/fastapi/pull/15388 Seemingly the main blocker is not creating a fix (there are several valid approaches), but knowing which direction is deemed acceptable.

Rony-ZenAlden:
I spent time reproducing and analyzing this — I can confirm the fix works for the case it targets, and I found a related case it *doesn't* cover that may be worth noting before merge.

**Confirmed:** on current `main`, a sync endpoint with a `response_model` and a sync `yield` dependency holding a blocking resource deadlocks under load. Reproduced deterministically by shrinking the AnyIO thread limiter to 5 tokens and firing 20 concurrent requests. Root cause is as described — response-model validation (`serialize_response`) competes for the shared threadpool with endpoint threads blocked on the resource. Applying this PR's dedicated `CapacityLimiter(1)` for validation resolves it (same repro then completes).

**One gap:** this covers the case where the resource is acquired in the *endpoint* (only validation is starved). If the resource is instead acquired in the dependency **setup** (before `yield`) and released in teardown, the deadlock persists **even with this patch** — because the starved operation becomes the endpoint execution itself (`run_endpoint_function` -> `run_in_threadpool(dependant.call)`), which can't be given a dedicated limiter without removing the shared pool's concurrency cap. I verified this still deadlocks (5 tokens / 20 requests) with the PR applied.

That second case likely isn't fixable the same way (it's inherent to running endpoint code in a bounded shared pool), so it may be out of scope here — but documenting it would help, since users hitting connection-pool deadlocks may structure dependencies either way. Happy to share the repro scripts.

PR #12067, opened 2024-08-24
📝 Clarify `response_class` parameter, validations, and returning a response directly

tiangolo:
📝 Clarify `response_class` parameter, validations, and returning a response directly


FedericoMassaioli:
Much clearer, thank you for the changes.

PR #12069, opened 2024-08-25
📝 Update docs about serving FastAPI: ASGI servers, Docker containers, etc.

tiangolo:
📝 Update docs about serving FastAPI: ASGI servers, Docker containers, etc.


idan3:
Should replacing one server with another be seamless? Let's say using Granian instead of the default.

tiangolo:
It would be better to check each project's docs

PR #12070, opened 2024-08-25
🌐 Update Chinese translation for `docs/zh/docs/how-to/index.md`

synthpop123:
The `小技巧` label in the admonition was causing rendering issues in MkDocs, so replacing it with the standard `tip` label fixes this problem.

### DIFF
--- a/docs/zh/docs/how-to/index.md
+++ b/docs/zh/docs/how-to/index.md
@@ -6,7 +6,7 @@
 
 如果某些内容看起来对你的项目有用，请继续查阅，否则请直接跳过它们。
 
-/// tip
+/// tip | 小技巧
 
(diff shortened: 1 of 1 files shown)


alejsdev:
Hi @synthpop123, you're right! This is an error, but we usually prefer to keep the original (for the styling) and the translation, so it would be something more like this:

/// tip | translated-word

Could you update it, please? :smile: 

Thank you :bowing_woman:

synthpop123:
> Hi @synthpop123, you're right! This is an error, but we usually prefer to keep the original (for the styling) and the translation, so it would be something more like this:
> 
> /// tip | translated-word
> 
> Could you update it, please? 😄
> 
> Thank you 🙇‍♀️

Thank you for the quick review! I’ve updated the PR with the suggested change.

Aruelius (approved):

svlandeg (approved):
Nice catch, thanks! 🙏

alejsdev:
Thank you! @synthpop123 @svlandeg :tada:

PR #12079, opened 2024-08-27
🧪 Add test for segmentation fault when using `asyncio.eager_task_factory`

07pepa:
This add tests for  https://github.com/fastapi/fastapi/discussions/11868. (disabled because they cause segfault)

TLDR: using eager_task_factory from asyncio causes segfault and crash
Cause: unknown

Here are witness runs verifiing this is real issue https://github.com/fastapi/fastapi/actions/runs/10573940061/job/29294417766?pr=12079
(3.8 error is me using too new syntax for 3.8)

I at least added test that are curently skiped
There is also known workaround so i could document that.

### DIFF
--- a/tests/test_asyncio_eager_task_factory.py
+++ b/tests/test_asyncio_eager_task_factory.py
@@ -1,0 +1,30 @@
+import asyncio
+from contextlib import asynccontextmanager
+from typing import Dict
+
+import pytest
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+
+
+@asynccontextmanager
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks, it's definitely useful to have a test that captures the segfault behaviour.

As we don't currently have a fix at hand, I would be in favor or merging the skipped test into the library until we are able to fix the underlying issue. I'll yet double check with Tiangolo though.

Kludex:
Does it happen if you use `httpx.AsyncClient`?

PR #12117, opened 2024-09-02
♻️ Refactor deciding if `embed` body fields, do not overwrite fields, compute once per router, refactor internals in preparation for Pydantic models in `Form`, `Query` and others

tiangolo:
♻️ Refactor deciding if `embed` body fields, do not overwrite fields, compute once per router


fastapi/dependencies/utils.py
```diff
@@ -704,21 +714,41 @@ def request_params_to_args(
...
+        # If it explicitly specifies it is embedded, it's not top level
+        or getattr(first_field.field_info, "embed", None)
+        # If it's a Form (or File) field, it has to be a BaseModel to be top level
+        # otherwise it has to be embedded, so that the key value pair can be extracted
+        or (
+            isinstance(first_field.field_info, params.Form)
+            and not isinstance(first_field.type_, BaseModel)
+        )
+    )
+    return embed
```
includeamin, line 734: NIT 🙈

no logic changes, just a small simplification

```suggestion
def _should_embed_body_fields(fields: List[ModelField]) -> bool:
    if not fields:
        return False
    
    # A top level field has to be a single field, not multiple
    if len(fields) > 1:
        return True
    
    first_field = fields[0]
    # If it explicitly specifies it is embedded, it's not top level
    if getattr(first_field.field_info, "embed", None):
        return True
     
    # If it's a Form (or File) field, it has to be a BaseModel to be top level
    # otherwise it has to be embedded, so that the key value pair can be extracted
    if isinstance(first_field.field_info, params.Form) and not isinstance(first_field.type_, BaseModel):
        return True
        
     return False
     
```

includeamin: Oops, my bad! 😅 The PR was still a draft. Just ignore it for 🙈

tiangolo: Thanks! Good idea. I needed to tweak it a bit, but yep, it simplifies it.

PR #12118, opened 2024-09-02
💚 Set `include-hidden-files` to `True` when using the `upload-artifact` GH action

svlandeg:
The CI was failing because hidden files are automatically excluded since `actions/upload-artifact` [v4.4.0](https://github.com/actions/upload-artifact/releases/tag/v4.4.0).

This PR specifically sets the parameter to `True` so the `.coverage` files are found.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -113,6 +113,7 @@ jobs:
         with:
           name: docs-site-${{ matrix.lang }}
           path: ./site/**
+          include-hidden-files: true
 
   # https://github.com/marketplace/actions/alls-green#why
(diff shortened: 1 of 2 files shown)


svlandeg:
There's one more occurrence of this action in the `build-docs.yml`, but I think we can stick to the defaults for that one?

tiangolo:
Nice! Thank you! :rocket: 

Can you do it in `build-docs.yml` too? Those generate a `.cache` directory with social preview images for each page... I'm not sure if those files are used or not, but better to be safe.

tiangolo (approved):
Nice! Thank you! :rocket:

PR #12129, opened 2024-09-05
✨ Add support for Pydantic models in `Form` parameters

tiangolo:
✨ Add support for Pydantic models in `Form` parameters

---

Restores https://github.com/fastapi/fastapi/pull/12127 (reverts the revert https://github.com/fastapi/fastapi/pull/12128).


rmawatson:
Hi,

This commit breaks [this](https://github.com/fastapi/fastapi/discussions/13690) use case. There are other reports of the same thing.

Thanks.

Issue #1661, opened 2020-06-30
Allow None to be used as a valid json value.


bharel:
### First check

* [X] I added a very descriptive title to this issue.
* [X] I used the GitHub search to find a similar issue and didn't find it.
* [X] I searched the FastAPI documentation, with the integrated search.
* [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [X] I already read and followed all the tutorial in the docs and didn't find an answer.
* [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [] After submitting this, I commit to one of:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Description

ATM, if you receive a post with the following json: `{data: null}`

And you have the following definition:
```
@app.post("/")
def post_data(data: Optional[str] = Body(...)):
  pass
```

FastAPI will say `data` was not passed. That's because it internally uses `None` as a placeholder for non-existing fields, instead of a sentinel value.

I wish to explicitly require `null` as one of the possible values, and not put a default value of `None`.

I've attempted to [open a PR](https://github.com/tiangolo/fastapi/pull/1660), but it failed. It looks like the issue is with `dependencies.utils.request_body_to_args` using `None` as a default value for `value`.

### Environment

* OS: [e.g. Linux / Windows / macOS]: macOS
* FastAPI Version [e.g. 0.3.0]: 0.58.1
* Python version: 3.8.3

Kludex:
Well, in your example `data` is not a json. You're trying to use it as a `Query`.

You're probably aiming for:
```
from typing import Optional
from fastapi import FastAPI
from pydantic import BaseModel

app = FastAPI()

class Data(BaseModel):
    data: Optional[str]

@app.post("/")
def post_data(data: Data):
    return data
```
Which works just fine.

bharel:
@Kludex Meant to use it with Body(...). Perhaps it works only in a model and not as a body param?

Kludex:
[...]

PR #12135, opened 2024-09-05
✨ Pass none instead of the default value for optional parameters when null is received

merlinz01:
This PR addresses the case where a parameter is optional but has a default that is not `None`, and the user wishes to detect that a `null` value was passed, versus the value not being defined at all.

### Use case

This is an example of a use case for which this behavior is crucial:
```python
@app.patch('/users/{user_id}')
def edit_user(user_id: int, name: Annotated[str | None, Body(embed=True)] = None, organization_id: Annotated[int | None, Body(embed=True)] = -1):
    # user_id is the ID of a user in the database
    user = lookup_user(user_id)
    # If a name was given, save the user's name
    if name is not None:
        user.name = name
    # If an organization id was given, edit the organization to which the user belongs
    # If organization_id is -1, we know that no value has been received
    if organization_id != -1:
        # The user might not belong to any organization and we want to handle that case
        if organization_id is not None:
            lookup_organization(organization_id)
        user.organization_id = organization_id # int | None
    save_user(user)
    return {"success": True}
```

### Previous behavior

```
PATCH /users/123
{}
```
would result in `organization_id == -1` (as expected), but:
```
PATCH /users/123
{
  "organization_id": null
}
```
would also result in `organization_id == -1`, which provides no way to distinguish the two.

### New behavior

```
PATCH /users/123
{}
```
results in `organization_id == -1` (as before), but:
```
PATCH /users/123
{
  "organization_id": null
}
```
now results in `organization_id is None`.

### Additional comments

Resolves #6058 and #1661 .

Tests included, all tests passing.


merlinz01:
Lint and test on all Python+Pydantic versions is succeeding as far as this is concerned. I'm seeing the test failure for the OpenAPI schema on the main branch also, so I don't think it's relevant for this.

Let's put @bharel's world [back together](https://github.com/fastapi/fastapi/issues/1661#issuecomment-652086321)! :smile:

merlinz01:
Added a test for full coverage.

svlandeg:
> I'm seeing the test failure for the OpenAPI schema on the main branch also, so I don't think it's relevant for this.

This was indeed an unrelated CI failure, and is fixed now after merging in the latest changes from `master` 🎉

merlinz01:
Is there a changelog that should be added?

bharel:
> Let's put @bharel's world [back together](https://github.com/fastapi/fastapi/issues/1661#issuecomment-652086321)! :smile: 

Two wars and a global pandemic later, at least now we will be able to differentiate nulls from unspecified values 😅

It's all about the little wins really...

merlinz01:
Great job on the recent releases, when do you think you can incorporate this PR? I am waiting on this to implement some fixes on my own project. I just updated this to the latest commits on `master`, and all tests and checks are passing.

merlinz01:
Any updates?

merlinz01:
@tiangolo  @svlandeg  @Kludex @alejsdev Can I please have some indication of what your plan is for this PR? I've been waiting for five months and I still need this.

merlinz01:
@tiangolo @svlandeg @Kludex @alejsdev Half a year now and I still haven't heard a peep. :raised_eyebrow:

Please tell me if there's anything I need to do to get this moving along.

svlandeg:
Hi @merlinz01, thanks for your contribution, and apologies for the delay in response.

We have quite a bit of backlog of PRs to review (and many other maintenance tasks to do), and haven't been able to review this one in detail yet. Unfortunately it's also really difficult to provide an ETA for that. But this PR is in our queue and we'll write here when we've been able to review it in detail. In the meantime, please refrain from pinging a list of maintainers, as this makes our GH notifications very difficult to use and actually costs us more time that could be spent on maintenance tasks. If you'd like to speed up the process overall, feel free to [help](https://fastapi.tiangolo.com/help-fastapi/#help-others-with-questions-in-github) with answering questions on the discussion forum, reviewing other PRs, etc. Thanks 🙏

9cb14c1ec0 (approved):
Looks good to me.  The tests are still passing.  Would like to get this merged as I need this fix also.

merlinz01:
I did a rebase on top of the latest master changes to clean out all the auto-merge commits and a number of formatting-only commits. All tests are still passing.

tiangolo:
Hey there, from the code example in the description it seems the app expects a query parameter `organization_id` but you are sending a body with a field `organization_id`.

For query parameters (and headers, cookies), there's no concept of `None` or `NULL`, it's all strings, so there's no simple way to explicitly pass `None` there, only in body requests.

I have some refactors I need to make that will also affect this, I have to make that first, and that will cause a bunch of conflicts, so it will be better to implement anything necessary afterwards.

For your specific use case, it would be better to start with a GitHub Discussion.

I'll now close this one, thanks! :coffee:

merlinz01:
Body requests are specifically what this PR is meant to address, as shown in the example requests. I have updated the example code to show this clearly.

I continue to need this feature, so it would be great to know when the refactors are complete.

YuriiMotov:
@merlinz01, I think this use case is covered by this set of tests: https://github.com/fastapi/fastapi/pull/14829 (see [‎tests/test_request_params/test_body/test_nullable_and_defaults.py](https://github.com/fastapi/fastapi/pull/14829/changes#diff-219099351b973dded13078b19166c43a1742b207bc4210e1a929f2c36d3ff70e) starting from line 756).

Could you please check it?

I hope that set of tests will be later used to fix this issue and all other similar issues related to request parameters

merlinz01:
That looks great, but I don't see any tests for where a request body parameter has a non-null default value, as in my example code.

YuriiMotov:
> That looks great, but I don't see any tests for where a request body parameter has a non-null default value, as in my example code.

I think this is it:

([‎tests/test_request_params/test_body/test_nullable_and_defaults.py](https://github.com/fastapi/fastapi/pull/14829/changes#diff-219099351b973dded13078b19166c43a1742b207bc4210e1a929f2c36d3ff70e) lines 1004-1038)
```py
@pytest.mark.parametrize(
    "path",
    [
        pytest.param(
            "/nullable-with-non-null-default",
            marks=pytest.mark.xfail(
                reason="Null values are treated as missing for non-model Body parameters"
            ),
        ),
        "/model-nullable-with-non-null-default",
    ],
)
def test_nullable_with_non_null_default_pass_null(path: str):
    client = TestClient(app)

    with patch(f"{__name__}.convert", Mock(wraps=convert)) as mock_convert:
        response = client.post(
            path,
            json={
                "int_val": None,
                "str_val": None,
                "list_val": None,
            },
        )

    assert mock_convert.call_count == 3, "Validator should be called for each field"
    assert response.json() == {
        "int_val": None,
        "str_val": None,
        "list_val": None,
        "fields_set": IsOneOf(
            None, IsList("int_val", "str_val", "list_val", check_order=False)
        ),
    }
```

merlinz01:
Oh yes, I missed that. But model vs non-model parameters should work the same, not xfail for non-model.

YuriiMotov:
> But model vs non-model parameters should work the same, not xfail for non-model.

Of course! it's added because this test fails on master. The goal will be to make it pass and remove the `xfail`

merlinz01:
Great! I'll be watching that PR then.

PR #12140, opened 2024-09-06
🌐 Fix Korean translation for ```docs/ko/docs/*.md```

sodaMelon:
I corrected minor Markdown syntax errors and fixed bold formatting as per the original text. 

also modified some typo. 

(most problem appeared by markdown, because of korean postposition grammer. anyway maybe it will be better)


sodaMelon:
i did mistake at branch 😱 will split

PR #12194, opened 2024-09-13
🐛 Fix form field regression with `alias`

tonnico:
Fix for https://github.com/fastapi/fastapi/discussions/12139

With 0.113.0 an alias in a Form stopped working.

This is the suggestion from https://github.com/fastapi/fastapi/discussions/12139#discussioncomment-10566577 with a test.

Thanks to @sinisaos for the suggestion.


tiangolo:
Great, good catch, thank you! :cake: Thanks for putting the effort into writing proper tests. :bow: 

This will be available in FastAPI 0.114.2 released in the next few hours. :rocket:

Issue #12133, opened 2024-09-05
Your upgrade of 113 broken all gradio apps - a fix needed ASAP this is devastating 


FurkanGozukara:
a very basic just gradio having app

fast api 111 fixes issue

all of my followers now reporting errors and so hard to fix all

i don't know how can you publish such a devastating bug having version please fix it ASAP

```
2024-09-06 00:12:20,515 - INFO - HTTP Request: GET https://api.gradio.app/gradio-messaging/en "HTTP/1.1 200 OK"
Running on local URL:  http://127.0.0.1:7860
2024-09-06 00:12:20,542 - INFO - HTTP Request: GET http://127.0.0.1:7860/startup-events "HTTP/1.1 200 OK"
2024-09-06 00:12:20,552 - INFO - HTTP Request: HEAD http://127.0.0.1:7860/ "HTTP/1.1 200 OK"

To create a public link, set `share=True` in `launch()`.
2024-09-06 00:12:20,673 - INFO - HTTP Request: GET https://checkip.amazonaws.com/ "HTTP/1.1 200 "
2024-09-06 00:12:21,337 - INFO - HTTP Request: GET https://api.gradio.app/pkg-version "HTTP/1.1 200 OK"
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "R:\Image_Caption_Editor_v6\venv\lib\site-packages\pydantic\type_adapter.py", line 209, in __init__
    core_schema = _getattr_no_parents(type, '__pydantic_core_schema__')
  File "R:\Image_Caption_Editor_v6\venv\lib\site-packages\pydantic\type_adapter.py", line 98, in _getattr_no_parents
    raise AttributeError(attribute)
AttributeError: __pydantic_core_schema__

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "R:\Image_Caption_Editor_v6\venv\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 406, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "R:\Image_Caption_Editor_v6\venv\lib\site-packages\uvicorn\middleware\proxy_headers.py", line 70, in __call__
    return await self.app(scope, receive, send)
  File "R:\Image_Caption_Editor_v6\venv\lib\site-packages\fastapi\applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
[...]

PR #10469, opened 2023-10-18
📝 Update docs, remove references to removed `pydantic.Required` in `docs/en/docs/tutorial/query-params-str-validations.md`

yogabonito:
Pydantic v2 [does not contain Required anymore](https://docs.pydantic.dev/latest/migration/#removed-in-pydantic-v2), so there is no need to mention it in FastAPI's docs.

PR #12197, opened 2024-09-13
🐛 Remove `Required` shadowing from fastapi using Pydantic v2

pachewise:
Re:
* https://github.com/vllm-project/vllm/issues/8212
* https://github.com/vllm-project/vllm/issues/8450
* https://github.com/fastapi/fastapi/discussions/11952

(and potentially others)

I think I found the underlying cause. We're still supporting `pydantic.Required` in FastAPI, which has the possibility to shadow the new `Required` type. To me, this seems like a Pydantic + fastapi bug. Here's how 0.113.0 introduced it:

1. `fastapi` has the new `get_model_fields()` function, which deals with `TypeAdapter()`, and is called for every field
2. TypeAdapter gets a `globalns` from the `sys._getframe()`, meaning it gets the ns from `_compat`. This is important because pydantic bases their type evaluations from that `globalns`. Why they are doing this is a question for later.
3. Because `Required = PydanticUndefined` is in _compat.py, we are getting that value set in the `globalns`
4. Independently, `openai` uses `TypedDict` with the `Required` annotations in its fields.
4. `vllm` (and potentially other libs) uses `fastapi` + `pydantic` + `openai` TypedDict classes

The smoking gun was: `TypeError: 'pydantic_core._pydantic_core.PydanticUndefinedType' object is not subscriptable`

Focusing on https://github.com/vllm-project/vllm/issues/8450, the offending type annotation was `Required[Union[str, Iterable[X]]`. It needed to have been `Required` that was a `PydanticUndefinedType`, otherwise it would be complaining about `Union[str, Iterable[X]]`, or some other annotation. I ran the snippet [here](https://github.com/fastapi/fastapi/issues/12133#issuecomment-2342858880) using `fastapi 0.114.1, pydantic 2.9.0, vllm 0.6.1, python = 3.8.20`, and was able to reproduce the issue again. Using `pdb.pm()`, I was able to inspect the globalns and connect the dots.

# How to test

Run the code example in [this](https://github.com/fastapi/fastapi/issues/12133#issuecomment-2342858880) comment Python 3.8. It will fail without the changes, and pass with the changes.

### DIFF
--- a/docs_src/query_params_str_validations/tutorial006d_an_py39.py
+++ b/docs_src/query_params_str_validations/tutorial006d_an_py39.py
@@ -1,13 +1,12 @@
 from typing import Annotated
 
 from fastapi import FastAPI, Query
-from pydantic import Required
 
 app = FastAPI()
 
 
 @app.get("/items/")
-async def read_items(q: Annotated[str, Query(min_length=3)] = Required):
+async def read_items(q: Annotated[str, Query(min_length=3)] = ...):
     results = {"items": [{"item_id": "Foo"}, {"item_id": "Bar"}]}
     if q:
         results.update({"q": q})
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -312,6 +312,7 @@ else:
     from pydantic.fields import (  # type: ignore[no-redef,attr-defined]
         ModelField as ModelField,  # noqa: F401
     )
+
     # from pydantic.fields import (  # type: ignore[no-redef,attr-defined]
     #     Required as Required,  # noqa: F401
     # )
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -71,7 +71,8 @@ if PYDANTIC_V2:
             general_plain_validator_function as with_info_plain_validator_function,  # noqa: F401
         )
 
-    RequiredParam = Undefined = PydanticUndefined
+    RequiredParam = PydanticUndefined
+    Undefined = PydanticUndefined
     UndefinedType = PydanticUndefinedType
     evaluate_forwardref = eval_type_lenient
     Validator = Any
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -71,7 +71,7 @@ if PYDANTIC_V2:
             general_plain_validator_function as with_info_plain_validator_function,  # noqa: F401
         )
 
-    Undefined = PydanticUndefined
+    RequiredParam = Undefined = PydanticUndefined
     UndefinedType = PydanticUndefinedType
     evaluate_forwardref = eval_type_lenient
     Validator = Any
@@ -316,7 +316,7 @@ else:
     # from pydantic.fields import (  # type: ignore[no-redef,attr-defined]
     #     Required as Required,  # noqa: F401
     # )
-    Required: Any = Ellipsis
+    RequiredParam: Any = Ellipsis
     from pydantic.fields import (  # type: ignore[no-redef,attr-defined]
         Undefined as Undefined,
     )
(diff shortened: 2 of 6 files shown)


fastapi/_compat.py
```diff
@@ -313,9 +312,11 @@ def get_model_fields(model: Type[BaseModel]) -> List[ModelField]:
...
         ModelField as ModelField,  # noqa: F401
     )
-    from pydantic.fields import (  # type: ignore[no-redef,attr-defined]
-        Required as Required,  # noqa: F401
-    )
+
+    # from pydantic.fields import (  # type: ignore[no-redef,attr-defined]
+    #     Required as Required,  # noqa: F401
+    # )
+    Required = Ellipsis
```
pachewise, line 319: not sure if we'd like to keep this or completely remove it

svlandeg:
Thanks for the PR and the detailed description and links! It looks like you've done some great digging 🙏 
We'll get this reviewed by the team in more detail and will update here once we do.

docs_src/query_params_str_validations/tutorial006d_an_py39.py
```diff
@@ -1,13 +1,12 @@
...
 
 from fastapi import FastAPI, Query
-from pydantic import Required
 
 app = FastAPI()
 
 
 @app.get("/items/")
-async def read_items(q: Annotated[str, Query(min_length=3)] = Required):
+async def read_items(q: Annotated[str, Query(min_length=3)] = ...):
```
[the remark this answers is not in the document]
svlandeg: This has been deleted from the (English) docs in PR #10469, but the Russian documentation does in fact still reference it in `docs/ru/tutorial/query-params-str-validations.md`. Ideally, we probably should remove the reference from the Russian docs as well, and then remove this file altogether, but I would suggest to do that in a separate PR to avoid mixing up different issues - it's always easier to review atomic PRs.

fastapi/_compat.py
```diff
@@ -71,8 +71,7 @@
             general_plain_validator_function as with_info_plain_validator_function,  # noqa: F401
         )
 
-    Required = PydanticUndefined
-    Undefined = PydanticUndefined
+    RequiredParam = Undefined = PydanticUndefined
```
svlandeg, line 74:
```suggestion
    RequiredParam = PydanticUndefined
    Undefined = PydanticUndefined
```
This might be a matter of personal preference, but I would use binary assignments instead of tertiary ones. I also think it makes maintenance of code easier on the long run, as you can change one of the assignments and only that particular line will show up in the diff, rather than doing a mix of things on one line.

fastapi/_compat.py
```diff
@@ -313,9 +312,11 @@ def get_model_fields(model: Type[BaseModel]) -> List[ModelField]:
     from pydantic.fields import (  # type: ignore[no-redef,attr-defined]
         ModelField as ModelField,  # noqa: F401
     )
-    from pydantic.fields import (  # type: ignore[no-redef,attr-defined]
-        Required as Required,  # noqa: F401
-    )
+
+    # from pydantic.fields import (  # type: ignore[no-redef,attr-defined]
+    #     Required as Required,  # noqa: F401
+    # )
```
svlandeg, line 318:
```suggestion
```

svlandeg (approved):
Thanks again for the digging! The proposed fix here makes sense to me and avoids further confusion with `Required`. I only had a few minor nitpicking comments. I'll ask Tiangolo to review as well 🙏

MoritzLaurer:
Will this be merged soon? I'm encountering a [related issue with vLLM](https://github.com/vllm-project/vllm/issues/8450)

tiangolo:
Great job with all the detective work @pachewise! And thanks for the very clear explanation and references.

Thanks a lot @svlandeg for all your help and review! :bow: 

This will be available in the next hours in FastAPI `0.115.1` :rocket:

PR #12199, opened 2024-09-13
✨ Add support for Pydantic models for parameters using `Query`, `Cookie`, `Header`

tiangolo:
✨ Add support for Pydantic models for parameters using `Query`, `Cookie`, `Header`


docs/en/docs/tutorial/cookie-param-models.md
```diff
@@ -0,0 +1,154 @@
...
+
+/// note
+
+This is supported since FastAPI version `0.115.0`. 🤓
+
+///
+
+/// tip
+
+This same technique applies to `Query`, `Cookie`, and `Heder`. 😎
```
Wurstnase, line 15:
```suggestion
This same technique applies to `Query`, `Cookie`, and `Header`. 😎
```

tiangolo: Good catch, thanks!

docs_src/query_param_models/tutorial001.py
```diff
@@ -0,0 +1,19 @@
...
+from typing_extensions import Literal
+
+app = FastAPI()
+
+
+class FilterParams(BaseModel):
+    limit: int = Field(100, gt=0, le=100)
+    offset: int = Field(0, ge=0)
+    order_by: Literal["created_at", "updated_at"] = "created_at"
+    tags: List[str] = []
```
Wurstnase, line 14: It's generally not a good idea to use mutables as defaults.

```suggestion
    tags: List[str] = Field(default_factory=list)
```

docs_src/query_param_models/tutorial001.py
```diff
@@ -0,0 +1,19 @@
...
+from pydantic import BaseModel, Field
+from typing_extensions import Literal
+
+app = FastAPI()
+
+
+class FilterParams(BaseModel):
+    limit: int = Field(100, gt=0, le=100)
+    offset: int = Field(0, ge=0)
+    order_by: Literal["created_at", "updated_at"] = "created_at"
```
Wurstnase, line 13: Should we probably add an alias for this tests?

```suggestion
    order_by: Literal["created_at", "updated_at"] = Field(alias="orderBy", default="created_at")
```

What is about `Header` and `Cookie`?

tonnico (commented):
I'm looking forward for this PR. This will makes some parts much more simple.

Thanks in advance 🤗

docs_src/query_param_models/tutorial001.py
tiangolo: Thanks! Pydantic does a deep copy of default mutables, so in this case it's not a problem.

docs_src/query_param_models/tutorial001.py
tiangolo: Why do you think so? :thinking: 

Maybe you can suggest it in a PR on top after this. :nerd_face:

docs_src/query_param_models/tutorial001.py
Wurstnase: Oh, good to know 💡

docs_src/query_param_models/tutorial001.py
Wurstnase: After some issues between 0.112.4 and 0.114.1 with the alias I thought it could be a good idea.
I will make a suggestion.

seabean1999:
Hi，it seems the Nested Pydantic Models don't work for parameters using Query. Is this the expected design?

oogetyboogety:
@tiangolo  is there a way we can work together on extending support for
a) https://fastapi.tiangolo.com/tutorial/header-params/#automatic-conversion for CommonModel, docs show the underscores still remain underscores:
image01
I realized you were saying that the usability of this functionality in Depends works, but coincidentally works. we have noticed some odd behavior (input parameter because a json object) when trying to use Depends and Header model together. It works as expected by only including "repeated" header parameter in Depends. I would still like for automatic conversion to happen on these Pydantic parameters models though, even if included via Depends.

ollie-bell:
Is it possible to mix "vanilla" query parameters with the new model parameters?

e.g. the [example from the docs](https://fastapi.tiangolo.com/tutorial/query-param-models/#query-parameters-with-a-pydantic-model) no longer works with an additional `other_param: str = "foo"` 

```python
from typing import Annotated, Literal

app = FastAPI()


class FilterParams(BaseModel):
    limit: int = Field(100, gt=0, le=100)
    offset: int = Field(0, ge=0)
    order_by: Literal["created_at", "updated_at"] = "created_at"
    tags: list[str] = []


@app.get("/items/")
async def read_items(filter_query: Annotated[FilterParams, Query()], other_param: str = "foo"):
    return filter_query
```

PR #12241, opened 2024-09-21
♻️ Refactor request handler to make it extensible for other libraries

zmievsa:
FastAPI is rather hard to extend by frameworks such as [cadwyn](https://docs.cadwyn.dev) because an important part of FastAPI's logic is defined using functions within functions. So if I want to reuse any of that logic in Cadwyn -- I have to copy it manually, making it harder to keep compatibility with FastAPI.

This PR fixes the issue without changing any of the underlying interfaces or behaviors.
Labels are `refactor` and probably `internal` but I have no permissions to add them.


zmievsa:
I now understand that https://fastapi.tiangolo.com/how-to/custom-request-and-route/#create-a-custom-gziprequest-class is the perfect solution to my problem so this PR is no longer necessary

PR #12264, opened 2024-09-25
📝 Update link to Swagger UI configuration docs

makisukurisu:
I've noticed that the [old link](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration) now leads to the "404 Not found" page on swagger.io, found the new link and updated it wherever possible (there appear to be only 4 languages that even have `configure-swagger-ui.md` page).

Even though it appears to be a problem that should be resolved on swagger.io side, I think that updating the link may be better.

P.S. I've created a [bug report](https://github.com/swagger-api/swagger.io-docs/issues/362) in the swagger.io-docs repo.
P.S.S. I still believe it's best to have the trailing slashes at the end of the link, because without them, users are getting 308 Permanent Redirect to other webpage, so adding a / will make life easier for both parties.


Ryandaydev (commented):
No objection to the changes if it adds to consistency. However, I don't receive an error when using the original link. It seems to work fine for me.

makisukurisu:
@raccoonyy, indeed, as noted in the linked [issue](https://github.com/swagger-api/swagger.io-docs/issues/362), they (swagger.io team) have resolved the redirection issue.

Thanks for your review, I'm absolutely OK if you decide to close the PR and decline the suggested changes as non-necessary.

raccoonyy:
@makisukurisu I think I was mentioned in the wrong place. :)

pjsully (approved):
I would suggest this change should be merged.  If it works as well as without the trailing /, and also works if redirection for some reason isn't working, it seems like a better solution.

alejsdev:
Thank you @makisukurisu :sparkles:

PR #12281, opened 2024-09-28
👷 Use uv in CI

tiangolo:
👷 Use uv in CI


Kludex:
I had to install the python version. I don't remember if it was only because of python 3.13...

https://github.com/Kludex/python-multipart/blob/f92851f1fd4f32350063bc69f8b21d9da681da6e/.github/workflows/main.yml#L27-L28

PR #12369, opened 2024-10-01
Update docker.md wording

dlrice:


pjsully (approved):

tiangolo:
Thanks for the interest @dlrice, I think it's fine as it is right now and conveys the meaning I intended, I don't see a grammatical error, I suspect it's a change only in taste (not sure), so I'll pass on this one, but thanks for the interest! :coffee:

dlrice:
Hi @tiangolo!  No worries. I do however believe it is a grammatical error and perhaps a more straightforward fix I could have suggested would be just to swap "is" and "that":

Let me tell you why **is that**.
→
Let me tell you why **that is**.

PR #11691, opened 2024-06-06
 ✨ Add custom GenerateJsonSchema

maurerle:
Supersedes #9873 which removes the generation of a AnyOf type None when using `| None` in method signatures, introduced through pydantic v2.

Has tests fixed in contrast to #9873 and is rebased onto latest master

Closes #9709 

Closes discussions:
https://github.com/tiangolo/fastapi/discussions/10836
https://github.com/tiangolo/fastapi/discussions/10481

PR #12375, opened 2024-10-02
 ✨ Add custom GenerateJsonSchema

maurerle:
Supersedes #9873 which removes the generation of a AnyOf type None when using `| None` in method signatures, introduced through pydantic v2.

Has tests fixed in contrast to #9873 and is rebased onto latest master

Closes #9709 
Supersedes #11691

Closes discussions:
https://github.com/tiangolo/fastapi/discussions/10836
https://github.com/tiangolo/fastapi/discussions/10481


michalalicki:
One question. I used the solution when replacing for example `Optional[str]` with `Annotated[Union[str, SkipJsonSchema[None]], Field(None, json_schema_extra=lambda x: x.pop("default"))]` however I encounter the issue when I wanted to add value constrain for example max_length. If I was passing explicit None value for example having response json like:
```
{
   "foo": "bar",
   "constrained_optional": None
}
```
then Pydantic validation raises TypeError because cannot apply constrain on None value. Could you please double check your solution if this kind of error will not appear?

jordantshaw:
Any updates on this?

maurerle:
@michalalicki sorry for not getting back on this.
So you experience this error only with this branch and not with the current master branch of fastapi?
Can you provide a MWE of it, I did not fully understand what you were trying to do.

@jordantshaw I rebased this onto latest master - though I am currently not actively using this anymore.

Anyway, an MWE of the feature is basically:

```
from fastapi import FastAPI

app = FastAPI()


@app.get("/")
def root(
    q: int | None = Query(None),
):
    return q
```

Maybe this helps to create an MWE for @michalalicki ?

jordantshaw:
@tiangolo is this something we can potentially get merged and released?

fastapi/_compat.py
```diff
@@ -72,6 +81,75 @@
...
+                    key.field_info, Param
+                )
+                self.generate_inner(schema)
+
+            definitions_remapping = self._build_definitions_remapping()
+
+            json_schemas_map: Dict[Tuple[JsonSchemaKeyT, JsonSchemaMode], DefsRef] = {}
+            for key, mode, schema in inputs:
+                self._mode = mode
+                self.skip_null_schema = isinstance(key, ModelField) and isinstance(
```
jiohyoo, line 140: @maurerle I tried out this version, and I see this custom schema is only effective when the field is a parameter in FastAPI endpoints. Can we extend this to fields of pydantic models as well?
When I modify the condition to `isinstance(key, ModelField) or isinstance(key.field_info, Param)`, it seems working for both cases

maurerle: should be working now

maurerle: Nope, this does not work out with various test cases.
Can you give an MWE of what you think is still wrong?

YuriiMotov (commented):
This solution seems to be quite fragile since it uses non-public Pydantic function `_sort_json_schema`.

As I understand it, the main intention of this PR is to solve the problem with optional parameters being incorrectly shown in Swagger.

Things have been changed a little, and now

```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/items")
async def get_items(item_id: int | None = None):
    pass
```

is now shown as

image

That's cryptic (as I understand it, this is an issue on Swagger UI side) and can be improved, but IMO it's not invalid.

Considering that implementing solid solution seems to be non-trivial task, I suggest we document current situation and workarounds:

**Non-required parameter**
```python
@app.get("/items")
async def get_item(item_id: int | SkipJsonSchema[None] = None):
    return {"item_id": item_id}
```
image


**Nullable required parameter**
TODO
Does anybody know how the schema should look like to make it possible to send empty value?

maurerle:
@YuriiMotov - yes

```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/items")
async def get_items(item_id: int | None = None):
    pass
```

does show types better, but does not validate the int in the browser (e.g. you can type in "hello", request to the api, which then responds with 422 can't process).
So this is basically the underlaying root issue here.

Your other example

```python

@app.get("/items")
async def get_item(item_id: int | SkipJsonSchema[None] = None):
    return {"item_id": item_id}
```

works as expected. Documenting this usage would be a solution instead of this PR.

>  Does anybody know how the schema should look like to make it possible to send empty value?

Without the `| None` the item_id is considered required:

image

Otherwise it can be left empty, in which case null is transmitted.

tiangolo:
Let's start with a GitHub Discussion with your specific use case, with a minimal reproducible example, to evaluate what you need in the specific use case and then take it from there.

Thank you! 🍰

PR #12395, opened 2024-10-07
🌐 Fix path code file in docs

JavierSanchezCastro:
From https://github.com/fastapi/fastapi/discussions/12393


JavierSanchezCastro:
Already fixed in last push.

Before:
image

Now:
image

I assume including the new "markdown-include-variants" fixed that.

Closing

PR #12406, opened 2024-10-08
🐛 Fix duplicate special dependency handling

volfpeter:
See the bug description [here](https://github.com/fastapi/fastapi/discussions/12403)


volfpeter:
Hi guys! Any chance of a review? This is a very trivial change, although with a very tiny performance impact (could be removed by complicating the code a bit). It's also okay if you decide to close/reject the PR because you don't want to support this use-case and it's not hard to work around. Thanks.

svlandeg:
Hi @volfpeter, thanks for the PR! Yes - we will get this reviewed, but we're currently facing quite a big backlog of PRs. We're currently trying to manage & prioritize everything. I'll put this on the team's queue for review, thanks for your patience! 🙏

volfpeter:
Thanks @svlandeg for the update and for your effort.

volfpeter:
Hi again! It seems not much is happening in the project these days (months). Is it reasonable to wait for a review for this bug fix? It's been 8 months now. If not, feel free to close the PR. Thanks!

fastapi/dependencies/models.py
```diff
@@ -22,12 +22,12 @@ class Dependant:
     security_requirements: List[SecurityRequirement] = field(default_factory=list)
     name: Optional[str] = None
     call: Optional[Callable[..., Any]] = None
-    request_param_name: Optional[str] = None
-    websocket_param_name: Optional[str] = None
-    http_connection_param_name: Optional[str] = None
-    response_param_name: Optional[str] = None
-    background_tasks_param_name: Optional[str] = None
-    security_scopes_param_name: Optional[str] = None
+    request_param_names: List[str] = field(default_factory=list)
```
YuriiMotov: Is there a reason not to use `set` here?
Seems like `set` should work a bit faster

YuriiMotov (commented):
@volfpeter, thank you for working on this!

LGTM in general, I think this fix is quite useful (error might be quite confusing for those who don't know about it).
~~I would only change `List` to `Set` if there is no reason not to do so.~~ (upd: done)
But let's first wait for Sebastián to validate the idea

fastapi/dependencies/models.py
volfpeter: No, you're totally right. I'll update it tomorrow.

volfpeter: Done. Thanks for your review! I really should have thought about it...

YuriiMotov: Thank you!

YuriiMotov (approved):
LGTM

I think this fix is quite useful PR - error might be quite confusing for those who don't know about it.
Example use case: custom route decorator that accesses `Response` before\after executing the endpoint. For now it's hard to implement such decorator as we should be sure that we use the same name for special dependency parameter as the endpoint uses.

<details open><summary>Code example</summary>

```py
import inspect
from functools import wraps
from typing import Callable, Concatenate, ParamSpec, TypeVar

P = ParamSpec("P")
R = TypeVar("R", bound=str)


def add_id_header(func: Callable[P, R]) -> Callable[Concatenate[Response, P], R]:
    @wraps(func)
    def wrapper(response: Response, *args: P.args, **kwargs: P.kwargs) -> R:
        name_id = func(*args, **kwargs)
        headers = {"etag": name_id}
        response.headers.update(headers)
        return name_id

    # Modify wrapper's signature to add `response: Response`
    sig = inspect.signature(func)
    if "response" not in sig.parameters:
        new_params = list(sig.parameters.values()) + [
            inspect.Parameter(
                name="response",
                kind=inspect.Parameter.KEYWORD_ONLY,
                annotation=Response,
            )
        ]
        wrapper.__signature__ = sig.replace(parameters=new_params)  # type: ignore attr-defined
        return wrapper


app = FastAPI()


@app.get("/")
@add_id_header
def get_index(r: Response):  # Removing `Response` parameter or renaming it to `response` will solve this
    return "123"

```

</details>

volfpeter:
Thanks @svlandeg for merging master!

fastapi/dependencies/utils.py
```diff
@@ -158,12 +158,12 @@ def get_flat_dependant(
...
-        http_connection_param_name=dependant.http_connection_param_name,
-        response_param_name=dependant.response_param_name,
-        background_tasks_param_name=dependant.background_tasks_param_name,
-        security_scopes_param_name=dependant.security_scopes_param_name,
+        request_param_names=dependant.request_param_names.copy(),
+        websocket_param_names=dependant.websocket_param_names.copy(),
+        http_connection_param_names=dependant.http_connection_param_names.copy(),
+        response_param_names=dependant.response_param_names.copy(),
+        background_tasks_param_names=dependant.background_tasks_param_names.copy(),
+        security_scopes_param_names=dependant.security_scopes_param_names.copy(),
```
YuriiMotov, line 165: Not sure about this part.
This way for flat dependant we have:
* `path_params` (and other params) will include all path params of sub-dependants
*  `request_param_names` (and other "param names") will only include param names of this dependant (will not include param names from sub-dependants)

This doesn't affect us currently, but maybe we should think how to prevent confusion in the future

YuriiMotov:
We need to add tests with sub-dependencies

volfpeter:
I forgot to answer this:

> For now it's hard to implement such decorator as we should be sure that we use the same name for special dependency parameter as the endpoint uses.

It's not that hard, all you need to do is create a dependency that returns the request, and use a runtime alias for the return value. Here's an example: https://github.com/volfpeter/fasthx/blob/main/fasthx/dependencies.py

Nevertheless, the issue is really not obvious and most users will not know how to work around it.

volfpeter:
The pre-commit check failed because the `setup-uv` ended in a timeout (other PRs have the same issue). Could you trigger a re-run, or should I do push an empty commit?

volfpeter:
Hi @YuriiMotov What's the status of this PR? It's been open for 16 months now, it would be really nice if it could be merged (or closed).

YuriiMotov:
> Hi @YuriiMotov What's the status of this PR? It's been open for 16 months now, it would be really nice if it could be merged (or closed).

This PR is awaiting for the final review by Sebastian

volfpeter:
Hi guys, sorry for pushing, but would it be possible to merge this PR? It's been waiting here for almost 20 months. It's been approved 10 months ago.

Issue #12402, opened 2024-10-08
[BUG] The pydantic model that has declared an alias cannot correctly receive query parameters


insistence:
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

### Issue Content

Thank you for all the work you have done. I have initiated the [discussion ](https://github.com/fastapi/fastapi/discussions/12401)as requested, but I think this issue is quite important. Initiating this issue is just to prevent the discussion from being drowned out, and I apologize for any offense.

### Example Code

```python
import uvicorn
from typing import Literal

from fastapi import FastAPI, Query
from pydantic import BaseModel, ConfigDict, Field
from pydantic.alias_generators import to_camel

app = FastAPI()


class FilterParams(BaseModel):
    model_config = ConfigDict(alias_generator=to_camel)

    limit: int = Field(100, gt=0, le=100)
    offset: int = Field(0, ge=0)
    order_by: Literal['created_at', 'updated_at'] = 'created_at'
    tags: list[str] = []


@app.get('/items/')
async def read_items(filter_query: FilterParams = Query()):
    return filter_query


if __name__ == '__main__':
    uvicorn.run(app='app:app')
```

### Description

Running the code in the example above, I encountered an incorrect result when accessing http://127.0.0.1:8000/items/?offset=1&orderBy=updated_at in the browser, orderBy did not receive successfully.
```
{
    "limit": 100,
    "offset": 1,
    "orderBy": "created_at",
    "tags": []
}
```
The correct result should be as follows
```
{
    "limit": 100,
    "offset": 1,
    "orderBy": "updated_at",
    "tags": []
}
```

### Operating System

Windows

### FastAPI Version

0.115.0

### Pydantic Version

2.9.2

### Python Version

3.9.19

aless10:
Did you try with the `Annotated` version as in the docs?
```
from fastapi import FastAPI, Query
from pydantic import BaseModel, Field
from typing_extensions import Annotated, Literal

app = FastAPI()


class FilterParams(BaseModel):
    limit: int = Field(100, gt=0, le=100)
    offset: int = Field(0, ge=0)
    order_by: Literal["created_at", "updated_at"] = "created_at"
    tags: list[str] = []


@app.get("/items/")
async def read_items(filter_query: Annotated[FilterParams, Query()]):
    return filter_query
```

Usernamelesss:
Hello everyone! Can you confirm that it is not possible to have other query params along with the pydantic model ?
[...]

PR #12411, opened 2024-10-08
🐛 Fix alias bug in query parameter models

sinisaos:
Related to #12401 and #12402


Alxe:
@alejsdev Is there any apparent reason why this was not merged?

I don't know if this fix was ill-fitting, but this issue is affecting my use case and I wanted to take a jab at it.
If this is the way, @sinisaos could perhaps reopen the PR and have it reviewed.

PR #12421, opened 2024-10-10
✨ Add relative docs everywhere

gshmu:
In some situations, you might need to use a proxy server like Traefik or Nginx with a configuration that adds an extra path prefix that is not seen by your application.

The [Behind a Proxy](https://fastapi.tiangolo.com/advanced/behind-a-proxy/)  handle these specific cases.

However, the relative path I obtained could be a more optimal solution.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -27,7 +27,7 @@ from fastapi.openapi.docs import (
     get_swagger_ui_html,
     get_swagger_ui_oauth2_redirect_html,
 )
-from fastapi.openapi.utils import get_openapi
+from fastapi.openapi.utils import calculate_relative_url, get_openapi
 from fastapi.params import Depends
 from fastapi.types import DecoratedCallable, IncEx
 from fastapi.utils import generate_unique_id
@@ -679,6 +679,7 @@ class FastAPI(Starlette):
                 """
             ),
         ] = True,
+        relative_docs: bool = False,
         responses: Annotated[
             Optional[Dict[Union[int, str], Dict[str, Any]]],
             Doc(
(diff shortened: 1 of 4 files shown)


gshmu:
@tiangolo please add some label to this pr.

IvDmNe:
+1, I also find this feature useful.
@gshmu Can you explain please, what do you mean here?

> The [Behind a Proxy](https://fastapi.tiangolo.com/advanced/behind-a-proxy/) handle these specific cases.

On this page it is stated that this preifx should be explicitly passed as ```root_path``` parameter in FastAPI(), so It must know the prefix.

YuriiMotov:
@gshmu, thanks for your interest!

Is the intention to avoid being have to specify `root_path`?

`relative_docs=True` will only solve the problem for Docs, but `request.url_for()` will not work correctly, right?
I guess in some cases there might be problems with mounted apps and static files.

I think we should follow the ASGI spec here and specify `root_path`. It could be specified either as an argument to `FastAPI()` or as an argument to Uvicorn (`uvicorn --root-path /api`).

gshmu:
@YuriiMotov  thanks.

Using relative paths for static assets as well makes it very convenient and flexible to compose multiple services with Nginx.

gshmu:
```
app = FastAPI(root_path="/mysub")

sub_app = FastAPI()  # sub app for external docs: manageable and isolated
app.mount("/sub-app", sub_app)
```

YuriiMotov:
@gshmu, sorry, but I didn't get it..
Could you please elaborate on why you think the proposed way is more convenient? For me it will bring additional cognitive load for people (they will need to decide which way to choose).
Also, please address my comments: https://github.com/fastapi/fastapi/pull/12421#issuecomment-3068893447

As far as I understand, Sebastian tends not to add new features if there is already way to achieve this. Here we already have `root_path` that effectively solves this.
So, you should provide strong arguments. Or, let's close this

PR #12431, opened 2024-10-12
⬆️ Upgrade Starlette to `>=0.37.2,<0.41.0`

tiangolo:
⬆️ Upgrade Starlette to `>=0.37.2,<0.41.0`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
-    "starlette>=0.37.2,<0.39.0",
+    "starlette>=0.37.2,<0.41.0",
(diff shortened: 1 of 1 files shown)


vfazio:
I don't see Starlette 0.40 released but it's now included in the supported range? Do they guarantee compatibility for a 2 version range? or was this supposed to be <0.40.0? (pre-coffee comment, so i could be missing something)

musicinmybrain:
> I don't see Starlette 0.40 released but it's now included in the supported range? Do they guarantee compatibility for a 2 version range? or was this supposed to be <0.40.0? (pre-coffee comment, so i could be missing something)

Now that [Starlette 0.40](https://github.com/encode/starlette/releases/tag/0.40.0) has been released, and it turned out to be a security fix ([GHSA-f96h-pmfr-66vw](https://github.com/encode/starlette/security/advisories/GHSA-f96h-pmfr-66vw), [CVE-2024-47874](https://nvd.nist.gov/vuln/detail/CVE-2024-47874)), this unusual “forward-compatibility” in FastAPI makes sense.

PR #12446, opened 2024-10-13
🌐 Just a peace message ... to remember

ghost:
An other way to fight for **peace** (and only for peace), on internet.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@
 pre-commit >=2.17.0,<4.0.0
 # For generating screenshots
 playwright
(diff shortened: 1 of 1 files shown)


peace-and-love-py:
@tiangolo What do you think of the initiative? I understand that it doesn't belong here. However, desperate situations call for desperate actions.

tiangolo:
Thanks for the interest, but yep, it doesn't belong here, it's an external dependency with unknown untrusted code, etc. So no, thanks, I'll pass on this. :coffee:

PR #12455, opened 2024-10-14
✨ Add schema generator class param to FastAPI

thomas-touhey:
This feature allows setting a custom JSON schema generator at app-level, in order to customize JSON schema from pydantic-core schema generation as described in the following pydantic guide:

https://docs.pydantic.dev/2.1/usage/json_schema/#customizing-the-json-schema-generation-process

:warning: This feature is only available when using pydantic V2. Setting the additional parameter while using pydantic V1 will not have any effect.

For more context, the schema `Literal["something"]` generates a schema with both `enum` and `const`, which can be traced back to [literal_schema in pydantic/json_schema.py](https://github.com/pydantic/pydantic/blob/18d39febf00cb07954e19ebbdbd8168ef20e3df1/pydantic/json_schema.py#L736) brought by pydantic [commit 18d39febf00cb07954e19ebbdbd8168ef20e3df1](https://github.com/pydantic/pydantic/commit/18d39febf00cb07954e19ebbdbd8168ef20e3df1). It was asked from my team to either present `const` or `enum`, but not both, as it raises errors in some schema validation processes, which we can't do unless we can override the JSON schema generator, as described in the guide above, hence this pull request.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -14,6 +14,7 @@ from typing import (
 )
 
 from fastapi import routing
+from fastapi._compat import GenerateJsonSchema
 from fastapi.datastructures import Default, DefaultPlaceholder
 from fastapi.exception_handlers import (
     http_exception_handler,
@@ -755,6 +756,17 @@ class FastAPI(Starlette):
                 """
             ),
         ] = True,
+        schema_generator_class: Annotated[
+            Type[GenerateJsonSchema],
+            Doc(
+                """
+                Schema generator to use for the OpenAPI specification.
+                You probably don't need it, but it's available.
+
+                This affects the generated OpenAPI (e.g. visible at `/docs`).
+                """
+            ),
+        ] = GenerateJsonSchema,
         swagger_ui_parameters: Annotated[
             Optional[Dict[str, Any]],
             Doc(
@@ -862,6 +874,7 @@ class FastAPI(Starlette):
         self.root_path_in_servers = root_path_in_servers
         self.docs_url = docs_url
         self.redoc_url = redoc_url
+        self.schema_generator_class = schema_generator_class
         self.swagger_ui_oauth2_redirect_url = swagger_ui_oauth2_redirect_url
         self.swagger_ui_init_oauth = swagger_ui_init_oauth
         self.swagger_ui_parameters = swagger_ui_parameters
@@ -1071,6 +1084,7 @@ class FastAPI(Starlette):
                 servers=self.servers,
                 separate_input_output_schemas=self.separate_input_output_schemas,
                 external_docs=self.openapi_external_docs,
+                schema_generator_class=self.schema_generator_class,
             )
         return self.openapi_schema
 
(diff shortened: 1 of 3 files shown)


thomas-touhey:
I did not take pydantic v1 into account apparently. Will look into this in more detail later today (French time).

**EDIT**: Nevermind, finished it in the end.

thomas-touhey:
Hi @tiangolo @alejsdev, any chance I could have review or merge on this PR please? :)

thomas-touhey:
Still up for some review on this PR! :)

thomas-touhey:
_(just rebasing on master to ensure the PR is still mergeable, still waiting for some review)_

YuriiMotov:
@thomas-touhey, thanks for your interest in FastAPI!

There were several similar PRs and they were closed due to weak reasoning.

You already provided a use case, but maybe we can elaborate on this and add more cases?
For example, I saw people wanted to change items sorting order in openapi (or avoid sorting them), and as far as I understand it can be achieved using this feature.

Let's bring all use-cases together and add them to PR's description!

Also, since it only works with Pydantic V2, I think we should mention this in the docstring and in the description of the PR.
I would even show a warning message if this parameter is specified with Pydantic V1

thomas-touhey:
Hi @YuriiMotov, sorry for the late reply, at this point I wasn't expecting any answer on this PR and I was quite busy for the last weeks. ^^

I don't really have more cases myself than the one provided, the reasoning was mostly "if pydantic fucks up some checks / compatibility from other tools also using OpenAPI specs (such as the one I mentioned), this could be a quick and accessible way to fix it without having to fork FastAPI altogether".

I'm not working at the company I was when this PR was useful, and the issue I was trying to resolve stayed stale: since the API gateway we were using could not validate the app's schema, for some time the team fixed the schema produced by FastAPI manually, then for quicker releases the team ended up abanding the API gateway for the frontend to hit the API directly. The situation was not very pretty, but it was either that or having to maintain a FastAPI fork with that PR merged, which would have taken too much work AFAIK; all things considered, not the best situation. But what else can you do when a PR that passes CI checks doesn't receive answers for8 months...

I haven't encountered the other use cases you have seen, so I prefer not to guarantee their legitimacy.

I'll add the Pydantic V2 warning in the PR's description. If you have remarks on the actual implementation, feel free to make them so I can fix them, otherwise I don't really see what else I can do on the matter :)

tiangolo:
Thanks for the interest and effort put into this!

I needed to support Pydantic v1 models as well, which was implemented in https://github.com/fastapi/fastapi/pull/14168, but the new JSON Schema Generator from Pydantic v2 is not compatible with v1, so I had to do it in a different way. This makes customizing the JSON Schema Generator only for Pydantic v2 not feasible yet. That might happen later when dropping support for Pydantic v1, but for now, it wouldn't be possible. And I needed to prioritize first backwards compatibility and providing a way for people to migrate to Pydantic v2 before adding extra features exclusive to Pydantic v2.

Anyway, given that, I'll now close this one, but thanks for the interest and discussion! :coffee:

thomas-touhey:
Thanks for the answer @tiangolo, at least it gets this PR out of limbo :)

PR #12464, opened 2024-10-15
cleared up misinformation

wisderfin:
While translating the documentation about virtual environments into Russian, I came across information that I couldn't believe. I double-checked the information and found no justification for the following statement.

> Every time you install a new package in that environment, activate the environment again.
> 
> This makes sure that if you use a terminal (CLI) program installed by that package, you use the one from your virtual environment and not any other that could be installed globally, probably with a different version than what you need.

 _I do not deny that I may have erred in my judgment, and if this is indeed the case please provide me with more information about this process to support the statement._

### DIFF
--- a/docs/en/docs/virtual-environments.md
+++ b/docs/en/docs/virtual-environments.md
@@ -162,14 +162,6 @@ $ source .venv/Scripts/activate
 
 ////
 
-/// tip
-
-Every time you install a **new package** in that environment, **activate** the environment again.
-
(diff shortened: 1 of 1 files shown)


fernandosmither:
> I double-checked the information and found no justification for the following statement.

How exactly did you double-check this?

It makes sense because command aliases are created usually at the beginning of a terminal session. If you add a CLI package for a virtual environment that uses certain command, if that command was already mapped to another executable via a previous global packagae, you'll most likely end up calling the old global one, not the new, until you restart.

image

You are one prompt away from a very easy and quick first-pass fact-check

wisderfin:
I apologize, apparently I misinterpreted the question when translating it from English.

wisderfin:
Issue resolved

PR #12490, opened 2024-10-18
🧑‍💻 Improve logging for custom exceptions in routing

avaika:
fastapi was not properly working in my environment due to readonly /tmp filesystem. The app kept returning the error `There was an error parsing the body`. Which was not helpful at all. Had `{e}` been included to the error message, I would've saved quite a bit of time. 

Most likely I'm not the only one with that problem. Hopefully it might be helpful for others.


avaika:
Can this be merged please?

fastapi/routing.py
```diff
@@ -283,7 +283,7 @@ async def app(request: Request) -> Response:
                 raise
             except Exception as e:
                 http_error = HTTPException(
-                    status_code=400, detail="There was an error parsing the body"
+                    status_code=400, detail=f"There was an error parsing the body: {e}"
```
svlandeg, line 286: Wouldn't it be nicer to just print `e.msg`?

YuriiMotov (commented):
This way you expose internal details to user (exception message may contain such details), that's not good..

@avaika, do you agree?

PR #12494, opened 2024-10-19
📝 Fix type of `username` in `/docs_src/security` to allow `None`

vittxr:
In the docs > [security section](https://fastapi.tiangolo.com/tutorial/security/oauth2-jwt/#hash-and-verify-the-passwords), there is a type issue 

`username: str = payload.get("sub")`

Since `.get()` may return None, the `username` variable should accommodate this possibility by using:

`username: str | None = payload.get("sub")`

image

I've fixed `/docs_src/security` snippets


svlandeg (approved):
Thanks for catching this! Yes, makes sense.

svlandeg:
As a quick note, it looks like the test suite is failing due to something outside of this PR. We're looking into it!

vittxr:
All tests passed. I think it's ok now

alejsdev:
Hi @vittxr, thanks for your interest in contributing. This was covered in https://github.com/fastapi/fastapi/pull/13378, so I'll close this one, but thanks for your effort. :bowing_woman:

PR #12495, opened 2024-10-20
📝 Fix broken link in docs

eltonjncorreia:
Hi, I saw a link that is no longer available online. So I moved to the Hug project repository.

 I am available to correct any errors that I may have caused. Thanks for the FastAPI


alejsdev:
Thank you! @eltonjncorreia :rocket:

PR #12504, opened 2024-10-21
🐛 Propagate Pydantic's model config

RB387:
Related discussion:
https://github.com/fastapi/fastapi/discussions/12503


RB387:
@alejsdev @tiangolo hey hey folks 👋 
Is there any chance that this PR can be merged any time soon? 
It's blocking us at Wolt unfortunately from upgrading FastAPI to the latest version 😞

tiangolo:
I haven't checked this thoroughly but I suspect it should be fixed in Pydantic in `TypeAdapter`, and that should be what respects the internal configs. E.g. `TypeAdapter(list[Model])` should also work, but I would suspect it's not working.

I haven't played with it locally yet, running the code, etc. I might be wrong.

RB387:
Yep, you are correct. 

It seems that the problem is more complex than I initially thought. `TypeAdapter` indeed does use the model config, but the issue is that the config is not applied to the root model. This occurs because the root model is not wrapped by the TypeAdapter, and all fields are extracted separately from it. Therefore, this is why Pydantic is unaware of any config related to the root model.

So yep, my PR does not work. I added to this PR test case that should pass

shrey1605:
Is this issue still open? Looking to contribute

RB387:
@shrey1605 yep, issue is still open

rmawatson:
Some more info on this [here](https://github.com/fastapi/fastapi/discussions/13690). Worth noting it seems you can work around this by adding a second dummy parameter to
```
@app.post("/")
async def read_root(a: Model, b :Model = Model()) -> bool:
    return True
```

The issue lies with [this](https://github.com/fastapi/fastapi/blob/0.113.0/fastapi/dependencies/utils.py#L807-L810) , from [this](https://github.com/fastapi/fastapi/pull/12129) commit.

rmawatson:
[this PR](https://github.com/fastapi/fastapi/pull/13694) should fix the problem discussed here.

YuriiMotov:
#13694 contains the test case from this PR and approaches to solve the issue.
Let's close this PR in favor of #13694.

@RB387, thank you for your efforts!

PR #12507, opened 2024-10-22
♻️ Updating HTTP auth error codes to 401 while complying to RFC standards

mrPaintMan:
**TL;DR** This PR aims to change default error code for HTTPBearer and HTTPDigest to from 403_Forbidden to 401_Unauthorized.

# Goal of this PR
This PR aims to address the issues of having status code 403_forbidden for the HTTPBearer and HTTPDigest classes, and change them to 401_Unauthorized. From what we understand, the reason for why this hasn't been fixed already is what @tiangolo states in [his comment](https://github.com/fastapi/fastapi/pull/2120#issuecomment-1236165134) in PR #2120. This PR aims to comply to the error handling described in RFC standards [6750](https://datatracker.ietf.org/doc/html/rfc6750) (Bearer auth), [7616](https://datatracker.ietf.org/doc/html/rfc7616) (Digest auth) and [7617](https://datatracker.ietf.org/doc/html/rfc7617) (Basic auth).

This PR does __not__ aim to provide a working implementation of the Bearer or the Digest authentication schemas, but rather to improve the existing error handling.

## Changed behavior
In order to comply to these standards, we have introduced a default value (the value `global`) to the `realm` challenge of the ```WWW-Authenticate``` header, since it MUST be present for Basic auth, and MAY be present for Bearer and Digest auth. On top of this, Bearer auth MUST have at least one challenge in the ```WWW-Authenticate``` header, but no specific challenge is required, thus making the realm challenge suitable for this case as well.

Also, the challenge `qop` (quality of protection) MUST be present on an erroneous request to an endpoint protected by Digest auth. We solved this by defaulting this value to `auth` since we __interpreted__ that this is the default value according to the standard.  

## Additional changes
* Changed default error message to better correspond to default errors
* Added the components `qop` and `realm` to the openapi docs according to how other components are handled
* Added tests covering the usages of the field `realm` for Basic, Bearer and Digest auth
* Added tests covering the usages of the field `qop` for Digest auth

## Disclosure
We have very little experience with contributing to open source projects, and are very open to suggestions and feedback on this work!


svlandeg:
Hi @mrPaintMan, thanks for the detailed and thoughtful PR! We've had quite a few open PRs for this issue, but have recently decided that we'll likely go with https://github.com/fastapi/fastapi/pull/13786, which provides an additional fall-back mechanism for users that need more time to update (as this is a breaking change). As such, I'll go ahead and close this one. If you feel like it, always welcome to help review #13786 !

PR #12508, opened 2024-10-22
🐛 Ensure that dependencies can always catch validation errors

anton-ryzhov:
Hello everyone,

According to the [documentation](https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/#dependencies-with-yield-and-httpexception) dependencies should be able to catch and handle (postprocess, ignore, overwrite, …) exceptions.

In fact that works for almost all cases, but not for pydantic validation errors for input parameters. It looks to me that the code block was just accidentally deindented and was placed after `async with`-block not inside.

First commit adds new tests where 2 of 6 are failing. The second fixes the issue.


anton-ryzhov:
Any feedback would be highly appreciated

anton-ryzhov:
A kind reminder 😉

anton-ryzhov:
@tiangolo could you please take a look?

anton-ryzhov:
I'd like to have this bug fixed. Is there a way I can help you with this PR?

YuriiMotov (commented):
Simplified test that shows the idea of this PR:
```python

app = FastAPI()

errors = []

def dep_with_yield():
    try:
        yield "_"
    except RequestValidationError as e:
        print("Caught RequestValidationError")
        errors.append(e)
        raise

@app.get("/")
def get_(param: bool, _: str = Depends(dep_with_yield)):
    pass

client = TestClient(app)

def test_():
    errors.clear()
    client.get("/?param=invalid_value")
    assert len(errors) == 1  # assert 0 == 1

    assert isinstance(errors[0], RequestValidationError)
```
For now we can't catch `RequestValidationError` using `try..except` block inside the dependency with `yield`. This PR makes it possible.

I don't think that in current implementation _"code block was just accidentally deindented and was placed after async with-block not inside"_ (<= quote from the initial comment). For me it looks like it was designed this way intentionally.
At the same time I couldn't find any downsides of this new behavior except the fact that it may brake the code of existing projects if they catch exceptions and rewrite them (in some cases it may lead to hiding request validation errors). But it's easily fixable.

**UPD:** here is the [commit](https://github.com/fastapi/fastapi/pull/11191/commits/59deab57d3f3bd8b1e19bdb2724ad93ef282a366) that introduced the `if not errors:` block

anton-ryzhov:
> Simplified test that shows the idea of this PR:

Yeah, that's exactly the problem I'm trying to solve here.

> I don't think that in current implementation "code block was just accidentally deindented and was placed after async with-block not inside" (<= quote from the initial comment). For me it looks like it was designed this way intentionally.

That was just my assumption. I think this is not very important now.

> UPD: here is the [commit](https://github.com/fastapi/fastapi/pull/11191/commits/59deab57d3f3bd8b1e19bdb2724ad93ef282a366) that introduced the if not errors: block

This block was originally added [here](https://github.com/fastapi/fastapi/commit/a4aa79e0b4cacc6b428d415d04d234a8c77af9d5#diff-4d573079004a9f3d148baa4658e68e82b8a3d1a95d603fee8177daa92cf65c93R287-R291) and `RequestValidationError` was raised within the async stack.

And [this commit](https://github.com/fastapi/fastapi/pull/11191/commits/59deab57d3f3bd8b1e19bdb2724ad93ef282a366) has put it out for some reason.

anton-ryzhov:
What's next? Are there any chances of this to be merged?

svlandeg:
@anton-ryzhov: we have put this on Sebastián's internal board to solicit feedback from him. Thanks for your patience 🙏

anton-ryzhov:
https://github.com/fastapi/fastapi/pull/14099 has fixed the original issue. Do you still want to merge new tests for that case?

tiangolo:
Thanks for the effort @anton-ryzhov! :cake: 

If the original issue is now solved, then I think we can close this one. Thanks for reporting back that as well! :coffee:

anton-ryzhov:
I've re-tested it again and noticed an issue with the current implementation. If a handler returns success value, a dependency can't overwrite it with an Exception / HTTPException.

This returns 200:
```py
from fastapi import Depends, FastAPI, HTTPException

async def dependency():
    try:
        yield
    finally:
        raise HTTPException(status_code=400, detail="Dependency raises")

app = FastAPI(dependencies=[Depends(dependency)])

@app.get("/")
async def root():
    return None
```
and logs that error:
```
  File "fastapi/fastapi/routing.py", line 115, in app
    await wrap_app_handling_exceptions(app, request)(scope, receive, send)
  File "fastapi/.venv/lib/python3.11/site-packages/starlette/_exception_handler.py", line 56, in wrapped_app
    raise RuntimeError("Caught handled exception, but response already started.") from exc
RuntimeError: Caught handled exception, but response already started.
```

If a handler raises `HTTPException`, another `Exception` or has an input validation error — then the dependency can catch and handle/overwrite the error successfully

anton-ryzhov:
This breaks scenario when a dependency manages DB session and commits transaction after all. Currently an error on such commit will be ignored and false successful result will be returned

tiangolo:
Please check the new docs for dependencies with scope: https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/#early-exit-and-scope

PR #12513, opened 2024-10-23
Update cors.md

Sunnnner:
When setting ['*'], a TypeError is prompted: argument of type 'PydanticDescriptorProxy' is not iterable, modify it to ['*,']


Sunnnner:
My problem, adding cors middleware needs to be at the top of all middleware

PR #12528, opened 2024-10-24
✨ Accept a list of strings as value for JSON Schema "type"

dlax:
According to the [JSON Schema specification], the "type" keyword may be an array of strings (with unique items). Previously, the `fastapi.openapi.models.Schema` definition did not allow this and would thus prevent users from using custom JSON Schema definitions in their field such as `Annotated[..., WithJsonSchema({"type": ["string", "null"]})` as an alternative to the `anyOf` which is usually generated (a ValidationError was raised when generated the OpenAPI definition).

We now accept this "type" variant, as illustrated in modified tests.

[JSON Schema specification]: https://json-schema.org/draft/2020-12/json-schema-validation#name-type


fastapi/openapi/models.py
```diff
@@ -157,7 +157,7 @@ class Schema(BaseModelWithConfig):
     unevaluatedProperties: Optional["SchemaOrBool"] = None
     # Ref: JSON Schema Validation 2020-12: https://json-schema.org/draft/2020-12/json-schema-validation.html#name-a-vocabulary-for-structural
     # A Vocabulary for Structural Validation
-    type: Optional[str] = None
+    type: Optional[Union[str, Set[str]]] = None
```
YuriiMotov, line 160: We have to decide here whether we really need to use `set` or we can use `list`.
We ensure that all elements are unique, but we lose the order of elements.
I would say it's better to keep the order and leave the responsibility for uniqueness of elements to developer.
Let's think about this.
Maybe we can use list with the validator or constraints?

tests/test_custom_schema_fields.py
```diff
@@ -42,7 +54,18 @@ def foo():
...
+            {
+                "title": "Description",
+                "type": IsList("string", "null", check_order=False),
+            }
+            if PYDANTIC_V2
+            else {
+                "title": "Description",
+                "type": "string",
+            }
+        ),
```
YuriiMotov:
```suggestion
        "description": {
            "title": "Description",
            "type": (
                IsList("string", "null", check_order=False)
                if PYDANTIC_V2
                else "string"
            ),
        },
```
I think it's better to simplify it. Only the value of `type` field is different

YuriiMotov (commented):
Good idea in general, but I'm not sure about using `set`.
I noticed some people don't like that the order of parameters may change.
I've seen a comment that changing the order of elements may lead to Swagger interprets schema differently (can't find that comment now).

There is an alternative PR #13639 that uses `list` instead of `set` and also specifies possible values for elements

tiangolo:
Thanks for the work on this! :coffee: 

I'm gonna go with https://github.com/fastapi/fastapi/pull/13639, so I'll close this one now, but thanks for the effort! :cake:

PR #12529, opened 2024-10-24
✨ Add `lifespan` dependency scope

UltimateLobster:
### Support for lifespan scoped dependencies

This feature adds a scope to  `Depends`: `"lifespan"`.

If it is `"lifespan"` the dependency will be evaluated at the application startup (the value will be reused in all endpoints), and will be teared down at the application shutdown. It will be useful when defining db connections which need to be maintained and reused throughout the application's up-time:

```python
def mongo_client():
    with MongoClient(SOME_CONNECTION_STRING) as client:
        yield client

ActiveMongoClient = mongo_client: Annotated[MongoClient, Depends(mongo_client, scope="lifespan")]

# A single instance of MongoClient will be evaluated when the application starts. That same instance will be used for all
# requests to both endpoints. When the application shuts down, that single instance will be cleaned up as well.

@app.post("/users")
def insert_user(
    mongo_client: ActiveMongoClient,
    username: str,  
) -> str:
    inserted_id = mongo_client["SomeDb"]["Users"].insert_one({"username": username})
    return user_id

@app.get("/users/{user_id}/username")
def get_username(
    mongo_client: ActiveMongoClient,
    user_id: Annotated[str, Path()],
) -> str:
    user = mongo_client["SomeDb"]["Users"].find_one({"_id": user_id})
    return user["username"]
```

The `use_cache` argument works in the same way for dependencies with `"lifespan"` scope as it is for dependencies with `"request"` or `"function"` scope:
```python
def mongo_client():
    with MongoClient(SOME_CONNECTION_STRING) as client:
        yield client

DedicatedMongoConnection = mongo_client: Annotated[MongoClient, Depends(
    mongo_client, 
    scope="lifespan",
    use_cache=False
)]

# Each of these endpoints would have their own, dedicated mongo client which be reused for all requests to its respective endpoint. So overall, 2 instances will be created when the application starts. Those same 2 will be cleaned when the application shuts down.

@app.post("/users")
def insert_user(
    mongo_client: DedicatedMongoConnection ,
    username: str,  
) -> str:
    inserted_id = mongo_client["SomeDb"]["Users"].insert_one({"username": username})
    return user_id

@app.get("/users/{user_id}/username")
def get_username(
    mongo_client: DedicatedMongoConnection ,
    user_id: Annotated[str, Path()],
) -> str:
    user = mongo_client["SomeDb"]["Users"].find_one({"_id": user_id})
    return user["username"]
```

### Implementation details to note:

- There is a new, internal lifespan passed to the application's main router, which will be responsible to evaluate the dependencies.
- If the user already has specified a `lifespan` argument, it will be merged with the internal lifespan using the recently added `_merge_lifespan_context` utility
- Starlette forbids the use of startup/shutdown events in combination with the `lifespan` argument (if a lifespan is supplied, the events will never be handled). Since this PR makes it so we always have a `lifespan` arugment passed, it would've meant the `on_event`, `on_startup` and `on_shutdown` features would break. I solved it by manually calling them in the application's internal lifespan.
- The lifespan scoped dependencies are gathered by iterating over each of the application's routes recursively. I added a property to APIRoutes which exposes this information, and updated the implementation of `get_flat_dependent` to concat dependencies as well.

### ~~Things left to do:~~
- ~~There are still tests missing from this PR which I'm still working on (Specifically, I need to add tests to cover interactions with websockets, as well as interactions with dependency overrides).~~
- ~~I still need to add the necessary documentation and examples.~~


dolfinus:
You can just create client instance, and then pass it to application endpoints using `dependency_overrides`:

```python
app = FastAPI()
client = ...
app.dependency_overrides[MongoClient] = client
```

UltimateLobster:
> You can just create client instance, and then pass it to application endpoints using `dependency_overrides`:
> 
> ```python
> app = FastAPI()
> client = ...
> app.dependency_overrides[MongoClient] = lambda: client
> ```

There are a few problems with this approach which this feature should be able to address:
1. The client is never cleaned up at the server shutdown or on a reload. 
2. The creation of a resource during the import is problematic when you work with things like gunicorn, which may need to run the application in a separate process (There are clients which don't handle a fork of the connection object well).
3. The creation may need to be asynchronous which further complicates the ability to do so in the global scope.
4. Using dependency_overrides may lead to unexpected behavior when testing your application and one dependency overrides will take precedence over the other.
5. Dependencies defined with `use_cache=False` will still use the exact same client (they should have a separate client just for them).
6. When dealing with larger applications and lots of dependencies, it might be easy to forget to override the global dependency (since it's defined in a different place), and it will be a problem that might be harder to identify at the development stage since technically it will cause no errors.

Further discussion on the subject exists [here](https://github.com/fastapi/fastapi/issues/617).

dillondaudert:
Is this planned to be merged?

UltimateLobster:
> Is this planned to be merged?

I have no idea. I definitely hope so though it seems to not be prioritized by the maintainers right now. (Considering the lack of CR)
However, if they request any changes or updates to the branch, I'm available to do so.

UltimateLobster:
Updated the PR so it uses the new `scope` parameter added to `Depends` in [version 0.121.0](https://github.com/fastapi/fastapi/pull/14262)

docs/en/docs/tutorial/dependencies/lifespan-scoped-dependencies.md
```diff
@@ -0,0 +1,108 @@
...
+called, they will be using the previously saved connection. Once the application
+shuts down, **FastAPI** will make sure to gracefully close the connection object.
+
+## The `use_cache` argument
+
+The `use_cache` argument works similarly to the way it worked with endpoint
+scoped dependencies. Meaning, as **FastAPI** gathers lifespan scoped dependencies, it
+will cache dependencies it already encountered before. However, you can disable
+this behavior by passing `use_cache=False` to `Depends`. This will cause a new lifespan
+dependency to be created for every endpoint/dependency/router where it shows up:
```
YuriiMotov, line 61: I think this is not correct. For other dependency scopes, `use_cache` works for this request only. Here, `use_cache` has different meaning.
I think for this reason new parameter should be introduced instead of `use_cache`. And passing `use_cache`  should raise warning as it's useless.

Actually, I'm not sure it's needed.
Do you have any particular use cases for this option?

YuriiMotov (commented):
@UltimateLobster, thank you for work you've done! Looks like you've put a lot of efforts into it!

This feature requires quite a lot of changes. So, let's first wait for Sebastian to validate the general idea.

UltimateLobster:
> @UltimateLobster, thank you for work you've done! Looks like you've put a lot of efforts into it!
> 
> This feature requires quite a lot of changes. So, let's first wait for Sebastian to validate the general idea.

Thanks for taking a look at this!
I'm available to do whatever changes needed so this feature may get in. So hopefully Sebastian will like it :)

docs/en/docs/tutorial/dependencies/lifespan-scoped-dependencies.md
UltimateLobster: I guess I meant to say it works the same way for the respective scope. The relationship between the request dependencies and the request cache, is the same as between lifespan dependencies and lifespan cache. I should've found a better way to explain this.

As for its usefulness, I imagined a case of someone wanting a dedicated evaluation of the dependency. The example I gave in the documentation was a dedicated connection that was only used for requests of a certain route (which would usually require to create a new dependency function for that route).

But I agree it might be somewhat of a weak reason to support `use_cache=False` (along with wanting to allow a behavior that a reader of `Depends`'s signature might expect).

If you think it would be better to disallow `use_cache=False` until a better argument arises, I won't mind doing it .

nidhishgajjar:

## Summary

This PR adds a new "lifespan" dependency scope to FastAPI, allowing dependencies to be evaluated once at application startup, shared across all requests, and cleaned up at application shutdown. This is a significant feature for managing resources like database connections that should persist for the application's entire lifetime.

The implementation involves:
- Splitting the `Dependant` class into `EndpointDependant` and `LifespanDependant`
- Adding a new `fastapi/lifespan.py` module to manage lifespan-scoped dependencies
- Modifying the application initialization to always use an internal lifespan context
- Extensive test coverage (922+ lines) for various scenarios
- Comprehensive documentation with examples

## Architecture

### Core Changes

**1. Dependency Model Refactor** (`fastapi/dependencies/models.py`):
- Introduced `_BaseDependant` base class with shared logic
- `LifespanDependant`: For dependencies evaluated at startup/shutdown
- `EndpointDependant`: For per-request dependencies (renamed from `Dependant`)
- Maintains backwards compatibility via `Dependant: TypeAlias = EndpointDependant`

**2. New Lifespan Management** (`fastapi/lifespan.py`):
```python
async def resolve_lifespan_dependants(
    *, app: FastAPI, async_exit_stack: AsyncExitStack
) -> Dict[LifespanDependencyCacheKey, Callable[..., Any]]
```
- Iterates through all routes to collect lifespan dependencies
- Uses `AsyncExitStack` to ensure proper cleanup
- Supports dependency overrides

**3. Application Lifecycle Changes** (`fastapi/applications.py`):
- Always uses an internal lifespan context (`_internal_lifespan`)
- Merges user-provided lifespan with internal lifespan
- Manually calls startup/shutdown event handlers (since Starlette won't when lifespan is set)

**4. Type System** (`fastapi/types.py`):
```python
EndpointDependencyScope: TypeAlias = Union[Literal["request", "function"], None]
LifespanDependencyScope: TypeAlias = Literal["lifespan"]
DependencyScope: TypeAlias = Union[LifespanDependencyScope, EndpointDependencyScope]
```

### Dependency Resolution Flow

1. **At Application Startup**:
   - Iterate through all routes
   - Extract `lifespan_dependencies` from each route
   - Deduplicate by cache key
   - Solve each lifespan dependency
   - Store in `lifespan_scoped_dependencies` in app state

2. **During Request Handling**:
   - Endpoint dependencies access lifespan dependencies from app state
   - Lifecycle: function → request → lifespan (widest scope)

## Issues

### Critical Issues

**None** — No critical bugs that would prevent the feature from working.

### Warnings

**1. Potential Race Condition in Lifespan Dependency Access**

Location: `fastapi/routing.py` (request handler implementation)

The PR doesn't show where lifespan dependencies are accessed from request handlers. However, based on the architecture, there's a potential race condition:

```python
# In request handler
lifespan_deps = request.state.get("__fastapi__", {}).get("lifespan_scoped_dependencies", {})
db_connection = lifespan_deps.get(cache_key)
```

**Issue**: If a request arrives before the internal lifespan has finished initializing all dependencies, `lifespan_scoped_dependencies` might not be fully populated. FastAPI's `TestClient` might not expose this, but production ASGI servers could send requests during startup.

**Suggested Fix**: Ensure lifespan dependencies are fully initialized before accepting requests:
```python
@asynccontextmanager
async def _internal_lifespan(self) -> AsyncGenerator[Dict[str, Any], None]:
    async with AsyncExitStack() as exit_stack:
        lifespan_scoped_dependencies = await resolve_lifespan_dependants(
            app=self, async_exit_stack=exit_stack
        )
        # Ensure dependencies are ready before yielding
        yield {
            "__fastapi__": {
                "lifespan_scoped_dependencies": lifespan_scoped_dependencies
            }
        }
```

Actually, looking at the code more carefully, this appears to be handled correctly - the context manager yields after dependencies are resolved. However, there's no verification that the yielded state is actually used by request handlers.

**2. Missing Error Handling in Lifespan Dependency Resolution**

Location: `fastapi/lifespan.py:32-44`

```python
async def resolve_lifespan_dependants(
    *, app: FastAPI, async_exit_stack: AsyncExitStack
) -> Dict[LifespanDependencyCacheKey, Callable[..., Any]]:
    lifespan_dependants = _get_lifespan_dependants(app)
    dependency_cache: Dict[LifespanDependencyCacheKey, Callable[..., Any]] = {}
    for lifespan_dependant in lifespan_dependants:
        solved_dependency = await solve_lifespan_dependant(
            dependant=lifespan_dependant,
            dependency_overrides_provider=app,
            dependency_cache=dependency_cache,
            async_exit_stack=async_exit_stack,
        )
        dependency_cache.update(solved_dependency.dependency_cache)
    return dependency_cache
```

**Issue**: If one lifespan dependency fails to initialize (e.g., database connection error), the entire application fails to start. There's no way to mark some dependencies as "optional" or handle failures gracefully.

**Impact**: A single failing dependency prevents the application from starting, even if other endpoints don't use it.

**Suggested Fix**: Add error handling with configuration for optional dependencies:
```python
@dataclass
class Depends:
    # ...
    scope: DependencyScope = None
    optional: bool = False  # NEW: Allow graceful failure

# In resolve_lifespan_dependants:
for lifespan_dependant in lifespan_dependants:
    try:
        solved_dependency = await solve_lifespan_dependant(...)
        dependency_cache.update(solved_dependency.dependency_cache)
    except Exception as e:
        if getattr(lifespan_dependant.call, "_optional", False):
            logger.warning(f"Optional lifespan dependency failed: {e}")
            dependency_cache[dependant.cache_key] = None
        else:
            raise
```

**3. Potential Memory Leak with Uncleaned Dependencies**

Location: `fastapi/lifespan.py:32-44`

**Issue**: If `solve_lifespan_dependant` succeeds for some dependencies but fails for others, the `AsyncExitStack` might not clean up the successfully initialized dependencies properly.

**Impact**: Resources (database connections, file handles, etc.) from partially initialized dependencies could leak.

**Suggested Fix**: The current implementation uses `async with AsyncExitStack()`, which should handle cleanup on exception. However, verify this behavior:
```python
# Add test for partial initialization failure
def test_partial_lifespan_dependency_failure():
    calls = []
    
    async def good_dep():
        calls.append("good_init")
        try:
            yield "good"
        finally:
            calls.append("good_cleanup")
    
    async def bad_dep():
        calls.append("bad_init")
        raise RuntimeError("Failed to initialize")
    
    app = FastAPI()
    @app.get("/")
    def endpoint(
        g: Annotated[str, Depends(good_dep, scope="lifespan")],
        b: Annotated[str, Depends(bad_dep, scope="lifespan")],
    ):
        return {"g": g}
    
    with pytest.raises(RuntimeError):
        with TestClient(app):
            pass
    
    # Verify good_dep was cleaned up
    assert "good_cleanup" in calls
```

**4. Compatibility Issue with Python 3.8 and 3.9**

Location: `fastapi/dependencies/models.py:76`

```python
@dataclass
class LifespanDependant(_BaseDependant):
    scope: LifespanDependencyScope
    caller: Callable[..., Any]
    call: Callable[..., Any]
    dependencies: List["LifespanDependant"] = field(default_factory=list)
    use_cache: bool = True
    name: Optional[str] = None
    index: Optional[int] = None
```

**Issue**: The dataclass doesn't use `kw_only=True` (not available in Python 3.8/3.9), which could lead to confusion when instantiating with positional arguments.

**Impact**: Users might accidentally pass arguments in wrong order, leading to subtle bugs.

**Suggested Fix**: Already addressed in commit `a4984b83`: "Removed kw_only from dataclass since it's not supported in python 3.8 and 3.9". ✅

**5. Missing Validation for Cyclic Dependencies**

Location: `fastapi/dependencies/utils.py:258-314`

```python
def get_lifespan_dependant(
    *,
    caller: Callable[..., Any],
    call: Callable[..., Any],
    scope: LifespanDependencyScope,
    ...
) -> LifespanDependant:
    # ...
    for param_name, param in signature_params.items():
        param_details = analyze_param(...)
        if param_details.depends is None:
            raise DependencyScopeError(...)
        
        if param_details.depends.scope != "lifespan":
            raise DependencyScopeError(...)
        
        sub_dependant = get_lifespan_dependant(...)
        dependant.dependencies.append(sub_dependant)
    return dependant
```

**Issue**: There's no detection for cyclic dependencies (e.g., A depends on B, B depends on A). This would cause infinite recursion.

**Impact**: Application hangs during startup with stack overflow.

**Suggested Fix**: Add cycle detection:
```python
def get_lifespan_dependant(
    *,
    caller: Callable[..., Any],
    call: Callable[..., Any],
    scope: LifespanDependencyScope,
    name: Optional[str] = None,
    use_cache: bool = True,
    index: Optional[int] = None,
    _visited: Optional[Set[Tuple[Callable, str]]] = None,  # NEW
) -> LifespanDependant:
    if _visited is None:
        _visited = set()
    
    visit_key = (call, name or "")
    if visit_key in _visited:
        raise DependencyScopeError(
            f"Cyclic dependency detected: {caller.__name__} -> {call.__name__}"
        )
    _visited.add(visit_key)
    
    try:
        # ... existing code ...
        for param_name, param in signature_params.items():
            # ...
            sub_dependant = get_lifespan_dependant(
                ..., _visited=_visited.copy()  # Don't share across siblings
            )
            dependant.dependencies.append(sub_dependant)
        return dependant
    finally:
        _visited.remove(visit_key)
```

### Suggestions

**1. Add Metrics for Lifespan Dependency Initialization**

Track how long lifespan dependencies take to initialize and how many are created:
```python
import time

@asynccontextmanager
async def _internal_lifespan(self) -> AsyncGenerator[Dict[str, Any], None]:
    async with AsyncExitStack() as exit_stack:
        start_time = time.time()
        lifespan_scoped_dependencies = await resolve_lifespan_dependants(
            app=self, async_exit_stack=exit_stack
        )
        init_time = time.time() - start_time
        logger.info(
            f"Initialized {len(lifespan_scoped_dependencies)} "
            f"lifespan dependencies in {init_time:.3f}s"
        )
        # ...
```

**2. Add Support for Lifespan Dependency Health Checks**

Allow checking if lifespan dependencies are still healthy:
```python
@dataclass
class Depends:
    # ...
    health_check: Optional[Callable[[Any], Awaitable[bool]]] = None

# In a health check endpoint:
@app.get("/health")
async def health_check(
    request: Request,
):
    lifespan_deps = request.state.get("__fastapi__", {}).get("lifespan_scoped_dependencies", {})
    unhealthy = []
    for cache_key, (depends, value) in lifespan_deps.items():
        if depends.health_check:
            is_healthy = await depends.health_check(value)
            if not is_healthy:
                unhealthy.append(cache_key)
    
    if unhealthy:
        raise HTTPException(
            status_code=503,
            detail=f"Unhealthy dependencies: {unhealthy}"
        )
    return {"status": "healthy"}
```

**3. Consider Adding Thread-Safety for Lifespan Dependencies**

If lifespan dependencies are accessed from multiple threads (e.g., background tasks), ensure thread-safety:
```python
# In request handler or background task
lifespan_deps = request.state.get("__fastapi__", {}).get("lifespan_scoped_dependencies", {})
db_connection = lifespan_deps.get(cache_key)

# If db_connection needs thread-safety:
if not getattr(db_connection, "_is_thread_safe", True):
    logger.warning(
        f"Lifespan dependency {cache_key} may not be thread-safe. "
        "Consider using thread-safe implementations."
    )
```

**4. Add Documentation About Lifespan Dependency Limitations**

The documentation should clearly state:
- Lifespan dependencies cannot use request-specific parameters (path, query, body, etc.)
- Lifespan dependencies are shared across all requests (be careful with mutable state)
- Lifespan dependencies are resolved before the application accepts requests
- Failed lifespan dependencies prevent application startup

**5. Consider Adding Warmup/Reload Support**

Allow reinitializing specific lifespan dependencies without restarting the entire application:
```python
@app.post("/admin/reload-dependencies")
async def reload_dependencies(
    request: Request,
    dependency_names: List[str] = Body(...),
):
    # Reload specific dependencies
    # This would require significant architecture changes
    pass
```

## Cross-file Impact

### Breaking Changes

**1. Dependant Class Split**

- Old: Single `Dependant` class
- New: `EndpointDependant` and `LifespanDependant`
- **Backwards Compatible**: `Dependant: TypeAlias = EndpointDependant`

**Impact**: Most code should continue to work. However, code that explicitly type-hints `Dependant` and assumes it has all fields might break.

**2. Always-Use-Lifespan**

The application now always uses an internal lifespan context, even if the user doesn't provide one.

**Impact**: 
- `on_event`, `on_startup`, `on_shutdown` decorators now call handlers manually instead of letting Starlette handle them
- This should be transparent to users, but edge cases might exist

**Testing**: The PR includes updated `test_router_events.py` tests that verify event handlers still work correctly. ✅

### Public API Changes

**New**:
- `Depends(scope="lifespan")` — New scope option
- `LifespanDependant` class (internal)
- `EndpointDependant` class (internal)
- `DependencyScope`, `EndpointDependencyScope`, `LifespanDependencyScope` types (internal)
- `InvalidDependencyScope` exception
- `UninitializedLifespanDependency` exception
- `merge_lifespan_context()` function (renamed from `_merge_lifespan_context`, made public)

**Modified**:
- `Dependant` → `EndpointDependant` (with backwards compatibility alias)
- `Dependant.dependencies` property now returns tuple of both endpoint and lifespan dependencies
- `FastAPI._internal_lifespan` (private method)
- `FastAPI.on_event()` implementation changed

**Removed**:
- None (all changes are additive)

### Callers Affected

**Direct Callers**: None (all changes are internal to FastAPI)

**Indirect Callers**: Users who:
- Import `Dependant` from `fastapi.dependencies.models` (should still work via type alias)
- Rely on specific implementation details of dependency resolution
- Have custom routers that expect old `Dependant` structure

### Test Coverage

**Excellent** — The PR includes comprehensive test coverage:

1. **test_lifespan_scoped_dependencies/test_endpoint_usage.py** (922 lines):
   - Basic lifespan dependency usage
   - Use cache enabled/disabled
   - Sub-dependencies
   - Dependency overrides
   - Scope conflicts (lifespan depends on request-scoped)
   - WebSockets support
   - Multiple endpoints sharing dependencies

2. **test_lifespan_scoped_dependencies/test_dependency_overrides.py**:
   - Override lifespan dependencies
   - Override with endpoint-scoped parameters (should fail)
   - Override with endpoint-scoped sub-dependencies (should fail)

3. **tests/test_router_events.py**:
   - Verify `on_event`, `on_startup`, `on_shutdown` still work
   - Test class-based event handlers

4. **Tutorial tests** (tutorial013a-d):
   - Test all documentation examples

**Missing**:
- Performance tests for high-volume scenarios
- Tests for cyclic dependency detection (if added)
- Tests for partial initialization failure
- Integration tests with real database connections

### Dependencies

**Added**:
- None (uses existing dependencies)

**Removed**:
- None

**Updated**:
- None mentioned in PR

### User Impact

**Positive**:
- Enables efficient resource sharing (database connections, connection pools, etc.)
- Reduces overhead for expensive initialization
- Clean, intuitive API (`scope="lifespan"`)
- Comprehensive documentation and examples
- Backwards compatible with existing code

**Negative**:
- Slightly increased complexity in dependency resolution
- New failure mode: application won't start if lifespan dependency fails
- Learning curve for understanding when to use "lifespan" vs "request" scope

## Assessment

✅ **Approve**

This is a well-designed and thoroughly tested implementation of a highly requested feature. The code demonstrates:

**Strengths**:
- Clean architecture with clear separation of concerns
- Backwards compatible design (type aliases, property-based compatibility)
- Comprehensive test coverage (922+ lines of tests)
- Excellent documentation with practical examples
- Proper use of async context managers for resource management
- Handles edge cases (dependency overrides, scope conflicts, websockets)
- Maintains compatibility with existing `on_event` handlers

**Minor Areas for Future Enhancement**:
1. Add cycle detection for lifespan dependencies (prevents infinite recursion)
2. Consider optional lifespan dependencies (graceful failure handling)
3. Add metrics/monitoring for lifespan dependency initialization
4. Add health check support for lifespan dependencies

The implementation is production-ready and provides significant value for FastAPI applications that need to share resources across the application lifetime. The suggested enhancements are optional improvements that could be added in follow-up PRs.

**Testing Recommendation**:
- Test with real database connections (MongoDB, PostgreSQL, etc.)
- Test under load to verify no race conditions during startup
- Test with applications that have many routes and dependencies
- Test failure scenarios (database down, network issues, etc.)

PR #12534, opened 2024-10-24
✅ Use `inline-snapshot` to support different Pydantic versions in the test suite

15r10nk:
This pull-request contains an example how inline-snapshot can be used with different snapshots for different pydantic versions.

The following code can be used to create different snapshots for each pydantic version (pytest has to be run with both pydantic version).

``` python
assert some_value == snapshot_pydantic(v1=snapshot(),v2=snapshot())
```

both snapshots will be created (see example in the pull-request).

@tiangolo
I would be interested if this approach looks useful to you, or if you prefer the current IsDict(...)|IsDict(...) pattern for some reason.

The next version of inline-snapshot will support things like `snapshot_pydantic()` inside of snapshots. which will make the following code work:

``` python
    assert response.json() == snapshot(
        {
            "openapi": "3.1.0",
            "paths": {
                "/items/": {
                    "get": {
                        "summary": "Read Items",
                        "operationId": "read_items_items__get",
                        "parameters": [
                            {
                                "name": "session_id",
                                "in": "cookie",
                                "required": True,
                                "schema": {"type": "string", "title": "Session Id"},
                            },
                            {
                                "name": "fatebook_tracker",
                                "in": "cookie",
                                "required": False,
                                "schema": snapshot_pydantic(v1=snapshot(),v2=snapshot()),
                            },
```

Thank you for using inline-snapshot :smiley:.


tiangolo:
This is great! :rocket: 

I would love to have this in all the tests, but that's a giant task, you don't have to do it. :sweat_smile: 

Having this first one as a template is already great. :rocket:

tests/utils.py
```diff
@@ -2,10 +2,33 @@
...
 needs_pydanticv1 = pytest.mark.skipif(PYDANTIC_V2, reason="requires Pydantic v1")
+
+
+def pydantic_snapshot(
+    *,
+    v2: Snapshot,
+    v1: Snapshot,
+):
+    """
+    this function should be used like:
```
svlandeg, line 21:
```suggestion
    This function should be used like this:
```

tests/utils.py
```diff
@@ -2,10 +2,33 @@
...
+
+    >>> assert value == pydantic_snapshot(v2=snapshot(),v1=snapshot())
+
+    inline-snapshot will create the snapshots when pytest is executed for each versions of pydantic.
+
+    It is also possible to use the function inside snapshots for version specific values.
```
svlandeg, line 27:
```suggestion
    It is also possible to use the function inside snapshots for version-specific values.
```

tests/utils.py
```diff
@@ -2,10 +2,33 @@
...
+
+    >>> assert value == snapshot({
+        "data": "some data",
+        "vesion_specific": pydantic_snapshot(v2=snapshot(),v1=snapshot()),
```
svlandeg, line 31:
```suggestion
        "version_specific": pydantic_snapshot(v2=snapshot(),v1=snapshot()),
```

tests/test_tutorial/test_sql_databases/test_tutorial002.py
```diff
@@ -117,14 +117,14 @@ def test_crud_app(client: TestClient):
         )
         assert response.status_code == 200, response.text
         assert response.json() == snapshot(
-            {"name": "Dog Pond", "age": None, "id": hero_id}
+            {"name": "Dog Pond", "age": None, "id": Is(hero_id)}
```
svlandeg, line 120: What is the added advantage of using `Is()` here?

tests/test_tutorial/test_cookie_param_models/test_tutorial002.py
```diff
@@ -71,9 +77,8 @@ def test_cookie_param_model_invalid(client: TestClient):
                     }
                 ]
             }
-        )
-        | IsDict(
-            # TODO: remove when deprecating Pydantic v1
```
svlandeg, line 76: We should keep this TODO statement somehow

svlandeg (commented):
This looks great, just left a few small suggestions & questions.

tests/test_tutorial/test_cookie_param_models/test_tutorial002.py
15r10nk: Can we add it to the function? 

"Remove v1 argument when deprecating Pydantic v1"

Maybe we can use [codecrumbs](https://15r10nk.github.io/codecrumbs/introduction/) in the future to clean things up. 

There are a lot of refactorings missing which would be needed for this. But it would be a reason to continue this project :-)

15r10nk:
@svlandeg, thank you for the review. I added the missing comment.

tests/utils.py
```diff
@@ -2,10 +2,33 @@
...
     sys.version_info < (3, 10), reason="requires python3.10+"
 )
 needs_pydanticv2 = pytest.mark.skipif(not PYDANTIC_V2, reason="requires Pydantic v2")
 needs_pydanticv1 = pytest.mark.skipif(PYDANTIC_V2, reason="requires Pydantic v1")
+
+
+def pydantic_snapshot(
+    *,
+    v2: Snapshot,
+    v1: Snapshot,  # Remove v1 argument when deprecating Pydantic v1
```
svlandeg, line 18:
```suggestion
    v1: Snapshot,  #TODO: remove v1 argument when deprecating Pydantic v1
```
(in same format as other TODO's to facilitate simple ctrl-F'ing 😉 )

tests/utils.py
```diff
@@ -2,10 +2,33 @@
...
     sys.version_info < (3, 10), reason="requires python3.10+"
 )
 needs_pydanticv2 = pytest.mark.skipif(not PYDANTIC_V2, reason="requires Pydantic v2")
 needs_pydanticv1 = pytest.mark.skipif(PYDANTIC_V2, reason="requires Pydantic v1")
+
+
+def pydantic_snapshot(
+    *,
+    v2: Snapshot,
+    v1: Snapshot,  #TODO: remove v1 argument when deprecating Pydantic v1
```
svlandeg, line 18:
```suggestion
    v1: Snapshot,  # TODO: remove v1 argument when deprecating Pydantic v1
```

svlandeg (approved):
This PR looks good to merge to me, but I'll run it internally by Tiangolo for a final review. Thanks again @15r10nk!

tiangolo (approved):
This is awesome, thank you! :rocket: 

I need to copy this over to SQLModel too. :sunglasses: 

And thank you @svlandeg! :raised_hands:

PR #12569, opened 2024-10-26
Update include syntax for 'README.md'

bilalqv:


lucaromagnoli:
I believe the changes are meant to be only for python files import from docs_src not for Inline Python code blocks

Issue #12554, opened 2024-10-26
Update docs include syntax for source examples


tiangolo:
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

### Issue Content

This is a good first contribution. :nerd_face: 

The code examples shown in the docs are actual Python files. They are even tested in CI, that's why you can always copy paste an example and it will always work, the example is tested.

The way those examples are included in the docs used a specific format. But now there's a new format available that is much simpler and easier to use than the previous one, in particular in complex cases, for example when there are examples in multiple versions of Python.

But not all the docs have the new format yet. The docs should use the new format to include examples. That is the task. :nerd_face: 

**It should be done as one PR per page updated.**

## Simple Example

Before, the format was like:

````markdown
```Python hl_lines="3"
{!../../docs_src/first_steps/tutorial001.py!}
```
````

Now the new format looks like:

````markdown
{* ../../docs_src/first_steps/tutorial001.py hl[3] *}
````

* Instead of `{!` and `!}` it uses `{*` and `*}`
* It no longer has a line above with:

````markdown
```Python
````

* And it no longer has a line below with:

````markdown
```
````

* The highlight is no longer a line with e.g. `hl_lines="3"` (to highlight line 3), but instead in the same line there's a `hl[3]`.

An example PR: https://github.com/fastapi/fastapi/pull/12552

## Multiple Python Versions

There are some cases where there are variants of the same example for multiple versions of Python, or for using `Annotated` or not.

In those cases, the current include examples have syntax for tabs, and notes saying `Annotated` should be preferred. For example:

````markdown
//// tab | Python 3.9+

```Python hl_lines="4  8  12"
{!> ../../docs_src/security/tutorial006_an_py39.py!}
```

////

//// tab | Python 3.8+

```Python hl_lines="2  7  11"
{!> ../../docs_src/security/tutorial006_an.py!}
```

////

//// tab | Python 3.8+ non-Annotated

/// tip

Prefer to use the `Annotated` version if possible.

///

```Python hl_lines="2  6  10"
{!> ../../docs_src/security/tutorial006.py!}
```

////
````

In these cases, it should be updated to only include the first one (the others will be included automatically :sunglasses: ):

````markdown
{* ../../docs_src/security/tutorial006_an_py39.py hl[4,8,12] *}
````

* The syntax for tabs is also removed, all the other variants are included automatically.
[...]

PR #12571, opened 2024-10-26
📝 Update include syntax for `docs/az/docs/index.md`

bilalqv:


alejsdev:
Hi @bilalqv, thanks for your interest in contributing to FastAPI! However, the syntax is wrong, you can see in the preview that it's not working as expected. Can you update it with the proper syntax? The details are explained in #12554 
Thanks in advance! :raised_hands:

bilalqv:
will create new one

PR #12558, opened 2024-10-26
📝 Update includes in `docs/fr/docs/python-types.md`

Ismailtlem:
Update docs for docs/fr/docs/python-types/

PR #12587, opened 2024-10-26
Update includes in docs/fr/docs/python-types.md

kantandane:
Update includes in docs/fr/docs/python-types.md

Part of #12554


kantandane:
Ok sorry, it's a Duplicate of #12558

PR #12656, opened 2024-10-28
🌐 Add Korean translation for `docs/ko/docs/advanced/translated_using_request_directly.md`

kwang1215:


alejsdev:
Hi @kwang1215, thanks for your interest in contributing to FastAPI!
It would be better if you separate the other file into its own PR. Can you do it? 😄
Thanks in advance!

sodaMelon:
@kwang1215 hi 
could you check original docs name?
i think your work from [here: fast-api-docs/advanced/using-request-directly/](https://fastapi.tiangolo.com/advanced/using-request-directly/)

maybe......... you need to change your work's file name

after than, could you ping me?

PR #12718, opened 2024-10-29
Fix: code visualization

chaush-server:
Learn > Tutorial - User Guide > Request Body > Use the model
code visualization fix, based on other file references, it is supposed to look like this.


alejsdev:
Hi @chaush-server,
The original doesn't have a highlight, so we'll keep it as it is right now. Based on this, we'll close this PR, but thanks for your interest! :raised_hands:

PR #12742, opened 2024-11-01
📝 Remove mention of Celery in the project generators

david-caro:
In [full-stack-fastapi-template!694 ](https://github.com/fastapi/full-stack-fastapi-template/pull/694)celery was removed from the template project, it makes no sense anymore mentioning in the docs that is an example on how to use celery.


david-caro:
Being a simple removal of text, I bundled all the pages together, but if it's easier to review them one-by-one let me know and I'll split the change.

svlandeg (approved):
Makes sense to me - thanks for being so thorough through all the languages! In a case like this, it's definitely fine to keep all the changes in the same PR, as they're highly related.

I'll pass this on internally to Tiangolo for a final review 🙏

alejsdev (approved):

tiangolo (approved):
Nice, makes sense, thanks @david-caro! :cake: :coffee: 

And thanks for the reviews @svlandeg and @alejsdev :bow:

PR #12757, opened 2024-11-02
📝 Update includes for `docs/en/docs/tutorial/body.md`

gsheni:
- I noticed the [code snippet in the tutorial](https://fastapi.tiangolo.com/tutorial/body/#use-the-model) was not formatted correct so I fixed it
- Replace {!> and !} with {* and *}


docs/en/docs/tutorial/body.md
```diff
@@ -126,7 +126,7 @@ It improves editor support for Pydantic models, with:
 
 Inside of the function, you can access all the attributes of the model object directly:
 
-{!> ../../docs_src/body/tutorial002_py310.py!}
+{* ../../docs_src/body/tutorial002_py310.py hl[19] *}
```
tiangolo, line 129: Thanks! Let's keep the original highlighting please (so, no highlighting). :pray: 

It's not super clear what's the correct section to highlight here because almost all the code references the Pydantic model, so I prefer to leave no highlight there. :nerd_face:

tiangolo (changes requested):

docs/en/docs/tutorial/body.md
gsheni: Sure, done:
https://github.com/fastapi/fastapi/pull/12757/commits/07cb7a229362dc94ef7d36e166c46b42045577b3

tiangolo (approved):
Great, thank you! :rocket: :cake:

PR #12792, opened 2024-11-07
Update first-steps.md

Whoeza:
I changed this

*path operation decorator* --> *path operation* decorator

and

**"path operation decorator"** --> **"path operation" decorator**

because the "path operation" part is the specification of the current decorator, and not the full name of this feature. I didn't want new readers to call it "path operation decorator" but rather "decorator".

I don't know what you think about this small change. You'll let me know.


tiangolo:
Thanks for the interest! But it's intentional the way it is. a "path operation decorator" is one specific line, a "path operation function" starts exactly the next line, and a "path operation" includes both. So I'll pass on this one, but thanks! :coffee:

PR #12914, opened 2024-11-12
♻️ Update internal checks to support Pydantic 2.10

tiangolo:
♻️ Update internal checks to support Pydantic 2.10


tamird:
The tests are broken with the release of Pydantic 2.10. I sent #12971 to fix them.

PR #12915, opened 2024-11-12
🔧 Remove duplicate actions/checkout in `notify-translations.yml`

tinyboxvk:


alejsdev:
Good catch, thank you! @tinyboxvk :mag:

PR #12935, opened 2024-11-15
🐛 Fix `TypeError` when encoding a decimal with a `NaN` or `Infinity` value

kentwelcome:
Once the `jsonable_encoder` tries to encode a dict containing `Decimal('NaN')` or `Decimal('Infinity')`, the exception `TypeError: '>=' not supported between instances of 'str' and 'int'` will be thrown.

#### Example Code
```python
def encode_decimal_nan():
    jsonable_encoder({"value": Decimal("NaN")})  
```
```
  File "/Users/kent/Dev/Python/FastAPI/nan.py", line 8, in <module>
    encode_decimal_nan()
  File "/Users/kent/Dev/Python/FastAPI/nan.py", line 5, in encode_decimal_nan
    jsonable_encoder({"value": Decimal("NaN")})
  File "/Users/kent/Dev/Python/FastAPI/.venv/lib/python3.10/site-packages/fastapi/encoders.py", line 289, in jsonable_encoder
    encoded_value = jsonable_encoder(
  File "/Users/kent/Dev/Python/FastAPI/.venv/lib/python3.10/site-packages/fastapi/encoders.py", line 318, in jsonable_encoder
    return ENCODERS_BY_TYPE[type(obj)](obj)
  File "/Users/kent/Dev/Python/FastAPI/.venv/lib/python3.10/site-packages/fastapi/encoders.py", line 52, in decimal_encoder
    if dec_value.as_tuple().exponent >= 0:  # type: ignore[operator]
TypeError: '>=' not supported between instances of 'str' and 'int'
```

It's caused by the implementation of `decimal_encoder` in `fastapi/encoders.py:52`

```Python
if dec_value.as_tuple().exponent >= 0:  # type: ignore[operator]
        return int(dec_value)
    else:
        return float(dec_value)
```

The type of `exponent` defined in `DecimalTuple` could be `int` or `str` (ref: [python/typeshed](https://github.com/python/typeshed/blob/5052fa2f18db4493892e0f2775030683c9d06531/stdlib/decimal.pyi#L38))
```python
class DecimalTuple(NamedTuple):
    sign: int
    digits: tuple[int, ...]
    exponent: int | Literal["n", "N", "F"]
```

If the decimal value is `NaN` or `Infinity,` the exponent will be `str,` and it doesn't support the `>=` operator.
We should only cast the decimal value as `int` if the type of `exponent` is `int` type.

Reference discussion: https://github.com/fastapi/fastapi/discussions/10641


svlandeg (approved):
Thanks for the PR! Very much appreciate the clear explanation of what's going on and how to fix it.

I've had another look at the `Decimal` API, and the fix you're proposing feels like the best solution to me. All three tests fail on `master` and succeed with this PR.

The only comment I have is that we probably want to adjust the phrasing "Taken from Pydantic v1 as is" - I'll push that edit now.

svlandeg:
@kentwelcome: I don't seem to have the permission to push to your branch directly. Can you update the docstring to this:

```
# Adapted from Pydantic v1
# TODO: pv2 should this return strings instead?
def decimal_encoder(dec_value: Decimal) -> Union[int, float]:
    """
    Encodes a Decimal as int if there's no exponent, otherwise float

    This is useful when we use ConstrainedDecimal to represent Numeric(x,0)
    where an integer (but not int typed) is used. Encoding this as a float
    results in failed round-tripping between encode and parse.
    Our Id type is a prime example of this.

    >>> decimal_encoder(Decimal("1.0"))
    1.0

    >>> decimal_encoder(Decimal("1"))
    1

    >>> decimal_encoder(Decimal("NaN"))
    nan
    """
```
These should be the changes in the diff:

image

kentwelcome:
> @kentwelcome: I don't seem to have the permission to push to your branch directly. Can you update the docstring to this:
> 
> ```
> # Adapted from Pydantic v1
> # TODO: pv2 should this return strings instead?
> def decimal_encoder(dec_value: Decimal) -> Union[int, float]:
>     """
>     Encodes a Decimal as int if there's no exponent, otherwise float
> 
>     This is useful when we use ConstrainedDecimal to represent Numeric(x,0)
>     where an integer (but not int typed) is used. Encoding this as a float
>     results in failed round-tripping between encode and parse.
>     Our Id type is a prime example of this.
> 
>     >>> decimal_encoder(Decimal("1.0"))
>     1.0
> 
>     >>> decimal_encoder(Decimal("1"))
>     1
> 
>     >>> decimal_encoder(Decimal("NaN"))
>     nan
>     """
> ```
> 
@svlandeg  Thanks for your feedback! I have modified my branch based on the diff you provided!

fastapi/encoders.py
```diff
@@ -32,14 +32,14 @@ def isoformat(o: Union[datetime.date, datetime.time]) -> str:
     return o.isoformat()
 
 
-# Taken from Pydantic v1 as is
+# Adapted from Pydantic v1 as is
```
svlandeg:
```suggestion
# Adapted from Pydantic v1
```

svlandeg (approved):
Great, this looks good to merge to me. I'll pass it on to Tiangolo for a final review. Thanks again, @kentwelcome!

fastapi/encoders.py
kentwelcome: fixed

tiangolo (approved):
Awesome, thank you @kentwelcome! 🙌 

And thanks @svlandeg for the help. ☕ 

This will be available in FastAPI 0.123.1 in the next few hours. 🎉

PR #12944, opened 2024-11-16
✨ Allow to have multiple Query parameter models

uriyyo:
Currently, it's allowed to have only one model as a query/headers/etc model parameter, in this PR I propose to remove this restriction. This PR should be backward-compatible.

For instance, example below will not work:
```py
from typing import Annotated

app = FastAPI()


class FilterParams(BaseModel):
    name: str
    age: int


class PageParams(BaseModel):
    limit: int
    offset: int


@app.get("/")
async def route(
    filters: Annotated[FilterParams, Query()],
    page: Annotated[PageParams, Query()],
):
    return {}
```

And same here:
```py
from typing import Annotated

app = FastAPI()


class FilterParams(BaseModel):
    name: str
    age: int


@app.get("/")
async def route(
    filters: Annotated[FilterParams, Query()],
    limit: Annotated[int, Query()],
    offset: Annotated[int, Query()],
):
    return {}
```

WIth this PR both examples above will work as expected. Also, it will generate the correct OpenAPI schema:
image

Discussion - https://github.com/fastapi/fastapi/discussions/12212


wu-clan:
Need additional test cases?

wu-clan:
This looks similar to #12481

uriyyo:
I agree, It does solve the same issue. I didn't see it when I started working on this PR(

wu-clan:
Hi, @svlandeg 
I wonder if anyone has paid attention to this, which is a very anticipated function.

wu-clan:
This is really a very good feature, we want to include it in our 1.0.0 version of the project, so I have been waiting for it. Compared to testing, the priority of the new feature should be higher, I hope you can see this @tiangolo, and I am very sorry for the ping.

jperuggia:
Hope this gets approved and released soon !

fastapi/dependencies/utils.py
```diff
@@ -212,11 +212,14 @@ def get_flat_dependant(
...
-        return fields_to_extract
-    return fields
+
+    fields_to_extract = []
+    for f in fields:
+        if lenient_issubclass(f.type_, BaseModel):
+            fields_to_extract.extend(get_cached_model_fields(f.type_))
+        else:
+            fields_to_extract.append(f)
+    return fields_to_extract
```
Ale-Cas, line 222: I think that this could be simplified a little bit: 
```python
fields_to_extract = [get_cached_model_fields(f.type_) for f in fields if lenient_issubclass(f.type_, BaseModel)]
fields.extend(fields_to_extract)
return fields
```

fastapi/dependencies/utils.py
```diff
@@ -747,15 +750,19 @@ def request_params_to_args(
...
 
+    model_fields = [
+        field for field in fields if lenient_issubclass(field.type_, BaseModel)
+    ]
+    if model_fields:
+        fields_to_extract = [
+            cached_field
+            for field in model_fields
+            for cached_field in get_cached_model_fields(field.type_)
+        ]
```
Ale-Cas, line 764: I think you might be able to combine those 2 list comprehensions: 
```python
fields_to_extract = [
    cached_field
    for field in fields if lenient_issubclass(field.type_, BaseModel)
    for cached_field in get_cached_model_fields(field.type_)
]
```

Mulugruntz:
This looks awesome! Really looking forward to it!
Do we know if/when it'll be approved and released?

Vaserty:
Do we know when this will be published? It would make things easier.

wu-clan:
Unfortunately, this feature didn't attract any maintainers.

fastapi/dependencies/utils.py
uriyyo: I'm not sure it will work, because we will have both base model and all fields, we want to remove based model and keep only it fields.

fastapi/dependencies/utils.py
uriyyo: Yup, it looks better, thanks for suggestion

uriyyo:
Just merged with master and fixed tests, @Kludex any chance you can take a look on this PR? 🙏

bassemkaroui:
@Ale-Cas can you please take a look at this ?

bobbyo:
is this merging soon?  would be nice to have pagination and query on models in FastAPI, which [is blocked ](https://github.com/uriyyo/fastapi-pagination/issues/1314) on merging this PR ....

bjmc:
Just chiming in that my team are also interested in this feature. :pray:

bjmc:
I've just noticed that the same limitation seems to affect `Form()`, could this fix be extended to cover form POST as well?

uriyyo:
Hi @tiangolo,

Any chance you can take a look at this PR? 🙏

wu-clan:
The maintenance team seems to be moving forward with #12481, but still have to wait for a response from tiangolo.

henri9813:
Hello, 

Is there any news on this ?

bassemkaroui (approved):

henri9813 (approved):

JustBipin:
any update on this?

fastapi/dependencies/utils.py
JustBipin: I does looks less lines of code but at the cose of readability, better keep it the way it is

PR #12955, opened 2024-11-19
✨ Add support for ReDoc parameters

SteinRobert:
This PR add simple support for ReDoc parameters - see here https://redocly.com/docs/redoc/config

Example config:
```python
{
   "disable-search": "true"
}
```

Will look like this:

```python
app = FastAPI(
    title="My API",
    docs_url=None,
    redoc_url="/docs",
    openapi_url="/docs/openapi.json",
    redoc_ui_parameters={
        "disable-search": "true"
    }
)
```

Will be added to the generated html:
```html
<redoc spec-url="/docs/openapi.json" disable-search="true"></redoc>
```

ReDoc will pickup the parameter and act accordingly.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -444,6 +444,15 @@ class FastAPI(Starlette):
                 """
             ),
         ] = "/redoc",
+        redoc_ui_parameters: Annotated[
+            Optional[Dict[str, Any]],
+            Doc(
+                """
+                Parameters to configure ReDoc documentation (by default at `/redocs`).
+                Use kebab-case for [parameters](https://redocly.com/docs/redoc/config).
+                """
+            ),
+        ] = None,
         swagger_ui_oauth2_redirect_url: Annotated[
             Optional[str],
             Doc(
@@ -833,6 +842,7 @@ class FastAPI(Starlette):
         self.root_path_in_servers = root_path_in_servers
         self.docs_url = docs_url
         self.redoc_url = redoc_url
+        self.redoc_ui_parameters = redoc_ui_parameters
         self.swagger_ui_oauth2_redirect_url = swagger_ui_oauth2_redirect_url
         self.swagger_ui_init_oauth = swagger_ui_init_oauth
         self.swagger_ui_parameters = swagger_ui_parameters
@@ -1043,7 +1053,9 @@ class FastAPI(Starlette):
                 root_path = req.scope.get("root_path", "").rstrip("/")
                 openapi_url = root_path + self.openapi_url
                 return get_redoc_html(
-                    openapi_url=openapi_url, title=f"{self.title} - ReDoc"
+                    openapi_url=openapi_url,
+                    title=f"{self.title} - ReDoc",
+                    redoc_ui_parameters=self.redoc_ui_parameters,
                 )
 
             self.add_route(self.redoc_url, redoc_html, include_in_schema=False)
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -448,7 +448,7 @@ class FastAPI(Starlette):
             Optional[Dict[str, Any]],
             Doc(
                 """
-                Parameters to configure ReDoc documentation (by default at `/redocs`).
+                Parameters to configure ReDoc documentation (by default at `/redoc`).
                 Use kebab-case for [parameters](https://redocly.com/docs/redoc/config).
                 """
             ),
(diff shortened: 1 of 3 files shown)


SteinRobert:
I'd be happy to change things based on feedback. Let me know what you guys think :)

SteinRobert:
Should I add some docs as well @svlandeg ?

maliayas:
Great PR. I'd love to see it merged soon. Currently there is no way to customize ReDoc.

fastapi/applications.py
```diff
@@ -444,6 +444,18 @@ async def read_items():
                 """
             ),
         ] = "/redoc",
+        redoc_ui_parameters: Annotated[
+            Optional[Dict[str, Any]],
+            Doc(
+                """
+                Parameters to configure ReDoc documentation (by default at `/docs`).
```
```suggestion
                Parameters to configure ReDoc documentation (by default at `/redoc`).
```

fastapi/applications.py
```diff
@@ -444,6 +444,18 @@ async def read_items():
...
+                Use kebab-case for [parameters](https://redocly.com/docs/redoc/config).
+
+                Read more about it in the
+                [FastAPI docs for Metadata and Docs URLs](https://fastapi.tiangolo.com/tutorial/metadata/#docs-urls).
```
YuriiMotov: The link is not really related

tests/test_local_docs.py
```diff
@@ -65,3 +65,17 @@ def test_google_fonts_in_generated_redoc():
...
+        openapi_url="/docs",
+        title="title",
+        with_google_fonts=False,
+        redoc_ui_parameters={"disable-search": "true"},
+    ).body.decode()
+    assert 'disable-search="true"' in body_with_parameters
+    body_without_parameters = get_redoc_html(
+        openapi_url="/docs", title="title", with_google_fonts=False
+    ).body.decode()
+    assert 'disable-search="true"' not in body_without_parameters
```
YuriiMotov: Better split into two separate tests

YuriiMotov (commented):
LGTM in general

I think this can be merged as is.
We should probably add docs about this feature.

Let's first wait for Sebastián to approve the idea of the feature

SteinRobert:
> Let's wait for Sebastián to approve the idea of feature, then we will need to make a few improvements and add docs

Alright. In the meantime I incorporated your feedback. Thank you very much.

SteinRobert:
Thanks for fixing the typo!

nidhishgajjar:

## Review of PR #12955: Add support for ReDoc parameters

Nice feature addition — bringing ReDoc customization in line with the existing `swagger_ui_parameters` support. The pattern is consistent and the implementation is straightforward.

### What looks good ✅

- **Consistent API**: Follows the same pattern as `swagger_ui_parameters` — Dict parameter on `FastAPI.__init__`, passed through to HTML generation
- **Tests included**: Both with and without parameters cases tested
- **Docstring**: Mentions kebab-case convention which matches ReDoc's API
- **Minimal change**: Clean, focused implementation with no unnecessary changes

### Security observation 🟡

The HTML attribute injection doesn't sanitize values:
```python
for key, value in redoc_ui_parameters.items():
    config_string += f' {key}="{value}"'
```

While this is developer-controlled config (not user input), a value containing `"` could break out of the attribute. For robustness, consider escaping:
```python
from html import escape
config_string += f' {escape(key)}="{escape(str(value))}"'
```

This is low risk since it's server-side config, but `swagger_ui_parameters` avoids this issue by using `JSON.stringify()` in JavaScript. Since ReDoc uses HTML attributes, explicit escaping would be a safer approach.

### Value type handling 💡

F-string conversion means `True` becomes `"True"` (capital T) but ReDoc expects `"true"` (lowercase). Consider converting booleans explicitly:
```python
if isinstance(value, bool):
    value = str(value).lower()
```

Or document that values should be pre-converted to strings.

### Summary

Solid feature addition that addresses a real gap in ReDoc customization. The HTML attribute injection pattern would benefit from value escaping and boolean handling, but the overall approach is clean and consistent with FastAPI's existing patterns.

**Assessment: comment** (HTML attribute escaping, boolean value handling)

tiangolo:

I want to make ReDoc optional, not by default, and make the configuration setup be different, by including a router, that depends on this refactor I have planned.

Thank you! 🍰

SteinRobert:
Makes sense. Thank you for the update. If you manage to write down your ideas in more detail I'm happy to assist with the ReDoc refactor.
Maybe it would even be possible to find a more generic approach to support something like https://scalar.com ?

PR #12956, opened 2024-11-19
⬆️ Bump version of `python-multipart` to 0.0.17

juliangilbey:
The fastapi code now uses the module name `python_multipart`, and this PR changes the dependency in `pyproject.toml` to require a version post-name-change (otherwise the tests fail).


tiangolo:
Thanks! Not sure if it was using only `python_multipart` at some point, but currently it's not: https://github.com/fastapi/fastapi/blob/502c928ab1441f1bf7c1e43e6da9d207d722fca6/fastapi/dependencies/utils.py#L91-L115, so it's probably fine as it is right now. Given that I'll close this one, but thanks! :coffee:

PR #12970, opened 2024-11-22
⬆️ Update mypy to 1.14.1

tamird:
Depends on #12971.


fastapi/encoders.py
```diff
@@ -240,7 +240,7 @@ def jsonable_encoder(
             custom_encoder=encoders,
             sqlalchemy_safe=sqlalchemy_safe,
         )
-    if dataclasses.is_dataclass(obj):
+    if dataclasses.is_dataclass(type(obj)):
```
tiangolo, line 243: Is this needed? :thinking:

tamird: Yep, see the commit message.

tamird: https://github.com/python/typeshed/blob/e1c74f08f12b8bec6afe266951f0756dc1b43ebe/stdlib/dataclasses.pyi#L217-L223

tamird: @tiangolo how can I move this forward?

pyproject.toml
```diff
@@ -121,6 +121,7 @@ source-includes = [
 name = "fastapi-slim"
 
 [tool.mypy]
+plugins = ["pydantic.mypy"]
```
Kludex, line 124: Is this necessary?

Kludex (approved):

pyproject.toml
tamird: This is what enabled the suppressions in fastapi/security/api_key.py to be removed.

tamird:
@Kludex thanks for the approval. Are you able to land this please?

Kludex:
No. I don't merge PRs.

tamird:
@tiangolo could you please have a look? you merged #12971 quickly but perhaps forgot about this one.

YuriiMotov (approved):
LGTM

Maybe we should extract the part related to plugin installation into a separate PR, but doesn't seem critical to me

tamird:
Thank you @YuriiMotov. Are you planning to merge this?

svlandeg (commented):
Ok, so in an attempt to get this merged, I've moved the pydantic.mypy plugin changes to a separate PR #14081. This makes it easier to review which changes are causing what, as several reviewers had questions about what exactly is needed and what relates to what.

For this PR, the goal is simply to update to `mypy` 1.14.1. When doing so, we're getting 2 similar errors:

```
fastapi\encoders.py:244: error: Argument 1 to "asdict" has incompatible type "DataclassInstance | type[DataclassInstance]"; expected "DataclassInstance"  [arg-type]
```
What happens here is that first, we're calling `dataclasses.is_dataclass(obj)`.

Due to [the typing of this method](https://github.com/python/typeshed/blob/e1c74f08f12b8bec6afe266951f0756dc1b43ebe/stdlib/dataclasses.pyi#L217-L223) in their `dataclasses.pyi`, `mypy` will go and type `obj` as `DataclassInstance | type[DataclassInstance]`. This then doesn't align with the signature of the following `dataclasses.asdict(obj)` statement, which expects a `DataclassInstance`.

So, these are the possible fixes for this:
- add an ignore statement
- add a `type()` call before calling `is_dataclass`, as @tamird suggests. This isn't actually necessary, because the implementation of `dataclasses.is_dataclass` will call `type()` when necessary. However, by doing this, we trick `mypy` into assuming that `obj` is NOT a `type[DataclassInstance]` and then it won't error on the following `asdict` call
- inbetween the original `is_dataclass` check and the `asdict` call, add the following: `assert not isinstance(obj, type)`. This again will tell `mypy` not to assume the `obj` is a type. To me, this is the most direct/clear solution, but I think it's a matter of taste.

Will run this by Tiangolo to get his opinion.

tiangolo:
Thanks a lot for the investigation and clarification with all the context @svlandeg! Yep, it makes more sense to me to use `assert` there instead of using `type()` inside of it.

tiangolo (approved):
Nice, thanks! 🚀 

And thanks @svlandeg for the analysis, discussion and work on this! 🙌

PR #11542, opened 2024-05-06
🦺 Update implementation and docs for `RequestValidationError` in Pydantic v2

tamird:
Change the signature of `ValidationException.errors` to return
`Sequence[pydantic_core.ErrorDetails]` in Pydantic V2.

Update the documentation to explain that `RequestValidationError` isn't
literally a subclass since Pydantic V2.

Closes (**UPD YuriiMotov:** actually not) #10424.
Closes #11176.

PR #12972, opened 2024-11-22
🐛 Add missing `ValidationException.__str__` functionality

tamird:
Fixes https://github.com/fastapi/fastapi/discussions/12125.

**UPD YuriiMotov:**
The issue is: the code example [in docs](https://fastapi.tiangolo.com/tutorial/handling-errors/#override-request-validation-exceptions) should show text version
```
1 validation error
path -> item_id
value is not a valid integer (type=type_error.integer)
```
But it shows JSON
```
[{'type': 'int_parsing', 'loc': ('path', 'item_id'), 'msg': 'Input should be a valid integer, unable to parse string as an integer', 'input': 'foo'}]
```

### DIFF
--- a/fastapi/exceptions.py
+++ b/fastapi/exceptions.py
@@ -1,10 +1,23 @@
-from typing import Any, Dict, Optional, Sequence, Type, TypedDict, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Optional,
+    Sequence,
+    Type,
+    TypedDict,
+    Union,
+)
 
 from annotated_doc import Doc
+from fastapi._compat import display_errors
 from pydantic import BaseModel, create_model
 from starlette.exceptions import HTTPException as StarletteHTTPException
 from starlette.exceptions import WebSocketException as StarletteWebSocketException
 from typing_extensions import Annotated
+
+if TYPE_CHECKING:  # pragma: nocover
+    from fastapi._compat import ErrorDict
 
 
 class EndpointContext(TypedDict, total=False):
@@ -164,7 +177,7 @@ class DependencyScopeError(FastAPIError):
 class ValidationException(Exception):
     def __init__(
         self,
-        errors: Sequence[Any],
+        errors: Sequence["ErrorDict"],
         *,
         endpoint_ctx: Optional[EndpointContext] = None,
     ) -> None:
@@ -177,7 +190,7 @@ class ValidationException(Exception):
         self.endpoint_file = ctx.get("file")
         self.endpoint_line = ctx.get("line")
 
-    def errors(self) -> Sequence[Any]:
+    def errors(self) -> Sequence["ErrorDict"]:
         return self._errors
 
     def _format_endpoint_context(self) -> str:
@@ -193,8 +206,7 @@ class ValidationException(Exception):
 
     def __str__(self) -> str:
         message = f"{len(self._errors)} validation error{'s' if len(self._errors) != 1 else ''}:\n"
-        for err in self._errors:
-            message += f"  {err}\n"
+        message += display_errors(self._errors)
         message += self._format_endpoint_context()
         return message.rstrip()
 
@@ -202,7 +214,7 @@ class ValidationException(Exception):
 class RequestValidationError(ValidationException):
     def __init__(
         self,
-        errors: Sequence[Any],
+        errors: Sequence["ErrorDict"],
         *,
         body: Any = None,
         endpoint_ctx: Optional[EndpointContext] = None,
@@ -224,7 +236,7 @@ class WebSocketRequestValidationError(ValidationException):
 class ResponseValidationError(ValidationException):
     def __init__(
         self,
-        errors: Sequence[Any],
+        errors: Sequence["ErrorDict"],
         *,
         body: Any = None,
         endpoint_ctx: Optional[EndpointContext] = None,
--- a/tests/test_tutorial/test_handling_errors/test_tutorial004.py
+++ b/tests/test_tutorial/test_handling_errors/test_tutorial004.py
@@ -1,24 +1,46 @@
-from fastapi.testclient import TestClient
+import inspect
 
-from docs_src.handling_errors.tutorial004 import app
+from fastapi.testclient import TestClient
+from inline_snapshot import snapshot
+
+import docs_src.handling_errors.tutorial004 as tutorial004
+from tests.utils import pydantic_snapshot
+
+app = tutorial004.app
 
 client = TestClient(app)
 
 
 def test_get_validation_error():
+    source_file = inspect.getsourcefile(tutorial004.read_item)
+    line_number = inspect.getsourcelines(tutorial004.read_item)[1]
+    func_name = tutorial004.read_item.__name__
+
     response = client.get("/items/foo")
     assert response.status_code == 400, response.text
-    # TODO: remove when deprecating Pydantic v1
-    assert (
-        # TODO: remove when deprecating Pydantic v1
-        "path -> item_id" in response.text
-        or "'loc': ('path', 'item_id')" in response.text
-    )
-    assert (
-        # TODO: remove when deprecating Pydantic v1
-        "value is not a valid integer" in response.text
-        or "Input should be a valid integer, unable to parse string as an integer"
-        in response.text
+    assert response.text == pydantic_snapshot(
+        v2=snapshot(
+            "\n".join(
+                [
+                    "1 validation error:",
+                    "path -> item_id",
+                    "  Input should be a valid integer, unable to parse string as an integer (type=int_parsing)",
+                    f'  File "{source_file}", line {line_number}, in {func_name}',
+                    "    GET /items/{item_id}",
+                ]
+            )
+        ),
+        v1=snapshot(
+            "\n".join(
+                [
+                    "1 validation error:",
+                    "path -> item_id",
+                    "  value is not a valid integer (type=type_error.integer)",
+                    f'  File "{source_file}", line {line_number}, in {func_name}',
+                    "    GET /items/{item_id}",
+                ]
+            )
+        ),
     )
 
 
--- a/fastapi/_compat/__init__.py
+++ b/fastapi/_compat/__init__.py
@@ -1,3 +1,5 @@
+from typing import TYPE_CHECKING
+
 from .main import BaseConfig as BaseConfig
 from .main import PydanticSchemaGenerationError as PydanticSchemaGenerationError
 from .main import RequiredParam as RequiredParam
@@ -34,6 +36,7 @@ from .may_v1 import CoreSchema as CoreSchema
 from .may_v1 import GetJsonSchemaHandler as GetJsonSchemaHandler
 from .may_v1 import JsonSchemaValue as JsonSchemaValue
 from .may_v1 import _normalize_errors as _normalize_errors
+from .may_v1 import display_errors as display_errors
 from .model_field import ModelField as ModelField
 from .shared import PYDANTIC_V2 as PYDANTIC_V2
 from .shared import PYDANTIC_VERSION_MINOR_TUPLE as PYDANTIC_VERSION_MINOR_TUPLE
@@ -48,3 +51,6 @@ from .shared import (
 from .shared import lenient_issubclass as lenient_issubclass
 from .shared import sequence_types as sequence_types
 from .shared import value_is_sequence as value_is_sequence
+
+if TYPE_CHECKING:  # pragma: nocover
+    from .may_v1 import ErrorDict as ErrorDict
--- a/fastapi/_compat/may_v1.py
+++ b/fastapi/_compat/may_v1.py
@@ -1,5 +1,16 @@
 import sys
-from typing import Any, Dict, List, Literal, Sequence, Tuple, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    List,
+    Literal,
+    Sequence,
+    Tuple,
+    Type,
+    TypedDict,
+    Union,
+)
 
 from fastapi.types import ModelNameMap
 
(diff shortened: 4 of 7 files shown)


svlandeg (commented):
Hi @tamird, thanks for the contribution! Can you add a unit test to this PR that showcases the incorrect behaviour on `master`, while succeeding on this PR? You can probably take inspiration from the original post in https://github.com/fastapi/fastapi/discussions/12125 🙏

tamird:
Hi @svlandeg, turns out there is already a test, it was just too permissive. I have tightened it up and fixed the implementation to make it pass.

tamird:
@svlandeg are you able to merge this please?

tamird:
@alejsdev could you please take a look? This is relatively trivial and contains a test as requested by @svlandeg.

fastapi/exceptions.py
```diff
@@ -153,6 +153,21 @@ def __init__(self, errors: Sequence[Any]) -> None:
     def errors(self) -> Sequence[Any]:
         return self._errors
 
+    def __str__(self) -> str:
+        message = f"{len(self._errors)} validation error"
+        if len(self._errors) != 1:
+            message += "s"
```
tamird, line 159: smokeshow coverage is red because of this line. =/

tamird: It's harder to review because it lacks coverage on this line that pluralizes a string? With all due respect, give me a break. If this one is too hard, please consider reviewing https://github.com/fastapi/fastapi/pull/12970.

svlandeg: I find your comment pretty disrespectful.

I have quite some years of experience in open-source maintenance, including writing, reviewing and merging PRs. Countless times I've seen PRs being submitted that were "trivial" with comments such as just "please just merge". But as soon as tests get properly added (incl. tests that ensure the coverage stays 100%), suddenly little bugs pop up. That's fine, that's a normal dev process, but it's the main reason why we generally ask for adequate tests ànd a passing CI before the team starts reviewing.

I want to iterate that we really appreciate user contributions and PRs. Unfortunately there's a huge load of open PRs (280) and we do our best to manage this queue. If users can help by (also) adding adequate tests, keeping the branch up-to-date wist `master` and ensuring the CI passes, then yes that will certainly help us in the reviewing process.

svlandeg: I will put this PR in draft as long as the CI is failing.

tamird: > I find your comment pretty disrespectful.

I hope you can understand that the disrespect was not the intent. I won't make excuses, you can see for yourself [how long I've been waiting for reviews](https://github.com/fastapi/fastapi/pulls/tamird).

> I have quite some years of experience in open-source maintenance, including writing, reviewing and merging PRs. Countless times I've seen PRs being submitted that were "trivial" with comments such as just "please just merge". But as soon as tests get properly added (incl. tests that ensure the coverage stays 100%), suddenly little bugs pop up. That's fine, that's a normal dev process, but it's the main reason why we generally ask for adequate tests ànd a passing CI before the team starts reviewing.

I too have quite some years of experience in open-source. I understand, of course, that you want to keep things healthy and coverage is part of that. But requests of this kind inherently require trust between the author and the reviewer; you ask that I invest more time _in advance_ of you investing _any_ time. Coupled with the prior experience of waiting for reviews for the past 7 months, this trust has not been earned. [Other contributors](https://github.com/fastapi/fastapi/pulls?q=is%3Apr+is%3Aopen+sort%3Acomments-asc) have not fared much better.

> I want to iterate that we really appreciate user contributions and PRs. Unfortunately there's a huge load of open PRs (280) and we do our best to manage this queue. If users can help by (also) adding adequate tests, keeping the branch up-to-date wist `master` and ensuring the CI passes, then yes that will certainly help us in the reviewing process.

Sure. I have now addressed your request. Thank you.

tamird:
Hello! This is ready for review.

tiangolo:
@tamird there's no need to tag individual people nor to keep posting comments.

Your aggressive and demeaning attitude and comments towards others here are not well received and are not appreciated.

This is open source, MIT licensed, free. The result of any work from others here is received by you and/or your employer for free. Neither you nor your employer (it seems Google, from your GitHub) are paying anything to receive this open source project or our work for free. You actually have no particular right to demand reviews, merges, timelines, etc.

I request that you abstain from adding unrequested comments, in particular tagging others or with the only intention of pressing us to give a higher priority to reviewing your PRs, handling your needs, or reporting a status on the process of handling them. In particular, I request you to abstain from making any comment that could be in any way interpreted as demeaning towards others.

tamird:
@tiangolo your points are well taken. The last comment here was 5 weeks ago.

tests/test_tutorial/test_handling_errors/test_tutorial004.py
```diff
@@ -8,18 +9,20 @@
...
-        or "Input should be a valid integer, unable to parse string as an integer"
-        in response.text
-    )
+    if PYDANTIC_VERSION_MINOR_TUPLE < (2, 0):
+        assert (
+            response.text
+            == """1 validation error
+path -> item_id
+  value is not a valid integer (type=type_error.integer)"""
+        )
```
YuriiMotov:
```suggestion
        assert response.text == (
            "1 validation error\n"
            "path -> item_id\n"
            "  value is not a valid integer (type=type_error.integer)"
        )
```

tests/test_tutorial/test_handling_errors/test_tutorial004.py
```diff
@@ -8,18 +9,20 @@
...
+    else:
+        assert (
+            response.text
+            == """1 validation error
+path -> item_id
+  Input should be a valid integer, unable to parse string as an integer (type=int_parsing)"""
+        )
```
YuriiMotov:
```suggestion
        assert response.text == (
            "1 validation error\n"
            "path -> item_id\n"
            "  Input should be a valid integer, unable to parse string as an integer "
            "(type=int_parsing)"
        )
```
I think this is more readable

YuriiMotov (changes requested):
@tamird, thanks for your contribution!

I have a couple of things to discuss:

1) Current implementation also changes the output in case of Response validation error:

Before
```
fastapi.exceptions.ResponseValidationError: 2 validation errors:
  {'type': 'int_parsing', 'loc': ('response', 'item_id'), 'msg': 'Input should be a valid integer, unable to parse string as an integer', 'input': 'sss'}
  {'type': 'missing', 'loc': ('response', 'name'), 'msg': 'Field required', 'input': {'item_id': 'sss'}}
```

After
```
fastapi.exceptions.ResponseValidationError: 2 validation errors
response -> item_id
  Input should be a valid integer, unable to parse string as an integer (type=int_parsing)
response -> name
  Field required (type=missing)
```

I think we should only change `RequestValidationError` here.
Proposed implementation hides `input` field that might be useful for debug reasons.

2) What if we import `display_errors` from pydantic instead of duplicating it here?

```python
from pydantic.version import VERSION as PYDANTIC_VERSION

if PYDANTIC_VERSION.startswith("2."):
    from pydantic.v1.error_wrappers import ErrorDict, display_errors
else:
    from pydantic.error_wrappers import (  # type: ignore[no-redef]
        ErrorDict,
        display_errors,
    )
```
In this case we also need to cast `errors` to `list[ErrorDict]` in `f'{display_errors(errors)}'`:
```python
    def __str__(self) -> str:
        errors = self.errors()
        no_errors = len(errors)
        return (
            f'{no_errors} validation error{"" if no_errors == 1 else "s"}\n'
            f'{display_errors(cast(List[ErrorDict], errors))}'
        )
```
I checked the code above and it passes tests.
I can push these changes if you don't mind

tamird:
@YuriiMotov I've implemented the changes you suggested. Note that there is no test coverage for the `ResponseValidationError.__str__` behavior you asked to preserve; you might consider adding some in another PR.

I had to change your code a bit because `ErrorDict` is defined iff `TYPE_CHECKING`.

PTAL.

fastapi/exceptions.py
```diff
@@ -159,6 +176,18 @@ def __init__(self, errors: Sequence[Any], *, body: Any = None) -> None:
         super().__init__(errors)
         self.body = body
 
+    def __str__(self) -> str:
+        errors: List[ErrorDict]
+        if TYPE_CHECKING:
+            errors = cast(List[ErrorDict], self.errors())
+        else:
+            errors = self.errors()
```
YuriiMotov:
```suggestion
        errors = cast(List[ErrorDict], self.errors())
```
Cast has very low performance impact at runtime.
Considering this part of code is not executed very frequently, I think we can avoid using `if TYPE_CHECKING` here.

fastapi/exceptions.py
```diff
@@ -1,10 +1,27 @@
...
 from typing_extensions import Annotated, Doc
 
+PYDANTIC_VERSION_MINOR_TUPLE = tuple(int(x) for x in PYDANTIC_VERSION.split(".")[:2])
+PYDANTIC_V2 = PYDANTIC_VERSION_MINOR_TUPLE[0] == 2
+
+if PYDANTIC_V2:
+    from pydantic.v1.error_wrappers import display_errors
+
+    if TYPE_CHECKING:
+        from pydantic.v1.error_wrappers import ErrorDict
```
YuriiMotov: Mypy doesn't show me any issues if I just import it without `if TYPE_CHECKING` and tests pass.
Are you sure we need this?

fastapi/_compat.py
```diff
@@ -19,17 +19,20 @@
     cast,
 )
 
-from fastapi.exceptions import RequestErrorModel
+from fastapi.exceptions import (
+    PYDANTIC_V2 as PYDANTIC_V2,
+)
+from fastapi.exceptions import (
+    PYDANTIC_VERSION_MINOR_TUPLE as PYDANTIC_VERSION_MINOR_TUPLE,
+)
```
YuriiMotov: I think there is no need in this. We can just use `if PYDANTIC_VERSION.startswith("2.")` in `fastapi.exceptions`

YuriiMotov (changes requested):
Thanks for quick response!

The problem with coverage is not related to `ResponseValidationError.__str__`, but to `RequestValidationError.__str__` and top-level code in `exceptions` module.
https://smokeshow.helpmanual.io/3w5t2y4p045l145e0y0d/function_index.html

I think that problems are with `if TYPE_CHECKING` statements.
Do you think they are necessary? I think we can simplify code by removing all these `if TYPE_CHECKING`. I don't see any issues without them locally

fastapi/_compat.py
tamird: Why is that better? That amounts to duplication. This is just moving from `_compat.py` because that file depends on this one.

fastapi/exceptions.py
tamird: Mypy is lying to you. From mypy's perspective, `TYPE_CHECKING` is always true.

See https://github.com/pydantic/pydantic/blob/ffc084053b1ed87927c4c10e6dc2d17f6a4f2cac/pydantic/v1/error_wrappers.py#L7-L22

This type is only defined `if TYPE_CHECKING`.

fastapi/exceptions.py
tamird: This isn't about performance. The type `ErrorDict` only exists `if TYPE_CHECKING`. See https://github.com/pydantic/pydantic/blob/ffc084053b1ed87927c4c10e6dc2d17f6a4f2cac/pydantic/v1/error_wrappers.py#L7-L22.

tamird (commented):
> I think that problems are with if TYPE_CHECKING statements.
Do you think they are necessary? I think we can simplify code by removing all these if TYPE_CHECKING. I don't see any issues without them locally

Did you actually run the tests? Importing `ErrorDict` unconditionally crashes, see my individual responses.

Thank you for the review.

fastapi/_compat.py
YuriiMotov: I just think that `_compat` means "compatibility" and this is the right place where things like `PYDANTIC_V2` should be defined and imported from.
For me it just seems logically wrong to place these things in `exceptions`.

I understand we cant import `PYDANTIC_V2` to `exceptions` from `_compat` and I suggest using just `if PYDANTIC_VERSION.startswith("2.")` there. We don't need `PYDANTIC_VERSION_MINOR_TUPLE` in `exceptions`.
This way we also minimize diff of this PR.

If you are still not convinced, let's keep it as it is for now

tamird: Yeah, I agree that it _should_ be in `_compat` - and it is still re-exported from there. That's why all the other imports of this constant didn't have to change.

I'd prefer to keep it as-is. Thanks for meeting me halfway.

fastapi/exceptions.py
YuriiMotov: I apologize - you are right here.
It seems I messed things up while jumping between tasks.. I was quite sure I had tested it, but apparently I didn’t.
Sorry for wasting your time with this 🙏

fastapi/exceptions.py
YuriiMotov: Got it, thanks! 🙏
We can simplify it this way (this time I'm sure I tested it :) ):
```python
errors = cast(List["ErrorDict"], self.errors())
```

tamird: Looks like that does work. Done.

YuriiMotov (approved):
I still don't like that `PYDANTIC_V2` and `PYDANTIC_VERSION_MINOR_TUPLE` are now defined in `exceptions` and imported to `_compat`.
I think we can avoid this and just use `if PYDANTIC_VERSION.startswith("2."):` in `exceptions` module.
But author prefers to keep it this way.

Except for the moment mentioned above, everything is good!

@tamird, thanks for your patience!

tamird:
I think the proper thing to do is just to avoid importing exceptions from compat. That would require moving a bit of code from company into exceptions. Would you like me to try that?

tamird:
@YuriiMotov FYI I decided to hold off on #11542 until this is merged. Let me know how you'd like to proceed.

YuriiMotov:
> I think the proper thing to do is just to avoid importing exceptions from compat. That would require moving a bit of code from company into exceptions. Would you like me to try that?

As I said, I think we should minimize the diff, so that Sebastian, when review this PR, would have as few questions as possible.
I would just use `if PYDANTIC_VERSION.startswith("2."):` in `exceptions` module. If you have better idea - let's try it, but keep in mind the idea of minimizing the diff (especially unnecessary and questionable changes)

tamird:
@YuriiMotov looks like the controversial changes were easy to remove now that #14168 has landed.

YuriiMotov:
Looks like it's not so easy now as we run tests on Python 3.12 and Pydantic V1 is not supported there..

tamird:
> Looks like it's not so easy now as we run tests on Python 3.12 and Pydantic V1 is not supported there..

Just needed a little copy & paste.

YuriiMotov:
@tamird, thanks for resolving the merge conflict!

If I recall correctly, previously we switched to using existing `ErrorDict` and `display_errors` (import them from `pydantic.v1`) for simplicity. But now it turned out we can't just use them and have to duplicate that code.
Would it simplify the code if we switch back to just using our own implementations?

tamird:
> @tamird, thanks for resolving the merge conflict!
> 
> If I recall correctly, previously we switched to using existing `ErrorDict` and `display_errors` (import them from `pydantic.v1`) for simplicity. But now it turned out we can't just use them and have to duplicate that code. Would it simplify the code if we switch back to just using our own implementations?

Sure, it would probably be simpler. However this PR has now been open for over a year, and there's no indication that there's a desire to merge it. I hope you understand that I'm reluctant to invest further effort at this time without some indication that it will be accepted (or rejected).

PR #12976, opened 2024-11-22
Add note for async fixtures

invisibleroads:


graingert:
This isn't correct. The best approach is to move global objects that bind the event loop into lifespan_context

tiangolo:
As this is about encountering a specific type of error, the best place would probably be in a GitHub Discussion. I'll pass on this for now, but thanks for the interest! :coffee:

invisibleroads:
> This isn't correct. The best approach is to move global objects that bind the event loop into lifespan_context

@graingert Are you suggesting that instead of creating instances as fixtures, we should just put those instances in the app fixture?

```
OLD
@pytest.fixture(scope='session')
async def app():
    async with LifespanManager(main) as manager:
        yield manager.app

@pytest.fixture(scope='session')
async def user():
    # ...
    yield user
    # ...

NEW
@pytest.fixture
async def app():
    async with LifespanManager(main) as manager:
        app = manager.app
        app.user = await make_user()
        yield app
```

graingert:
It's not related to the fixture, it's how you define your create_app

invisibleroads:
@graingert

I figured out that the problem was where redis was being initialized. Initially, I was initializing redis (from redis.asyncio import Redis; redis = Redis()) in a global namespace.

By postponing the redis initialization to happen inside of the lifespan context instead, I was able to resolve the "Event loop is closed" errors disappear! Not only that, but I am able to use async fixtures again.

The main point is that async objects should be initialized inside the lifespan context.

```
@asynccontextmanager                                                                                                                              
async def get_lifespan(application):                                                                                                              
    await initialize_cache()      
    # ...                                                                                                                         
    yield
    # ...                                                                                                                         
```

Thanks for the clue.

PR #12997, opened 2024-11-27
✨ Replace `dict` by `Mapping` on `HTTPException.headers`

rijenkii:
HTTP Headers cannot be represented by a `dict`, as they can have duplicate entries, as shown in the example: two `set-cookie` headers:
```python
from pprint import pprint
import fastapi

response = fastapi.Response()
response.delete_cookie("cookie1")
response.delete_cookie("cookie2")

pprint(list(response.headers.items()))
# [('content-length', '0'),
#  ('set-cookie',
#   'cookie1=""; expires=Thu, 07 Nov 2024 08:26:12 GMT; Max-Age=0; Path=/; '
#   'SameSite=lax'),
#  ('set-cookie',
#   'cookie2=""; expires=Thu, 07 Nov 2024 08:26:12 GMT; Max-Age=0; Path=/; '
#   'SameSite=lax')]

headers = fastapi.datastructures.Headers(
    raw=[(b"set-cookie", x.encode()) for x in response.headers.getlist("set-cookie")]
)

exc = fastapi.HTTPException(400, headers=headers)
# Argument of type "Headers" cannot be assigned to parameter "headers" of type "Dict[str, str] | None" in function "__init__"
#   Type "Headers" is not assignable to type "Dict[str, str] | None"
#     "Headers" is not assignable to "Dict[str, str]"
#     "Headers" is not assignable to "None"PylancereportArgumentType
```

As such type of `fastapi.HTTPException`'s `headers` field should be `typing.Mapping` instead of `dict`.

In the example I tried assigning `fastapi.datastructures.Headers` to the parameter `headers`. It worked at runtime, but pyright is screaming about wrong types in the editor and CLI.

`fastapi.Request`'s `headers` field already has the type `MutableHeaders`, which is a distant descendant of `typing.Mapping` already.

A fix has also been submitted to starlette and released as part of 0.41.3, which is why its version is also bumped.

See also:
- https://github.com/fastapi/fastapi/discussions/12786
- https://github.com/encode/starlette/pull/2749


Kludex:
Does it need the starlette bump tho?

rijenkii:
I have bumped starlette to `>=0.41.3` in this PR (https://github.com/fastapi/fastapi/pull/12997/commits/fca6d2dbe80cae20cc8872accb8da5801254fd7d), if needed I can separate it into a different PR.

rijenkii:
Yes, otherwise this typing error happens in fastapi source:
image

fastapi/exceptions.py
```diff
@@ -54,7 +54,7 @@ def __init__(
             ),
         ] = None,
         headers: Annotated[
-            Optional[Dict[str, str]],
+            Optional[Mapping[str, str]],
```
youneshenniwrites, line 57: `Mapping` (like `Dict`) doesn't inherently support duplicate keys. This change wouldn't solve the problem. To handle this scenario properly, I suggest replacing 

`Optional[Mapping[str, str]]` 

with 

`Optional[List[Tuple[str, str]]]`

Let me know your thoughts.

youneshenniwrites (changes requested):

fastapi/exceptions.py
rijenkii: `Mapping` was chosen because starlette already uses it in every place where a generic header type is required:
- `starlette.datastructures.Headers` inherits from `Mapping[str, str]` and receives an optional `headers: Mapping[str, str]` in its `__init__`: https://github.com/encode/starlette/blob/8a99adfb5839b37efee33f2f49c1ba27a954fcbd/starlette/datastructures.py#L490-L500
- `starlette.responses.Response` also receives `Mapping[str, str]` in its `__init__`: https://github.com/encode/starlette/blob/8a99adfb5839b37efee33f2f49c1ba27a954fcbd/starlette/responses.py#L36

Also, `Mapping` is commonly used as a base class for such containers with non-unique keys:
- httpx: https://github.com/encode/httpx/blob/26d48e0634e6ee9cdc0533996db289ce4b430177/httpx/_types.py#L44-L50 (also features `Sequence[Tuple[str, str]]` though)
- requests: https://github.com/python/typeshed/blob/b0768596d2f7b2c12aa61d8a9714325284410b4b/stubs/requests/requests/api.pyi#L9
- django: https://github.com/django/django/blob/5e998d717f7b4220a1728bd49b66ca0162e2a6cb/django/http/request.py#L468 (`CaseInsensitiveMapping` inherits from `Mapping`)

I think being compatible with starlette is an important reason in and of itself.

YuriiMotov (approved):
LGTM!

I think we can skip pinning Starlette version - we only run CI with latest Starlette version, and everything will work in runtime even with older version (since we only changed type hint)

youneshenniwrites (approved):

rijenkii:
> I think we can skip pinning Starlette version

Reverted the minimum Starlette version bump.

tiangolo (approved):
Thanks! This will be available in FastAPI 0.128.7, released in the next few hours. 🚀

PR #13011, opened 2024-11-30
✏️ Fix typo in README installation instructions

dave-hay:
Fixed a missing word in the installation instructions:

Before: "it comes the standard group"
After: "it comes with the standard group"

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,7 +459,7 @@ FastAPI depends on Pydantic and Starlette.
 
 ### `standard` Dependencies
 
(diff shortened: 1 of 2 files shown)


hamzaaitbenyissa (approved):

svlandeg (approved):
Thanks, good catch! This was also wrong in the original `index.md` file, so I added that fix as well.

alejsdev:
Good catch! Thank you @dave-hay :mag: 
And thanks for the update @svlandeg :smile_cat:

PR #13013, opened 2024-11-30
Replace depricated functions for pydantic 2.5+

Abdujabbar:
Hey, hope you are doing well. 

I have seen that in the requirements, we already use the pydantic version 2.5.3+, and as in these documentation I found that, for models, instead of model.dict(...) we should use model_dump, and instead of  attribute __fields__ we should use model_fields, and instead of update_forward_refs we should use model_rebuild function. So, I added these changes in source where I found them. 

Here is the links for details: 

https://docs.pydantic.dev/2.5/concepts/models/#rebuild-model-schema

https://docs.pydantic.dev/2.5/concepts/serialization/#modelmodel_dump

It's just minor changes, don't know if it matters to add or not, but I decided maybe it's better to change deprecated functions.

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -442,12 +442,12 @@ else:
         return use_errors
 
     def _model_rebuild(model: Type[BaseModel]) -> None:
-        model.update_forward_refs()
+        model.model_rebuild()
 
     def _model_dump(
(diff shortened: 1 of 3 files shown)


Abdujabbar:
Argh, it looks we can broke backward compatibility if we can change this. I think, maybe it's not time to change these, if so then I will close these MR.

PR #13037, opened 2024-12-04
📝 Add another example of an exception handler for internal server errors

NanoNulla:
The PR is meant to add another example to the custom exception handler for internal server errors.
It was inspired by https://github.com/fastapi/fastapi/discussions/4798

The PR was open together with @SamueleRomano

Thanks


docs_src/handling_errors/tutorial004.py
```diff
@@ -11,6 +12,19 @@ async def http_exception_handler(request, exc):
...
+        status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+        content=jsonable_encoder({"message": "Internal Server Error"}),
+    )
+
+
+@app.exception_handler(RequestValidationError)
+async def validation_exception_handler(request, exc):
+    return PlainTextResponse(str(exc), status_code=400)
+
+
```
svlandeg, line 27: This part of the file got duplicated by accident and should be removed again.
```suggestion
```

YuriiMotov (changes requested):
@CyberMullah, thanks for the interest in FastAPI!

As I understand it, the idea is to show how developers can override exception handler for server error.

For now changes brake the highlights for the modified code example.
image

Also, we should add an explanation to this code.

@CyberMullah, would you like to work on it?

PR #13061, opened 2024-12-12
📝 Change the word "unwrap" to "unpack" in `docs/en/docs/tutorial/extra-models.md`

timothy-jeong:
While 'unwrap' (or 'unwrapping') is a Pydantic-specific convention, its usage in the documentation might lead to confusion. The document mentions 'unwrap' with the prefix 'Python', which could incorrectly imply that 'unwrap' is official Python terminology.

The behavior of 'unwrap' described in the tutorial/extra-models document aligns with the "dictionary unpacking" feature in Python dictionary displays (https://docs.python.org/3/reference/expressions.html#dictionary-displays).

Therefore, if the intended behavior is indeed Python's dictionary unpacking, I suggest replacing 'unwrap' with 'unpack' for clarity and consistency with official Python terminology.


hard-coders (approved):
Sounds reasonable.

svlandeg (approved):
Makes sense to me!

tiangolo (approved):
Makes sense, thanks! :coffee:

PR #13100, opened 2024-12-21
🐛 Replace underscores in parameter fields to ensure they get recognized correctly

goehls:
**When using the header parameter model, the parameter fields displayed in the swagger-ui interface documentation are incorrect, which leads to the inability to use the swagger-ui documentation for interface testing.**
使用header参数模型时，swagger-ui接口文档展示的参数字段错误，从而导致无法使用swagger-ui文档进行接口测试。

```py
class CommonHeaderSchema(BaseModel):
    main_account_id: int
    shop_id: int
    merchant_id: Optional[int] = None

async def common_header_parameters(headers: Annotated[CommonHeaderSchema, Header()]):
    return headers

router = APIRouter(dependencies=[Depends(common_header_parameters)])

@router.post("/create_auto_operation_config")
def create_auto_operation_config(region_item_config_id: int):
    return ResponseInfo.success(data=auto_config_service.create_auto_operation_config(region_item_config_id))
```
**If there is a CommonHeaderSchema, the swagger-ui interface documentation is displayed as shown in the figure.**
如存在一个CommonHeaderSchema，swagger-ui接口文档展示的如图所示：
image
**The reasons are as follows: Since the swagger-ui documentation does not perform conversion on fields with underscores (converting underscores to dashes), fields with underscores cannot be recognized.**
原因如下：由于swagger-ui文档未对带有下划线的字段做转换（下划线转换为破折号），导致无法识别带下划线的字段。

after bug fix
代码修复后：
image

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -713,7 +713,9 @@ def _validate_value_with_model_field(
         return v_, []
 
 
-def _get_value(field: ModelField, values: Mapping[str, Any], alias: Union[str, None] = None) -> Any:
+def _get_value(
+    field: ModelField, values: Mapping[str, Any], alias: Union[str, None] = None
+) -> Any:
     if is_sequence_field(field) and isinstance(values, (ImmutableMultiDict, Headers)):
         value = values.getlist(alias)
     else:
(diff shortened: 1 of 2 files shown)


bask209:
So, after reading this, I removed the underscores from my headers just to test and it now can read the headers as needed.

Anyone got any more info on this?

svlandeg (commented):
It looks like this change is making the test suite fail, so that will have to be resolved before we can properly review this. I'll put the PR in draft in the meantime.

YuriiMotov:
@goehls, thank you for your efforts!

This was fixed in 0.115.12.

PR #13118, opened 2024-12-29
🌐 Update existing Spanish translations using the new LLM-assisted system using PydanticAI

tiangolo:
🌐 Update existing Spanish translations using the new LLM-assisted system using PydanticAI

I manually reviewed all the text and tweaked the prompts and then the content to ensure it was okay.

Then @alejsdev reviewed it all as well to give it another check and add some extra tweaks.

PR #13122, opened 2024-12-30
🌐 Add new Spanish translations for all docs with new LLM-assisted system using PydanticAI

tiangolo:
🌐 Add new Spanish translations for all docs with new LLM-assisted system using PydanticAI

I manually read everything, tweaked the prompts, regenerated the content a few times, and then tweaked the result to ensure it was all okay.

Then @alejsdev also reviewed that everything looked visually okay.

Depends on https://github.com/fastapi/fastapi/pull/13118

### DIFF
--- a/docs/en/docs/release-notes.md
+++ b/docs/en/docs/release-notes.md
@@ -17,6 +17,7 @@ hide:
 
 ### Translations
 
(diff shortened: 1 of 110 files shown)


alejsdev (approved):
LGTM in general :smile:

PR #13135, opened 2025-01-01
👷‍♀️ Add script for GitHub Topic Repositories and update External Links

alejsdev:
👷‍♀️ Add script for GitHub Topic Repositories and update External Links

### DIFF
--- a/docs/en/docs/js/custom.js
+++ b/docs/en/docs/js/custom.js
@@ -1,25 +1,3 @@
-const div = document.querySelector('.github-topic-projects')
-
-async function getDataBatch(page) {
-    const response = await fetch(`https://api.github.com/search/repositories?q=topic:fastapi&per_page=100&page=${page}`, { headers: { Accept: 'application/vnd.github.mercy-preview+json' } })
-    const data = await response.json()
-    return data
-}
-
-async function getData() {
-    let page = 1
-    let data = []
-    let dataBatch = await getDataBatch(page)
-    data = data.concat(dataBatch.items)
-    const totalCount = dataBatch.total_count
-    while (data.length < totalCount) {
-        page += 1
-        dataBatch = await getDataBatch(page)
-        data = data.concat(dataBatch.items)
-    }
-    return data
-}
-
 function setupTermynal() {
     document.querySelectorAll(".use-termynal").forEach(node => {
         node.style.display = "block";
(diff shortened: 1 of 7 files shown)


docs/en/docs/js/custom.js
```diff
@@ -158,20 +136,6 @@ async function showRandomAnnouncement(groupId, timeInterval) {
...
-        const ul = document.querySelector('.github-topic-projects ul')
-        data.forEach(v => {
-            if (v.full_name === 'fastapi/fastapi') {
-                return
-            }
-            const li = document.createElement('li')
-            li.innerHTML = `<a href="${v.html_url}" target="_blank">★ ${v.stargazers_count} - ${v.full_name}</a> by <a href="${v.owner.html_url}" target="_blank">@${v.owner.login}</a>`
-            ul.append(li)
-        })
-    }
```
tiangolo, line 173: Nice! I wanted to get rid of this. :fire:

scripts/topic_repos.py
```diff
@@ -0,0 +1,82 @@
...
+    logging.info("Finished")
+
+    # Local development
+    # repos_path = Path("../docs/en/data/topic_repos.yml")
+    repos_path = Path("./docs/en/data/topic_repos.yml")
+    repos_old_content = repos_path.read_text(encoding="utf-8")
+    new_repos_content = yaml.dump(data, sort_keys=False, width=200, allow_unicode=True)
+    if repos_old_content == new_repos_content:
+        logging.info("The data hasn't changed. Finishing.")
+        sys.exit(0)
```
tiangolo, line 58:
```suggestion
        return
```

I see this was from a copy of the other scripts that were run with `if __name__ == "__main__"`, but as this is now in a function, we can just return to finish the execution here, we don't have to call `sys.exit(0)`

scripts/topic_repos.py
```diff
@@ -0,0 +1,82 @@
...
+            Repo(
+                name=repo.name,
+                html_url=repo.html_url,
+                stars=repo.stargazers_count,
+                owner_login=repo.owner.login,
+                owner_html_url=repo.owner.html_url,
+            )
+        )
+    data = [repo.model_dump() for repo in final_repos]
+    logging.info("Finished")
```
tiangolo, line 49:
```suggestion
```

This is probably an old copy-paste log, by this point it's not finished yet. :sweat_smile:

There's already a log with "Finished" at the end so we can just remove this one.

tiangolo (commented):
Nice! :rocket: 

Just a couple of nitpicks. :nerd_face:

scripts/topic_repos.py
alejsdev: My bad :sweat_smile:

tiangolo (approved):
Nice! :rocket: :sunglasses:

PR #13155, opened 2025-01-04
✏️ Update Strawberry integration docs

kinuax:
Docs regarding `Strawberry` integration are updated with the currently proper way. I also propose to remove the websocket reference because it is not needed as part of a minimum working example.


alejsdev (approved):

patrick91 (approved):

alejsdev:
Nice, thank you! @kinuax :smile: :100:

PR #13157, opened 2025-01-04
Fix ruff B039 warning: avoid using mutable default values for ContextVar

nillvitor:
Fix B039 warning: avoid using mutable default values for ContextVar.

This fix addresses the B039 warning in ruff version 0.8.4. It is necessary to allow upgrading the ruff version in the requirements to 0.8.4, which previously failed in the tests for pull request #13131.


alejsdev:
Hello @nillvitor, thanks for your interest in contributing. However, this is no longer necessary since the Peewee tutorial was removed a while ago. These files were already removed in another PR, but thanks for your effort! :raised_hands: :sparkles:

PR #13162, opened 2025-01-07
fixed incorrect typehint for username (possible None) in 004 and 005 security tutorials

AlexeyTarasov77:


alv2017 (commented):
1) I think that it is not a good idea to mix code from different tutorials.  I would prefer to work with tutorials source code per tutorial, providing a link to the tutorial itself.

2) I do not support this change. I do not see any problem with defining a username to be a string, and actually this is the preferred solution. If we are using typing, then we should give a preference to a stricter version of the code whenever it is possible. All in all my preference would be to stick with the existing version of the code.

3) Nevertheless, I think that there is still a bug, and it lives in `get_current_user()` function. 
The `username` will never take a value of None, however it can be an empty string, and we are not handling this in our code. 

```
async def get_current_user(token: str = Depends(oauth2_scheme)):
        headers={"WWW-Authenticate": "Bearer"},
    )
    try:
        username: str = payload.get("sub")
        if username is None:
            raise credentials_exception
        token_data = TokenData(username=username)
    except InvalidTokenError:
        raise credentials_exception
    user = get_user(fake_users_db, username=token_data.username)
    if user is None:
        raise credentials_exception
    return user
```

AlexeyTarasov77:
But if we define username as just string, check for username is None seems to be unreachable, isn't it?

alv2017:
> But if we define username as just string, check for username is None seems to be unreachable, isn't it?

`username` is None is not reachable, because of the logic implemented in ` get_current_user()` function, and not because of type defined for the variable `username` (recall that Python is a dynamically typed language)

AlexeyTarasov77:
Ok, closing

PR #13163, opened 2025-01-07
📝 Update async tests and documentation

oleg-korshunov:


docs/en/docs/advanced/async-tests.md
```diff
@@ -54,46 +59,31 @@ $ pytest
...
 /// tip
 
-If you encounter a `RuntimeError: Task attached to a different loop` when integrating asynchronous function calls in your tests (e.g. when using <a href="https://stackoverflow.com/questions/41584243/runtimeerror-task-attached-to-a-different-loop" class="external-link" target="_blank">MongoDB's MotorClient</a>), remember to instantiate objects that need an event loop only within async functions, e.g. an `'@app.on_event("startup")` callback.
+If you encounter a `RuntimeError: Task attached to a different loop` when integrating asynchronous function calls in your tests, you can override the default pytest event loop using the following fixture:
+
+```python
+@pytest.fixture(scope="session")
+def event_loop() -> Generator[AbstractEventLoop, None, None]:
+    """Overrides pytest default function scoped event loop"""
+    policy = asyncio.get_event_loop_policy()
```
graingert, line 83: event loop policy is deprecated in 3.14 and will be removed

oleg-korshunov: corrected description 
image

docs/en/docs/advanced/async-tests.md
```diff
@@ -54,46 +59,31 @@ $ pytest
...
+As the testing function is now asynchronous, you can now also call (and await) other async functions apart from sending requests to your FastAPI application in your tests, exactly as you would call them anywhere else in your code.
 
 /// tip
 
-If you encounter a `RuntimeError: Task attached to a different loop` when integrating asynchronous function calls in your tests (e.g. when using <a href="https://stackoverflow.com/questions/41584243/runtimeerror-task-attached-to-a-different-loop" class="external-link" target="_blank">MongoDB's MotorClient</a>), remember to instantiate objects that need an event loop only within async functions, e.g. an `'@app.on_event("startup")` callback.
+If you encounter a `RuntimeError: Task attached to a different loop` when integrating asynchronous function calls in your tests, you can override the default pytest event loop using the following fixture:
+
+```python
+@pytest.fixture(scope="session")
+def event_loop() -> Generator[AbstractEventLoop, None, None]:
```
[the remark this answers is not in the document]
oleg-korshunov: corrected description 
image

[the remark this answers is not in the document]
graingert-coef: This advice is wrong, the existing advice was correct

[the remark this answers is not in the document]
graingert: This advice is wrong, the existing advice was correct

docs/en/docs/advanced/async-tests.md
graingert: The deprecated event loop policy is still being used here

oleg-korshunov: corrected it to asyncio.Runner

docs/en/docs/advanced/async-tests.md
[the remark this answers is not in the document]
oleg-korshunov: Returned the old description

docs/en/docs/advanced/async-tests.md
```diff
@@ -54,46 +59,34 @@ $ pytest
...
 
-If your application relies on lifespan events, the `AsyncClient` won't trigger these events. To ensure they are triggered, use `LifespanManager` from <a href="https://github.com/florimondmanca/asgi-lifespan#usage" class="external-link" target="_blank">florimondmanca/asgi-lifespan</a>.
+```python
+import asyncio
+import pytest
+from collections.abc import Generator
 
-///
+@pytest.fixture(scope="session")
+def event_loop() -> Generator[asyncio.AbstractEventLoop, None, None]:
```
graingert, line 85: This just scopes the same loop for the whole session, each test function should get a fresh loop. The only way to fix this issue is to fix the underlying code that uses a global database connection

oleg-korshunov: Yes, you are absolutely right, I specifically used this option as it is simpler, I made the code in accordance with your review, and also added a note about a simpler setup

docs/en/docs/advanced/async-tests.md
```diff
@@ -54,46 +59,40 @@ $ pytest
...
 
-/// tip
+```python
+import asyncio
+import pytest
+from collections.abc import Generator
 
-Note that we're using async/await with the new `AsyncClient` - the request is asynchronous.
+@pytest.fixture(scope="function")
+def event_loop() -> Generator[asyncio.AbstractEventLoop, None, None]:
```
YuriiMotov, line 85: Not sure about this part.
We recommend using `pytest.mark.anyio`, but this fixture is tied to `asyncio` (`pytest-asyncio`?).

YuriiMotov (commented):
@oleg-korshunov, thank you for your interest and efforts!

I think we shouldn't replace existing code example with more complex code example with `lifespan`.
There is already a warning about `lifespan` events and `LifespanManager`. And, I personally think this is enough. 
But if we want to add code example, I think we should add it after explaining basic simple example. This iterative approach will make it simpler for beginners.

Could you please explain why you decided not to use the approach with `LifespanManager`?

Also, we should try to minimize unnecessary changes. This would increase chances for this PR to be merged.

Issue #13150, opened 2025-01-03
Simplify tests for variants


tiangolo:
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

## Summary

Simplify tests for variants, from multiple test files (one test file per variant) to a single test file with parameters to test each variant.

## Background

Currently, we have multiple source example variants for different Python versions:

* Python 3.8
* Python 3.9
* Python 3.10

And we have versions using `Annotated` and without using it.

Combining that, for each source app, we end up with different variants.

For example, for `docs_src/query_params_str_validations/tutorial010.py`, for this same `tutorial010`, we have these variants:

* `docs_src/query_params_str_validations/tutorial010_an_py39.py`
    * Using `Annotated`, Python 3.9.
* `docs_src/query_params_str_validations/tutorial010_an_py310.py`
    * Using `Annotated`, Python 3.10.
* `docs_src/query_params_str_validations/tutorial010_an.py`
    * Using `Annotated`, Python 3.8 (as 3.8 is the oldest, this one doesn't have a part in the name like `py38`).
* `docs_src/query_params_str_validations/tutorial010_py310.py`
    * Python 3.10, not using `Annotated` (as not using `Annotated` is the oldest form, it just doesn't have the `an` part in the file name.
* `docs_src/query_params_str_validations/tutorial010.py`
    * Not using `Annotated`, Python 3.8.

Each of these files represent the same FastAPI app, but with the improved syntax for Python 3.9, or 3.10, or using `Annotated`, but in the end, the same app.

We want to keep these files like this because they have the different ways to create an app, the different supported syntaxes, including backward-compatible ones. They are shown in the docs and tested on CI.

Then, we have tests for that... currently, we just have a test file per variant file, so, we have:

* `tests/test_tutorial/test_query_params_str_validations/test_tutorial010_an_py39.py`
* `tests/test_tutorial/test_query_params_str_validations/test_tutorial010_an_py310.py`
* `tests/test_tutorial/test_query_params_str_validations/test_tutorial010_an.py`
* `tests/test_tutorial/test_query_params_str_validations/test_tutorial010_py310.py`
* `tests/test_tutorial/test_query_params_str_validations/test_tutorial010.py`

But then, each of the files is almost exactly the same code, only with Pytest "markers" to define that something should only be run on Python 3.10, etc. but apart from that, they have the same code.

## The Task

[...]

PR #13174, opened 2025-01-07
✅ Simplify tests for dependencies

alejsdev:


tests/test_tutorial/test_dependencies/test_tutorial006.py
```diff
@@ -1,12 +1,29 @@
...
+def get_client(request: pytest.FixtureRequest):
+    mod = importlib.import_module(f"docs_src.dependencies.{request.param}")
 
-client = TestClient(app)
+    client = TestClient(mod.app)
+    return client
 
 
-def test_get_no_headers():
+@needs_py39
```
svlandeg, line 25: I'm a little confused why we have additional `needs_py39` statements throughout the test file - this means that the whole file won't test on Python 3.8, right? But then the "normal" `tutorial006` also isn't being tested on Python 3.8?

alejsdev: My bad, this is not necessary now with the refactor :sweat_smile: (same in #13182)
I already fixed both, thank you!

svlandeg (approved):

tiangolo (approved):
Nice, thank you! :fire:

PR #13182, opened 2025-01-08
✅ Simplify tests for request_files

alejsdev:
#13150


tests/test_tutorial/test_request_files/test_tutorial002.py
```diff
@@ -1,12 +1,36 @@
...
+
+
+@pytest.fixture(name="client")
+def get_client(app: FastAPI):
+    client = TestClient(app)
+    return client
 
 
-def test_post_form_no_body():
+@needs_py39
```
svlandeg, line 32: why do we need this `needs_py39` here?

svlandeg (approved):

tests/test_tutorial/test_request_files/test_tutorial003.py
```diff
@@ -1,11 +1,45 @@
...
-client = TestClient(app)
+file_required = {
+    "detail": [
+        {
+            "loc": ["body", "files"],
+            "msg": "field required",
+            "type": "value_error.missing",
+        }
+    ]
+}
```
tiangolo, line 39:
```suggestion
```

Is this needed? :thinking: 

I suspect this was copy-pasted from the other file and it was no longer needed there either. :sweat_smile: 

But in that case, we could remove it now. :fire:

luke14025:
thanks

tiangolo (approved):
Nice! Thanks! :fire: :rocket:

PR #13185, opened 2025-01-08
✅ Simplify tests for request_forms_and_files

alejsdev:
#13150


svlandeg (approved):

tiangolo (approved):
:fire:

njsbhu:
I reviewed test_tutorial001.py and would suggest to add Type Hints to Fixtures 

```
@pytest.fixture(name="client")
def get_client(app: FastAPI):
    client = TestClient(app)
    return client
```

instead we can change it to below
```
from typing import Generator
@pytest.fixture(name="client")
def get_client(app: FastAPI) -> Generator[TestClient, None, None]:
        yield client
```

PR #13194, opened 2025-01-10
⬆️ Upgrade `jinja2` to >=3.1.5

DanielKusyDev:
Avoid versions affected by https://github.com/advisories/GHSA-q2x7-8rv6-6q7h

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ standard = [
     # For the test client
     "httpx >=0.23.0",
     # For templates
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks! Makes sense to me. This might be slightly annoying for users who might need to update their `jinja2` versions quite drastically, but then again they probably should anyway.

tiangolo (approved):
Makes sense, thank you! :lock:

PR #13199, opened 2025-01-12
🔧 Add Pydantic 2 trove classifier

johnthagen:
FastAPI also supports Pydantic 2, so add the appropriate trove classifier.

https://github.com/fastapi/fastapi/blob/09a0295bf3bd604eb0acd7eb1f9d9cd4852f3805/pyproject.toml#L45


svlandeg (approved):
Thanks, good catch! 🙏

tiangolo (approved):
Makes sense, thank you! :cake:

PR #13207, opened 2025-01-14
🐛 Fix OpenAPI schema support for computed fields when using `separate_input_output_schemas=False`

vgrafe:
This demonstrates how the output schema differs when the FastAPI app runs with different values for `separate_input_output_schemas` - namely, the computed fields are omitted.

Initially reported here: https://github.com/fastapi/fastapi/discussions/10342


svlandeg:
Thanks! We'll look into this and get back to you 🙏  

(I'll put this PR in draft while we work on it)

vgrafe:
Hi all. Is it possible to have another look into this?

svlandeg:
@vgrafe: apologies for the delay. Yes, I'll look into this in the next few weeks.

svlandeg:
Ok, so I looked into this and @ldthorne was right that the issue occurs at https://github.com/fastapi/fastapi/blob/073e7fc950f1af24634b304afbb34bf212756dff/fastapi/_compat.py#L186-L188

Specifically, when `separate_input_output_schemas` is `False`, this `override_mode` will set all fields to `"validation"` mode. However, to preserve computed fields, they should remain in `"serialization"` mode. The thing is, we can't just change this override mode, as this would prevent the unification of input/output schemas we'd expect from setting `separate_input_output_schemas` to `False`.

So, I'm finding it to be a bit of a philosophical question as to what `separate_input_output_schemas=False` actually means. If it means that the input and the output schema should (always?) be **_identical_**, then really we should be excluding any generated fields, in which case we don't need this PR and it's not a bug but a feature 😬

svlandeg:
@vgrafe: for me to better understand your use-case: why are you not able to use `separate_input_output_schemas=True`?

fastapi/_compat.py
```diff
@@ -238,6 +235,59 @@ def get_definitions(
...
+        # Set override to 'validation' as baseline
+        field_mapping, base_definitions = _get_definitions_with_override_mode(
+            fields=fields,
+            schema_generator=schema_generator,
+            override_mode="validation",
+        )
+
+        # Set override to 'serialization' to be able to add computed fields to the output
+        # Create a new schema generator as it can't be reused
+        schema_generator_2 = GenerateJsonSchema(ref_template=REF_TEMPLATE)
```
svlandeg: Not too happy with this solution, perhaps we should copy the given `schema_generator` instead?

svlandeg:
Update: OK, so I ended up trying to implement a solution that would keep the unifying behaviour, while still adding extra computed fields if there are any. The tests succeed, but we definitely need a few more additional tests to cover various edge cases (and to ensure the coverage test succeeds).

I'm still not 100% sure we want to go down this route though, so I'll first wait and sollicit feedback from either @tiangolo, @YuriiMotov or another maintainer before I continue here.

YuriiMotov:
The behavior I would expect here is: force separate schema creation if model contains computed field.

So

https://github.com/fastapi/fastapi/blob/master/fastapi/_compat.py#L197-L199

Should be something like:
```py
        override_mode: Union[Literal["validation"], None] = (
            None if (separate_input_output_schemas or has_computed_fields(field)) else "validation"
        )
```

And the meaning of `separate_input_output_schemas=False` could be described as: do not create separate input\output schemas if not mandatory

svlandeg:
> The behavior I would expect here is: force separate schema creation if model contains computed field.

Thanks for the review! That is basically the same behaviour as what I had previously implemented, but using your feedback I've now pushed an updated implementation that is much cleaner and easier to maintain / understand.

svlandeg:
Fixed the conflict with https://github.com/fastapi/fastapi/pull/14168. 

<s>Now looking into why the test suite fails for Python 3.14 🤔 
[UPDATE]: the failing test suite is unrelated to this PR, cf https://github.com/fastapi/fastapi/pull/14186
[UPDATE]: test suite green again!</s>

tiangolo (approved):
Nice, thank you @vgrafe! 🚀 

And thanks for the help and work here @svlandeg and @YuriiMotov 🙌 🍰 

This will be available in FastAPI 0.123.4, released in the next hours. 🎉

PR #13224, opened 2025-01-17
👥 Update FastAPI People - Sponsors

tiangolo:
👥 Update FastAPI People - Sponsors


tiangolo:
Created only for debugging.

PR #13230, opened 2025-01-18
📝 Update Request Body's `tutorial002` to deal with `tax=0` case

togogh:
Before this commit, price_with_tax was not included in the response body when tax=0, because the "if item.tax" condition gets set to false

image

Changing the condition to "if item.tax is not None" solves this issue

image


svlandeg (approved):
Thanks, makes sense to fine-tune this docs example to cover the case where `tax` is explicitely `0.0`. I've gone ahead and pushed the same edit also to `tutorial002.py` to keep both files in sync.

tiangolo (approved):
Thanks! :nerd_face:

PR #13241, opened 2025-01-21
✨ Support multiple exc class or status codes in exception handlers

mr-mapache:
I have this annoying situation where I have to handle a lot of exceptions with the same handler. I added an extra argument in the exception_handler method in order to be able to do something like this:

```python
@app.exception_handler(UserNotFound, AccountNotFound, EmailNotFound, SessionNotFound)
async def handle_not_found(request, exc):
    return HTTPException(
        status_code=status.HTTP_404_NOT_FOUND,
        detail=str(exc)
    )

@app.exception_handler(EmailAlreadyExists, UserAlreadyExists)
async def email_already_exists(request, exc):
    return HTTPException(
        status_code=status.HTTP_409_CONFLICT,
        detail=str(exc)
    )

```
Instead of writting a handler for every single exception type or status code. It doesn't break backward compatiblity.


alv2017:
1) Actually, I like the feature, but I find the implementation a bit risky. 
2) Probably it would be safer to implement it as an alternative exception handler in a separate method, then it wouldn't affect the backward compatibility
3) I think the tests are needed as well
4) Maybe it would make sense to open a discussion, then other users could have a say as well. 

(IMHO as of a regular FastAPI user)
:blush:

mr-mapache:
> 1. Actually, I like the feature, but I find the implementation a bit risky.
> 
>     2. Probably it would be safer to implement it as an alternative exception handler in a separate method, then it wouldn't affect the backward compatibility
> 
>     3. I think the tests are needed as well
> 
>     4. Maybe it would make sense to open a discussion, then other users could have a say as well.
> 
> 
> (IMHO as of a regular FastAPI user) 😊

Don't really think it's risky because till now the exception handlers had only one parameter and people were using it like that, and adding an extra *args don't really have any efect.

YuriiMotov:
I opened a #13614 with alternative implementation (`exc_class_or_status_code` is either a single element or a sequence of similar elements).

Feel free to chose which approach is more conventional.

Also, ready to help with adding tests and docs to this PR if you find this implementation more preferable.

mr-mapache:
> I opened a #13614 with alternative implementation (`exc_class_or_status_code` is either a single element or a sequence of similar elements).
> 
> Feel free to chose which approach is more conventional.
> 
> Also, ready to help with adding tests and docs to this PR if you find this implementation more preferable.

This implementation has issues, since it passed all tests It looks like it work but when writting tests found out that there are not enough tests cases for exceptions, so if merged will break the exception handlers silently.  

Maybe this PR should be closed.

tiangolo:

I'm undecided about having the feature in FastAPI, but for now, let's better continue in https://github.com/fastapi/fastapi/pull/13614 as that one has more tests and docs, etc.

Given that, I'll close this one for now. Thanks! :coffee:

PR #13250, opened 2025-01-23
♻️ Replace float with annotated_types in `params.py` and `params_functions.py`

junah201:
# Change

This commit refactors the type hints in `params.py` and `params_functions.py`, changing the following fields:

- `gt: Optional[float] = None` → `gt: Optional[SupportsGt] = None`
- `ge: Optional[float] = None` → `ge: Optional[SupportsGe] = None`
- `lt: Optional[float] = None` → `lt: Optional[SupportsLt] = None`
- `le: Optional[float] = None` → `le: Optional[SupportsLe] = None`

# Justification

The original type hints in Pydantic are:

https://github.com/pydantic/pydantic/blob/b0ef7d043f0c63c9cf25781e34597f43348aaafb/pydantic/fields.py#L774C1-L777C52
```py

import annotated_types

@overload
def Field(
    ...
    gt: annotated_types.SupportsGt | None = _Unset,
    ge: annotated_types.SupportsGe | None = _Unset,
    lt: annotated_types.SupportsLt | None = _Unset,
    le: annotated_types.SupportsLe | None = _Unset,
    ...
)
```

Given `annotated_types` is available from `python>=3.8`, there are no compatibility issues.
Also, pydantic already includes `annotated_types` as a dependency.


alv2017:
Below is an excerpt from the `fastapi/param_functions.py`
```
def Path(  # noqa: N802
    ...
    gt: Annotated[
        Optional[float],
        Optional[SupportsGt],
        Doc(
            """
            Greater than. If set, value must be greater than this. Only applicable to
            numbers.
            """
        ),
    ] = None,
    ...
```
The docstring says, that if the `gt` parameter is set, then it needs to be numeric. I think that you were not taking this into consideration.

junah201:
Thank you for your review!
I will update the hints  in the code and also documentation.

Viicos (commented):
Using `annotated_types` relies on the assumption that Pydantic depends on it. This could thus break at any time if Pydantic removes `annotated-types` as dependency. So either `annotated_types` should be added explicitly as a fastapi dependency, or (and I would suggest going this way), [the protocols](https://github.com/python/typeshed/blob/527d724bf4b9cd13ee0b1c7a0cd07837a49b9df5/stdlib/_typeshed/__init__.pyi#L87-L97) provided by [typeshed](https://github.com/python/typeshed/tree/main/stdlib/_typeshed) could be used.

Ale-Cas (commented):
Overall LGTM, I would also replace annotated_types with the built-in protocols for comparison, and would remove the "not just numbers" in the docstrings.

alv2017:
I'm wondering if comparison validations are going to work with non-floats. :smiley:

YuriiMotov (commented):
I think we are limited by openapi spec (JSON SCHEMA spec, actually) here.
Specifying `gt`, `ge`, etc.. should result in adding `maximum`, `exclusive_maximum`, etc.. in openapi schema.
But these parameters are only supported for numbers:
* https://swagger.io/docs/specification/v3_0/data-models/data-types/#numbers
* https://json-schema.org/draft/2020-12/json-schema-validation#name-validation-keywords-for-num

tiangolo:
I'll pass on this one for now, thanks! :coffee:

PR #13264, opened 2025-01-28
✨ Raise error with clear error message when generator didn't yield anything

Jaza:
If an async generator dependency raises `RuntimeError: generator didn't yield`, make `solve_dependencies` catch and re-raise it, to more easily identify the dependency responsible for the error, and to provide more information on how to fix the dependency.


Jaza:
This resolves #13238, by making the raised exception more informative, and by adding an example of a badly-behaving async dependency to the docs.

fastapi/dependencies/utils.py
```diff
@@ -557,7 +558,24 @@ async def solve_generator(
         cm = contextmanager_in_threadpool(contextmanager(call)(**sub_values))
     elif is_async_gen_callable(call):
         cm = asynccontextmanager(call)(**sub_values)
-    return await stack.enter_async_context(cm)
+
+    try:
+        solved = await stack.enter_async_context(cm)
+    except RuntimeError as ex:
+        if "generator didn't yield" not in f"{ex}":
```
```suggestion
        if str(ex) == "generator didn't yield":
```
This should be faster

fastapi/dependencies/utils.py
```diff
@@ -557,7 +558,24 @@ async def solve_generator(
...
+            raise ex
+
+        dependency_name = getattr(call, "__name__", "(unknown)")
+        raise FastAPIError(
+            f"Dependency {dependency_name} raised: {ex}. There's a high chance that "
+            "this is a dependency with yield that has a block with a bare except, or a "
+            "block with except Exception, and is not raising the exception again. Read "
+            "more about it in the docs: "
```
```suggestion
            "this is a dependency with yield that catches an exception using except, "
            "but doesn't raise the exception again. Read more about it in the docs: "
```

YuriiMotov (commented):
@Jaza, thanks for your interest!

I think adding a clear error message is a good idea.
As for updating docs - I think we don't need these code examples. Everything is clear in current version.

We can just add tests for `sync` and `async` dependencies to make sure it works

Jaza:
@YuriiMotov any chance we can re-open this PR? I'm happy to go with your suggested changes.

fastapi/dependencies/utils.py
Jaza: I guess you meant `if str(ex) != "generator didn't yield":` . I've updated it to that.

Jaza:
@YuriiMotov I've updated the PR, removed all docs changes, added a test for when an exception is not re-raised. Can you please review again.

tests/test_dependency_runtime_errors.py
```diff
@@ -0,0 +1,55 @@
...
+from fastapi import Depends, FastAPI
+from fastapi.testclient import TestClient
+
+app_reraises = FastAPI()
+app_doesnt_reraise = FastAPI()
+
+
+async def get_username_reraises():
+    try:
+        async with await open_file("/path/to/sanchez.txt", "r") as f:
```
YuriiMotov:
```suggestion
        async with await open_file("/non_existing/path.txt", "r") as f:
```
Maybe better use self-descriptive path?

tests/test_dependency_runtime_errors.py
```diff
@@ -0,0 +1,55 @@
...
+            yield await f.read()  # pragma: no cover
+    except OSError as ex:
+        raise RuntimeError("File something something, wubba lubba dub dub!") from ex
```
YuriiMotov:
```suggestion
        raise RuntimeError("File read error") from ex
```
I think more neutral text would be better

tests/test_dependency_runtime_errors.py
```diff
@@ -0,0 +1,55 @@
+import pytest
+from anyio import open_file
+from fastapi import Depends, FastAPI
+from fastapi.testclient import TestClient
+
+app_reraises = FastAPI()
+app_doesnt_reraise = FastAPI()
```
YuriiMotov: We could just add 2 endpoints to the same app, but it's not critical

YuriiMotov (approved):
LGTM in general!

I would only change texts in test to more neutral and self-descriptive.

@Jaza, thank you!

tests/test_dependency_runtime_errors.py
Jaza: Done

Jaza:
@YuriiMotov updated. Please review again.

Jaza:
This has been approved. It's been nearly a year since I submitted this PR. Any chance it can be merged soon?

Jaza:
Approved, conflicts resolved, has been sitting here for over a year. @tiangolo thanks for fixing latest conflicts, any chance you can merge this?

PR #13290, opened 2025-01-31
✅ Detect blocking calls in coroutines using BlockBuster

cbornet:
This PR uses the [blockbuster](https://github.com/cbornet/blockbuster) library to detect blocking calls made in the asyncio event loop during unit tests.
Avoiding blocking calls is hard as these can be deeply buried in the code or made in 3rd party libraries.
Blockbuster makes it easier to detect them by raising an exception when a call is made to a known blocking function (eg: `time.sleep`).


tests/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+
+@pytest.fixture(autouse=True)
+def blockbuster(request):
+    with blockbuster_ctx("fastapi") as bb:
+        bb.functions["io.BufferedReader.read"].can_block_in("starlette/testclient.py", "receive")
+        for func in [
+            "os.stat",
+            "io.TextIOWrapper.read",
+        ]:
+            bb.functions[func].can_block_in("pydantic/networks.py", "import_email_validator")
```
cbornet: Pydantic's `import_email_validator` calls `metadata.version` which does a blocking read.
The result is cached so the impact of blocking the event loop is not big.

tests/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+        for version in ["001", "002"]:
```
cbornet, line 18: The tutorials do blocking calls.
So I've put some exemptions.
Would you prefer to fix the tutorials ?

tests/test_tutorial/test_events/test_tutorial002.py
```diff
@@ -3,6 +3,11 @@
 from fastapi.testclient import TestClient
 
 
+@pytest.fixture(autouse=True)
+def deactivate_blockbuster(blockbuster):
+    blockbuster.deactivate()
```
cbornet, line 8: starlette `Router`'s `shutdown` doesn't execute the shutdown event in a thread.
And `tutorial002` writes blockingly to the log file in its `shutdown_event`.
Here I deactivate blockbuster for the `tutorial002` test but maybe the tutorial could be changed to be fully async ? Or should starlette's `Router` defer the events execution to a thread ?

tests/test_tutorial/test_sql_databases/test_tutorial001.py
```diff
@@ -12,6 +12,11 @@
 from tests.utils import needs_py39, needs_py310
 
 
+@pytest.fixture(autouse=True)
+def deactivate_blockbuster(blockbuster):
```
cbornet, line 16: Same as before but here it's `tutorial001` `on_startup` which does a blocking call to `create_db_and_tables`

tests/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+    with blockbuster_ctx("fastapi") as bb:
+        bb.functions["io.BufferedReader.read"].can_block_in(
+            "starlette/testclient.py", "receive"
+        )
+        for func in [
+            "os.stat",
+            "io.TextIOWrapper.read",
+        ]:
+            bb.functions[func].can_block_in(
+                "pydantic/networks.py", "import_email_validator"
```
cbornet, line 16: Pydantic's import_email_validator calls metadata.version which does a blocking read.
The result is cached so the impact of blocking the event loop is not big.

tests/conftest.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from blockbuster import blockbuster_ctx
+
+
+@pytest.fixture(autouse=True)
+def blockbuster(request):
+    with blockbuster_ctx("fastapi") as bb:
+        bb.functions["io.BufferedReader.read"].can_block_in(
+            "starlette/testclient.py", "receive"
```
cbornet, line 9: Starlette's  `_TestClientTransport` `receive` calls `httpx.Request.read()`. I think it should use `httpx.Request.aread()` ?

alv2017:
@cbornet: Thanks for demo! Looks interesting. :+1:

cbornet:
> It might be useful to put this on Discussion board as well

Done : https://github.com/fastapi/fastapi/discussions/13296

svlandeg (commented):
Thanks for the PR!

Pardon my ignorance, but a genuine question: what is the actual issue this PR is aiming to solve? Is the test suite currently not running fine?

cbornet:
Thiis PR aims at protecting from writing code that blocks the event loop.
Like doing a sync network call, a time.sleep, and so on.
If you do so and the code path is tested, it will raise an Exception and fail the test.

cbornet:
Updated blockbuster version to fix CI issues.

cbornet:
Hi. Could this PR be reviewed (or discussed in the discussion thread if needed) ?

svlandeg (commented):
I personally don't really see the necessity for having this, and adding another dependency to the test suite. But I'll leave this up for review for Yurii or Tiangolo to chime in.

tiangolo:
Thanks for the interest! I don't think we currently need this in FastAPI itself so I'll pass on this, but thanks! :coffee:

PR #13313, opened 2025-02-03
📝 Fix test badge

esadek:
Before:
image

After:
image


alejsdev:
Hello @esadek 

Thanks for your interest in contributing to FastAPI. However, this change has to be done in `/home/user/code/fastapi/docs/en/docs/index.md` as well. Could you update this PR? Thanks :smile:

esadek:
@alejsdev Updated the docs as well

svlandeg (approved):
Thanks for the fix, @esadek! I wonder whether the format of this has changed recently, because it definitely was working before 🤔 

Either way, the format you're using is the one currently [documented](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/adding-a-workflow-status-badge#using-the-workflow-file-name) so looks good. Thanks again! 

I will merge this and follow up with a separate PR #13329 to make sure this gets fixed in all other languages as well.

PR #13329, opened 2025-02-06
🩺 Unify the badges across all tutorial translations

svlandeg:
Currently, the displayed badges can be quite different between translations, even potentially showing a different status, like is the case right now: https://fastapi.tiangolo.com/ vs. https://fastapi.tiangolo.com/bn/

image

image

This PR ensures consistency of the displayed badges:

- Using only pushes to `master` to determine the status of the `Test` badge
- Use the new format as introduced in #13313 for all languages
- Using `coverage-badge.samuelcolvin.workers.dev` for the `Coverage` badge
- Including a badge for the supported Python versions on all pages.


svlandeg:
Note that currently, the `Test` badge displayes "no status" which seems wrong as [the runs](https://github.com/fastapi/fastapi/actions?query=workflow%3ATest+event%3Apush+branch%3Amaster) are all green. While we probably need to investigate that, too, I don't think it's directly relevant to this PR, which simply unifies everything.

svlandeg (commented):
Since submitting this PR, the coverage badge seems to be broken on `master` as well;

image

Again - I don't think this is related to this PR in particular, it seems like a wider problem.

[UPDATE:] This should now be fixed

alv2017:
I would suggest to fix the English version first. :hammer_and_wrench: 
(Currently on the main website `Test` and `coverage` badges are not working)

alv2017:
@svlandeg 

Applied fixes looks good. :smiley:

Missing fixes: 
- it
- ja

alejsdev (approved):

tiangolo (approved):
Nice, thank you! :rocket:

PR #13331, opened 2025-02-07
📝 Explicitly check for `None` in the docs when testing query `q`

DanielYang59:
### Summary

- Use `if q is not None:` over `if q:` for explicit `None` check
- Avoid assigning when it's immediately returned

I really appreciate the highly comprehensive tutorial in the first place. And free feel to let me know if my change is not a valid concern in the web framework context (as I'm just starting to learn `fastapi`).

In this case, I think it's better to explicitly check for `None` instead of checking for truthiness, but I'm not entirely sure if `q=""` (empty string) is a valid/useful query (e.g. `http://127.0.0.1:8000/items/1234?q=`)

---

While this may be less of an issue for str, I generally think using truthiness to check for None is not ideal. Imagine a scenario where a query parameter is an integer:

```python
@app.get("/items/{item_id}")
async def get_items(item_id: str, q: int | None = None):
    if q:
        return {"item_id": item_id, "q": q}
    return {"item_id": item_id}
```

In this case `q=0` would lead to surprising behaviour.


alv2017:
@DanielYang59: 

1) what about trying to run examples with q="". do they make sense?
2) what about trying to add tests for q=""? Do they pass? 

:smiley:

DanielYang59:
Hi @alv2017 thanks for the quick response!

> what about trying to run examples with q="". do they make sense?

I don't think I understand this question here (I'm just starting to learn `fastapi` and I don't really know what are the "examples" to run?). But AFAIK, `q = None` and `q = ""` serve different purposes, with the former as a default value when `q` is not given (optional), and the latter as `q` is given but as an empty string (though I'm not sure if `q = ""` is a valid use case in this context).

> what about trying to add tests for q=""? Do they pass?

Sure thing. Happy to add tests for these.

alv2017:
@DanielYang59 
Do not add anything for now, just try to think over the examples, and experiment with tutorials code and tests on your machine.

One of the condition that we are after is q = ""
:smiley:

alv2017:
I mean that in this PR you are proposing to change more than 70 tutorials, and it would be good to run the changed code and observe the results. :nerd_face: 

The case of concern is setting q to be an empty string.

DanielYang59:
> it would be good to run the changed code and observe the results. 

This sounds exactly the job of unit tests, but looks like many "duplicate" tests are removed in #13150. Should this be added into the test suite?

> Do not add anything for now

It's very confusing, and I don't understand whether you're requesting more tests or not (I think having automated tests is safer than requesting people to test some behaviour locally)?

alv2017:
I'm trying to say, that my comments are not a call to action of changing the PR. :smile:
(It is rather a proposal to rethink it and understand the code and the implemented changes better.)

DanielYang59:
> my comments are not a call to action of changing the PR

I agree with you that tests are needed.

> proposal to rethink it and understand the code and the implemented changes better

I still don't understand this. Please point out if there's any issue with the change, otherwise I believe my intention for this change is clear enough, i.e. using explicit `None` check over truthiness check.

svlandeg:
@DanielYang59: thanks for your interest in contributing to `fastapi`! 🚀 

I understand the general point you're trying to make, that checking for truthiness is not the same as checking for `None`. Your point about doing this for integers makes sense, as this would be problematic in the case the value is `0`. However, the situation is quite different with strings, as an empty string may or may not have actual meaning.

Consider the URL `http://127.0.0.1:8000/items/342?q=`. Can we really say we're defining a parameter `q` here? In most cases, it feels to me like this can be considered as if `q` is non-existent, or `None`. I think that's why the current tutorials all simply check `if q`.

Also - they are just tutorials. If your specific application does require a distinction between `None` and `""` then clearly you are free to program your `app` however you like.

To add my 2 cents to the discussion with @alv2017: what I think they are trying to say (correct me if I'm wrong @alv2017) is that you should _first_ play around with the code and the consequences of changes _locally_ and truly understand the current & proposed behaviour before making a PR that would change 70+ files. I think they're just trying to give you some good advice 😉 

For instance - think back about the integer `0` case. It's not relevant here, but imagine you would have a code base where this issue had crept in. Then, you could have added a unit test for the edge case where the value is `0`, and you could have demonstrated that the behaviour on `master` is unexpected, while your PR would fix it. Do you see? In the same way, @alv2017 is suggesting to think about how tests with `q=""` would look like. What behaviour would they test? Would this not be working as expected on `master`? What, specifically, would your proposed PR fix? These type of experiments & thoughts would help us come to the right conclusion on whether or not we need the changes proposed in the PR.

DanielYang59:
Hi @svlandeg thanks for your reply!

> Consider the URL http://127.0.0.1:8000/items/342?q=. Can we really say we're defining a parameter q here? In most cases, it feels to me like this can be considered as if q is non-existent, or None. I think that's why the current tutorials all simply check if q.

Yes this is [what I was concerned about from the start](https://github.com/fastapi/fastapi/pull/13331#issuecomment-2646303411), i.e. is passing `q=""` in query a valid use case in this context?

svlandeg:
> If in most cases it's safe to ignore query as an empty string, then maybe my concern (and this PR) is not valid. On the other hand, if there're many use cases where we need to handle such empty string query, then it's better that we carefully differentiate them vs None.

I've briefly discussed with the team internally and we've decided that we'll keep the tutorial code as is, assuming that in most cases the empty string does not really carry semantic information. But again, if anyone does require special treatment of such an empty string, they can implement things differently for their specific app.

So to conclude, I'll go ahead and close this PR, but thanks again for the contribution @DanielYang59 and thanks to @alv2017 for the feedback and discussion! 🙏

DanielYang59:
Hi thanks for discussing over this topic!

> assuming that in most cases the empty string does not really carry semantic information

Sure great to know. That is my concern as I'm new to fastapi

And thanks again for the comprehensive tutorial :)

PR #13339, opened 2025-02-09
🐛 Fix handling of raw strings in endpoint docstrings and truncate at form feed

kanav03:
…form feed

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -528,15 +528,15 @@ class APIRoute(routing.Route):
             self.response_field = None  # type: ignore
             self.secure_cloned_response_field = None
         self.dependencies = list(dependencies or [])
-        
+
         # Get docstring and handle raw strings
         docstring = inspect.cleandoc(self.endpoint.__doc__ or "")
         if docstring.startswith('r"""') or docstring.startswith("r'''"):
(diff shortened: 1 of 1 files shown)


fastapi/routing.py
```diff
@@ -528,10 +528,15 @@ def __init__(
...
             self.secure_cloned_response_field = None
         self.dependencies = list(dependencies or [])
-        self.description = description or inspect.cleandoc(self.endpoint.__doc__ or "")
-        # if a "form feed" character (page break) is found in the description text,
-        # truncate description text to the content preceding the first "form feed"
-        self.description = self.description.split("\f")[0].strip()
+
+        # Get docstring and handle raw strings
+        docstring = inspect.cleandoc(self.endpoint.__doc__ or "")
+        if docstring.startswith('r"""') or docstring.startswith("r'''"):
```
dolfinus, line 534: This will not work, `r'''...'''` is already handled by interpreter, and there is no difference between this form, `'''...'''` and `'...'` at all. This could be done only at AST level

svlandeg (commented):
@kanav03, thanks for contibuting to `fastapi`!

You've labeled this as a bug, so could you add a unit test that breaks on `master` but succeeds with this PR?

The coverage test will also need to be brought back to 100%, meaning we test each new code line.

PR #13366, opened 2025-02-13
📝 Minor additions to the documentation

nfsknight:
I made a few small additions and grammatical corrections to your documentation.  Please see if you think they are useful.


.python-version
```diff
@@ -0,0 +1 @@
+fastapi
```
svlandeg, line 1: Can you explain why this file was added?

svlandeg, line 1: Can you explain why this file was added?

svlandeg (commented):
Thanks for your interest in contributing to `fastapi`!

.python-version
nfsknight: I noticed that file, too.  I didn't add it myself, at least not that I'm aware of.  I thought it had something to do with the setup steps on your Contributing page.  Maybe it has to do with the pyenv virtual environment I created to work on the branch.  Pyenv manages which python versions are used in each virtual environment.  Do you possibly have any ideas?

svlandeg: I don't know how you manage your Python environments, but this is the sort of file that you should ignore locally and shouldn't be committed as part of the PR. Can you delete it again?

docs/en/docs/advanced/response-change-status-code.md
```diff
@@ -2,7 +2,9 @@
 
 You probably read before that you can set a default [Response Status Code](../tutorial/response-status-code.md){.internal-link target=_blank}.
 
-But in some cases you need to return a different status code than the default.
+You also read that you signal errors (400-range status codes) by raising an `HTTPException`.
+
+But in some cases you need to return a status code other than the default or an error.
```
svlandeg, line 7: I don't think the additional explanation here is helpful - instead it made sense to me to just focus on the default status code and how you can change that. What was the reasoning to add the bit around `HTTPException`? This isn't otherwise used or mentioned in the remainder of the page.

docs/en/docs/tutorial/body.md
```diff
@@ -26,9 +26,9 @@ First, you need to import `BaseModel` from `pydantic`:
 
 ## Create your data model
 
-Then you declare your data model as a class that inherits from `BaseModel`.
+Then you declare your data model as a class that inherits from `BaseModel`. These classes will be the containers for the various requests and responses you intend to implement.
 
-Use standard Python types for all the attributes:
+`BaseModel` will handle the data validation of all requests and responses for you.  All you do is declare what type each attribute is supposed to have.  Use standard Python types for the attributes:
```
svlandeg, line 31:
```suggestion
`BaseModel` will handle the data validation of all requests and responses for you.  All you have to do is declare what type each attribute is supposed to have.  Use standard Python types for the attributes:
```

docs/en/docs/tutorial/schema-extra-example.md
```diff
@@ -24,7 +24,7 @@ That extra info will be added as-is to the output **JSON Schema** for that model
 
 //// tab | Pydantic v2
 
-In Pydantic version 2, you would use the attribute `model_config`, that takes a `dict` as described in <a href="https://docs.pydantic.dev/latest/api/config/" class="external-link" target="_blank">Pydantic's docs: Configuration</a>.
+In Pydantic version 2, you would use the attribute `model_config`. `model_config` formats the behavior of an entire model, as opposed to `Field` which formats the behavior of individual attributes. It takes a `dict` as described in <a href="https://docs.pydantic.dev/latest/api/config/" class="external-link" target="_blank">Pydantic's docs: Configuration</a>.
```
svlandeg, line 27:
```suggestion
In Pydantic version 2, you would use the attribute `model_config`, which formats the behavior of an entire model, as opposed to `Field` which formats the behavior of individual attributes. The `model_config` takes a `dict` as described in <a href="https://docs.pydantic.dev/latest/api/config/" class="external-link" target="_blank">Pydantic's docs: Configuration</a>.
```

svlandeg (commented):
Added a few more minor comments.

.python-version
nfsknight: Of course.  I'm pretty sure it has to do with how pyenv manages local environments.  I will delete it and push the changes.

docs/en/docs/advanced/response-change-status-code.md
nfsknight: I wanted to make it clear that one does not signal 400-range status codes using the `Response` parameter even though one might think that intuitively because they also could be considered divergences from default status codes set in path operation decorators.  That was my reasoning.  I'm happy to remove it if you prefer.

svlandeg: Got it - thanks for clarifying. I'll leave it to Tiangolo for the final decision!

svlandeg (approved):
Thanks for the contribution - this mostly looks good to me! Leaving it to Tiangolo for a final review.

docs/en/docs/advanced/response-change-status-code.md
nfsknight: Ok, thank you for reviewing my changes.  Have a great day!

tiangolo:
Thanks for the interest @nfsknight! But for now, I prefer to keep the docs in the slimmer version they are. Thanks! :coffee:

nfsknight:
Ok no problem. Just suggestions.

Sent with [Proton Mail](https://proton.me/mail/home) secure email.

PR #13379, opened 2025-02-16
📝 Enhance docstrings: fix typos and improve clarity.

isgin01:


svlandeg (commented):
Hi, thanks for you interest in contributing to `fastapi`!

These edits only concern internal docstrings and they're mostly about personal preference instead of really correcting typos or incorrect information. I think either way of phrasing is mostly fine. I'll leave the final decision to Tiangolo whether we want to make these changes or not.

PR #13391, opened 2025-02-18
♻️ Refactor tests in `tests/test_openapi_separate_input_output_schemas.py`

alv2017:
This PR refactors `tests/test_openapi_separate_input_output_schemas.py`

1) It allows to get rid of `get_app_client(separate_input_output_schemas: bool = True)` function, which essentially plays the role of the `pytest fixture`.

2) The `pytest fixture` for the FastAPI application with parameter is implemented instead. 

3) Suggests a method on testing the impact of changing FastAPI parameters. 

As a result:
1) We get a slightly improved test coverage, and 
2) The tests become clearer and more readable
3) When testing we can explicitly test for the value of the parameter in test
4) The tests also become more "pytestic" :blush:

Major Disadvantage:
Performance: `scope="function"`, hence the fixture is invoked once per test function.

For reviewers: 
when reviewing compare both versions side by side, otherwise it might be difficult.


tests/test_openapi_separate_input_output_schemas.py
```diff
@@ -45,47 +47,40 @@ def read_items() -> List[Item]:
...
-    client = TestClient(app)
-    return client
+    return app
 
 
-def test_create_item():
-    client = get_app_client()
-    client_no = get_app_client(separate_input_output_schemas=False)
+@pytest.mark.parametrize("separate_input_output_schemas", [True, False])
+def test_create_item(app: FastAPI, separate_input_output_schemas: bool):
```
valentinDruzhinin, line 54: I would rename the testcase to be smt like: `test_create_item_when_separate_input_output_schemas_is`.
In that case we will have the clearer log output.

tests/test_openapi_separate_input_output_schemas.py
```diff
@@ -94,46 +89,44 @@ def test_create_item_list():
...
+            "sub": {"subname": "subname", "sub_description": None, "tags": []},
+        },
+        {"name": "Plumbus", "description": None, "sub": None},
+    ]
 
 
 @needs_pydanticv2
-def test_openapi_schema():
-    client = get_app_client()
+@pytest.mark.parametrize("separate_input_output_schemas", [True])
```
valentinDruzhinin, line 124: Can we join 2 test cases `test_openapi_schema` and `test_openapi_schema_no_separate` into one with parameterized (`True, False`)? The logic in both test cases are the same.

valentinDruzhinin (commented):
Looks good, just a couple of suggestions

tests/test_openapi_separate_input_output_schemas.py
alv2017: I do not think so, the output schemas are different for `test_openapi_schema` and `test_openapi_schema_no_separate`

This is the excerpt from the code:

```python
        separate_input_output_schemas: Annotated[
            bool,
            Doc(
                """
                Whether to generate separate OpenAPI schemas for request body and
                response body when the results would be more precise.

                This is particularly useful when automatically generating clients.

                For example, if you have a model like:

                ```python

                class Item(BaseModel):
                    name: str
                    tags: list[str] = []
                ```

                When `Item` is used for input, a request body, `tags` is not required,
                the client doesn't have to provide it.

                But when using `Item` for output, for a response body, `tags` is always
                available because it has a default value, even if it's just an empty
                list. So, the client should be able to always expect it.

                In this case, there would be two different schemas, one for input and
                another one for output.
                """
            ),
        ] = True,
```

valentinDruzhinin: I just recommend to join both test cases, because they the same. And we can remove the duplication of these code lines:
```
    client = TestClient(app)

    assert app.separate_input_output_schemas == separate_input_output_schemas
```
Since you are doing the refactoring of the test file and touched these test cases, I think it would be cool to remove the duplicated code, right?

alv2017: I was thinking about it, but I rejected this idea. I prefer to skip schema revision in this PR. :smiley:

valentinDruzhinin (approved):
LGTM! Left the comments up to the author to decide.

tiangolo:
Thanks for the work on this @alv2017!

Indeed, this uses more features of pytest. Neverhteless, I feel the logic is a bit more complex, it needs to think about the fixture that will be called automatically by pytest, but that needs as a parameter another input, and that input is coming from the same test function with a parametrize decorator... it's a long explanation that needs to be understood to work with it.

The alternative is to leave the code as it is right now, for me, it's a bit simpler to think about it as it is, and there's no clear disadvantage. So, for now, I'll pass on this one, but thanks for the interest! :coffee:

alv2017:
@tiangolo: I agree, this is a bit too extreme. I created this PR rather for demonstration purposes. :smiley:

njsbhu:
In test_openapi_separate_input_output_schemas.py, I reviewed and suggest to add type hints for test functions like below

`def test_read_items(app: FastAPI, separate_input_output_schemas: bool) -> None:`

PR #11140, opened 2024-02-13
✨ Add support for PEP695 TypeAliasType

dev-loki:
Fixes: #10719
Discussed (by others) in: #10662

Adds support for [PEP695](https://peps.python.org/pep-0695/). 

Sidesupport (please check if intended!): Only loads `*_py312` files for python3.12 and later, as only the test introduces syntax not compatible for python < 3.12.

```python
def some_complex_function():
    return 42
    
type DependedVariable = Annotated[int, some_complex_function]
```

Using `DependedVariable` in route dependencies will now work :).

PR #13401, opened 2025-02-20
✨ Add type alias syntax support for dependency injection

kilian-hu:
This PR adds support for dependency injection with types defined by the type parameter syntax which was introduced with python 3.12 ([PEP 695](https://peps.python.org/pep-0695/)).

It makes it possible to write
```python
async def get_foo() -> str:
    return "foo"

type Foo = Annotated[str, Depends(get_foo)]

@app.get("/foo")
def my_endpoint(foo: Foo):
    return foo
```
which previously was unsupported.


tests/test_type_alias_depends.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+async def get_bar2(y: str = "y") -> str:
+    return f"bar2_{y}"
+
+
+# Equivalent to:
+# type Bar1 = Annotated[str, Depends(get_bar1)]
+# type Bar2 = Annotated[str, Depends(get_bar2)]
+Bar1 = TypeAliasType("Bar1", Annotated[str, Depends(get_bar1)])
```
Kludex, line 29: I think it's possible to test the other syntax, I don't remember how.

Kludex (approved):

tests/test_type_alias_depends.py
kilian-hu: It's possible to make the tests work with older python versions by using conditional imports, but problem is that when we use the actual `type Foo = ...` syntax the coverage report fails because it parses all the files. Unless I'm missing something. I'd also be happier to use the actual syntax :)

Kludex: You can create a file, then use `importlib.import_module` to import, so you can test on python 3.12+.

Use/create the `needs_py312` fixture.

kilian-hu: Thanks, I added `needs_py312` and used `importlib.import_module` like you suggested.

I see that you already have some entries in `pyproject.toml -> [tool.coverage.run] -> omit` ✨ so I suppose it's fine that I added the file that cannot be parsed by python < 3.12 ?

s5181981464:
> [此 PR 增加了對依賴注入的支持，其類型由在 python 3.12（ PEP 695](https://peps.python.org/pep-0695/) ）中引入的類型參數語法定義。
> 
> 這使得編寫
> 
> ```python
> async def get_foo() -> str:
>     return "foo"
> 
> type Foo = Annotated[str, Depends(get_foo)]
> 
> @app.get("/foo")
> def my_endpoint(foo: Foo):
>     return foo
> ```
> 
> 此前，此功能不受支援。

Viicos:
This is overlapping with https://github.com/fastapi/fastapi/pull/11140, and this PR suffers from the same issue I mentioned [here](https://github.com/fastapi/fastapi/pull/11140/files#r1940968983).

kilian-hu:
@Viicos thanks! Then I'm close this in light of https://github.com/fastapi/fastapi/pull/11140

PR #13411, opened 2025-02-22
✅ Fix a minor bug in the test `tests/test_modules_same_name_body/test_main.py`

alv2017:
Description: 
1.
```
data = response.json()
```
has been replaced with the assertion statement
```
assert data == response.json()
```

2.
Tests parametrization applied to reduce the duplicated code


svlandeg (approved):
Good catch with the missing `assert`! 

Also a good idea to use `parametrize` here and avoid code duplication to test all corner cases.

Should be ready to merge 🥳

tiangolo (approved):
Thank you! :rocket: :tada:

njsbhu:
I reviewed test.main.py, to get rid off from repeated data, we can adapt test fixtures, like below logic

```
import pytest
from .app.main import app

client = TestClient(app)
@pytest.fixture
def valid_data():
       -----------
@pytest.fixture
def invalid_data():
      -----------
@pytest.mark.parametrize(
    "path", [-----]
)
def test_post(path, valid_data):
    response = client.post(path, json=valid_data)
    assert valid_data == response.json()

@pytest.mark.parametrize(
    "path", [-----]
)
def test_post_invalid(path, invalid_data):
    response = client.post(path, json=invalid_data)
    assert response.status_code == 422, response.text
```

PR #13413, opened 2025-02-23
✅ Refactor `tests/test_empty_router.py` to avoid code duplication

valentinDruzhinin:
Use `pytest.mark.parametrize` to reduce code duplicates and make tests more clear


svlandeg (approved):
Thanks!

tiangolo:
Thanks! Given it's only two lines of more or less duplicated code, but are the two lines that give away the difference, the test is more clear to me as it currently is, so I'll pass on this one, but thanks! :coffee:

PR #13415, opened 2025-02-24
👷 Use `wrangler-action` v3

joakimnordling:
The fix to the wrangler-action was released already and dependabot has been updating it couple of times to newest version already, so should be OK to switch back to the @v3 and get rid of the TODO about it.


svlandeg (approved):
Good catch, thanks!

tiangolo (approved):
Thanks! :cake:

PR #13417, opened 2025-02-24
✅ Fix `tests/test_query.py` to verify the exact response based on the Pydantic version

valentinDruzhinin:
Right now we check that at least one response is correct. Which is not good, because we can pass test with response for pydantic v1 when pydantic v2 is used. To fix this and check on the exact value `needs_pydanticv1` and `needs_pydanticv2` decorators was used.


svlandeg:
Thanks! We'll get this reviewed by the team and get back to you 🙏

valentinDruzhinin:
Hi @YuriiMotov, may I ask you to take a look, please? 🙏

YuriiMotov:
I think we don't need these changes.

For now these tests will run twice (with Pydantic V1 and with Pydantic V2).
When Pydantic V1 is used, the result will either match the second `dict` (which is marked with `# TODO: remove when deprecating Pydantic v1`) or don't match any of them (in case something went wrong).
When Pydantic V2 is used, the result will either match the first `dict` or don't match any of them (in case something went wrong).

The only potential issue that may happen here is that tests will still pass if one of the following things happen:

**1.** Pydantic V1 starts returning error in a way that Pydantic V2 does
**2.** Pydantic V2 starts returning error in a way that Pydantic V1 does

But I guess this is not something that we need to test.

@valentinDruzhinin, do you agree? Or, am I missing something?

valentinDruzhinin:
> I think we don't need these changes.
> 
> For now these tests will run twice (with Pydantic V1 and with Pydantic V2). When Pydantic V1 is used, the result will either match the second `dict` (which is marked with `# TODO: remove when deprecating Pydantic v1`) or don't match any of them (in case something went wrong). When Pydantic V2 is used, the result will either match the first `dict` or don't match any of them (in case something went wrong).
> 
> The only potential issue that may happen here is that tests will still pass if one of the following things happen:
> 
> **1.** Pydantic V1 starts returning error in a way that Pydantic V2 does **2.** Pydantic V2 starts returning error in a way that Pydantic V1 does
> 
> But I guess this is not something that we need to test.
> 
> @valentinDruzhinin, do you agree? Or, am I missing something?

@YuriiMotov thank you. My concern also is that we also allow to pass the test if Pydantic V1 returns the same result as for Pydantic V2 and vice versa. Maybe that is overthinking but having more precise tests is better than having tests which can allow to pass the wrong results. If you think we don't need this, I'm fine with closing this PR :)
Thanks

tiangolo:
Thanks for the interest! The codebase actually used to look like you propose, and it was a bigger burden to maintain, so I refactored it all with `IsDict`, etc. :sweat_smile: 

So, I'll pass on this one for now, but thanks! :coffee:

Issue #13400, opened 2025-02-20
Header parameter and model handling does not work as expected


DuplosFidibuss:
### Discussed in https://github.com/fastapi/fastapi/discussions/13160

<div type='discussions-op-text'>

<sup>Originally posted by **freud14-tm** January  6, 2025</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/pydantic/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Annotated, TypeAlias

from fastapi import FastAPI, Header
from pydantic import BaseModel

app = FastAPI()


class CommonHeaders(BaseModel):
    host: str
    save_data: bool
    if_modified_since: str | None = None
    traceparent: str | None = None
    x_tag: list[str] = []


CommonHeadersAnnotated: TypeAlias = Annotated[CommonHeaders, Header()]


@app.get("/items/")
async def read_items(headers: CommonHeadersAnnotated):
    return headers


@app.get("/items2/")
async def read_items2(x_prd_code: Annotated[str, Header()], headers: CommonHeadersAnnotated):
    return headers
```

### Description

There seems to be a few problems when using [Header Parameter Models](https://fastapi.tiangolo.com/tutorial/header-param-models/) as documented. The code above and the screenshots bellow show it. First, it seems that the Swagger UI doesn't convert underscores into dashes when using a Header Parameter Model. It works fine otherwise. Second, when using another header in the function, FastAPI interprets the parameter model as an object to be sent as a single header with the parameters as a comma-separated list instead of expecting the single parameters.

![image](https://github.com/user-attachments/assets/7f387f1d-e737-492a-98ae-752dafa7438b)

![image](https://github.com/user-attachments/assets/30e5eaea-2d5f-4595-b989-46fb6bf89e00)

### Operating System

Linux

### FastAPI Version

0.115.6

[...]

PR #13427, opened 2025-02-27
🐛 Fix extraction of BaseModel fields for openapi

jpjacobpadilla:
Fixes second half of issue #13400


fastapi/dependencies/utils.py
```diff
@@ -210,13 +210,16 @@ def get_flat_dependant(
...
+    extracted_fields = []
+
+    for field in fields:
+        if lenient_issubclass(field.type_, BaseModel):
+            model_fields = get_cached_model_fields(field.type_)
+            extracted_fields.extend(model_fields)
+        else:
+            extracted_fields.append(field)
+
+    return extracted_fields
```
Ale-Cas, line 222: can't you just do a for comprehension and return it? 

```suggestion
    return [
        *get_cached_model_fields(field.type_) 
        if lenient_issubclass(field.type_, BaseModel) 
        else field
        for field in fields
    ]
```

jpjacobpadilla: I don't think you can't unpack an iterable object inside of a list comprehension like that...?

Ale-Cas: You're right that's a syntax error, you could still nest the 2 for loops in a comprehension or use itertools, but your approach is more readable at that point

Ale-Cas (approved):

Whizzo123:
Just testing this change on my end I can see that we are getting a 422 response from the server due to it no longer seeing the header field in the `recieved_params` parameter in the `request_params_to_args` func as we now unpack the model field into it's separate fields here. 
It might be useful to have a test to catch this and see how we can fix the validation step failing in `request_params_to_args` func inside `dependencies/utils.py`

Whizzo123:
Something like this added to the _get_multidict_value func

```python
if value is None and lenient_issubclass(field.type_, BaseModel):
        model_fields = get_cached_model_fields(field.type_)
        value = dict()
        for model_field in model_fields:
            if get_origin(model_field.field_info.annotation) is list:
                value[model_field.alias] = values.get(alias).split(',')
            else:
                value[model_field.alias] = values.get(alias)
```

allows us to grab the values from the recieved params add build a dictionary of values to pass to the model for validation

YuriiMotov:
@jpjacobpadilla, thanks for your interest and efforts!

There are 2 older PRs that aim to solve this problem and they seem to be more developed (cover more cases and have tests):
* https://github.com/fastapi/fastapi/pull/12944 (has unresolved conflicts with `master`)
* https://github.com/fastapi/fastapi/pull/12481 (covers Query, Cookie and Header parameters)

Let's focus on reviewing those two

njsbhu:
I reviewed utils.py, Problem could be with `@dataclass` it doesn’t check if the data you put inside is valid and It just stores values, even if they’re the wrong type. Instead we can use `Pydantic BaseModel` for `ParamDetails ` and `SolvedDependency` this will enable validation.

PR #13444, opened 2025-03-02
📝 Update `docs/en/docs/tutorial/middleware.md`

Rishat-F:
Hello!
Since the “Background Tasks” section already exists, we can add a link to it in "Middleware" section

### DIFF
--- a/docs/en/docs/tutorial/middleware.md
+++ b/docs/en/docs/tutorial/middleware.md
@@ -15,7 +15,7 @@ A "middleware" is a function that works with every **request** before it is proc
 
 If you have dependencies with `yield`, the exit code will run *after* the middleware.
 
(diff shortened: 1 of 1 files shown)


alejsdev (approved):

tiangolo (approved):
Thanks! the note was referring to "you will see the documentation later on this same tutorial), not as a task to do later. :sweat_smile: 

I didn't want people reading it for the first time to feel like they had to go and read those docs right away, that's why I didn't add the link first.

With the new update from @alejsdev I think it keeps that meaning of "documented later" while also adding the link. :nerd_face: 

Thanks all! :cake:

Rishat-F:
@tiangolo Oh, indeed i thought it meant that "Background Tasks" section was absent when "Middleware" section was created😅😅😅

Issue #2619, opened 2021-01-08
Behavior of synchronous path operation functions


pascaldanek:
Hello,

Coming from Scala where the management of asynchronous/synchronous operations is quite explicit, I am struggling to understand what's the impact of using def instead of async def for path operation functions.

My main motivation by using an async framework is to be able to scale a lot on a single machine, being able to support a high number of concurrent http requests with a limited number of system threads.

Using a mature database access library like SQL Alchemy will require to use non async functions a lot, as it does not support async yet (I know there are options but I am not sure I want to use it). The doc claims that then the function runs on a separate thread pool.

Does that mean:
1/ a new thread will be created for each new http client request: in that case I guess the async nature of fastapi just disappear...

2/ a new thread will be created just for a new sql connection/session while keeping the requests handlers async on the main pool ? In that case what kind of thread is it ? is it a fixed thread pool ? is the number of threads configurable ? How will it be managed regarding the number of connections in the connection pool (dead locks ...) ?

Thanks for your help !
--pascal

falkben:
Sqlalchemy has support for async in beta (1.4). See docs [here](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#asynchronous-io-support-for-core-and-orm).

1. A new thread is not created for each new request, if you use a sync function, it is run inside a threadpool.
2. The threadpool size is set by the standard library. It is activated by starlette's [`run_in_threadpool`](https://github.com/encode/starlette/blob/e4307065ea6dbe708fba9643d14fe7adfff06d46/starlette/concurrency.py#L27-L40) which calls [`loop.run_in_executor()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) which by default uses [`ThreadPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor), which you can see has a default max size, depending on Python version.

I think the [async docs](https://fastapi.tiangolo.com/async/) in FastAPI do a very good job of explaining the practical effects of using async. If you're looking for more information, this [video series](https://youtu.be/Xbl7XjFYsN4) by Łukasz Langa, one of the python release managers, does a great job of going into more depth on it.

pikeas:
+1, I was just about to write up a similar question.

I'm familiar with threading and async, but even after reading https://fastapi.tiangolo.com/async/, it's not clear to me what magic FastAPI is doing under the hood when it comes to making async def/def just work.

The docs say: 

[...]

Issue #953, opened 2020-02-07
[FEATURE] Ability to change contextvars in sync dependencies


smagafurov:
Sync dependencies now run in thread pool with `starlette.concurrency.run_in_threadpool`.

`run_in_threadpool` run in the context copy (`contextvars.copy_context().run(...)`)

```
async def run_in_threadpool(
    func: typing.Callable, *args: typing.Any, **kwargs: typing.Any
) -> typing.Any:
    loop = asyncio.get_event_loop()
    if contextvars is not None:  # pragma: no cover
        # Ensure we run in the same context
        child = functools.partial(func, *args, **kwargs)
        context = contextvars.copy_context()  # <------------------ here
        func = context.run
        args = (child,)
    elif kwargs:  # pragma: no cover
        # loop.run_in_executor doesn't accept 'kwargs', so bind them in here
        func = functools.partial(func, **kwargs)
    return await loop.run_in_executor(None, func, *args)
```

But if we change contextvars inside dependency (func), this changes are lost. We don't see this changes in route func.

But I want to see this changes :)

May be we can propagate (copy) contextvars changes to main thread contextvars  after dependency finish?

dmontagu:
I think this should be a feature request for `starlette` rather than FastAPI; it might be worth creating an issue there. @tomchristie might know of a good pattern for accomplishing this (or whether it is even possible), or might be able to provide reasons why you should avoid this pattern.

tomchristie:
(No idea)

smagafurov:
> I think this should be a feature request for `starlette` rather than FastAPI; it might be worth creating an issue there. @tomchristie might know of a good pattern for accomplishing this (or whether it is even possible), or might be able to provide reasons why you should avoid this pattern.

dependency - is a subject of fastapi not starlette

Logically dependency should have the same context as route handler.

If dependency is async - it works correct (have the same context).

Sync dependency and async dependency have the same logic.

So async dependency should have the same context too.

There is problem with sync dependency only, because of technical realization of sync dependency (they called with run_in_threadpool). So we should think about this problem in fatsapi (not in starlette)

tiangolo:
Oh, that's a heavy hardcore subject :sweat_smile: 

[...]

Issue #555, opened 2019-09-22
Check if the request context is available


raghavio:
Is there a way to check if the `request` variable is available to use?
I'm coming from Flask and my codebase gets run from the API and inside RQ Workers (background queue).
Inside RQ Workers, the SQLAlchemy db session is maintained per worker and for API it's at the request level.
To get the session I have a function that looks something like this:

```
def get_session():
    if has_request_context():
        # Get it from the request. In case of Fast this will be probably be Depends(request.state.db)
    elif has_worker_context():
        # Pick it from a local variable
```

To run a query we simply call the above function to get the session.

How can I achieve something like this in Fastapi?

dmontagu:
It depends on how you have things implemented.

In general, even if it is somehow possible, I think it may force an awkward design if you try to use the *same* `get_session` function to get a session inside both your fastapi endpoints and inside your workers.

I would recommend you write functions that *accept* a session as an argument, and then use a fastapi `Depends` to inject the session in your endpoints, and use an alternative approach to get a session in your workers.

This *could* be as simple as calling the fastapi endpoint functions while specifying a non-default value for the session (so that the `Depends` is not passed to the function) in the workers. Or it could mean having your fastapi endpoints basically just being wrappers around a single set of functions either the api or workers can call, if some minor amount of api-specific preparation is required.

I think it might be hard to give a better answer than that without a little more detail about the api and worker design.

dmontagu:
If you *really* want to go down this route, you could maybe do something along the following lines:
```python
def get_fastapi_request(request: Request):
    return request

def get_session(request: Request = Depends(get_fastapi_request)) -> Session:
    if isinstance(request, fastapi.params.Depends):
        # default value was passed, so in worker context
        ...
    else:
        assert isinstance(request, Request)
        # fastapi injected the value, so in a dependency
        ...
```
But again, I would encourage you to think if there is a way to design this that avoids this check.

RaghavPro:
Thanks for this.

--
We have 130k lines of code and doing the refactor of injecting the session through function parameters would be too much of a work.

The `get_session` is simply a helper function that picks the session depending on where the code was run from. Which can be 3 places. From the api request, worker and python console or jupyter environment. It allows to run the exact same code without any pre-requisites of starting an app or instantiating some objects.

It works well for us in Flask and I'd like to stick to it unless there's something very wrong with this design being used with Fast.

dmontagu:
I understand if it is due to legacy reasons.

[...]

PR #13453, opened 2025-03-05
📝 Add caveat about ContextVar in dependencies

JP-Ellis:
As discovered and discussed in #13382, ContextVars set in a _synchronous_ dependency is not available to the rest of the request due to the way FastAPI handles that dependency.

The documentation did not make any mention of this, and it was only after I asked that I discovered this was intended behaviour. So I am updating the docs to make this clear to others.

Ref: #13382


JP-Ellis:
Note, I've added AI-generated translations of the ContextVar warning. These might be an acceptable starting point, but should be separately reviewed.

docs/en/docs/tutorial/dependencies/index.md
```diff
@@ -156,6 +156,12 @@ If you don't know, check the [Async: *"In a hurry?"*](../../async.md#in-a-hurry)
 
 ///
 
+/// warning
+
+If you need to rely on `ContextVar` values across multiple steps in a request, **use an async dependency** or another approach. This is because sync dependencies are executed in the mainthread*, while `async` dependencies run in the same thread pool as the one handling the request. As `ContextVar` provides thread-local storage, any value set inside a sync dependency will not be available in the rest of the request handling.
```
YuriiMotov: I think this is not clear enough.
It's suggested to use `async` dependency, but what if our endpoint function is `sync`? Will we be able to access the context var that was set in dependency from the endpoint function?

It's probably worth adding a dedicated page to describe the behavior of `ContextVar` in FastAPI with examples.
Such short explanation will be confusing for newcomers.

Alternatively, I would add the admonition to avoid using `ContextVar` unless the developer is experienced enough and understands how FastAPI works internally

YuriiMotov (changes requested):
@JP-Ellis, thanks for the interest in FastAPI and for the efforts!

Translations should be updated in separate PRs (it will be done automatically later). We should only update the original document (En) in this PR.

Also, I think this explanation is not clear enough. Let's discuss it

JP-Ellis:
Thanks for the review! I can certainly remove the translations from the PR.

As for the explanations itself, let me address some points you raise.

> It's suggested to use async dependency, but what if our endpoint function is sync? Will we be able to access the context var that was set in dependency from the endpoint function?

Short answer: yes. This is, at least, based on my testing done in:

- https://github.com/fastapi/fastapi/discussions/13382

> It's probably worth adding a dedicated page to describe the behaviour of ContextVar in FastAPI with examples.
> Such short explanation will be confusing for newcomers.

I think that is somewhat beyond the scope of this initial PR that is merely adding a caveat. I am also not intimately familiar with the exact internals of FastAPI and Starlette. I can certainly expand on the explanation here, especially given that the fix is quite straightforward.

> Alternatively, I would add the admonition to avoid using ContextVar unless the developer is experienced enough and understands how FastAPI works internally

In general, I tend to shy away from exclusionary language like "Don't do X unless you know what you're doing". I am more than happy to elaborate on the _why_ though.

What about the following admonition which expands on what I had originally, while providing more detail about the _why_?

> [!WARNING]
>
> **ContextVar Propagation in FastAPI**
>
> Python's [`ContextVar`](https://docs.python.org/3/library/contextvars.html) provides thread-safe shared storage as documented in [PEP 567](https://peps.python.org/pep-0567/). 
>
> - `ContextVar` values set in synchronous (`def`) dependencies are **not** accessible in the rest of the request lifecycle as FastAPI executes them in isolated threads.
> - To ensure `ContextVar` values are available throughout the request (including other dependencies, provided they run in the right order), use **asynchronous (`async def`) dependencies**.
> - If you must use synchronous dependencies, prefer passing data explicitly or using [`Request.state`](https://fastapi.tiangolo.com/advanced/middleware/#using-the-request-object-in-a-middleware) for request-scoped storage.
> - **Troubleshooting tip:** If a `ContextVar` value is unexpectedly `None` or missing, check whether it was set in a synchronous dependency.

YuriiMotov:
The fact that contextvars works in combination of `async` dependency and `sync` endpoint function surprized me.
I'm not sure this behavior is stable. To state this we should find tests that cover this case.

New version of Warning is definitely more clear, but.. I still think we should either investigate this in-depth and create a separate page with code examples or add an admonition to refrain from using `ContextVars` with dependencies if it's possible.

Below (in the details) there are several comments of Sebastián about `ContextVars` I could find.

<details>

https://github.com/fastapi/fastapi/pull/5489#issuecomment-1310957653

> ContextVars are a very delicate and nuanced subject/tool. It's very easy to end up with hard to debug problems, even more when interacting with threads, different async task contexts, etc. And you do all that with FastAPI, knowingly or unknowingly. For example, if you have any Middleware based on BaseHTTPMiddleware, as that creates a new task group, it creates a copy of the context vars context for the task, so, changes are not propagated upwards. The same way, you cannot set a context var in a sync function as that would be executed in a thread worker, it would have to be set in an async function that is run on the main thread running the event loop, etc. ...so, I recommend avoiding them when possible and only using them when strictly necessary.
---

https://github.com/fastapi/fastapi/issues/953#issuecomment-586006249

This one looks outdated

---

https://github.com/fastapi/fastapi/issues/555#issuecomment-546698061


---

https://github.com/fastapi/fastapi/issues/2619#issuecomment-1359044461

> But, you should probably stay away from thread-locals and contextvars and all that stuff as much as possible, use local variables, pass values in parameters, etc. It will make your life soooo much easier, your code easier to debug, and the extra time you spend typing the extra parameter in the function will save you orders of magnitude the same time debugging why a pseudo-global is not working correctly (from personal experience 😅 ).

---

</details>

You can at least remove the changes in translations and improve the Warning message. Then we will ask Sebastián to decide

JP-Ellis:
> I still think we should either investigate this in-depth 

I have done an early investigation on the permutations in [this discussion](https://github.com/fastapi/fastapi/discussions/13382) which is why I think this warning is warranted.

> and create a separate page with code examples or add an admonition to refrain from using ContextVars with dependencies if it's possible.

While I wholeheartedly agree that they can be difficult to use, make some logic flows difficult, and are inherently hard to debug, they have their place. It isn't always possible to avoid their use.

As a concrete example, if you want to insert request-specific information into _all_ logging calls (from all libraries), then it is best to define a custom log formatter/handler which inserts this additional data from a `ContextVar` with every call to the logger.

YuriiMotov (commented):
Still not convinced that this is the right way to go, but let's ask for third opinion

YuriiMotov (commented):
Still not convinced that this is the right way to go, but let's ask for third opinion

nidhishgajjar:

## Review of PR #13453: Add caveat about ContextVar in dependencies

This is a valuable documentation addition that addresses a real and confusing gotcha with ContextVars in FastAPI dependencies.

### What looks good ✅

- **Accurate problem description**: Sync dependencies run in isolated threads via `run_in_threadpool`, and while the context is *copied* to the thread, mutations (new values set) don't propagate back to the calling coroutine. This is a subtle behavior that's easy to miss.
- **Practical workarounds**: Suggesting `async def` dependencies, explicit parameter passing, and `Request.state` covers the main alternatives
- **Troubleshooting tip**: "If a ContextVar value is unexpectedly `None` or missing, check whether it was set in a synchronous dependency" — this is exactly the kind of hint that saves hours of debugging
- **Well-placed**: Right after the sync/async explanation section, where it directly relates to the threadpool behavior

### Minor suggestion 💡

The `Request.state` link uses an absolute URL (`https://fastapi.tiangolo.com/...`). FastAPI docs typically use relative references to stay consistent and work across deployments. Consider converting to a relative link if the target exists in the local docs.

### Summary

Clear, accurate documentation of a subtle but important behavior difference between sync and async dependencies. This will save developers significant debugging time.

**Assessment: approve**

PR #13457, opened 2025-03-05
🌐 Add Romanian translation for `docs/ro/docs/index.md`

ivintoiu:


docs/en/mkdocs.yml
```diff
@@ -6,7 +6,7 @@ theme:
   name: material
   custom_dir: ../en/overrides
   palette:
-  - media: "(prefers-color-scheme)"
+  - media: (prefers-color-scheme)
```
svlandeg, line 9: Typically, there shouldn't be any changes to `docs/en/mkdocs.yml` when contributing a translation page for another language. Can you revert these, to allow easier reviewing of the Romanian part?

svlandeg: So yes, I would expect the link to be added, but I'm confused why other formatting has happened, and comments have been removed etc.

svlandeg: This has now been fixed with https://github.com/fastapi/fastapi/pull/13542 😉

tiangolo:
Thank you very much for the interest and effort! :cake: 

Sadly, there are many incomplete and obsolete translations, so we want to update how translations are handled. We will have some main languages, where the translations are automated with LLMs and reviewed by humans. You can read more about it here: https://fastapi.tiangolo.com/contributing/#translations

If there are several people (let's say 5) commenting on https://github.com/fastapi/fastapi/discussions/13458 and committing to help review new (generated) translations, I would make it one of the official (automatically translated) languages.

For now, I will close this one, but thanks for the interest! :coffee:

PR #13463, opened 2025-03-06
📝 Add windows example for running fastapi to `docs/en/docs/tutorial/first-steps.md`

holts-shoe:
Add termy tab for windows users to make it easier to run fastapi for the first time


alv2017:
FastAPI CLI is a Python program, that gets installed together with FastAPI, hence I would expect that it uses the same commands on Linux, MacOS and Windows.

tiangolo:
Thanks for the interest! This is a bit specific to one tool in Windows, and I would actually prefer to teach people to install uv, so I'll pass on this one for now. :coffee:

PR #13464, opened 2025-03-07
🐛 Don't prefill default values on form input

### LINKED ISSUE #13399: Dependency Models created from Form input data are loosing metadata(field set) and are enforcing validation on default values.
### Discussed in https://github.com/fastapi/fastapi/discussions/13380

<sup>Originally posted by **sneakers-the-rat** February 16, 2025</sup>
### First Check

### Commit to Help

### Example Code

fastapi_defaults_bug.py
```python
import uvicorn
from typing import Annotated

class ExampleJsonModel(BaseModel):
    sample_field_1: Annotated[bool, Field(default=True)]
    sample_field_2: Annotated[bool, Field(default=False)]
    sample_field_3: Annotated[bool, Field(default=None)]
    sample_field_4: Annotated[str, Field(default=0)] # This is dangerous but can be used with a validator


class ExampleFormModel(BaseModel):
    sample_field_1: Annotated[bool, Form(default=True)]
    sample_field_2: Annotated[bool, Form(default=False)]
    sample_field_3: Annotated[bool, Form(default=None)]
    sample_field_4: Annotated[str, Form(default=0)] # This is dangerous but can be used with a validator


class ResponseSampleModel(BaseModel):
    fields_set: Annotated[list, Field(default_factory=list)]
    dumped_fields_no_exclude: Annotated[dict, Field(default_factory=dict)]
    dumped_fields_exclude_default: Annotated[dict, Field(default_factory=dict)]
    dumped_fields_exclude_unset: Annotated[dict, Field(default_factory=dict)]



app = FastAPI()


@app.post("/form")
async def form_endpoint(model: Annotated[ExampleFormModel, Form()]) -> ResponseSampleModel:
    return ResponseSampleModel(
        fields_set=list(model.model_fields_set),
        dumped_fields_no_exclude=model.model_dump(),
        dumped_fields_exclude_default=model.model_dump(exclude_defaults=True),
        dumped_fields_exclude_unset=model.model_dump(exclude_unset=True)
    )

@app.post("/json")
async def form_endpoint(model: ExampleJsonModel) -> ResponseSampleModel:
    return ResponseSampleModel(
        fields_set=list(model.model_fields_set),
        dumped_fields_no_exclude=model.model_dump(),
        dumped_fields_exclude_default=model.model_dump(exclude_defaults=True),
        dumped_fields_exclude_unset=model.model_dump(exclude_unset=True)
    )


if __name__ == "__main__":

```

Test File: test_fastapi_defaults_bug.py
```python
import pytest

from fastapi_defaults_bug import (
    app,
    ExampleFormModel,
    ExampleJsonModel,
    ResponseSampleModel
)

@pytest.fixture(scope="module")
def fastapi_client():
    with TestClient(app) as test_client:
        yield test_client


################
# Section 1: Tests on Form model -> no fastapi, pydantic model
################

def test_form_model_pydantic_only_defaults():
    f_model = ExampleFormModel()
    for field_name, field in f_model.model_fields.items():
        assert getattr(f_model, field_name) == field.default

def test_form_model_pydantic_all_unset():
    f_model = ExampleFormModel()
    assert not f_model.model_fields_set

def test_form_model_pydantic_set_1():
    f_model = ExampleFormModel(sample_field_1=True) # Those set have the same value of default
    assert "sample_field_1" in f_model.model_fields_set
    assert len(f_model.model_fields_set) == 1

def test_form_model_pydantic_set_2():
    f_model = ExampleFormModel(sample_field_1=True, sample_field_2=False) # Those set have the same value of default
    assert "sample_field_1" in f_model.model_fields_set
    assert "sample_field_2" in f_model.model_fields_set
    assert len(f_model.model_fields_set) == 2

def test_form_model_pydantic_set_all():
    f_model = ExampleFormModel(
        sample_field_1=True,
        sample_field_2=False,
        sample_field_3=True,
        sample_field_4=""
    ) # Those set could have different values from default
    assert not set(f_model.model_fields).difference(f_model.model_fields_set)

################
# Section 2: Same Tests of Form on Json model -> they are the same on different model
################

def test_json_model_pydantic_only_defaults():
    j_model = ExampleJsonModel()
    for field_name, field in j_model.model_fields.items():
        assert getattr(j_model, field_name) == field.default

def test_json_model_pydantic_all_unset():
    j_model = ExampleJsonModel()
    assert not j_model.model_fields_set

def test_json_model_pydantic_set_1():
    j_model = ExampleJsonModel(sample_field_1=True) # Those set have the same value of default
    assert "sample_field_1" in j_model.model_fields_set
    assert len(j_model.model_fields_set) == 1

def test_json_model_pydantic_set_2():
    j_model = ExampleJsonModel(sample_field_1=True, sample_field_2=False) # Those set have the same value of default
    assert "sample_field_1" in j_model.model_fields_set
    assert "sample_field_2" in j_model.model_fields_set
    assert len(j_model.model_fields_set) == 2

def test_json_model_pydantic_set_all():
    j_model = ExampleJsonModel(
        sample_field_1=True,
        sample_field_2=False,
        sample_field_3=True,
        sample_field_4=""
    ) # Those set could have different values from default
    assert not set(j_model.model_fields).difference(j_model.model_fields_set)

def test_form_json_model_share_same_default_behaviour():
    f_model = ExampleFormModel()
    j_model = ExampleJsonModel()

    for field_name, field in f_model.model_fields.items():
        assert getattr(f_model, field_name) == getattr(j_model, field_name)


################
# Section 3: Tests on Form model with fastapi
################


def test_submit_form_with_all_values(fastapi_client: TestClient):

    form_content = {
        "sample_field_1": "False",
        "sample_field_2": "True",
        "sample_field_3": "False",
        "sample_field_4": "It's a random string"
    }

    response = fastapi_client.post("/form", data=form_content)

    response_model = ResponseSampleModel(**response.json())
    assert len(response_model.fields_set) == 4
    assert not set(form_content).symmetric_difference(set(response_model.fields_set))


def test_submit_form_with_not_all_values(fastapi_client: TestClient):
    """
    This test should pass but fails because fastapi is preloading default and pass those values
    on model creation, losing the ability to know if a field has been set.

    :param fastapi_client:
    :return:
    """

    form_content = {
        "sample_field_1": "False",
        "sample_field_3": "False",
        "sample_field_4": "It's a random string"
    }

    response = fastapi_client.post("/form", data=form_content)

    response_model = ResponseSampleModel(**response.json())
    assert len(response_model.fields_set) == 3 # test will fail here and below
    assert not set(form_content).symmetric_difference(set(response_model.fields_set))


def test_submit_form_with_no_values(fastapi_client: TestClient):
    """
    This test should pass but fails because fastapi is preloading default and pass those values
    on model creation, losing the ability to not have validation on default value.

    :param fastapi_client:
    :return:
    """

    form_content = {}

    response = fastapi_client.post("/form", data=form_content)
    assert response.status_code == 200 # test will fail here and below -> will raise 422

    response_model = ResponseSampleModel(**response.json())
    assert len(response_model.fields_set) == 0
    assert not set(form_content).symmetric_difference(set(response_model.fields_set))


################
# Section 4: Tests on Json model with fastapi
################


def test_submit_json_with_all_values(fastapi_client: TestClient):

    json_content = {
        "sample_field_1": False,
        "sample_field_2": True,
        "sample_field_3": False,
        "sample_field_4": "It's a random string"
    }

    response = fastapi_client.post("/json", json=json_content)

    response_model = ResponseSampleModel(**response.json())
    assert len(response_model.fields_set) == 4
    assert not set(json_content).symmetric_difference(set(response_model.fields_set))


def test_submit_json_with_not_all_values(fastapi_client: TestClient):
    """
    This test will pass but the same not happen with Form.

    :param fastapi_client:
    :return:
    """

    json_content = {
        "sample_field_1": False,
        "sample_field_3": False,
        "sample_field_4": "It's a random string"
    }

    response = fastapi_client.post("/json", json=json_content)

    response_model = ResponseSampleModel(**response.json())
    assert len(response_model.fields_set) == 3 # This time will not fail
    assert not set(json_content).symmetric_difference(set(response_model.fields_set))


def test_submit_json_with_no_values(fastapi_client: TestClient):
    """
     This test will pass but the same not happen with Form.

    :param fastapi_client:
    :return:
    """

    json_content = {}

    response = fastapi_client.post("/json", json=json_content)
    assert response.status_code == 200 # This time will not fail

    response_model = ResponseSampleModel(**response.json())
    assert len(response_model.fields_set) == 0
    assert not set(json_content).symmetric_difference(set(response_model.fields_set))

```

### Description

This is a generalized version of the issue reported in https://github.com/fastapi/fastapi/discussions/13380 .
This issue do not affect body json data.

For models created from a form, during the parsing phase their default values are preloaded and passed to the validator to create the model. 
1) This leads to a loss of information regarding which fields have been explicitly set, since default values are now considered as having been provided.
12) Consequently, validation is enforced on default values, which might not be the intended behavior and anyway different from the one from json body.

### Operating System

macOS - Linux

### Operating System Details

_No response_

### FastAPI Version

0.115.8

### Pydantic Version

2.10.6

### Python Version

Python 3.11 - Python 3.13.1

sneakers-the-rat:
Fix: https://github.com/fastapi/fastapi/issues/13399
Discussed in: https://github.com/fastapi/fastapi/discussions/13380

## summary

Context in above issue and discussion.

This PR makes JSON and Form data behave consistently, not prefilling default values but instead allowing the model to do that so that model validators are able to differentiate between unset and default values. 

## motivation

This is particularly important for form encoded data, as some fields like checkboxes do not have behavior that would be expected given the model type, and so it becomes impossible to decouple the semantics of the model (where it is entirely reasonable to have defaults for boolean fields, e.g. if one was programmatically generating a form from a model, one would use the default to set the initial state of a checkbox) from form parsing.

## impl notes

I did not include all the test cases from #13399 since many of them were just testing pydantic model behavior which seems out of scope for fastapi to test, but i did make use of some of the ideas from there (thanks for getting this started @luzzodev !!!), and it should be easy to add them in if we want to have them.

edit: re: [this comment](https://github.com/fastapi/fastapi/pull/13464#discussion_r2209382217) requesting clarification on changes to [_extract_form_body](https://github.com/fastapi/fastapi/pull/13464/files#diff-aef3dac481b68359f4edd6974fa3a047cfde595254a4567a560cebc9ccb0673fR841), 
see [this comment explaining the problem](https://github.com/fastapi/fastapi/pull/13464#issuecomment-2708378172) and [my comment explaining the resolution](https://github.com/fastapi/fastapi/pull/13464#issuecomment-2708617396). `values` is populated in two stages - first any values in `received_body` that have defined annotations from a model/field (`body_fields`) are extracted using the annotation, and second any values that do *not* have a corresponding annotation are added as-is for downstream validation. Previously, the first stage would populate `values` with defaults, and so in the second stage we could check for "extra" values with `if key not in values`. However since now we are *not* prepopulating the defaults (the point of the PR), in the case of empty strings (a form-encoding artifact), `if key not in values` would just add them back in (defeating the whole point of the PR), and so we needed a different method of preserving the `received_body` fields without accompanying `body_fields` annotations. So the change is incredibly minor - we just get a set of field aliases and check against that, rather than checking against what is already in `values`. this is a narrowly focused change that should *only* make it so we don't include empty strings in `values` when processing form input.

## future directions

This is the most minimal form of the proposed changes, which allows someone to use a `model_validator` in `before` mode to differentiate unset and default values. 

A more complete fix would be to a) make a `@form_validator` that receives the raw form input before normal model validation, and b) if no `@form_validator` is declared for a model, use the field annotation information to infer how to process form encoded data (e.g. for a `bool` field, if the key is absent in the input, interpret as `False`. for a `list[str]` field, treat as an empty list, etc.). I would be happy to continue on and implement these improvements, which would make form handling a lot more explicit, but i'll hold here until i get approval bc this also is sufficient for my purposes for now.


sneakers-the-rat:
There is a failing check from not having a PR label, i see other bugs have the 🐛 emoji prefix (cute! love it!) so I added it, but it doesn't look like the labeler action detects that, and i don't have privs to add a label.  didn't see anything in the contributing docs on this, but sorry if i missed something <3

luzzodev:
Hi @sneakers-the-rat, thanks for the PR first of all.

About your proposal: while on one side this could offer the tools to handle what you asking at https://github.com/fastapi/fastapi/discussions/13380, on the other side it will not address the more general issue https://github.com/fastapi/fastapi/issues/13399.

As i said on the issue discussion, a good and more robusts solution could be harder and require more work than you think.
Since the changes are on a critical function that is used by half of the possibile inputs ( Headers, Path, Query, etc) first i would write a lot more tests to be sure that the actual behaviour will be preserved.

Technical Comments:
I haven't take a look in depth, so there could be more but with your proposal there will be for sure a breaking change on
https://github.com/fastapi/fastapi/blob/643d2845de09a437b955a163ab25eb0887bfd0e1/fastapi/dependencies/utils.py#L727-L729

I attach a small simple test that should pass but it will not. 
In that specific case you will have a different behaviour between alias and the original field name.

``` python

class SimpleForm(BaseModel):
    foo: Annotated[str, Field(default="bar")]
    alias_with: Annotated[str, Field(alias="with", default="nothing")]


@app.post("/simple-form")
def form_endpoint(model: Annotated[SimpleForm, Form()]) -> dict:
    return model.model_dump()


def test_casted_empty_defaults(fastapi_client: TestClient):
    form_content = {
        "foo": "",
        "alias_with": ""
    }
    response = fastapi_client.post("/simple-form", data=form_content)
    response_content = response.json()
    assert response_content["foo"] == "bar"  # Expected :'bar' -> Actual   :''
    assert response_content["alias_with"] == "nothing"  # ok
```

sneakers-the-rat:
> As i said on the issue discussion, a good and more robusts solution could be harder and require more work than you think.

Great, yeah i was relying on the tests to be protective of any behavior that needs to be preserved, so if there are things that would break that aren't tested then of course list them and we can add tests.

> on the other side it will not address the more general issue https://github.com/fastapi/fastapi/issues/13399.

I am not sure how what you described there is more general than what i intitially proposed or have PR'd here, but ofc please let me know what needs to be fixed. This PR passes all the tests in #13399 except for `test_submit_form_with_no_values`, which emits an warning because the pydantic model is invalid. Specifically, it's a serialization warning, where a model that is correctly constructed with the defaults can't be serialized  - it being a warning, the correct value would be returned from the client anyway, the tests just treat warnings as failures. 

The current code before this PR also fails this test. So if we want to make fixing that an additional bug we handle in this PR, i can do that, but in my opinion this is actually correct behavior for a misspecified model.

> Since the changes are on a critical function that is used by half of the possibile inputs ( Headers, Path, Query, etc) first i would write a lot more tests to be sure that the actual behaviour will be preserved.

The change i have made only affects the `Form` input and is equivalent to the current behavior for the other types. 

The old code was

```python
if field.required:
    return
else:
    return deepcopy(field.default)
```

The new code is

```python
if return_default and not field.required:
    return deepcopy(field.default)
else:
    return None
```

Notice how when `return_default == True` the behavior is identical, the ordering has just been reverse to group the "deepcopy the default" behavior in the same block as evaluating "return_default". 

the default value of `return_default` is `True`, and it is only explicitly made false inside of `_extract_form_body`, which I hope is only called for Form data input.


> I haven't take a look in depth, so there could be more but with your proposal there will be for sure a breaking change on

This is not a breaking change, as this also fails in the current code, but again i can fix the handling of aliases as an additional bug if we want to attach it to this PR as well.

luzzodev:
Hello @sneakers-the-rat , i-m sorry, i have mixed tests, the following one is the right one for your case.

``` python
app = FastAPI()


@pytest.fixture(scope="module")
def fastapi_client():
    with TestClient(app) as test_client:
        yield test_client


class SimpleForm(BaseModel):
    foo: Annotated[str, Form(default="bar")]
    alias_with: Annotated[str, Form(alias="with", default="nothing")]


@app.post("/simple-form")
def form_endpoint(model:Annotated[SimpleForm, Form()]) -> dict:
    return model.model_dump()


def test_casted_empty_defaults(fastapi_client: TestClient):
    form_content = {
        "foo": "",
        "with": ""
    }
    response = fastapi_client.post("/simple-form", data=form_content)
    response_content = response.json()
    assert response_content["foo"] == "bar"  # Expected :'bar' -> Actual   :''
    assert response_content["alias_with"] == "nothing"  # ok
```

Right now is possible to use the given format and in some way, this is the one used for File and UploadFile.
Technically your changes will be reflected on the body params extraction phase on inputs that comes from: 
- 'multipart/form-data'
- 'application/x-www-form-urlencoded'

With your implementation you are disabling any possible handling of 
https://github.com/fastapi/fastapi/blob/643d2845de09a437b955a163ab25eb0887bfd0e1/fastapi/dependencies/utils.py#L724-L732

And this will allow original values to come back in 
https://github.com/fastapi/fastapi/blob/643d2845de09a437b955a163ab25eb0887bfd0e1/fastapi/dependencies/utils.py#L870-L873

This is exactly what i reported in last test.

sneakers-the-rat:
Ah i see. 

So the 

These lines: https://github.com/fastapi/fastapi/blob/643d2845de09a437b955a163ab25eb0887bfd0e1/fastapi/dependencies/utils.py#L870-L873

seem incorrect: the purpose they seem to serve is to get any additional fields that are passed in the form body, but are not included in the form model. currently they are too general - the test is "`if key not in values`", which would make sense in the old behavior where defaults would be auto-filled (and explains a bit why they were). 

It should actually be `if key not in body_fields` (but since `body_fields` can't do `in` checks like that, the aliases have to be extracted out).

I added your test and now it passes. 

---

The core of the problem appears to be regarding empty string input, so i added an additional test that tests empty string input against all the possible ways of specifying a form. The docs show the ability to specify form data in two ways:

1) [as parameters in the endpoint signature](https://fastapi.tiangolo.com/tutorial/request-forms/#define-form-parameters):

```python
@app.post("/login/")
async def login(username: Annotated[str, Form()], password: Annotated[str, Form()]):
    return {"username": username}
```

2) [as pydantic models](https://fastapi.tiangolo.com/tutorial/request-form-models/):

```python
class FormData(BaseModel):
    username: str
    password: str

@app.post("/login/")
async def login(data: Annotated[FormData, Form()]):
    return data
```

(I actually can't find the example case you're using of specifying `Annotated[type, Form()]` within a pydantic model in the docs. That seems like it should be technically possible, but from a data modeling POV would be a violation of separation of concerns - the *model* shouldn't care about the encoding of its inputs, that's something that should be handled outside the model, in the framework. regardless, the current version in the PR supports it.)

So when extracting the data, the current check of whether the *field* is a `Form()` type is inadequate: https://github.com/fastapi/fastapi/blob/643d2845de09a437b955a163ab25eb0887bfd0e1/fastapi/dependencies/utils.py#L727

We also care about whether the *containing field* is a `Form()` (i.e., in the (2) case of specifying form input with a model where the model fields are not themselves `Form()`s)

So see the test run for [`7fade13`](https://github.com/fastapi/fastapi/pull/13464/commits/7fade13ac8696e529c070a42349951cdba8b4987): 
https://github.com/fastapi/fastapi/actions/runs/13743551022/job/38435686000 - the test above passes with the `Form()` field annotations, but fails for the rest of the methods of specifying a form.

So I revised the prior impl semantics of `return_default` to instead explicitly specify whether we are in the context of extracting form data. This, again, only applies to form input and does not affect other input types. 

To be clear, again - this is a bug in the current implementation, not introduced by this PR. The tests in `test_empty_string_inputs` fail in the current implementation.

fastapi/dependencies/utils.py
```diff
@@ -867,8 +871,10 @@ async def process_fn(
             value = serialize_sequence_value(field=field, value=results)
         if value is not None:
             values[field.alias] = value
+
+    # preserve extra keys not in model body fields for validation
     for key, value in received_body.items():
-        if key not in values:
+        if key not in field_aliases:
```
luzzodev, line 877: This is the right thing to do here =) and will solve even already existing bug, where None is returned =)

fastapi/dependencies/utils.py
```diff
@@ -867,8 +871,10 @@ async def process_fn(
             value = serialize_sequence_value(field=field, value=results)
```
luzzodev, line 871: Probably we need more tests. Here things has changing because we are not returning defaults anymore.

luzzodev (changes requested):
Overall i think this will solve a lot of issues related to FormFields with default, is now possibile to keep track if a field has been set or not. Is even possible to define a default placeholder that is not compatible with defined type. This option need to be well documented because default values are sent to the doc.

fastapi/dependencies/utils.py
```diff
@@ -724,14 +727,14 @@ def _get_multidict_value(
     if (
         value is None
         or (
-            isinstance(field.field_info, params.Form)
+            (isinstance(field.field_info, params.Form) or form_input)
             and isinstance(value, str)  # For type checks
             and value == ""
         )
         or (is_sequence_field(field) and len(value) == 0)
     ):
```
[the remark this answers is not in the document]
sneakers-the-rat: not sure what you mean? that `or form_input` is necessary for the case where one does

```python
class MyModel(BaseModel):
    value: str = "default"

@app.post('/endpoint')
async def myendpoint(model: Annotated[MyModel, Form()]):
    pass
```

> So when extracting the data, the current check of whether the field is a Form() type is inadequate:
> 
> [fastapi/fastapi/dependencies/utils.py](https://github.com/fastapi/fastapi/blob/643d2845de09a437b955a163ab25eb0887bfd0e1/fastapi/dependencies/utils.py#L727)
> 
> Line 727 in [643d284](https://github.com/fastapi/fastapi/commit/643d2845de09a437b955a163ab25eb0887bfd0e1)
>  isinstance(field.field_info, params.Form) 
> 
> We also care about whether the containing field is a Form() (i.e., in the (2) case of specifying form input with a model where the model fields are not themselves Form()s)

See the test run for https://github.com/fastapi/fastapi/actions/runs/13743551022/job/38435686000

fastapi/dependencies/utils.py
```diff
@@ -724,14 +727,14 @@ def _get_multidict_value(
...
-        if field.required:
-            return
+        if form_input or field.required:
+            return None
```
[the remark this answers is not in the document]
sneakers-the-rat: it's a style question, but keeping the None is clearer to me - this isn't the code that actually differentiates `None` from unset, that happens here: https://github.com/fastapi/fastapi/blob/643d2845de09a437b955a163ab25eb0887bfd0e1/fastapi/dependencies/utils.py#L868

so since the check there is for `is None`, returning an explicit None is clearer. When not used for Form data, a `None` has a different meaning than unset.

this is also the only bare `return` statement in the module, everywhere else uses `return None`.

they're equivalent, imo this is clearer, seems like a wash to me

fastapi/dependencies/utils.py
```diff
@@ -835,11 +838,12 @@ async def _extract_form_body(
     received_body: FormData,
 ) -> Dict[str, Any]:
     values = {}
+    field_aliases = {field.alias for field in body_fields}
```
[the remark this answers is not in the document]
sneakers-the-rat: iteration without side effects is super cheap. i bet if you profiled this code this wouldn't even show up since it's such a small use of time. this keeps the creation of the variable in one place. we certainly could do this in the for loop, inconsequential perf benefit and minor modularity loss

tests/test_forms_single_model.py
```diff
@@ -104,13 +104,13 @@ def test_no_data():
                     "type": "missing",
                     "loc": ["body", "username"],
                     "msg": "Field required",
-                    "input": {"tags": ["foo", "bar"], "with": "nothing"},
+                    "input": {},
```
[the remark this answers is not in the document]
sneakers-the-rat: it's just what the old behavior was, we are changing the behavior here, so if this PR is accepted, the tests should change too

fastapi/dependencies/utils.py
sneakers-the-rat: sure, name whatever tests need to be written. if the tests aren't already there to protect existing functionality, i'm not sure how i would know what should break. happy to write more tests but at some point we do need to limit the scope of this PR so it doesn't turn into "fix every problem with form handling"

fastapi/dependencies/utils.py
```diff
@@ -714,7 +714,10 @@ def _validate_value_with_model_field(
 
 
 def _get_multidict_value(
-    field: ModelField, values: Mapping[str, Any], alias: Union[str, None] = None
+    field: ModelField,
+    values: Mapping[str, Any],
+    alias: Union[str, None] = None,
+    form_input: bool = False,
```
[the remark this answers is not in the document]
sneakers-the-rat: the reason i changed it is because the behavior we need isn't just falling back on default, it's also treating empty strings as `None`s, which is specific to form input - Form fields in HTML submit an empty string when they're empty, but JSON input doesn't do that, so we handle empty strings differently there. If we need additional variation on this function we can deal with it when we need to imo.

sneakers-the-rat:
will add docs tomorrow, i had already written a short example in the discussion, so i can improve that and include it in docs.

luzzodev (commented):
The proposed solution resolves the mentioned bug and includes additional tests covering the expected behavior. I'm still somewhat unsure about the chosen names for certain parameters, as they might benefit from being more general, but I'll defer that judgment to the final review before merging. In particular, I'd like to highlight the behavior change regarding the exclusion of default values from validation error messages, which I believe is the correct choice. Overall, I think this is excellent work. Thanks for your contribution!

luzzodev (approved):

sneakers-the-rat:
Happy to change any names, and also I will spend a bit of time this afternoon describing this behavior in the docs :)

sneakers-the-rat:
OK I added docs for this here: https://github.com/fastapi/fastapi/pull/13464/commits/f63e983b60cfd63a49ba03f53a7695e92e53436f

While i was doing that, i realize i couldn't really write satisfying docs because the feature is arguably incomplete - We still needed a way to detect if we're being validated as a form (or being used elsewhere in the program) which was the whole root of the problem, so I added [this](https://github.com/fastapi/fastapi/pull/13464/files#diff-943afc4e4de9d9d5403c162b885e560d6ca36029bcde0a920e916532c2a36c50R130) into the field validation wrapper:

```python
self._type_adapter.validate_python(
    value, from_attributes=True, context={"fastapi_field": self}
),
```

where we just pass the fastapi `ModelField` in the validation context. This should be a totally non-breaking change with near-zero perf consequences: all it does is allow a validator to opt-in to receiving the field as an additional field: https://docs.pydantic.dev/dev/concepts/validators/#validation-context

the reason that was necessary is hopefully obvious from the added docs, because otherwise it would be impossible to add special validation behavior *only when handling form data* without using `inspect` and getting the enclosing frames (awkward!). 

Previously I had proposed a special `@form_validator` decorator that would serve this purpose, but haven't heard back on that, can still implement that.

I recognize that is a late addition to the PR, but it is still directly related to the purpose of the pr (being able to handle form data when it differs from e.g. json or model use within python). I can remove it if we don't like that.

fastapi/dependencies/utils.py
```diff
@@ -724,14 +727,14 @@ def _get_multidict_value(
...
             and value == ""
         )
         or (is_sequence_field(field) and len(value) == 0)
     ):
-        if field.required:
-            return
+        if form_input or field.required:
+            return None
         else:
             return deepcopy(field.default)
```
YuriiMotov: How about changing this to:
```python
        if form_input:
            return None
        return None if field.required else deepcopy(field.default)
```
or
```python
        if form_input:
            return None
        if field.required:
            return None
        else:
            return deepcopy(field.default)
```

IMO, this way it will be clearer that we treat form fields in a different way.
Maybe we can also add the comment like:
```python
        if form_input:  # Handle empty string for Form models
            return None
```

fastapi/_compat.py
```diff
@@ -126,7 +126,9 @@ def validate(
         ) -> Tuple[Any, Union[List[Dict[str, Any]], None]]:
             try:
                 return (
-                    self._type_adapter.validate_python(value, from_attributes=True),
+                    self._type_adapter.validate_python(
+                        value, from_attributes=True, context={"fastapi_field": self}
+                    ),
```
YuriiMotov: This is not related to the main purpose of this PR, I suggest to remove this from this PR.

Also, I think this is not necessary to add the context here at all.
To understand whether the model is used for Form data or JSON data, you can create new model, inherit it from existing model and add the validator to this new model only. Then use this new model for Form data.

See code example in details:

<details>

```python
from typing import Annotated

from pydantic import BaseModel, model_validator

class ExampleModel(BaseModel):
    field_1: bool = True

class ExampleFormModel(ExampleModel):
    @model_validator(mode="before")
    def validate_inputs(cls, values: dict) -> dict:
        if values.get("field_1") is None:
            values["field_1"] = False
        return values

app = FastAPI()

@app.post("/form")
async def form_endpoint(model: Annotated[ExampleFormModel, Form()]):
    return {
        "fields_set": list(model.model_fields_set),
        "field_1": model.field_1,
    }

client = TestClient(app)

def test_form_model_fields_empty_is_false():
    resp = client.post("/form", data={})
    assert resp.status_code == 200, resp.text
    resp_json = resp.json()
    assert resp_json["field_1"] is False

def test_form_model_fields_explicit_true():
    resp = client.post("/form", data={"field_1": True})
    assert resp.status_code == 200, resp.text
    resp_json = resp.json()
    assert resp_json["field_1"] is True

def test_form_model_fields_explicit_false():
    resp = client.post("/form", data={"field_1": False})
    assert resp.status_code == 200, resp.text
    resp_json = resp.json()
    assert resp_json["field_1"] is False
```

</details>

fastapi/dependencies/utils.py
YuriiMotov: Not related to the main purpose of this PR. I think this should be moved to a separate PR

fastapi/dependencies/utils.py
YuriiMotov: Not related to the main purpose of this PR. I think this should be moved to a separate PR

YuriiMotov (commented):
@sneakers-the-rat, @luzzodev, thanks for working on this!

I think to get this merged we need to split these changes into several PRs.
First, we need to fix the issue with the default values being passed as input values to Form model (that causes the inability to check which fields were set).

I would focus on this main issue in this PR and when it's merged then we can open new PRs to add docs about quirks with form-encoded data.

I haven't properly reviewed tests yet. As I see it now, we basically need to make the following pass (in details):

<details>

```python
from typing import Annotated

class ExampleModel(BaseModel):
    field_1: bool = True

app = FastAPI()

@app.post("/form")
async def form_endpoint(model: Annotated[ExampleModel, Form()]):
    return {
        "fields_set": list(model.model_fields_set),
        "field_1": model.field_1,
    }

client = TestClient(app)

def test_form_model_fields_set_empty():
    resp = client.post("/form", data={})
    assert resp.status_code == 200, resp.text
    resp_json = resp.json()
    assert resp_json["fields_set"] == []
    assert resp_json["field_1"] is True
```

</details>

fastapi/_compat.py
sneakers-the-rat: sure, but the whole underlying purpose is that we want to be able to use models in a way that is agnostic to the input type (form, json, etc.). We *could* just make different models for every endpoint, and that would avoid the entire problem, but what we want is to be able to reuse models within the program as well as for form input so that we can avoid making a million separate models. the change here is extraordinarily minimal, is general-purpose (there are probably other circumstances when we would want to access the fastapi field within a validator), and fully backwards compatible.

I can remove this from this PR, but it means that we can't really document the changes here in a way that makes sense - "you want to be able to handle quirks of form data... there's actually no satisfying way to do that, but here is an awkward hack that involves duplicating your model classes..."

fastapi/dependencies/utils.py
sneakers-the-rat: sure, second version is equivalently readable to me as current impl, but if that is preferable will make that change now

sneakers-the-rat: done in https://github.com/sneakers-the-rat/fastapi/commit/fbb0480732756d794fc7de0138098bac71d386bd

fastapi/dependencies/utils.py
sneakers-the-rat: It actually is essential to the main purpose of this PR, see: https://github.com/fastapi/fastapi/pull/13464#issuecomment-2708617396

the previous behavior made sense when we did prefill default values - so then checking `if x in values` would be equivalent to the behavior in this PR because defaults would be present in `values`. 

since we are no longer prefilling default values, now the purpose is to preserve values that are not specified in the model but should still be validated (e.g. if extra values are allowed).

fastapi/dependencies/utils.py
sneakers-the-rat: see: https://github.com/fastapi/fastapi/pull/13464/files#r2208912386

fastapi/_compat.py
YuriiMotov: > the change here is extraordinarily minimal, is general-purpose (there are probably other circumstances when we would want to access the fastapi field within a validator), and fully backwards compatible.

This is not necessary for the purpose of this PR, right? You can open new PR for this.

My idea is: the less unnecessary changes this PR contains, the more chances it will be approved by Sebastián and merged

fastapi/dependencies/utils.py
YuriiMotov: Oh, now I see.. Thanks for pointing out!
I think we need to add some reasoning about this to the description of PR to make it clear why we need this

YuriiMotov:
If we want to get this reviewed by Sebastián and merged, our task is to prepare this PR so that the reviewing process would be as simple as possible for him.

I think for this we need to:
* Exclude unnecessary changes (not related to the main purpose)
* Revise tests (remove unnecessary, simplify, add new if needed)
* Update description
* Resolve and hide comments in conversation (maybe even close this PR and open new to have clean comment history)

fastapi/dependencies/utils.py
sneakers-the-rat: sure. added to main PR body.

fastapi/_compat.py
sneakers-the-rat: I understand the amount of labor it takes to maintain a widely used package, so my full sympathies there, and i don't mean to be harsh, but i would be much more willing to split things off into a separate PR if PRs moved with any velocity in this package.

If i copy just the text parts of all my comments from the original discussion that were required to justify raising an issue, then comments in the issue justifying the change, and then comments here, i'm at >3,600 words over five months. That's all for a PR that changes seven (7!!!!!) lines of code in the package to fix something that is inarguably a bug in the mismatch between the handling of form encoded data and model annotations. If i have to go through all that again to change literally a single line, the odds i actually go through with that are almost 0 - i've already adapted my code to just never use form encoded data because this took so long, so it no longer affects me, i'm just trying to improve the state of the package for everyone since i was already started. 

Again this change *is* necessary for the purpose of this PR: *ideally* fastAPI would just handle form data reasonably and use the information it already has from the type annotations to handle inconsistency between form encoded data and python types (where, e.g. booleans are encoded as false -> not present, true -> "on"), but since i was waved down from making it actually work well, the *fallback* solution that touched the least amount of code was to be able to detect when a model is being validated as a form vs when it's being validated from json or used in the python code. Without this change, it would *still* be impossible to handle the inconsistencies between form encoding and json (the root of the problem!) because we would be unable to tell when a model was being used with form input vs. other contexts. There would be effectively nothing to document, and no bug would be fixed, because the docs would just go "handling defaults in forms is weird, but there is no solution, sorry! just make more models despite using your models everywhere from the API through the database being one of the main features of fastAPI!" I could have made this more relevant to the PR by just passing `{"fastapi_form": True}` in the context, but that would be far, far less broadly useful, so my approach here was trying to both address the immediate needs of the PR while providing additional useful functionality, which is the source of it appearing unrelated to the PR. 

Along the way I proposed several other, imo better resolutions to that problem (e.g. a specific `@form_validator` decorator, "just handling it correctly"), but since the whole process is designed to whittle down contribs to as few lines as possible, those better options got pruned and this is what we have left. I've received extensive input from @luzzodev (thanks again!) and now from you, but it seems like in both cases none of us are really sure what would be permitted, and y'all don't seem to have discretionary power to make calls about what should go and what shouldn't. so it seems like the result of a process that seems to be intended to spare Sebastián from having to waste time triaging a bunch of junk issues and PRs (again! all sympathy there, totally understood how much work that must be) seems to actually be producing a very low volume of half-measure PRs because decisionmaking power isn't actually being delegated to y'all who are longterm maintainers, and people who want to contribute fall out of the process or accept half-measures at each step of the funnel from the forced discussion through PR process. The velocity of non-docs, non-bot PRs in a project that is this widely used i think speaks to that bottleneck.

So anyway. this has already been a very long journey for such a tiny *tiny* bugfix. I could move this to another PR if it wouldn't take another 5 months to go through the whole process, but pre-emptively removing changes that appear to be off-topic because the single person with decisionmaking power will only have 5 milliseconds to review the PR seems like a pretty maladaptive dev process to me, and i would probably just not bother and abandon the PR.

YuriiMotov: I totally understand your frustration about long time waiting for this to be reviewed.
On the other side, I know that sometimes even tiny changes may introduce more bugs than fix.
Also, it took me more than 1h to go through all comments (issue, discussion and this PR) to make sure I understand the problem and don't miss anything. Off course it will take much less time for Sebastián, but it's still not "5 milliseconds" and it's something that can be improved from our side.
Let's stop creating even more comments. :)
I'll find time to open an alternative version of this PR (anybody else who is willing to contribute, feel free to take this over) and let Sebastián decide which one to review and merge.

Thanks again for your contribution!

sneakers-the-rat: Yeah, right, the process needing to loop through the long route it did makes it hard to review in addition to hard to contribute, and needing to have a bunch of people retrace that context instead of the maintainer (luzzo) who followed it being able to make a call multiplies that. splitting off to competing approaches is demotivating in a similar way, where if i am trying to contribute and push something through a long process and then at the end my contribs just get dropped in favor of an internal contributor duplicating my work, i wonder why bother. in both cases it's the indeterminacy of "here is a diagnosed problem, but we don't really know what kind of fix could get merged" that is causing lots of extra work. 

like i said, if what you're saying is "in order for this PR to be merged, split it into two, and then we could review independently and merge them together as different parts of addressing the same problem" then i would happily do it right now. but if the request is "split this PR into two to double the process, that *might* make this incomplete PR mergeable, and then 5 months later maybe the thing that finally resolves the issue can be merged" then that's when I write comments like the above rather than posting 🫡 and opening the pr.

YuriiMotov: > in favor of an internal contributor duplicating my work

I think this part is not true. I'm not going to duplicate your work. I'm going to take a part of your work (I will off course point to your PR as a source of this changes and explain why I created new PR), review it carefully, revise tests and probably add others, add concise description.
I think it's normal for opensource. This is how great things get created.
If somebody else is willing to do this, It will be even better for me!

Yes, unfortunately I can't predict the decisions of Sebastián yet. But, to make this PR reviewed by Sebastián, someone who is responsible for sorting PRs should take the responsibility to approve it and send to Sebastián for review. In current state I'm not ready to take this responsibility, sorry..

sneakers-the-rat: alright in that case, all yours, i'm just going to fork.

svlandeg:
@sneakers-the-rat & @luzzodev: thanks so much for your time and effort put into this. I also understand your frustration with the PR reviewing process. As you mentioned, it really is a huge task going through all the PRs in detail. Considering the amount of users of the library, we are very wary of introducing new bugs with new features, or even with bug fixes, and so we really do need to be very careful whenever we merge any PR. That said, I agree that the process should be more fluent especially when it comes to bug fixes. I also want to say I really appreciate how much work you've both put into creating solid tests etc, this is really crucial for the code quality.

With respect to Sebastián reviewing the "final" PR (whenever e.g. Yurii or I sign off of it), I can assure you he would take a lot more than 5 milliseconds to go over it. In fact, he is very meticulous in his reviewing process and goes through every detail, thinks of every edge case and often finds issues even when I for instance thought it was 100% good to merge. He knows this library inside-out like no one else, and that's why his final review is so important. But at the same time we try to make his job as efficient as possible to have each PR address only one atomic fix/feature. Hence Yurii's requests.

Anyway let's try and keep the momentum for this PR, as it would be unfortunate to lose this work. Ideally, we should try and work off the original PR to ensure the original contributor gets recognition as well. @YuriiMotov: do you have time to work on this?

sneakers-the-rat:
appreciate everyone has limited time, and i am not trying to be a burden or ornery about having things my way, but i still maintain that the change that's being objected to is directly related to, and necessary to resolve the underlying issue for this PR.

The change is [this single line](https://github.com/fastapi/fastapi/pull/13464/files/fec0a068edc153b46f9ce26bad9fd2b36449e345#diff-943afc4e4de9d9d5403c162b885e560d6ca36029bcde0a920e916532c2a36c50R130) that gives the outer type to the validation context as `fastapi_field`. the validation context is an opt-in object that a validator only gets if it requests it, this is exactly what it's for, it provides additional useful information for validators that could be used outside this case, it doesn't conflict with any current fastapi behavior, and it's effectively impossible that there is another key `fastapi_field` in that dictionary that would make this conflict with downstream use.

The reason this is core to the purpose of this PR is pretty simple: 
- a) situation: form data needs to be validated differently than JSON data in some cases, e.g. booleans from checkboxes
- b) problem: we don't have enough information to apply validation logic only to form data because 
  - b1) we can't access the original values to do the form-specific validation, and
  - b2) we can't tell when we're receiving form data anyway
- c) solution: we provide that information by 
  - c1) not prefilling default values and allowing the model to fill its defaults in the normal validation flow
  - c2) allowing a validator to determine when it's being called with form data.
  
if we don't have (c2) then we don't solve the problem - we might be able to tell when a value with a default is missing, but that doesn't really matter because there's no way to distinguish form data from other contexts like in-program validation or validation from JSON! 

Again, I had proposed several other means of accomplishing this like a `@form_validator` decorator, or using the information we already have from the `Form()` annotation on a `bool` field to just automatically handle the differences between json and form data, so this one line change is the most minimal change i could find that accomplished (c2). I would prefer the other means, but alas minimizing PR size precluded better fixes.

> Revise tests (remove unnecessary, simplify, add new if needed)

I wrote tests that covered exactly the changes in the PR, and through review was asked to add several more. The docstrings in the tests describe what they are testing, and hopefully the parameterization is obvious enough. lmk which tests are unnecessary or how you would simplify them, but i am not seeing anything unnecessary there reading again.

> Update description

i did update the description as requested. also let me know what needs to be updated there because to me it just describes the changes and provides links to context. 

> Resolve and hide comments in conversation

sure if closing and reopening the same PR to clear its history is a thing that's done around here, i can do that, no problem.

YuriiMotov:
>   * c2) allowing a validator to determine when it's being called with form data.
> 
> if we don't have (c2) then we don't solve the problem - we might be able to tell when a value with a default is missing, but that doesn't really matter <ins>because there's no way to distinguish form data from other contexts like in-program validation or validation from JSON</ins>!

I already suggested a way to do this without adding context: https://github.com/fastapi/fastapi/pull/13464#discussion_r2208569314

I know you didn't like it, but _"there's no way to distinguish ..."_ is not true

sneakers-the-rat:
One could also directly parse the binary from the `Request`, but I personally find adding a free, zero-maintenance feature that allows us to consistently use FastAPI's own semantics of `Annotated[Model, Form()]` without needing to litter my code with duplicate models as a workaround is sort of a win-win, but yes i suppose that is true.

if this is actually the way all the maintainers want things to work, i'm sort of worn down after 6 months, so I removed the one controversial line

nidhishgajjar:

## Review of PR #13464: Don't prefill default values on form input

This is a well-targeted bug fix that addresses a real pain point with form models — the inability to distinguish between "user didn't send this field" and "user sent the default value." The HTML checkbox case is the classic example, and this fix handles it correctly.

### Core change analysis ✅

The fix in `_get_multidict_value` is clean and minimal:
```python
if form_input or field.required:
    return None
```
By returning `None` for unset form fields, the value is omitted from the dict (filtered by `if value is not None`), letting Pydantic's own default mechanism take over. This properly preserves `model_fields_set` semantics.

### What looks good ✅

- **Correct semantics**: Unset form fields are now truly absent from the input dict, so Pydantic validators can differentiate unset vs. default values
- **Extra keys fix**: Changed from `if key not in values` to `if key not in field_aliases` for preserving extra form keys — this correctly avoids re-adding processed keys that were filtered as `None`
- **Test coverage**: Excellent — covers all-unset, partially-set, empty-string, aliased fields, inlined params, Pydantic v1/v2, and comparison with JSON behavior
- **Documentation**: Clear explanation of the checkbox problem and the recommended model-duplication workaround
- **No effect on JSON endpoints**: The change only applies to form inputs via `form_input=True`

### Minor observations 💡

1. **Validation error `input` change**: The `test_forms_single_model.py` change shows validation errors will now report `input: {}` instead of the pre-filled defaults. This is more correct (shows what was actually received) but is a visible behavior change in error responses.

2. **Inlined form params unaffected**: Parameters using `Form()` directly on endpoints (not via a model) still get defaults filled in, since they don't go through `_extract_form_body`. The tests verify this correctly. This difference in behavior between model-based and inlined form params is worth documenting.

3. **The `test_casted_empty_defaults` test**: Good catch of the alias + empty string edge case. The test comment `# Expected :'bar' -> Actual :''` suggests this was a known issue that this PR fixes.

### Summary

Clean, focused bug fix with thorough test coverage and helpful documentation. The change is minimal (3 lines of core logic) but solves a significant usability issue with form models. The behavior is now consistent with how JSON body parsing works.

**Assessment: approve**

sneakers-the-rat:
Thanks orb reviewer, i think it's pretty good too, or at least it was a year ago when it was opened. a few of the ideas from here have already made their way into the code, but the problem remains unsolved, so I just updated the PR to match the new style for the docs, removed pydantic v1 support, etc. the day may yet come where the hundreds of comments and thousands of words i have written pay off and we merge 3 lines of code to fix a bug that makes form models with defaults possible to use. still crossing my fingers.

PR #13466, opened 2025-03-07
✨ Allow any type for `APIRoute.description`

miloserdev:


alv2017:
Can you explain why is this needed? and what are use cases?

miloserdev:
> Can you explain why is this needed? and what are use cases?
We just need easy data parse in our project
shot_Mar 8-15-43-48
shot_Mar 8-15-45-01

peidrao (approved):

Kludex:
Thanks for the PR, but the description is always a string on the OpenAPI specification.
You can use `str()` on the object before passing to it.

See: https://swagger.io/specification/.

PR #13469, opened 2025-03-08
♻️ Refactor tests in `tests/test_tutorial/test_cors/test_tutorial001.py`

alv2017:
The goal of this PR is to improve the existing test structure, and test more rigorously by adding extra test cases.

What has been done?

1) Test class has been created, and multiple test cases where added as separate tests into the class
2) Added a test case for `preflight request` when origin is not listed as allowed origin.
3) Added a test case for `simple request` when origin is not listed as allowed origin.
4) A wider list of CORS headers have been added to the tests.


nadzhou (changes requested):
Sorry, I don't really get these scenarios, can you clarify these tests a bit more?

alv2017:
@nadzhou 

- `test_preflight_with_allowed_origin`: imitates a browser sending to the API a preflight request for the allowed origin
- `test_preflight_with_not_allowed_origin`: imitates a browser sending to the API a preflight request for the non-allowed origin
- `test_simple_response_with_allowed_origin`: imitates a browser sending to the API a simple request for the allowed origin
- `test_simple_response_with_not_allowed_origin`: imitates a browser sending to the API a simple request for the non-allowed origin
- `test_non_cors_response`: sends a regular request to the API

So, what changes are you asking for? :smile:

tests/test_tutorial/test_cors/test_tutorial001.py
```diff
@@ -1,37 +1,103 @@
...
+    @pytest.mark.parametrize("allowed_origin_url", origins)
+    def test_preflight_with_allowed_origin(self, client, allowed_origin_url):
+        origin_url = allowed_origin_url
+        headers = {
+            "Origin": origin_url,
+            "Access-Control-Request-Method": "GET",
+            "Access-Control-Request-Headers": "X-Example",
+        }
+        response = client.options("/", headers=headers)
+        assert origin_url in self.allowed_origins
```
YuriiMotov, line 24: Don't see a point in this assertion. Do we need this?

tests/test_tutorial/test_cors/test_tutorial001.py
```diff
@@ -1,37 +1,103 @@
...
-        "Access-Control-Request-Headers": "X-Example",
-    }
-    response = client.options("/", headers=headers)
-    assert response.status_code == 200, response.text
-    assert response.text == "OK"
-    assert (
-        response.headers["access-control-allow-origin"]
-        == "https://localhost.tiangolo.com"
-    )
-    assert response.headers["access-control-allow-headers"] == "X-Example"
```
YuriiMotov, line 21: Could you please explain why you removed this line?

tests/test_tutorial/test_cors/test_tutorial001.py
```diff
@@ -1,37 +1,103 @@
...
+        # response
+        assert response.status_code == 200
+        # response headers: cors
+        assert "access-control-allow-methods" in response.headers
+        assert "access-control-allow-credentials" in response.headers
+        assert "access-control-max-age" in response.headers
+        assert "access-control-allow-headers" in response.headers
+        # response headers: cors: origin
+        assert "access-control-allow-origin" in response.headers
+        assert response.headers["access-control-allow-origin"] == origin_url
```
YuriiMotov, line 39:
```suggestion
        assert response.headers["access-control-allow-origin"] == origin_url
```
I understand your point, but I think it's better to simplify such assertions

tests/test_tutorial/test_cors/test_tutorial001.py
```diff
@@ -1,37 +1,103 @@
...
+        }
+        response = client.options("/", headers=headers)
+        assert origin_url in self.allowed_origins
+        # response
+        assert response.status_code == 200
+        # response headers: cors
+        assert "access-control-allow-methods" in response.headers
+        assert "access-control-allow-credentials" in response.headers
+        assert "access-control-max-age" in response.headers
+        assert "access-control-allow-headers" in response.headers
```
YuriiMotov, line 32:
```suggestion
        for header_key in MANDATORY_HEADERS:
            assert header_key in response.headers
```
I think this will reduce code duplication and even make code more readable

YuriiMotov (commented):
@alv2017, thanks for your efforts!

I think we should convert tests back to just functions (don't group them into class). At least because it makes it harder to review changes. Also, IMO, there is no need for grouping them in class here.

Also, I think we should try to make tests a bit simpler as I suggested in the comments

njsbhu:
I reviewed test_tutorial001.py, there is pytest code that checks FastAPI’s CORS behavior, but it repeats with the same steps, we can use fixtures to organize the tests.

```
@pytest.fixture(scope="class")
def client():
    return TestClient(app)
```

PR #13475, opened 2025-03-10
✨ Support passing context to Pydantic serializers via `jsonable_encoder`

MaximeGoyette:
It would be handy if we could pass some context to the Pydantic serializers through the `jsonable_encoder` function.

> You can pass a context object to the serialization methods which can be accessed from the `info` argument to decorated serializer functions. This is useful when you need to dynamically update the serialization behavior during runtime. For example, if you wanted a field to be dumped depending on a dynamically controllable set of allowed values, this could be done by passing the allowed values by context:

-- Pydantic docs on [Serialization Context](https://docs.pydantic.dev/latest/concepts/serialization/#serialization-context)

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -126,6 +126,15 @@ def jsonable_encoder(
             """
         ),
     ] = None,
+    context: Annotated[
+        Optional[Any],
+        Doc(
+            """
+            Pydantic's `context` parameter, passed to Pydantic serializers.
+            This only works with Pydantic V2.
+            """
+        ),
+    ] = None,
     by_alias: Annotated[
         bool,
         Doc(
@@ -225,6 +234,7 @@ def jsonable_encoder(
             mode="json",
             include=include,
             exclude=exclude,
+            context=context,
             by_alias=by_alias,
             exclude_unset=exclude_unset,
             exclude_none=exclude_none,
@@ -234,6 +244,7 @@ def jsonable_encoder(
             obj_dict = obj_dict["__root__"]
         return jsonable_encoder(
             obj_dict,
+            context=context,
             exclude_none=exclude_none,
             exclude_defaults=exclude_defaults,
             # TODO: remove when deprecating Pydantic v1
@@ -247,6 +258,7 @@ def jsonable_encoder(
             obj_dict,
             include=include,
             exclude=exclude,
+            context=context,
             by_alias=by_alias,
             exclude_unset=exclude_unset,
             exclude_defaults=exclude_defaults,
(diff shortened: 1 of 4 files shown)


svlandeg (commented):
Hi, thanks for your contribution! It looks like the test suite is failing for Pydantic v1, could you look into that? I'll put the PR in draft in the meantime - feel free to mark as "Ready for review" once the test suite is green!

tests/test_jsonable_encoder.py
```diff
@@ -316,3 +320,25 @@ class Model(BaseModel):
 def test_encode_pydantic_undefined():
     data = {"value": Undefined}
     assert jsonable_encoder(data) == {"value": None}
+
+
+@needs_pydanticv2
```
svlandeg, line 325: Thanks for looking into the issue with Pydantic v1!

I think we should still include a test for Pydantic v1 that ensures that the `kwargs.pop` functionality works correctly, and that things won't break unexpectedly when using Pydantic v1.

MaximeGoyette: Good point! Comment adressed.

tests/test_jsonable_encoder.py
```diff
@@ -316,3 +320,25 @@ class Model(BaseModel):
...
+    from pydantic import SerializationInfo, field_serializer
+
+    class ModelWithContextualSerializer(BaseModel):
+        value: int
+
+        @field_serializer("value")
+        def serialize_value(value: int, info: SerializationInfo) -> int:
+            if info.context is not None and isinstance(
+                value_from_context := info.context.get("value"), int
+            ):
```
YuriiMotov: May be better split into two nested ifs to improve readability?

fastapi/encoders.py
```diff
@@ -126,6 +126,14 @@ def jsonable_encoder(
             """
         ),
     ] = None,
+    context: Annotated[
+        Optional[Any],
+        Doc(
+            """
+            Pydantic's `context` parameter, passed to Pydantic serializers.
```
YuriiMotov, line 133: Should we add a note that this only works with Pydantic V2?

YuriiMotov (approved):
LGTM in general, added a couple of optional suggestions.

This PR seems to be **not very useful** by its own, but it will be a base for providing serialization context to serialize the results with response model ([this PR](https://github.com/fastapi/fastapi/pull/11634) attempts to implement it)

tests/test_jsonable_encoder.py
MaximeGoyette: Done!

fastapi/encoders.py
MaximeGoyette: Done!

tiangolo:

I want to implement a few big refactors and features that will probably interact with this logic and might cover the intention here.

Thank you! 🍰

PR #13481, opened 2025-03-12
✨ Relax HTTPBearer call requirement to support websockets

florian-sattler:
Currently, using `HTTPBearer` is not possible for websockets as it requires an HTTP request.

However, its functionality only ever uses the HTTP headers, which are also present in websockets, as both HTTP requests and websocket requests are based on the `HTTPConnection` class.

This PR relaxes the call requirement of `HTTPBearer` to only require a `HTTPConnection`, which allows using this feature in a type-safe manner for websockets.

### DIFF
--- a/fastapi/security/http.py
+++ b/fastapi/security/http.py
@@ -8,7 +8,7 @@ from fastapi.openapi.models import HTTPBearer as HTTPBearerModel
 from fastapi.security.base import SecurityBase
 from fastapi.security.utils import get_authorization_scheme_param
 from pydantic import BaseModel
-from starlette.requests import Request
+from starlette.requests import Request, HTTPConnection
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@florian-sattler, thanks for your efforts!

There is already #10147 that attempts to solve the same problem and it's more elaborate.

njsbhu:
In code http.py, I would suggest to use `HTTPConnection `everywhere for type consistency. e.g -
Some` __call__` methods use `Request`, other than `HTTPConnection ` and for security dependencies, `HTTPConnection  `is usually sufficient.
We can align status code like - For all the HTTP auth schemas, we can use 401 for "invalid Credential" and 403 for "authenticated but not allowed"

PR #13503, opened 2025-03-20
📝 Add missing type hints for `docs_src/background_tasks`

valentinDruzhinin:


YuriiMotov (commented):
@valentinDruzhinin, thanks for your efforts, but I think we don't need to do this (at least now).

First, this adds a bit of unnecessary complexity for readers.

Second, if we really decide to add return type annotations we should do it consistently across entire codebase. That's a lot of work and now it's probably not the right thing to spend our time on.

valentinDruzhinin:
@YuriiMotov got it, thanks for the feedback :) I'll close the PR

PR #13506, opened 2025-03-20
Simplify tests for schema_extra_example #13150

valentinDruzhinin:
#13150


valentinDruzhinin:
closing, payload for pydanticv1 is different. Better to leave it as it is.

njsbhu:
In test_tutorial001.py, I see the OpenAPI check is very strong and makes sure we don’t accidentally break things as per my review.
But right now it looks at the whole schema, which can be a bit fragile. If some unrelated field changes (like metadata or optional fields), the test might fail even though the important parts are fine.
A small improvement would be to assert only the key parts we care about (like certain paths or components). That way, the test still catches real problems but won’t break just because of minor, unrelated changes.

PR #13507, opened 2025-03-20
✅ Simplify tests for `validate_response_recursive`

valentinDruzhinin:
#13150


valentinDruzhinin:
@YuriiMotov can you please take a look?

YuriiMotov (approved):
LGTM

tiangolo (approved):
Nice, clever, thank you! 🚀 🍰

njsbhu:
In app.py, I see uses a mutable default argument ([]), which I think is unsafe in Python - the below one
```
class RecursiveItem(BaseModel):
    sub_items: List["RecursiveItem"] = []
    name: str
```

Instead I would suggest to use like below 
```
class RecursiveItem(BaseModel):
    name: str
    sub_items: List["RecursiveItem"] = Field(default_factory=list)
```
    
and same applies to RecursiveSubitemInSubmodel & RecursiveItemViaSubmodel

PR #13509, opened 2025-03-20
✨ Add HEAD method on GET routes

### LINKED ISSUE #1773: Automatically support HEAD method for all GET routes, as Starlette does

Kludex:
- Closes https://github.com/fastapi/fastapi/issues/1773

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -492,7 +492,7 @@ class APIRoute(routing.Route):
         if methods is None:
             methods = ["GET", "HEAD"]
         elif "GET" in methods:
-            methods = set(methods) | {"HEAD"}
+            methods = set(methods).union({"HEAD"})
         self.methods: Set[str] = {method.upper() for method in methods}
         if isinstance(generate_unique_id_function, DefaultPlaceholder):
             current_generate_unique_id: Callable[[APIRoute], str] = (
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -490,7 +490,9 @@ class APIRoute(routing.Route):
         self.name = get_name(endpoint) if name is None else name
         self.path_regex, self.path_format, self.param_convertors = compile_path(path)
         if methods is None:
-            methods = ["GET"]
+            methods = ["GET", "HEAD"]
+        elif "GET" in methods:
+            methods = set(methods) | {"HEAD"}
         self.methods: Set[str] = {method.upper() for method in methods}
         if isinstance(generate_unique_id_function, DefaultPlaceholder):
             current_generate_unique_id: Callable[[APIRoute], str] = (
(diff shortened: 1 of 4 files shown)


fastapi/routing.py
```diff
@@ -490,7 +490,9 @@ def __init__(
         self.name = get_name(endpoint) if name is None else name
         self.path_regex, self.path_format, self.param_convertors = compile_path(path)
         if methods is None:
-            methods = ["GET"]
+            methods = ["GET", "HEAD"]
+        elif "GET" in methods:
+            methods = set(methods).union({"HEAD"})
```
alv2017, line 495: Mixing lists and sets doesn't look great ... I would prefer a consistent processing approach and stick to sets.

tests/test_tutorial/test_first_steps/test_tutorial001.py
```diff
@@ -18,6 +18,9 @@ def test_get_path(path, expected_status, expected_response):
     assert response.status_code == expected_status
     assert response.json() == expected_response
 
+    response = client.head(path)
+    assert response.status_code == expected_status
+
```
alv2017, line 23: We also need to test:

a) Response to HEAD request does not contain body data
b) The response headers to HEAD request are the same as the headers to GET request

alv2017 (commented):
1) I'm wondering what is going to happen if HEAD handlers are already defined manually. Probably this needs to be mentioned in the PR description.

2) Would it be possible to disable automatically generated HEAD method?

tests/test_tutorial/test_first_steps/test_tutorial001.py
Kludex: > Response to HEAD request does not contain body data

We don't. This is tested on the underneath packages.

alv2017: Why do we need this?

This is the tutorial tests, and in tutorial tests we want to demonstrate basic functionalities to package users. This is sort of a supplement to documentation, and from my perspective, this would be a nice to have.

ocodo:
If anyone needs a hand, I'm happy to help get this done.

njsbhu:
In utils.py, I see the code uses `assert ` statements to check assumptions. But assert is usually meant for debugging, not for permanent runtime checks and in some environments, assert can even be disabled, which means those checks won’t run.

PR #13510, opened 2025-03-20
📝 Add note about Cookies and JavaScript on `tutorial/cookie-params.md`

Kludex:


svlandeg (approved):
The same note is already present in `cookie-param-models.md`, but it makes sense to me to have it in `cookie-params.md` as well, as this page occurs earlier on in the tutorial.

tiangolo (approved):
Nice, thanks! 🍪

Issue #13471, opened 2025-03-09
Basic auth `realm` is REQUIRED but handled as optional


tvogel:
According to https://datatracker.ietf.org/doc/html/rfc7617#autoid-3 , specification of the `realm` parameter is REQUIRED, so making `realm` optional here

https://github.com/fastapi/fastapi/blob/643d2845de09a437b955a163ab25eb0887bfd0e1/fastapi/security/http.py#L195

and at related places, is misleading and makes it easy to create non-conforming APIs. It bit me when a Shelly device refused (rightfully) to authenticate against the API which was using the default (missing) `realm` behavior.

Whizzo123:
Yeah this does look like a mistake in the API to have "realm" as optional

Whizzo123:
Yeah this does look like a mistake in the API to have "realm" as optional looks to be getting fixed here https://github.com/fastapi/fastapi/pull/12507

tvogel:
Ah, good that this is already in the works! However, the fix to default it to "global" is questionable to me because that is a string that in general gets presented to users as a prompt to provide credentials, so shouldn't have a rather meaningless default. A good framework should make the framework user provide relevant detail like the realm. If they use basic auth, they need to define a realm.

CODEKRUNCH:
Has The Issue Been Resolved Uptill Now ??

YuriiMotov:
I think we should close this issue.

As Sebastian mentioned in [this comment](https://github.com/fastapi/fastapi/pull/13786#pullrequestreview-3501695067), `realm` seems to be not required in latest version of RFC.

> I also noted that the spec says that the realm is now optional, but only in a footer note (https://datatracker.ietf.org/doc/html/rfc7235#appendix-A). 😂 Anyway, we are good with that now.


I checked this and agree:

The https://datatracker.ietf.org/doc/html/rfc7617#autoid-3 states:

> The authentication parameter 'realm' is REQUIRED ([RFC7235], Section 2.2).

But if we go to Section 2.2 of RFC7235, we have:

> [2.2](https://datatracker.ietf.org/doc/html/rfc7235#section-2.2).  Protection Space (Realm)
> 
>    The "realm" authentication parameter is reserved for use by
>    authentication schemes that wish to indicate a scope of protection.
> 
>    A protection space is defined by the canonical root URI (the scheme
>    and authority components of the effective request URI; see [Section](https://datatracker.ietf.org/doc/html/rfc7230#section-5.5)
[...]

PR #13511, opened 2025-03-20
🐛 Add realm to HTTP Basic Authentication

Kludex:
- Related to https://github.com/fastapi/fastapi/issues/13471


nehaanndev (approved):

YuriiMotov:
I think we don't need to do it. See: https://github.com/fastapi/fastapi/issues/13471#issuecomment-3596388412

tiangolo:
Yep, I'll pass on this, thanks! :coffee:

PR #13514, opened 2025-03-20
🐛 Use 401 instead of 403 in security classes

Kludex:


svlandeg:
HI @Kludex, thanks for the PR! I'll go ahead and close this in favour of the more extensive PR https://github.com/fastapi/fastapi/pull/13786 which should hopefully address this issue once and for all. Reviews on #13786 welcome if you should feel like it! 🙏

njsbhu:
In http.py, I see the way of authentication works is clear and consistent
One small improvement could be to move the repeated code for authorization parsing and error handling into the base class to avoid duplication.

PR #13525, opened 2025-03-22
📝 Update middleware tutorial and examples

alv2017:
1. `tutorial/middleware.md` updates:
- added a remark, that the tutorial focuses on HTTP middleware.
- added new section: *Class-Based Middleware*. It explains how to create a custom `http` middleware using classes.

2. Source code update: `docs_src/middleware/tutorial001.py`:
- added endpoint to the test app (this will allow to write tests against the middleware)

3. New source code added: `docs_src/middleware/tutorial002.py`. It contains a test app for the class-based middleware.

4. Tests added:
- `tests/test_tutorial/test_middleware/test_tutorial001.py`
- `tests/test_tutorial/test_middleware/test_tutorial002.py`


YuriiMotov (commented):
The `BaseHTTPMiddleware` has some limitations (https://www.starlette.io/middleware/#limitations, also it's [much slower than pure ASGI middleware](https://github.com/fastapi/fastapi/discussions/6985)) and it's advised to use ASGI middlewares instead.

So, I'm not sure we need to focus on documenting how to use it. I would better document these limitations (function style middleware that is described in docs also uses `BaseHTTPMiddleware` under the hood).

tiangolo (commented):
Thanks for the interest!

But yep, I prefer to keep the docs as they are for now. I'll close this one, but thanks for the effort! ☕

PR #13528, opened 2025-03-23
👽️ Address deprecation of `BaseModel.dict()` in Pydantic v2

harsh-haria:
- Have replaced `.dict()` with `.model_dump()` as `.dict()` is deprecated in Pydantic V2.

- Changes have just been made to the scripts in the docs_src/body folder.

- Here is the official documentation for reference: https://docs.pydantic.dev/latest/migration/#migration-guide:~:text=model_copy()-,dict(),-model_dump()


svlandeg (approved):
Looks good to me, thanks!

harsh-haria:
Any update on this ?

tiangolo:
Thanks for the interest!

If you see the relevant page in the docs you will see the note:

Whenever we move to use the new way of declaring models we should do it in one sweep in all the docs source examples, and update the docs to let people know of the difference.

So, for now I'll pass on this one, but thanks for the interest! :coffee:

njsbhu:
In tutorial002.py - I would suggest to use modern typing like `str | None` instead of `Union[str, None]` like below

```
description: str | None = None
tax: float | None = None
```

PR #13529, opened 2025-03-23
fix: Includes Optional in parameter type and removes Union

victorflipe:
- Replaced Union by Optional


svlandeg (commented):
Hi @victorflipe, thanks for your interest in contributing to FastAPI!

As stated in the [docs](https://fastapi.tiangolo.com/python-types/#using-union-or-optional), we actually prefer using `Union[X, None]` over `Optional[X]`. So even though both are equivalent, we'll pass on this change for consistency reasons.

PR #12502, opened 2024-10-21
🐛 Fix support for nullable form fields

maxclaey:
Fix support for nullable form fields, differentiate between unspecified and specified as None. See related bug ticket https://github.com/fastapi/fastapi/issues/12245

PR #13537, opened 2025-03-25
🐛 Fix support for form values with empty strings interpreted as missing (`None` if that's the default), for compatibility with HTML forms

### LINKED ISSUE #13533: Multiple regressions in the handling of forms & form validation
### Privileged issue

- [x] I'm @tiangolo or he asked me directly to create an issue here.

### Issue Content

We have found at least 2 regression in the way forms are handled while trying to migrate to v0.115.11 from version v0.112.4. After investigation we were able to identify 2 commits that introduce 2 separate bugs that break the handling of default values in both `x-form-urlencoded` and `mutlipart` froms. Partial fixes have been proposed that also address a CVE, but the PR has been hanging since october 24.

# Breaking in the handling of default value in `x-url-encoded` forms

This bug has been introduced in https://github.com/fastapi/fastapi/pull/12134.

MRE:
```python
app = FastAPI()


@app.post("/")
def root(
    name: Annotated[Optional[str], Form(embed=True)] = None,
):
    print(name)
    return name
```

with the following request:
```
❯ http post -v --form localhost:8888 name=""
POST / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, br, zstd
Connection: keep-alive
Content-Length: 5
Content-Type: application/x-www-form-urlencoded
Host: localhost:8888
User-Agent: xh/0.24.0

name=

HTTP/1.1 200 OK
Content-Length: 2
Date: Sun, 23 Mar 2025 21:01:49 GMT
Server: uvicorn

""
```
The expected result is that `None` is printed, but instead `""` is printed.

this is the offending bit from the linked PR:
```
diff --git a/fastapi/dependencies/utils.py b/fastapi/dependencies/utils.py
index 6083b731..db7eedba 100644
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -789,9 +789,9 @@ async def _extract_form_body(
         if value is not None:
             values[field.name] = value
-    for key, value in received_body.items():
-            values[key] = value
+    # for key, value in received_body.items():
+    #     if key not in values:
+    #         values[key] = value
     return values
```
This patch solves the x-form-urlencoded  case. So we indeed have two different regressions.
This _get_multi_dict_value  gets called for all fields in the schema: https://github.com/fastapi/fastapi/blob/4633b1bca933e68dac5c3bcce797ff5963debe2a/fastapi/dependencies/utils.py#L765-L765
we see here the special case for empty fields: https://github.com/fastapi/fastapi/blob/4633b1bca933e68dac5c3bcce797ff5963debe2a/fastapi/dependencies/utils.py#L695-L700
This bit here just blindly re-adds all fields from the body:    https://github.com/fastapi/fastapi/blob/4633b1bca933e68dac5c3bcce797ff5963debe2a/fastapi/dependencies/utils.py#L794-L796

This would be partially fixed by https://github.com/fastapi/fastapi/pull/12502/files, but also requires that ignored fields are not re-added to the form.

# Breaking in the handling of default value in `multipart` forms

The breaking of multipart form originates from https://github.com/fastapi/fastapi/pull/12117 where the check for empty field was dropped https://github.com/fastapi/fastapi/pull/12117/files#diff-aef3dac481b68359f4edd6974fa3a047cfde595254a4567a560cebc9ccb0673fR669

MRE:
```python
app = FastAPI()


@app.post("/")
def root(
     file: Annotated[Optional[bytes], File()] = None,
    name: Annotated[Optional[str], Form(embed=True)] = None,
):
    print(name)
    print(file)
    return name
```

The same request as the previous bug yields name printing an empty string.

Both commits were identified by bisecting using the MREs in this issue

This also solves https://github.com/fastapi/fastapi/discussions/13255, but scanning through the issues of the repo it seems that there's a bunch of issue related to handling of default values in forms that I suspect may be linked to this issue.

MarinPostma:
There was a bug in the validation of form data for optional params with None default value. This is a regression that was introduced by https://github.com/fastapi/fastapi/pull/12134.

## Explaination:
The `_get_multidict_value` uses `None` to represent both the absence a value in the dict, and a default value of `None` for that field. When the values are collected in `_extract_form_body`, they are only added to the values if they are not `None`. Later (#12134), some code was added that attempted to complement the model values with values not explicitely in the model, if values for those fields were not already processed earlier. The issue is that since we didn't add values to the initial value dict if they were none, we attempted to add them again in that second iteration. This generally wasn't an issue, except for the fast that a form field set to the empty string (`""`) is interpreted as the absence of that field. This resulted in `_get_multidict_value` picking the default value, that would happen to be `None`, so it wasn't added to the values dict in the first iteration. In the second iteration, we would find that value in the request params again, but set to `""`, and add that to the initial value dict. Later, validation would choke on that empty string.

The fix is to keep track of already visited field, rather that relying on the content of the values dict to complement the missing field. Such thing was already done in `request_params_to_args` for example.

closes #13533


MarinPostma:
Hello @tiangolo :) Do you think one will get merged? We're trying to figure out a way forward to be able to bump fastapi somehow, but if this is not a fix you'd merge, we'd need to think of a different approach

Thanks!

quiqueg (approved):
I just ran this patch against my full test suite and I can confirm that it fixed all of my tests that had started failing when upgrading from `fastapi` version `0.111.0` to `0.115.12`.

Nice work, @MarinPostma!

MarinPostma:
@quiqueg well I hope we can eventually merge it :)

tests/test_form_default.py
```diff
@@ -0,0 +1,47 @@
...
+
+@app.post("/urlencoded")
+async def post_url_encoded(age: Annotated[Optional[int], Form()] = None):
+    return age
+
+
+def test_form_default_url_encoded():
+    response = client.post("/urlencoded", data={"age": ""})
+    assert response.status_code == 200
+    assert response.text == "null"
```
PidgeyBE, line 19: This looks odd to me.
If you pass a string to an integer field, I would assume a `4xx` error code.

schneebuzz: This is a form POST and empty numeric inputs would be submitted as empty strings. IMO it's reasonable that they are accepted and treated like *no value* (`None` / `null`) when the field is an optional integer.

schneebuzz (approved):

JordanChow51 (approved):

YuriiMotov (approved):
LGTM!

I removed the `post_multi_part_file_first` test as it was unrelated to the current issue and that issue (that the order of `File` and `Form` matters) has already been fixed: https://github.com/fastapi/fastapi/pull/11194

@MarinPostma, thanks you for your work and patience!

YuriiMotov:
There is also https://github.com/fastapi/fastapi/pull/12502 that solves close issue - inability to pass `None` if form field is nullable, but default value is not `None`:

<details><summary>Code example in the details</summary>

```py
from typing import Optional
from uuid import UUID, uuid4

from typing_extensions import Annotated

app = FastAPI()

default_uuid = uuid4()


@app.post("/form-optional")
def post_form_optional(
    test_id: Annotated[Optional[UUID], Form()] = default_uuid,
):
    return test_id


def test_pass_none() -> None:
    client = TestClient(app)
    response = client.post("/form-optional", data={"test_id": None})
    assert response.json() is None


def test_pass_empty_str() -> None:
    client = TestClient(app)
    response = client.post("/form-optional", data={"test_id": ""})
    assert response.json() is None

```

</details>

I think it's Ok if we address these cases separately, but it would be nice to investigate if we can easily address them both in one PR.

tiangolo (approved):
Sorry for the delay handling this, and thank you for the explanation and fix @MarinPostma! 🙌 

Thanks for the help @YuriiMotov! 🍰 

---

I made a local quick example with an HTML form to confirm that browsers by default would send the fields even if they are empty, and will send them with an empty string. That convinced me this fix is necessary.

---

This will be available in FastAPI 0.123.2, released in the next few hours. 🎉

PR #13543, opened 2025-03-26
♻️ Refactor tests in `tests/test_tutorial/test_background_tasks`

alv2017:
Currently when running tests from the `tests/test_tutorial/test_background_tasks` a `log.txt` file is created.

This PR proposes a cleaner solution, that allows to remove the `log.txt` after the execution of every single test (located in the `tests/test_tutorial/test_background_tasks` directory):

- Created a file `conftest.py` containing a pytest fixture, that performs a tear down task.
-  The fixture is re-used in both tests: `tests/test_tutorial/test_background_tasks/test_tutorial001.py`,   `tests/test_tutorial/test_background_tasks/test_tutorial002.py`,

### DIFF
--- a/tests/test_tutorial/test_background_tasks/conftest.py
+++ b/tests/test_tutorial/test_background_tasks/conftest.py
@@ -1,6 +1,7 @@
 import os
-import pytest
 from pathlib import Path
+
+import pytest
 
 
 @pytest.fixture(name="path_to_log_file")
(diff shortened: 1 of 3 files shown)


YuriiMotov (changes requested):
@alv2017, thanks!

`log.txt` is still created by `docs_src/events/tutorial002.py`, so we need to fix this one too

njsbhu:
In conftest.py - I would suggest Use `Path.unlink()` instead of `os.remove` (something like below)
```
# os.remove(log)
log.unlink()
```

PR #13571, opened 2025-04-01
📝 Fix liblab client generation doc link

EFord36:
In [this section of the docs](https://fastapi.tiangolo.com/advanced/generate-clients/#client-and-sdk-generators-sponsor), currently clicking the liblab link gives a 'Page Not Found' response:

Screenshot 2025-04-01 at 15 41 49

The link in the docs is https://developers.liblab.com/tutorials/sdk-for-fastapi/?utm_source=fastapi but it would seem it should be https://developers.liblab.com/tutorials/sdk-for-fastapi?utm_source=fastapi without the last `/` before the `?`. This modified link works as expected:

Screenshot 2025-04-01 at 15 43 07


svlandeg (approved):
Thanks @EFord36! I verified that the link was incorrect and is fixed with this PR 🎉

PR #13572, opened 2025-04-01
🐛  Fix/http basic realm

### LINKED ISSUE #13471: Basic auth `realm` is REQUIRED but handled as optional

shyam-ramani:
fixes #13471


svlandeg (commented):
Hi @shyam-ramani, thanks for your interest in contributing to FastAPI!

It looks like this specific issue is already covered by PR https://github.com/fastapi/fastapi/pull/13511. As such, I would suggest to close this one and continue the review process on the other one.

PR #13253, opened 2025-01-24
📝 Remove references to `#app.on_event` in the SQL database section

IdleSys:
# Overview

This pull request updates the FastAPI documentation to reflect the deprecation of the `@app.on_event` decorator in the SQL database section of tutorial. The documentation has been revised to guide users towards using the new lifespan event handling system for managing application startup and shutdown events.

# Changes Made

## Updated Documentation:
- Removed references to @app.on_event in python `doc_src` files.
- Code demonstration updated in documentation markdown files for English. 

# Testing

- Reviewed all modified documentation for accuracy and clarity.
- Ensured that all examples are functional and correctly illustrate the new usage patterns.

PR #13576, opened 2025-04-02
📝 Use lifespan events instead of @app.on_event annotation in sql example code

dbejo:
I started to learn about FastAPI and followed the docs when I found that on_event annotation is deprecated. This PR changes the example codes for SQL databases to use the suggested lifespan event handlers.

### DIFF
--- a/docs_src/sql_databases/tutorial001.py
+++ b/docs_src/sql_databases/tutorial001.py
@@ -1,3 +1,4 @@
+from contextlib import asynccontextmanager
 from typing import List, Union
 
 from fastapi import Depends, FastAPI, HTTPException, Query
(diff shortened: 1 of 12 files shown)


svlandeg (commented):
Hi @dbejo, thanks for the contribution! We appreciate it, and hope you've been having fun learning FastAPI 😉 

It looks like all the edits from this particular PR are also included in https://github.com/fastapi/fastapi/pull/13253, so we'll probably go ahead and close this one as it's considered to be a duplicate. But thanks again! 🙏

tiangolo:

I want to implement a big refactor and some features in how to handle these things, I'm not fully happy with the existing interface.

I'll do that first, and then update the docs to reflect the recommended way to do things. I don't want to update the docs yet and suggest doing something that later I will suggest doing in a different way.

So for now, I will close this one. Thank you! 🍰

PR #13583, opened 2025-04-03
⚡️ Speed up test suite via caching and fixture scopes to make it ~24% faster

dikos1337:
While running the tests, I noticed a couple of spots where we could speed things up a bit based on profiling.

1.  **Caching `verify_password`:**
    *   The `verify_password` helper (using bcrypt) can be a bit slow, and I saw it was getting called with the same inputs multiple times during the test run.
    *   I added `@lru_cache()` to it. Since the results are always the same for the same inputs, this is safe.
    *   In my run, this cut its calls from ~70 down to ~40, and its share of the total test time dropped from **~26.9%** to **~16.2%**.

2.  **Optimizing Fixture Scopes:**
    *   I also saw two fixtures that were being set up and torn down *a lot* (like 166 and 86 times!), taking up around **8.7%** and **4.9%** of the total test time respectively.
    *   Since they didn't seem to need a completely fresh state for *every single test function*, I changed their scope to `scope="module"`.
    *   Now they run way less often, and their time usage basically disappeared from the profile.

**Overall Speedup:**

The profiling breakdown suggested these specific parts contributed to a relative reduction of ~24.3% of the original time.

**Actual measured impact:** Running the full test suite (2334 passed, 130 skipped) on my machine took **24.37s** before these changes and now takes **16.85s**.

That's a time saving of 7.52s, resulting in an overall test suite speedup of **~30.9%** (`(24.37 - 16.85) / 24.37 * 100%`).

Hopefully, this makes the test suite noticeably faster for everyone! Let me know what you think. 🙂

image

image

image


alv2017:
> 1. Caching `verify_password`:

Introduces a security vulnerability into the application.

svlandeg (commented):
Thanks for the PR @dikos1337! Definitely appreciate you trying to make the test suite more efficient 🙏 

With respect to @alv2017's comment, about the security of using `@lru_cache` for `verify_password`. I'm also unsure about this change: even if we would consider the usage safe in the context of running our test suite, we also have to consider that these code snippets are used verbatim in the tutorial (e.g. [oauth2-scopes](https://74ae3147.fastapitiangolo.pages.dev/advanced/security/oauth2-scopes)), and as such we need to consider whether this is the "best practices" approach we want to show users. 

Another important point: by adding lines to the `tutorialXXX.py` files, the highlighting of specific lines in the tutorial will be wrong. For instance, on the `oauth2-scopes` link above you can see that e.g. the lines around the `get_current_user` function are highlighted wrongly (compare with the [current](https://fastapi.tiangolo.com/advanced/security/oauth2-scopes) version)).

dikos1337:
@svlandeg Maybe I should just leave optimizing fixture scopes in this pull request for now? Shouldn't that be a problem?

alv2017:
@dikos1337, which tool were you using to measure the performance of the tests?

dikos1337:
@alv2017 [pytest-profiling](https://pypi.org/project/pytest-profiling/)

dikos1337:
> Could you be more specific about which parts of my comments you have addressed? Did you see my comment on line highlighting?


I don't quite understand what's wrong with highlighting now. I don't quite understand what's wrong with highlighting now. Can you show a screenshot of the problem?

svlandeg:
I'm actually having a hard time reviewing the changes because you force pushed changes, meaning that I can't see the original commits anymore, or exactly what changed inbetween versions and inbetween comments. It also looks like your original post is unchanged and doesn't reflect the current state of the PR anymore, and the only comment you gave was that you found a way to fix "this" - I don't even know what "this" refers to.

So let's please take a step back, address the comments on the PR as they were and explain your current changes and their implications? And update the outdated information in the PR? 🙏

dikos1337:
@svlandeg The original post is fully consistent with the changes in the current pull request, so I did not change it.

"this" refers to both the issues that the use of @lru_cache was visible in the user documentation and the highlighting of specific lines in the tutorial was wrong.

I changed the speedup percentage from 31% to 24% in the pull request name because it matches the calculations based on the profiler data, and I also compared the speed of running tests in the current pull request and in other pull requests and it is actually closer to 24% than 31%.

svlandeg (approved):
These changes look fine to me, but I'll defer to Tiangolo for a final review. Thanks again for the contribution and explanations, @dikos1337 !

nidhishgajjar:

## Review of PR #13583: Speed up test suite via caching and fixture scopes

Thanks for this well-motivated optimization! The profiling data and measured speedup (~30%) make a strong case. Here are my observations:

### What looks good ✅

- **Proper cleanup**: The `cache_verify_password` fixture correctly restores the original function after yielding, preventing test pollution.
- **Engine disposal**: Adding `mod.engine.dispose()` in the SQL tests is good practice for releasing database connections.
- **Well-scoped changes**: All changes are confined to test code, so no production risk.
- **Clear profiling data**: The before/after screenshots and measured times (24.37s → 16.85s) are compelling.

### Minor observations 💡

1. **Monkey-patching `verify_password`**: The approach of replacing `mod.verify_password` with a cached version works because `lru_cache` keys on input arguments, so it remains correct even if future tests add new user credentials. This is safe.

2. **`scope="module"` and shared database state**: With `get_client` now module-scoped, `clear_sqlmodel()` runs once per module instead of per test. This is fine as long as the test functions within each module don't require independent clean database states. Looking at `test_crud_app`, it tests a full CRUD sequence in a single function, so there shouldn't be inter-test interference.

3. **`cache_verify_password` assert**: The `assert hasattr(mod, 'verify_password')` is a nice defensive check. Consider using `pytest.skip()` instead of a bare `assert` — a hard failure might be confusing if this fixture is accidentally included in a module that doesn't have `verify_password`. That said, since it's in a specific test file, this is fine as-is.

### Summary

A solid, well-measured optimization to the test suite with proper cleanup and no production impact. The ~30% speedup is a nice win for developer experience.

**Assessment: approve**

tiangolo (approved):
Nice, thank you! 🚀 

And thanks for the review @svlandeg 🙌

roman-postnov:
@tiangolo Thanks for merging my PR. I wanted to point out that apparently, since the PR was created, mod.engine.dispose() was added to the code. I also added mod.engine.dispose() in this PR, so now it's being called twice.

tiangolo:
Oh, dang, I'll check it, thanks!

PR #13589, opened 2025-04-06
♻️ Convert all instances of assert for safer error handling

saltnepperson:


svlandeg (commented):
Hi @saltnepperson, thanks for the PR!

Usually, there's a good reason to use an `assert` instead of an `if` check or similar. Asserts are used when we assume that a certain condition holds at that point - usually because other parts of the code have ensured it. They ensure programmatic consistency, whereas `if` checks usually check user input and are more user-facing.

As there are no comments in the PR description, I don't think we should accept a blanket change of all `assert` statements to `if` checks. I do think that some of the edits in this PR could be valuable, but reviewing them all at once without further comments or reasoning is not ideal. Some of these could be accepted as separate PRs though, especially when `assert` statements are user-facing.

As also the tests are failing and the branch has several conflicts, I'll go ahead and close this one.

saltnepperson:
@svlandeg Thank you for the reply and feedback! I intended to come back to this PR and clean it up, but unfortunately did not get the time to do so. I appreciate the comments and will use them to make a more targeted PR for more reasonable changes to the assert logic.

PR #13605, opened 2025-04-11
🐛 Ensure handlers receive `fastapi.UploadFile` instances, not Starlette's

mlb-:
To start, thank you all for FastAPI! It's been a a joy to use FastAPI at $DAYJOB.

## The problem
While writing a unit test for an endpoint that handles a file, I stumbled across a discrepancy in what FastAPI [documentation describes](https://fastapi.tiangolo.com/tutorial/request-files/#file-parameters-with-uploadfile) and what actually gets passed to the handler (a `starlette.datastructures.UploadFile` via `starlette.requests.Request.form`).

The coding standard at $DAYJOB includes [typeguard](https://github.com/agronholm/typeguard) (via the `pytest-typeguard` plugin). In the unit test, typeguard enforcement fails because the upload file was not a FastAPI instance, but a Starlette instance. (Of course, bypassing typeguard allows the tests to pass, as expected.)
> typeguard.TypeCheckError: argument "file" (starlette.datastructures.UploadFile) is not an instance of fastapi.datastructures.UploadFile

A quick look at the implementation reveals the `UploadFile` creation is handled by `starlette.requests` and then passed to FastAPI's `run_function_endpoint`, without any transforms to a `fastapi.datastructures.UploadFile`. Thankfully, in practice, this doesn't appear to be an issue at runtime, since the FastAPI UploadFile merely extends the Starlette UploadFile to add Pydantic support.

## A solution
- An `UploadFile.from_starlette` classmethod is added to transform from a `StarletteUploadFile` to Starlette's `UploadFile`. FastAPI is already aware of the Starlette UploadFile implementation because it's extending it, so adding the transform to the FastAPI class felt like adding a missing puzzle piece
- This can be applied in `fastapi.routing.run_endpoint_function` to check the `values` dict for non-FastAPI `UploadFile`s and convert them prior to handling off to any endpoint function
- Finally, a quick test that doesn't require typeguard to demonstrate the type signature is consistent

## Thanks again!

I look forward to feedback!


svlandeg:
Hi, thanks for this contribution! 

It looks like the tests are failing for Python 3.8, have you looked into that?

mlb-:
@svlandeg : Hey, thanks for following up! Let me know if that looks good to you now 🙏

TheCaffinatedDeveloper:
This is great! I need this fix too actually - @svlandeg @mlb- can we get this merged?

YuriiMotov:
As an idea: wouldn't it be better to ask Starlette to add a config option like `file_upload_class: type = UploadFile`?
So that Starlette would create instances of proper class and we don't need to re-create them?

mlb-:
@YuriiMotov You'll need something more precise (probably a `Protocol` and builder/factory pattern). Starlette controls how `UploadFile` instances are constructed and needs to ensure they conform to expected async behavior and internal interfaces.
Allowing arbitrary subclass injection might be more convenient for us as a consumer, but it’s not a trivial change; it would incur maintenance and support costs for Starlette.

Since FastAPI is the one customizing `UploadFile`, it’s appropriate for us to take responsibility for the explicit conversion.

YuriiMotov:
> Allowing arbitrary subclass injection might be more convenient for us as a consumer, but it’s not a trivial change; it would incur maintenance and support costs for Starlette.

What are possible problems you see if they restrict this class to be inherited from their `UploadFile`?

mlb-:
> > Allowing arbitrary subclass injection might be more convenient for us as a consumer, but it’s not a trivial change; it would incur maintenance and support costs for Starlette.
> 
> What are possible problems you see if they restrict this class to be inherited from their `UploadFile`?

As currently implemented, none. If you could get them to accept that feature, we'll need to modify this PR to use your proposed configuration and make sure the tests still pass.

mlb-:
@tiangolo: Do you think this is an easy enough bugfix you could merge?

fastapi/routing.py
```diff
@@ -215,6 +216,19 @@ async def run_endpoint_function(
     # facilitate profiling endpoints, since inner functions are harder to profile.
     assert dependant.call is not None, "dependant.call must be a function"
 
+    # Convert all Starlette UploadFiles to FastAPI UploadFiles
+    for key, value in values.items():
+        if isinstance(value, StarletteUploadFile) and not isinstance(value, UploadFile):
```
YuriiMotov: Do we really need that `and not isinstance(value, UploadFile)` part in condition?

YuriiMotov (dismissed):
We finally discussed with Kludex the idea of adding `upload_file_class` parameter to Starlette, and it's now clear this is unlikely to be implemented on their side. So, let's go the way suggested in this PR then!

All looks good to me, let's just discuss whether we really need the `and not isinstance(value, UploadFile)` part in the conditions.

**UPD:** we should extend the implementation to also cover `UploadFile` parameters in dependency functions

fastapi/routing.py
YuriiMotov: I investigated a bit more and came up with the following use case (quite unrealistic):

```py
from fastapi import Depends, FastAPI, File, UploadFile

app = FastAPI()

async def dep(uploadfile: UploadFile = File(...)):
    return UploadFile.from_starlette(uploadfile)

@app.post("/uploadfile")
async def uploadfile(uploadfile: UploadFile = Depends(dep)):
    assert isinstance(uploadfile, UploadFile)
    return {}
```

In this case, the value passed to endpoint function will already be `fastapi.UploadFile`.

And, this use case exposed an issue in this implementation - it only converts values for endpoint function, but in dependencies we will still have `starlette.UploadFile`.

@mlb-, will you be able to take a look at this?

YuriiMotov (approved):
LGTM!

PR #13614, opened 2025-04-14
✨ Accept multiple status codes or exceptions in `exception_handler`

YuriiMotov:
## Description

This PR allows passing multiple status codes or exceptions to `exception_handler`.

```
@app.exception_handler((401, 403))
def auth_error_handler(request, exc):
    pass
    
@app.exception_handler((FileTooLargeError, UnsupportedFileTypeError))
def file_upload_error_handler(request, exc):
    pass
```

## Motivation

This feature is requested by #13242 and seems to be quite nice to have.

## Implementation details

In this PR I implemented it as a Union type. This way seems to be widely used in popular libraries like `unittest`, `pytest`, `backoff`, `tenacity`.

You can pass parameters as any `Sequence` (`tuple` or `list`).

I intentionally limited it to only allow passing a sequence of items of similar type (it's not allowed to mix status codes with exception classes at the same time). If you think this limitation is not needed, I will update PR to remove it.

```
# For now the following line will cause typing error
@app.exception_handler((413, UnsupportedFileTypeError))
def file_upload_error_handler(request, exc):
    pass
```

## Alternatives
There is already existing #13241 that does it in a bit different way (it accepts additional status codes or exceptions as `*extra_exc_classes_or_status_codes`).

It's already possible to re-use handlers for different errors by adding multiple decorators:
```
@app.exception_handler(401)
@app.exception_handler(403)
def auth_error_handler(request, exc):
    pass
```

## Checklist

- [x] Tests added (including tests for code examples)
- [x] Docs updated


YuriiMotov:
Feel free to close it if you think we don't need this.
I created it just as an alternative to #[13241](https://github.com/fastapi/fastapi/pull/13241)

YuriiMotov:
I don't think this will ever be accepted - there is already a way to add exception handler for multiple codes ([see](https://github.com/fastapi/fastapi/discussions/13242#discussioncomment-12826955)) and we have other priorities now.
I'll close this.

PR #13618, opened 2025-04-14
📝 Add event-driven HowTo section in the documentation

Flosckow:
Event-driven patterns (e.g., microservices, streaming data) are increasingly common. I think we should add information about event-driven service & hybrid services development to FastAPI documentation. Providing official guidance reduces friction for FastAPI users adopting messaging systems.

Also, I suggest adding information about FastStream framework integration to such a section. This framework has been growing for 2 years already, continuously adding new features and increasing installations number. For now, it is a mature project for event-driven system development, used by hundreds of companies all over the world.

Moreover FastStream has a perfect integration with FastAPI  – https://faststream.airt.ai/latest/getting-started/integrations/fastapi/

Documenting the framework's integration helps users build hybrid systems that handle both HTTP requests and asynchronous event processing seamlessly. Both frameworks leverage modern Python (async/await, Pydantic, type hints). By showcasing their compatibility, we empower developers to combine REST APIs with real-time event-driven logic without redundant boilerplate.


kumaranvpl (approved):

tiangolo:
Thanks for the interest! I currently prefer to keep the docs as they are, so I'll pass on this. :coffee:

PR #13626, opened 2025-04-18
✨ Add tags to websockets

jmahoney-eab:
Added parity for websocket api routes to contain tags.

Websockets do not seem to be included in openapi schemas, so those tests did not need updates.


jmahoney-eab:
Happy to add any additional tests or missed functionality.

jmahoney-eab:
I need to add a test for adding tags to a websocket route that is part of a router which also has tags

PR #13627, opened 2025-04-19
✨ Add support for Starlette's scoped middleware

FatimahAlhumaidhi:
@Kludex created a nice way to allow scoped middleware in starlette:
*  https://github.com/encode/starlette/pull/2349

this PR aims to bringing this very helpful feature to FastAPI :)

I've added docs, tutorial, and a test to the tutorial


Kludex:
I'm all pro to this, but @tiangolo sais he wanted to do something before... I don't recall what.

Tonkonozhenko:
It would be great to have this MR released

Tonkonozhenko:
@tiangolo, could you review the PR, please?
jfyi: our use case is that we have an app, we have an api router that handles incoming webhooks, and we want to implement request-response logging using middleware to have it in 1 place in the code

FatimahAlhumaidhi:
> @tiangolo, could you review the PR, please? jfyi: our use case is that we have an app, we have an api router that handles incoming webhooks, and we want to implement request-response logging using middleware to have it in 1 place in the code

@Tonkonozhenko  I think you can do this with a dependency
somehting like:
```python
async def logging(
    request: Request,
    response: Response,
):
    yield
    # log here
```

I had a similar use case when I suggested this pr, I wanted to track events on some endpoints (while also collecting some context)
so I created this dependency:
```python
def track_event(
    event_type: EventType,
    initial_props: Optional[dict[str, Any]] = None,
):
    async def _tracker(
        request: Request,
        response: Response,
        background_tasks: BackgroundTasks,
    ):
        start_time = time.monotonic()
        props = dict(initial_props) if initial_props else {}
        try:
            yield props
        finally:
            elapsed_ms = (time.monotonic() - start_time) * 1000
            background_tasks.add_task(
                collect_event,
                request,
                event_type,
                props,
                response.status_code,
                response_time_ms=elapsed_ms,
            )

    return Depends(_tracker)
```
and added something like `_track: dict[str, Any] = track_event(EventType.MY_EVENT),` in the route function

Tonkonozhenko:
@FatimahAlhumaidhi I'll definitely try it

tiangolo:

Thank you! 🍰

PR #13628, opened 2025-04-19
📝 Add docs for IoC containers using Dishka as an example

EgorOnishchuk:
With the **dependency injection** (DI) pattern becoming increasingly common — especially due to its popularization through FastAPI — it's important to clarify when the built-in `Depends` mechanism is sufficient, and when using a third-party DI library is more appropriate.

This PR adds a new guide to the _«How To»_ section that explores the difference between **dependency injection** and **dependency inversion**, and how they relate. It combines a relatable real-life analogy with practical, detailed code examples.

I chose **Dishka** as the third-party integration example because:

* **it has strong community support and many integrations, including with FastAPI;**

* **it builds on lessons learned from earlier DI libraries;**

* **it efficiently resolves complex dependencies graphs;**

* **it fully complies with the Dependency Inversion Principle (DIP);**

* **it is async- and type-oriented — just like FastAPI;**

* **it provides an extensive, flexible API ready for an almost boundless scalability even with a monolith.**

The guide is written in a friendly yet technical tone, consistent with the rest of the documentation. It includes helpful links to related sections and aims to encourage developers — including newcomers — to adopt _best practices_ around dependency management.

Special thanks for the inspiration to make this PR:

* **@Tishka17 and all the Dishka contributors;**

* **a great [article](https://habr.com/ru/articles/867040/) on DI in FastAPI (in a non-English language);**

* **a helpful [curated list](https://github.com/sfermigier/awesome-dependency-injection-in-python) of Python DI tools.**


YuriiMotov:
What if we simplify this part of code example:

```python
def get_city_client(settings: ClientSettingsDep) -> CityClient:
    return CityClient(settings=settings)


CityClientDep = Annotated[CityClient, Depends(get_city_client)]


def get_village_client(settings: ClientSettingsDep) -> VillageClient:
    return VillageClient(settings=settings)


VillageClientDep = Annotated[VillageClient, Depends(get_village_client)]


geo_router = APIRouter(prefix="/geo", tags=["Geo"])


@geo_router.get(
    "",
    response_class=PlainTextResponse,
    summary="Fetches geodata.",
)
async def get(client: CityClientDep) -> str:
    return await client.request()
```

to this:

```python
def get_client(settings: ClientSettingsDep) -> Client:
    return CityClient(settings=settings)


ClientDep: TypeAlias = Annotated[Client, Depends(get_client)]


geo_router = APIRouter(prefix="/geo", tags=["Geo"])


@geo_router.get(
    "",
    response_class=PlainTextResponse,
    summary="Fetches geodata.",
)
async def get(client: ClientDep) -> str:
    return await client.request()
```

We don't need `VillageClientDep` and `CityClientDep` at the same time. If we need to switch between them, we can just modify `return CityClient(settings=settings)` to `return VillageClient(settings=settings)`.

Now, most of the following points are irrelevant:

> 
> So, we need:
> 
> A factory function for each dependency.
> A construction with Annotated for each dependency to avoid duplicating Depends if one dependency will be used in multiple path operations.
> A unique name for each construction with Annotated so as not to shadow the origin class name.
> A lru_cache decorator for each singleton.
> While manageable at first, this list grows quickly and spreads across your codebase — making maintenance more difficult.
> 
> And the most important, path operations depend on the concrete implementation, not the abstraction — violating the [Dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle).

EgorOnishchuk:
Thank you for feedback, @YuriiMotov

> ```python
> def get_city_client(settings: ClientSettingsDep) -> CityClient:
> ```

You are right — I forgot to add `async`, but I'll fix that in the next commit to avoid creating an expensive thread in the pool.

> We don't need `VillageClientDep` and `CityClientDep` at the same time. If we need to switch between them, we can just modify `return CityClient(settings=settings)` to `return VillageClient(settings=settings)`.

This approach is relevant for simple scenarios. Since this page covers _advanced DI_, it is better to assume that the reader may encounter more complex ones now or in the future. Here are two real cases.

1. Imagine we need to check something related to the endpoint (testing, benchmarking, etc.) with each of the clients. We can do it like this:

```python
def app_for_check(dep: Callable) -> FastAPI:
    app = main()
    app.dependency_overrides[get_city_client] = dep

    return app


async def check(dep: Callable, route: str, method: str) -> Response:
    async with AsyncClient(
        transport=ASGITransport(app=app_for_check(dep)), base_url="http://test"
    ) as session:
        return await session.request(method=method, url=route)


async def main_() -> None:
    responses = await asyncio.gather(
        *(
            check(dep=dep, route="/api/geo", method="GET")
            for dep in (get_village_client,)
        )
    )

    print(*(response.text for response in responses), sep='\n')
```

Since in your code the «options for swapping» are not pre-defined, we'll need to declare them on the fly, violating the _single responsibility principle_, and duplicate declarations in other parts of the application where runtime overriding is needed.

```python
async def main_() -> None:
    result = await asyncio.gather(
        *(
            check(dep=dep, route="/api/geo", method="GET")
            for dep in (
                lambda settings=Depends(get_client_settings): VillageClient(
                    settings=settings
                ),
            )
        )
    )

    print(*(response.text for response in result), sep="\n")
```

2. Even if you don't need overriding, modifying the function body code introduces a lot of mechanical input errors. Replacing `return CityClient(settings=settings)` with `return VillageClient(settings=settings)` seems simple, but even implementations of the same interface can have different number of subdependencies. In this case, you will have to change the factory signature as well, and remember (or spend time searching for) the code of all the swaps. Hypothetical example:
```python
def get_city_client(settings: ClientSettingsDep) -> CityClient:
    return CityClient(settings=settings)

def get_village_client(settings: ClientSettingsDep, ip_service: IpServiceDep) -> VillageClient:
    ip = ip_service.get_ip()
    return VillageClient(settings=settings, ip=ip)
```

I didn't include these examples in the documentation because long ones are hard to consume and overload readers. There is a warning at the beginning of the page about the need to have a good understanding of DI, so I don't see the reason to give them explicitly. If you think the benefits of multiple factories are not obvious to those who have learn the whole [DI section](https://fastapi.tiangolo.com/tutorial/dependencies), let me know — I will happy to make improvements.

YuriiMotov:
@EgorOnishchuk , thanks for the reply!

If the tasks is defined this way, I would create two separate dependency functions and assign one of them to `get_client` variable:

```python
async def get_city_client(settings: ClientSettingsDep) -> CityClient:
    return CityClient(settings=settings)


async def get_village_client(
    settings: ClientSettingsDep,
    ip_service: Annotated[IpService, Depends(get_ip_service)],
) -> VillageClient:
    ip = ip_service.get_ip()
    return VillageClient(settings=settings, ip=ip)


get_client = get_village_client
```

This way it will be easy to switch between those dependencies both in main code (just change `get_client = get_village_client` to `get_client = get_city_client`).

And, code for tests\benchmarking can be rewritten as follows:

```python
CLIENT_DEP_OPTIONS = (get_city_client, get_village_client)


def get_app() -> FastAPI:
    return main()


async def check(client_dep_override: Any, route: str, method: str) -> Response:
    app = get_app()
    app.dependency_overrides[get_client] = client_dep_override

    async with AsyncClient(
        transport=ASGITransport(app=app), base_url="http://test"
    ) as session:
        return await session.request(method=method, url=route)


async def main_() -> None:
    tasks = [
        check(client_dep_override=dep, route="/api/geo", method="GET")
        for dep in CLIENT_DEP_OPTIONS
    ]
    responses = await asyncio.gather(*tasks)

    print(*(response.text for response in responses), sep="\n")


if __name__ == "__main__":
    asyncio.run(main_())
```

See full code in details:

<details>

```python
import asyncio
import random
from abc import ABC, abstractmethod
from functools import lru_cache
from typing import Annotated, Any, TypeAlias

from fastapi.params import Depends
from fastapi.responses import PlainTextResponse
from httpx import ASGITransport, AsyncClient, Response


@lru_cache
@dataclass
class ClientSettings:
    timeout: int


@dataclass
class Client(ABC):
    settings: ClientSettings

    async def _connect(self) -> None:
        await asyncio.sleep(min(random.randint(1, 5), self.settings.timeout))

    @abstractmethod
    async def request(self) -> str:
        raise NotImplementedError


class CityClient(Client):
    async def request(self) -> str:
        await self._connect()
        return "Response"


@dataclass
class VillageClient(Client):
    ip: str

    async def request(self) -> str:
        await self._connect()

        return f"Response, {self.ip}"


async def get_client_settings() -> ClientSettings:
    return ClientSettings(timeout=3)


ClientSettingsDep: TypeAlias = Annotated[ClientSettings, Depends(get_client_settings)]


@lru_cache
@dataclass
class IpService:
    ip: str

    def get_ip(self) -> str:
        return self.ip


async def get_ip_service() -> IpService:
    return IpService(ip="192.168.0.1")


async def get_city_client(settings: ClientSettingsDep) -> CityClient:
    return CityClient(settings=settings)


async def get_village_client(
    settings: ClientSettingsDep,
    ip_service: Annotated[IpService, Depends(get_ip_service)],
) -> VillageClient:
    ip = ip_service.get_ip()
    return VillageClient(settings=settings, ip=ip)


get_client = get_village_client


geo_router = APIRouter(prefix="/geo", tags=["Geo"])


@geo_router.get(
    "",
    response_class=PlainTextResponse,
    summary="Fetches geodata.",
)
async def get(client: Annotated[Client, Depends(get_client)]) -> str:
    return await client.request()


def main() -> FastAPI:
    app = FastAPI(title="FastAPI Dependsency Injection is good enough 🙂")

    for router in (geo_router,):
        app.include_router(router, prefix="/api")

    return app


# ====================================================================================

CLIENT_DEP_OPTIONS = (get_city_client, get_village_client)


def get_app() -> FastAPI:
    return main()


async def check(client_dep_override: Any, route: str, method: str) -> Response:
    app = get_app()
    app.dependency_overrides[get_client] = client_dep_override

    async with AsyncClient(
        transport=ASGITransport(app=app), base_url="http://test"
    ) as session:
        return await session.request(method=method, url=route)


async def main_() -> None:
    tasks = [
        check(client_dep_override=dep, route="/api/geo", method="GET")
        for dep in CLIENT_DEP_OPTIONS
    ]
    responses = await asyncio.gather(*tasks)

    print(*(response.text for response in responses), sep="\n")


if __name__ == "__main__":
    asyncio.run(main_())

```

</details>

YuriiMotov:
IMHO, the only strong argument is that FastAPI's `Depends` can't be used outside of FastAPI endpoints and sub-dependencies.
So, I would focus on such use-cases to create really good code example.

Anyway, I would recommend not to expect this PR will be merged soon (just my observations). Better post this as an article on external resource and add the link\description to Discussions ("Show and tell") and to the list of external resources (https://fastapi.tiangolo.com/external-links/)

Tishka17:
Let me, as an author of dishka, add some points here. Fastapi Depends works quite well untill it doesn't. In my cases I've encounterd several problems:

* It doesn't distinguish input data and dependencies. So, you can get unexpected things in swagger which were expected to be dependencies.
* It requires from you to put `Depends` in factory, so you cannot actually use extisting class as-is
* It cannot be used outside of fastapi application
* It cannot create "singletons", and `lrucache` is not suitable in tests
* It cannot create dependencies per websocket message, only for connection
* There are bugs like incorrect anylisis of `Protocol` classes

So, I implemented several ideas in my lib:
* Container should be detached from web framework so it can be used in a wide range of applications
* There should be several "Scopes" meaning lifetime of objects
* It should not implicitly take decisions so user can control how each thing is created
* No modification of original classes should be required

I hope these thoughts can be somehow covered in extended docs.

tiangolo:
Thanks for the interest! For now I prefer to keep the docs as they are, so I'll pass on this one. :coffee:

PR #13639, opened 2025-04-23
✨ Allow array values for OpenAPI schema `type` field

sammasak:
# Allow Union Types in OpenAPI Schemas
This PR addresses an issue with how union types for the "type" field are validated in FastAPI's OpenAPI generation. Currently, only a string is allowed for the type definition even though the JSON Schema specification (Section 6.1.1) permits the value to be either a string or an array of unique string values. This discrepancy can lead to confusion or incorrect schema validation when using union types.

Example here that throws pydantic validation error:

```python
# example.py 
# uv run --no-project example.py
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "fastapi",
#     "pydantic",
# ]
# ///


from fastapi.openapi.models import OpenAPI
from pydantic import ValidationError

openapi_dict = {
    "openapi": "3.1.0",
    "info": {"title": "Type Examples with Union Types", "version": "1.0.0"},
    "components": {
        "schemas": {
            "UnionStringOrNumber": {
                "type": ["string", "number"],
                "description": "An instance is valid if it is either a string or a number.",
                "example": "example as string",
            },
        }
    },
}


def test_openapi_type() -> None:
    try:
        _ = OpenAPI(**openapi_dict)
    except ValidationError as e:
        # Expecting this to raise a ValidationError
        print(f"Validation error: {e}")
        return

    raise NotImplementedError("This should not be reached.")


if __name__ == "__main__":
    test_openapi_type()
    print("Test passed!")
```

Addressed by introducing`TypeValue`, defined as an annotated literal that encompasses [the valid JSON Schema types.](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.1.1)

And updating the Schema model to use the literal values.

After these changes the validation error is removed.

### DIFF
--- a/tests/test_openapi_schema_type.py
+++ b/tests/test_openapi_schema_type.py
@@ -1,8 +1,7 @@
 from typing import List, Optional, Union
 
 import pytest
-from fastapi.openapi.constants import TypeValue
-from fastapi.openapi.models import Schema
+from fastapi.openapi.models import Schema, SchemaType
 
 
 @pytest.mark.parametrize(
@@ -14,9 +13,9 @@ from fastapi.openapi.models import Schema
     ],
 )
 def test_allowed_schema_type(
-    type_value: Optional[Union[TypeValue, List[TypeValue]]],
+    type_value: Optional[Union[SchemaType, List[SchemaType]]],
 ) -> None:
-    """Test that Schema accepts TypeValue, List[TypeValue] and None for type field."""
+    """Test that Schema accepts SchemaType, List[SchemaType] and None for type field."""
     schema = Schema(type=type_value)
     assert schema.type == type_value
 
--- a/tests/test_openapi_schema_type.py
+++ b/tests/test_openapi_schema_type.py
@@ -1,50 +1,27 @@
-import itertools
-from typing import List
+from typing import List, Optional, Union
 
 import pytest
 from fastapi.openapi.constants import TypeValue
 from fastapi.openapi.models import Schema
 
-# Define all possible type values
-TYPE_VALUES: List[TypeValue] = [
-    "array",
-    "boolean",
-    "integer",
-    "null",
-    "number",
-    "object",
-    "string",
-]
 
-# Generate all combinations of 2 or more types
-TYPE_COMBINATIONS = [
-    list(combo)
-    for size in range(2, len(TYPE_VALUES) + 1)
-    for combo in itertools.combinations(TYPE_VALUES, size)
-]
+@pytest.mark.parametrize(
+    "type_value",
+    [
+        "array",
+        ["string", "null"],
+        None,
+    ],
+)
+def test_allowed_schema_type(
+    type_value: Optional[Union[TypeValue, List[TypeValue]]],
+) -> None:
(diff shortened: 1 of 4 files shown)


sammasak:
Hey @svlandeg thanks for adding the label, what does the review process look like and can I help out in anyway? 😄

YuriiMotov:
Similar, but `set[str]` is used: https://github.com/fastapi/fastapi/pull/12528

I personally like the approach with `list` more.

@sammasak, could you please add tests?

sammasak:
@YuriiMotov Added some tests, but the change in this PR is quite small and is only made to allow for type to be a str and list. But since there is no field validator for type its still possible to break the json schema. (by setting type to a empty list or by adding duplicates. https://json-schema.org/draft/2020-12/json-schema-validation#section-6.1.1 ) 

 But I think this is a good start and more validation can be added later. wdyt?

tests/test_openapi_schema_type.py
```diff
@@ -0,0 +1,50 @@
...
+    """Test that Schema accepts single TypeValue for type field."""
+    schema = Schema(type=type_val)
+    assert schema.type == type_val
+
+
+@pytest.mark.parametrize("type_list", TYPE_COMBINATIONS)
+def test_schema_type_multiple_type_value(type_list: List[TypeValue]) -> None:
+    """Test all possible combinations of TypeValue for Schema type field."""
+    schema = Schema(type=type_list)
+    assert schema.type == type_list
```
YuriiMotov:
```suggestion
def test_schema_type_multiple_type_value() -> None:
    """Test all possible values of TypeValue for Schema type field."""
    schema = Schema(type=TYPE_VALUES)
    assert schema.type == TYPE_VALUES
```

I think now it's too much tests for such small feature. 🤣
Let's just test one combination of all possible values?

YuriiMotov (commented):
Let's reduce the number of test a bit :)
Just test that all valid types are accepted and maybe add one test for validation error on passing incorrect type

> Added some tests, but the change in this PR is quite small and is only made to allow for type to be a str and list. But since there is no field validator for type its still possible to break the json schema. (by setting type to a empty list or by adding duplicates
> But I think this is a good start and more validation can be added later. wdyt?


I agree, no need to add this validation now.

tests/test_openapi_schema_type.py
```diff
@@ -0,0 +1,27 @@
...
+)
+def test_allowed_schema_type(
+    type_value: Optional[Union[TypeValue, List[TypeValue]]],
+) -> None:
+    """Test that Schema accepts TypeValue, List[TypeValue] and None for type field."""
+    schema = Schema(type=type_value)
+    assert schema.type == type_value
+
+
+def test_invlid_type_value() -> None:
```
YuriiMotov:
```suggestion
def test_invalid_type_value() -> None:
```

YuriiMotov (approved):
LGTM

@sammasak, thanks!

tiangolo (approved):
Sounds good, thanks! 🚀 

I tweaked it a bit to put the OpenAPI types and related on the same file, it should be ready now. 🍰

PR #13641, opened 2025-04-23
📝 Remove unnecessary bullet from docs

Adamowoc:
Removing a bullet for doc uniformity.


alejsdev:
Makes sense, thank you @Adamowoc :raised_hands:

PR #13642, opened 2025-04-23
📝 Clarify guidance on using `async def` without `await`

swastikpradhan1999:
### Summary
This PR improves the clarity of the **"Concurrency and async/await"** section in the documentation.

The original sentence: *"If your application (somehow) doesn't have to communicate with anything else and wait for it to respond, use async def."*

has been updated to:

*"If your application doesn't need to communicate with anything else or await a response, it's still safe to use `async def` -- even if there's nothing to `await`."*

### Why this change?
The original phrasing could be misinterpreted as recommending `async def` *even when it’s not useful*. The new wording keeps the reassuring tone but clarifies that using `async def` without `await` is safe — not required.

 Notes
- This update aims to reduce confusion, especially for developers new to asynchronous programming in Python.
- No functional or structural changes were made to the documentation layout.

### DIFF
--- a/docs/en/docs/async.md
+++ b/docs/en/docs/async.md
@@ -40,7 +40,7 @@ def results():
 
 ---
 
-If your application doesn't need to communicate with anything else or await a response, it's still safe to use `async def` -- even if there's nothing to `await`.
+If your application (somehow) doesn't have to communicate with anything else and wait for it to respond, use `async def`, even if you don't need to use `await` inside.
 
 ---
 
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks!

tiangolo:
Thanks! I updated it a bit to revert some of the changes but clarify the intention.

The idea is that when people have something that could block the event loop, they *should* use regular `def` functions, because those are run in a thread worker.

But when they don't have something that would block the event loop, they would get better performance with async functions, because those are not run in a thread worker.

That's why the intention of the phrasing is more like a "should" than a "could".

Thanks! :coffee:

tiangolo (approved):

PR #13650, opened 2025-04-26
📝 Add "Global Responses" section

timonrieger:
This PR adds a new section to the [Additional Responses](https://fastapi.tiangolo.com/advanced/additional-responses) page. It documents how to register responses globally, so that they are applied to every endpoint.

### DIFF
--- a/docs/en/docs/advanced/additional-responses.md
+++ b/docs/en/docs/advanced/additional-responses.md
@@ -191,6 +191,20 @@ But if you have specified a custom response class with `None` as its media type,
 
 ///
 
+## Global Responses
+
+You can also register responses at the application level.
+This automatically adds the specified response models to every endpoint, without needing to define them individually.
+Use the same syntax as demonstrated above.
+
+```Python
+class MyErrorModel(BaseModel):
+    message: str
+    code: str
+
+app = FastAPI(responses={"default": {"model": MyErrorModel, "description": "Error response"}})
+```
+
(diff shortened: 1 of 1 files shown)


docs/en/docs/advanced/additional-responses.md
```diff
@@ -191,6 +191,20 @@ But if you have specified a custom response class with `None` as its media type,
 
 ///
 
+## Global Responses
+
+You can also register responses at the application level.
```
YuriiMotov, line 196: It's also possible to register responses at router level as well

docs/en/docs/advanced/additional-responses.md
```diff
@@ -191,6 +191,20 @@ But if you have specified a custom response class with `None` as its media type,
...
+This automatically adds the specified response models to every endpoint, without needing to define them individually.
+Use the same syntax as demonstrated above.
+
+```Python
```
YuriiMotov, line 200: Code example should be moved to a source file in `docs_src/additional_responses/` and included here.
That file should covered by tests (look how it's done for other code examples).
I think it can be one file for both, application level and router level (that I believe should be also described here) responses

YuriiMotov (changes requested):
@timonrieger, thanks for your interest and efforts!

All code examples should be placed in separate modules and should be covered by tests.
Also, for the sake of completeness, I think we need to mention router level responses.

PR #13674, opened 2025-05-01
✅ Improve test coverage

Walancy:


YuriiMotov (changes requested):
@Walancy, thanks for the interest and efforts!

Could you please provide some reasoning on why we need all these tests? Have you checked all these cases are missed?

Also, we must use English language only. Using other languages makes it hard for others to review and maintain this code.

Lastly, some tests failed with Pydantic V1

PR #13680, opened 2025-05-03
📝 Swap install and run code sections order in `docs/tutorial/index.md`

liby:
## Description

This PR changes the order of sections in the tutorial index files across multiple language translations. It moves the "Install FastAPI" section before the "Run the code" section, which is more logical since users need to install FastAPI before they can run any code examples.

## Changes

- Reordered sections in `docs/tutorial/index.md` files for the following languages:
  - English (en)
  - Simplified Chinese (zh)
  - Traditional Chinese (zh-hant)
  - Spanish (es)
  - German (de)
  - Portuguese (pt)
  - Indonesian (id)
  - Japanese (ja)
  - Korean (ko)
  - Polish (pl)
  - Russian (ru)
  - Ukrainian (uk)
  - Vietnamese (vi)
  - Emoji version (em)
  - French (fr)

## Motivation

The original order was inconsistent with the natural flow of onboarding new users. It makes more sense to first explain how to install FastAPI before showing users how to run code examples, as installation is a prerequisite for running the code.

## Checklist

- [x] I have read the [Contributing Guide](https://fastapi.tiangolo.com/contributing/)
- [x] Documentation content is consistent across all translations
- [x] No code examples were modified, only the section order


YuriiMotov (commented):
@liby, thanks for the suggestion!

I think there is no need to change the order of sections here.

"Run the code" section just shows how the process looks like, it doesn't require any immediate actions from user (there is no code at this point). So, readers don't need to know how to install FastAPI to understand this section

liby:
If we’re already putting the code demo and installation guide together in the same section, why not arrange them in a way that’s more ergonomic and user-friendly?

YuriiMotov:
> If we’re already putting the code demo and installation guide together in the same section, why not arrange them in a way that’s more ergonomic and user-friendly?

I understand your frustration..
But to merge this we need to review it twice (I should review it carefully and approve, then Sebastián will need to take a look and take a final decision).
We have a lot of PRs and limited time, so we need to set priorities.

Hope for your understanding

PR #13689, opened 2025-05-06
♻️ Create `dependency-cache` dict in `solve_dependencies` only if `None` (don't re-create if empty)

bokshitsky:
Hello,
Current code:
1) Creates new `dependency_cache` dictionary inside `solve_dependencies` function if passed `dependency_cache` parameter is None or empty.
2) As a result of (1) `dependency_cache.update(solved_result.dependency_cache)` is required at every iteration of dependencies solving, because `solved_result.dependency_cache` returned by recursion can be either same as `dependency_cache` (if at least one dependency was already stored in cache) or new dict (on first iterations of recursion where empty cache is passed to recursive function)

This behaviour adds a bit complexity to the code and spends (albeit a little) more CPU and memory resources.

My proposal:
1) always create dependency cache on first iteration of recursive dependency solving and pass created cache instance to all subsequent calls. Do not exchange cache dict on first steps of recursion iterations if passed dict was empty.
2) Now `dependency_cache.update(solved_result.dependency_cache)` is not required, because one `dependency_cache ` is used over all dependency solving process.


bokshitsky:
@tiangolo hi, may I ask you or any other maintainer to look my PR?

bokshitsky:
@YuriiMotov sorry for mention you, can you please give any advice that can I do to move this PR further?

I think these changes a pretty safe and improve a bit performance and mem-print, although in my company we need them to remove hacks relates to manual solve_dependencies usage out of fastapi-http-scope

YuriiMotov (approved):
LGTM

These changes make code clearer and less error-prone.

@bokshitsky, I would recommend you update the description to make it more concise

bokshitsky:
@YuriiMotov thanks, I updated, I think changes are clearer now.

bokshitsky:
@tiangolo hello, may you look please - can I do something to include patch in following releases?

tiangolo (approved):
Sounds good! 🚀 

Thanks for the detailed explanation. 🍰

PR #13695, opened 2025-05-07
✨ Allow including custom scripts to Swagger UI

vk496:
This fix the "StandaloneLayout" layout + enable user to add any other scripts to the Swagger UI through **swagger_ui_scripts**.

Before this PR, this code would never loaded
```
from fastapi import FastAPI

app = FastAPI(
    swagger_ui_parameters={
        "layout": "StandaloneLayout",
    },
)


@app.get("/")
async def root():
```

Now it will show the Swagger layout:
image

NOTE: The main bug comes from bad copy-paste of https://swagger.io/docs/open-source-tools/swagger-ui/usage/installation. The example used in FastAPI come from pure node. The actual example that must be used (JS Vanilla) includes **swagger-ui-standalone-preset.js**

### DIFF
--- a/docs/de/docs/how-to/configure-swagger-ui.md
+++ b/docs/de/docs/how-to/configure-swagger-ui.md
@@ -61,7 +61,7 @@ FastAPI umfasst auch diese Nur-JavaScript-`presets`-Einstellungen:
 ```JavaScript
 presets: [
     SwaggerUIBundle.presets.apis,
-    SwaggerUIBundle.SwaggerUIStandalonePreset
+    SwaggerUIStandalonePreset
 ]
 ```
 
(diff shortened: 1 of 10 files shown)


YuriiMotov:
@vk496, thank you for the interest in FastAPI! 

Would it make sense to turn `swagger_ui_scripts` into a list to preserve the order of scripts?

In this implementation, `https://cdn.jsdelivr.net/npm/swagger-ui-dist@5/swagger-ui-standalone-preset.js` script will always be loaded. There is no way to replace it. It's only possible to add new scripts.
How about turning it from set of default values to separate parameter (something like `swagger_ui_preset_url`) with default value? So that it would be possible to replace it.

Should we mention this new feature in docs?

YuriiMotov:
@vk496, are you still interested in working on this?

PR #13698, opened 2025-05-08
🐛 Fix truncating the model's description with form feed (`\f`) character for Pydantic V2

YuriiMotov:
## Problem description 

```python
app = FastAPI()

class MyModel(BaseModel):
    """
    A model with a form feed character in the title.
    \f
    Text after form feed character.
    """

@app.get("/foo")
def foo(v: MyModel):
    pass
```

The code above will produce different **description** for `MyModel` in schema depending on whether **Pydantic V1** or **Pydantic V2** is used.

With **Pydantic V1** the description will be truncated starting from `\f` character (only first line will be in schema).
With **Pydantic V2** it will not be truncated (all 3 lines will be in schema).

Feature was introduced in #3032 .
But when the support of **Pydantic V2** was added to FastAPI this feature wasn't added to the code related to Pydantic V2 (I think it was just oversight).

## Related discussion(s):
https://github.com/fastapi/fastapi/discussions/9038

## Remark about `cast`
I added commit that improves typing of new part of code by using `cast`. Not sure it's needed. If not, it can be easily reverted.

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -231,6 +231,9 @@ if PYDANTIC_V2:
         field_mapping, definitions = schema_generator.generate_definitions(
             inputs=inputs
         )
+        for item_def in definitions.values():
+            if "description" in item_def:
+                item_def["description"] = item_def["description"].split("\f")[0]
         return field_mapping, definitions  # type: ignore[return-value]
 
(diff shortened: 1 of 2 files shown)


schneebuzz (approved):

tiangolo (approved):
Nice, thank you! 🙌

PR #13699, opened 2025-05-09
📝 Clarify the middleware execution order in docs

YuriiMotov:
At the moment, it's not obvious for newcomers what the order of middleware execution is when more than one middleware is added.

This PR adds clarification on the stacking behavior of middlewares in FastAPI.

Specifically, it explains that the last middleware added is the outermost, which runs first on the request and last on the response.

### DIFF
--- a/docs/en/docs/tutorial/middleware.md
+++ b/docs/en/docs/tutorial/middleware.md
@@ -65,6 +65,29 @@ Here we use <a href="https://docs.python.org/3/library/time.html#time.perf_count
 
 ///
 
+## Multiple middleware execution order
+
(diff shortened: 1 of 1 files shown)


schneebuzz (approved):
Very good addition, I would even add a link to the new section on the top bulletpoints, e.g.


    - Then it passes the request to be processed by the rest of the application (by some path operation *or the next middleware, see [...]*).

jingyugao:
very good mr. Now AI will reply  wrong answer about middleware execute order because lack of documents.

tiangolo (approved):
Nice, thank you! 🚀

PR #13703, opened 2025-05-10
:book: Update typing import for `Global Dependencies` example

rockokw:
This PR updates an example from the tutorial to import `Annotated` from `typing` instead of `typing_extensions` for Python 3.9+. `typing.Annotated` was made available since 3.9.

### DIFF
--- a/docs_src/dependencies/tutorial012_an_py39.py
+++ b/docs_src/dependencies/tutorial012_an_py39.py
@@ -1,5 +1,6 @@
-from fastapi import Depends, FastAPI, Header, HTTPException
 from typing import Annotated
+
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@rockokw, thanks for the interest in FastAPI!

Although it's true that we can import `Annotated` from `typing` in Python 3.9, it's not a mistake.
Also, there are many more code examples where `Annotated` or `Literal` are imported from `typing_extensions`. If we decide to change it, we need to do it consistently across entire codebase.
But we have a lot of PRs to review, some of them are quite important fixes.

Considering all this, I think this is not something we should spend our efforts on now.

I'll close this PR.
Thanks again!

rockokw:
Understood, thanks for the feedback

PR #13713, opened 2025-05-13
✨ Add OpenAPI `external_docs` parameter to `FastAPI`

cmtoro:
This PR adds an external_docs parameter to the __init__ method of the FastAPI class. Currently, the external_docs parameter is not being passed when generating the OpenAPI documentation, which I believe is an oversight. By including this parameter in the FastAPI class constructor, the external_docs field will be correctly included in the root of the generated OpenAPI JSON or YAML file.

Reasoning:

The external_docs field provides valuable external documentation links and should be part of the OpenAPI metadata. Including it at the root level of the OpenAPI spec enhances the usability and completeness of the generated documentation, especially for providing references to external resources.


fastapi/applications.py
```diff
@@ -810,6 +810,47 @@ class Item(BaseModel):
                 """
             ),
         ] = True,
+        openapi_external_docs: Annotated[
+            Optional[Dict[str, Any]],
+            Doc(
+                """
+                This field allows you to provide additional external documentation links
+                for an API operation or component. It’s an optional field and can be added
+                to an operation or a tag. If provided, it must be a dictionary containing:
```
```suggestion
                This field allows you to provide additional external documentation links.
                If provided, it must be a dictionary containing:
```

fastapi/applications.py
```diff
@@ -810,6 +810,47 @@ class Item(BaseModel):
...
+
+                * `description`: A brief description of the external documentation, which may use
+                [CommonMark syntax](https://commonmark.org/) for rich text formatting.
```
YuriiMotov: I think we need to make sure it works (CommonMark syntax for rich text formatting)

fastapi/applications.py
```diff
@@ -810,6 +810,47 @@ class Item(BaseModel):
...
+                from fastapi.openapi.models import Doc
+                from typing_extensions import Annotated
+
+                external_docs: Annotated[Optional[Dict[str, Any]], Doc()] = {
+                    "description": "Detailed API Reference",
+                    "url": "https://example.com/api-docs",
+                }
+
+                app = FastAPI(openapi_external_docs=external_docs)
+
```
```suggestion
                from fastapi import FastAPI

                external_docs = {
                    "description": "Detailed API Reference",
                    "url": "https://example.com/api-docs",
                }

                app = FastAPI(openapi_external_docs=external_docs)
```
No need to specify type explicitly here

fastapi/applications.py
```diff
@@ -810,6 +810,47 @@ class Item(BaseModel):
...
+                ```
+
+                ### Explanation:
+                - `description`: Provides a short description explaining what this external documentation covers.
+                - `url`: Points to the full external documentation.
+
+                **Notes**:
+                - This field is **optional**; if not included, no external documentation will be linked.
+                - You can use this feature to link to full API documentation, tutorials, or any other
```
```suggestion
```
I think we don't need this. The above explanation should be enough

tests/test_openapi_examples.py
```diff
@@ -5,7 +5,12 @@
 from fastapi.testclient import TestClient
 from pydantic import BaseModel
 
-app = FastAPI()
+external_docs = {
+    "description": "External API documentation.",
+    "url": "https://docs.example.com/api-general",
+}
+
+app = FastAPI(openapi_external_docs=external_docs)
```
YuriiMotov: Not sure this is right place for this. Maybe add this to `tests/main.py` and test in `tests/test_application.py`?

YuriiMotov (changes requested):
@cmtoro, thanks for your interest and efforts!

I think that description should be simplified.
Please review my suggestions

tests/test_openapi_examples.py
cmtoro: Hi Yurii, thanks a lot for your comments—I really appreciate you taking the time to review this!

Could you help me understand a bit more why you think this might not be the right place for the example? It seemed like a good fit to include it here, since the file is already dedicated to OpenAPI scenarios and includes examples that demonstrate the various OpenAPI-related options and attributes.

YuriiMotov: Hi @cmtoro! You are welcome!

`test_openapi_examples.py` is intended to test `examples` and `openapi_examples` parameters. Look at tests that are already there and you will see that all of them are about these parameters.

`tests/main.py` includes a lot of different use-cases and `tests/test_application.py` tests openapi schema of app from `tests/main.py`.

YuriiMotov (approved):
LGTM

Not sure we need this parameter, but if we want to add it, then this PR effectively does it.

@cmtoro, thanks!

tiangolo (approved):
Looks good, thank you! 🚀

PR #13714, opened 2025-05-14
🐛 Prevent scroll-to-top on restart/fast buttons in `termynal.js`

Ashish-Pandey62:
**Issue**: Clicking the "restart" or "fast" buttons in the terminal widget causes the page to scroll to the top due to the href="#" attribute, which disrupts the user experience by interrupting their reading flow.

**Fix**: Changed href="#" to href="javascript:void(0)" in the generateRestart and generateFinish methods of termynal.js. This prevents the scroll behavior while keeping the UI and functionality intact.

**Testing**: Tested locally on the FastAPI docs page. The buttons no longer cause the page to scroll to the top, and their functionality (restarting or speeding up the terminal animation) remains unchanged.


YuriiMotov (approved):
It works! (I mean, this PR fixes the [problem](https://github.com/fastapi/fastapi/discussions/12420))

Ashish-Pandey62:
@YuriiMotov Just wanted to check if this is ready to be merged?

YuriiMotov:
Looked at this PR one more time.
@Ashish-Pandey62, ~~could you please explain why you removed indentations and added `;`?~~
~~I think we should avoid adding unnecessary changes. Especially if it makes code less readable (this is about removing the indentations).~~

Answering your question. I checked that it works, but I think this PR should be reviewed by person who is more familiar with JS than me.
@alejsdev, could you please take a look when you have time?

Ashish-Pandey62:
@YuriiMotov I've fixed the indentation issues. Regarding the semicolons, I added them as it's generally recommended for consistency and to prevent potential issues, even though they're optional in JS.

YuriiMotov:
> @YuriiMotov I've fixed the indentation issues. Regarding the semicolons, I added them as it's generally recommended for consistency and to prevent potential issues, even though they're optional in JS.

My bad I didn't explain it clearly. My idea was that we should minimize the diff (changes) so that it would be easier for reviewers to see what was changed exactly.
Now we have 20 updated lines of code, some (most?) of these changes are just adding `;` or moving lines up or down.
But if you look at the diff that GitHub shows, it's not obvious what changes you made in each of those lines. So, reviewers will have to compare line by line.
If you made only necessary changes, it would take a little bit less time and efforts to review it. Thus this PR would have more chances to be reviewed and merged soon.

Ashish-Pandey62:
@YuriiMotov Thank you for the feedback! I've updated the PR with minimal changes as suggested - now only modifying the href attributes in both functions while keeping all other code intact.

The changes are now focused solely on:

    Changing href="#" to href="javascript:void(0)" in generateRestart() and generateFinish()

    Maintaining all existing formatting and structure

This should make the diff much cleaner for review while still fully resolving the scrolling issue. Please let me know if there's anything else that needs adjustment!

YuriiMotov (approved):
LGTM

To check:
* Go to https://9f494309.fastapitiangolo.pages.dev/#installation
* Wait until the bash example has finished playing and the "restart" button appears
* Press "restart" button and notice it doesn't scroll to the top
* In [current docs](https://fastapi.tiangolo.com/#installation) pressing "restart" causes page scroll to the top

tiangolo (approved):
Nice, thank you! 🚀 

And thanks a lot @YuriiMotov for the review, the detailed step by step to confirm the issue and the fix is super useful! :raised_hands:

PR #13729, opened 2025-05-22
🐛 Fix response status code which were misconfigured to show 403

AditSoni:
Replaced 403 status code with 401 ; as 403 represents correct authentication but incorrect authorization. 

And the cases here are about missing creds in auth headers so 401 makes more sense. 

The response body is displaying the correct message but the response status code was 403.

### DIFF
--- a/fastapi/security/api_key.py
+++ b/fastapi/security/api_key.py
@@ -4,7 +4,7 @@ from fastapi.openapi.models import APIKey, APIKeyIn
 from fastapi.security.base import SecurityBase
 from starlette.exceptions import HTTPException
 from starlette.requests import Request
-from starlette.status import HTTP_403_FORBIDDEN
+from starlette.status import HTTP_401_UNAUTHORIZED
 from typing_extensions import Annotated, Doc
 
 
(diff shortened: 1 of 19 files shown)


mkbhanda:
Looks good. 

A better practice is to use constants everywhere exceptions are created and asserts like HTTP_401_UNAUTHORIZED and INVALID_AUTH_CREDENTIALS_STR to avoid errors from typos.

svlandeg:
Thanks for the contribution @AditSoni 🙏. I'm closing this one in favour of https://github.com/fastapi/fastapi/pull/13786, which is more extensive and should (hopefully) address this issue once and for all.

njsbhu:
Would be good if we can add docstring in check_api_key() for clarity, something like below -

def check_api_key(api_key: str) -> bool:
    """
    This Validate the provided API key against the expected value.
    Returns:
        bool: True if the key is valid, False otherwise.
    Notes:
        The "Not authenticated" message is used consistently across query, header, and cookie checks to make authentication failures clear and uniform.
    """

Issue #10370, opened 2023-10-02
Roadmap


tiangolo:
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

Kludex:
I've locked, so people subscribed to this issue only receive updates from the maintainers.

Kludex:
I've locked this issue so subscribers are only updated by @tiangolo.

iudeen:
This is lovely! 

A suggestion...will it be possible to have a list of PRs mapped with these roadmap items, so that we can check and contribute to those which are not started.

elsakeirouz:
Don't you plan on adding RefreshUrl to the OAuth2 Password Flow?

wangxin688:
Is there any plans to add class based view natively support by fastapi?

antonrh:
Small comment regarding `Errors based on RFC 7807`: it has already been obsoleted by RFC 9457 - https://www.rfc-editor.org/rfc/rfc9457.html.

afallou:
A note on the maintenance section.
> I have requested to still review each PR myself.

That is a major issue IMO. We've deployed a FastAPI app in production for one of our microservices, as an exploratory project to switch over from Flask. We started hitting issues like https://github.com/MagicStack/asyncpg/issues/978, which https://github.com/tiangolo/fastapi/pull/5707 is attempting to fix (see also related issue https://github.com/tiangolo/fastapi/discussions/8424).

I consider that issue be a major one, because it really touches on what makes async programming in Python hard and a little risky without a good framework on top – the errors you get are somewhat probabilistic, the stracktraces cryptic and the issues hard to debug. That PR is almost a year old, and there has been no activity of any kind on it from you, and that worries me. Because of that, my current conclusion from this exploratory project is that FastAPI isn't something I want to rely on in a production environment yet.

Your project is awesome and I love it, like many others. I have a lot of respect for your work here. But with 64k stars it's far beyond a "one-man shop"  OSS project, and having your review every PR isn't realistic anymore. Maybe you could start with a 2nd core reviewer, and see how that works?

PR #13733, opened 2025-05-24
📝 Add warning for deprecated methods in sql-databases.md

crummyh:
The `app.on_event` method has been deprecated, so I added a warning to the docs.

### DIFF
--- a/docs/en/docs/tutorial/sql-databases.md
+++ b/docs/en/docs/tutorial/sql-databases.md
@@ -105,6 +105,12 @@ We will create the database tables when the application starts.
 
 {* ../../docs_src/sql_databases/tutorial001_an_py310.py ln[32:37] hl[35:37] *}
 
+/// warning
+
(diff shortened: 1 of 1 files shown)


YuriiMotov (commented):
@crummyh, thank you for your interest and efforts!

I think this warning might be misleading for readers. `create_db_and_tables` function is synchronous, but `lifespan` is `async`. So, just replacing `on_startup` with `lifespan` would be bad practice (calling `sync` blocking function inside `async` coroutine). Then we would need to explain here that this function should be called with `run_in_threadpool` inside `lifespan`.

There is a "Updated SQL tutorials (sync and async) based on SQLModel." task in FastAPI [Roadmap](https://github.com/fastapi/fastapi/issues/10370). I believe this question along with many others will be addressed there. 

I suggest we leave it as it is now and wait for this whole section to be updated (I hope with `async` example).

crummyh:
I completely agree with you, and I'll take a look at that task.

PR #13755, opened 2025-06-01
docs: add use_cache example

varma887:
Added a usage example showing how to use the `use_cache` parameter in FastAPI route decorators.


YuriiMotov:
Sorry, but there is no `use_cache` parameter of route decorator.

PR #13756, opened 2025-06-02
⚡️ Parallelize dependency resolution

archfz:
This resolves a serious performance issue in fastapi: async dependencies resolving sequentially. #639

I have attempted to pick up this previous work https://github.com/fastapi/fastapi/pull/3902 but I thought it was too complicated. So I am giving it a more simpler localized approach. Two things to mention:

1. Since this approach breaks when contextvars are used with async generators, subtrees of the dependency graph are excluded from parallelization. This would be a limitation with this solution.
2. For cached dependencies Future is used so await can be called multiple times on it.

Testing wise I'd need suggestions. Not sure how to test this and I also expect that many tests already cover the changes. Asserting that things are done in parallel would be beneficial but I'd need help with that.

EDIT:

Added in the meantime an option, just like cache, on the dependency definition called parallelizable. This is by default opt-out style, meaning it is True. Maybe to prevent breaking changes we can ship this with opt-in style. Let community experiment with it and in time if no issues are creeping up it can be considered to switch to opt-out style with a major release. For Security however I am thinking that maybe we should have it always False.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -281,7 +281,7 @@ def get_dependant(
     name: Optional[str] = None,
     security_scopes: Optional[List[str]] = None,
     use_cache: bool = True,
-    parallelizable: bool = True,
+    parallelizable: Optional[bool] = None,
 ) -> Dependant:
     path_param_names = get_path_param_names(path)
     endpoint_signature = get_typed_signature(call)
@@ -589,13 +589,19 @@ class DependencySolveException(Exception):
         self.errors = errors
 
 
-def is_context_sensitive(dependant: Dependant) -> bool:
-    if dependant.parallelizable is False or (
+def is_context_sensitive(dependant: Dependant, *, default_parallelizable: bool) -> bool:
+    effective_parallelizable = (
+        default_parallelizable if dependant.parallelizable is None else dependant.parallelizable
+    )
+    if effective_parallelizable is False or (
         dependant.call is not None
         and (is_gen_callable(dependant.call) or is_async_gen_callable(dependant.call))
     ):
         return True
-    return any(is_context_sensitive(sub) for sub in dependant.dependencies)
+    return any(
+        is_context_sensitive(sub, default_parallelizable=default_parallelizable)
+        for sub in dependant.dependencies
+    )
 
 
 def is_context_with_background_task(dependant: Dependant) -> bool:
@@ -659,6 +665,7 @@ async def solve_dependencies(
                 call=call,
                 name=sub_dependant.name,
                 security_scopes=sub_dependant.security_scopes,
+                parallelizable=sub_dependant.parallelizable,
             )
 
         def resolve_cached() -> Optional["Future[Any]"]:
@@ -750,10 +757,16 @@ async def solve_dependencies(
             if name is not None:
                 values[name] = value
 
+
+    app = request.app  # type: ignore[attr-defined]
+    default_parallelizable = bool(
+        getattr(app, "depends_default_parallelizable", False)
+    )
+
     sequential_deps = []
     parallel_deps = []
     for sub in dependant.dependencies:
-        if is_context_sensitive(sub):
+        if is_context_sensitive(sub, default_parallelizable=default_parallelizable):
             sequential_deps.append(sub)
         else:
             parallel_deps.append(sub)
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -1,5 +1,7 @@
+import asyncio
 import inspect
 import sys
+from asyncio import Future
 from contextlib import AsyncExitStack, contextmanager
 from copy import copy, deepcopy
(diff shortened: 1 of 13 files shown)


tests/test_dependency_cache.py
```diff
@@ -29,7 +29,7 @@ async def get_sub_counter(
 
 @app.get("/sub-counter-no-cache/")
 async def get_sub_counter_no_cache(
-    subcount: int = Depends(super_dep),
+    subcount: int = Depends(dep_counter),
```
archfz: Here I had to change this to the topmost dep, because one level lower already resolves just a tiny bit later and in that case this test is not valid anymore.

archfz:
@tiangolo Can you please cancel the actions that were triggered by this PR? They are blocking the whole runners as I had a bug that blocked tests completely. Cancel all but last. The first run is still running since 2+ hours :/ I was confused why the tests are not running on later pushes, so I ended up pushing more as I though I made something incorrectly with commits.

archfz:
Tested this in a medium side production backend with fastapi setup. Found several issues, more notably around caching and exception handling. Fixed these, however these warrant for additional test cases.

fastapi/dependencies/utils.py
```diff
@@ -612,36 +651,112 @@ async def solve_dependencies(
...
+            else:
+                dependency_cache[cache_key].set_result(value)
+
+        try:
+            resolved = await task
+        except DependencySolveException as exc:
+            ensure_cache(None)
+            return sub_dependant.name, None, exc.errors, None
+        except BaseException as resolution_exc:
+            ensure_cache(resolution_exc)
```
archfz, line 741: Tests case needed.

fastapi/dependencies/utils.py
```diff
@@ -612,36 +651,112 @@ async def solve_dependencies(
...
+    if parallel_deps:
+        parallel_results = await asyncio.gather(
+            *[resolve_sub_dependant(sub) for sub in parallel_deps]
+        )
+        results.extend(parallel_results)
+
+    for name, result, sub_errors, sub_exception in results:
+        # Ensures order of exception based on dependency order.
+        if sub_exception:
+            raise sub_exception
```
archfz: Tests case needed.

YuriiMotov:
@archfz, thanks for your interest and efforts!

Could you please tell me what the current state of this PR is and what are your future plans on it?
Do you have any open questions?
If this is a work in progress, how about converting this into a Draft PR?

zoltanfodor-archfz:
@YuriiMotov My plan is to complete it, get it merged and released. This is already used in production, through a fork of fastapi. I need however feedback from the maintainers / or atleast from the community, code review. I will continue aligning, adding necessary tests only after that.

archfz:
Still no reviews. @tiangolo please make it move.

archfz:
@tiangolo @YuriiMotov I have added:

1. A global app config to enable the default of dependency parallelization. It is Opt-In now, default False, which means existing apps should not be affected if any missing cases are missed in testing.
2. I have added extensive testing.
3. I have added documentation about this option.

Please review and get it moving.

archfz:
@tiangolo @YuriiMotov Can you guys please review. I will fix conflicts once this gets attention, cause like this there is no point, PR is open 6months+.

archfz:
@tiangolo @YuriiMotov Bump. We have been using this in production now for 9 months. Would be nice to get it in.

YuriiMotov:
@archfz, we are waiting for Sebastian to take a look and validate the idea of this PR.

> I will fix conflicts once this gets attention, cause like this there is no point, PR is open 6months+.

You are right - we should wait for idea validation first

Mohamed-elgypaly:
As a Senior Architect focused on high-performance API design, I find the shift towards parallel dependency resolution in #13756 critical for reducing TTB (Time To Benefit) in complex I/O-bound graphs. However, from a production stability standpoint, three areas require further architectural hardening:

1. **ContextVar Safety**: The current reliance on `asyncio.gather` can be precarious for sub-dependencies that utilize `contextvars` for transaction tracing or tenancy isolation. I recommend a wrapper that explicitly validates context propagation or a 'conservative' execution mode that identifies `contextvar`-sensitive nodes during graph construction.
2. **Security Sequencing**: Security dependencies (e.g., OAuth2 scopes, RBAC checks) should ideally remain in a 'sequential-only' tier. Mixing security validation with general business logic dependencies in a parallel pool could lead to subtle race conditions if a security check relies on a state set by a previous check.
3. **Resource Saturation**: While parallelization improves latency, it increases memory pressure by spawning multiple coroutines simultaneously per request. A 'Concurrency Governor' (perhaps a semaphore-based limit) would prevent resource exhaustion in environments with high request volume and deep dependency trees.

This PR is a major step forward, but ensuring these 'safety rails' are robust will be key to its adoption in enterprise-scale FastAPI applications.

tiangolo:

Thank you! 🍰

Issue #976, opened 2020-02-12
[FEATURE] APIKeyHeader for Websockets


Catastropha:
### Is your feature request related to a problem

Is your feature request related to a problem?

Yes. It would be nice to use Security with APIKeyHeader (maybe APIKeyCookie, APIKeyQuery as well) for websocket connections.

### The solution you would like

Consistent with the current http way of doing it

### Describe alternatives you've considered

At the moment our websocket endpoints are protected with a simple api_key: str = Header(None),
and then we do
if not api_key:
        await websocket.close(code=WS_1008_POLICY_VIOLATION)
        return None

If someone can give me some guiding on how to structure this feature I might give it a try and do a PR

tiangolo:
Yes, you can, it's documented here: https://fastapi.tiangolo.com/advanced/websockets/#using-depends-and-others

Also, have in mind that browsers don't have a way to set headers when creating WebSocket connections: https://stackoverflow.com/a/4361358/219530

Catastropha:
You are right!

Out of curiosity, (the project is mobile only) I followed the docs and here is what I have:

`router = APIRouter()
@router.websocket("/ws/")
async def ws_endpoint(
        websocket: WebSocket,
        origin: dict = Depends(get_ws_origin),
        ):
    await websocket.accept()
`

`async def get_ws_origin(
        websocket: WebSocket,
        api_key: str = Security(APIKeyHeader(name=API_KEY_NAME)),
        ):
`
Is this the correct way to do websockets with FastAPI?

I get this error:

`Traceback (most recent call last):
  File "/env/lib/python3.6/site-packages/uvicorn/protocols/websockets/websockets_impl.py", line 153, in run_asgi
    result = await self.app(self.scope, self.asgi_receive, self.asgi_send)
  File "/env/lib/python3.6/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/env/lib/python3.6/site-packages/fastapi/applications.py", line 142, in __call__
    await super().__call__(scope, receive, send)  # pragma: no cover
  File "/env/lib/python3.6/site-packages/starlette/applications.py", line 134, in __call__
    await self.error_middleware(scope, receive, send)
  File "/env/lib/python3.6/site-packages/starlette/middleware/errors.py", line 143, in __call__
    await self.app(scope, receive, send)
[...]

PR #13758, opened 2025-06-03
✨ Fix `APIKeyQuery` for WebSocket endpoints

Harry-Lees:
Currently, the Security classes outlined in the documentation do not work for WebSocket endpoints. This has raised several questions on GitHub discussions/ issues.

* https://github.com/fastapi/fastapi/discussions/8503
* https://github.com/fastapi/fastapi/discussions/9132
* https://github.com/tiangolo/fastapi/issues/2587
* https://github.com/fastapi/fastapi/issues/976

There have been several PRs attempting to fix this issue, but it was always deemed not possible due to the WebSocket exception handling. I think this has changed with several recent updates to Starlette and FastAPI.

This uses the same theory behind the example in the Starlette [docs](https://www.starlette.io/websockets/) on WebSockets.

### DIFF
--- a/fastapi/security/api_key.py
+++ b/fastapi/security/api_key.py
@@ -3,7 +3,7 @@ from typing import Optional
 from fastapi.openapi.models import APIKey, APIKeyIn
 from fastapi.security.base import SecurityBase
 from starlette.exceptions import HTTPException
-from starlette.requests import Request
+from starlette.requests import HTTPConnection
 from starlette.status import HTTP_403_FORBIDDEN
 from typing_extensions import Annotated, Doc
 
@@ -107,7 +107,7 @@ class APIKeyQuery(APIKeyBase):
         self.scheme_name = scheme_name or self.__class__.__name__
         self.auto_error = auto_error
 
-    async def __call__(self, request: Request) -> Optional[str]:
+    async def __call__(self, request: HTTPConnection) -> Optional[str]:
         api_key = request.query_params.get(self.model.name)
         return self.check_api_key(api_key, self.auto_error)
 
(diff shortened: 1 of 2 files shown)


BitWeaverDev:
I ran this branch locally against current master. The core fix is correct: `HTTPConnection` is the right shared base for `Request` and `WebSocket`, and FastAPI's dependency resolver already special cases it in `fastapi/dependencies/utils.py`. All 3 tests in this PR pass, and I ran all 136 existing `tests/test_security*.py` tests with no regressions. I also wrote extra tests for `APIKeyQuery` and `APIKeyHeader` over WebSocket, since only `APIKeyCookie` is covered here, and both pass, so the fix genuinely works for all three classes.

A few things I think need attention before this can merge.

1. This is unmergeable against current master (shows as CONFLICTING). Master has since modernized this file (`Optional[str]` became `str | None`, `typing_extensions.Doc` became `annotated_doc.Doc`, and #13786 changed the missing key status from 403 to 401). The underlying problem this PR fixes is still present on master today, so it just needs a rebase.

2. Test coverage only covers `APIKeyCookie`. `APIKeyQuery` and `APIKeyHeader` are changed too but have no tests. Worth adding those so this doesn't rely on a manual check like the one I did.

3. The title says "Fix `APIKeyQuery`" but the diff touches all three API key classes. Might be worth updating for clarity.

4. The issues and discussions linked in the description ask about Security classes on WebSocket in general, not just API keys. `fastapi/security/http.py` (HTTPBasic, HTTPBearer, HTTPDigest) and `fastapi/security/oauth2.py` (OAuth2PasswordBearer, SecurityScopes) still hard type `request: Request` and remain broken for WebSocket. This PR closes part of the gap those issues describe, so it might be worth either extending the fix to those or being explicit in the description that this is scoped to API key auth only.

5. No docs update. Nothing in `docs/en/docs/advanced/websockets.md` or `docs/en/docs/advanced/security/` mentions that Security dependencies now work on WebSocket routes.

Overall the change itself is small, correct, and low risk for what it covers. I think it just needs a rebase, tests for the two untested classes, and either a scope note or a docs mention before it is ready to merge.

PR #13760, opened 2025-06-03
✨ Add support for BSON response

WrldEngine:
# BSON Response  Feature

Discussion: https://github.com/fastapi/fastapi/discussions/13768

## Summary
Implemented a custom Response class for supporting BSON as a response serialization format in FastAPI. This enables seamless integration for applications using BSON instead of JSON, commonly used in MongoDB-based services or performance-critical APIs.

### Additionally
1. updated .gitignore to include missing .lock files from various package managers.
2. wrote unit tests [tests/test_bson_response_class.py](https://github.com/fastapi/fastapi/pull/13760/commits/308c762d6ebf68cfe356755d759fa0589508227d#diff-809ec7275fbdb456d14feb9a84ef7aa5c4d752e5f3bf8adaafbc0f14bd6f30ba)

## Motivation
BSON is a binary-encoded serialization of JSON-like documents, designed to be more efficient in both size and speed for certain use cases. FastAPI currently does not provide built-in support for returning responses in BSON format. This addition provides a clean and reusable way to return application/bson responses.

## Next Steps
A Rust wrapper is planned soon. The existing BSON utility library (currently in Python) will be rewritten in Rust to achieve improved performance and tighter memory control, especially important for high-throughput applications. The Rust implementation will be exposed via bindings for Python, allowing FastAPI to use this enhanced BSON handling behind the scenes.


YuriiMotov:
Thanks for your interest in FastAPI!

To increase the chances of this PR being merged, please avoid mixing multiple changes in a single PR. It's best to follow the principle: one PR = one fix/feature.

Additionally, it would be helpful to include some context - for example, why you believe this change is important, and any related links to discussions or issues.

I couldn't find any existing discussions about introducing a BSONResponse class. It might be a good idea to start a discussion first to gather feedback from the community.

WrldEngine:
> Thanks for your interest in FastAPI!
> 
> To increase the chances of this PR being merged, please avoid mixing multiple changes in a single PR. It's best to follow the principle: one PR = one fix/feature.
> 
> Additionally, it would be helpful to include some context - for example, why you believe this change is important, and any related links to discussions or issues.
> 
> I couldn't find any existing discussions about introducing a BSONResponse class. It might be a good idea to start a discussion first to gather feedback from the community.

Thank you for reply, so, am i right?
I need to:
1. Describe more about this feature in comments
2. Start disscussion and collect positive feedbacks from the community

YuriiMotov:
> 1. Describe more about this feature in comments

Better edit initial description

> 2. Start disscussion and collect positive feedbacks from the community

Yes

Also, exclude and don't include any unrelated things. For example, editing `.gitignore` looks not necessary for this PR.

WrldEngine:
> > 1. Describe more about this feature in comments
> 
> Better edit initial description
> 
> > 2. Start disscussion and collect positive feedbacks from the community
> 
> Yes
> 
> Also, exclude and don't include any unrelated things. For example, editing `.gitignore` looks not necessary for this PR.

Thank you, but can you said when you or other contributors can label my pr?
When disscussion will be positive?

wbarnha:
I would appreciate support for BSON in FastAPI! I think having https://github.com/mongodb-labs/python-bsonjs as an optional dependency for faster BSON parsing would be nice to include in this PR, at least until the Rust implementation is ready.

WrldEngine:
> I would appreciate support for BSON in FastAPI! I think having https://github.com/mongodb-labs/python-bsonjs as an optional dependency for faster BSON parsing would be nice to include in this PR, at least until the Rust implementation is ready.

Okay, no problem, but i think i want to make some benchmarks to compare bson & bsonjs with _serialisation_/deserialisation . As i was looking, this lib using c++ backend, so im sure about benchmark results

pyproject.toml
```diff
@@ -46,6 +46,7 @@ dependencies = [
     "starlette>=0.40.0,<0.47.0",
     "pydantic>=1.7.4,!=1.8,!=1.8.1,!=2.0.0,!=2.0.1,!=2.1.0,<3.0.0",
     "typing-extensions>=4.8.0",
+    "python-bsonjs>=0.5.0",
```
dolfinus: It's definitely should not be added to core framework dependencies

WrldEngine: No problem, i will edit, but, if it so, ujson/orjson also shouldn't be in core dependencies

dolfinus: > ujson/orjson also shouldn't be in core dependencies

They are a part of extras `pip install "fastapi[all]"`, not a part of core `pip install fastapi`

WrldEngine: Now i can see, sorry, my bad, i already added this to fastapi[all] but forgot to remove from core deps, because of pdm dependency management. I will edit

WrldEngine:
@dolfinus , I accidentally requested a review. Dear contributor, if you're not responsible for merges, please feel free to ignore this. The review above isn't meaningful — it's just a simple mistake caused by inattention and PDM's poor dependency management.

WrldEngine:
@YuriiMotov hi, how many likes should be in discussion for pr to merge to happen? Tests and other checks already passed successfully

YuriiMotov:
> @YuriiMotov hi, how many likes should be in discussion for pr to merge to happen? Tests and other checks already passed successfully

Please be patient!
When time comes, we will take a look at that discussion and see whether this feature is supported by community or not.

We appreciate your interest in FastAPI, but we have a lot of PRs we need to review.
If you want to speed up this process, you can help us by [reviewing PRs](https://fastapi.tiangolo.com/help-fastapi/#review-pull-requests), this is very much appreciated!

fastapi/responses.py
```diff
@@ -46,3 +59,35 @@ def render(self, content: Any) -> bytes:
         return orjson.dumps(
             content, option=orjson.OPT_NON_STR_KEYS | orjson.OPT_SERIALIZE_NUMPY
         )
+
+
+class BSONResponse(Response):
+    """
+    BSON response using the current Python bson library for data serialization.
+
+    Note: This is a temporary solution. Soon, a custom Rust wrapper will be implemented for BSON serialization/deserialization to improve performance and reliability.
```
dolfinus, line 68: Why this temporary solution should become a part of the framework? Why not make a dedicated `fastapi-bson` package which can be installed and imported by user if needed?

WrldEngine: I think it's best that I delete that part in the comment. But again, back to the old topic of dependencies. According to your thinking, I should remove one of the two dependencies related to bson, but then look at orjson and ujson, in 99% of cases everyone would choose orjson because ujson has no superior features compared to it. That's why I decided to add two different dependency options too. Bson is not Json.

Otherwise it would be better to open a discussion on removing the ujson response.

tiangolo:
Thanks for the interest. I think this should be done with a custom response class if needed: https://fastapi.tiangolo.com/advanced/custom-response/#custom-response-class

I'll pass on this, thanks! :coffee:

Issue #3225, opened 2021-05-14
Query parameter descriptions not picked up from docstring?


flashgorman-chorus:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to one of:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Description

There is a fair amount of talk in FastAPI docs about picking up API descriptions from the docstring.  I expected that would include descriptions of query parameter args nested underneath the docstring's `Args:` block but that does not appear to be the case.

Since the sole purpose of the `Args:` block is to provide descriptions of args, it seemed quite natural to me that FastAPI would parse this to auto-generate descriptions for each query parameter found so I just wanted to confirm that this is, indeed, the expected behavior:

> While FastAPI will pick up your API's description from your docstring, FastAPI will ignore all query parameter descriptions found in the docstring `Args:` block.

Can anyone here confirm or deny?

Hultner:
I can confirm that it does not pick up the description from the docstrings.
You can use the description field instead like this:

```python
@app.get("/xyz")
def xyz(
   *,
   param: str = Query(..., description="This will show up in OpenAPI spec",
): ...
```

flashgorman-chorus:
Got it.  Probably a good enhancement request, for reasons cited above, but will close this issue as the question has been answered.  Thank you!

tiangolo:
Thanks for the help here @Hultner ! 👏  🙇 

Thanks for reporting back and closing the issue @flashgorman-chorus  👍

Yep, there are many docstring formats, each with its own micro syntax and none is a "standard", so there's no straightforward way to implement it in a way that is generic enough.

[...]

PR #13767, opened 2025-06-04
✨ Extract descriptions of parameters and responses from docstring

JustKiddingCode:
This pull request implements parsing docstrings via griffe and sets the relevant openapi descriptions to the corresponding docstring part.

The openapi has 3 description fields: the route itself, each parameter and the response.
With docstrings, you can only set one: The route itself. You can set the description for parameter and response using attributes in the router or annotations.
If I describe the function with all arguments, e.g. to build a documentation via mkdocs, the arguments that get autofilled by depends will appear in the openapi spec. If I want the description in the openapi spec and in the docstring there will be redundancy. Doing it only in the fastapi part, it will not be part of mkdocs (or similar).
As described by #3225, using docstrings is complicated, because the structure is not standardized. So I solve this using griffe, which can parse the most common formats (google, numpy, sphinx). Currently i set it to google format, but the plan is to make this configurable. However, if the description is set by fastapi attributes, this will take precedence.

```python
@router.get(
    "/events/{eventId}",
    status_code=status.HTTP_200_OK,
    summary="Get event details",
)
@alog
async def get_event_details(
    auth: Annotated[Auth, Depends(authorize(AuthStrategy.HYBRID))],
    db: Annotated[AsyncSession, Depends(get_db)],
    event_id: Annotated[int | uuid.UUID, Path(alias="eventId")],
) -> AddEditGetEventResponse:
    """Retrieve details of a specific event either by its event ID, or via its UUID.

    args:
        auth: the user's authentification status
        db: the current database
        event_id: the ID or UUID of the event

    returns:
        the event details
    """
    return await _get_event_details(db, event_id, auth.user)
```

will be converted to:

image

Before i continue to work on this, i want to know if adding a dependency is ok. If so, i would allow to set the parse format of the docstring (griffe supports google, numpy and sphinx).

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -141,7 +141,7 @@ class FastAPI(Starlette):
             ),
         ] = None,
         description: Annotated[
-            str,
+            Optional[str],
             Doc(
                 '''
                 A description of the API. Supports Markdown (using
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -1064,7 +1064,7 @@ class FastAPI(Starlette):
         dependencies: Optional[Sequence[Depends]] = None,
         summary: Optional[str] = None,
         description: Optional[str] = None,
-        response_description: str = "Successful Response",
+        response_description: Optional[str] = None,
         responses: Optional[Dict[Union[int, str], Dict[str, Any]]] = None,
         deprecated: Optional[bool] = None,
         methods: Optional[List[str]] = None,
@@ -1122,7 +1122,7 @@ class FastAPI(Starlette):
         dependencies: Optional[Sequence[Depends]] = None,
         summary: Optional[str] = None,
         description: Optional[str] = None,
-        response_description: str = "Successful Response",
+        response_description: Optional[str] = None,
         responses: Optional[Dict[Union[int, str], Dict[str, Any]]] = None,
         deprecated: Optional[bool] = None,
         methods: Optional[List[str]] = None,
@@ -1574,7 +1574,7 @@ class FastAPI(Starlette):
             ),
         ] = None,
         response_description: Annotated[
-            str,
+            Optional[str],
             Doc(
                 """
                 The description for the default response.
@@ -1582,7 +1582,7 @@ class FastAPI(Starlette):
                 It will be added to the generated OpenAPI (e.g. visible at `/docs`).
                 """
             ),
-        ] = "Successful Response",
+        ] = None,
         responses: Annotated[
             Optional[Dict[Union[int, str], Dict[str, Any]]],
             Doc(
@@ -1947,7 +1947,7 @@ class FastAPI(Starlette):
             ),
         ] = None,
         response_description: Annotated[
-            str,
+            Optional[str],
             Doc(
                 """
                 The description for the default response.
@@ -1955,7 +1955,7 @@ class FastAPI(Starlette):
                 It will be added to the generated OpenAPI (e.g. visible at `/docs`).
                 """
             ),
-        ] = "Successful Response",
+        ] = None,
         responses: Annotated[
             Optional[Dict[Union[int, str], Dict[str, Any]]],
             Doc(
@@ -2325,7 +2325,7 @@ class FastAPI(Starlette):
             ),
         ] = None,
         response_description: Annotated[
-            str,
+            Optional[str],
             Doc(
                 """
                 The description for the default response.
@@ -2333,7 +2333,7 @@ class FastAPI(Starlette):
                 It will be added to the generated OpenAPI (e.g. visible at `/docs`).
                 """
             ),
-        ] = "Successful Response",
+        ] = None,
         responses: Annotated[
             Optional[Dict[Union[int, str], Dict[str, Any]]],
             Doc(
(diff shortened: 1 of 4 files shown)


YuriiMotov:
Could you please explain what problem this PR solves?

JustKiddingCode:
> Could you please explain what problem this PR solves?

Sure.
The openapi has 3 description fields: the route itself, each parameter and the response.
With docstrings, you can only set one: The route itself. You can set the description for parameter and response using attributes in the router or annotations.
If I describe the function with all arguments, e.g. to build a documentation via mkdocs, the arguments that get autofilled by depends will appear in the openapi spec. If I want the description in the openapi spec and in the docstring there will be redundancy. Doing it only in the fastapi part, it will not be part of mkdocs (or similar).
As described by #3225, using docstrings is complicated, because the structure is not standardized. So I solve this using griffe, which can parse the most common formats (google, numpy, sphinx). Currently i set it to google format, but the plan is to make this configurable. However, if the description is set by fastapi attributes, this will take precedence.

HenriBlacksmith:
Just a personal opinion, I am not a maintainer but as a FastAPI user:
- API documentation is already supported using OpenAPI, it does not make much sense to use mkdocs docs for API endpoints.
- At most the new dependency must be optional, FastAPI is a micro framework and by default it should not pull tons of dependencies.
- As for dependencies it looks this addition impacts the existing flow though I'll let more skilled reviewers checking that.

JustKiddingCode:
>     * API documentation is already supported using OpenAPI, it does not make much sense to use mkdocs docs for API endpoints.

Were using mkdocs with [Modern MISP](https://github.com/Modern-MISP/api). Documenting only the API would leave out integral parts of the software, doing documentation mixed (openapi for api, rest mkdocs) would be error-prone. Also if I'm using api functions outside the api, how should i document them?

>     * At most the new dependency must be optional, FastAPI is a micro framework and by default it should not pull tons of dependencies.

I agree with the later part, however, making it optional would require much additional code and i doubt the benefit would outweigh it.

>     * As for dependencies it looks this addition impacts the existing flow though I'll let more skilled reviewers checking that.

Could you elaborate that? I don't understand it.

YuriiMotov:
@JustKiddingCode, let's first wait for Sebastián to look at this PR and validate the idea.
Then (in case of idea is accepted) we will need to add more tests and describe this feature in docs.

nidhishgajjar:

## Review of PR #13767: Extract descriptions of parameters and responses from docstring

This is a useful feature — automatically generating OpenAPI descriptions from docstrings reduces duplication and keeps documentation closer to the code. However, there are several concerns worth addressing.

### Major concerns 🔴

1. **Adding `griffe` as a required dependency**: Griffe is a substantial library. Adding it as a required dependency for all FastAPI users is significant — even those who don't use docstring-based OpenAPI generation will pay the install cost. Consider making this an optional dependency (e.g., `pip install fastapi[docstring-openapi]`) and gracefully falling back when not available.

2. **No tests**: The PR modifies core OpenAPI generation logic but includes no tests. This is critical for:
   - Verifying Google-style docstrings produce correct OpenAPI schemas
   - Ensuring explicit `Field(description=...)` still takes priority
   - Testing edge cases (empty docstrings, malformed docstrings, non-Google style)
   - Verifying the `\f` form-feed behavior is preserved

### Design observations 💡

3. **Hardcoded Google-style parser**: 
   ```python
   Docstring(self.docstring, parser="google", parser_options={"warnings": False})
   ```
   Users with Sphinx-style or NumPy-style docstrings won't benefit. This should either be configurable or documented as a limitation.

4. **Subtle default change**: `response_description` changes from `"Successful Response"` (always present) to `None` (conditionally filled from docstring or fallback). This could cause empty `description` fields in OpenAPI schemas for routes without docstrings or `Returns:` sections.

5. **Form-feed handling change**: The `\f` truncation is now applied to explicitly-set `route.description` too:
   ```python
   operation["description"] = operation["description"].split("\f")[0].strip()
   ```
   Previously it was only applied to auto-extracted descriptions. If a user deliberately includes `\f` in their explicit description, this would now truncate it.

6. **Response description joining**: Multiple return entries are joined with commas:
   ```python
   ",".join(x.description for x in ret[0])
   ```
   This may produce odd formatting in OpenAPI docs. Consider `\n\n` or just taking the first entry.

### What looks good ✅

- Correct priority: explicit `Field(description=...)` > docstring description
- Clean separation: parsed docstring stored on route, used during OpenAPI generation
- The overall concept is valuable and would reduce documentation boilerplate

### Summary

The feature idea is solid, but the implementation needs work before merging. The main blockers are: (1) the required dependency addition should be optional, (2) tests are essential for OpenAPI schema generation changes, and (3) the hardcoded parser style limits usefulness. The subtle behavior changes around defaults and form-feed handling should also be verified.

**Assessment: request-changes** (dependency strategy, tests needed, parser configurability)

tiangolo:
Thanks for the interest here. I personally don't like how regular conventional docstrings work as it's a microsyntax in a string, so much I don't use that style in the internals of FastAPI. So I'll pass on this one, thanks! :coffee:

PR #13774, opened 2025-06-06
✏️ Remove redundant words in docs/zh/docs/python-types.md

cwandev:
Remove redundant words

### DIFF
--- a/docs/zh/docs/python-types.md
+++ b/docs/zh/docs/python-types.md
@@ -228,7 +228,7 @@ John Doe
 
 ## Pydantic 模型
 
-<a href="https://docs.pydantic.dev/" class="external-link" target="_blank">Pydantic</a> 是一个用来用来执行数据校验的 Python 库。
+<a href="https://docs.pydantic.dev/" class="external-link" target="_blank">Pydantic</a> 是一个用来执行数据校验的 Python 库。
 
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@CharleeWa, thanks, but let's keep it as it is since it's closer to English version.

If you believe we really need to make this change, I'l re-open this. But we will need to get confirmation from 2 native speakers

CharleeWa:
> @CharleeWa, thanks, but let's keep it as it is since it's closer to English version.
> 
> If you believe we really need to make this change, I'l re-open this. But we will need to get confirmation from 2 native speakers

Thanks! I'm a native Chinese speaker, and I can confirm the repetition is unnecessary and doesn't add meaning. Removing it makes the sentence clearer and more natural. Happy to clarify if needed.

cwandev:
> @CharleeWa, thanks, but let's keep it as it is since it's closer to English version.
> 
> If you believe we really need to make this change, I'l re-open this. But we will need to get confirmation from 2 native speakers

Thanks! I'm a native Chinese speaker, and I can confirm the repetition is unnecessary and doesn't add meaning. Removing it makes the sentence clearer and more natural. Happy to clarify if needed.

waketzheng (approved):

Zhongheng-Cheng (approved):

PR #13794, opened 2025-06-16
docs: add Dart OpenAPI generator to client list

sandeshkumatgi:
This is my first open source contribution.


YuriiMotov:
@sandeshkumatgi, thanks for the interest in FastAPI!

I think this document doesn't aim to list all possible generators. There are many of them.
So, there is no need to include Dart OpenAPI Generator.

I'll close this PR.
Thanks for understanding!

PR #13800, opened 2025-06-18
✏️ Fix grammar mistake in `docs/en/docs/advanced/response-directly.md`

NavesSapnis:


YuriiMotov (approved):
It was an obvious mistake. Double-checked with ChatGPT

PR #13803, opened 2025-06-18
🐛 Improve `jsonable_encoder` for other types of iterables

Lob26:
## What was changed
Updated the jsonable_encoder function to correctly handle any iterable type.

Reordered conditionals to ensure the iterable check occurs later, allowing appropriate handling to other type-specific branches (which previously caused failures).

## Why this change is necessary
The previous contribution at #30 did not (fully) serialize objects like `sqlalchemy.engine.Row`, which inherit from `collections.abc.Sequence` but behave differently than standard lists or tuples. This led to unexpected failures or incorrect encoding.

## Testing
Manually tested the updated function against a wide variety of data structures
Behavior appears consistent and correct across all tested inputs.

## Notes
While the function now gracefully handles many iterables, I'm still wary there may be edge cases I haven't considered. Further review and test cases are welcome.


fastapi/encoders.py
```diff
@@ -14,14 +14,26 @@
...
+    Union,
+)
 from uuid import UUID
 
 from fastapi.types import IncEx
 from pydantic import BaseModel
-from pydantic.color import Color
 from pydantic.networks import AnyUrl, NameEmail
 from pydantic.types import SecretBytes, SecretStr
+from pydantic_extra_types.color import Color
```
dolfinus: This package supports pydantic 2.x only

rhealakhotia:
Hi @Lob26,  I’m new to open source but would love to help with this.  
Could you clarify:  
- Which edge cases still need testing?  
- How to reproduce the SQLAlchemy row issue locally?  

I’ll try to write a test case and debug the CI failures. Thanks!

fastapi/encoders.py
dolfinus: Same for `Coordinate` type

svlandeg: Note that this is not an issue anymore, as we've dropped support for Pydantic v1 in the meantime.

YuriiMotov:
@Lob26, thanks for your interest and efforts!

Do you have plans to continue working on this PR?

Lob26:
> **[rhealakhotia](/rhealakhotia) **

Sorry for the (very) late response, yes, I'll appreciate more tests for this PR, previously I encountered errors logic-related, so after a few iterations I achieve my desired result for the encoding, but I'm open for more test cases

tiangolo:
Thanks for the interest here, I understand this adds a new parameter `named_tuple_as_dict` that would be used manually by anyone that wants to opt in to that behavior.

I would consider this out of scope for now, so I'll pass on this. Thanks! :coffee:

PR #13824, opened 2025-06-24
⚡️ Re-use result of `is_sequence_field` in `_get_multidict_value` to improve performance

yinziyan1206:
The function `is_sequence_field` in `_get_multidict_value` is called twice. And it affects about 15%~20% performance.

## Test Script
```python
@app.get("/test/{num:int}")
async def test(num: int) -> dict:
    return {"echo": num}

async def send(message):
    pass

async def receive() -> dict:
    return {}

scope = {
    "type": "http",
    "asgi": {"version": "3.0", "spec_version": "2.3"},
    "http_version": "1.1",
    "server": None,
    "client": None,
    "scheme": "http", 
    "root_path": "",
    "headers": [(b"token", b"1")],
    "state": {},
    "path": "/test",
    "raw_path": b"/test",
    "query_string": b"a=1&b=2",
    "method": "GET"
}

async def main():
    p = Profiler(async_mode="disabled")
       
    async with app.router.lifespan_context(app):
        with p:
            await asyncio.gather(*(app(scope | {"path": scope["path"] + f"/{i}"}, receive, send) for i in range(10000)))
        p.print(show_all=True)


asyncio.run(main())
```

## Before
```
2.666 app  fastapi\routing.py:239
├─ 1.936 solve_dependencies  fastapi\dependencies\utils.py:572
│  ├─ 1.059 request_params_to_args  fastapi\dependencies\utils.py:740
│  │  ├─ 0.807 _get_multidict_value  fastapi\dependencies\utils.py:716
│  │  │  ├─ 0.764 is_sequence_field  fastapi\_compat.py:243
```

## After
```
1.255 app  fastapi\routing.py:239
├─ 0.787 solve_dependencies  fastapi\dependencies\utils.py:572
│  ├─ 0.404 request_params_to_args  fastapi\dependencies\utils.py:741
│  ├─ 0.266 _get_multidict_value  fastapi\dependencies\utils.py:716
│  │  ├─ 0.243 is_sequence_field  fastapi\_compat.py:243
```


YuriiMotov (approved):
Results of performance test in the description look wrong.
The time difference between _Before_ and _After_ is actually not as big as it's shown:
I have something like `1.720` VS `1.550`.
But it's still ~10% improvement for this test-case. And, as I see it, the test case is quite realistic.

The code of test to copy and run in the details:

<details>

```python
import asyncio

from fastapi import FastAPI
from pyinstrument import Profiler

app = FastAPI()

@app.get("/test/{num:int}")
async def test(num: int) -> dict:
    return {"echo": num}

async def send(message):
    pass

async def receive() -> dict:
    return {}

scope = {
    "type": "http",
    "asgi": {"version": "3.0", "spec_version": "2.3"},
    "http_version": "1.1",
    "server": None,
    "client": None,
    "scheme": "http",
    "root_path": "",
    "headers": [(b"token", b"1")],
    "state": {},
    "path": "/test",
    "raw_path": b"/test",
    "query_string": b"a=1&b=2",
    "method": "GET"
}

async def main():
    p = Profiler()

    async with app.router.lifespan_context(app):
        with p:
            await asyncio.gather(*(app(scope | {"path": scope["path"] + f"/{i}"}, receive, send) for i in range(10000)))
        p.print()


asyncio.run(main())

```
</details>

yinziyan1206:
Yes. It can reduce half of the `is_sequence_field` execution time  And the more `_get_multidict_value` is executed the more performance is affected.

YuriiMotov:
@yinziyan1206, please don't update branch if there are no merge conflicts. We will do it before merging. Only update it if there are conflicts that you need to resolve

tiangolo:
Thanks for the interest! This is a delicate area, the `ModelField` is only there to overcome that Pydantic v2 doesn't have an equivalent, but several internal parts expect it. Changes in this could affect one or the other.

If I was doing optimizations like this, I would instead add `@lru_cache` decorators on top of `is_sequence_field()` and probably others.

But for now I prefer to keep it as is, there are several refactors I would rather do before adding extra complexity to tweak a bit of performance, e.g. supporting both Pydantic v2 and v1 at the same time (with `pydantic.v1`).

For now I'll pass on this, but thanks! :coffee:

PR #13825, opened 2025-06-24
📝 Fix docs UI asset paths to respect `root_path` in Custom Docs code example

bertomaniac:
## Description

Updates docs for https://fastapi.tiangolo.com/how-to/custom-docs-ui-assets/ with better examples for handling `root_path` setups.

Updates tests to validate static assets when `root_path` is used in conjunction with self-hosting swagger/redoc assets.


YuriiMotov (changes requested):
Great!

But, there are a few things:
* It would be nice to avoid creating `test_offline_docs` and implement those tests in the existing `tests/test_tutorial/test_custom_docs_ui/test_tutorial002.py`
* lines highlighting needs to be fixed

Regarding the first point, main question is how to parameterize the app to pass `root_path` to its init. I think we can modify code example by adding the ENV variable that defaults to empty string. This way we will be able patch in tests:

Before:
```python
app = FastAPI(docs_url=None, redoc_url=None)
```

Suggestion:
```python
import os
root_path = os.getenv("ROOT_PATH", "")
app = FastAPI(docs_url=None, redoc_url=None, root_path=root_path)
```
And then we need to figure out how to patch this variable in test before importing the module, so that it would take effect.

@bertomaniac, what do you think?

bertomaniac:
@YuriiMotov: Ahh...I didn't realize there was a test already for `test_custom_docs_ui`. 

I'll move things over to there.

bertomaniac:
@YuriiMotov: Changes pushed. For some reason the `deploy-docs` check failed but I can't find a way to re-run it.

YuriiMotov:
@bertomaniac, as I see it, we need to test 5 cases:
* no `root_path` specified
* `root_path` specified via parameter to `FastAPI()`
    * access routes using paths with `/api` prefix
    * access routes using paths without prefix
* `root_path` specified via parameter to `TestClient()` (simulate `uvicorn --root-path /api`)
    * access routes using paths with `/api` prefix
    * access routes using paths without prefix

Could you please review the changes I've made?
Do they look clear? Do you agree with those changes?

We will need to make the same with `tutorial001` (add `root_path` to code example and update tests) since it also has the same problem

bertomaniac:
@YuriiMotov 

I think your updates look good and make sense.

I'll take a deeper look within the week and look to make the changes to the other test case.

bertomaniac:
@YuriiMotov 

I updated `tutorial_001.py` and updated the tests for it to handle the different `root_path/request_path/asgi_path` params. I also updated the highlighting line numbers on the `how-to/custom-docs-ui-assets` page.

I think this is probably in a good spot now.

YuriiMotov (approved):
LGTM

This PR fixes the code examples in [Custom Docs UI Static Assets (Self-Hosting)](https://fastapi.tiangolo.com/how-to/custom-docs-ui-assets/).
Before changes those examples fail if we run app under the path prefix (with `root_path`). In this case URLs of assets don't include path prefix and Swagger UI fails to load them in some cases.

<details><summary>Instructions to reproduce</summary>

Take a code example from [docs](https://fastapi.tiangolo.com/how-to/custom-docs-ui-assets/#disable-the-automatic-docs) and add `root_path="/api"` parameter:

```py
from fastapi import FastAPI
from fastapi.openapi.docs import (
    get_redoc_html,
    get_swagger_ui_html,
    get_swagger_ui_oauth2_redirect_html,
)

app = FastAPI(docs_url=None, redoc_url=None, root_path="/api")

@app.get("/docs", include_in_schema=False)
async def custom_swagger_ui_html():
    return get_swagger_ui_html(
        openapi_url=app.openapi_url,
        title=app.title + " - Swagger UI",
        oauth2_redirect_url=app.swagger_ui_oauth2_redirect_url,
        swagger_js_url="https://unpkg.com/swagger-ui-dist@5/swagger-ui-bundle.js",
        swagger_css_url="https://unpkg.com/swagger-ui-dist@5/swagger-ui.css",
    )

@app.get(app.swagger_ui_oauth2_redirect_url, include_in_schema=False)
async def swagger_ui_redirect():
    return get_swagger_ui_oauth2_redirect_html()

@app.get("/redoc", include_in_schema=False)
async def redoc_html():
    return get_redoc_html(
        openapi_url=app.openapi_url,
        title=app.title + " - ReDoc",
        redoc_js_url="https://unpkg.com/redoc@2/bundles/redoc.standalone.js",
    )

@app.get("/users/{username}")
async def read_user(username: str):
    return {"message": f"Hello {username}"}
```

Run app using `uvicorn main:app` and go to [http://localhost:8000/api/docs](http://localhost:8000/api/docs).
Open a source code of HTML page and see that URLs of `openapi.json` and `oauth2RedirectUrl` don't include path prefix, it's just `/openapi.json` and `window.location.origin + '/docs/oauth2-redirect'`. Remember we want to run our app under `/api` path prefix, so our proxy will not be able to route such request properly (it expects `/api` prefix).

Things are even worst if we use the code example with self-hosted scripts (and added `root_path="/api"`) - static files don't work at all (swagger UI is not loaded).


</details>

---

Updated code examples are in line with the current implementations of default `swagger_ui_html` ([link](https://github.com/fastapi/fastapi/blob/7179d48fd774bcbc5f5013bb8f4df61b8c36795f/fastapi/applications.py#L1014-L1026)) and `redoc_html` ([link](https://github.com/fastapi/fastapi/blob/7179d48fd774bcbc5f5013bb8f4df61b8c36795f/fastapi/applications.py#L1042-L1047)) routes.

Checked line highlights - all correct.

Related discussion: https://github.com/fastapi/fastapi/discussions/11148

PR #13827, opened 2025-06-24
🐛 Fix support for unions when using `Form`

patrick91:


tiangolo (approved):
Nice, thank you! 🚀 

This will be available in FastAPI 0.115.14, released in the next hours. 🤓

jaw997 (commented):
Be smart to getmore

jaw997 (commented):
More energy

njsbhu:
For function: _get_multidict_value(
As I can see this function pulling out request in one go. The logic will works fine no issue, but since it handles several steps together, a short note about the overall flow will make it easier for future readers to follow for the team.

In function: is_union_of_base_models() or _should_embed_body_fields()
I can see the way this handles form field unions is correct and yes It will covers an important edge case.  But Since unions of BaseModel need special treatment, adding a quick note here may help future readers to understand why they’re handled differently.

PR #13833, opened 2025-06-29
Fix ModelField alias property

pressogh:
Hello!

I've encountered an issue in FastAPI's `openapi.json` generation for parameter inputs (`Query`, `Header`, `Cookie`). When using `alias` for field names in Pydantic models, the alias appears correctly in the generated schema. However, when using `validation_alias`, it is ignored and the original field name is shown instead.

To demonstrate this behavior, I’ve included minimal reproducible examples along with corresponding Swagger UI screenshots.

### With alias
```python
from typing import Annotated

from fastapi import FastAPI, Header, Query, Cookie


class DataModel(BaseModel):
    foo: str = Field(..., alias="bar")


app = FastAPI()


@app.get("/test")
def test(
        query: Annotated[DataModel, Query(...)],
        header: Annotated[DataModel, Header(...)],
        cookie: Annotated[DataModel, Cookie(...)]
):
    return "Received: " + str(query) + ", " + str(header) + ", " + str(cookie)
```
image

### With validation_alias
```python
from typing import Annotated

from fastapi import FastAPI, Header, Query, Cookie


class DataModel(BaseModel):
    foo: str = Field(..., validation_alias="bar")


app = FastAPI()


@app.get("/test")
def test(
        query: Annotated[DataModel, Query(...)],
        header: Annotated[DataModel, Header(...)],
        cookie: Annotated[DataModel, Cookie(...)]
):
    return "Received: " + str(query) + ", " + str(header) + ", " + str(cookie)
```
image

As shown above, even though validation_alias="bar" is set, the OpenAPI schema still uses "foo".

Upon inspecting the parameter generation logic, I found that the name used in the schema is determined via the field alias:
https://github.com/fastapi/fastapi/blob/ebdeda2de6e17036e3048940b7a9e725ef6a95b7/fastapi/openapi/utils.py#L137

This uses Pydantic’s ModelField logic, where alias is returned if present; otherwise, the field name is used:
https://github.com/fastapi/fastapi/blob/ebdeda2de6e17036e3048940b7a9e725ef6a95b7/fastapi/_compat.py#L88-L97

### Proposed Change
This PR modifies the alias resolution logic in ModelField to prioritize validation_alias or serialization_alias, depending on mode, instead of defaulting to alias alone.

Thanks for taking the time to review this!

### DIFF
--- a/fastapi/_compat.py
+++ b/fastapi/_compat.py
@@ -93,13 +93,18 @@ if PYDANTIC_V2:
 
         @property
         def alias(self) -> str:
-            match self.mode:
-                case "validation" if self.field_info.validation_alias is not None:
-                    a = self.field_info.validation_alias
-                case "serialization" if self.field_info.serialization_alias is not None:
-                    a = self.field_info.serialization_alias
-                case _:
-                    a = self.field_info.alias
+            if (
+                self.mode == "validation"
+                and self.field_info.validation_alias is not None
+            ):
+                a = self.field_info.validation_alias
+            elif (
+                self.mode == "serialization"
+                and self.field_info.serialization_alias is not None
+            ):
+                a = self.field_info.serialization_alias
+            else:
+                a = self.field_info.alias
             return a if a is not None else self.name
 
         @property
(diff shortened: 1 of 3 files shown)


YuriiMotov:
This has been fixed! 🎉

`validation_alias` works well now:

image

@pressogh, thanks for you efforts!

PR #13861, opened 2025-07-03
🐛 Fix `generate_unique_id` to produce consistent IDs for operations with multiple methods

mohiuddin-khan-shiam:
## Description
[generate_unique_id](cci:1://file:///d:/Github/fastapi/fastapi/utils.py:178:0-183:23) picked `list(route.methods)[0]`, which depends on the internal,
non-deterministic order of a `set`.  
The resulting `operationId`s could oscillate between launches (e.g. `get_items_get`
vs `get_items_post`), breaking caches and client generation.

The fix sorts the HTTP methods first and then selects the first value, ensuring a
stable, predictable `operationId` across runs while preserving the original logic.

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -179,8 +179,15 @@ def generate_operation_id_for_path(
 def generate_unique_id(route: "APIRoute") -> str:
     operation_id = f"{route.name}{route.path_format}"
     operation_id = re.sub(r"\W", "_", operation_id)
+    # Ensure deterministic generation of the operation_id
+    # Sets are unordered, so converting directly to a list could give
+    # different results between runs, causing unstable OpenAPI specs.
+    # Sort the HTTP methods to guarantee a consistent order and pick the first
+    # one. This keeps the original behaviour (choosing one method) while
+    # making the result predictable.
     assert route.methods
-    operation_id = f"{operation_id}_{list(route.methods)[0].lower()}"
+    method = sorted(route.methods)[0].lower()
+    operation_id = f"{operation_id}_{method}"
     return operation_id
 
 


YuriiMotov:
@mohiuddin-khan-shiam, thanks for your interest in FastAPI!

I just found [similar PR](https://github.com/fastapi/fastapi/pull/4851). It was rejected with the reason:

> Using multiple methods for the same function is not actually supported. @router.api_route() is there as a semi-internal utility, and because Starlette supports multiple methods, the interface is there. But the official way to do it in FastAPI and have all the FastAPI features (including the automatic OpenAPI integration, with operation IDs) is by using the dedicated methods, like app.post(), app.get(), etc.

So, the proposed solution is to avoid using `.api_route("/", methods=["PUT", "POST"])`. Use multiple decorators with single methods instead:

```python
@app.put("/")
@app.post("/")
async def po() -> MyModel:
    return {}
```

Pavanmanikanta98:
Hi @tiangolo, @YuriiMotov ,

I've noticed that multiple users have created issues or PRs (like #4851 and #13861) related to using `@app.api_route()` or `@router.api_route()` with multiple HTTP methods (e.g., `methods=["POST", "PUT"]`). These lead to non-deterministic `operationId`s and confusion around FastAPI's expected routing behavior.

To help reduce future duplicates and improve the DX (developer experience), I'd like to propose adding a **clear subsection in the documentation**, either under:

- **"Path Operation Decorators"** (most ideal),
- or **"Dependencies in Path Operation Decorators"**, if that's more appropriate.

This subsection would clarify that FastAPI **does not officially support multiple HTTP methods on a single function**, even if Starlette does. Instead, users should define each method with its own decorator.

For example:

❌ Don't do this:
```python
@app.api_route("/items/", methods=["GET", "POST"])
async def get_or_create_item():
    ...
```

✅ Do this instead:
```python
@app.get("/items/")
async def read_items():
    return [{"item_id": "Foo"}]

@app.post("/items/")
async def create_item():
    return {"item_id": "Foo", "status": "created"}

```
Even if both routes share logic, users can extract it into a helper function to stay consistent with FastAPI's OpenAPI handling.

Let me know if you're open to a PR for this! I’d be happy to contribute the exact wording and placement based on your preferences. 

Thanks for the awesome framework 🙌

YuriiMotov:
@Pavanmanikanta98, I think we need such PR. I can't guarantee it will be merged, but I will support it.

Just one thing. I would implement the code example as one endpoint function with 2 path decorators.
```python
@app.put("/")
@app.post("/")
async def po(request: Request):
    if reques.method == "PUT":
        # Do something
    else:
        # Do something else
    return {}
```
I would also add docstring to `api_route` method with the warning and the link to the docs

hozn:
I was about to submit the same PR, so I'm happy someone beat me to it :)

IMO, the official recommendation of splitting into many methods is really ugly.  And this can be fixed with 1 line ....

In my case we have a diagnostic endpoint that supports GET/POST/PUT/DELETE/PATCH/OPTIONS/HEAD and there are detailed summary, description, examples for these.  Refactoring that into 7 methods that all have the same docs, etc. results in a lot of boilerplate code that is much more easily solved by just using the _public_ `api_route()` method.

tiangolo:
Thanks for the interest! But yes, `api_route` is not in the documentation, so it's not supported. It's only there because it's inherited from Starlette.

But this makes me thing maybe we could add a `@deprecated()` decorator to tell users not to use it and instead use the HTTP-method-specific decorators.

For now I'll pass on this, thanks! :coffee:

PR #13867, opened 2025-07-06
Create python-publish.yml

mortezave:


mortezave:
mortezave:mortezave-patch-1

YuriiMotov:
@mortezave, thanks for your interest in FastAPI!

I think we already have such script: https://github.com/fastapi/fastapi/blob/master/.github/workflows/publish.yml

Do you think your version is better? If yes, could you explain its advantages?

PR #13879, opened 2025-07-09
🐛 Fix typing for `CustomORJSONResponse.media_type`

BrunoTanabe:
🐛 Fix typing for `CustomORJSONResponse.media_type`

* Declare `media_type` as `ClassVar[str]`, importing `ClassVar` from `typing`.
* Treats the attribute as a class-level constant, matching FastAPI’s built-in response classes and silencing mypy/ruff strict-mode errors.
* Prevents rare `AttributeError` scenarios when `Response` subclasses are instantiated before the attribute is set.
* No runtime behaviour change; purely a correctness/compatibility fix.

### DIFF
--- a/docs_src/custom_response/tutorial009c.py
+++ b/docs_src/custom_response/tutorial009c.py
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, ClassVar
 
 import orjson
 from fastapi import FastAPI, Response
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@BrunoTanabe, thanks for your interest in FastAPI!

Could you please elaborate on "Prevents rare AttributeError scenarios when Response subclasses are instantiated before the attribute is set"? Could you give an example?

As I noticed, Sebastián tends not to increase the complexity of code examples in docs if it's not necessary. So, the arguments should be strong.

IMO, if people use mypy strict mode, they should know how to fix such type of errors. For most people it's not needed.

PR #13884, opened 2025-07-11
⬆️ Upgrade Starlette supported version range to `>=0.40.0,<0.48.0`

tiangolo:
⬆️ Upgrade Starlette supported version range to `>=0.40.0,<0.48.0`


mintuhouse:
Can you do a new release of fastapi with this change in 
Github started flagging it as vulnerability 
https://github.com/advisories/GHSA-2c2j-9gv5-cj73

PR #13887, opened 2025-07-11
fixed CWE-79 on docs.py

LucaPaterlini:


LucaPaterlini:
probably was needed to make the flaky test pass for the documentation...

YuriiMotov:
@LucaPaterlini, thanks for the interest in FastAPI!
Could you please add more reasoning on why we need to add this?

LucaPaterlini:
https://cwe.mitre.org/data/definitions/79.html
there is a CWE about not checking the parsed input.
Shall we assume the input is safe on Template?

Thanks for your replies was not expecting this celery response, great fan of your work guys

YuriiMotov:
> https://cwe.mitre.org/data/definitions/79.html there is a CWE about not checking the parsed input. Shall we assume the input is safe on Template?

So, as I understand it, the intruder can open a PR that injects malicious code in docs. And if somebody tries opening automatically generated docs preview that injected code may be executed on client side. Right?
I don't think it's possible that we have such malicious code in `master`, but the situation described above looks possible.

Do you have any idea why "Building Docs" action fails?

YuriiMotov:
As an alternative approach, we can disable automatic docs preview deployments for PRs and do it only after maintainers check the diff and explicitly request it

tiangolo:
Thanks, I don't think this is currently needed, that script is only used internally by us to generate the docs, it never receives external input. I'll pass on this one for now, thanks! :coffee:

PR #13890, opened 2025-07-13
Replace deprecated `asyncio.iscoroutinefunction` with `inspect.iscoroutinefunction`

edgarrmondragon:
Deprecated in Python 3.14 and will be removed in Python 3.16.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1,4 +1,3 @@
-import asyncio
 import dataclasses
 import email.message
 import inspect
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -231,7 +231,7 @@ def get_request_handler(
     embed_body_fields: bool = False,
 ) -> Callable[[Request], Coroutine[Any, Any, Response]]:
     assert dependant.call is not None, "dependant.call must be a function"
-    is_coroutine = asyncio.iscoroutinefunction(dependant.call)
+    is_coroutine = inspect.iscoroutinefunction(dependant.call)
     is_body_form = body_field and isinstance(body_field.field_info, params.Form)
     if isinstance(response_class, DefaultPlaceholder):
         actual_response_class: Type[Response] = response_class.value


YuriiMotov:
@edgarrmondragon, thanks for your interest!

I think it's early to do this refactor. We don't have any warning and without it we will miss occurrences of `asyncio.iscoroutinefunction` in new PRs and merge them.

Do you think it's right time for this? I think we can wait till FastAPI supports Python 3.16 or other versions start to emit deprecation warnings

edgarrmondragon:
> I think it's early to do this refactor. We don't have any warning and without it we will miss occurrences of `asyncio.iscoroutinefunction` in new PRs and merge them. Do you think it's right time for this?

That's a good point. Final release of Python 3.14.0 is scheduled for [2025-10-07](https://peps.python.org/pep-0745/), and the first RC is only a little over a week away at this point, so I think

- the short time frame
- the fact that more people will start testing their FastAPI apps with Python 3.14 in the coming weeks now that pydantic-core supports it (that's how I discovered this)
- there's no functional differences between the two methods

are good reasons to do this at this point.

> I think we can wait till FastAPI supports Python 3.16 or other versions start to emit deprecation warnings

If you want to wait, the earliest I'd recommend is when final 3.14.0 comes out, but that's of course completely up to you.

YuriiMotov (commented):
I think we should wait until all pending PRs (or most of them) have been reviewed and merged, then make sure we haven't introduced any new cases where deprecated call is used and then merge this.
Also, we need to check for other deprecated methods used as well.

I think this will happen automatically when Python 3.16 is released and FastAPI officially supports it (will start running tests on Python 3.16)

OrangeDoro (commented):
Hi! I'm a grad student working on a research project about using large language models to automate code review. Based on your commit 66ec6495205961a830848a388f5a16036416be62 and the changes in fastapi/routing.py, my tool generated this comment:
1. Verify that the removal of the `asyncio` import does not affect any existing asynchronous code paths. If there are any async functions or operations that depend on `asyncio`, they should be reviewed to ensure they are still functioning correctly.
2. Conduct thorough testing to ensure that the removal does not introduce any race conditions or deadlocks in the code, especially if the codebase previously relied on asynchronous execution.
3. Ensure that there are existing tests that cover the functionality that may have previously relied on `asyncio`. If there are no tests that validate the behavior of the code that interacts with asynchronous operations, it would be prudent to add them.
4. Review the test coverage for the affected file. If the tests do not adequately cover the scenarios where `asyncio` was previously utilized, you should add tests to ensure that all code paths are validated.
5. Ensure that each of the remaining imports (`dataclasses`, `email.message`, and `inspect`) is necessary for the functionality of the code. If any of them are not used, consider removing them to enhance clarity.

As part of my research, I'm trying to understand how useful these comments are in real-world development. If you have a moment, I'd be super grateful if you could quickly reply to these two yes/no questions:


1. Does this comment provide suggestions from a dimension you hadn’t considered?

2. Do you find this comment helpful?

Thanks a lot for your time and feedback! And sorry again if this message is a bother.

YuriiMotov:
See also: https://github.com/fastapi/fastapi/discussions/14021

edgarrmondragon:
Closed in favor of #14022

PR #13892, opened 2025-07-14
md 수정22222

fastapi25:
2222


fastapi25:
tt

PHclaw:
For the dependency injection lifecycle issue:

Use `asynccontextmanager` for proper cleanup:
```python

@asynccontextmanager
async def get_db_session():
    session = Session()
    try:
        yield session
        await session.commit()
    except Exception:
        await session.rollback()
        raise
    finally:
        await session.close()
```

This ensures proper cleanup even if the request fails.

PR #13896, opened 2025-07-15
Fix typo: excption -> exception

CuberQAQ:
thx for checking


YuriiMotov:
@CuberQAQ, thanks for your interest in FastAPI!

We're planning to update how we handle documentation translations soon.
To avoid bloating the Git history, we've decided not to merge small typo fixes in translations for now. These issues will hopefully be resolved once the new translation mechanism is in place.

With that, I'll close this PR.
Thanks again for your efforts!

PR #13898, opened 2025-07-16
Add minimal reproduction for form field metadata bug in FastAPI

ASR1015:
This PR adds a minimal test case to reproduce a bug in FastAPI where metadata like `description` from `Form()` is lost when used inside a dependency with `Depends()`.

Included files:
- `form_dependency_test.py`: Demonstrates the issue using `Depends()` with `Form()`.
- `form_metadata_repro.py`: Shows correct behavior when `Form()` is used directly.

Also includes a `README.md` with explanation and instructions to run.

This is related to a known issue and can be helpful for debugging or submitting to upstream.

### DIFF
--- a/test_issue_13471/form_dependency_test.py
+++ b/test_issue_13471/form_dependency_test.py
@@ -1,0 +1,19 @@
+from fastapi import FastAPI, Form, Depends
+from pydantic import BaseModel
+from fastapi.responses import JSONResponse
+
+app = FastAPI()
+
+class User(BaseModel):
+    name: str
+    age: int
+
+def get_user(
+    name: str = Form(..., description="The user's name"),
+    age: int = Form(..., description="The user's age")
+) -> User:
+    return User(name=name, age=age)
+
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@ASR1015, both files you posted are identical. Could you take a look?
And, I can't reproduce the problem using the code you provided

Screenshot from 2025-07-17 08-32-16

ASR1015:
Thanks for your feedback! I just re-tested the example and confirmed that the form metadata is correctly being shown under the request body.
The issue might have been caused earlier due to not being in the right working directory or module not importing properly.
You can now see the form field descriptions under /submit in Swagger UI — please let me know if you’d like a video demo or if any clarification is needed!

YuriiMotov:
As I understand it, the issue has been resolved and we can close this PR now, right?

ASR1015:
Yes, the issue is resolved. Thank you for your support! You can go ahead and close this PR.

PR #12452, opened 2024-10-14
♻️ Refactor loop to list comprehension for performance optimization

0xsirsaif:
This PR refactored loop to list comprehension for minor performance optimization in `jsonable_encoder`, referencing [this explanation](https://stackoverflow.com/questions/30245397/why-is-a-list-comprehension-so-much-faster-than-appending-to-a-list).

PR #13902, opened 2025-07-18
perf: optimize sequence encoding with list comprehension

killerdevildog:
Replace for-loop with append() in jsonable_encoder for better performance when encoding sequences (list, tuple, set, frozenset, deque, GeneratorType).

🚀 FastAPI List Comprehension Optimization Benchmark
=======================================================
🎯 Focus: Sequence encoding performance (lists, tuples, sets)

📊 Generating sequence-heavy test data (1000 items)...
✅ Generated 1000 test items with heavy sequence usage

📈 Dataset contains ~660 primary sequences

⏱️  Running benchmarks (100 iterations each)...

🐌 Benchmarking OLD approach (for-loop + .append())...
🚀 Benchmarking NEW approach (list comprehensions)...

📈 BENCHMARK RESULTS
========================================

🐌 OLD (for-loop + .append()):
   Average: 3.125ms ± 0.964ms
   Range:   2.383ms - 6.784ms

🚀 NEW (list comprehensions):
   Average: 2.591ms ± 0.335ms
   Range:   2.345ms - 4.189ms

🎯 PERFORMANCE IMPROVEMENT
==============================
⚡ Improvement: 17.09% faster
📊 Speedup:     1.206x

✅ List comprehension optimization shows significant improvement!
🏆 This optimization is worthwhile for production use.

HYPER FINE BENCHMARK 

Benchmark 1: python benchmark_test/hyperfine_test.py old

Time (mean ± σ): 26.9 ms ± 2.9 ms [User: 23.1 ms, System: 3.7 ms]
Range (min … max): 24.9 ms … 38.9 ms (99 runs)

Benchmark 2: python benchmark_test/hyperfine_test.py new

Time (mean ± σ): 26.3 ms ± 2.0 ms [User: 22.4 ms, System: 3.8 ms]
Range (min … max): 25.0 ms … 38.4 ms (116 runs)

Performance Benchmark Results:

- Dataset: 1,000 items with heavy sequence usage (~660 sequences)
- Test Method: 100 iterations each, sequence-focused encoding
- Before (for-loop + .append()): 3.12ms ± 0.96ms
- After (list comprehensions): 2.59ms ± 0.34ms
- Performance Gain: 17.1% faster (1.21x speedup)

The optimization shows consistent improvement in sequence encoding,
which is a common operation in FastAPI JSON serialization.

🔍 VERIFICATION:
---------------
✅ Both implementations produce identical results
Benefits:
- Reduced function call overhead
- Optimized C implementation
- Better memory allocation
- 100% backward compatible

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -297,22 +297,20 @@ def jsonable_encoder(
                 encoded_dict[encoded_key] = encoded_value
         return encoded_dict
     if isinstance(obj, (list, set, frozenset, GeneratorType, tuple, deque)):
-        encoded_list = []
-        for item in obj:
-            encoded_list.append(
-                jsonable_encoder(
-                    item,
-                    include=include,
-                    exclude=exclude,
-                    by_alias=by_alias,
-                    exclude_unset=exclude_unset,
-                    exclude_defaults=exclude_defaults,
-                    exclude_none=exclude_none,
-                    custom_encoder=custom_encoder,
-                    sqlalchemy_safe=sqlalchemy_safe,
-                )
+        return [
+            jsonable_encoder(
+                item,
+                include=include,
+                exclude=exclude,
+                by_alias=by_alias,
+                exclude_unset=exclude_unset,
+                exclude_defaults=exclude_defaults,
+                exclude_none=exclude_none,
+                custom_encoder=custom_encoder,
+                sqlalchemy_safe=sqlalchemy_safe,
             )
-        return encoded_list
+            for item in obj
+        ]
 
     if type(obj) in ENCODERS_BY_TYPE:
         return ENCODERS_BY_TYPE[type(obj)](obj)


valentinDruzhinin:
Very interesting change!
But I think, that a regular for loop was used intentionally for better reading.
However, I don't see any issue with using list comprehension, but for future maintainability I would stay with current approach. Because we can add new params to the `jsonable_encoder` or we may prefer to do some calculations before calling the `jsonable_encoder`. So this part will be rewritten again if we add a little change.
Performance in ~20% is good, but I don't think it's very sufficient.
Just a thought :)
What do you think @YuriiMotov ?

YuriiMotov:
This PR is a duplicate of https://github.com/fastapi/fastapi/pull/12452.

I tried testing the #12452 and i got a decrease in performance instead of increase. See [my comment](https://github.com/fastapi/fastapi/pull/12452#pullrequestreview-3000759067).
The author of #12452 hasn't replied yet.
@killerdevildog, could you take a look and address it?

PR #13906, opened 2025-07-20
📝 Add deprecation info block about `dict()` in `docs/tutorial/body.md`

jomkv:
### **Description**

Updates docs for https://fastapi.tiangolo.com/tutorial/body/#use-the-model with an info about .dict() being deprecated, and that .model_dump() should be used instead.

Refer to discussion #13904  for more details.

### **Change Preview (refer to the info at the bottom)**
image


YuriiMotov:
@jomkv, thanks for you interest!

I remember I saw such note somewhere. I think it would be nice to find it and use the same wording for consistency.
Could you please try to find it?

jomkv:
> @jomkv, thanks for you interest!
> 
> I remember I saw such note somewhere. I think it would be nice to find it and use the same wording for consistency. Could you please try to find it?

I actually just copied it from there :) forgot to mention that, sorry. [Here ](https://fastapi.tiangolo.com/tutorial/extra-models/#multiple-models) is where I found it, just below the code snippet of "Multiple Models".

YuriiMotov (commented):
LGTM

I checked the wording is exactly the same that is used in other places, and this is the only place where `.dict()` is used without such warning

valentinDruzhinin (approved):

Shihasz:
Pydantic V2 is released long ago and fastapi is no more using V1. I think it is better to change code itself to use model_dump() instead of dict() method which is deprecated already and not recommended to be used in especially in tutorials for new users of the framework. If absolutely necessary, note/info can be added about the method dict() had to be used in v1.

YuriiMotov:
> Pydantic V2 is released long ago and fastapi is no more using V1. I think it is better to change code itself to use model_dump() instead of dict() method which is deprecated already and not recommended to be used in especially in tutorials for new users of the framework.

FastAPI still supports Pydantic V1 and all code examples from docs are tested with both, Pydantic V1 and V2.
If we change it to `.model_dump()` then tests with Pydantic V1 will fail.
We could solve this by creating separate code examples for Pydantic V1 and V2 and showing V2 version by default.

tiangolo (approved):
Great, thanks! 🚀

PR #13918, opened 2025-07-23
✨ Allow custom `JsonSchemaGenerator` for OpenAPI generation

stickm4n:
# Intro
OpenAPI generators have grom a lot by themselves and despite all comply with base OpenAPI specs, many have implemented extra features that would be good to have when getting the schema from `FastAPI`.

# Example use case:
```python
class Gender(IntEnum):
    male = auto()
    female = auto()
    other = auto()
    not_given = auto()

ta = TypeAdapter(Gender)
ta_schema = ta.json_schema()

# This generates the following
assert ta_schema == {
    "enum": [1, 2, 3, 4],
    "title": "Gender",
    "type": "integer"
}
```

The inconvenience here comes because OpenAPI specs does not hace naming for `IntEnum` class attributes and most be created by generators without any naming convention...
TypeScript generated code with `openapi-generator-cli` would look like this and is completely abstract:

```ts
export const Gender= {
    NUMBER_1: 1,
    NUMBER_2: 2,
    NUMBER_3: 3,
    NUMBER_4: 4
} as const;
export type Gender = typeof Gender[keyof typeof Gender];
```

These generators have implemented extra definitions to provide these attributes naming association. This varies depending on the generator but if we keep on with the one we are using already (`openapi-generator-cli`) that would be the `x-enum-varnames`.

With a simple implementation, as intended by `pydantic`, we could get this done in a blink of an eye.
```python
class CustomJsonSchemaGenerator(GenerateJsonSchema):
    def enum_schema(self, schema: EnumSchema) -> JsonSchemaValue:
        generated_schema = super().enum_schema(schema)
        generated_schema['x-enum-varnames'] = [enum.name for enum in schema['members']]

        return generated_schema
```
With the output being:
```ts
export const Gender= {
    MALE: 1,
    FEMALE: 2,
    OTHER: 3,
    NOT_GIVEN: 4
} as const;
export type Gender = typeof Gender[keyof typeof Gender];
```

# Conclusion
`FastAPI` already has the interface to make this, it just lacks the parameter of the generator to be used. So, adding that to its implementation would be all what is required to have this feature, as shown in the code changes. The rest of `FastAPI`-ways for openapi overrideshown in the doc will remain the same and the doc will not require update

### DIFF
--- a/fastapi/_compat/main.py
+++ b/fastapi/_compat/main.py
@@ -35,7 +35,7 @@ if PYDANTIC_V2:
 else:
     from .v1 import BaseConfig as BaseConfig  # type: ignore[assignment]
     from .v1 import FieldInfo as FieldInfo
-    from .v1 import GenerateJsonSchema as GenerateJsonSchema
+    from .v1 import GenerateJsonSchema as GenerateJsonSchema  # type: ignore[assignment]
     from .v1 import (  # type: ignore[assignment]
         PydanticSchemaGenerationError as PydanticSchemaGenerationError,
     )
--- a/fastapi/_compat/main.py
+++ b/fastapi/_compat/main.py
@@ -3,6 +3,7 @@ from typing import (
     Any,
     Dict,
     List,
+    Optional,
     Sequence,
     Tuple,
     Type,
@@ -19,6 +20,7 @@ from .model_field import ModelField
 if PYDANTIC_V2:
     from .v2 import BaseConfig as BaseConfig
     from .v2 import FieldInfo as FieldInfo
+    from .v2 import GenerateJsonSchema
     from .v2 import PydanticSchemaGenerationError as PydanticSchemaGenerationError
     from .v2 import RequiredParam as RequiredParam
     from .v2 import Undefined as Undefined
@@ -33,6 +35,7 @@ if PYDANTIC_V2:
 else:
     from .v1 import BaseConfig as BaseConfig  # type: ignore[assignment]
     from .v1 import FieldInfo as FieldInfo
+    from .v1 import GenerateJsonSchema
     from .v1 import (  # type: ignore[assignment]
         PydanticSchemaGenerationError as PydanticSchemaGenerationError,
     )
@@ -231,6 +234,7 @@ def get_definitions(
     fields: List[ModelField],
     model_name_map: ModelNameMap,
     separate_input_output_schemas: bool = True,
+    schema_generator: Optional[GenerateJsonSchema] = None,
 ) -> Tuple[
     Dict[Tuple[ModelField, Literal["validation", "serialization"]], v1.JsonSchemaValue],
     Dict[str, Dict[str, Any]],
@@ -251,6 +255,7 @@ def get_definitions(
             fields=v2_fields,
             model_name_map=model_name_map,
             separate_input_output_schemas=separate_input_output_schemas,
+            schema_generator=schema_generator,
         )
         all_definitions = {**v1_definitions, **v2_definitions}
         all_field_maps = {**v1_field_maps, **v2_field_maps}
--- a/fastapi/_compat/main.py
+++ b/fastapi/_compat/main.py
@@ -20,7 +20,7 @@ from .model_field import ModelField
 if PYDANTIC_V2:
     from .v2 import BaseConfig as BaseConfig
     from .v2 import FieldInfo as FieldInfo
-    from .v2 import GenerateJsonSchema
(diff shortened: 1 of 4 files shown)


YuriiMotov:
@stickM4N, thanks for your interest!

This is a duplicate of https://github.com/fastapi/fastapi/pull/12455.
Not sure the author of that PR is still interested in it.
[This comment](https://github.com/fastapi/fastapi/pull/12455#issuecomment-3069767849) is also applicable to your PR

stickm4n:
@YuriiMotov Thanks for the quick feedback. When I was doing research about this, I saw that PR (and also an [old branch](https://github.com/fastapi/fastapi/tree/feat/custom-generate-json-schema) still in the repo, probably deprecated...)
Their approach to this is interacting with `FastAPI` logic directly, and though i like that approach I opted by just interacting with the utility it gives to form the requests.
It preserved the `FastAPI`-way to do the schema generation override, reducing the impacts on other parts of the module and still giving the feature.
Probably a better solution is to use the other approach and pass that content directly to App initialization...
I think this is a good feature that con be implemented quickly using the current interface (already prepared for it)

YuriiMotov:
> I opted by just interacting with the utility it gives to form the requests. It preserved the `FastAPI`-way to do the schema generation override, reducing the impacts on other parts of the module and still giving the feature.

I agree, this should be enough.

We still need to add test for this feature

YuriiMotov (approved):
I think this is quite useful feature that was requested many times.

This will allow solving (overcome) multiple issues with schema generation:
* adding nice names for ENUM parameters (described in this PR)
* fix schema for `Literal["something"]` described in [this PR](https://github.com/fastapi/fastapi/pull/12455)
* fix schema for optional Query parameters described in those PRs: #12375, #9873
* change sorting order, described in [this discussion](https://github.com/fastapi/fastapi/discussions/13263) and others
* I believe there are many other use-cases where it might be useful

In the future we may want to switch to using our own customized implementation of `GenerateJsonSchema`, but for now it would be nice to provide the way for users to workaround some of mentioned (and other) issues.

I'm not sure whether we need to add docs for this feature. I think we can skip it for now and add it later.

stickm4n:
Thanks for the review, but I'll add tests soon and I would like to discuss with you another point that is worth taking into account.
So please don't merge it yet...

The idea I want to check with you:
FastAPI uses a custom schema ref location that is passed to the generator, is it required for all generators to use it? I think not, but wanted to put that into discussion before making any assumptions...

stickm4n:
@YuriiMotov Ohh I just realized you added tests by yourself...
This method is just partially documented, so no need to update any docs. It could be merged already.

Thanks for the quick review and feedback @YuriiMotov.
I'll be waiting for this features release.

YuriiMotov:
@stickM4N, you are welcome)
So, can we mark [this](https://github.com/fastapi/fastapi/pull/13918#issuecomment-3144163005) as resolved?
You said you wanted to discuss something related to "ref location"

stickm4n:
@YuriiMotov Yes we could mark it as resolved!
I think it was covered by your test, so it's not necessary to use REF_TEMPLATE for all generators, right?

mwildehahn:
can this get merged?

stickm4n:
@mwildehahn Here you have an example to have the functionality while its merged...
Just call `CustomJsonSchemaGenerator.inject_fastapi()` after creating your app.

```python
from sys import modules

from fastapi.openapi.utils import get_openapi
from pydantic.json_schema import GenerateJsonSchema, JsonSchemaValue
from pydantic_core.core_schema import EnumSchema


class CustomJsonSchemaGenerator(GenerateJsonSchema):

    @classmethod
    def inject_fastapi(cls) -> None:
        modules[get_openapi.__module__].GenerateJsonSchema = cls

    def enum_schema(self, schema: EnumSchema) -> JsonSchemaValue:
        # Override this or any other method you want.
        ...
```

stickm4n:
@YuriiMotov @tiangolo This PR got broken after yesterdays changes to support both `pydantic` v1 and v2 in the same app, so I will adapt this code to the new structure.

stickm4n:
tests are pending after refactor, will do it next!

nsoranzo:
@mvdbeek This is another attempt similar to your https://github.com/fastapi/fastapi/pull/10903 which should solve [our failing tests](https://github.com/galaxyproject/galaxy/actions/runs/18748573516/job/53481856385) with FastAPI>=0.119.0 .

tiangolo:

Thank you! 🍰

PR #13793, opened 2025-06-15
📝 Update `docs/en/docs/advanced/generate-clients.md`

mrlubos:
I've made the following changes to improve this guide.
- update title to reference "SDKs" over "clients" as people are more familiar with this term
- highlight other use cases for OpenAPI beyond generating SDKs
- add a link to OpenAPI.Tools so people can discover more tools
- highlight that FastAPI uses OpenAPI 3.1, so tooling needs to support that version
- clarify the sponsors section
- update Stainless domain
- use the `npx` command for `@hey-api/openapi-ts` so it's faster to get started
- fix broken "copy to clipboard" buttons
- simplify right-hand side navigation from 3 to 2 levels
- reference Hey API docs for more details, so it's hopefully easier to maintain this guide
- add a note encouraging people to open an issue if they want us to update the remaining out-of-date references

PR #13919, opened 2025-07-23
📄🪲 Fix: incorrect documentation on openapi-ts (Generating Typescript Clients)

Ali-Parandeh:
This PR fixes the incorrect command pointing to the wrong location for `openapi.json` on the FastAPI server. It also removes the need for needing to install additional axios plugin as it's not needed. fetch API is more than enough as default client.

### DIFF
--- a/docs/en/docs/advanced/generate-clients.md
+++ b/docs/en/docs/advanced/generate-clients.md
@@ -81,7 +81,7 @@ It could look like this:
   "description": "",
   "main": "index.js",
   "scripts": {
-    "generate-client": "openapi-ts --input http://localhost:8000/openapi.json --output ./src/client --client axios"
+    "generate-client": "openapi-ts --input http://localhost:8000/api/openapi.json --output ./src/client"
   },
   "author": "",
   "license": "",
@@ -100,12 +100,12 @@ After having that NPM `generate-client` script there, you can run it with:
 $ npm run generate-client
 
 frontend-app@1.0.0 generate-client /home/user/code/frontend-app
-> openapi-ts --input http://localhost:8000/openapi.json --output ./src/client --client axios
+> openapi-ts --input http://localhost:8000/api/openapi.json --output ./src/client
 ```
 
 </div>
 
-That command will generate code in `./src/client` and will use `axios` (the frontend HTTP library) internally.
+That command will generate code in `./src/client`.
 
 ### Try Out the Client Code
 


docs/en/docs/advanced/generate-clients.md
```diff
@@ -81,7 +81,7 @@ It could look like this:
   "description": "",
   "main": "index.js",
   "scripts": {
-    "generate-client": "openapi-ts --input http://localhost:8000/openapi.json --output ./src/client --client axios"
+    "generate-client": "openapi-ts --input http://localhost:8000/api/openapi.json --output ./src/client"
```
YuriiMotov, line 84: The application showed above serves `openapi.json` at default path, which is `http://localhost:8000/openapi.json`. No need to add `api/`

YuriiMotov (commented):
@Ali-Parandeh, thanks for your interest!

I think there is no need to change this.
`--client` argument is optional, but it's not a mistake to use it.
Even if we update it here, we will have old version in translations of this section.

So, let's avoid making unnecessary changes and focus on more important things!

Ali-Parandeh:
Hi @YuriiMotov Totally understand. Too much effort to update some details in the advanced guide pages. In particular, if you have to update it in all languages.

Just note that the typescript client mentioned in this document is now working totally different so all the code examples won't work out of box. Also you're pointing to the wrong path for the openapi.json in the docs. If someone who's not savvy with how FastAPI and typescript works tries to follow this page in documentation, they'll just get errors and can get frustrated as code examples don't work out of box.

YuriiMotov:
@Ali-Parandeh, thanks for reply!

> Also you're pointing to the wrong path for the openapi.json in the docs.

By default `openapi.json` is served by FastAPI app at `http://localhost:8000/openapi.json`.
The code example in this section of docs doesn't override `openapi.json` path. So, attempt to access it at `http://localhost:8000/api/openapi.json` will result in `404 Not found`.
Am I missing something?

As for

> typescript client mentioned in this document is now working totally different so all the code examples won't work out of box. 

Does this PR solve this?
As I see now, to use axios we need to add `@hey-api/client-axios` plugin. Are you talking about this problem or are there any other issues?

alejsdev:
Hello @Ali-Parandeh,

Thanks for your interest in contributing. However, this is already being addressed in #13793, which was started earlier by @mrlubos, the maintainer of [hey-api/openapi-ts](https://github.com/hey-api/openapi-ts). We'll likely prioritize that PR for consistency and continuity. Given that, I'll close this one but thanks for the effort!

mrlubos:
I better get on that huh 😀

mrlubos:
Ready https://github.com/fastapi/fastapi/pull/13793

PR #13920, opened 2025-07-24
✨ Add support for PEP695 `TypeAliasType`

### LINKED ISSUE #10719: `Annotated` dependencies are interpreted incorrectly when using `PEP 695`-style type alias.
### Discussed in https://github.com/tiangolo/fastapi/discussions/10662

<sup>Originally posted by **Rockmizu** November 16, 2023</sup>
### First Check

### Commit to Help

### Example Code

```python

from typing import Annotated, TypeAlias

app = FastAPI()


# dependency
async def some_value() -> int:
    return 123

# This works.
DependedValue: TypeAlias = Annotated[int, Depends(some_value)]

# This won't work.
# type DependedValue = Annotated[int, Depends(some_value)]


@app.get('/')
async def get_with_dep(value: DependedValue) -> str:
    print(f'{type(value) = !r}')
    print(f'{value = !r}')

    assert isinstance(value, int), '`value` should be an integer.'
    assert value == 123, '`value` should be 123.'

    return f'value: {value}'
```

### Description

When using the method in line 15:
```python
DependedValue: TypeAlias = Annotated[int, Depends(some_value)]
```
to declare a type alias, FastAPI correctly interprets it and fills the `value` with `123` for `get_with_dep`.

However, when using the new type alias syntax introduced in Python 3.12 (line 18):
```python
type DependedValue = Annotated[int, Depends(some_value)]
```
FastAPI interprets it as a URL query parameter and responds with HTTP 422 Unprocessable Entity when accessing `/`.

## Expected Response

HTTP 200 OK

```
"value: 123"
```

## Actual Response

HTTP 422 Unprocessable Entity

```
{"detail":[{"type":"missing","loc":["query","value"],"msg":"Field required","input":null,"url":"https://errors.pydantic.dev/2.5/v/missing"}]}
```

### Operating System

Windows

### Operating System Details

_No response_

### FastAPI Version

0.104.1

### Pydantic Version

2.5.1

### Python Version

Python 3.12.0

### Additional Context

_No response_</div>

cstruct:
This is a continuation of #11140 with the comment by @Viicos fixed by vendoring a copy of `is_typealiastype` from [`typing-inspection`](https://github.com/pydantic/typing-inspection).


cstruct:
I'm unsure how to get pre-commit.ci to accept the Python 3.12 PEP695 syntax without it trying to update to the new union syntax. Would love some guidance if anyone has any ideas! If I run ruff locally with 3.12 it behaves as expected but updating `.pre-commit-config` to have `default_language_version` for `python` set to `python3.12` does not help. What is also confusing is that the error states Python 3.8 but as far as I can se it is configured to run with 3.10?

As for the failing 3.8 builds I'm also unsure with how to proceed, `typing-inspection` was recommended by @Viicos but unfortunately only supports Python 3.9 and above. When is the support window for Python 3.8 ending for FastAPI since it is already eol?

Viicos:
> As for the failing 3.8 builds I'm also unsure with how to proceed, `typing-inspection` was recommended by @Viicos but unfortunately only supports Python 3.9 and above. When is the support window for Python 3.8 ending for FastAPI since it is already eol?

I don't know if there is any policy, but I would strongly suggest the team to drop support for 3.8 (it is eol for almost a year now). Anyway, if this happens, it should be split off in a separate PR.

As an alternative, FastAPI can ship the `typing-inspection` implementation if `is_typealiastype()`.

> I'm unsure how to get pre-commit.ci to accept the Python 3.12 PEP695 syntax

You can use `TypeAliasType` directly, it is valid syntax in all Python versions:

```python
MyType = TypeAliasType('MyType', list[T], type_params=(T,))
```

cstruct:
Thank you for the feedback and guidance @Viicos! I did not think of using the `TypeAliasType` directly.

fastapi/dependencies/utils.py
```diff
@@ -1010,3 +1020,26 @@ def get_body_field(
         field_info=BodyFieldInfo(**BodyFieldInfo_kwargs),
     )
     return final_field
+
+
+def _is_typealiastype(tp: Any, /) -> TypeGuard[TypeAliasType]:
+    in_typing = hasattr(typing, "TypeAliasType")
```
```suggestion
    """
    This implementation is in line with the implementation of `is_typealiastype` in `typing-inspection`:
    See:
    - https://github.com/pydantic/typing-inspection/blob/8db011350942f33ac4b5d7db60d4d9ea83ab480f/src/typing_inspection/typing_objects.py#L488-L499
    - https://github.com/pydantic/typing-inspection/blob/8db011350942f33ac4b5d7db60d4d9ea83ab480f/src/typing_inspection/typing_objects.py#L105-L134
    """

    # Check if TypeAliasType exists in typing and/or typing_extensions.
    in_typing = hasattr(typing, "TypeAliasType")
```

fastapi/dependencies/utils.py
```diff
@@ -1010,3 +1020,26 @@ def get_body_field(
...
+    in_typing_extensions = hasattr(typing_extensions, "TypeAliasType")
+    is_typealiastype = False
```
```suggestion

    is_typealiastype = False  # Default: assume not a TypeAliasType
```

fastapi/dependencies/utils.py
```diff
@@ -1010,3 +1020,26 @@ def get_body_field(
...
+    if in_typing and in_typing_extensions:
+        if getattr(typing, "TypeAliasType", None) is getattr(
+            typing_extensions, "TypeAliasType", None
+        ):  # pragma: no cover
+            is_typealiastype = isinstance(tp, typing.TypeAliasType)  # type: ignore [attr-defined]
```
```suggestion
            # Case 1: Both implementations are the same object.
            # Checking against one of them.
            is_typealiastype = isinstance(tp, typing.TypeAliasType)  # type: ignore [attr-defined]
```

fastapi/dependencies/utils.py
```diff
@@ -1010,3 +1020,26 @@ def get_body_field(
...
+        else:
+            is_typealiastype = isinstance(
```
```suggestion
            # Case 2: Implementations are different objects.
            # Need to check against both versions.
            is_typealiastype = isinstance(
```

fastapi/dependencies/utils.py
```diff
@@ -1010,3 +1020,26 @@ def get_body_field(
...
+                tp,
+                (typing.TypeAliasType, typing_extensions.TypeAliasType),  # type: ignore [attr-defined]
+            )
+    elif in_typing and not in_typing_extensions:  # pragma: no cover
+        is_typealiastype = isinstance(tp, typing.TypeAliasType)  # type: ignore [attr-defined]
```
```suggestion
        # Case 3: Only typing.TypeAliasType exists.
        is_typealiastype = isinstance(tp, typing.TypeAliasType)  # type: ignore [attr-defined]
```

fastapi/dependencies/utils.py
```diff
@@ -1010,3 +1020,26 @@ def get_body_field(
...
+    elif not in_typing and in_typing_extensions:
+        is_typealiastype = isinstance(tp, typing_extensions.TypeAliasType)
```
```suggestion
        # Case 4: Only typing_extensions.TypeAliasType exists.
        is_typealiastype = isinstance(tp, typing_extensions.TypeAliasType)
```

fastapi/dependencies/utils.py
```diff
@@ -1010,3 +1020,26 @@ def get_body_field(
...
+    if sys.version_info[:2] == (3, 10):
```
```suggestion

    # Special case for Python 3.10:
    # On Python 3.10, parameterized PEP 695 type aliases are represented as `GenericAlias`
    # instead of proper TypeAliasType. We must exclude those.
    if sys.version_info[:2] == (3, 10):
```

fastapi/dependencies/utils.py
```diff
@@ -364,6 +371,9 @@ def analyze_param(
     depends = None
     type_annotation: Any = Any
     use_annotation: Any = Any
+    if _is_typealiastype(annotation):
+        # unpack in case py3.12 type syntax is used
```
```suggestion
        # unpack in case PEP 695 type syntax is used
```

tests/test_dependency_pep695.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
```
YuriiMotov: Do we really need this import?

YuriiMotov (approved):
LGTM!

I suggest adding comments to the code to simplify review process.

@cstruct, thanks you for working on this!

Related issue: #10719

tests/test_dependency_pep695.py
stickm4n: I also think is redundant and can be removed

stickm4n:
Thanks for the details in this PR that helped me complete [thisone](https://github.com/fastapi/sqlmodel/pull/1628). I would recommend to take a look to may simplify the logic a little with more typing definitions support and add more tests to each of them

YuriiMotov (approved):
LGTM!

This resolves issue https://github.com/fastapi/fastapi/issues/10719

The following type alias syntax will be supported:
```
type DependedValue = Annotated[int, Depends(some_value)]
```

Since we dropped the support of Python 3.8 we can now use `is_typealiastype` from `typing-inspection`. So, I updated PR to do so.

`typing-inspection` is already a sub-dependency of Pydantic, but I added it explicitly just in case (min version as in [Pydantic's pyproject.toml](https://github.com/pydantic/pydantic/blob/12c94d9753a77fe76111ef2161a33f5e2d5e41f3/pyproject.toml#L51C6-L51C30))

tiangolo (approved):
Awesome, thanks all! 🙌 

This will be available in FastAPI 0.128.2, released in the next few hours. 🚀

cstruct:
Thank you to everyone who helped push this through! I apologize that I dropped the ball keeping the PR up-to-date, and I'm grateful to @YuriiMotov for pushing fixes and keeping it up-to-date!

PR #13922, opened 2025-07-24
Improve parameter type inference for Depends and Security

mdev-new:
This pull request improves the final type inference for the Depends and Security functions.

```python
@router.post("/login")
def login(
  database = Depends(get_database)
):
  pass
```

Previously, the database parameter had the default `Any` type. Now, with this change, it inherits the return type of the `get_database` function.

### DIFF
--- a/fastapi/param_functions.py
+++ b/fastapi/param_functions.py
@@ -2217,9 +2217,9 @@ def File(  # noqa: N802
     )
 
 
-def Depends(  # noqa: N802
+def Depends[T](  # noqa: N802
     dependency: Annotated[
-        Optional[Callable[..., Any]],
+        Optional[Callable[..., T]],
         Doc(
             """
             A "dependable" callable (like a function).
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@mdev-new, thanks for your interest!

There are already a couple of PRs that aim to address this:
* https://github.com/fastapi/fastapi/pull/11255
* https://github.com/fastapi/fastapi/pull/5856
* https://github.com/fastapi/fastapi/pull/10053

I suggest we close this.
It's recommended now to use the approach with `Annotated`:

```python
@router.post("/login")
def login(
  database: Annotated[DBSession, Depends(get_database)],
):
  pass
```

YuriiMotov:
Since one more similar PR [was closed](https://github.com/fastapi/fastapi/pull/11255#issuecomment-3121580194) recently, I think we should close this to avoid wasting time.

> the recommended approach now is to use Annotated, this solves this and other issues, so it should be better.
> 
> There's also other case of class dependencies, where Depends() is empty, and having these stricter type annotations probably conflicts with that, so I'll pass on this one for now. 🍰

@mdev-new, thanks for your efforts!

PR #13929, opened 2025-07-26
🌐 Add German translations for recent new pages (using LLMs)

tiangolo:
🌐 Add German translations for recent new pages (using LLMs)

While reviewing these files, instead of suggesting specific changes and tweaks, it's better to tweak the LLM prompt (https://github.com/fastapi/fastapi/pull/13931) to make the LLM generate better translations.

PR #13930, opened 2025-07-26
🌐 Update German translations for recently updated docs (using LLMs)

tiangolo:
🌐 Update German translations for recently updated docs (using LLMs)

This is done with the German LLM prompt file (https://github.com/fastapi/fastapi/pull/13931), if there are things that are wrong in this PR, things that shouldn't be updated or things missing, it would be better to tweak that prompt file to instruct the LLM to better handle those cases than to review this PR directly.

---

These are the existing translations that have been updated in the original files in English since the last update in German.

These are probably the most delicate, because the prompt is adjusted to try both:

* Update the existing translation to cover whatever is new in the original docs in English.
* Keep what's already fine, avoid just rewriting everything every time, otherwise this could run over and over again generating something completely new each time, and any update would make the diff giant, instead of focusing on what actually changed.

Both objectives are at odds, one would expect the LLM to change more things, the other would expect the LLM to change fewer things.

There are some pages that have changed quite a bit in the original in English since they were translated, so I imagine there's a chance the LLM would end up defaulting to changing more in those files.

There are also some files that have changed only in specific sections, for example adding or removing things at the end. It seems the LLM can get these fairly well, only adding (or removing) that final section but respecting the rest.

If there are some files that are changed too much unnecessarily, or not working well, maybe a note with which files are okay to merge would work, so that I can remove the bad ones to tweak them further and keep the good translation files.


nilslindemann (approved):
Good 👍

nilslindemann:
@YuriiMotov, I can add the anchors in the rest of the documents too, if you want. Edit: already did, see [here](https://github.com/fastapi/fastapi/pull/13968#issuecomment-3156809552) Edit 2: this comment can be marked as outdated.

docs/de/docs/index.md
```diff
@@ -434,41 +439,63 @@ Für ein vollständigeres Beispiel, mit weiteren Funktionen, siehe das <a href="
...
 * Viele zusätzliche Funktionen (dank Starlette) wie:
     * **WebSockets**
-    * extrem einfache Tests auf Basis von `httpx` und `pytest`
+    * extrem einfache Tests auf Basis von HTTPX und `pytest`
     * **CORS**
     * **Cookie Sessions**
     * ... und mehr.
 
-## Performanz
+## <a id="performance"></a> Performanz
```
YuriiMotov:
```suggestion
## Performanz {% raw %}{#performance}{% endraw %}
```

nilslindemann (approved):
This is good 👍 and can be merged as is, if you want, I will add the permalinks in a later PR.

nilslindemann:
#13931, #13929, and #13930 (this one) can be closed, as I have applied them in #14015 and their changes will be committed when that one is committed. Also, I have resolved merge conflicts there already.

PR #13934, opened 2025-07-27
🌐Add Russian translations with missing pages (using LLMs)

tiangolo:
🌐Add Russian translations with missing pages (using LLMs)

This uses the LLM prompt in https://github.com/fastapi/fastapi/pull/13936

Instead of asking for specific changes in these files, it would be better to tweak that LLM prompt file to make the LLM generate better output.

For example, for Spanish, there are instructions to translate certain terms in certain way: https://github.com/fastapi/fastapi/blob/master/docs/es/llm-prompt.md


YuriiMotov:
I'll start new PR and re-generate translations with new prompt

PR #13938, opened 2025-07-27
feat: add comprehensive custom API tests

gengbo123456:
- Add TestCustomAPI class with multiple test scenarios
- Include parametrized tests for different item IDs
- Add fixture examples for reusable test data
- Test both valid and invalid input validation
- Cover GET, POST endpoints with success/error cases

### DIFF
--- a/tests/test_custom_api.py
+++ b/tests/test_custom_api.py
@@ -2,12 +2,13 @@
 自定义API测试
 这是一个示例测试文件，展示如何在FastAPI项目中编写测试
 """
+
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@gengbo123456, hello!
Thanks for your interest!

Could you please clarify why you think we need to add these tests. Which of these tests are missed in current version?

Second, we need to use only English language. All comments must be in English.

If this PR is not finished yet (work in progress), let's convert it into Draft for now

PR #13935, opened 2025-07-27
🌐Update Russian translations that were changed after the last commit (ussing LLMs)

tiangolo:
🌐Update Russian translations that were changed after the last commit (ussing LLMs)

This uses the LLM prompt in https://github.com/fastapi/fastapi/pull/13936

Instead of asking for specific changes in these files, it would be better to tweak that LLM prompt file to make the LLM generate better output.

For example, for Spanish, there are instructions to translate certain terms in certain way: https://github.com/fastapi/fastapi/blob/master/docs/es/llm-prompt.md

PR #13941, opened 2025-07-28
🌐 Update translations for ru

tiangolo:
🌐 Update translations for ru


YuriiMotov:
I think we don't need this anymore, will continue work in https://github.com/fastapi/fastapi/pull/13935

Issue #10180, opened 2023-08-31
Mounting sub-applications under APIRouter


Kludex:
### Discussed in https://github.com/tiangolo/fastapi/discussions/8682

<div type='discussions-op-text'>

<sup>Originally posted by **chriswhite199** November 19, 2021</sup>
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from fastapi import FastAPI, APIRouter
from starlette.testclient import TestClient

app = FastAPI()
api_router = APIRouter(prefix="/api")


@api_router.get("/app")
def read_main():
    return {"message": "Hello World from main app"}


app.include_router(api_router)

subapi = FastAPI()


@subapi.get("/sub")
def read_sub():
    return {"message": "Hello World from sub API"}


api_router.mount("/subapi", subapi)

client = TestClient(app)

assert client.get('/api/app').status_code == 200

# this next assert fails
assert client.get('/api/subapi/sub').status_code == 200
```

### Description

Is it possible to mount a sub-application under an APIRouter? APIRouter itself has a mount function and accepts similar arguments to mounting a sub-application on a FastAPI instance, but I can't get the routing to actually work (nor can i get the openapi docs or spec to come back from that I would assume are the correct URLs.

The docs for sub applications note that the sub-application will have it's root_path correctly set, and I've tried a few combinations of manually setting the root_path on the subapi instance, but to no avail.

### Operating System

macOS

### FastAPI Version

0.70.0

### Python Version

3.9.7

### Additional Context

_No response_</div>

achaayb:
I dont see a reason of binding two FastAPI instances, perhaps in your usecase you're trying to merge two applications, in this case you should have a service that extracts endpoints and mounts them into an APIRouter

[...]

PR #13944, opened 2025-07-28
✨ Support mounting `StaticFiles` to sub-routers

YuriiMotov:
For now if you try mounting the sub-application to `APIRouter` it will be silently ignored.

See issue: https://github.com/fastapi/fastapi/issues/10180

This PR makes it working for `StaticFiles` only. At the same time mounting other apps to APIRouter will raise `FastAPIError` with clear error message:

```
APIRouter does not support mounting ASGI applications other than StaticFiles.
```

For non-`StaticFiles` sub-apps it's recommended to mount them to root app instead of router.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -1133,10 +1133,8 @@ class FastAPI(Starlette):
             scope["root_path"] = self.root_path
         await super().__call__(scope, receive, send)
 
-
     def mount(self, path: str, app: ASGIApp, name: str | None = None) -> None:
         self.router._mount(path, app=app, name=name)
-
 
     def add_api_route(
         self,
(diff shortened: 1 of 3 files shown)


mrcghil:
Having bumped on this silent bug today I think it is a pity that this PR has been sitting here for so long! Anything we can do to push this through?
Best,
Marco

YuriiMotov:
Updated PR to support mounting `StaticFiles` to sub-routers instead of deprecating `APIRouter.mount()` completely

BharatShukla30:
I tested a few additional scenarios:

- multiple mounts on the same router (/static and /assets)
- nested routers (/api + /v1)
- mount path with trailing slash (/static/)

In all cases the composed paths were correctly propagated to the final app routing table. The implementation and tests look consistent with the intended behavior.

tiangolo:
I prefer to delay this until after the big router refactor as that might interact with how this works, and supporting this now would mean needing to keep backwards compatibility with it, so I'll pass on this for now. :coffee:

PR #13946, opened 2025-07-29
✅ Run tests on free threaded python 3.14

patrick91:


svlandeg:
Had a look into this.

I suggest to skip `3.13t` which was in beta and not supported by several libraries anyway. For instance, `trio` only supports `3.14t` (and up), and `pydantic-core` also [removed testing](https://github.com/pydantic/pydantic-core/pull/1738) for `3.13t` .

To be able to run our test suite on `3.14t` in the CI, as far as I can see, we'd need 

- [x] `httptools` to release wheels built against Cython 3.1+ (I verified that it installs fine [locally](https://github.com/MagicStack/httptools/issues/118#issuecomment-3165444672))
- [x] `pydantic` 2.12.0 to be released with 314t wheels
- [x] `bcrypt` <s>to release [3.14 support](https://github.com/pyca/bcrypt/pull/1073)</s> -> 5.0.0 is now released with free-threaded Python 3.14 support
- [ ] `orjson` to start supporting free threading (some relevant discussion [here](https://github.com/ijl/orjson/issues/591)) 

I'll open a PR to start testing fastapi on (normal) Python 3.14 so we can work on that separately: <s>https://github.com/fastapi/fastapi/pull/14110</s> https://github.com/fastapi/fastapi/pull/14165

CC @ngoldbaum: as I've seen many contributions from you on various libraries on this topic, does the above ☝️  sound about right? Any other feedback or pointers is appreciated 🙏

ngoldbaum:
> does the above ☝️ sound about right? Any other feedback or pointers is appreciated 🙏

`httptools` isn't a blocker for me personally so I haven't touched it yet. I think it should be straightforward to update if anyone wants to take that on. It'd be really nice to add free-threaded wheel builds to their build automation.

Both pydantic and bcrypt should have releases with 3.14t support soon, I think.

orjson is trickier, I don't think there will be free-threaded support before 3.15. I know that pydatic-core uses https://github.com/pydantic/jiter rather than orjson. I'm not sure how tightly coupled fastapi is to orjson.

svlandeg (commented):
This currently runs when explicitely disabling the GIL, which we need to do for `orjson` & `ujson` for now, cf also https://github.com/ultrajson/ultrajson/pull/681/ and https://github.com/ijl/orjson/issues/591#issuecomment-3145838844.

svlandeg:
This PR has gone pretty stale and much has changed since I last worked on it, so I'm continuing this work in #15149.

PR #13951, opened 2025-07-30
⚡️ Use `TypeAdapter.validate_json` instead of `json.loads`

dolfinus:
#14962 replaced `json.dumps` with pydantic v2 `TypeAdapter.serialize_json()`. This is the same, but for body parsing - replace `json.loads` with `TypeAdapter.validate_json()`.
See also https://github.com/fastapi/fastapi/discussions/13949

Now this gives only a moderate speedup, nothing huge. The larger the request body, the larger the speedup.
Currently optimization can be applied only to POST handler with just one body field.

<details>

<summary>Small benchmark</summary>

requirements.txt

```pip
fastapi
locust
py-spy
```

app.py
```python
from http import HTTPStatus

app = FastAPI()

class SimpleModel(BaseModel):
    a: str


@app.post("/data")
async def create_item(data: SimpleModel):
    return Response(status_code=HTTPStatus.NO_CONTENT)

if __name__ == "__main__":
    import uvicorn
```

locustfile.py
```python
import json

from locust import FastHttpUser, TaskSet, task


class FastAPITask(TaskSet):
    data = json.dumps({"a": "abc" * 10000})
    headers = {
        "Content-Type": "application/json",
    }

    @task
    def push_data(self):
        with self.user.rest(
            "POST",
            "data",
            data=self.data,
            headers=self.headers,
            name="/data",
        ) as response:
            response.js


class FastAPIUser(FastHttpUser):
    host = "http://localhost:8000/"
    tasks = [FastAPITask]

```

```bash
locust --web-host=0.0.0.0 --processes 4 FastAPIUser
```

```bash
record --duration 600 --output flamegraph.svg -- python -m uvicorn app:app --host 0.0.0.0 --port 8000 --log-level=warning
```

</details>

Before - 2228RPS, `json.loads` took 9.8% and `type_adapter.validate_python` took 1% = 11% combined
[fastapi_with_json_loads.tar.gz](https://github.com/user-attachments/files/25499087/fastapi_with_json_loads.tar.gz)
fastapi_with_json_loads

After - 2570RPS, `type_adapter.validate_json` took 7.17%
[fastapi_with_typeadapter_json.tar.gz](https://github.com/user-attachments/files/25499127/fastapi_with_typeadapter_json.tar.gz)
fastapi_with_typeadapter_json

Caveats - if route handle contains multiple fields with Body() annotation, json will be parsed for each field. Changing this requires substantial rewrite of body parsing, e.g. merging all body fields into one pydantic class and using it's `model_validate_json()` method for body parsing. I'm not ready for this type of change, so this can be considered as proof-of-concept.

Bump pydantic minimal version to 2.10.0 because of [this issue](https://github.com/pydantic/pydantic-core/pull/1479) with validate_json appears on versions below pydantic-core 2.24.1

### DIFF
--- a/docs/de/docs/tutorial/handling-errors.md
+++ b/docs/de/docs/tutorial/handling-errors.md
@@ -205,13 +205,11 @@ Sie erhalten eine Response, die Ihnen sagt, dass die Daten ungültig sind und di
         "size"
       ],
       "msg": "value is not a valid integer",
-      "type": "type_error.integer"
+      "type": "type_error.integer",
+      "input": "XL"
     }
   ],
-  "body": {
-    "title": "towel",
-    "size": "XL"
-  }
+  "body": "{\n  \"title\": \"towel\",\n  \"size\": \"XL\"\n}"
 }
 ```
 
--- a/docs/en/docs/tutorial/handling-errors.md
+++ b/docs/en/docs/tutorial/handling-errors.md
@@ -205,13 +205,11 @@ You will receive a response telling you that the data is invalid containing the 
         "size"
       ],
       "msg": "value is not a valid integer",
-      "type": "type_error.integer"
+      "type": "type_error.integer",
+      "input": "XL"
     }
   ],
-  "body": {
-    "title": "towel",
-    "size": "XL"
-  }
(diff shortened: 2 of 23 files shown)


dolfinus:
This bot is quite annoying... #15617

PR #13965, opened 2025-08-03
✨ Add `TypedState` special dependency

YuriiMotov:
This PR aims to solve the problem of `request.state._state` being untyped (it's `dict[str, Any]` now).

## Current situation

For now if we want to get an object from request state, we need to use `request.state._state` and cast extracted objects to get type hints.

```py
class MyState(TypedDict):
    param_1: str
    param_2: int

@app.get("/")
async def read_state(request: Request):
    state = cast(MyState, request.state._state)
    a = state["param_1"]
```

## With this PR

With changes implemented in this PR we can inject request state object into our endpoint function (or dependency function) with:

```py
from fastapi import TypedState

class MyState(TypedDict):
    param_1: str
    param_2: int

@app.get("/")
async def read_state(state: TypedState[MyState]):
    a = state._state["param_1"]  # <= You have type hints and autocompletions here
```

Full code example in details

<details>

```py
from typing import TypedDict

from fastapi import FastAPI, TypedState

class MyState(TypedDict):
    param_1: str
    param_2: int

@asynccontextmanager
async def lifespan(app: FastAPI):
    state = MyState(param_1="example", param_2=42)
    yield state

app = FastAPI(lifespan=lifespan)

@app.get("/")
async def read_state(state: TypedState[MyState]):
    return {"param_1": state._state["param_1"], "param_2": state._state["param_2"]}
```

</details>

And it will be typed. So, you will have type hints and autocompletion:

---

Screenshot from 2025-08-03 15-54-16

 ---

Screenshot from 2025-08-03 15-54-59

---

## Discussion

Let's discuss it in this discussion: https://github.com/fastapi/fastapi/discussions/13966


mahimairaja:
LGTM

wu-clan:
Will this change its direction? Please see：https://github.com/Kludex/starlette/pull/3036

YuriiMotov:
Closing this in favor of https://github.com/fastapi/fastapi/pull/14863

PR #13977, opened 2025-08-07
📝 Add `llms.txt` file for LLMs better understand

vldmrdev:
llms.txt is a proposed standard file that helps Large Language Models (LLMs) better understand and interact with your website's documentation and content structure. It's similar to robots.txt, but specifically designed for AI language models and in markdown format.

More details: https://llmstxthub.com/faq
References:  
Pydantic: https://docs.pydantic.dev/latest/llms.txt
Docker: https://docs.docker.com/llms.txt
Svetle: https://svelte.dev/llms.txt
and other https://llmstxthub.com/websites


vldmrdev:
Hi! Should I create an issue for this?

YuriiMotov:
@vldmrdev, thanks for your interest!

I think this file should be automatically generated and updated to make sure it's up to date.
So, we probably need to commit not `llms.txt` itself, but a script that generates\updates this `llms.txt`.

vldmrdev:
> So, we probably need to commit not llms.txt itself, but a script that generates\updates this llms.txt.

@YuriiMotov Hi, I think it is possible to do it based on this file https://github.com/fastapi/fastapi/blob/master/docs/en/mkdocs.yml, but:
1. How often new sections appear in the documentation and the short description changes? If not, perhaps it will be overengineering?
2. If yes, do all sections and subsections have a short description?
3. Also, I suppose, a short description should be highlighted syntactically in each file for parsing.

I have already made llms.txt for another project and in half of the cases I had to make the description manually, including because of the problems described above.

tiangolo:
Thanks for the interest @vldmrdev.

Indeed, as @YuriiMotov says, this would need to be autogenerated and account for several other things, so I'll pass on this for now. :coffee:

PR #13980, opened 2025-08-08
📝 Rename functions for params tutorial and minor grammar tweaks

nikmaxott:
The functions in the [query parameter tutorial](https://fastapi.tiangolo.com/tutorial/query-params/) were renamed to read_user_item and then never changed back when user_id is no longer a parameter

### DIFF
--- a/docs/en/docs/tutorial/body.md
+++ b/docs/en/docs/tutorial/body.md
@@ -10,7 +10,7 @@ To declare a **request** body, you use <a href="https://docs.pydantic.dev/" clas
 
 /// info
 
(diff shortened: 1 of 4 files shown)


nikmaxott:
Should we be using the new .model_dump() instead of the deprecated .dict() in the tutorial?

tiangolo:
Thanks for the interest! This PR has several things at once. For the textual rewordings, I would prefer to pass on them.

The rename of the function names, I could take it, but it would have to be a separate PR, with the tests passing. :sweat_smile: 

So, for now, I'll pass on this one. But thanks for the interest! :coffee:

PR #13982, opened 2025-08-09
⚡️ Refactor `jsonable_encoder` with performance bump

Ben-Brady:
Continuation of https://github.com/fastapi/fastapi/pull/5462

- Moved implemntation to `encode_value` function to avoid reprocession parameters each time
- Moved dict encoding to `encode_dict`
- Added NoneType constant
- Replaced dataclass and final catchall `jsonable_encoder` calls with `encode_dict` as they're guarenteed to be `dict`
- Removed explicit PurePath, UndefinedType, Enum checks and put them in ENCODERS_BY_TYPE similar to other misc types.

# Benchmarks

| Before | After |
| - | - |
| image | image |


fastapi/encoders.py
```diff
@@ -230,114 +290,105 @@ def jsonable_encoder(
...
             include=include,
             exclude=exclude,
             by_alias=by_alias,
             exclude_unset=exclude_unset,
-            exclude_defaults=exclude_defaults,
             exclude_none=exclude_none,
             custom_encoder=custom_encoder,
             sqlalchemy_safe=sqlalchemy_safe,
         )
-    if isinstance(obj, Enum):
```
Ben-Brady, line 257: Enum was already contained in ENCODERS_BY_TYPE, so this was never needed

fastapi/encoders.py
```diff
@@ -230,114 +290,105 @@ def jsonable_encoder(
...
+    allowed_keys = set(obj.keys())
+
+    if include is not None:
+        allowed_keys &= set(include)
+
+    if exclude is not None:
+        allowed_keys -= set(exclude)
+
+    for key, value in obj.items():
+        if sqlalchemy_safe and isinstance(key, str) and key.startswith("_sa"):
```
Ben-Brady: Replaced the original complicated boolean expression with multiple continue statemenets
original for comparsion:
```py
if (
    (
        not sqlalchemy_safe
        or (not isinstance(key, str))
        or (not key.startswith("_sa"))
    )
    and (value is not None or not exclude_none)
    and key in allowed_keys
):
```

Ben-Brady: Although the _sa check was the original behaviour, I believe it would be more valid for the _sa to be checked after encoding the key? As it may be possible a user would define a custom encoder that would generate a key with a "_sa" prefix

tests/test_jsonable_encoder.py
```diff
@@ -88,6 +88,42 @@ def test_encode_dict():
...
+    assert jsonable_encoder(
+        {CustomString("foo"): "bar"},
+        custom_encoder={CustomString: lambda v: v.value}
+    ) == {"foo": "bar"}
+
+def test_encode_dict_with_custom_encoder_keys():
+    assert jsonable_encoder(
+        {"foo": "bar"},
+        custom_encoder={str: lambda v: "_" + v}
+    ) == {"_foo": "_bar"}
```
Ben-Brady, line 109: This test was added as my original version had a bypass for primative key types in `encode_dict`, this gave a 10~20% the performance bump however is incorrect behaviour as it ignores a custom_encoder. This test was added to ensure that the behaviour is not accidentally broken.

fastapi/encoders.py
```diff
@@ -230,114 +290,105 @@ def jsonable_encoder(
...
     if type(obj) in ENCODERS_BY_TYPE:
         return ENCODERS_BY_TYPE[type(obj)](obj)
+
     for encoder, classes_tuple in encoders_by_class_tuples.items():
         if isinstance(obj, classes_tuple):
             return encoder(obj)
 
     try:
-        data = dict(obj)
+        obj_dict = dict(obj)
```
Ben-Brady, line 308: renamed to obj_dict to be more consistent

Ben-Brady:
also here's the benchmark I used to test it, based on an application I made a while ago

```

from pydantic import BaseModel as _PydanticBaseModel, Field
from enum import Enum
import time

class BaseModel(_PydanticBaseModel):
    class Config:
        json_encoders = { bytes: lambda _bytes: _bytes.hex() }


class MediaType(str, Enum):
    image = "image"
    animation = "animation"
    video = "video"


class Rating(str, Enum):
    unrated = "unrated"
    safe = "safe"
    sensitive = "sensitive"
    mature = "mature"
    explicit = "explicit"


class BaseMedia(BaseModel):
    url: str = Field(...)
    mimetype: str = Field(...)
    height: int = Field(...)
    width: int = Field(...)
    type: MediaType


class Image(BaseMedia):
    type: MediaType = Field(default=MediaType.image)


class Animation(BaseMedia):
    duration: float = Field(...)
    frame_count: int = Field(...)
    duration:float = Field(...)
    type:MediaType = Field(default=MediaType.animation)


class Video(BaseMedia):
    has_sound: bool = Field(...)
    duration: float = Field(...)
    fps: str = Field(...)
    type:MediaType = Field(default=MediaType.video)


class Hashes(BaseModel):
    md5s: list[bytes] = Field(default_factory=list)
    sha256s: list[bytes] = Field(default_factory=list)
    phashes: list[bytes] = Field(default_factory=list)


GenericMedia = Image | Animation | Video


class Post(BaseModel):
    id: int = Field(..., lt=2**64)
    created_at: float = Field(default_factory=time.time)
    uploader: int|None = Field(default=None)
    deleted: bool = Field(default=False)
    source: str = Field(default="")
    rating: Rating = Field(default=Rating.unrated)

    full: GenericMedia = Field(...)
    preview: GenericMedia|None = Field(default=None)
    thumbnail: Image = Field(...)

    media_type: MediaType = Field( ...)
    hashes: Hashes = Field(default_factory=Hashes)

    tags: list[str] = Field(default_factory=list)
    comments: list[int] = Field(default_factory=list)

    upvotes: int = Field(default=0)
    downvotes: int = Field(default=0)

sample_data = {"id": 3377992785,"created_at": 1666691800.2500513,"uploader": 1552510247,"deleted": False,"source": "https://twitter.com/AdvosArt/status/1584520218507759616","rating": "unrated","full": {"url": "https://api.openbooru.org/media/pK5zTCOH30.webp","mimetype": "image/webp","height": 1024,"width": 1024,"type": "image"},"preview": {"url": "https://api.openbooru.org/media/35-uazKb9L.webp","mimetype": "image/webp","height": 750,"width": 750,"type": "image"},"thumbnail": {"url": "https://api.openbooru.org/media/GchBJhBbDU.webp","mimetype": "image/webp","height": 180,"width": 180,"type": "image"},"media_type": "image","hashes": {"md5s": ["56dcedb90a8fb6ad71a860eedc0ff45b","d23734ecbbcbdd188a7517dbed5cdbeb","7a2ad52ad62bb072deb581ffd47db3f6","48c965935f573a611b739688a7c18ccb"],"sha256s": ["0bbb7631d0c623f20c9aec346e65abe5272c03fa7ec68dd59771c86f21beed01","1dcd78fe0fb5c32f4caeb93c13773f3121acd15b1c74f53ee1f134c08a14870a","dae6dc9c6928c9f2cfb79a2f4652593e28922b98f186b817295945634312af24","8afff20da8b889ced57945505394db037b5f4505efbcd091323c57c2419cfbbe"],"phash": [ ]},"tags": ["image","advosart"],"comments": [ ],"edits": [ ],"upvotes": 0,"downvotes": 0}
post = Post.model_validate(sample_data)

for _ in range(50_000):
    jsonable_encoder(post)
```

fastapi/encoders.py
```diff
@@ -230,114 +294,105 @@ def jsonable_encoder(
...
+
+    if exclude is not None:
+        allowed_keys -= set(exclude)
+
+    for key, value in obj.items():
+        if sqlalchemy_safe and isinstance(key, str) and key.startswith("_sa"):
+            continue
+        if value is None and exclude_none:
+            continue
+        if key not in allowed_keys:
```
dolfinus: Instead of iterating all object keys and rejecting some, can we just iterate allowed keys in the firdt place?

Ben-Brady: good point I didn't spot that

Do you recon this approach is better then?
```py
for key in allowed_keys:
    if key not in obj:
        continue

    value = obj[key]
```

dolfinus: There is a line at the beginning of this function:
```python
allowed_keys = set(obj.keys())
```
So this is redundant

Ben-Brady: Actually I don't think this is an improvement. I think it's slightly worse to read, and also the performance isn't any better than a slightly tweaked version.

Although it did initially show some performance improvements, if I moved the `if key not in allowed_keys` check to before the others (since it's cheaper than the `isinstance` check) I actually got better performance (~1%) than iterating over the allowed_keys.

I've implemented the re-arranged check

dolfinus: I see, `key, value in dict.items()` is faster than `value = dict[key]`, as it doesn't need to hash key + check if key is present

fastapi/encoders.py
```diff
@@ -201,25 +205,85 @@ def jsonable_encoder(
...
+    custom_encoder: Optional[Dict[Any, Callable[[Any], Any]]] = None,
+    sqlalchemy_safe: bool = True,
+) -> Any:
+    if custom_encoder:
+        if type(obj) in custom_encoder:
+            return custom_encoder[type(obj)](obj)
+
+        for encoder_type, encoder_instance in custom_encoder.items():
+            if isinstance(obj, encoder_type):
+                return encoder_instance(obj)
```
dolfinus:
```suggestion
        # fastpath for exact class match
        obj_type = type(obj)
        if obj_type in custom_encoder:
            return custom_encoder[obj_type](obj)

        # fallback to isinstance which uses MRO
        for encoder_type, encoder_instance in custom_encoder.items():
            if isinstance(obj, encoder_type):
                return encoder_instance(obj)
```
This also can be changed at the bottom of this method

dolfinus (approved):

fastapi/encoders.py
```diff
@@ -201,25 +205,88 @@ def jsonable_encoder(
...
+
+        # fallback to isinstance which uses MRO
+        for encoder_type, encoder_instance in custom_encoder.items():
+            if isinstance(obj, encoder_type):
+                return encoder_instance(obj)
+
+    if isinstance(obj, (str, int, float, NoneType)):  # type: ignore[arg-type, misc]
+        return obj
+
+    if isinstance(obj, (list, set, frozenset, GeneratorType, tuple, deque)):
```
dolfinus, line 226: These hardcoded values can be excluded from ENCODERS_BY_TYPE, to make less iterations

fastapi/encoders.py
dolfinus: Perhaps this was an optimization because enums are used more frequently than other entries in ENCODERS_BY_TYPE

fastapi/encoders.py
```diff
@@ -201,25 +205,88 @@ def jsonable_encoder(
...
+    exclude_defaults: bool = False,
+    exclude_none: bool = False,
+    custom_encoder: Optional[Dict[Any, Callable[[Any], Any]]] = None,
+    sqlalchemy_safe: bool = True,
+) -> Any:
+    if custom_encoder:
+        # fastpath for exact class match
+        obj_type = type(obj)
+        if obj_type in custom_encoder:
+            return custom_encoder[obj_type](obj)
```
dolfinus:
```suggestion
        encoder = custom_encoder.get(type(obj))
        if encoder:
            return encoder(obj)
```
To avoid getting a type hash twice.

Same on lines 326-327

Ben-Brady:
> Perhaps this was an optimization because enums are used more frequently than other entries in ENCODERS_BY_TYPE

I tried benchmarking it and it only makes a very slight difference

### No Explicit Check

image 

### Explicit Check

image 

I can add the check back in, but I don't think it's worth it for such a small performance bump

fastapi/encoders.py
Ben-Brady: I've implemented this. Also since this code is duplicated for both the ENCODERS_BY_TYPE and custom_encoders, I've moved it into it's own function.
I also benchmarked it to check if it was affect performance and it has a negligible impact to performance

dolfinus (approved):

fastapi/encoders.py
```diff
@@ -230,114 +287,118 @@ def jsonable_encoder(
...
     if dataclasses.is_dataclass(obj):
         obj_dict = dataclasses.asdict(obj)
-        return jsonable_encoder(
+        return encode_dict(
             obj_dict,
             include=include,
             exclude=exclude,
             by_alias=by_alias,
             exclude_unset=exclude_unset,
-            exclude_defaults=exclude_defaults,
```
YuriiMotov, line 252: Could you please clarify why you removed this parameter?
Seems that it may break the logic if we have dataclass with Pydantic model as nested field:

```py
class A(BaseModel):
    a: str = "123"

@dataclasses.dataclass
class B:
    sub: A

obj = B(sub=A(a="123"))
obj_dict = dataclasses.asdict(obj)
# {'sub': A(a='123')}
```

fastapi/encoders.py
```diff
@@ -230,114 +287,118 @@ def jsonable_encoder(
...
-    return jsonable_encoder(
-        data,
+
+    return encode_dict(
+        obj_dict,
         include=include,
         exclude=exclude,
         by_alias=by_alias,
         exclude_unset=exclude_unset,
-        exclude_defaults=exclude_defaults,
```
YuriiMotov, line 340: Could you please clarify why you removed this parameter?
Seems that it may break the logic if we have nested Pydantic model

fastapi/encoders.py
```diff
@@ -98,6 +96,8 @@ def generate_encoders_by_class_tuples(
 
 encoders_by_class_tuples = generate_encoders_by_class_tuples(ENCODERS_BY_TYPE)
```
YuriiMotov: We don't need `generate_encoders_by_class_tuples` and `encoders_by_class_tuples` anymore

YuriiMotov (commented):
@Ben-Brady, thank you for working on this!

Could you please take a look at my comments?

Also, I think we should refactor this PR to make it easier to review.
For now diff is not very informative and it takes a lot of time to review it carefully.
I think I can help with this. Let's discuss my questions first

fastapi/encoders.py
Ben-Brady: I think this was a bug in the original implementation that I've copied over.

So the two places that are guarenteed to be `dict`s, here and at the end, I've swapped over to encode_dict calls.

The diff makes it hard to read, but the original implementation for encoding a dictionary was this:

```py
for key, value in obj.items():
    if (...):
      encoded_key = jsonable_encoder(
        key,
        by_alias=by_alias,
        exclude_unset=exclude_unset,
        exclude_none=exclude_none,
        custom_encoder=custom_encoder,
        sqlalchemy_safe=sqlalchemy_safe,
    )
    encoded_value = jsonable_encoder(
        value,
        by_alias=by_alias,
        exclude_unset=exclude_unset,
        exclude_none=exclude_none,
        custom_encoder=custom_encoder,
        sqlalchemy_safe=sqlalchemy_safe,
    )
    encoded_dict[encoded_key] = encoded_value
```

For some reason it doesn't pass down the `exclude_defaults` parameter, so I tried to preserve that behavior to avoid it being a breaking change. But if it is a bug, I can just add it back in.

fastapi/encoders.py
Ben-Brady: see previous comment

fastapi/encoders.py
Ben-Brady: yup removed, forgot to take it out when I changed the resolve logic

fastapi/encoders.py
YuriiMotov: I suggest we leave parts about `dict`s as it is (and as it was), but add this parameter back here (for dataclass) and [there](https://github.com/fastapi/fastapi/pull/13982/files#r2334497197).
So, we will have the same behavior (regarding passing the `exclude_defaults` parameter) as it was before changes.
We will think about fixing this for `dict`s later

Ben-Brady: The current behavior applies to both dataclasses and fallback encoding. Since those although  those originally called `jsonable_encoder` with the `exclude_defaults` parameter. But they were guaranteed to be `dict`, so the parameter would always be ignored.

So I can't really change it to so that the dataclass and fallback encoding use the `exclude_defaults`, because even if they did it would be ignored.

YuriiMotov: I think that not passing `exclude_defaults` in case of `dict`s is a bug and I think we will fix it later. Do you agree with this?

Assuming you agree with the statement above, I believe we should revert removing the `exclude_defaults` in case of dataclass and fallback. As you correctly noted (I didn't noticed it previously), It will be ignored for now, but it will be in line with previous implementation and it will start working correctly after we fix the issue with not passing `exclude_defaults` for `dict`s.

Ben-Brady: Yeah that makes sense, but I think we should add an exclude_defaults parameter to encode_dict, even if it's not used yet.
When we fix the bug, it'll only be the encode_dict function that'll have to change, not any of it's callers. Also the reason it uses encode_dict here instead of encode_value, is because the type is guaranteed so we can skip all the type checks.

Sorry, if I've been a little argumentative on this, I hadn't initially realised this was a bug in the original.

YuriiMotov: Agree, let's add the `exclude_defaults` parameter to `encode_dict`

Ben-Brady: Done, it's implemented.

Ben-Brady:
> @Ben-Brady, thank you for working on this!
> 
> Could you please take a look at my comments?
> 
> Also, I think we should refactor this PR to make it easier to review. For now diff is not very informative and it takes a lot of time to review it carefully. I think I can help with this. Let's discuss my questions first

If you need the diff clean to review properly, I would need some help on how to do that as it's something I've never done that before

fastapi/encoders.py
```diff
@@ -201,25 +185,83 @@ def jsonable_encoder(
     Read more about it in the
     [FastAPI docs for JSON Compatible Encoder](https://fastapi.tiangolo.com/tutorial/encoder/).
     """
-    custom_encoder = custom_encoder or {}
```
YuriiMotov, line 204: I think we should return this line back and then change the type of `custom_encoder` parameter to `encode_value` and `encode_dict` to be non-optional `Dict` - I think this will be more type-safe

YuriiMotov (approved):
LGTM!

These changes give ~50% performance increase with the test provided in [the comment](https://github.com/fastapi/fastapi/pull/13982#issuecomment-3170658318).
The real performance increase will definitely depend on the input data and can be lower. We should probably test it with different input data.

I created a [branch](https://github.com/fastapi/fastapi/compare/master...YuriiMotov:fastapi:Ben-Brady_master_2) with the copy of these changes, where I broke down changes into logical peaces, every peace is in separate commit. Should be easier to review commit-by-commit.
Total diff of those commits is equal to the diff of this PR except one thing I [suggested to change in review comment](https://github.com/fastapi/fastapi/pull/13982/files#r2419265228).

Commits:
* [commit 1](https://github.com/fastapi/fastapi/commit/d68d7ec39c3590dd683a5eb26b1c539ab0de3836) Avoid repeating the input parameters coercion (5% performance improvement)
* [commit 2](https://github.com/fastapi/fastapi/commit/5ab3bb5e8f6cd7c6f7801d332d44957c9e42e2ef) Introduce `encode_dict` and use it when object is `dict` (`dict`, `dataclass`, default catchall) (just little performance improvement)
* [commit 3](https://github.com/fastapi/fastapi/commit/36b0a1aae2a95d43396e2d3497ce0bc6b664e0cc) Encapsulate custom encoder resolution logic in `find_encoder` (it's just refactoring)
* [commit 4](https://github.com/fastapi/fastapi/commit/ed8fa7071c0258e29ba1e008a452d65a94a1deb8) Re-order IF-blocks in `encode_value` (this brings the biggest performance improvement - about 25-30%)

<details><summary>Simplified diff for "Re-order IF-blocks" in the details</summary>

```diff
    if custom_encoder:
        ...

-    if isinstance(obj, BaseModel):
-        ...
-    if dataclasses.is_dataclass(obj):
-        ...
    if isinstance(obj, Enum):
        return obj.value
    if isinstance(obj, PurePath):
        return str(obj)
    if isinstance(obj, (str, int, float, type(None))):
        return obj
    if isinstance(obj, UndefinedType):
        return None
-    if isinstance(obj, dict):
-        return encode_dict(...)
    if isinstance(obj, (list, set, frozenset, GeneratorType, tuple, deque)):
        ...
+    if isinstance(obj, dict):
+        return encode_dict(...)
+    if isinstance(obj, BaseModel):
+        ...
+    if dataclasses.is_dataclass(obj):
+        ...

    encoder = find_encoder(obj, ENCODERS_BY_TYPE)
    if encoder:
        return encoder(obj)

    try:
        data = dict(obj)
    except Exception as e:
        ...
    return encode_dict(...)
```

</details>

* [commit 5](https://github.com/fastapi/fastapi/commit/91cc20baa4756d9047c27a3e64c2ed6cd33094cb) Move `PurePath` and `UndefinedType` to `ENCODERS_BY_TYPE` (this gave ~7% performance improvement)
* [commit 6](https://github.com/fastapi/fastapi/commit/9fcd1c346416eea9ddd3b00cb52e4d9f5304e0aa) Remove `frozenset`, `deque`, `GeneratorType`, `set` from `ENCODERS_BY_TYPE` (just a clen-up - we have check them explicitly earlier)
* [commit 7](https://github.com/fastapi/fastapi/commit/9fcd1c346416eea9ddd3b00cb52e4d9f5304e0aa) Remove explicit `isinstance(obj, Enum)`, refactor `obj is None` check (mostly refactoring)
* [commit 8](https://github.com/fastapi/fastapi/commit/68595b5827f455a69dd05c178ef9c1689d686619) Some refactoring, doesn't change logic
* [commit 9](https://github.com/fastapi/fastapi/commit/d0f8f7ae07ee49e446a04118d0437c2e0880f6f9) update the type of `custom_encoder` parameter in `encode_dict` (make it non-optional)

YuriiMotov:
One more thing - we should probably make those new internal functions private

wu-clan:
Can we push this forward quickly? Why not?

YuriiMotov:
> Can we push this forward quickly? Why not?

Unfortunately it's not so easy..

As I see the plan is:
* Wait for Sebastian to validate the idea
* When idea is accepted, we will need to resolve merge conflicts and update the set of commits [I created previously](https://github.com/fastapi/fastapi/pull/13982#pullrequestreview-3322502720) (this will simplify the final review by Sebastian)
* Then we will ask Sebastian to review this PR and hopefully merge it

nidhishgajjar:

## Summary

This PR refactors `jsonable_encoder` for performance with documented benchmarks showing improvement. However, the scope is massive (2,885 files changed, 184,317 deletions) and includes significant breaking changes including removal of Pydantic v1 compatibility and deletion of documentation files. While the performance work appears valuable, the PR combines multiple breaking changes that should be separated.

## Architecture

**Stated Purpose:** Refactor jsonable_encoder with performance bump

**Actual Scope:**
- 2,885 files changed (95,513 insertions, 184,317 deletions)
- **Performance refactor:** Refactored `fastapi/encoders.py` with benchmarked improvements
- **Breaking changes:** 
  - Removed Pydantic v1 compatibility (`fastapi/_compat/` directory deleted)
  - Removed documentation files (multiple .md files deleted)
  - Removed GitHub workflow files
- **Net loss:** 88,804 lines of code

## Issues

### Critical Issues

**Massive Scope - Multiple Breaking Changes Combined:**

This PR combines at least three separate changes:
1. Performance refactor of jsonable_encoder (legitimate feature)
2. Removal of Pydantic v1 compatibility (breaking change)
3. Documentation cleanup (potentially breaking)

**Impact:** Impossible to review properly. Each change should be a separate PR.

**Pydantic v1 Compatibility Removed:**

```python
# Removed imports:
- from fastapi._compat import Url, is_pydantic_v1_model_instance
- from fastapi.exceptions import PydanticV1NotSupportedError
- from pydantic_core import PydanticUndefinedType

# Deleted entire _compat directory:
D	fastapi/_compat/__init__.py
D	fastapi/_compat/shared.py
D	fastapi/_compat/v2.py
```

**Impact:** Users on Pydantic v1 will experience breaking changes. This should be a major version bump with clear migration guide.

**Documentation Files Deleted:**

```
D	docs/en/docs/advanced/advanced-python-types.md
D	docs/en/docs/advanced/json-base64-bytes.md
D	docs/en/docs/advanced/stream-data.md
D	docs/en/docs/how-to/migrate-from-pydantic-v1-to-pydantic-v2.md
... (many more documentation files deleted)
```

**Impact:** Loss of valuable documentation. The migration guide from Pydantic v1 to v2 is deleted while Pydantic v1 support is also removed.

### Warnings

**Type Hint Changes:**

```python
# Before:
def isoformat(o: datetime.date | datetime.time) -> str:
def decimal_encoder(dec_value: Decimal) -> int | float:

# After:
def isoformat(o: Union[datetime.date, datetime.time]) -> str:
def decimal_encoder(dec_value: Decimal) -> Union[int, float]:
```

**Impact:** Changed from modern type hints (`|`) to older style (`Union`). This is a step backward in code modernization.

**Large Diff from Merge Commits:**

Similar to other PRs reviewed, this includes many merge commits from master, making the diff difficult to review.

## Cross-file Impact

**Breaking Changes:**
- Pydantic v1 compatibility removed
- Documentation deleted
- Type hints changed to older style

**Public API:** 
- jsonable_encoder performance improved (positive)
- Pydantic v1 support removed (breaking)

**Test Coverage:** 
- No information on test changes (diff too large to analyze)

**User Impact:**
- **Positive:** Performance improvements for jsonable_encoder
- **Negative:** Breaking changes for Pydantic v1 users
- **Negative:** Loss of documentation
- **Unclear:** Migration path for Pydantic v1 users

## Assessment

**Comment Only**

This PR contains valuable performance work but is too large and combines multiple breaking changes.

### Recommendations

**Split into Multiple PRs:**

#### PR 1: Performance Refactor (Merge First)

**Scope:** ONLY jsonable_encoder performance changes
- Refactor encode_value, encode_dict functions
- Add NoneType constant
- Update ENCODERS_BY_TYPE
- Include benchmarks
- **NO breaking changes**
- **NO documentation deletions**

#### PR 2: Remove Pydantic v1 Compatibility (Major Version Bump)

**Scope:** ONLY Pydantic v1 removal
- Remove _compat directory
- Update error handling
- Add deprecation warnings first, then remove in next major version
- Update version to 1.0.0 (or next major)
- Include migration guide
- Update documentation

#### PR 3: Documentation Cleanup

**Scope:** ONLY documentation changes
- Review and update deleted docs
- Ensure critical documentation is preserved
- Add new docs for changes in PR 1 and PR 2

### Specific Concerns

1. **No Migration Guide:** Pydantic v1 users need clear guidance on how to upgrade
2. **Version Bump:** Removing Pydantic v1 compatibility requires a major version bump
3. **Documentation Loss:** Ensure no critical information is lost in doc cleanup
4. **Reviewability:** Current scope makes thorough review impossible

The performance work is valuable and should be merged. The Pydantic v1 removal is also likely necessary, but should be a separate, clearly communicated major version change.

**Recommendation:** Split this PR into at least 2-3 separate, reviewable PRs. Merge the performance refactor first, then handle breaking changes in a major version update.

tiangolo:
I see this would optimize for some datatypes at the expense of others, e.g. checking for Pydantic last, it could also increase the risk of exposing things that could break edge cases, so I prefer to skip this. Thanks! :coffee:

Ben-Brady:
I think it's very funny that I had just started Uni when I did the original version of this PR and 4 years later I've finished my degree just as it gets closed.

PR #13984, opened 2025-08-11
🔧 Update Speakeasy sponsor graphic

chailandau:
The image we supplied before was reading very low res on retina screens, this should fix that :)


svlandeg (approved):
Thanks!

PR #13997, opened 2025-08-17
📝 Fix code include for Pydantic models example in `docs/zh/docs/python-types.md`

anfreshman:
The Pydantic part of python-type in the Chinese documentation references the wrong file, and this PR fixes this error


YuriiMotov:
@anfreshman, thanks!

PR #14006, opened 2025-08-21
📝 Remove links to site callbackhell.com that doesn't exist anymore

dennybiasiolli:
I noticed them walking through the "Learn" doc in the website. The website callbackhell.com does not exists anymore.

Links pointing to callbackhell.ru are still working, but if you want I can remove them as well for consistency.

### DIFF
--- a/docs/fa/docs/async.md
+++ b/docs/fa/docs/async.md
@@ -383,7 +383,7 @@ Starlette (و **FastAPI**) بر پایه <a href="https://anyio.readthedocs.io/e
 
 توی نسخه‌های قبلی پایتون، می‌تونستی از نخ‌ها یا <a href="https://www.gevent.org/" class="external-link" target="_blank">Gevent</a> استفاده کنی. ولی کد خیلی پیچیده‌تر می‌شه برای فهمیدن، دیباگ کردن و فکر کردن بهش.
 
-توی نسخه‌های قبلی NodeJS / جاوااسکریپت مرورگر، از "کال‌بک‌ها" استفاده می‌کردی. که می‌رسید به <a href="http://callbackhell.com/" class="external-link" target="_blank">جهان کال‌بک‌ها</a>.
+توی نسخه‌های قبلی NodeJS / جاوااسکریپت مرورگر، از "کال‌بک‌ها" استفاده می‌کردی. که می‌رسید به "جهان کال‌بک‌ها".
 
 ## کروتین‌ها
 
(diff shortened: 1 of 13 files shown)


svlandeg (commented):
Thanks, this looks good!

I would also remove the links to callbackhell.ru just for consistency. 

Then, I worry a bit that the current text is not super informative:
> In previous versions of NodeJS / Browser JavaScript, you would have used "callbacks". Which leads to callback hell.

Before, the link was there to explain more, but now to someone unfamiliar to this concept, this explanation is not really sufficient. I would either

- put "callback hell" in quotes, to clarify it's a common concept that people could google
- change the link to a new one, maybe https://www.geeksforgeeks.org/javascript/what-to-understand-callback-and-callback-hell-in-javascript/ ?
- add a (very brief) description to the sentence, something like "... to callback hell, a problem occurring due to excessive nesting of functions", or something such. The downside then is that we'd need to translate this to all the languages that are being edited here 😬 

What do you think, @dennybiasiolli ?

>

dennybiasiolli:
@svlandeg I think the first option is the easiest, but I can even change the link to the new one. The only (future) problem could be that the link goes offline again, so maybe it's better to leave "callback hell" in quotes and people could google that.

I can do those changes right now if you are happy with them, and also removing the callbackhell.ru links

svlandeg:
That'd be great, thanks. Yea let's stick to having the concept in quotes.

docs/fa/docs/async.md
```diff
@@ -383,7 +383,7 @@ Starlette (و **FastAPI**) بر پایه <a href="https://anyio.readthedocs.io/e
 
 توی نسخه‌های قبلی پایتون، می‌تونستی از نخ‌ها یا <a href="https://www.gevent.org/" class="external-link" target="_blank">Gevent</a> استفاده کنی. ولی کد خیلی پیچیده‌تر می‌شه برای فهمیدن، دیباگ کردن و فکر کردن بهش.
 
-توی نسخه‌های قبلی NodeJS / جاوااسکریپت مرورگر، از "کال‌بک‌ها" استفاده می‌کردی. که می‌رسید به <a href="http://callbackhell.com/" class="external-link" target="_blank">جهان کال‌بک‌ها</a>.
+توی نسخه‌های قبلی NodeJS / جاوااسکریپت مرورگر، از "کال‌بک‌ها" استفاده می‌کردی. که می‌رسید به "جهان کال‌بک‌ها".
```
svlandeg, line 386: This _looks_ incorrect when viewing the github diff (one of the quote marks seem to be placed incorrectly) but when viewing the actual doc page, it seems fine.

svlandeg (approved):
Thanks, looks good to me!

docs/fa/docs/async.md
dennybiasiolli: This should be related to a right-to-left text thing.
On my localhost using VS Code, the diff is like this

image

svlandeg: Yep, it seems to be a bug with how Github displays the bug, but the edit is actually correct.

tiangolo (approved):
Great, thank you! 🚀

PR #14013, opened 2025-08-25
📝 Add APIKeyHeader usage example to security documentation

Gaurapatil1:
This PR adds a working example and explanation of how to use APIKeyHeader with the Security dependency in FastAPI.

It helps clarify usage for beginners and enhances the API Key Security Schemes section in the documentation.

### DIFF
--- a/docs/en/docs/reference/security/index.md
+++ b/docs/en/docs/reference/security/index.md
@@ -34,6 +34,35 @@ from fastapi.security import (
 
 ::: fastapi.security.APIKeyHeader
 
+### Using APIKeyHeader
+
+`APIKeyHeader` is a security tool for extracting API keys from HTTP headers.
+
+Here’s a simple example of how to use it in FastAPI:
+
+```python
+from fastapi import FastAPI, Security, HTTPException
+from fastapi.security import APIKeyHeader
+from starlette.status import HTTP_403_FORBIDDEN
+
+app = FastAPI()
+
+API_KEY = "mysecretapikey"
+API_KEY_NAME = "access_token"
+api_key_header = APIKeyHeader(name=API_KEY_NAME, auto_error=False)
+
+@app.get("/secure-data")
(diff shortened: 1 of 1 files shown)


nidhishgajjar:

## Review of PR #14013: Add APIKeyHeader usage example to security documentation

Good initiative adding a practical example for `APIKeyHeader` — this is a commonly used security utility that benefits from a concrete usage example.

### Security concern 🟡

The example hardcodes the API key directly in the source:

```python
API_KEY = "mysecretapikey"
```

Since this is security documentation, it's especially important to model best practices. Consider loading from environment variables instead:

```python
import os
API_KEY = os.environ.get("API_KEY")
```

Or at minimum, add a comment like `# In production, load from environment variables`.

Additionally, the direct comparison `api_key == API_KEY` is technically vulnerable to timing attacks. For production code, `secrets.compare_digest()` would be the proper approach. While this is a simple example, since it's specifically in a security reference section, demonstrating the safer pattern would be valuable.

### Minor issues

- **Extra blank lines**: There are 4 trailing blank lines before the next section; 1-2 would be conventional
- **`auto_error=False` handling**: With `auto_error=False`, the `api_key` parameter could be `None` when the header is missing. The current code handles this implicitly (since `None != "mysecretapikey"`), but an explicit `if not api_key or api_key != API_KEY:` would be clearer

### What looks good ✅

- Clean, concise example that demonstrates the key concepts
- Proper use of `Security()` dependency
- Good use of `HTTP_403_FORBIDDEN` status code for authentication failures
- Well-placed in the reference documentation

### Summary

A helpful documentation addition. The hardcoded API key is the main concern — since this is security documentation, it should ideally demonstrate secure practices like loading secrets from environment variables.

**Assessment: comment** (suggestion to improve security practices in the example)

tiangolo:

I want to implement a few big refactors and features that might solve many of these use cases and simplify how to use security.

Thank you! 🍰

PR #14032, opened 2025-09-01
🐛 Ensure that ASGIApp is mounted correctly to the sub-routes

sujin502:
During the use of FastAPI, I noticed a problem with sub-routes, My project structure is as follows:
```
┌─example
│ ├─main.py
│ └─sub_dir
│   ├─__init__.py
│   └─static
│     └─index.html
```
The code is as follows:

```python
# example/main.py
from fastapi import FastAPI

from sub_dir import sub_dir

app = FastAPI()
app.include_router(sub_dir, prefix="/app")

if __name__ == "__main__":
    import uvicorn

    uvicorn.run(app, host="127.0.0.1", port=8000)

```

```python
# example/sub_dir/__init__.py
from pathlib import Path as libPath

from fastapi import APIRouter
from fastapi.staticfiles import StaticFiles

skip = APIRouter()
skip.mount("/static", StaticFiles(directory=libPath(__file__).resolve().parent / "static"), name="skip")

__all__ = ["skip"]
```

After running `main.py` and accessing `http://127.0.0.1:8000/app/sub_dir/index.html`, one should have seen the content of `example/sub_dir/static/index.html`. However, it did not display as expected, Log entry as follows:
```
INFO:     Started server process [60576]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     127.0.0.1:59030 - "GET /app/sub_dir/index.html HTTP/1.1" 404 Not Found
```

By examining the source code, it was discovered that `APIRouter.include_router` failed to properly handle `routing.Mount`.

After the repair, we ran `main.py`, and we could see that the content in `example/sub_dir/static/index.html` could be displayed normally. The log is as follows:
```
INFO:     Started server process [78196]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     127.0.0.1:58825 - "GET /app/sub_dir/index.html HTTP/1.1" 200 OK
```

This will be extremely useful in complex routing scenarios!!!

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1355,6 +1355,8 @@ class APIRouter(routing.Router):
                 self.add_websocket_route(
                     prefix + route.path, route.endpoint, name=route.name
                 )
+            elif isinstance(route, routing.Mount):
+                self.mount(prefix + route.path, route.app, name=route.name)
         for handler in router.on_startup:
             self.add_event_handler("startup", handler)
         for handler in router.on_shutdown:
--- a/tests/test_additional_response_extra.py
+++ b/tests/test_additional_response_extra.py
@@ -1,4 +1,7 @@
+from pathlib import Path as libPath
+
 from fastapi import APIRouter, FastAPI
+from fastapi.staticfiles import StaticFiles
 from fastapi.testclient import TestClient
 
 router = APIRouter()
@@ -6,6 +9,11 @@ router = APIRouter()
 sub_router = APIRouter()
 
 app = FastAPI()
+
+
+sub_router.mount(
+    "/static", StaticFiles(directory=libPath(__file__).resolve().parent), name="static"
+)
 
 
 @sub_router.get("/")
@@ -18,6 +26,14 @@ router.include_router(sub_router, prefix="/items")
 app.include_router(router)
 
 client = TestClient(app)
+
+
+def test_sub_router_mount():
+    response = client.get(f"/items/static/{libPath(__file__).resolve().stem}.py")
+    assert response.status_code == 200, response.text
+
+
+test_sub_router_mount()
 
 
 def test_path_operation():


svlandeg:
Hi @sujin502, thanks for your contribution and detailed explanation!

Can you design a minimal unit test that would fail on `master` but would succeed on this PR? Then it will be 100% clear that this PR fixes that use-case, and it will also ensure that the test suite coverage remains at 100% (the CI will be red otherwise).

I'll put this PR in draft until the CI goes green.

sujin502:
I'm very sorry. Due to my busy work schedule, I didn't notice this until now. The unit test code has been completed. There are absolutely no errors now!

Finally, I would like to express my deepest gratitude to you and all the contributors for your efforts. It is truly wonderful.

fastapi/routing.py
```diff
@@ -1355,6 +1355,8 @@ def read_users():
                 self.add_websocket_route(
                     prefix + route.path, route.endpoint, name=route.name
                 )
+            elif isinstance(route, routing.Mount):
```
YuriiMotov, line 1358:
```suggestion
            elif isinstance(route, routing.Mount) and isinstance(
                route.app, StaticFiles
            ):
```
I'm not sure we should add this for any `mount`. Mounted apps can also have sub-apps. If we want to support this I think we should add much more tests.
I suggest we only support this for `StaticFiles` for now

YuriiMotov (changes requested):
@sujin502, thank you for your interest and efforts!

The test should be moved to a separate module (e.g. `test_sub_router_mount.py`) - `test_additional_response_extra.py` is not correct place..
And, test can be simplified a bit:

<details><summary>code in the details</summary>

```py
from pathlib import Path as libPath

from fastapi.staticfiles import StaticFiles

CURRENT_DIR = libPath(__file__).resolve().parent


def test_sub_router_mount():

    app = FastAPI()

    sub_router = APIRouter()
    sub_router.mount("/static", StaticFiles(directory=CURRENT_DIR), name="static")

    app.include_router(prefix="/sub", router=sub_router)

    client = TestClient(app)
    response = client.get("/sub/static/__init__.py")

```

</details>

I also think we should **limit** this to **only support mounting `StaticFiles`** in sub-routers (I think it only makes sense for StaticFiles and we should discourage mounting sub-apps to the sub-routes).
The title would be something like: "🐛 Fix mounting `StaticFiles` to sub-routers"

### Related issue and discussions:
* #10180
* https://github.com/fastapi/fastapi/discussions/9070
* https://github.com/fastapi/fastapi/discussions/8555

PR #14033, opened 2025-09-01
🛠️ Add `add-permalinks` and `add-permalinks-page` to `scripts/docs.py`

YuriiMotov:
Add missing permalinks to all En docs:
```
typer scripts/docs.py run add-permalinks
```

Re-generate permalinks for all En docs:
```
typer scripts/docs.py run add-permalinks --update-existing
```

Add missing permalinks to specific page in En docs:
```
typer scripts/docs.py run add-permalinks-page docs/en/docs/advanced/behind-a-proxy.md
```

---

At the moment we have 1 missed permalink (I committed it an then reverted): https://github.com/fastapi/fastapi/pull/14033/commits/6bf43ba276329a5f65434f9243bf29045f49eada

Can we fix it in this PR or should I open separate PR for this?

---

Also, I don't know if we need to add some logging?


svlandeg (commented):
How will we keep things up-to-date, will this script run on Github actions when a PR changes/adds documentation?

YuriiMotov:
> How will we keep things up-to-date, will this script run on Github actions when a PR changes/adds documentation?

Yes, it's supposed to be used by GitHub actions

svlandeg:
Then I wouldn't worry about the one missing link, and let the script deal with it on the next update.

scripts/docs.py
```diff
@@ -27,7 +29,7 @@
 {!../../docs/missing-translation.md!}
 """
 
-non_translated_sections = [
+non_translated_sections = (
     "reference/",
```
svlandeg: To be OS agnostic:
```suggestion
    "reference",
```
assuming we have no "normal" pages starting with "reference" (I don't think that's a real issue)

svlandeg: Why is that we exclude the `reference` directory?

YuriiMotov: > Why is that we exclude the reference directory?

I guess that's because it's generated from sources? If we decided to translate it, we would need to update all translations every time code changes..

As for slash in the end - good catch!
But instead of removing it, I would go with `f"reference{os.sep}",`

YuriiMotov: Looked at code and realized that there are a lot of other occurrences of hardcoded `/`.
So, we need to decide whether we want to refactor it in this PR on in separate PR?

svlandeg: I suggest we (only) update the hardcoded slashes of the new code that is contributed with this PR. Because when I run this new script on my Windows machine, it starts generating permalinks for the `reference` docs and we don't want to do deal with any PRs of users that get the same thing.

With respect to hardcoded slashes already in the current code base: let's edit those in a separate PR.

svlandeg: > But instead of removing it, I would go with f"reference{os.sep}",

That's fine too by me!

YuriiMotov: Done!

svlandeg (approved):
LGTM! 

Ran the script on a Windows box and that works good too (`behind-a-proxy.md` gets the missing permalink) and also `--update-existing` works well (finds one occassion where a space should be added).

As this script has already been tested extensively before and the results of it merged previously, I think this one is good to merge now.

tiangolo (approved):
Excellent job, thank you! 🚀

PR #14036, opened 2025-09-02
👽️ Ensure compatibility with Pydantic 2.12.0

cjwatson:
`tests/test_multi_body_errors.py::test_openapi_schema` failed with pydantic 2.12.0a1 due to
https://github.com/pydantic/pydantic/pull/11987.  Since we're testing the exact contents of the JSON schema, the easiest fix seems to be to add version-dependent handling for this.


YuriiMotov:
I think we should try to avoid adding new dependency.

Maybe something with `dirty_equals`?
Take a look at the code in the details

<details>

```py
from dirty_equals import IsOneOf

v1 = {
    "age": {
        "title": "Age",
        "anyOf": [
            {"exclusiveMinimum": 0.0, "type": "number"},
            {"type": "string"},
        ],
    }
}

v2 = {
    "age": {
        "title": "Age",
        "anyOf": [
            {"exclusiveMinimum": 0.0, "type": "number"},
            {"type": "string", "pattern": r"^(?!^[-+.]*$)[+-]?0*\d*\.?\d*$"},
        ],
    }
}


expected = {
    "age": {
        "title": "Age",
        "anyOf": [
            {"exclusiveMinimum": 0.0, "type": "number"},
            IsOneOf(
                {"type": "string"},
                {"type": "string", "pattern": r"^(?!^[-+.]*$)[+-]?0*\d*\.?\d*$"}
            ),
        ],
    }
}

assert v1 == expected
assert v2 == expected

```

</details>

cjwatson:
> I think we should try to avoid adding new dependency.
> 
> Maybe something with `dirty_equals`? Take a look at the code in the details

Ah yes, that works too, thanks.  Amended.

YuriiMotov:
I just tried running this branch with pydantic 2.12.0a1, and I have 4 failed tests
```
FAILED tests/test_forms_single_model.py::test_send_all_data - pydantic.warnings.UnsupportedFieldAttributeWarning: The 'alias' attribute with value 'with' was provided to the `Field()` function, which has no effect in the context it was used. 'alias' is field-spec...
FAILED tests/test_forms_single_model.py::test_defaults - pydantic.warnings.UnsupportedFieldAttributeWarning: The 'alias' attribute with value 'with' was provided to the `Field()` function, which has no effect in the context it was used. 'alias' is field-spec...
FAILED tests/test_forms_single_model.py::test_invalid_data - pydantic.warnings.UnsupportedFieldAttributeWarning: The 'alias' attribute with value 'with' was provided to the `Field()` function, which has no effect in the context it was used. 'alias' is field-spec...
FAILED tests/test_forms_single_model.py::test_no_data - pydantic.warnings.UnsupportedFieldAttributeWarning: The 'alias' attribute with value 'with' was provided to the `Field()` function, which has no effect in the context it was used. 'alias' is field-spec...
```

cjwatson:
> I just tried running this branch with pydantic 2.12.0a1, and I have 4 failed tests
> 
> ```
> FAILED tests/test_forms_single_model.py::test_send_all_data - pydantic.warnings.UnsupportedFieldAttributeWarning: The 'alias' attribute with value 'with' was provided to the `Field()` function, which has no effect in the context it was used. 'alias' is field-spec...
> FAILED tests/test_forms_single_model.py::test_defaults - pydantic.warnings.UnsupportedFieldAttributeWarning: The 'alias' attribute with value 'with' was provided to the `Field()` function, which has no effect in the context it was used. 'alias' is field-spec...
> FAILED tests/test_forms_single_model.py::test_invalid_data - pydantic.warnings.UnsupportedFieldAttributeWarning: The 'alias' attribute with value 'with' was provided to the `Field()` function, which has no effect in the context it was used. 'alias' is field-spec...
> FAILED tests/test_forms_single_model.py::test_no_data - pydantic.warnings.UnsupportedFieldAttributeWarning: The 'alias' attribute with value 'with' was provided to the `Field()` function, which has no effect in the context it was used. 'alias' is field-spec...
> ```

These are due to https://github.com/pydantic/pydantic/pull/12028.  In that PR, @Viicos wrote:

> I've reached out to the FastAPI team regarding the failing tests. They will need to address them in some way.

But I really don't follow what FastAPI is doing here; it seems to be something to do with `ModelField.__post_init__` in `fastapi/_compat.py`?  Somebody who's actually an expert in that code would need to weigh in, I think.  Maybe @tiangolo?

Viicos:
I discussed internally with the FastAPI team a while ago about this. The fix I proposed was to use the [`warning.catch_warnings`](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) context manager when creating type adapters in FastAPI. Pydantic emits a specific `UnsupportedFieldAttributeWarning` in this case, so it can safely be filtered while still having other potential warnings raised.

cjwatson:
> The fix I proposed is to use [warning.catch_warnings](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) context manager when creating type adapters in FastAPI.

OK, thanks for the advice.  https://github.com/fastapi/fastapi/pull/14036/commits/a33cc0a5a136d469d72ca34981297d4a698008ae implements that.

fastapi/_compat.py
```diff
@@ -109,9 +110,24 @@ def type_(self) -> Any:
...
 
         def __post_init__(self) -> None:
-            self._type_adapter: TypeAdapter[Any] = TypeAdapter(
-                Annotated[self.field_info.annotation, self.field_info]
-            )
+            with warnings.catch_warnings():
+                # Pydantic >= 2.12.0a1 warns about this when building
+                # TypeAdapters from field information that uses aliases.
+                # The Pydantic team recommends ignoring this in this case:
+                # https://github.com/fastapi/fastapi/pull/14036#issuecomment-3316045587
```
Viicos:
```suggestion
                # Pydantic >= 2.12.0 warns about field specific metadata that is unused
                # (e.g. `TypeAdapter(Annotated[int, Field(alias='b')])`). In some cases, we
                # end up building the type adapter from a model field annotation so we
                # need to ignore the warning:

File: fastapi/_compat.py [hunk 2]
```diff
@@ -109,9 +110,24 @@ def type_(self) -> Any:
...
+                    )
+                except ImportError:  # pragma: no cover
+                    pass
+                else:  # pragma: no cover
+                    warnings.simplefilter(
+                        "ignore", category=UnsupportedFieldAttributeWarning
+                    )
+                self._type_adapter: TypeAdapter[Any] = TypeAdapter(
+                    Annotated[self.field_info.annotation, self.field_info]
+                )
```
Viicos: - I believe you need to use the `catch_warnings` context manager, because as it stands the warning will always be ignored from now on.
- I would suggest using an explicit version check on the `pydantic` version instead of a `try..except` statement, so that it can easily be unconditionally imported once FastAPI drops support for Pydantic < 2.12.

tests/test_multi_body_errors.py
```diff
@@ -185,7 +185,15 @@ def test_openapi_schema():
                                 "title": "Age",
                                 "anyOf": [
                                     {"exclusiveMinimum": 0.0, "type": "number"},
-                                    {"type": "string"},
+                                    IsOneOf(
+                                        # pydantic < 2.12.0a1
```
Viicos:
```suggestion
                                        # pydantic < 2.12.0
```

tests/test_multi_body_errors.py
```diff
@@ -185,7 +185,15 @@ def test_openapi_schema():
...
+                                        {"type": "string"},
+                                        # pydantic >= 2.12.0a1
```
Viicos:
```suggestion
                                        # pydantic >= 2.12.0
```

Viicos (dismissed):

svlandeg:
I changed the base branch of this PR to `feature/py314` where we've got a CI run for Python 3.14 and Pydantic 2.12 set up - this way we can actually see the failing tests 😉

cjwatson:
@svlandeg That seems to have resulted in a pile of errors that have nothing to do with this branch.  Or am I missing something?

fastapi/_compat.py
cjwatson: I'm already using the `catch_warnings` context manager just above, so I'm confused about your first point.

I've made your suggested second change.

svlandeg:
> @svlandeg That seems to have resulted in a pile of errors that have nothing to do with this branch. Or am I missing something?

I think these are exactly the errors that need resolving - the test suite now actually runs Pydantic 2.12.0a1 on Python 3.14 (rc)?

(they're really all mostly versions of the same error, from a quick glance)

svlandeg:
PS: @cjwatson: it's fine if you don't want to continue working on those other errors - just let us know - and we'll take over this branch because they need fixing anyway 😉

cjwatson:
Those errors don't appear to happen in my local setup with pydantic 2.12.0a1 on Python 3.13 (which was what I was originally trying to fix with this PR).  I'd appreciate it if somebody else could take over the rest of those fixes, yes.

svlandeg:
> Those errors don't appear to happen in my local setup with pydantic 2.12.0a1 on Python 3.13 (which was what I was originally trying to fix with this PR)

I think you're right - those additional errors are new in Python 3.14. 

> I'd appreciate it if somebody else could take over the rest of those fixes, yes.

Sure thing - I'm going to look into them now. I'll probably merge this PR into our feature branch `feature/py314` and continue from there.

svlandeg (approved):
Ok, just to recap and apologies for the earlier confusion!

This PR ensures compatibility with Pydantic 2.12.0, which will be released soon.

In particular, `pydantic` 2.12.0a1 causes 2 types of failures in our test suite:

- `test_openapi_schema` fails due to https://github.com/pydantic/pydantic/pull/11987 as explained by OP. The test is adjusted to work on both new and older versions of Pydantic.
- warnings about using the `alias` attribute with value `with` in the `Field()` function, cf PR https://github.com/pydantic/pydantic/pull/12028. This PR suppresses those warnings when creating type adapters in FastAPI, which seems like a reasonable solution to me.

I had intended to merge these fixes as part of a larger PR to get ready for Python 3.14, but have reverted that and suggest we merge & release this separately to ensure compatibility with the upcoming Pydantic 2.12.0.

Thanks again for your work on this, @cjwatson !

fastapi/_compat.py
```diff
@@ -109,9 +110,22 @@ def type_(self) -> Any:
...
         def __post_init__(self) -> None:
-            self._type_adapter: TypeAdapter[Any] = TypeAdapter(
-                Annotated[self.field_info.annotation, self.field_info]
-            )
+            with warnings.catch_warnings():
+                # Pydantic >= 2.12.0 warns about field specific metadata that is unused
+                # (e.g. `TypeAdapter(Annotated[int, Field(alias='b')])`). In some cases, we
+                # end up building the type adapter from a model field annotation so we
+                # need to ignore the warning:
+                if PYDANTIC_VERSION_MINOR_TUPLE >= (2, 12):  # pragma: no cover
```
svlandeg: Once Pydantic 2.12 is released, we'll be able to remove the `no cover` statement here.

Viicos: (Which should happen very soon, hopefully today).

fastapi/_compat.py
```diff
@@ -109,9 +110,22 @@ def type_(self) -> Any:
...
+                    from pydantic.warnings import (  # type: ignore[attr-defined]
+                        UnsupportedFieldAttributeWarning,
+                    )
+
+                    warnings.simplefilter(
```
Viicos, line 121: This is still incorrect, as per my first point [here](https://github.com/fastapi/fastapi/pull/14036#discussion_r2369201471).

fastapi/_compat.py
svlandeg: (I know, but we have a bit of a chicken-or-egg situation going on here 😉)

Viicos: Yeah no worries, just wanted to give you a release estimate!

fastapi/_compat.py
svlandeg: Hi @Viicos, thanks for the additional review! 🙏 

Can you elaborate on your point?

The current code uses `warnings.simplefilter` within a `with warnings.catch_warnings()` context manager. It is my understanding that, once the context manager exits, after L128, the warnings filter is restored to its original state when the context was entered (L113) ?

Viicos: My bad, I always get confused by this `warnings` API, as the context manager isn't opened right on top of the filter. All good!

svlandeg: No worries, thanks for double checking! 🙏

tiangolo (approved):
Nice, thank you @cjwatson! 🚀 

And thanks a lot for all the help here @YuriiMotov, @Viicos, @svlandeg, @patrick91 🙌 

So cool to come and see all the conversations already done, the code simplified to the minimum, all the context easy to get quickly. Great job, thank you all! 🍰 

I'll release this in the next hours/minutes in FastAPI 0.118.1 🎉

PR #14059, opened 2025-09-10
📝 Remove outdated formatting from `path-params-numeric-validations.md` for languages `en`, `es` and `uk`.

svlandeg:
This was a leftover after the docs reformatting in https://github.com/fastapi/fastapi/pull/12825


YuriiMotov (approved):
LGTM

We don't have the annotated version for this code example, so, this tip is not needed

tiangolo (approved):
Nice, thank you! ☕

PR #14060, opened 2025-09-10
👽️ Update `Color` import depending on the version of Pydantic

SaisakthiM:
Description:

### Summary
This PR addresses the deprecation warning for `Color` in FastAPI when using Pydantic v2 while maintaining backward compatibility with Pydantic v1.  

Previously, `fastapi/encoders.py` imported `Color` directly from `pydantic.color`, which is deprecated in Pydantic v2, leading to warnings for users upgrading.  

This change ensures that the correct `Color` class is used depending on the installed Pydantic version.

---

### Changes
- Conditional import of `Color`:
  ```python
  try:
      # Pydantic v2
      from pydantic_extra_types.color import Color
  except ImportError:
      # Fallback for Pydantic v1
      from pydantic.color import Color


No other functionality is altered.

Fully backward-compatible with both Pydantic v1 and v2.

Checklist

 I have verified that the change fixes the deprecation warning.
 Code follows existing FastAPI style and conventions.
 Tests pass for Pydantic v2.
 Tests pass for Pydantic v1.
 No breaking changes introduced for either Pydantic version.
 Documentation updated if necessary (none needed for this fix).

Testing

Environment 1 – Pydantic v2 (current default)
pytest tests/test_jsonable_encoder.py -v
21 tests collected
18 passed, 3 skipped (v1-specific tests)


Environment 2 – Pydantic v1
pip install "pydantic<2.0"
pytest tests/test_jsonable_encoder.py -v
21 tests collected
17 passed, 4 skipped (v2-specific tests)

Results: ✅ All relevant tests pass in both environments. Only version-specific tests are skipped, as expected.

Motivation

Remove Pydantic v2 deprecation warnings.
Maintain backward compatibility with Pydantic v1.
Ensure smooth user experience for projects upgrading to Pydantic v2.


svlandeg (commented):
Hi @SaisakthiM, thanks for the PR!

It looks like the tests are failing for Pydantic v1, could you have a look?

Also, could you add a test that checks the deprecation warning isn't thrown anymore? (basically we'd need a test that fails on `master` but succeeds on this PR, to showcase the change in functionality)

fastapi/encoders.py
```diff
@@ -19,13 +19,20 @@
 
 from fastapi.types import IncEx
 from pydantic import BaseModel
-from pydantic.color import Color
 from pydantic.networks import AnyUrl, NameEmail
 from pydantic.types import SecretBytes, SecretStr
+from pydantic_extra_types.color import Color  # ✅ updated import
```
svlandeg: You probably need to import this only when you need it for Pydantic v2, as this will crash on Pydantic v1.

SaisakthiM: thanks, so should i add a test for this import

SaisakthiM:
hi @svlandeg, 

finally after many trial and error
i found pydantic v1 doesn't have Color module so i first imported the v2 one and for the fallback used a dummy class 
and fixed the linting using # type : ignore
the error is fixed
and added a test file too 

the result in my PR

```
(.venv-12) PS C:\Users\saika\OneDrive\Documents\GitHub\fastapi> pytest tests/test_color_compat.py
========================================================== test session starts ===========================================================
platform win32 -- Python 3.12.3, pytest-8.4.2
collected 1 item                                                                                                                          

tests\test_color_compat.py .                                                                                                        [100%]

=========================================================== 1 passed in 0.30s ============================================================
```

the result in the master branch, 

```
E           pydantic.error_wrappers.ValidationError: 1 validation error for Model
E           c
E             instance of DummyColor expected (type=type_error.arbitrary_type; expected_arbitrary_type=DummyColor)
```
i got a validation error i don't know why but i got a error

SaisakthiM:
this is not working at all now as i updated it
better to give up

PR #14062, opened 2025-09-10
✅ Simplify tests for response_model

dynamicy:


svlandeg:
Hi! Could you look into the failing tests for Python 3.8?

dynamicy:
Hi @svlandeg,
I’ve fixed the issues causing the Python 3.8 test failures — everything should be working now. 
Please let me know if you spot anything else!

svlandeg (dismissed):
Thanks for looking into this, @dynamicy!

I'm not quite sure I understand why you're rewriting 
`client = TestClient(app)` 

to a `get_client` function that yields several `TestClient`objects, even when there's only one. In fact, I don't think we need to change the test files that don't have variants of them (which is the focus of #13150)

Have a look at https://github.com/fastapi/fastapi/pull/13197/files for instance. The `get_client` method becomes something like this:

```
@pytest.fixture(
    name="client",
    params=[
        "tutorial001",
        pytest.param("tutorial001_py310", marks=needs_py310),
    ],
)
def get_client(request: pytest.FixtureRequest):
    mod = importlib.import_module(f"docs_src.schema_extra_example.{request.param}")

    client = TestClient(mod.app)
    return client
```

I suggest to keep to this style for consistency across the code base. Then most of the edits of this PR can be removed, and we can focus on the files that actually have code variants for Python 3.10 and clean those up.

dynamicy:
Thanks for the feedback, @svlandeg! Thanks for pointing me to the right direction!

svlandeg (approved):
This now looks good to me, thanks @dynamicy !

svlandeg:
@dynamicy: this was already approved and ready to merge, but now I see that a change has been force-pushed. I generally dislike force pushes because it doesn't allow a reviewer to track the history. What have you changed on this PR since my approval?

Just to clarify: there's really no need to obliterate the history of a PR. We squash right before merging anyway.

dynamicy:
Thanks for the heads-up! GitHub showed a notice that the branch was out of date with the base, so I used the Rebase and update option in the UI. That resulted in a force-push of the rebased commits.
There are no functional changes since the approval—only the base was updated (rebased). Sorry for the noise and any inconvenience.
I’m happy to avoid rebasing/force-pushing here going forward and leave base updates to maintainers (we’ll squash on merge anyway). Please let me know if you’d like me to do anything else to make the review easier.

svlandeg (approved):
Ok cool, thanks for the explanation! Had another look and confirmed all is still good to merge here 😎

tiangolo (approved):
Nice, thanks! 🚀 

And thanks @svlandeg for the review and help! ☕

PR #14064, opened 2025-09-10
Fix: Add redact_error_details flag to FastAPI app to hide sensitive i…

shriyashaThakur:
Problem:

FastAPI’s default request and WebSocket validation handlers expose detailed internal error information.
This can leak sensitive data structures in production environments, posing a security risk.

Solution:

Added a new flag redact_error_details to the FastAPI app.
When enabled (redact_error_details=True), validation errors return generic messages instead of exposing internal details:
HTTP request validation (422) → "Request validation failed"
WebSocket validation errors → "WebSocket validation failed"
Default behavior remains unchanged if the flag is not set.

Files Modified:
fastapi/applications.py → added redact_error_details parameter
fastapi/exception_handlers.py → updated handlers to respect the flag

Impact:
Improves security by preventing sensitive data leakage in production.
Safe for debugging if the flag is left False.


YuriiMotov:
Request validation errors don't contain **"internal error information"**. They are based on openapi schema, that should be open (otherwise, how users know the interface of your API?).

If in your specific case you need to hide these error details, you can easily achieve this by setting the error handler for `RequestValidationError`:

```py
async def request_validation_exception_handler(
    request: Request, exc: RequestValidationError
) -> JSONResponse:
    content = {"detail": "Request validation failed"}
    return JSONResponse(status_code=HTTP_422_UNPROCESSABLE_ENTITY, content=content)
```

Having this, I'll close this PR.
Thanks for your efforts!

PR #14065, opened 2025-09-10
Fix #13533 : Fix/form defaults

### LINKED ISSUE #13533: Multiple regressions in the handling of forms & form validation

shriyashaThakur:
Restores the expected behaviour where empty form fields are treated as missing and default values are used, fixing regressions in x-www-form-urlencoded and multipart forms (Issue #13533). Includes 11 new tests to verify functionality.


YuriiMotov:
Please, clear the branch from unrelated changes

PR #14066, opened 2025-09-10
Fix #13471: Enforce required realm in HTTP Basic Auth

### LINKED ISSUE #13471: Basic auth `realm` is REQUIRED but handled as optional

shriyashaThakur:
### Summary
This PR fixes issue #13471 by making the HTTP Basic Auth `realm` parameter required in FastAPI, ensuring compliance with RFC 7617.

### Problem
The `realm` parameter in HTTP Basic Auth was optional, which is non-compliant with RFC 7617. Clients that require a `realm` could fail authentication.

Solution
- Made `realm` a required parameter in the `HTTPBasic` class.
- Added validation to prevent empty or whitespace-only `realm` values.
- Updated `WWW-Authenticate` headers to always include `realm`.
- Updated documentation and examples to highlight the required `realm` parameter.

Tests
- Created `tests/test_security_http_basic_realm_required.py` with 11 new test cases covering:
  - Missing, empty, and whitespace-only `realm` errors.
  - Proper inclusion of `realm` in headers.
  - Correct authentication behavior with valid credentials.
- Updated existing test `test_security_http_basic_optional.py` to include `realm`.

 Verification
- All 26 HTTP Basic Auth tests pass (11 new + 15 existing).
- No linting errors.
- Backward compatibility maintained for code that already specifies `realm`.

Compliance
- Fully compliant with RFC 7617:
  - `realm` is mandatory.
  - Headers always include `realm`.
  - Proper error messages for invalid or missing `realm`.


YuriiMotov:
@shriyashaThakur, I think you used wrong branch to open this PR. It doesn't contain changes you described.

UPD: seems that you started the branch not from the master, but from the branch with other changes. Still, this branch doesn't contain changes that you described

PR #13504, opened 2025-03-20
✅ Simplify tests for websockets

valentinDruzhinin:
#13150

PR #14074, opened 2025-09-12
test: parametrize response model tutorial websockets

dynamicy:


YuriiMotov:
We closed [previous similar PR](https://github.com/fastapi/fastapi/pull/13504) because we will likely drop the support of Python 3.8 in the nearest future.
But now I think it might be better to merge it since we haven't decided when to drop Python 3.8 support

PR #14075, opened 2025-09-13
✨ Add a `default_status_code` attribute to FastAPI response classes

dynamicy:
This PR adds explicit `default_status_code` attributes to all FastAPI response classes and updates the OpenAPI documentation generation logic to use these attributes safely.

The explicit `default_status_code` attributes also improve code clarity by making it obvious what status code each response type uses by default, and they're properly reflected in the generated OpenAPI schemas.

### DIFF
--- a/fastapi/responses.py
+++ b/fastapi/responses.py
@@ -1,12 +1,52 @@
-from typing import Any
+from typing import Any, Optional
 
-from starlette.responses import FileResponse as FileResponse  # noqa
-from starlette.responses import HTMLResponse as HTMLResponse  # noqa
-from starlette.responses import JSONResponse as JSONResponse  # noqa
-from starlette.responses import PlainTextResponse as PlainTextResponse  # noqa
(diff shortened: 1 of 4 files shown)


fastapi/responses.py
```diff
@@ -1,12 +1,52 @@
...
+from starlette.responses import StreamingResponse as StarletteStreamingResponse
+
+
+class Response(StarletteResponse):
+    default_status_code: int = 200
+
+    def __init__(
+        self,
+        content: Any = None,
+        *,
```
alaminopu, line 18: Looks like you are using [Keyword-Only Arguments](https://peps.python.org/pep-3102/) feature in python,  It might be a breaking changes for people are currently not setting `status_code` as keyword argument. Looking [Starlette code ](https://github.com/Kludex/starlette/blob/main/starlette/responses.py#L38), they also don't enforce it as keyword only argument. 

e.g. the response `Response("Hello", 404)` is not going to work anymore.

alaminopu (changes requested):

PR #14077, opened 2025-09-14
⬆️ Upgrade Starlette supported version range to >=0.40.0,<0.49.0

musicinmybrain:
Starlette 0.48.0 introduced `HTTP_422_UNPROCESSABLE_CONTENT` as a new name for `HTTP_422_UNPROCESSABLE_ENTITY` and deprecated the old name due to RFC 9110; see https://github.com/Kludex/starlette/pull/2939.

I updated all instances ~to the new name and increased the lower bound on the Starlette version~ to use the literal integer or string value `422`/`"422"` instead of either `HTTP_422_UNPROCESSABLE_CONTENT` or `HTTP_422_UNPROCESSABLE_ENTITY`, which makes FastAPI compatible with Starlette >=0.48 without deprecation warnings while preserving backwards compatibility with Starlette<0.48.

@Kludex, FYI.

### DIFF
--- a/docs_src/handling_errors/tutorial005.py
+++ b/docs_src/handling_errors/tutorial005.py
@@ -10,7 +10,7 @@ app = FastAPI()
 @app.exception_handler(RequestValidationError)
 async def validation_exception_handler(request: Request, exc: RequestValidationError):
     return JSONResponse(
-        status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+        status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
         content=jsonable_encoder({"detail": exc.errors(), "body": exc.body}),
     )
 
--- a/docs_src/handling_errors/tutorial005.py
+++ b/docs_src/handling_errors/tutorial005.py
@@ -10,8 +10,6 @@ app = FastAPI()
 @app.exception_handler(RequestValidationError)
 async def validation_exception_handler(request: Request, exc: RequestValidationError):
     return JSONResponse(
-        # 422 = fastapi.status.HTTP_422_UNPROCESSABLE_CONTENT (RFC 9110,
-        # Starlette >=0.48), previously HTTP_422_UNPROCESSABLE_ENTITY
         status_code=422,
         content=jsonable_encoder({"detail": exc.errors(), "body": exc.body}),
     )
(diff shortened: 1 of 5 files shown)


Kludex:
If we use 429 instead of the constant, there's no need to bump.

Maybe we should do that?

musicinmybrain:
> If we use 429 instead of the constant, there's no need to bump.
> 
> Maybe we should do that?

As in, just write literal integer `422`? Sure, that’s a sensible idea, and it would be easy enough to do. @tiangolo, do you have a preference one way or the other?

tiangolo:
Yep, I think a literal `422` should be enough, and that way we avoid having to restrict the lower bound limit, which is otherwise not necessary. :nerd_face:

docs_src/handling_errors/tutorial005.py
```diff
@@ -10,7 +10,9 @@
 @app.exception_handler(RequestValidationError)
 async def validation_exception_handler(request: Request, exc: RequestValidationError):
     return JSONResponse(
-        status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+        # 422 = fastapi.status.HTTP_422_UNPROCESSABLE_CONTENT (RFC 9110,
+        # Starlette >=0.48), previously HTTP_422_UNPROCESSABLE_ENTITY
+        status_code=422,
```
svlandeg, line 13: We have to be carefull here, this script is referenced in the docs [here](https://fastapi.tiangolo.com/tutorial/handling-errors/#use-the-requestvalidationerror-body), and by adding more lines, the wrong line will be highlighted in the docs (used to be L14 which is now L16).

Anyway, I don't think we need this comment here, or elsewhere where we're switching to 422. Let's just edit to `status_code=422` without the comments.

musicinmybrain: > We have to be carefull here, this script is referenced in the docs [here](https://fastapi.tiangolo.com/tutorial/handling-errors/#use-the-requestvalidationerror-body), and by adding more lines, the wrong line will be highlighted in the docs (used to be L14 which is now L16).
> 
> Anyway, I don't think we need this comment here, or elsewhere where we're switching to 422. Let's just edit to `status_code=422` without the comments.

Thanks. Your comment-removal commit looks good.

Kludex (approved):

tiangolo (approved):
Nice, thank you! 🚀 

And thanks @Kludex and @svlandeg! 🙌 

This will be available in FastAPI 0.116.2, released in the next hours.

Issue #12965, opened 2024-11-20
Will FastAPI support QUERY http method? "app.query"


Kludex:
### Discussed in https://github.com/fastapi/fastapi/discussions/6049

<div type='discussions-op-text'>

<sup>Originally posted by **FilipeMarch** December 16, 2022</sup>
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from fastapi import FastAPI

app = FastAPI()


@app.query('/query/subjects')
def query_subjects(schema: ArbitrarySchema):
    with Session(engine) as db:
        subjects = db.query(Subject).all()
        return schema(**subjects)

# something like this
```

### Description

There is a new HTTP method called QUERY, I discovered it this week and it is super interesting!

https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html

I was trying to make a GET route that would accept a request body, but started receiving an error on Swagger UI:

[TypeError: Request has method 'GET' and cannot have a body](https://github.com/swagger-api/swagger-ui/issues/5891#issuecomment-1340591857)

The idea is that sometimes we need to make a big or complex query, and this is the scenario in which we can see the advantages of GraphQL, and although I have seen that [we can integrate FastAPI with GraphQL](https://fastapi.tiangolo.com/advanced/graphql/), I was wondering if FastAPI will ever be able to simply accept a QUERY method like, for example, `app.query("/query/subjects")`.

For example, suppose I have this object:
```yml
{
   "id": 1,
   "name": "Math"
   "tags": [
       {
           "id": 1,
           "name": "Algebra",
           "number_of_clicks": 1,
           "number_of_questions": 7,
           "number_of_answers": 3,
           "number_of_comments": 2,
           "number_of_votes": 1,
       }]
[...]

PR #14083, opened 2025-09-17
📝 Clarify non-standard HTTP methods like QUERY

DivInstance:
- Clarifies that QUERY is not part of OpenAPI, so it won’t appear in docs/Swagger UI.
- Recommends POST for complex query bodies.
- Adds advanced example using api_route(..., methods=[\"QUERY\"], include_in_schema=False).
- References #12965.


YuriiMotov:
@DivInstance, thank you for your interest!

I think we don't need this clarification..
Since Openapi 3.2 has been released, `QUERY` method is a part of the standard now.
When Swagger supports openapi 3.2, FastAPI will probably also support it.

tiangolo:
Thanks for the interest, I'll pass on this for now. ☕

PR #14086, opened 2025-09-18
📌 Pin `httpx` to `>=0.23.0,<1.0.0`

YuriiMotov:
See discussion: https://github.com/fastapi/fastapi/discussions/14084

Currently we have `httpx >=0.23.0`, and impending httpx `1.x` release can break the api.

---

Also relaxed pins in dev dependencies and GH-actions


Kludex (approved):

tiangolo (approved):
Thank you! 🚀

PR #14089, opened 2025-09-18
✨ Improve JSON Parsing Error Messages

doganarif:
This PR enhances JSON parsing error messages to provide better debugging information.

## Changes

Modified JSON error handling in fastapi/routing.py to include additional context in the error response:
  - Line and column numbers where the error occurred (0-indexed)
  - Position in the JSON string
  - Error snippet (max 80 chars with ellipsis) showing the problematic JSON

## Error Response Format

The error location remains ["body", position] to maintain backward compatibility. Additional debugging information is provided in the ctx field:
```JSON
  {
    "loc": ["body", 47],
    "msg": "JSON decode error",
    "input": {},
    "ctx": {
      "error": "Expecting value",
      "position": 47,
      "line": 3,
      "column": 12,
      "snippet": "...\"price\": 'invalid'..."
    }
  }
```
## Testing

  - Added comprehensive test suite in tests/test_json_body_decode_error.py
  - Tests cover single-line errors, multi-line errors, snippet ellipsis behavior, and edge cases


svlandeg:
Hi @doganarif, thanks for the PR! It looks like the smokeshow coverage test is failing - which basically means that there are new lines of code that are not being executed by at least one unit test. Could you look into this and expand the tests so that all edge cases are tested? 🙏 

I'm putting the PR in draft as long as the CI fails - feel free to mark as "ready for review" once everything is green!

doganarif:
> Hi @doganarif, thanks for the PR! It looks like the smokeshow coverage test is failing - which basically means that there are new lines of code that are not being executed by at least one unit test. Could you look into this and expand the tests so that all edge cases are tested? 🙏
> 
> I'm putting the PR in draft as long as the CI fails - feel free to mark as "ready for review" once everything is green!

Hey @svlandeg 

Got it, fixed the coverage! thanks.

tests/test_tutorial/test_body/test_tutorial001.py
```diff
@@ -206,11 +206,14 @@ def test_post_broken_body(client: TestClient):
             "detail": [
                 {
                     "type": "json_invalid",
-                    "loc": ["body", 1],
-                    "msg": "JSON decode error",
-                    "input": {},
+                    "loc": ["body", 1, 2],
+                    "msg": "JSON decode error - Expecting property name enclosed in double quotes at line 1, column 2",
```
YuriiMotov:
```suggestion
                    "loc": ["body", 1],
                    "msg": "JSON decode error",
```

I think we don't need this duplication - having `error`, `line` and `column` in `ctx` should be enough.

Also, the semantic of `loc` field seems to be different - it should be indexes of list or dict

tests/test_tutorial/test_body/test_tutorial001.py
```diff
@@ -206,11 +206,14 @@ def test_post_broken_body(client: TestClient):
...
+                    "input": "{some broken json}",
```
YuriiMotov: As for `input` - I'm not sure, but it's probably incorrect to pass partial input string here:

For Pydantic the `input` field contains whole input value. So, it might be better to add this snippet to `ctx`

tests/test_tutorial/test_body/test_tutorial001.py
```diff
@@ -206,11 +206,14 @@ def test_post_broken_body(client: TestClient):
...
                     "ctx": {
-                        "error": "Expecting property name enclosed in double quotes"
+                        "error": "Expecting property name enclosed in double quotes",
+                        "position": 1,
+                        "line": 1,
+                        "column": 2,
```
YuriiMotov: There is an inconsistency here - `position` starts from `0`, but `line` and `column` start from `1`

YuriiMotov (commented):
@doganarif, thank you for your interest!

I like the idea in general - it will help the debugging process easier.

But I think we should follow the semantic of fields we use. Otherwise this might be breaking for some clients.
I also think we should avoid the duplication of information in the response details

@doganarif, what do you think?

doganarif:
**@YuriiMotov**, thanks for the review! You're right on all points - I should stick to existing conventions to avoid breaking changes.

**On `loc` and `msg`:** Agreed, I'll revert `loc` to `["body", position]` to keep its original semantic meaning. Line/column in `ctx` + the enhanced `msg` should be enough.

**On `input`:** Good catch - I'll keep the full document in `input` and move the snippet to `ctx` (maybe `ctx["snippet"]`?) to match Pydantic's convention.

**On position/line/column:** Makes sense. Since `position` is 0-based (from `JSONDecodeError.pos`) and line/column are 1-based for readability, I'll add a comment in the code to clarify this.

So the structure would be:
```python
"loc": ("body", e.pos),
"msg": f"JSON decode error - {e.msg} at line {line_number}, column {column_number}",
"input": e.doc,
"ctx": {
    "error": e.msg,
    "position": e.pos,  # 0-based
    "line": line_number,  # 1-based
    "column": column_number,  # 1-based
    "snippet": error_snippet,
}
```

Does this work? I can push these changes and resolve the conflicts if this looks good to you.Retry

YuriiMotov:
I personally think we should revert `msg` and `input` to how it was before this PR:
```py
                            "msg": "JSON decode error",
                            "input": {},
```
Especially for `input` as it can be huge.

As for `line` and `column` - I would make them 0-based for consistency. Few people will notice this in code comments.

Also, I didn't try it, but seems that `JSONDecodeError` has `lineno` and `colno` attributes ([docs](https://docs.python.org/3/library/json.html#json.JSONDecodeError)). Can we use them?

doganarif:
Here's a more natural response:

---

**@YuriiMotov**, yeah I was on the fence about `input` - it could definitely get huge with large payloads. Reverting both to the original is probably safer.

On line/column being 0-based - I initially went with 1-based since that's what most editors show, but consistency with `position` is more important here. And `JSONDecodeError` does have `lineno` and `colno` - way cleaner than calculating them from the doc string.

Updated structure:
```python
"loc": ("body", e.pos),
"msg": "JSON decode error",
"input": {},
"ctx": {
    "error": e.msg,
    "position": e.pos,
    "line": e.lineno - 1,  # Convert to 0-based
    "column": e.colno - 1,  # Convert to 0-based
    "snippet": error_snippet,
}
```

I'll push these changes.

doganarif:
@YuriiMotov, pushed the changes. Reverted loc, msg, and input, moved snippet to ctx, and using e.lineno/e.colno now.

tests/test_json_error_improvements.py
```diff
@@ -0,0 +1,162 @@
...
+    assert error["ctx"]["column"] == 26
+    assert "None" in error["ctx"]["snippet"]
+
+
+def test_json_decode_error_multiline():
+    invalid_json = """
+{
+  "name": "Test",
+  "price": 'invalid'
+}"""
```
YuriiMotov: I think it would be better to move this outside the function:
```py
INVALID_JSON = """
{
  "name": "Test",
  "price": 'invalid'
}"""

class Item(BaseModel):
    name: str
    ...
```

tests/test_json_error_improvements.py
```diff
@@ -0,0 +1,162 @@
...
+    assert error["input"] == {}
+    assert "position" in error["ctx"]
+    assert "line" in error["ctx"]
+    assert "column" in error["ctx"]
+    assert "snippet" in error["ctx"]
+
+
+def test_json_decode_error_in_middle_of_long_document():
+    # Create a JSON with error early in a long document (need >40 chars after error)
+    # The error is at position for "invalid" which needs at least 41 chars after it
```
YuriiMotov:
```suggestion
```
I think this comment is not needed.
I would parameterize this test to cover cases:
* snippet starts with `...`
* snippet ends with `...`
* snippet starts and ends with `...`
* snippet contains the whole input

Then we can delete `test_json_decode_error_shows_snippet` (it will be covered)

tests/test_json_error_improvements.py
```diff
@@ -0,0 +1,162 @@
...
+    response = client.post(
+        "/items/",
+        json={"name": "Test Item", "price": 19.99, "description": "A test item"},
+    )
+
+    assert response.status_code == 200
+    data = response.json()
+    assert data["name"] == "Test Item"
+    assert data["price"] == 19.99
+    assert data["description"] == "A test item"
```
YuriiMotov:
```suggestion
```
I think this is not needed here

YuriiMotov (approved):
LGTM

I would work a bit more on tests - I think we only need to test the functional that we changed, and we need to make it consistently (if we test `endswith("...")`, we should probably also test `startswith()` and also `startswith() and endswith()`).

I suggest the following set of tests:

<details><summary>see in the details</summary>

```py
import pytest
from fastapi import FastAPI

SHORT_JSON = """{
  "name": "abc",
  "price": 12.3,
}"""

LONG_JSON = """{
  "name": "loooooooooooooooooooooooooooooooooooooooooooooooooooong name",
  "price": 3.14159265359,
  "description": "looooooooooooooooooooooooooooooooooooooong description"
}"""


def _break_json(s: str, pos: int) -> str:
    return s[:pos] + "🐛" + s[pos:]


app = FastAPI()


class Item(BaseModel):
    name: str
    price: float
    description: str = None


@app.post("/items/")
async def create_item(item: Item):
    return item


client = TestClient(app)


@pytest.mark.parametrize(
    "body",
    [
        pytest.param('{"name": abc}', id="1-line-json-error"),
        pytest.param(_break_json(SHORT_JSON, 10), id="short-json-error-in-middle"),
    ],
)
def test_snippet_short_json(body: str):
    response = client.post(
        "/items/", content=body, headers={"Content-Type": "application/json"}
    )
    assert response.status_code == 422
    errors = response.json()["detail"]

    snippet = errors[0]["ctx"]["snippet"]
    assert snippet == body


@pytest.mark.parametrize(
    "body,expected_snippet",
    [
        pytest.param(
            _break_json(LONG_JSON, 0),
            '🐛{\n  "name": "looooooooooooooooooooooooo...',
            id="long-json-error-at-start",
        ),
        pytest.param(
            _break_json(LONG_JSON, 75),
            (
                '...ooooooooooooooooooooooooooooooong name",🐛\n  '
                '"price": 3.14159265359,\n  "descripti...'
            ),
            id="long-json-error-in-middle",
        ),
        pytest.param(
            _break_json(LONG_JSON, len(LONG_JSON)),
            '...ooooooooooooooooooooooong description"\n}🐛',
            id="long-json-error-at-end",
        ),
    ],
)
def test_snipped_long_json(body, expected_snippet):
    response = client.post(
        "/items/", content=body, headers={"Content-Type": "application/json"}
    )
    assert response.status_code == 422
    errors = response.json()["detail"]

    snippet = errors[0]["ctx"]["snippet"]
    assert snippet == expected_snippet


@pytest.mark.parametrize(
    "body,pos_line_col",
    [
        pytest.param(
            '{"name": 🐛}',
            (9, 0, 9),
            id="one-line-json-error",
        ),
        pytest.param(
            '🐛: abc,\n  "price": 12.3}',
            (0, 0, 0),
            id="multiline-json-error-at-start",
        ),
        pytest.param(
            '{\n  "name": "abc",\n  🐛: "abc"\n}',
            (21, 2, 2),
            id="multiline-json-error-in-middle",
        ),
        pytest.param(
            '{\n  "name": "abc",\n  "price": 12.3\n}🐛',
            (36, 3, 1),
            id="multiline-json-error-at-end",
        ),
    ],
)
def test_json_decode_error_position_line_col(body, pos_line_col):
    response = client.post(
        "/items/", content=body, headers={"Content-Type": "application/json"}
    )
    assert response.status_code == 422
    errors = response.json()["detail"]

    ctx = errors[0]["ctx"]

    assert ctx["position"] == pos_line_col[0]
    assert ctx["line"] == pos_line_col[1]
    assert ctx["column"] == pos_line_col[2]

```

</details>

I would also rename the test module to `test_json_body_decode_error.py`

doganarif:
Hi @YuriiMotov,

I made changes but,

I've kept test_successful_item_creation because it's needed to maintain 100% coverage - 
it's the only test that exercises the successful return path in create_item. 
> Without it, coverage drops to 98%.

tests/test_json_error_improvements.py
YuriiMotov: To fix coverage you can just add `# pragma: no cover`:
```py
@app.post("/items/")
async def create_item(item: Item):
    return item  # pragma: no cover
```

doganarif: Thanks! fixed :)

YuriiMotov:
@doganarif, thanks!
Could you please update the description to reflect latest changes?

doganarif:
@YuriiMotov  done! Thanks

YuriiMotov (approved):
LGTM

@doganarif, thanks!

YuriiMotov:
@doganarif, no need to merge `master` branch if there are no conflicts. It creates unnecessary noise

doganarif:
Hello @YuriiMotov ,

Will this PR will get migrated wdyt?

PR #14091, opened 2025-09-19
Add caching logic to get_dependant

dynamicy:
Summary
1. Propagated the original dependant’s use_cache flag when rebuilding dependency overrides so overridden dependencies retain their intended caching behavior.
2. Added regression coverage confirming overrides of Depends(..., use_cache=False) continue to call the dependency for each usage without caching.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -610,6 +610,7 @@ async def solve_dependencies(
                 call=call,
                 name=sub_dependant.name,
                 security_scopes=sub_dependant.security_scopes,
+                use_cache=sub_dependant.use_cache,
             )
 
         solved_result = await solve_dependencies(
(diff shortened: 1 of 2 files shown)


YuriiMotov:
As I can see it already works correctly.
Test added by this PR passes on `master` branch


I also tested it with the following code and it works as I expected:

<details><summary>See in the details</summary>

```py

app = FastAPI()

counter = {"calls": 0}


async def dep():
    counter["calls"] = counter["calls"] + 1
    return "dependency"


@app.get("/")
async def read_root(
    d1: str = Depends(dep, use_cache=False),
    d2: str = Depends(dep, use_cache=False),
):
    return counter



def test_():

    def dep_override():
        counter["calls"] = counter["calls"] + 1
        return "dependency"

    app.dependency_overrides[dep] = dep_override

    client = TestClient(app)

    counter["calls"] = 0
    response = client.get("/")
    assert response.json() == {"calls": 2}

```

</details>

PR #14099, opened 2025-09-21
🐛 Fix support for `StreamingResponse`s with dependencies with `yield` or `UploadFile`s, close after the response is done

tiangolo:
🐛 Fix support for `StreamingResponse`s with dependencies with `yield` or `UploadFile`s, close after the response is done

---

There's (again) an `AsyncExitStack` middleware, now used mainly to close objects after the response is done.

There's also a new wrapper duplicating some logic in Starlette but creating an `AsyncExitStack` for the exit code in dependencies with `yield`, for regular responses and WebSockets as well.

The exit code of dependencies with `yield` is now executed after the response is done.


anton-ryzhov:
This PR also fixes https://github.com/fastapi/fastapi/pull/12508

anton-ryzhov:
But introduces another bug as described [here](https://github.com/fastapi/fastapi/pull/12508#issuecomment-3706462807)

PR #14100, opened 2025-09-22
📝 Clarify the difference between an empty string and `None` query parameter

EnamSingh:
Problem: Readers expect ?q= to pass None, but it passes empty string.

Change: Add note clarifying omission (None) vs empty string.

Why: Matches HTTP semantics & FastAPI behavior; reduces confusion.

Testing: Built docs locally; ran tutorial tests subset.


tiangolo:
Thanks for the interest, there's another PR tweaking the wording of this section, I'll focus on that one. Thanks! :coffee:

PR #14102, opened 2025-09-22
♻️ Refactor sponsor image handling

alejsdev:
♻️ Refactor sponsor image handling

Hide the sponsor announcement area by default, wait for all sponsor images to load successfully, and only show the area if all images load properly


tiangolo (approved):
😅 🚀

PR #14104, opened 2025-09-22
tests: add empty - string ('?q=') case --> 422 for regex validated qu…

EnamSingh:
Problem: ?q= sends an empty string (not None) and should fail ^fixedquery$ → 422.

Change: Add a test next to existing “no_query” and “fixedquery” cases.

Why: Complements docs clarification; prevents regressions.

Testing: Ran file-only and tutorial subset locally.

### DIFF
--- a/tests/test_regex_deprecated_params.py
+++ b/tests/test_regex_deprecated_params.py
@@ -70,7 +70,38 @@ def test_query_params_str_validations_item_query_nonregexquery():
             ]
         }
     )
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@EnamSingh, thanks for your interest!

The right place to add such test would be `tests/test_query.py`. And the test should be just "sendint the empty query parameter results in empty string recieved"

PR #14105, opened 2025-09-23
docs: clarify unsupported QUERY method + test

ethicalzeus07:
This PR clarifies that FastAPI only supports standard HTTP methods (RFC 7231).

- Added a note in docs about unsupported non-standard methods like QUERY
- Added a small test to confirm that app.query is invalid

Related to issue #12965

### DIFF
--- a/docs/en/docs/advanced/path-operation-advanced-configuration.md
+++ b/docs/en/docs/advanced/path-operation-advanced-configuration.md
@@ -202,3 +202,9 @@ Here we reuse the same Pydantic model.
 But the same way, we could have validated it in some other way.
 
 ///
+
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@ethicalzeus07, thanks for you interest!

I think we don't need such clarifications..
We should describe in the docs features that are supported by FastAPI.
We have [API reference](https://fastapi.tiangolo.com/reference/fastapi/) - it should be clear that if there is no such method, then it's not supported.

If we choose the way of describing all things we don't support, documentation will get bloated and less readable.
People who don't find this in docs will just go to issues\discussions and find discussions about this.

PR #14107, opened 2025-09-23
fix(security): add realm to Basic Auth WWW-Authenticate header (RFC 7617 compliance)

jsibitoye:
**Description**

This PR fixes the WWW-Authenticate header for HTTP Basic Auth to comply with RFC 7617 §2, which requires a realm attribute in the authentication challenge.

**Motivation**

Without the realm attribute, some HTTP clients may reject the authentication challenge or behave inconsistently. Adding the realm ensures proper interoperability and standards compliance.

**Changes**

Updated fastapi/security/http.py to include realm="fastapi" in the WWW-Authenticate header.

Updated affected tests to match the expected header:

tests/test_httpbasic_realm.py

tests/test_security_http_basic_optional.py

tests/test_tutorial/test_tutorial006.py

**Notes**

All existing tests pass after modification.

### DIFF
--- a/fastapi/security/http.py
+++ b/fastapi/security/http.py
@@ -189,10 +189,10 @@ class HTTPBasic(HTTPBase):
     ) -> Optional[HTTPBasicCredentials]:
         authorization = request.headers.get("Authorization")
         scheme, param = get_authorization_scheme_param(authorization)
-        if self.realm:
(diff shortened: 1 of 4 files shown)


YuriiMotov:
@jsibitoye, thanks for your interest!

We already have https://github.com/fastapi/fastapi/pull/13511

Could you please explain why we need this one as it's a duplicate of #13511 ?

PR #14115, opened 2025-09-25
📝 Fix CORS tutorial inconsistency with `allow_credentials=True` and wildcard values in `allow_methods`/`allow_headers`

FlaviusRaducu:
# Overview
[The CORS documentation](https://github.com/fastapi/fastapi/blob/master/docs/en/docs/tutorial/cors.md) states the following:

> None of allow_origins, allow_methods and allow_headers can be set to ['*'] if allow_credentials is set to True. All of them must be [explicitly specified](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#credentialed_requests_and_wildcards).

However. [the CORS tutorial example](https://github.com/fastapi/fastapi/blob/master/docs_src/cors/tutorial001.py#L13), sets `allow_methods` and `allow_headers` to `['*']`.

Related discussions:
* https://github.com/fastapi/fastapi/discussions/14491


docs_src/cors/tutorial001.py
```diff
@@ -14,8 +14,8 @@
     CORSMiddleware,
     allow_origins=origins,
     allow_credentials=True,
-    allow_methods=["*"],
-    allow_headers=["*"],
+    allow_methods=["GET"],
+    allow_headers=["X-Example"],
```
YuriiMotov:
```suggestion
    allow_headers=["Authorization", "X-Example", "Content-Type"],
```
I think it would make sense to add `Authorization` header (as we specified `allow_credentials=True`) and probably `Content-Type` header (as it's very common).
What do you think?

FlaviusRaducu: Hi Yurii, thank you for taking the time to review this PR.

Yes, that makes sense; though since this example is a `simple request`, the `Content-Type` header is allowed by default, the docs mention that [here](https://github.com/fastapi/fastapi/blob/master/docs/en/docs/tutorial/cors.md?plain=1#L59C1-L59C75). I don't think we have to explicitly add it as it might seem a bit confusing on a first read regarding why we did that for `Content-Type` and not the other defaults.

What do you think? I'm happy to add it if you believe it would be helpful to have it there.

YuriiMotov: It's true for simple requests. But if you send a request with credentials, it will not be a simple request

simone-trubian: Hi Both! Looking at the code of both files I wonder if the confusing part is not so much this part
```python
    allow_methods=["*"],
    allow_headers=["*"],
```
but this
```python
allow_credentials=True,
```

I tried running the code commenting out `allow_credentials=True,` and the tests run fine. This is expected because the code doesn't include any login or credential exchange.

In the spirit of keeping the code clean and focused on a simple CORS example, I suggest removing `allow_credentials=True` and keeping the `allow_methods` and `allow_headers` parameters as they are.

What do you guys think?

YuriiMotov: @simone-trubian, good point :)

```
app.add_middleware(
    CORSMiddleware,
    allow_origins=origins,
    allow_credentials=False,
    allow_methods=["*"],
    allow_headers=["*"],
)
```

FlaviusRaducu: Yes, that's a good point. I will update the files; thanks for the review guys.

YuriiMotov:
Let's better add `allow_credentials=False,`?
Otherwise we will need to update the highlights in code include blocks (including translations)

image

FlaviusRaducu:
> Let's better add `allow_credentials=False,`? Otherwise we will need to update the highlights in code include blocks (including translations)

@YuriiMotov sorry; yes, that makes sense.

YuriiMotov (approved):
LGTM

johnson-earls:
This change makes it comply, but might make it a less useful example?  It's hard to imagine an API that doesn't use credentials.  It might be more useful to leave the `allow_credentials` set to `True` and give a reasonable example of the list of methods and headers that would be used.
```
    allow_credentials=True,
    allow_methods=['GET','POST','PUT','PATCH','DELETE'],
    allow_headers=['Content-Type','Authorization'],
```
(_I'm not sure what all headers should be included, which is why i didn't submit my own PR for this_.)

YuriiMotov:
> This change makes it comply, but might make it a less useful example? It's hard to imagine an API that doesn't use credentials.

But this specific code example doesn't use credentials. So, I think we should keep it simple.

We can later add one more code examples showing how to use `CORSMiddleware` in apps that use credentials

PR #14120, opened 2025-09-26
📝 Update Primary Key notes for the SQL databases tutorial to avoid confusion

FlaviusRaducu:
For developers unfamiliar with `SQLModel` ( might be just me though :) ) creating a nullable primary key can be quite confusing when reading this tutorial section.
I believe, adding a bit more information resolves that initial confusion.
~~The extra comments have been taken from [the official SQLModel docs](https://sqlmodel.tiangolo.com/tutorial/create-db-and-table/#primary-key-id).~~


simone-trubian:
Hi @FlaviusRaducu thank you for taking the time to submit the improvement. I think you're right in saying that section should be more explicit in why that type signature is required. I feel however you could reword the submission to make it read less "staccato" that is less of an edit and more like something that was always there. As an example, Copilot suggest re-wording the paragraph likes so:

Declaring the type as int | None tells SQLModel that the column should be an INTEGER in the SQL database and that it can be NULL. However, in practice, the id column is always required and cannot be NULL—so why use int | None?

The reason is that although the id is mandatory in the database, it's not provided by our code. Instead, it's automatically generated by the database itself.

docs/en/docs/tutorial/sql-databases.md
```diff
@@ -67,6 +67,10 @@ There are a few differences:
 
     By having the type as `int | None`, SQLModel will know that this column should be an `INTEGER` in the SQL database and that it should be `NULLABLE`.
 
+    But inside the SQL database, the `id` is always required and can't be `NULL`. So, why should we declare it with `int | None`?
+
+    The `id` will be required in the database, but it will be generated by the database, not by our code.
```
YuriiMotov: I think we can make it clearer..
How about the following version?

---

* `Field(primary_key=True)` tells SQLModel that the `id` is the **primary key** in the SQL database (you can learn more about SQL primary keys in the SQLModel docs).

    **Note:** We use `int | None` for the primary key field so that in Python code we can *create an object without an `id`* (`id=None`), assuming the database will *generate it when saving*. SQLModel understands that the database will provide the `id` and *enforces it as non-null  `INTEGER`*. See <a href="https://sqlmodel.tiangolo.com/tutorial/create-db-and-table/#primary-key-id" class="external-link" target="_blank">SQLModel docs on primary keys</a> for details.

---

Markdown:

```
* `Field(primary_key=True)` tells SQLModel that the `id` is the **primary key** in the SQL database (you can learn more about SQL primary keys in the SQLModel docs).

    **Note:** We use `int | None` for the primary key field so that in Python code we can *create an object without an `id`* (`id=None`), assuming the database will *generate it when saving*. SQLModel understands that the database will provide the `id` and *enforces it as non-null  `INTEGER`*. See <a href="https://sqlmodel.tiangolo.com/tutorial/create-db-and-table/#primary-key-id" class="external-link" target="_blank">SQLModel docs on primary keys</a> for details.
```

simone-trubian: Nice @YuriiMotov ! Even better!

FlaviusRaducu: Thank you guys for taking the time to review this PR!
Yeah, I think Yurii's version is definitely better.

FlaviusRaducu: @YuriiMotov I have updated `enforces` to `enforce`; assuming that you're talking about the database, please let me know if I'm wrong.

YuriiMotov: I'm not a native English speaker, so I may be wrong..
My idea was "SQLModel understands ... and enforces ...". Maybe "enforce" is not perfect verb here. Will "defines" be better?

```
SQLModel understands that the database will provide the `id` and *defines the column as a non-null `INTEGER`* in the database schema
```

simone-trubian: Hi @YuriiMotov @FlaviusRaducu I think "enforce" is a reasonable term but the "s" after it opens a question. If SQLModel feeds the id or the id _type_ to the database, then the subject of the sentence stays on SQLModel so it enforces (or "implements" can be used as a synonym) the id. However if the database does it on its own, then the subject of the "enforce" verb changes from SQLModel to the database and we should drop the "s". Hope this makes sense.

FlaviusRaducu: @YuriiMotov no worries at all and yes, I think that this updated version is better. It makes it clear that the SQLModel will define the expected primary key constraints on the schema. The previous version, as @simone-trubian pointed out, makes it sound a bit like SQLModel will enforce the primary key constraints at runtime.

YuriiMotov (approved):
LGTM!

tiangolo (approved):
Nice, thanks! 🚀

PR #14152, opened 2025-10-02
📝 Add External Link: Getting started with logging in FastAPI

itssimon:
I wrote an article on request logging and correlating application logs in FastAPI that I'd like to add to the External Links section in the docs. The topic is not covered by any of the existing links.

### DIFF
--- a/docs/en/data/external_links.yml
+++ b/docs/en/data/external_links.yml
@@ -1,5 +1,9 @@
 Articles:
   English:
+  - author: Apitally
+    author_link: https://apitally.io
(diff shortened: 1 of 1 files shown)


YuriiMotov:
We already have similar article, and it looks to me that this new article is a next edition of that old one. Should we replace that old with this new?

https://github.com/fastapi/fastapi/blob/56c0632f863b49b6224a8eea1b899ea8dacffa2e/docs/en/data/external_links.yml#L43-L46

PR #14159, opened 2025-10-05
App state property is not recognized by most type checkers because it's declared inside the constructor

diegoromerodev:
image

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -64,6 +64,24 @@ class FastAPI(Starlette):
     ```
     """
 
+    state: Annotated[
+        State,
+        Doc(
+            """
+                A state object for the application. This is the same object for the
+                entire application, it doesn't change from request to request.
+
+                You normally wouldn't use this in FastAPI, for most of the cases you
+                would instead use FastAPI dependencies.
+
+                This is simply inherited from Starlette.
+
+                Read more about it in the
+                [Starlette docs for Applications](https://www.starlette.io/applications/#storing-state-on-the-app-instance).
+                """
+        ),
+    ]
+
     def __init__(
         self: AppType,
         *,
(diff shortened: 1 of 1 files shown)


YuriiMotov:
What type checkers are you talking about?
I use VSCode with Pylance and I can see type hints and code completion:

image

diegoromerodev:
> What type checkers are you talking about? I use VSCode with Pylance and I can see type hints and code completion:
> 
> image

Ah, good to know, thanks for checking!
In my case, PyCharm fails to provide completions when using the default JetBrains type checker.

YuriiMotov:
Related discussions:
* https://youtrack.jetbrains.com/projects/PY/issues/PY-63306/False-positive-for-unresolved-reference-of-state-instance-field-in-FastAPI-app?utm_source=chatgpt.com  (marked as FIXED)
* https://youtrack.jetbrains.com/projects/PY/issues/PY-79866/Uninitialized-attribute-defined-in-init-are-not-suggested-in-completion?utm_source=chatgpt.com

If we want to fix this on FastAPI's side, we will need to make it consistently for all attibutes, not only `.state`.
But first we need to decide whether we should go this way or wait this is fixed on PyCharm's side

svlandeg:
As an additional datapoint: I use PyCharm Professional 2021.2.2 and this does work on my end:

image

And after I type "state", I can navigate to its original definition on L928 in `applications.py`.

YuriiMotov:
@diegoromerodev, please try updating your PyCharm and plugins to latest versions

svlandeg:
In the meantime, I suggest to close this PR as I don't think this is something that needs fixing on our side.

diegoromerodev:
@svlandeg Oh, good to know. I will update and re-check.

@YuriiMotov do you agree this should be considered to be a problem with this particular type checker?

If you both agree on this matter, I will close this PR. Let me know.

Thank you both for looking into this.

YuriiMotov:
Yes, it looks like a problem with particular version

PR #14168, opened 2025-10-08
✨ Add support for `from pydantic.v1 import BaseModel`, mixed Pydantic v1 and v2 models in the same app

tiangolo:
✨ Add support for `from pydantic.v1 import BaseModel`, mixed Pydantic v1 and v2 models in the same app


tiangolo:
This was just released in FastAPI 0.119.0 :tada:

PR #14171, opened 2025-10-10
✅ Expand test matrix to include Windows and MacOS

svlandeg:
Instead of testing on `ubuntu` only, run the test suite also on `macos` and `windows`. Alternate OS & Pydantic for older versions of Python in order not to blow up the combinations (& CI time)

### Ensure correct encoding

When `fastapi-cli` prints the 🚀 , this was causing a `UnicodeEncodeError`. Adding specifically `PYTHONIOENCODING=utf-8` to the environment before running the command, avoids the error.

### Use OS in coverage report name

By including the OS in the coverage file, we prevent overwriting / "file exists" errors.

### Fix combining coverage files from different OS's

When combining the different .coverage.XXX files from the different runs in the matrix, coverage needs to remap the locations. It's easier to just use the [relative_files](https://coverage.readthedocs.io/en/latest/config.html#config-run-relative-files) option as suggested in the Coverage docs.

2025-12-11 @tiangolo edit: update to not include coverage for all intermediate versions, to avoid max file size limit in Smokeshow.


svlandeg:
Not sure why the coverage test won't finish 😕

svlandeg:
[Maintenance update]: the failing CI tests are unrelated, cf https://github.com/fastapi/fastapi/pull/14325.

[Update]: those are now fixed 🎉

svlandeg:
I wonder why the coverage run doesn't end. I feel like I've seen this before on other PRs, but this one gets the issue consistently 🤔

tiangolo (approved):
Awesome, thank you! 🙌

tiangolo:
I just noticed why the coverage is not being reported: https://github.com/fastapi/fastapi/actions/runs/19846770073/job/56865669963

It's erroring out while uploading the files for exceeding the site size of 50MB:

```
     z_fee7577972855584_tutorial001_an_py.html ct=text/html size=16.8KB
    z_fee7577972855584_tutorial001_an_py310_py.html ct=text/html size=15.5KB
    z_b954e22f23c2e1a6_tutorial002_py39_py.html ct=text/html size=12.7KB
    ERROR! z_f0c103b61377e127_test_tutorial001_py.html status=429 response=429: You've exceeded the site size limit of 52428800.
    ERROR! z_86d9a3f0c3307d30_test_tutorial001_py.html status=429 response=429: You've exceeded the site size limit of 52428800.
```

---

I was checking another run from any other PR, and it seems we are very close in size already before this PR, the size is already 45.9MB: https://github.com/fastapi/fastapi/actions/runs/19846633836/job/56865298781

```
    z_e7126a6eca7662f4___init___py.html ct=text/html size=4.5KB
    z_a44f0ac069e85531_test_sub_callbacks_py.html ct=text/html size=88.8KB
    z_6de190ce15502a91_docs_py.html ct=text/html size=91.5KB
upload complete ✓ site size 45.9MB
go to https://smokeshow.helpmanual.io/303b1p370b0b5z0q2q2d/
setting status on github.com/fastapi/fastapi#19eea13, success: "Coverage 100.00%"
Smokeshow upload success!
```

Maybe we could ask for an increase in the limit, but I don't know what the size of the site we need is. Maybe we can try to figure that out to request that increase (or figure out an alternative). :thinking:

tiangolo:
I'm thinking maybe we can just wait for the next week after removing Python 3.8 that might be enough. And then the week after or so removing Pydantic v1, that would also reduce the total size in half or so.

We might just wait a bit.

tiangolo (approved):
Nice, thank you! 🙌 

I tweaked it to not upload coverage for all variants (that are already covered in other tests) to not pass the max file size in Smokeshow, I needed this for another PR as well. 😎

PR #14173, opened 2025-10-11
feat(middleware): expose BaseHTTPMiddleware and update docs

santoshvandari:
### Summary
This PR exposes `BaseHTTPMiddleware` directly from `fastapi.middleware.base` for convenience, and adds a new multiple-file custom middleware example in documentation.


YuriiMotov (commented):
@santoshvandari, thanks for your interest and efforts!

I think we shouldn't focus on adding more docs on `BaseHTTPMiddleware` - the use of it should be discouraged due to the facts that it has some limitations (https://www.starlette.io/middleware/#limitations, also it's [much slower than pure ASGI middleware](https://github.com/fastapi/fastapi/discussions/6985)).

PR #14186, opened 2025-10-14
🐛 Fix internal Pydantic v1 compatibility (warnings) for Python 3.14 and Pydantic 2.12.1

svlandeg:
The Python 3.14 tests started failing with the Pydantic v.2.12.1 release, because of the combination of https://github.com/fastapi/fastapi/pull/14168 and https://github.com/pydantic/pydantic/pull/12367.

Basically we were importing Pydantic `v1` functionality even when the user had (only) Pydantic v2 installed. This was done in https://github.com/fastapi/fastapi/pull/14168 to support more easy migration from v1 to v2 by allowing mixing versions within the same app.

Here, we prevent importing any `v1` functionality by creating dummy classes for Python 3.14 / Pydantic 2.12.0 onwards, which [drops support](https://github.com/pydantic/pydantic/releases/tag/v2.12.0) for `v1` and emits a [warning](https://github.com/pydantic/pydantic/pull/12367) when importing `v1` functionality.

### TODO

- [x] fix coverage
- [ ] check warning is still shown to users at some point, and not just swallowed by this code.


fastapi/_compat/may_v1.py
```diff
@@ -0,0 +1,123 @@
+import sys
+from typing import Any, Dict, List, Literal, Sequence, Tuple, Type, Union
+
+from fastapi.types import ModelNameMap
+
+if sys.version_info >= (3, 14):
```
svlandeg, line 6: Probably better to check directly for Pydantic 2.12.0+ ?

Then again, Pydantic themselves say 

> The core functionality of Pydantic V1 is not compatible with Python 3.14 or greater.

tiangolo (approved):
Awesome, thank you @svlandeg! 🙌 

I checked and confirmed that on Python 3.14, if users try to use Pydantic v1 they still get a warning on their own code, but not on older versions. So we're good to go! 🚀

PR #14188, opened 2025-10-14
🌐 Sync German docs

nilslindemann:
... with recent changes in one PR.


nilslindemann:
For the records, I am aware of the failing tests for python 3.14 – `from pydantic.v1 import BaseConfig as BaseConfig` raises `UserWarning: Core Pydantic V1 functionality isn't compatible with Python 3.14 or greater.`, which is probably expected – but not sure how to fix that.

Oh, I see this is already worked on in [14186](/fastapi/fastapi/pull/14186), Nice.

PR #14192, opened 2025-10-15
🐛 Fix ignoring background tasks added after `yield`

JP-Ellis:
This fixes a regression introduced in 0.118 whereby background tasks added _after_ a `yield` within a dependency don't get executed.

Resolves: #14137


JP-Ellis:
The issue with Python 3.12 and Pydantic should be resolved by #14186. Once that is merged, I will rebase this PR.

JP-Ellis:
Updated the test to ensure 100% coverage

ezeparziale:
Any update on this PR?

YuriiMotov (commented):
@JP-Ellis, thank you for your interest and efforts!

For me it looks incorrect that the logic of running background tasks is split into two parts\places - we run most of background tasks (those that were added before sending the response) in one place of code, but we run others (those that were added after `yield` in dependencies) in other place of code.

I think this should be solved differently - by running all background tasks in one place.

What do you think?

JP-Ellis:
I personally don't really care which way it is done, and am happy to defer to what people might expect.

I was under the impression that the current behaviour, that is, executing pre-yield background tasks early, was intentional and therefore there was a gap in the post-yield task, which this PR addresses.

If you want to have all background tasks executed post-yield, I can certainly amend this PR to do so.

YuriiMotov:
> If you want to have all background tasks executed post-yield, I can certainly amend this PR to do so

I personally think we should do it this way, but I haven't investigated how difficult it is to implement

JP-Ellis:
Great! I'll take a look at doing this in the next few days (don't have time today unfortunately).

JP-Ellis:
@YuriiMotov Looking into running all background tasks _after_, there would be some bigger changes required.

Background tasks in pre-yield dependencies get included as part of Starlette's background task handling, and get executed in [Starlette's `Response.__call__`](https://github.com/Kludex/starlette/blob/4941b4a04993d087dea66e9287fe9472babee879/starlette/responses.py#L155-L167):

```py
    prefix = "websocket." if scope["type"] == "websocket" else ""
    await send(
        {
            "type": prefix + "http.response.start",
            "status": self.status_code,
            "headers": self.raw_headers,
        }
    )
    await send({"type": prefix + "http.response.body", "body": self.body})


    if self.background is not None:
        await self.background()
```

When FastAPI runs the post-yield logic, the new background tasks get added to the same list, but there's no logic to eventually handle them.

I can see two ways forward:

1. Keep the PR as is, with some background tasks handled by Starlette, and some by FastAPI. Pragmatic, although perhaps not ideal.
2. Change the way background tasks are collected in FastAPI so that they are not passed to Starlette at all. I do not know if this will result in any other side-effects though.

YuriiMotov:
@JP-Ellis, thanks!
Then, let's wait for Sebastian to take a look and decide how we should proceed with this.

YuriiMotov:
I pushed some updates:
* merged master into PR's branch
* updated code to execute all bg tasks after sending response and exiting async stack
* updated tests - I think it's enough to test only async endpoint + async dependency. Also, I added asserts to ensure bg tasks executed only once and dependency is finalized at the moment of bg tasks execution

If you disagree with any of those changes - feel free to revert, I will then open alternative PR and we will ask Sebastian to decide.

The main question now is: do we want to allow users to use results of dependencies with yield in background tasks?
According to the docs, results of dependencies with yield are [not supposed to be used in background tasks](https://fastapi.tiangolo.com/advanced/advanced-dependencies/#background-tasks-and-dependencies-with-yield-technical-details). If it's not a mistake in docs, then we are on the right way

JP-Ellis:
Thanks for the update, looks good to me.

To answer you question:

> Do we want to allow users to use results of dependencies with yield in background tasks?

No, I don't think so. Starlette does not offer any way to access any results from background tasks, and I don't think FastAPI should deviate from that.

Even if you did somehow allow for the results to be accessed, where would that happen? It is executed after the async stack, so it is not available to any of the dependencies, nor the main handler for the endpoint.

tiangolo:
Thanks for the interest and work here! :coffee: 

I was checking https://github.com/fastapi/fastapi/discussions/14137 and replied there. I think your use case can be solved with the new Dependency scopes: https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/#early-exit-and-scope

---

I also need to review and improve all the docs around dependencies. :nerd_face: 

---

On the other hand, I think that there's a chance to improve further the idea of background tasks in a way that could go beyond what Starlette can offer by default, but that's something to explore in the future.

---

I think this specific use case would be solved by the new dependencies with scopes, so I'll pass on this PR, but thanks for the work and conversation everyone here! :cake:

JP-Ellis:
> I think this specific use case would be solved by the new dependencies with scopes, so I'll pass on this PR, but thanks for the work and conversation everyone here!

Unfortunately, in my particular use case, it is not. I have provided a more complete response to this in the linked discussion.

PR #14193, opened 2025-10-16
✨ Support Annotated dependencies in APIRoute, APIRouter, and FastAPI class

Profianton:
This PR adds support for `Annotated` dependencies in `APIRoute`, `APIRouter`, and the `FastAPI` class.

This allows for more flexible and expressive dependency definitions.


YuriiMotov (commented):
@Profianton, thanks for your interest!

This will make typing less strict and allow passing incompatible values. Look at the following code example:

```py
MyInt = Annotated[int, BeforeValidator(lambda x: x*10)]

router = APIRouter(dependencies=[MyInt])

@router.get("/")
def read_root(dep: MyInt):
    return {"dep": dep}
```
mypy will not argue on the code example above, but passing `MyInt` to `dependencies` is actually invalid.

At the same time I don't see any benefits it brings other than just the ability to reuse annotated dependencies that are used as parameters of functions.

So, without changes from this PR you would do:
```diff
ValueDep = Annotated[int, Depends(get_value)]

- router = APIRouter(dependencies=[ValueDep])
+ router = APIRouter(dependencies=[Depends(get_value)])

@router.get("/")
def read_root(dep: ValueDep):
    return {"dep": dep}
```

Profianton:
Hi @YuriiMotov,

Thanks for the feedback.

I agree with your first point.

However, I think the benefit of **reuseability** is a strong motivator here. Allowing the full `Annotated` definition to be used at the router/app level makes it far easier to reuse the `Depends`.


Thanks!

PR #14214, opened 2025-10-21
Update tutorial002.py

MikeiLL:
Replace deprecated `@app.on_event("startup")` with lifecycle event approach. 

There are a few more versions of this file in the repo with names like `tutorial002_an_py39.py`. I'm not sure what those are.

### DIFF
--- a/docs_src/sql_databases/tutorial002.py
+++ b/docs_src/sql_databases/tutorial002.py
@@ -1,4 +1,5 @@
 from typing import List, Union
+from contextlib import asynccontextmanager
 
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@MikeiLL, thanks for your interest!

We already have PR that attempts to solve this: https://github.com/fastapi/fastapi/pull/13253
And, here is my comment that describes why it's not so simple task: https://github.com/fastapi/fastapi/pull/13253#pullrequestreview-3047549301

MikeiLL:
Thank you very much. In the meantime, is the recommendation to just stick with the deprecated method?

PR #14217, opened 2025-10-22
✏️ Fix links and add missing permalink in docs

YuriiMotov:
Several minor fixes


mdrahat7120112054-hash:
Welcome you are choosing for the iPhone winner

PR #14224, opened 2025-10-24
Change "/" to "or" in docs->security->first steps

DanielYang59:
[The following](https://fastapi.tiangolo.com/tutorial/security/first-steps/#fastapis-oauth2passwordbearer) confused me for a second:
> This parameter doesn't create that endpoint / path operation, but declares that the URL /token will be the one that the client should use to get the token. That information is used in OpenAPI, and then in the interactive API documentation systems.

In this context `/` could be easily interpreted as "root", guess better to use "or"?

### DIFF
--- a/docs/en/docs/tutorial/security/first-steps.md
+++ b/docs/en/docs/tutorial/security/first-steps.md
@@ -144,7 +144,7 @@ Using a relative URL is important to make sure your application keeps working ev
 
 ///
 
-This parameter doesn't create that endpoint / *path operation*, but declares that the URL `/token` will be the one that the client should use to get the token. That information is used in OpenAPI, and then in the interactive API documentation systems.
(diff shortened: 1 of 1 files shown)


YuriiMotov:
"OR" here sounds like an alternative, but "path operation" is actually a clarification, not alternative.

I don't think we should change it (I don't find the version with slash confusing), but if we decide to do it, I think it should be something like:

> This parameter doesn't create that endpoint (*path operation*), but declares ...

DanielYang59:
Thanks for the quickly reply. From my understanding the docs means adding the `tokenUrl` param would not:
- create the path operation (HTTP method + path pattern + function handling the operation)
- nor create the endpoint (the function handling the route)

So as I understand it "path operation" doesn't really clarify "endpoint" (but I could likely be wrong here as I'm not very familar with fastapi terminologies).

Either case I think it's less confusing because when I quickly read through that docs I thought it means "This parameter doesn't create that endpoint _root_ path operation" and was confused why it has anything to do with the root path operation

Free feel to close this if you don't think it's an issue ;)

PR #14232, opened 2025-10-26
♻️ Simplify code in `fastapi/routing.py`

bokshitsky:
TLDR: just simplify internal logic and code a bit, doesn't change framework behaviour

split in in 4 commits:
1) remove unused variable initialization `errors: List[Any] = []`

2) simplify code: remove nested code, using early exit pattern

```
if errors:
  raise Exception()
```

3) instead of:
```
current_status_code = (
    status_code if status_code else solved_result.response.status_code
)
if current_status_code is not None:
    response_args["status_code"] = current_status_code
if solved_result.response.status_code:
    response_args["status_code"] = solved_result.response.status_code
```

write:
```
if solved_result.response.status_code:
    response_args["status_code"] = solved_result.response.status_code
elif status_code is not None:
    response_args["status_code"] = status_code
```

if remains exactly the old behaviour, but 
a) looks more clear (right now looks like `status_code if status_code else solved_result.response.status_code` is default value, but actually in the end default value is `solved_result.response.status_code`
b) it doesn't make any speed effect, but new version use less comparison and assignment operations

4) remove non used variable `response` initialization

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -463,9 +463,7 @@ def get_request_handler(
         errors = solved_result.errors
         assert dependant.call  # For types
         if errors:
-            raise RequestValidationError(
-                errors, body=body, endpoint_ctx=endpoint_ctx
-            )
+            raise RequestValidationError(errors, body=body, endpoint_ctx=endpoint_ctx)
 
         # Shared serializer for stream items (JSONL and SSE).
         # Validates against stream_item_field when set, then
@@ -520,9 +518,7 @@ def get_request_handler(
                         comment=item.comment,
                     )
 
-                return format_sse_event(
-                    data_str=_serialize_data(item).decode("utf-8")
-                )
+                return format_sse_event(data_str=_serialize_data(item).decode("utf-8"))
 
             if dependant.is_async_gen_callable:
                 sse_aiter: AsyncIterator[Any] = gen.__aiter__()
@@ -530,9 +526,7 @@ def get_request_handler(
                 sse_aiter = iterate_in_threadpool(gen)
 
             @asynccontextmanager
-            async def _sse_producer_cm() -> AsyncIterator[
-                ObjectReceiveStream[bytes]
-            ]:
+            async def _sse_producer_cm() -> AsyncIterator[ObjectReceiveStream[bytes]]:
                 # Use a memory stream to decouple generator iteration
                 # from the keepalive timer. A producer task pulls items
                 # from the generator independently, so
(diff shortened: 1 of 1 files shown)


fastapi/routing.py
```diff
@@ -372,7 +372,6 @@ async def app(request: Request) -> Response:
             raise http_error from e
 
         # Solve dependencies and run path operation function, auto-closing dependencies
-        errors: List[Any] = []
```
bokshitsky: variable is overwritten by `errors = solved_result.errors` a bit lower

fastapi/routing.py
```diff
@@ -277,6 +278,63 @@ async def serialize_response(
...
+    file_stack: AsyncExitStack[bool | None]
+) -> Optional[Union[Dict[str, Any], FormData, bytes]]:
+    # Read body and auto-close files
+    if not body_field:
+        return None
+    try:
+        is_body_form = body_field and isinstance(
+            body_field.field_info, (params.Form, temp_pydantic_v1_params.Form)
+        )
+        if is_body_form:
```
bokshitsky: doesn't change internal rules, just simplify code removing nesting

fastapi/routing.py
```diff
@@ -308,71 +366,20 @@ def get_request_handler(
 ) -> Callable[[Request], Coroutine[Any, Any, Response]]:
     assert dependant.call is not None, "dependant.call must be a function"
     is_coroutine = iscoroutinefunction(dependant.call)
-    is_body_form = body_field and isinstance(
```
bokshitsky: moved to get_request_body, where actually used

fastapi/routing.py
```diff
@@ -386,52 +393,41 @@ async def app(request: Request) -> Response:
...
-                response.headers.raw.extend(solved_result.response.headers.raw)
         if errors:
-            validation_error = RequestValidationError(
-                _normalize_errors(errors), body=body
-            )
-            raise validation_error
+            raise RequestValidationError(_normalize_errors(errors), body=body)
 
-        # Return response
-        assert response
```
bokshitsky, line 726: `assert response` was used to avoid forget response assignment inside if-else-branches, not required in proposed changes

bokshitsky:
@YuriiMotov hi!

I removed last commit (which extracted `get_request_body` function) since it requires more accurate refactoring and type-annotation updates (right now `body` [can be](https://github.com/tiangolo/fastapi/blob/030f01473ae61438abaf026545a42b8968a30245/fastapi/routing.py#L349-L349) of type `bytes`, but [solve_dependencies](https://github.com/tiangolo/fastapi/blob/030f01473ae61438abaf026545a42b8968a30245/fastapi/dependencies/utils.py#L611-L611) accepts only `body: Optional[Union[Dict[str, Any], FormData]]`)

So PR is ready, may you please remove `Draft` tag and help with review if you have time?

fastapi/routing.py
YuriiMotov: This was removed from this PR

YuriiMotov (approved):
LGTM!

With these changes code is more readable. Changes don't affect final result

@bokshitsky, thank you!

wu-clan (approved):

bokshitsky:
@YuriiMotov, I rebased code and fixed conficts, 
see some failing checks, not sure what's the reason, probably need rerun them, but I don't have rerun button in interface, do you have?
image

@tiangolo, may you please look, if this PR is acceptable and merge if everything ok?
I believe, since some merge-conflicts occured and the logic inside actually changes in new commits, the code is "live" and code simplification from that PR make sense

svlandeg:
@bokshitsky: unfortunately this PR has accumulated some non-trivial merge conflicts. Do you have time to look into them & resolve them by merging in the latest version of `master`?

bokshitsky:
@svlandeg, yes I rebased the branch on master, but I am not sure that anyone will merge current PR, since now the diff is much larger, and previous small diff have not been merge for half-year

tiangolo:

I want to implement a big refactor in how routers work.

Adding this change means exposing the possibility that it breaks someone's use case somewhere, and that risk is not worth now given there's that additional work coming.

Thank you! 🍰

PR #14233, opened 2025-10-26
♻️ Avoid creating an intermediate container in `get_value_or_default` for efficiency

arun-geek:
Refactoring to improve efficiency of the code

create a items tuple by combining first_items with extra_items and is not used in the reminder of the code. To make it more efficient we can check the items directly

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -247,8 +247,9 @@ def get_value_or_default(
 
     Otherwise, the first item (a `DefaultPlaceholder`) will be returned.
     """
-    items = (first_item,) + extra_items
-    for item in items:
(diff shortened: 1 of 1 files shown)


YuriiMotov (approved):
@arun-geek, thanks for your interest!

It was probably made this way to reduce the code duplication (`if not isinstance(first_item, DefaultPlaceholder)`).
The performance impact of this change is very tiny - this function is only called several times per route on application startup.

I don't think we should focus on such micro-optimizations, but let's Sebastian decide.

tiangolo:
Thanks for the interest! I think for now it's fine as it is, so I'll pass on this one. :coffee:

PR #14238, opened 2025-10-27
📝 Update the part of "Behind a proxy" section related to `root_path`

YuriiMotov:
There is a wrong statement in current version of docs:

> Passing the `root_path` to `FastAPI` would be the equivalent of passing the `--root-path` command line option to Uvicorn or Hypercorn.

This is described well in this discussion: https://github.com/fastapi/fastapi/discussions/14220

Here is my attempt to fix docs.

It will also resolve the [issues](https://stackoverflow.com/questions/79599624/fastapi-application-with-nginx-staticfiles-not-working) with mounted sub-apps and `StaticFiles` when app is run with `root_path` parameter passed to `FastAPI` - now it will be clear that running such configuration behind the **stripping path** proxy is not correct.

**Will resolve:**

<details>

* https://github.com/fastapi/fastapi/discussions/11153
* https://github.com/fastapi/fastapi/discussions/11147
* https://github.com/Kludex/starlette/discussions/2928
* https://github.com/Kludex/starlette/discussions/2557
* https://stackoverflow.com/questions/79599624/fastapi-application-with-nginx-staticfiles-not-working
* https://github.com/fastapi/fastapi/discussions/11033
* https://github.com/fastapi/fastapi/discussions/13590
* https://github.com/fastapi/fastapi/discussions/11959
* https://github.com/fastapi/fastapi/discussions/12151

</details>

---

Looking at this a few weeks later I realize that's quite a lot of changes (diff is big).
If you think we should try to minimize changes, I can try to create an alternative version with minimal changes


Pentusha:
Shouldn't this have been fixed on the starlette side?

simone-trubian:
Hi @YuriiMotov happy to help with the testing of the examples; for my understanding, which examples are you referring to?

YuriiMotov:
> Hi @YuriiMotov happy to help with the testing of the examples; for my understanding, which examples are you referring to?

Great!
We need to test all 3 approaches described in new version - configure Traefik using the config example provided for this approach, run ASGI server as it's described in that approach, send proper requests (with prefix included) and confirm that everything works as expected.

We can also modify the code example of the app to include sub-app and `StaticFiles` to make sure it all works correctly in all 3 configurations

KarelRuzicka (approved):
Looks great. Covers previously undocumented behaviour of root_proxy mentioned in #14220

YuriiMotov:
There is a problem with the third approach (**Using `X-Forwarded-Prefix` header**) - FastAPI uses cache for openapi schema, and `servers` property will only contain the server that corresponds to the `root_path` that was used for first request to that schema. So, for other `root_path`'s openapi will contain "wrong" server.

If we decide to support this approach, we will need to update openapi caching to use `root_path` as a cache key.

docs_src/behind_a_proxy/tutorial005.py
```diff
@@ -0,0 +1,31 @@
...
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+
+class ForwardedPrefixMiddleware:
+    def __init__(self, app: ASGIApp):
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
+        if scope["type"] in ("http", "websocket"):
+            scope_headers: list[tuple[bytes, bytes]] = scope.get("headers", [])
```
YuriiMotov: Will probably need to add variants for this code example.
Can't get why it doesn't fail on Python 3.8 🤔

Issue #12419, opened 2024-10-10
Question about "Required, can be None" parameter


Kludex:
### Discussed in https://github.com/fastapi/fastapi/discussions/12418

<div type='discussions-op-text'>

<sup>Originally posted by **bowenerchen** October 10, 2024</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/pydantic/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from typing import Annotated

from fastapi import FastAPI, Query

app = FastAPI()


@app.get("/items/")
async def read_items(q: Annotated[str | None, Query(min_length=3)] = ...):
    results = {"items": [{"item_id": "Foo"}, {"item_id": "Bar"}]}
    if q:
        results.update({"q": q})
    return results
```

### Description

https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#required-can-be-none
![QQ_1728534861112](https://github.com/user-attachments/assets/95e967aa-ea9f-46c1-80c8-f13dc7c52e6c)
how to pass a None value to the required query parameter q?

### Operating System

macOS

### Operating System Details

Sonoma 14.7 M2 Max

### FastAPI Version

0.115.0

### Pydantic Version

2.5.3

### Python Version

Python 3.12.4

### Additional Context

_No response_</div>

Kludex:
The documentation is wrong.

This is the file that is testing that part of the documentation: https://github.com/fastapi/fastapi/blob/529155e72e196f82e0289f173a185e8ddb2888ce/tests/test_tutorial/test_query_params/test_tutorial006_py310.py

As we can see, it doesn't pass the `None`...

Kludex:
This may be the same as https://github.com/fastapi/fastapi/issues/12245

zubairevangelist:
[...]

PR #14250, opened 2025-10-29
docs: clarify required query param that can be None

mahdighadiriii:
Fixes #12419
update docs to show how to make query param **required** but allow `None`
add tip: use `?q=null` to send `None`
add working example with string check
replace broken example with `tutorial006c_fixed.py`


mahdighadiriii:
@Kludex Ready for review!  
fixed the broken example and added clear tip: `?q=null` → `None`  
tested locally works in Swagger UI.

ivikasavnish (approved):

YuriiMotov:
See also: https://github.com/fastapi/fastapi/pull/14322

tiangolo:
Let's focus on https://github.com/fastapi/fastapi/pull/14322, I'll pass on this one for now, but thanks for the interest! :coffee:

PR #14254, opened 2025-10-30
♻️ Refactor internals of dependencies, simplify using dataclasses

tiangolo:
♻️ Refactor internals of dependencies, simplify using dataclasses

---

As dataclasses come with their own `repr()` there's no need for custom logic nor tests for that.


zmievsa:
@tiangolo this one broke Cadwyn because Depends and Security are no longer hashable. Please, see #14320

PR #14258, opened 2025-10-30
✨ Show a clear error on attempt to include router into itself

JavierSanchezCastro:
Previously discussed in https://github.com/fastapi/fastapi/discussions/11424 with the PR at https://github.com/fastapi/fastapi/pull/11425

Since I made modifications in the fork of the repo, the PR was automatically closed. This PR serves to add the circular reference check again. Discussed again in https://github.com/fastapi/fastapi/discussions/14251


tests/test_router_circular_import.py
```diff
@@ -0,0 +1,11 @@
...
+from fastapi import APIRouter, FastAPI
+
+
+def test_router_circular_import():
+    app = FastAPI()
+    router = APIRouter()
+
+    app.include_router(router)
+    with pytest.raises(AssertionError, match="Cannot include router into itself"):
+        router.include_router(router)
```
YuriiMotov, line 12:
```suggestion
import pytest
from fastapi import APIRouter


def test_router_circular_import():
    router = APIRouter()

    with pytest.raises(AssertionError, match="Cannot include router into itself"):
        router.include_router(router)
```

`app` is not needed here, and calling `router.include_router` after `app.include_router(router)` wrong

fastapi/routing.py
```diff
@@ -1333,6 +1333,7 @@ def read_users():
         app.include_router(internal_router)
         ```
         """
+        assert self is not router, "Cannot include router into itself"
```
```suggestion
        assert self is not router, (
            "Cannot include the same APIRouter instance into itself. "
            "Did you mean to include a different router?"
        )
```
How about this version?

YuriiMotov: Also, we should probably raise [`FastAPIError`](https://github.com/fastapi/fastapi/blob/1fc586c3a5ae457457c5c70adddb07c7e9185f16/fastapi/exceptions.py#L144) instead of `assert`.

UPD: maybe not. In other cases just `assert` is used..

YuriiMotov (approved):
LGTM!

@JavierSanchezCastro, thank you!

tiangolo (approved):
Makes sense, thank you! This will be available in FastAPI 0.128.7, released in the next few hours. 🚀

PR #14260, opened 2025-10-30
📝 Add EuroPython talk & podcast episode with Sebastián Ramírez

clytaemnestra:
I don't seem to be able to add a label myself - it should be `docs`.


YuriiMotov (approved):
LGTM

PR #14262, opened 2025-10-30
✨ Add support for dependencies with scopes, support `scope="request"` for dependencies with `yield` that exit before the response is sent

tiangolo:
✨ Add support for dependencies with scopes, support `scope="request"` for dependencies with `yield` that exit before the response is sent

This adds support for

* `Depends(func, scope="request")`, the default.
* `Depends(func, scope="function")`, early exit, after the function, but before sending the response back.

---

I'm considering an alternative name for the `scope` parameter of `mode`, to avoid any confusion with OAuth scopes, e.g. `Security(scopes=["blah"])`, which is a completely different idea.

Or a potential future `Depends(oauth_scopes=["blah"])` to replace `Security(scopes=["blah"])`.

That's the only reason to consider `mode` over `scope`.

---

I'm also considering the values, currently they are:

* `"function"`: start before the *path operation function*, exit after the function returns but **before** the response is sent.
* `"request"`: start before the *path operation function* (same as above), exit **after** the response is sent back.

I'm considering an alternative name for `"request"` of `"response"`, with the same behavior.

The rationale for using `"request"` as the value is that the dependency runs around (start and end) of the *request* cycle.

The rationale for using `"response"` as the value is that the main difference with `"function"` is when the exit code is run, in `"function"` it is run *after* the function is done, in this second form (`"request"` or `"response"`) it is run *after* the *response* is sent back.

---

I'm using this interface design because I also plan on (potentially, possibly) add a new `scope="lifespan"` that would allow using dependencies (or equivalent) functions that are run only once per app, and closed when the app ends. This way the same functionality for dependencies could be used for the app lifespan (ASGI `startup` and `shutdown` events). It should be more intuitive for users to work with. But these are future plans.

---

### 2025-10-03 Edit

After asking in several places, I concluded to keep the parameter name as `scope`.

I'll also keep the current values of `"function"`, similar to pytest, and `"request"`, to avoid confusing users who could think that the dependency will start running when the response is being sent, but it starts when the request starts.

### DIFF
--- a/docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
+++ b/docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
@@ -184,6 +184,51 @@ If you raise any exception in the code from the *path operation function*, it wi
 
 ///
 
+## Early exit and `scope` { #early-exit-and-scope }
+
+Normally the exit code of dependencies with `yield` is executed **after the response** is sent to the client.
+
+But if you know that you won't need to use the dependency after returning from the *path operation function*, you can use `Depends(scope="function")` to tell FastAPI that it should close the dependency after the *path operation function* returns, but **before** the **response is sent**.
+
+{* ../../docs_src/dependencies/tutorial008e_an_py39.py hl[12,16] *}
+
+`Depends()` receives a `scope` parameter that can be:
+
+* `"function"`: start the dependency before the *path operation function* that handles the request, end the dependency after the *path operation function* ends, but **before** the response is sent back to the client. So, the dependency function will be executed **around** the *path operation **function***.
+* `"request"`: start the dependency before the *path operation function* that handles the request (similar to when using `"function"`), but end **after** the response is sent back to the client. So, the dependency function will be executed **around** the **request** and response cycle.
+
+If not specified and the dependency has `yield`, it will have a `scope` of `"request"` by default.
+
+### `scope` for sub-dependencies { #scope-for-sub-dependencies }
+
+When you declare a dependency with a `scope="request"` (the default), any sub-dependency needs to also have a `scope` of `"request"`.
+
+But a dependency with `scope` of `"function"` can have dependencies with `scope` of `"function"` and `scope` of `"request"`.
+
+This is because any dependency needs to be able to run its exit code before the sub-dependencies, as it might need to still use them during its exit code.
+
+```mermaid
+sequenceDiagram
+
+participant client as Client
+participant dep_req as Dep scope="request"
+participant dep_func as Dep scope="function"
+participant operation as Path Operation
+
+    client ->> dep_req: Start request
+    Note over dep_req: Run code up to yield
+    dep_req ->> dep_func: 
+    Note over dep_func: Run code up to yield
+    dep_func ->> operation: Run path operation
+    operation ->> dep_func: Return from path operation
+    Note over dep_func: Run code after yield
+    Note over dep_func: ✅ Dependency closed
+    dep_func ->> client: Send response to client
+    Note over client: Response sent
+    Note over dep_req: Run code after yield
+    Note over dep_req: ✅ Dependency closed
+```
+
 ## Dependencies with `yield`, `HTTPException`, `except` and Background Tasks { #dependencies-with-yield-httpexception-except-and-background-tasks }
 
 Dependencies with `yield` have evolved over time to cover different use cases and fix some issues.
--- a/docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
+++ b/docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
@@ -217,7 +217,7 @@ participant operation as Path Operation
 
     client ->> dep_req: Start request
     Note over dep_req: Run code up to yield
-    dep_req ->> dep_func: 
+    dep_req ->> dep_func:
     Note over dep_func: Run code up to yield
     dep_func ->> operation: Run path operation
     operation ->> dep_func: Return from path operation
--- a/docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
+++ b/docs/en/docs/tutorial/dependencies/dependencies-with-yield.md
@@ -217,9 +217,9 @@ participant operation as Path Operation
 
     client ->> dep_req: Start request
     Note over dep_req: Run code up to yield
-    dep_req ->> dep_func:
+    dep_req ->> dep_func: Pass dependency
     Note over dep_func: Run code up to yield
-    dep_func ->> operation: Run path operation
+    dep_func ->> operation: Run path operation with dependency
     operation ->> dep_func: Return from path operation
     Note over dep_func: Run code after yield
     Note over dep_func: ✅ Dependency closed
--- a/docs_src/dependencies/tutorial008e.py
+++ b/docs_src/dependencies/tutorial008e.py
@@ -1,0 +1,15 @@
+from fastapi import Depends, FastAPI
+
+app = FastAPI()
+
+
+def get_username():
+    try:
+        yield "Rick"
+    finally:
+        print("Cleanup up before response is sent")
+
+
+@app.get("/users/me")
+def get_user_me(username: str = Depends(get_username, scope="function")):
+    return username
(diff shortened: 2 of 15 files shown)


potiuk:
Small comment - I am not sure how well I can comment on the code itself :) as I am vaguely familiar with it, but I think I can comment on the UX of it. 

* scope is way better choice than mode, more natural. And I think there is very little probability of people confusing `Dependency(scope=` with `Security(scope=`. Having both as keyword parameters and practically always being specified right after the class, namespaces them properly already. It very nicely reads as "dependency_scope" and "security_scope" - following nicely "literate programming" principles by Donald Knuth.

* I think maybe a better and more consistent naming for scopes would be "request" and "response" - where request would be an "early" exit, and "response" would be "late" exit after response is prepared. This is - I think way more natural and describes the expected behaviour way more intuitively for my taste. Having "function" and "request/response" as two scopes somehow mixes the abstraction layers and it's not too intuitive. 

UPDATE: comment - I have not spent a lot of time with FastAPI so my suggestions might be missing some of the context, this is mostly based on my general programming intuition.

Danstiv:
Hello!

Let's consider the following pseudo-example:

```python
def db_dep():
    # some code to create and yield a sqlalchemy session
    ...

def chat_repo_dep(db_session = Depends(db_dep)):
    # some code to create a chat repository
    ...

@router.post("/")
def create_chat(repo: ChatRepository = Depends(chat_repo_dep)):
    ...

@router.get("/")
def get_chat(repo: ChatRepository = Depends(chat_repo_dep)):
    ...

@router.patch("/")
def update_chat(repo: ChatRepository = Depends(chat_repo_dep)):
    ...

@router.post("/generate")
def generate_llm_response(repo: ChatRepository = Depends(chat_repo_dep)) -> StreamingResponse:
    # some code that returns a StreamingResponse
    ...
```

So, we have a dependency with a SQLAlchemy session, and also a chat repository dependency that requires the SQLAlchemy session.

We have many regular endpoints that return a response all at once, and the behavior for them is fine.
However, some endpoints might return a `StreamingResponse`, and the response iterator won't be able to work with the chat repository because it depends on the DB session, which was closed when `generate_llm_response` returned the `StreamingResponse`.

Also, we cannot influence this behavior when declaring the dependency, since `generate_llm_response` depends on the chat repository, not on `db_session` directly.

The only solution I've come to is to create a separate repository for `StreamingResponse` and use it in the iterator.

Therefore, it might be worth considering something like a `StreamingResponse` with dependencies.

YuriiMotov:
If I got the idea of use case described by @Danstiv correctly, then the solution could be:
* use `function` as a default scope for top level dependencies (that are not used as sub-dependencies anywhere)
* for sub-dependencies use the `scope` of their outer dependency

Code example to illustrate:
```py
def db_dep():
    # some code to create and yield a sqlalchemy session
    ...

def chat_repo_dep(db_session = Depends(db_dep)):
    # some code to create a chat repository
    ...

@router.post("/")
def create_chat(repo: ChatRepository = Depends(chat_repo_dep, scope="function")):
    ...

@router.post("/generate")
def generate_llm_response(repo: ChatRepository = Depends(chat_repo_dep)) -> StreamingResponse:
    # some code that returns a StreamingResponse
    ...
```

Since we don't specify `scope` for `chat_repo_dep` in `generate_llm_response`, the scope for it and all sub-dependencies will be `request`.
But in `create_chat` we specified `scope="function"` and it will be used as a default for all sub-dependencies

ashb:
I personally think something like "response" or "response-sent" would be more accurate than "request", since at least in my head it's the request-response cycle, and the request phase is over when the request has been sent. 

I realise this likely isn't a common pov. Just my 2p

Another option than scope might be to name the parameter something like `finalize_after`. Pro: more explicit what it's controlling for this case. Con: doesn't generalise to app lifecycle.

tiangolo:
This will be available in version `0.121.0`, released in the next hours. :tada: 

---

Thanks for the feedback @potiuk! :raised_hands:

I concluded I'll leave the current names, more details in the description.

---

@Danstiv it seems to me that the issue you are describing was the behavior before FastAPI [0.118.0](https://fastapi.tiangolo.com/release-notes/#01180). Please check the release notes and update your FastAPI version, that should be working already.

This PR is about allowing people to opt in to that old behavior in a case-by-case basis.

If you still have issues, please create a new discussion question following the template.

And thanks @YuriiMotov! :coffee: 

---

Thanks for the feedback @ashb! I decided to keep the names as they are in this PR, more details in the description.

potiuk:
Thanks - whatever decision, it's good to have the feature  :)

pierrejeambrun:
Just to mention that moving out `is_coroutine_callable` `is_gen_callable` etc from `dependencies.utils` broke some other third parties libraries. (Not sure if those were part of the public interface though, or if we have anything like this for `initial development` version of FastAPI)

For instance https://github.com/zmievsa/cadwyn.

tiangolo:
Thanks @pierrejeambrun, that is not part of a public API for FastAPI, is not documented.

It would be best if they added a feature request with an official way for them to change the things they need, explaining why is it needed to use and change internals of FastAPI.

YuriiMotov:
I still think that the [comment of Danstiv](https://github.com/fastapi/fastapi/pull/14262#issuecomment-3476579028) was valid. Consider the following code example. Imagine we have repository dependency that depends on `db_dep`. Currently we can't configure it to close session before sending the response for endpoints that don't need to interact with DB in response generator. The only way is to have two `chat_repo_dep` that use different scope for `db_dep`.
But now I think it's not so easy to solve..

<details>

```py
import asyncio
from datetime import datetime
from typing import Annotated

# Classes

class Session:
    def __init__(self):
        self.status = "closed"

    def __enter__(self):
        self.status = "open"
        print(f"open session ({datetime.now()})")
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.status = "closed"
        print(f"close session ({datetime.now()})")

class ChatRepository:
    def __init__(self, session: Session):
        self.session = session


# Dependencies

def db_dep():
    with Session() as session:
        yield session

def chat_repo_dep(db_session: Annotated[Session, Depends(db_dep)]):
    yield ChatRepository(session=db_session)


# Endpoints

app = FastAPI()

@app.get("/generate-needs-session")
def generate_llm_response_1(
    repo: Annotated[ChatRepository, Depends(chat_repo_dep)],
):
    async def gen_response(repo: ChatRepository):
        for i in range(5):
            yield f"response part {i}"
            await asyncio.sleep(0.1)

    return StreamingResponse(gen_response(repo=repo), media_type="text/plain")


@app.get("/generate-doesnt-need-session")
def generate_llm_response_2(
    repo: Annotated[ChatRepository, Depends(chat_repo_dep, scope="function")],
):
    async def gen_response():
        for i in range(5):
            yield f"response part {i}"
            await asyncio.sleep(0.1)

    return StreamingResponse(gen_response(), media_type="text/plain")
```

</details>


One more thing - we may have situations when regular dependency (with return) and `scope="request"` depends on dependency with `yield` which scope is `function`. This will not raise errors on app instantiation, but  may lead to runtime errors if the session is saved and used in streaming generator.
This is actually a user error (they should use dependency with `yield`), but it's easy to forget this and not so easy to debug.

<details>

```py
import asyncio
from datetime import datetime
from typing import Annotated

# Classes

class Session:
    def __init__(self):
        self.status = "closed"

    def __enter__(self):
        self.status = "open"
        print(f"open session ({datetime.now()})")
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.status = "closed"
        print(f"close session ({datetime.now()})")

class ChatRepository:
    def __init__(self, session: Session):
        self.session = session


# Dependencies

def db_dep():
    with Session() as session:
        yield session

def chat_repo_dep(db_session: Annotated[Session, Depends(db_dep, scope="function")]):
    return ChatRepository(session=db_session)  # Oops.. return instead of yield


# Endpoints

app = FastAPI()

@app.get("/generate-needs-session")
def generate_llm_response_1(
    repo: Annotated[ChatRepository, Depends(chat_repo_dep)],
):
    async def gen_response(repo: ChatRepository):
        for i in range(5):
            assert repo.session.status == "open"  # This will fail
            yield f"response part {i}"
            await asyncio.sleep(0.1)

    return StreamingResponse(gen_response(repo=repo), media_type="text/plain")
```
    
    
</details>

PR #14266, opened 2025-10-31
🐛 Fix security schemes in OpenAPI when added at the top level app

YuriiMotov:
🐛 Fix security schemes in OpenAPI when added at the top level app

Ref: https://github.com/fastapi/fastapi/discussions/14263
Ref: https://github.com/fastapi/fastapi/issues/14271


tiangolo (approved):
Good catch, awesome, thank you! 🚀 🙌 

I added a test based on the code examples from the report.

This will be available in FastAPI version 0.120.3, released in the next minutes/hours. 🎉

PR #14281, opened 2025-11-01
✨ Improve error messages for incorrect `Depends()` usage

timihack:
# Improve error messages for incorrect Depends() usage

## Description

This PR improves the developer experience when `Depends()` is used incorrectly by providing clear, actionable error messages instead of cryptic TypeErrors from deep in the stack.

## Problem

When new FastAPI users misuse the `Depends()` function, they encounter confusing error messages like:
- `TypeError: 'dict' object is not callable`
- `AttributeError: 'Depends' object has no attribute...`

These errors come from Starlette or Pydantic internals and don't clearly explain what went wrong or how to fix it.

## Solution

Added validation in the `Depends.__init__()` method to catch common mistakes early and provide helpful error messages.

### Common mistakes now caught:

**1. Calling the dependency function:**
```python
# ❌ Wrong
@app.get("/")
def route(user = Depends(get_user())):
    pass

# ✓ Correct
@app.get("/")
def route(user = Depends(get_user)):
    pass
```

**2. Nesting Depends() calls:**
```python
# ❌ Wrong
@app.get("/")
def route(val = Depends(Depends(get_value))):
    pass
```

**3. Passing non-callable values:**
```python
# ❌ Wrong
@app.get("/")
def route(val = Depends("not_a_function")):
    pass
```

### Modified `fastapi/params.py`:
- Added validation logic in `Depends.__init__()`
- Checks for nested `Depends()` calls
- Checks if dependency is callable
- Provides specific error messages with examples

### Added `tests/test_depends_validation.py`:
- Tests for all error cases (nested Depends, called functions, non-callables)
- Tests for correct usage still working
- Tests for error message quality and helpfulness
- Tests for callable classes and edge cases

### Before:
```
TypeError: 'dict' object is not callable
```

### After:
```
TypeError: Depends() expects a callable (function or class), but received dict.

It looks like you may have called the dependency function instead of passing it.
✓ Correct: Depends(my_function)
✗ Wrong: Depends(my_function())
```

## Benefits

- ✅ **Fails fast** - Catches errors at route definition time, not runtime
- ✅ **Clear guidance** - Shows exactly what's wrong and how to fix it
- ✅ **Minimal change** - Only modifies one method in `fastapi/params.py`
- ✅ **No breaking changes** - Correct usage is completely unaffected
- ✅ **Well tested** - Comprehensive test coverage for all scenarios
- ✅ **Improves DX** - Significantly better experience for new users

## Testing

- [x] Added comprehensive test coverage (`tests/test_depends_validation.py`)
- [x] All existing tests pass
- [x] Manual testing confirms improved error messages
- [x] Code formatted with Black and isort
- [x] Type checking passes

## Checklist

- [x] Clear, actionable error messages with examples
- [x] Catches errors at route definition time (early)
- [x] Maintains backward compatibility (correct usage unaffected)
- [x] Includes helpful examples in error messages (✓/✗ format)
- [x] Comprehensive test coverage (8 test cases)
- [x] Follows FastAPI code style and conventions

## Related Issues

This addresses a common pain point for new users learning FastAPI's dependency injection system. Related discussions:
- Users encountering cryptic `Depends()` errors in community channels
- Multiple Stack Overflow questions about these error messages

This enhancement aligns with FastAPI's philosophy of excellent developer experience and will help onboard new users more smoothly.


fastapi/params.py
```diff
@@ -767,6 +763,47 @@ class Depends:
...
+                error_msg = f"Depends() expects a callable (function or class), but received {dep_type}. "
+
+                # Add specific hints for common mistakes
+                if isinstance(dependency, (dict, list, str, int, float, bool)):
+                    error_msg += (
+                        "\n\nIt looks like you may have called the dependency function instead of passing it. "
+                        "\n✓ Correct: Depends(my_function)"
+                        "\n✗ Wrong: Depends(my_function())"
+                    )
+                elif callable(dependency):
```
YuriiMotov: This condition can't be True since we are in the `if not callable(dependency):` branch

svlandeg (commented):
Hi @tImIhAcK, was this PR vibe coded?

timihack:
Hi @svlandeg,
Not at all, this was carefully implemented.
The goal was to improve developer experience by validating invalid Depends() usages (like nested calls or passing non-callables) and providing explicit, friendly error hints.

fastapi/params.py
timihack: Good catch! You were right, that condition was redundant. I’ve removed it and cleaned up the logic in the latest commit. Thanks for pointing that out!

tiangolo:
I don't think this is currently needed, so I'll pass on this for now. :coffee:

PR #14282, opened 2025-11-01
⬆️ Bump Starlette to <`0.51.0`

musicinmybrain:
The key change in [Starlette 0.50.0](https://github.com/Kludex/starlette/releases/tag/0.50.0) is that it drops support for Python 3.9, which has now reached end-of-life.


svlandeg (approved):
Thanks!

PR #14293, opened 2025-11-04
✅ Add tests for body fields aliases

YuriiMotov:
### Tests for Pydantic V2 include:
* Basic alias functionality `Form(alias=...)` + the same for form model
* Validation alias functionality `Form(validation_alias=...)` + the same for form model
* Combined aliases `Form(alias=..., validation_alias=...)` + the same for form model
* Workaround `Form(alias=X, validation_alias=X)`

Each of test cases (except workaround) covers different field types:
* Required `str` field
* Optional `str` field
* `list` field
* Optional `list` field

Each of test cases covers different parameter types:
* `Form` field
* `Form` field declared as model (think about adding tests for `alias_generator`, `populate_by_name`, `alias_priority`, `alias_choices`)
* `File` field
* `File` field declared as model
* `Body` field
* field declared in `Body` model

For each type of parameter declaration, the following is checked:
* Send parameter by name
* Send parameter by `alias`
* Send parameter by `validation_alias` (if specified)
* Generate openapi schema

### Tests for Pydantic V1

Tests for Pydantic V1 duplicate tests for V2, excluding all tests with `validation_alias` (it's not implented for V1 and I don't think we need to add this)

### Currently tests expose several issues with Pydantic V2:
* Single fields (not model):
    * `alias` - for Form\File\Body fields works partially - isn't reflected in schema
    * `validation_alias`- for Form\File\Body fields is reflected in schema, but doesn't work
    * optional lists of files cause `TypeError` in `serialize_sequence_value` (fix - [PR #14297](https://github.com/fastapi/fastapi/pull/14297))
* Fields declared as model:
    * Pydantic Models with `validation_alias` for List of Form\File Fields fails due to some bug (`"Input should be a valid list","input":"456"`) - (fix [PR 14303](https://github.com/fastapi/fastapi/pull/14303))
    * parameters with `validation_alias` work, but they go through wrong path during value extraction phase - they are extracted as `extra` parameters
    * For `File` fields, it fails with validation error `"msg":"Input should be a valid bytes","input":{"filename":"upload","file":{...`

### Currently tests expose 1 issue with Pydantic V1:
* Same validation error for `File` fields: `{"detail":[{"loc":["body","file_alias"],"msg":"byte type expected","type":"type_error.bytes"}]}`

### Notes

* This PR includes 2 commits form [another PR](https://github.com/fastapi/fastapi/pull/14297) that fixes the bug with handling optional lists of bytes
  * [commit 1](https://github.com/fastapi/fastapi/pull/14293/commits/751935c6df540dd167d28b7b0ea394c72b8e6905)
  * [commit 2](https://github.com/fastapi/fastapi/pull/14293/commits/04c956b624c73a599d262325acaa5573ed06f5b8)

### Swagger uses `style="form", explode=False` by default to send form data, should use `explode=True`

Swagger UI sends multiple values comma-separated: `param=value1,value2`.

As a workaround, we can specify encoding style manually in openapi:

<details>

```py
@app.post(
    "/alias",
    openapi_extra={
        "requestBody": {
            "content": {
                "application/x-www-form-urlencoded": {
                    "encoding": {"param_val_alias": {"style": "form"}},
                }
            }
        }
    },
)
async def list_field_alias_model(data: ListFieldAliasModel = Form(...)):
    return {"param": data.param}
```

</details>

Links: https://github.com/swagger-api/swagger-ui/issues/7695, https://github.com/swagger-api/swagger-ui/issues/3494, https://github.com/swagger-api/swagger-ui/issues/10221

### python-multipart doesn't handle comma-separated form parameters

As stated above, Swagger sends multiple values for Form parameter as comma-separated list, but Starlette doesn't handle it and treats it as one string.

ToDo: search for related issue on Starlette repo


YuriiMotov:
Moved to https://github.com/fastapi/fastapi/pull/14358

PR #14297, opened 2025-11-05
🐛 Fix optional sequence handling in `serialize sequence value` with Pydantic V2

YuriiMotov:
Currently the following test code leads to `TypeError: issubclass() arg 1 must be a class` (with Python 3.14 error is different, but the root cause is the same):

```py
from typing import List, Optional

from fastapi import FastAPI, File

app = FastAPI()


@app.post("/files")
async def upload_files(files: Optional[List[bytes]] = File(None)):
    if files is None:
        return {"files_count": 0}
    return {"files_count": len(files), "sizes": [len(f) for f in files]}


def test_optional_bytes_list():
    client = TestClient(app)
    response = client.post(
        "/files",
        files=[("files", b"content1"), ("files", b"content2")],
    )
    assert response.json() == {"files_count": 2, "sizes": [8, 8]}
```

This is because for optional sequence the `origin_type` in the following line is `Union`:

https://github.com/fastapi/fastapi/blob/972a967d5d3fa24d450637c36c6b7ad1f1d8c1b6/fastapi/_compat/v2.py#L374

So, we need to unwrap that `Uinion`, take non-None type from its arguments and use it as `origin_type`


tiangolo (approved):
Nice, thank you @YuriiMotov! 🙌 

This will be available in FastAPI 0.123.3, released in the next hours. 🚀

PR #14299, opened 2025-11-05
Add PydanticJSONResponse

eltoder:
When using this response class, json serialization is done using Pydantic's built-in json serialization (`dump_json(r)`) instead of generating an intermediate dict that is later serialized using a json library (`json.dumps(dump_python(r))`).

In my testing this is 3-4x faster than using the standard json library (the default) and 50% faster than using orjson, without requiring any extra dependencies. This also allows configuring serialization behavior per model using Pydantic's model_config.


eltoder:
A simple benchmark script:
```python
from datetime import date, timedelta
import json, orjson, ujson
import timeit
from pydantic import BaseModel, TypeAdapter

class Response(BaseModel):
    values: dict[int, dict[date, float]]

def make_response(n: int, m: int) -> Response:
    base = date(2025, 1, 1)
    values = {
        i + 1: {base + timedelta(i + j): (i + j) / 12.3 for j in range(m)}
        for i in range(n)
    }
    return Response(values=values)

ta = TypeAdapter(Response)
r = make_response(2, 20)
cases = {
    "json": "json.dumps(ta.dump_python(r, mode='json'), ensure_ascii=False).encode()",
    "ujson": "ujson.dumps(ta.dump_python(r, mode='json'), ensure_ascii=False).encode()",
    "orjson": "orjson.dumps(ta.dump_python(r, mode='json'), option=orjson.OPT_NON_STR_KEYS)",
    "pydantic": "ta.dump_json(r)",
}
for case, code in cases.items():
    print(case, min(timeit.repeat(code, number=10000, globals=globals())))
```
Results:
```
json 0.23686759299016558
ujson 0.12213927501579747
orjson 0.09217712600366212
pydantic 0.06348165299277753
```

fastapi/routing.py
```diff
@@ -229,11 +229,13 @@ async def serialize_response(
     exclude_defaults: bool = False,
     exclude_none: bool = False,
     is_coroutine: bool = True,
+    to_json: bool = False,
 ) -> Any:
     if field:
         errors = []
         if not hasattr(field, "serialize"):
             # pydantic v1
+            assert not to_json, "PydanticJSONResponse requires pydantic v2"
```
```suggestion
            assert not to_json, "PydanticJSONResponse only works with pydantic v2 models"
```
It's possible to have Pydantic V2 installed, but use V1 models (from pydantic.v1 import BaseModel).
This way the message will be clearer about it

tests/test_pydantic_json_response_class.py
```diff
@@ -0,0 +1,39 @@
...
+from fastapi.responses import PydanticJSONResponse
+from fastapi.testclient import TestClient
+from pydantic import BaseModel
+
+from .utils import needs_pydanticv2
+
+app = FastAPI(default_response_class=PydanticJSONResponse)
+
+
+class Response(BaseModel):
```
YuriiMotov:
```suggestion
class ResponseModel(BaseModel):
```

tests/test_pydantic_json_response_class.py
```diff
@@ -0,0 +1,39 @@
...
+    today: date
+    numbers: List[float]
+
+
+@app.get("/basic_response")
+def get_basic_response() -> Response:
+    return Response(
```
YuriiMotov:
```suggestion
def get_basic_response() -> ResponseModel:
    return ResponseModel(
```

`Response` is very confusing, it took me some time to notice that this is not `Response` imported from `fastapi` but user-defined model

YuriiMotov (commented):
@eltoder, thanks you for working on this!

Looks good to me in general, but I think we should add a bit more tests:
* ~~test that it raises with Pydantic V1 models (when V1 installed or with `from pydantic.v1 import BaseModel`)~~ (done)
* I would add some tests with different data and values of `response_model_**` parameters to show the resulted JSON comparing to result with regular `Response` class. Results should be equal except some cases when `json.dumps` raises but Pydantic handles it without error.
* show that response is validated using response model

YuriiMotov:
Idea of how tests can be organized:

<details>

```py
import math
from datetime import date
from typing import List

import pytest
from fastapi import FastAPI
from fastapi.responses import PydanticJSONResponse

regular_response_app = FastAPI()
pydantic_response_app = FastAPI(default_response_class=PydanticJSONResponse)


class ResponseModel(BaseModel):
    today: date
    numbers: List[float]


@regular_response_app.get("/date-inf-nan")
@pydantic_response_app.get("/date-inf-nan")
def get_inf_nan() -> ResponseModel:
    return ResponseModel(
        today=date(2025, 11, 1),
        numbers=[3.14, math.inf, math.nan, 2.72],
    )


regular_response_app_client = TestClient(regular_response_app)
pydantic_response_app_client = TestClient(pydantic_response_app)


def test_date_inf_nan_regular_response():
    with pytest.raises(  # json.dumps fails to serialize inf/nan
        ValueError, match="Out of range float values are not JSON compliant"
    ):
        _response = regular_response_app_client.get("/date-inf-nan")


def test_date_inf_nan_pydantic_response():
    response = pydantic_response_app_client.get("/date-inf-nan")
    assert response.json() == {
        "today": "2025-11-01",
        "numbers": [3.14, None, None, 2.72],  # pydantic converts inf/nan to None
    }
```

</details>

eltoder:
@YuriiMotov thanks for taking a look! This PR was intentionally minimal just to demonstrate the idea. I'm happy to add more tests and update documentation if the fastapi team agrees on the approach, especially since this is a user visible feature.

eltoder:
@YuriiMotov I addressed your comments. Please take another look.

YuriiMotov:
I forgot to inform - we discussed this idea with Sebastian and he said he'd like to take a look at this later as he has his own thoughts regarding the implementation. So, let's wait.

eltoder:
@YuriiMotov sounds good. thanks for letting me know

eltoder:
Hi @YuriiMotov! Do you have any ETA on when this is going to be reviewed?

YuriiMotov:
> Hi @YuriiMotov! Do you have any ETA on when this is going to be reviewed?

Unfortunately no ETA

YuriiMotov:
Since FastAPI [0.130.0](https://github.com/fastapi/fastapi/releases/tag/0.130.0) if you specify response model and don't specify response class, FastAPI bypasses `json_serializer` and serialize response data on Pydantic's side.

See: https://github.com/fastapi/fastapi/pull/14962

So, this PR is not needed anymore and may be closed.

@eltoder, thanks for your efforts!

eltoder:
@YuriiMotov what if I need a custom content type?

YuriiMotov:
> @YuriiMotov what if I need a custom content type?

I see what you mean..

As I see it, the ability to bypass `jsonable_encoder` would be a solution here (something like `disable_jsonable_encoder` parameter. Or even `disable_jsonable_encoder` attribute of response class), right?
This would enable the ability to use your own response class and have full control over the serialization.

This would also solve use cases like https://github.com/fastapi/fastapi/discussions/14929

eltoder:
I agree that this should be controlled by the response class. This is what I did in this PR. Instead of a boolean attribute this is usually done with a marker class like `NoJsonableEncoderResponse` which one can inherit into their response classes. This is similar to [the current solution](https://github.com/fastapi/fastapi/discussions/14929#discussioncomment-15879330), but will be more ergonomic.

Back on the original subject. Note that automatically enabling pydantic json serialization like in #14962 is a backwards incompatible change. Pydantic's serialization behavior is different from the previous behavior.

eltoder:
@YuriiMotov it appears that #14962 was reverted. Please re-open this PR.

EDIT: I take it back. The revert commit https://github.com/fastapi/fastapi/commit/a0a2d841719fbcf70d11ccb73eba19e2c53e8fc7 is not on any branch.

YuriiMotov:
> @YuriiMotov it appears that #14962 was reverted. Please re-open this PR.
> 
> EDIT: I take it back. The revert commit [a0a2d84](https://github.com/fastapi/fastapi/commit/a0a2d841719fbcf70d11ccb73eba19e2c53e8fc7) is not on any branch.

It was in [this PR](https://github.com/fastapi/fastapi/pull/14971) just to test performance impact one more time. That PR was closed

YuriiMotov:
> Back on the original subject. Note that automatically enabling pydantic json serialization like in #14962 is a backwards incompatible change. Pydantic's serialization behavior is different from the previous behavior.

I think I agree it might be breaking, but could you please share a use case you have in mind?

PR #14301, opened 2025-11-06
🐛 Fix `Depends(func, scope='function')` for top level (parameterless) dependencies

luzzodev:
**Summary**

This PR enables proper support for `scope="function"` on `APIRoute` dependencies by passing the dependency’s scope through to `get_dependant`. Previously, parameter-less dependencies created via `get_parameterless_sub_dependant` ignored `depends.scope`, which led to unexpected behavior for function-scoped dependencies.

**Motivation**

Users expect `scope="function"` to work as defined, even when registered at the route level. Because `depends.scope` wasn’t forwarded, function scope behaved like the default(request). This change aligns behavior with user expectations and documentation around dependency scopes.
(Related: #14296)


YuriiMotov:
Hi! 👋

I think the test logic looks a bit complicated. Can we try to simplify it?

My suggestion:

<details>

```py
from typing import Any

from fastapi import APIRouter, Depends, FastAPI, HTTPException

app = FastAPI()


def raise_after_yield() -> Any:
    yield
    raise HTTPException(status_code=503, detail="Exception after yield")


router = APIRouter()


@router.get("/")
def get_index():
    return {"status": "ok"}


app.include_router(
    prefix="/router-scope-function",
    router=router,
    dependencies=[Depends(raise_after_yield, scope="function")],
)

app.include_router(
    prefix="/router-scope-request",
    router=router,
    dependencies=[Depends(raise_after_yield, scope="request")],
)


client = TestClient(app)


def test_router_level_dep_scope_function() -> None:
    response = client.get("/router-scope-function/")
    assert response.status_code == 503
    assert response.json() == {"detail": "Exception after yield"}


def test_router_level_dep_scope_request() -> None:
    client = TestClient(app, raise_server_exceptions=False)
    response = client.get("/router-scope-request/")
    assert response.json() == {"status": "ok"}


def test_app_level_dep_scope_function() -> None:
    app = FastAPI(dependencies=[Depends(raise_after_yield, scope="function")])

    @app.get("/app-scope-function")
    def get_app_scope_function():
        return {"status": "ok"}

    client = TestClient(app)

    response = client.get("/app-scope-function")
    assert response.status_code == 503
    assert response.json() == {"detail": "Exception after yield"}


def test_app_level_dep_scope_request() -> None:
    app = FastAPI(dependencies=[Depends(raise_after_yield, scope="request")])

    @app.get("/app-scope-request")
    def get_app_scope_request():
        return {"status": "ok"}

    client = TestClient(app, raise_server_exceptions=False)

    response = client.get("/app-scope-request")
    assert response.json() == {"status": "ok"}
```

</details>

Also, I think we should put these tests into existing `test_dependency_yield_scope.py` and `test_dependency_yield_scope_websockets.py`

What do you think?

YuriiMotov:
How about changing the title to
````
🐛 Fix `Depends(func, scope='function')` for top level (parameterless) dependencies
````
?

luzzodev:
I totally agree with all suggestions. I was unsure to use dependencies on include_router  but seems reasonable. I’ll update the PR later this evening 😁

YuriiMotov (approved):
LGTM!

Thanks!

tiangolo (approved):
Awesome! Great job, thank you @luzzodev! 🚀 

And thanks a lot @YuriiMotov! 🙌 

This will be available in the next hours in FastAPI version 0.121.1 🎉

PR #14303, opened 2025-11-06
🐛 Fix parsing extra `Form` parameter list

YuriiMotov:
## Description

Currently if we declare Form parameters as model with `extra="allow"` and then pass extra parameter with multiple values, only last value of such parameter will be passed to model validator:

```py

app = FastAPI()


class Model(BaseModel):
    param: str

    model_config = {"extra": "allow"}

@app.post("/")
async def model_with_extra(data: Model = Form()):
    return data


def test_pass_extra_list():
    client = TestClient(app)
    resp = client.post(
        "/",
        data={
            "param": "123",
            "param2": ["456", "789"],  # Pass a list of values as extra parameter
        },
    )
    assert resp.json() == {
        "param": "123",
        "param2": ["456", "789"],
    }
    # AssertionError: assert
    # {'param': '123', 'param2': '789'}
    # ==
    # {'param': '123', 'param2': ['456', '789']}
```

The part of code that adds extra parameters is:

https://github.com/fastapi/fastapi/blob/972a967d5d3fa24d450637c36c6b7ad1f1d8c1b6/fastapi/dependencies/utils.py#L901-L903

The problem is in `received_body.items()` - it returns only last value for each key.
To fix this we need to iterate over all keys in `received_body` and get the values using `received_body.getlist(key)`:

```py
    for key in received_body.keys():
        if key not in values:
            param_values = received_body.getlist(key)
            if len(param_values) == 1:
                values[key] = param_values[0]
            else:
                values[key] = param_values
```

## Alternative implementation

This solution parses single value as string

`extra=string` -> `{"extra": "string"}`

but multiple values as list of strings

`extra=string1&extra=string2` -> `{"extra": ["string1", "string2"]}`

We can consider always parse extra parameters as lists

```py
    for key in received_body.keys():
        if key not in values:
            values[key] = received_body.getlist(key)
```

This way for single value we would also have list with one element:

`extra=string` -> `{"extra": ["string"]}`

It's more consistent, but probably not what people expect for single value.

I think we should go with current implementation (single value -> single string, multiple values -> list of strings)

## Other parameters

Query\Header\Cookie parameters likely have the same issue, I will look at it a bit later and open separate PR to fix this


ivikasavnish (approved):

svlandeg (approved):
Nice catch, Yurii! Definitely tricky that `items()` of `ImmutableMultiDict` behaves this way 😬

tiangolo (approved):

And thanks @svlandeg for the review. 🍰 

This will be available in FastAPI 0.123.2, released in the next few hours. 🎉

PR #14318, opened 2025-11-09
✨ Add warning when route conflict detected in api routes

y3l1n4ung:
https://github.com/fastapi/fastapi/discussions/14314

### DIFF
--- a/tests/test_route_conflicts.py
+++ b/tests/test_route_conflicts.py
@@ -1,8 +1,9 @@
 import warnings
 
 from fastapi import FastAPI
-from fastapi.routing import APIRouter
+from fastapi.routing import APIRouter, _detect_route_conflicts
 from fastapi.testclient import TestClient
+from starlette.routing import Mount
 
 
 def test_route_conflict_warning_dynamic_before_static():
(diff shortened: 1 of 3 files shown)


tests/test_route_conflicts.py
```diff
@@ -0,0 +1,571 @@
...
+        @app.get("/items/sync")
+        def sync_items():
+            return {"action": "sync"}
+
+        @app.get("/items/{item_id}")
+        def get_item(item_id: str):
+            return {"item_id": item_id}
+
+        # Should warn about potential conflict
+        assert len(w) == 1
```
YuriiMotov, line 47: I think this is absolutely normal situation and FastAPI shouldn't warn in this case

YuriiMotov (commented):
@b14cknc0d3, thanks for your interest and efforts!

This implementation produces warnings for absolutely normal configurations, that actually work.
It's not a mistake when we declare more specific path before less specific.
If we have lot's of False warnings, users will just ignore them.

At the same time, case with sub-app that is shadowed by another endpoint with `{path:path}` is not handled (no warning), but the sub-app's endpoint is not reachable..

```py
import uuid
from fastapi import FastAPI


app = FastAPI()

@app.get("/case1/{item_id}/{action}")
def act(item_id: int, action: str):
    return {
        "endpoint": "act",
        "item_id": item_id,
        "action": action,
    }


# Catch-all endpoint for case1
@app.get("/case1/{path:path}")
def default(path: str):
    return {
        "endpoint": "default",
        "path": path,
    }

# The above endpoints work fine, but we have warning..
#  UserWarning: Route /case1/{path:path} (GET) may shadow existing route /case1/{item_id}/{action}
#  @app.get("/case1/{path:path}")


@app.get("/case2/{item_int_id:int}")
def int_ids(item_int_id: int):
    return {"endpoint": "int_ids"}

@app.get("/case2/{item_uuid_id:uuid}")
def uuid_ids(item_uuid_id: uuid.UUID):
    return {"endpoint": "uuid_ids"}

@app.get("/case2/{item_str_id:str}")
def str_ids(item_str_id: str):
    return {"endpoint": "str_ids"}


# The above endpoints work fine, but we have warnings..
# UserWarning: Route /case2/{item_str_id:str} (GET) may shadow existing route /case2/{item_int_id:int}
#   @app.get("/case2/{item_str_id:str}")
# UserWarning: Route /case2/{item_str_id:str} (GET) may shadow existing route /case2/{item_uuid_id:uuid}
#   @app.get("/case2/{item_str_id:str}")

sub_app = FastAPI()

@sub_app.get("/something")
def get_something():
    return {
        "endpoint": "get_something",
    }

app.mount(path="/case1", app=sub_app)

# No warnings, but get_something is not reachable (shadowed by `/case1/{path:path}`)

```

y3l1n4ung:
@YuriiMotov Thanks! I’ll update the implementation.

PR #14319, opened 2025-11-09
📝 Update user tutorial article categories

wu-clan:
This PR is for improving the user tutorial browsing experience.

`Path`, `Query`, `Body`, `Cookie`, `Header`, `Form`, `File` are self-explanatory in my opinion, they are universally recognized parameter names.

Before:

image

After:

image

### DIFF
--- a/scripts/docs.py
+++ b/scripts/docs.py
@@ -297,7 +297,7 @@ def live(
     if dirty:
         args.append("--dirty")
     subprocess.run(
-        args, env={**os.environ, "LINENUMS": "true"}, cwd=lang_path, check=True
+        args, env={**os.environ, "LINENUMS": "true", "PYTHONUTF8": "1"}, cwd=lang_path, check=True
     )
 
 
--- a/scripts/docs.py
+++ b/scripts/docs.py
@@ -297,7 +297,10 @@ def live(
     if dirty:
         args.append("--dirty")
     subprocess.run(
-        args, env={**os.environ, "LINENUMS": "true", "PYTHONUTF8": "1"}, cwd=lang_path, check=True
+        args,
+        env={**os.environ, "LINENUMS": "true", "PYTHONUTF8": "1"},
+        cwd=lang_path,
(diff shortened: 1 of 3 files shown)


scripts/docs.py
```diff
@@ -297,7 +297,10 @@ def live(
     if dirty:
         args.append("--dirty")
     subprocess.run(
-        args, env={**os.environ, "LINENUMS": "true"}, cwd=lang_path, check=True
+        args,
+        env={**os.environ, "LINENUMS": "true", "PYTHONUTF8": "1"},
```
wu-clan, line 301: Add `"PYTHONUTF8": "1"` is because by default, when executing `python ./scripts/docs.py live` on Windows, a `gbk` error occurs:
```
Error reading page 'advanced/path-operation-advanced-configuration.md': 'gbk' codec can't decode byte 0xa8 in position 288: illegal multibyte sequence
```

YuriiMotov (commented):
@wu-clan, thank you!

Personally, I agree that grouping pages makes the menu easier to navigate.

But, it changes the order of some sections and now we have a bit incorrect references:
 
`Path` goes before `Query`, but in **"Path Parameters and Numeric Validations"** we reference `Query` as if we've already learned it:

> In the same way that you can declare more validations and metadata for query parameters with `Query`, you can declare the same type of validations and metadata for path parameters with `Path`.

---

Let's ask Sebastian to validate the idea of such grouping. If he accepts the idea, I think we will need to check all cross-references and correct them.

Meanwhile, it would be nice if you went through pages in new order and checked how much efforts it will require. Is it worth it?

wu-clan:
I completely agree with your views on the matter of sequence.

I haven't tried to update the content because it would affect all languages, let's see what Sebastian has in mind first.

tiangolo:
The intention is to teach concepts gradually, building on top of each other, that's the reason for the current order. I prefer to leave it as it is. Thanks! :coffee:

PR #14332, opened 2025-11-10
📝 Update `tutorial/sql-databases.md` to use `lifespan` instead of deprecated `on_event`

KrzysztofCzapla:
This PR removes the deprecated `on_event()` method from SQL databases docs and instead adds `lifespan()` function inside @asynccontextmanager decorator [as specified in the documentation](https://fastapi.tiangolo.com/advanced/events/#lifespan).

The highlighted lines were updated accordingly in the `sql-databases.md` file. If the same should be done for the docs in other languages, please let me know.

I also added a link to the lifespan events page on line 110 for further reading for the consumers of the docs.


docs_src/sql_databases/tutorial001.py
```diff
@@ -27,12 +28,13 @@ def get_session():
         yield session
 
 
-app = FastAPI()
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    create_db_and_tables()
```
YuriiMotov: We shouldn't teach people to call **sync** blocking code inside **async** functions

KrzysztofCzapla:
@YuriiMotov 
Hi, thank you for the comment!

I fixed the code and now the sync function is used inside `asyncio.to_thread()` inside the lifespan function.
Regarding version 3.8 - I added `asyncio.run_in_executor()` because `to_thread` is not yet supported there.

I also fixed the highlighting of the lines in the `tutorial002` files, that I forgot about.

If you think anything should be changed or is wrong, please let me know.

docs/en/docs/tutorial/sql-databases.md
```diff
@@ -287,8 +289,8 @@ We receive in the request a `HeroCreate` *data model*, and from it, we create a
 This new *table model* `Hero` will have the fields sent by the client, and will also have an `id` generated by the database.
 
 Then we return the same *table model* `Hero` as is from the function. But as we declare the `response_model` with the `HeroPublic` *data model*, **FastAPI** will use `HeroPublic` to validate and serialize the data.
-
-{* ../../docs_src/sql_databases/tutorial002_an_py310.py ln[56:62] hl[56:58] *}
+7
```
YuriiMotov:
```suggestion

```

YuriiMotov (approved):
LGTM!

@KrzysztofCzapla, thank you!
~~We can also add a small note explaining why we need `await asyncio.to_thread()` here.~~ (Done)

I'll pass this PR to Sebastian for final review.

KrzysztofCzapla:
@YuriiMotov Thank you!

I added the small note you mentioned and also committed your suggestion.

tiangolo:

I want to implement a big refactor and some features in how to handle these things, I'm not fully happy with the existing interface.

I'll do that first, and then update the docs to reflect the recommended way to do things. I don't want to update the docs yet and suggest doing something that later I will suggest doing in a different way.

So for now, I will close this one. Thank you! 🍰

PR #14333, opened 2025-11-10
Simplifying control flow and tests for jsonable_encoder

sauravvenkat:
Changes:

1. Reduced unnecessary nesting in jsonable_encoder making logic appear simpler
2. Removed any pydanticv1 references in tests given we're on pydanticV2.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -210,13 +210,11 @@ def jsonable_encoder(
     [FastAPI docs for JSON Compatible Encoder](https://fastapi.tiangolo.com/tutorial/encoder/).
     """
     custom_encoder = custom_encoder or {}
-    if custom_encoder:
-        if type(obj) in custom_encoder:
-            return custom_encoder[type(obj)](obj)
-        else:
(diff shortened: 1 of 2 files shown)


sauravvenkat:
@tiangolo Can you please add a label to this. I don't have permissions to add a label.

YuriiMotov (commented):
@sauravvenkat, thank you for your interest!

I think that `if custom_encoder:` is here intentionally. Considering that in most cases `custom_encoder` will be `None` and `jsonable_encoder` is called recursively, it's valid performance optimization. And, logic is probably even clearer with `if custom_encoder:` (IMO).

> Removed any pydanticv1 references in tests given we're on pydanticV2.

FastAPI still supports Pydantic V1

---

Personally, I think we don't need to change this.

eltoder:
@YuriiMotov since `custom_encoder` is only used under `if custom_encoder:`, we can delete the
```python
```
line at the top, which is a minor optimization.

tiangolo:
I'll pass on this for now, thanks!

PR #14335, opened 2025-11-11
Added TypeVar to make Depends aware of dependency methods return type

ewen-lorimer:
This PR adds extended type hinting to the `Depends` param function. `Depends` now uses the return/generator value from the passed dependency method as the overall return type, removing the necessity for explicit type hints for route dependency arguments.

### DIFF
--- a/fastapi/param_functions.py
+++ b/fastapi/param_functions.py
@@ -1,4 +1,16 @@
-from typing import Any, Callable, Dict, List, Optional, Sequence, Union
+from typing import (
+    Any,
+    AsyncGenerator,
+    Awaitable,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    Sequence,
+    TypeVar,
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@ewen-lorimer, thanks for your interest!

Please, take a look at similar PRs that were closed:
* https://github.com/fastapi/fastapi/pull/11255
* https://github.com/fastapi/fastapi/pull/13922
* https://github.com/fastapi/fastapi/pull/10053

If you disagree with reasons those PRs were closed and can explain why we still need to apply these changes - please do explain it. Otherwise, let's close this PR.

And PR with alternative approach:
* https://github.com/fastapi/fastapi/pull/5856

tiangolo:
Thanks! Sadly this wouldn't be compatible with dependency classes like `Annotated[SomeClass, Depends()]`. So I'll pass on this one for now. :coffee:

PR #14337, opened 2025-11-11
Fix Pylance typing warnings in tutorial examples

clement-zentz:
### Summary
Fix Pylance type checking warnings in several tutorial examples by explicitly typing `results` as `dict[str, Any]` and using direct assignment instead of `.update()`.

### Reason
Recent versions of Python and Pylance infer `results` as `dict[str, list[dict[str, str]]]`,
causing false positives when adding a `str` value.

### Changes
- Updated multiple `docs_src/query_params_str_validations/*.py` files to fix Pylance typing warnings.

### DIFF
--- a/docs_src/query_params_str_validations/tutorial001.py
+++ b/docs_src/query_params_str_validations/tutorial001.py
@@ -1,4 +1,4 @@
-from typing import Union, Any
+from typing import Any, Union
 
 from fastapi import FastAPI
 
--- a/docs_src/query_params_str_validations/tutorial001.py
+++ b/docs_src/query_params_str_validations/tutorial001.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Union, Any
 
 from fastapi import FastAPI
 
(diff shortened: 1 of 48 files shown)


clement-zentz:
All linting issues have been fixed and all tests pass.  
This PR is ready for review and merge. ✅

YuriiMotov (commented):
@clement-zentz, thanks for your interest!

It's probably a good idea to fix typing mistakes in all docs code examples, but I think we should do it in a complex way for all code examples, and configure checking all code examples with mypy to prevent future mistakes.

Let's wait for Sebastian to take a decision about whether we should go this way. I guess this may require quite a lot of changes

clement-zentz:
> @clement-zentz, thanks for your interest!
> 
> It's probably a good idea to fix typing mistakes in all docs code examples, but I think we should do it in a complex way for all code examples, and configure checking all code examples with mypy to prevent future mistakes.
> 
> Let's wait for Sebastian to take a decision about whether we should go this way. I guess this may require quite a lot of changes

@YuriiMotov Thanks for the feedback!
Would it make sense if I set up a mypy check for all docs_src/* examples in a new PR, so we can systematically detect type issues?
I could start by adding a mypy.ini configuration and a CI step for that.

If the team decides to keep the current approach for now, I’m happy to focus this PR on these specific fixes.

YuriiMotov:
As I already said, let's wait for Sebastian to take the decision.
We have a lot of tasks in our backlog and we have to set priorities

clement-zentz:
Hi! Just a small ping to know if there are any updates on this PR.
Happy to adjust anything if needed.

tiangolo:
Thanks for the interest! I prefer to keep the examples simpler, without the explicit extra types, to make the point clearer. So I'll pass on this, but thanks! :coffee:

PR #14342, opened 2025-11-12
Fix usage of deprecated Pydantic methods in tutorial

coderliu:
Hi, I'm new to FastAPI.

I'm learning FastAPI through the fantastic documentation.

While trying the example at https://fastapi.tiangolo.com/tutorial/body/#use-the-model on my local environment, I noticed the following **deprecation warning** in VS Code:
image

Upon checking the Pydantic source code, I confirmed that the `dict` method of `BaseModel` is indeed deprecated in Pydantic V2. It appears the documentation needs to be updated.

I then referred to the Pydantic V2 migration guide: https://docs.pydantic.dev/latest/migration/#changes-to-pydanticbasemodel

I have searched all deprecated methods listed there within the `docs_src` path and replaced them with their corresponding methods for Pydantic V2 (e.g., replacing `.dict()` with `.model_dump()`).

### DIFF
--- a/docs_src/body/tutorial002.py
+++ b/docs_src/body/tutorial002.py
@@ -16,7 +16,7 @@ app = FastAPI()
 
 @app.post("/items/")
 async def create_item(item: Item):
-    item_dict = item.dict()
+    item_dict = item.model_dump()
     if item.tax is not None:
         price_with_tax = item.price + item.tax
         item_dict.update({"price_with_tax": price_with_tax})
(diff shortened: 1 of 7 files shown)


YuriiMotov:
Please, read the note below the code example:

> In Pydantic v1 the method was called `.dict()`, it was deprecated (but still supported) in Pydantic v2, and renamed to `.model_dump()`.
> 

coderliu:
@YuriiMotov 
You are right, I totally missed the note😅

Thank you for the clarify.

PR #14343, opened 2025-11-12
🐛 Fix request body mediatype for form model with `File` fields

YuriiMotov:
Currently if form parameter model contains `File` field, the content mediatype for this path operation in openapi will be wrong:

```py
from fastapi import FastAPI, Form, File

app = FastAPI()

class MyModel(BaseModel):
    file_model: bytes = File()

@app.post("/file-model")
async def create_item(item: MyModel = Form()):
    return {"file_size": len(item.file_model)}
```

Request body content mediatype in openapi will be `application/x-www-form-urlencoded` instead of `multipart/form-data`:
```json
        "requestBody": {
          "content": {
            "application/x-www-form-urlencoded": {
              "schema": {
                "$ref": "#/components/schemas/MyModel"
              }
            }
          },
          "required": true
        },
```

This PR enforces the content mediatype to be `multipart/form-data` for form parameter models that contain at least one `File` field.
I added `if request_media_type == "application/x-www-form-urlencoded"` condition to make it possible to override mediatype by passing a parameter to `Form`:

```
@app.post("/file-model")
async def create_item(item: MyModel = Form(media_type="whatever")):
    ...
```

### DIFF
--- a/docs/en/docs/tutorial/request-form-models.md
+++ b/docs/en/docs/tutorial/request-form-models.md
@@ -36,6 +36,14 @@ You can verify it in the docs UI at `/docs`:
 <img src="/img/tutorial/request-form-models/image01.png">
 </div>
 
+## Including `File` Fields { #including-file-fields }
+
+You can also include **file fields** inside a form model by using `UploadFile` or `bytes` as field types and annotating them with `File`.
+
+If there is at least one `File` field in the form model, **FastAPI** will automatically set the request content media type to `multipart/form-data` for this path operation.
+
+{* ../../docs_src/request_form_models/tutorial001b_an_py39.py hl[3,12,16] *}
+
 ## Forbid Extra Form Fields { #forbid-extra-form-fields }
 
 In some special use cases (probably not very common), you might want to **restrict** the form fields to only those declared in the Pydantic model. And **forbid** any **extra** fields.
@@ -75,4 +83,4 @@ They will receive an error response telling them that the field `extra` is not a
 
 ## Summary { #summary }
 
-You can use Pydantic models to declare form fields in FastAPI. 😎
+You can use Pydantic models to declare form fields - including file uploads - in FastAPI. 😎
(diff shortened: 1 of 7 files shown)


docs/en/docs/tutorial/request-form-models.md
```diff
@@ -36,6 +36,14 @@ You can verify it in the docs UI at `/docs`:
 <img src="/img/tutorial/request-form-models/image01.png">
 </div>
 
+## Including `File` Fields { #including-file-fields }
+
+You can also include **file fields** inside a form model by using `UploadFile` or `bytes` as field types and annotating them with `File`.
+
+If there is at least one `File` field in the form model, **FastAPI** will automatically set the request content media type to `multipart/form-data` for this path operation.
```
svlandeg, line 43: Should we add here a note to indicate from which version onwards, this is the case? Something like 
```suggestion
If there is at least one `File` field in the form model, **FastAPI** (from v.0.121.3 onwards) will automatically set the request content media type to `multipart/form-data` for this path operation.
```

tiangolo:
I'm unsure about this. :thinking: 

I wouldn't expect people to create Pydantic models where the fields are declared with something other than Pydantic's regular `Field()`. I wouldn't expect them to use `File()` in a Pydantic model field.

YuriiMotov:
> I'm unsure about this. 🤔
> 
> I wouldn't expect people to create Pydantic models where the fields are declared with something other than Pydantic's regular `Field()`. I wouldn't expect them to use `File()` in a Pydantic model field.


In the code example I already use `File` for model field:

```py
class MyModel(BaseModel):
    file_model: bytes = File()
```

YuriiMotov:
Closing this as explained [here](https://github.com/fastapi/fastapi/pull/14343#issuecomment-3600534954) and [here](https://github.com/fastapi/fastapi/pull/14358#discussion_r2607250726)

Issue #14344, opened 2025-11-13
AttributeError: 'dict' object has no attribute 'split'


Kludex:
Minimum Reproducible Example:

```py
# /// script
# dependencies = [
#   "pydantic",
#   "fastapi[standard]",
#   "starlette",
# ]
# ///

import pydantic
import starlette
import fastapi
from fastapi import FastAPI
import sys
from pydantic import BaseModel, Field, ConfigDict

# print the dependencies to stdout, one per line
print(f'pydantic version = {pydantic.__version__}')
print(f'starlette version = {starlette.__version__}')
print(f'fastapi version = {fastapi.__version__}')
# print python version
print(f'Python version = {sys.version}')


app = FastAPI(title='Logfire')

class ModelWithRef(BaseModel):
    ref: str = Field(validation_alias='$ref', serialization_alias='$ref')
    model_config = ConfigDict(validate_by_alias=True, serialize_by_alias=True)



@app.get('/')
async def read_root() -> ModelWithRef:
    return ModelWithRef(ref='some-ref')

print(app.openapi())
```

Run with `uv run main.py`.

Error:

```bash
❯ uv run main.py   
Installed 42 packages in 61ms
pydantic version = 2.12.4
starlette version = 0.49.3
fastapi version = 0.121.1
Python version = 3.10.15 (main, Oct 16 2024, 08:33:15) [Clang 18.1.8 ]
Traceback (most recent call last):
  File "/Users/marcelotryle/dev/main.py", line 36, in <module>
    print(app.openapi())
  File "/Users/marcelotryle/.cache/uv/environments-v2/main-84bd05b9e640fe68/lib/python3.10/site-packages/fastapi/applications.py", line 1060, in openapi
    self.openapi_schema = get_openapi(
  File "/Users/marcelotryle/.cache/uv/environments-v2/main-84bd05b9e640fe68/lib/python3.10/site-packages/fastapi/openapi/utils.py", line 504, in get_openapi
    field_mapping, definitions = get_definitions(
  File "/Users/marcelotryle/.cache/uv/environments-v2/main-84bd05b9e640fe68/lib/python3.10/site-packages/fastapi/_compat/main.py", line 292, in get_definitions
    v2_field_maps, v2_definitions = v2.get_definitions(
  File "/Users/marcelotryle/.cache/uv/environments-v2/main-84bd05b9e640fe68/lib/python3.10/site-packages/fastapi/_compat/v2.py", line 249, in get_definitions
[...]

PR #14346, opened 2025-11-13
Update v2.py

amitmoralwar:
Fixed AttributeError: 'dict' object has no attribute 'split' #14344 this issue

### DIFF
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -262,18 +262,18 @@ def _replace_refs(
     new_schema = deepcopy(schema)
     for key, value in new_schema.items():
         if key == "$ref":
-            # ref_name = schema["$ref"].split("/")[-1]
-            ref_name = schema["$ref"]
+            print(f"schema:{schema}")
+            #ref_name = schema["$ref"].split("/")[-1]
+            ref_name = schema["$ref"]            
             if isinstance(ref_value, str):
                 # Normal case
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Let's keep it as Draft until it's ready for review

amitmoralwar:
deletting because of duplication

njsbhu:
I would like to add/modify as below
1) _replace_refs — undefined ref_value and wrong handling of $ref
```
if key == "$ref":
    print(f"schema:{schema}")
    # ref_name = schema["$ref"].split("/")[-1]
    ref_name = schema["$ref"]
    if isinstance(ref_value, str):
        # Normal case
        ref_name = ref_value.split("/")[-1]
```
2)  ModelField.validate — mutable default argument
```
def validate(self,
        value: Any,
        values: Dict[str, Any] = {},  # noqa: B006
        *,
        loc: Tuple[Union[int, str], ...] = (),
    ) -> 
```
3) Fixed ModelField.validate (no mutable defaults)
```
def validate(
    self,
    value: Any,
    values: Dict[str, Any] | None = None,
    *,
    loc: Tuple[Union[int, str], ...] = (),
    values = {} if values is None else values
    try:
        validated = self._type_adapter.validate_python(value, from_attributes=True)
        return validated, None
    except ValidationError as exc:
        return None, may_v1._regenerate_error_with_loc(
            errors=exc.errors(include_url=False), loc_prefix=loc
        )
```

PR #14349, opened 2025-11-13
🐛 Fix handling of JSON Schema attributes named "$ref"

tiangolo:
Fix handling of JSON Schema attributes named "$ref"

Related to: https://github.com/fastapi/fastapi/issues/14344

### DIFF
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -262,12 +262,12 @@ def _replace_refs(
     new_schema = deepcopy(schema)
     for key, value in new_schema.items():
         if key == "$ref":
-            ref_name = schema["$ref"].split("/")[-1]
-            if ref_name in old_name_to_new_name_map:
-                new_name = old_name_to_new_name_map[ref_name]
-                new_schema["$ref"] = REF_TEMPLATE.format(model=new_name)
-            else:
-                new_schema["$ref"] = schema["$ref"]
+            value = schema["$ref"]
+            if isinstance(value, str):
+                ref_name = schema["$ref"].split("/")[-1]
+                if ref_name in old_name_to_new_name_map:
+                    new_name = old_name_to_new_name_map[ref_name]
+                    new_schema["$ref"] = REF_TEMPLATE.format(model=new_name)
             continue
         if isinstance(value, dict):
             new_schema[key] = _replace_refs(
--- a/tests/test_schema_ref_pydantic_v2.py
+++ b/tests/test_schema_ref_pydantic_v2.py
@@ -1,39 +1,38 @@
-import sys
 from typing import Any
 
+import pytest
 from fastapi import FastAPI
 from fastapi.testclient import TestClient
 from inline_snapshot import snapshot
-
-from tests.utils import skip_module_if_py_gte_314
-
-if sys.version_info >= (3, 14):
-    skip_module_if_py_gte_314()
-
 from pydantic import BaseModel, ConfigDict, Field
 
-app = FastAPI()
+from tests.utils import needs_pydanticv2
 
 
-class ModelWithRef(BaseModel):
-    ref: str = Field(validation_alias="$ref", serialization_alias="$ref")
-    model_config = ConfigDict(validate_by_alias=True, serialize_by_alias=True)
+@pytest.fixture(name="client")
+def get_client():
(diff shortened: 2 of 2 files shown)


farzbood:
Hi,
I'm trying to use the [Letta-ai letta (standalone package)](https://github.com/letta-ai/letta) that depends on `FastAPI` running the server in windows command-prompt cause raising **fastapi-related exceptions** (screenshots attached).
I'm not sure if this is related to **this fix** or not, just tried to run `letta-server` again with the **latest versions of all packages** but still the same, would you kindly consider having a look at this and let us know if discover anything helpful?
The [issue I've opened there](https://github.com/letta-ai/letta/issues/3053#issue-3560755389) has no reply yet!
Thanks in advance 🙏 

letta-fastapi-exception-01
letta-fastapi-exception-02

njsbhu:
1) I would like to update function - _replace_refs as below
```
if key == "$ref":
    print(f"schema:{schema}")
    # ref_name = schema["$ref"].split("/")[-1]
    ref_name = schema["$ref"]
    if isinstance(ref_value, str):
        # Normal case
        ref_name = ref_value.split("/")[-1]
    ...
```
2) ModelField.validate — mutable default argument
```
def validate(self,
        value: Any,
        values: Dict[str, Any] = {},  # noqa: B006
        *,
        loc: Tuple[Union[int, str], ...] = (),
    ) -> ...
```
3) serialize_sequence_value — unsafe issubclass usage
```
origin_type = get_origin(field.field_info.annotation) or field.field_info.annotation
assert issubclass(origin_type, shared.sequence_types)  # type: ignore[arg-type]
```

4) get_flat_models_from_model — does not use return of helper (style)
```
fields = get_model_fields(model)
get_flat_models_from_fields(fields, known_models=known_models)
return known_models
```
5) Many redundant / non-idiomatic imports
`from pydantic import PydanticSchemaGenerationError as PydanticSchemaGenerationError`

6) Fixed _replace_refs (robust, recursive, no prints)
```
def _replace_refs(
    *,
    schema: Dict[str, Any],
    old_name_to_new_name_map: Dict[str, str],
) -> Dict[str, Any]:
    """
    Replace $ref values in a json-schema-like dict.

    Handles cases where "$ref" value may be a string (normal) or a dict (custom),
    and recurses into nested dicts/lists. Returns a new dict (does not mutate input).
    """
    new_schema: Dict[str, Any] = {}

    for key, value in schema.items():
        if key == "$ref":
            ref_value = value
            # Normal case: "$ref": "#/definitions/ModelName" or "ModelName"
            ref_name = None
            if isinstance(ref_value, str):
                # Extract last path component
                ref_name = ref_value.split("/")[-1]
            elif isinstance(ref_value, dict):
                # Some tools may use a dict here; try to extract a title or name
                ref_name = ref_value.get("title") or ref_value.get("$ref") or None
                if isinstance(ref_name, str):
                    ref_name = ref_name.split("/")[-1]
            if ref_name and ref_name in old_name_to_new_name_map:
                new_name = old_name_to_new_name_map[ref_name]
                new_schema["$ref"] = REF_TEMPLATE.format(model=new_name)
            else:
                # leave as-is if we cannot remap
                new_schema["$ref"] = ref_value
            continue

        # Recurse into dicts and lists
        if isinstance(value, dict):
            new_schema[key] = _replace_refs(
                schema=value, old_name_to_new_name_map=old_name_to_new_name_map
            )
        elif isinstance(value, list):
            new_list = []
            for item in value:
                if isinstance(item, dict):
                    new_list.append(
                        _replace_refs(schema=item, old_name_to_new_name_map=old_name_to_new_name_map)
                    )
                else:
                    new_list.append(item)
            new_schema[key] = new_list
        else:
            new_schema[key] = value

    return new_schema
```

7) Fixed ModelField.validate (no mutable defaults)
```
def validate(
    self,
    value: Any,
    values: Dict[str, Any] | None = None,
    *,
    loc: Tuple[Union[int, str], ...] = (),
    values = {} if values is None else values
    try:
        validated = self._type_adapter.validate_python(value, from_attributes=True)
        return validated, None
    except ValidationError as exc:
        return None, may_v1._regenerate_error_with_loc(
            errors=exc.errors(include_url=False), loc_prefix=loc
        )
```

PR #14356, opened 2025-11-16
🐛 Fix parsing extra non-body parameter list

YuriiMotov:
Currently if we declare Query or Header parameters as model with `extra="allow"` and then pass extra parameter with multiple values, only last value of such parameter will be passed to model validator:

```py

app = FastAPI()


class Model(BaseModel):
    param: str

    model_config = {"extra": "allow"}


@app.get("/")
async def model_with_extra(data: Model = Query()):
    return data


def test_pass_extra_list():
    client = TestClient(app)
    resp = client.get(
        "/",
        params={
            "param": "123",
            "param2": ["456", "789"],  # Pass a list of values as extra parameter
        },
    )
    assert resp.json() == {
        "param": "123",
        "param2": ["456", "789"],
    }

# AssertionError: assert
# {'param': '123', 'param2': '789'} == {'param': '123', 'param2': ['456', '789']}
```

The reason is the same as with Form parameters. See #14303
There is also a note about implementation described in #14303. It's also relevant to this PR.

**As for Cookies:**
As far as I understand the spec, you can't pass multiple Cookie parameters with the same name (only last value will be used). The only way to pass a list of values is to pass them comma-separated, but it's not currently supported by Starlette


svlandeg (approved):
Looks good! The implementation is slightly more convoluted here than in #14303 because the type of `received_params` is `Union[Mapping[str, Any], QueryParams, Headers]`. The code in this PR addresses all the possibilities, as far as I can see.

tiangolo (approved):
Excellent job, thank you @YuriiMotov! 🙌 

And thanks @svlandeg for the review! 🍰 

This will be available in FastAPI 0.123.2, in the next few hours. 🎉

PR #14360, opened 2025-11-17
🐛 Fix Query\Header\Cookie parameter model alias

YuriiMotov:
When we declare Query\Header\Cookie parameter as model and specify `alias` for the field, there are currently 2 issues:
1. Value is extracted by alias, but it's put into `values` with the key `field.name` instead of `key.alias`. Then, when we attempt to validate the model passing `values`, it fails to find the value for that field (because it searches by alias) and we have validation error.
2. If we pass the value by name, we will correctly get validation error, but the `input` field will not contain the extra parameter (passed by name). That's because the `field.name` were mistakenly added to `processed_keys`. We should only add the parameter alias, not both.

This PR fixes these two issues


tiangolo (approved):
Excellent, thank you! 🙌 

This will be available in FastAPI 0.123.3, released in the next few hours. 🎉

PR #14361, opened 2025-11-17
🐛 Avoid accessing non-existing "$ref" key for Pydantic v2 compat remapping

svlandeg:
cf Discussion #14265


YuriiMotov (approved):
LGTM!

Can confirm that generated schema is equal to schema generated before 0.119.0.


### Some thoughts

According to spec, empty list of values is allowed (the word SHOULD is used):

> ### [6.1.2. enum](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.1.2)
> The value of this keyword MUST be an array. This array SHOULD have at least one element.

I think people would expect the schema for generic string enum to be `{"type": "string"}` (match any string enum).
`{"enum": []}` is useless as it will never match..

But still, fixing this issue opens a way to modify schema using `SkipJsonSchema` and `WithJsonSchema`

jerry-reevo:
For my case reported in https://github.com/fastapi/fastapi/discussions/14265 I am able to successfully generate an OpenAPI specification file without erroring out when using this branch of FastAPI.

Thanks!

svlandeg:
Thanks for reporting back, @jerry-reevo !

tsmathis:
Just adding in, my team was also facing the same issue discussed in this PR which has kept us stuck at `fastapi==0.118.3`. This branch has seemed to resolved this issue for me (at least for local testing).

zachary-walravens:
I can also confirm this PR fixes the issue for me. Thanks @svlandeg!

tiangolo (approved):
Nice, thank you for the fix! 🚀 

This will be available in FastAPI 0.123.1 in the next few hours. 🎉

PR #14358, opened 2025-11-17
✅ Add set of tests for request parameters and alias

YuriiMotov:
Tests for different types of parameters:
* `Query`
* `Path`
* `Header`
* `Cookie`
* `Form`
* `File`
* `Body`

Some of these tests currently fail on `master`.

Here is a list of PRs to fix issues:
* https://github.com/fastapi/fastapi/pull/14297
* https://github.com/fastapi/fastapi/pull/14360

* https://github.com/fastapi/fastapi/pull/14303
* https://github.com/fastapi/fastapi/pull/14343 (closed as we decided not to support this for now)
* https://github.com/fastapi/fastapi/pull/14356

* https://github.com/fastapi/fastapi/pull/14371

Fixing this will resolve:
* https://github.com/fastapi/fastapi/issues/10286
* https://github.com/fastapi/fastapi/issues/12402

---

There is still not fixed issue with Swagger UI (it's unable to send list of form parameters properly) - it sends them comma-separated, but Starlette can't parse them. Discussed [here](https://github.com/fastapi/fastapi/discussions/8741)

<details>

```py
from fastapi import FastAPI, Form

app = FastAPI()

@app.post("/")
def form_list(data: list[str] = Form()):
    return data
```

Swagger UI will send it as:
```
curl -X 'POST' \
  'http://localhost:8000/' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d 'data=string1,string2'

["string1,string2"]
```

But Starlette can't parse them (`style="form", explode=False` seems to be not supported by `python-multipart`).

To workaround this we can specify the encoding explicitly:
```py
from fastapi import FastAPI, Form

app = FastAPI()

@app.post(
    "/",
    openapi_extra={
        "requestBody": {
          "content": {
            "application/x-www-form-urlencoded": {
                "encoding": {"data": {"style": "form"}}
            }
          }
        }
    }
)
async def form_list(data: list[str] = Form()):
    return data
```
But that's a lot of code..

</details>

There are 3 ways to solve this:
* Fix this on Swagger's side
* Add the support for `style="form", explode=false`  to `python-multipart`
* Modify FastAPI to explicitly specify `style=form`  in openapi for form parameters (it's actually a default value, but Swagger fails if it's omitted)

I don't think this will be fixed by Swagger. Last approach (workaround on FastAPI's side) is the simplest

PR #14371, opened 2025-11-19
🐛 Fix parameter aliases

YuriiMotov:
This PR fixes the rest of failing tests in #14358.

---

**Note:** for `validation_aliases` only string value is currently supported. We can later think of supporting `AliasPath` and `AliasChoices`

**Will also fix issues:**

<details>

* https://github.com/fastapi/fastapi/discussions/8535
* https://github.com/fastapi/fastapi/discussions/10014

</details>


fastapi/params.py
```diff
@@ -115,6 +115,10 @@ def __init__(
         else:
             kwargs["deprecated"] = deprecated
         if PYDANTIC_V2:
+            if serialization_alias in (_Unset, None) and isinstance(alias, str):
```
YuriiMotov, line 118: This is how it's done in Pydantic's Field: https://github.com/pydantic/pydantic/blob/407934af10962473d8db0b8c693393cad3975eda/pydantic/fields.py#L1346-L1350

fastapi/dependencies/utils.py
```diff
@@ -1021,3 +1019,8 @@ def get_body_field(
         field_info=BodyFieldInfo(**BodyFieldInfo_kwargs),
     )
     return final_field
+
+
+def get_validation_alias(field: ModelField) -> str:
```
YuriiMotov, line 1065: Should we add cache?

tiangolo (approved):
Excellent job with this @YuriiMotov! Thank you! 🙌 

This will be available in FastAPI 0.124.4, released in the next few hours. 🎉

fastapi/dependencies/utils.py
tiangolo: It's fine like this for now, we'll soon remove support for Pydantic v1 and will be able to refactor this.

fastapi/params.py
tiangolo: Nice, thanks for the clarification!

rensoostenbachBL:
A little bit late to the party, but I am curious if https://github.com/fastapi/fastapi/discussions/9099 is also fixed by this PR. Either way, thanks for this work @YuriiMotov!

PR #14372, opened 2025-11-19
♻️ Make the result of `Depends()` and `Security()` hashable, as a workaround for other tools interacting with these internal parts

tiangolo:
♻️ Make the result of `Depends()` and `Security()` hashable, as a workaround for other tools interacting with these internal parts

---

This comes from https://github.com/fastapi/fastapi/pull/14320

As a note, FastAPI doesn't really have official support for extending its internals in ways that are not documented in the docs.

So this is not a long-term commitment to keep some specific behavior, but a quick workaround to make things easier for current tools that interact with FastAPI internals in some way.

---

At some point in the future, I'll refactor the internals of FastAPI to make them more explicitly private, and then also figure out the right ways to expose any points of connection with external components that currently need to interact with the internals in some way, and formalize that better.

---

Before that, I need to handle a few issues, bugs, features, refactors, deprecations, etc. And then I'll be able to come back to this idea of figuring out how and what to expose. :nerd_face:

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -1,3 +1,4 @@
+import dataclasses
 import inspect
 from contextlib import AsyncExitStack, contextmanager
 from copy import copy, deepcopy
@@ -428,7 +429,7 @@ def analyze_param(
     if depends is not None and depends.dependency is None:
         # Copy `depends` before mutating it
         depends = copy(depends)
-        depends.dependency = type_annotation
+        depends = dataclasses.replace(depends, dependency=type_annotation)
 
     # Handle non-param type annotations like Request
     if lenient_issubclass(
--- a/fastapi/params.py
+++ b/fastapi/params.py
@@ -762,13 +762,13 @@ class File(Form):  # type: ignore[misc]
         )
 
 
-@dataclass
+@dataclass(frozen=True)
 class Depends:
     dependency: Optional[Callable[..., Any]] = None
     use_cache: bool = True
     scope: Union[Literal["function", "request"], None] = None
 
 
-@dataclass
+@dataclass(frozen=True)
 class Security(Depends):
     scopes: Optional[Sequence[str]] = None
--- a/tests/test_depends_hashable.py
+++ b/tests/test_depends_hashable.py
@@ -1,0 +1,25 @@
+# This is more or less a workaround to make Depends and Security hashable
+# as other tools that use them depend on that
+# Ref: https://github.com/fastapi/fastapi/pull/14320
+
+from fastapi import Depends, Security
+
+
+def dep():
+    pass
+
+
+def test_depends_hashable():
+    dep()  # just for coverage
+    d1 = Depends(dep)
+    d2 = Depends(dep)
+    d3 = Depends(dep, scope="function")
+    d4 = Depends(dep, scope="function")
+
+    s1 = Security(dep)
+    s2 = Security(dep)
+
+    assert hash(d1) == hash(d2)
+    assert hash(s1) == hash(s2)
+    assert hash(d1) != hash(d3)
+    assert hash(d3) == hash(d4)


njsbhu:
I would like to update analyze_param function in - fastapi/dependencies/utils.py
1) In Extract Annotated info
```
def analyze_param(
    *,
    param_name: str,
    annotation: Any,
    value: Any,
    is_path_param: bool,
) -> ParamDetails:
    field_info = None
    depends = None
    type_annotation: Any = Any
    use_annotation: Any = Any
    if annotation is not inspect.Signature.empty:
        use_annotation = annotation
        type_annotation = annotation

    # Extract Annotated info
    if get_origin(use_annotation) is Annotated:
        annotated_args = get_args(use_annotation)  # <-- use the resolved value
        # first arg is the type
        type_annotation = annotated_args[0] if annotated_args else type_annotation
        fastapi_annotations = [
            arg
            for arg in annotated_args[1:]
            if isinstance(arg, (FieldInfo, may_v1.FieldInfo, params.Depends))
        ]

        fastapi_specific_annotations = [
            arg
            for arg in fastapi_annotations
            if isinstance(
                arg,
                (
                    params.Param,
                    temp_pydantic_v1_params.Param,
                    params.Body,
                    temp_pydantic_v1_params.Body,
                    params.Depends,
                ),
            )
        ]

        fastapi_annotation = fastapi_specific_annotations[-1] if fastapi_specific_annotations else None

        # Annotated FieldInfo handling
        if isinstance(fastapi_annotation, (FieldInfo, may_v1.FieldInfo)):
            # Copy `field_info` because we mutate `field_info.default` below.
            field_info = copy_field_info(field_info=fastapi_annotation, annotation=use_annotation)
            assert field_info.default in {
                Undefined,
                may_v1.Undefined,
            } or field_info.default in {RequiredParam, may_v1.RequiredParam}, (
                f"`{field_info.__class__.__name__}` default value cannot be set in"
                f" `Annotated` for {param_name!r}. Set the default value with `=` instead."
            )
            if value is not inspect.Signature.empty:
                assert not is_path_param, "Path parameters cannot have default values"
                field_info.default = value
            else:
                field_info.default = RequiredParam

        # Annotated Depends handling
        elif isinstance(fastapi_annotation, params.Depends):
            depends = fastapi_annotation
```

2) In _get_multidict_value function - safe _get_multidict_value for Headers
```
def _get_multidict_value(
    field: ModelField, values: Mapping[str, Any], alias: Union[str, None] = None
) -> Any:
    alias = alias or field.alias
    if is_sequence_field(field) and hasattr(values, "getlist"):
        # ImmutableMultiDict supports getlist; Headers does not.
        value = values.getlist(alias)  # type: ignore[attr-defined]
    elif is_sequence_field(field) and isinstance(values, Headers):
        # Headers in Starlette exposes getall (or multiple values in raw), try getall/getlist fallback
        try:
            value = values.getall(alias)  # type: ignore[attr-defined]
        except Exception:
            # Fallback: single header value -> put in list if present
            v = values.get(alias, None)
            value = [v] if v is not None else []
    else:
        value = values.get(alias, None)
    ...
```
3) safer response header deletion in solve_dependencies
Replace:
```
if response is None:
    response = Response()
    del response.headers["content-length"]
    response.status_code = None  # type: ignore
```
with:
```
if response is None:
    response = Response()
    # remove content-length if present; be tolerant if not
    response.headers.pop("content-length", None)
    response.status_code = None  # type: ignore
```
4)  Avoid AttributeError in _extract_form_body when reading files. Assume value.read exists for objects in value list. Improve check:
```
if is_bytes_sequence_field(field) and isinstance(field_info, (params.File, temp_pydantic_v1_params.File)) and value_is_sequence(value):
    # Ensure items are UploadFile-like and have read() coroutine method
    results: List[Union[bytes, str]] = []

    async def process_fn(fn):
        result = await fn()
        results.append(result)

        for sub_value in value:
            if hasattr(sub_value, "read"):
                tg.start_soon(process_fn, sub_value.read)
            else:
                # fallback: append raw value
                results.append(sub_value)
```

PR #14375, opened 2025-11-20
🔧 Upgrade Material for MkDocs and remove insiders

tiangolo:
🔧 Upgrade Material for MkDocs and remove insiders


njsbhu:
1) Timeouts & concurrency: add timeout-minutes per job and concurrency at workflow level so repeated pushes cancel older runs. Saves runner minutes.
```
concurrency:
  group: docs-${{ github.ref }}
  cancel-in-progress: true
```
2) timeout under jobs
```
jobs:
  changes:
    runs-on: ubuntu-latest
    timeout-minutes: 20
```
3) timeout under langs
```
  langs:
    needs: changes
    runs-on: ubuntu-latest
    timeout-minutes: 15
```
4) updated build-docs - Fail fast vs allowed skips: your docs-all-green job uses allowed-skips: build-docs. Decide whether skips are acceptable — it can cause branch protection false positives if build was intentionally skipped.
```
  build-docs:
    needs: [changes, langs]
    # Run only if docs changed AND langs list is non-empty
    if: ${{ needs.changes.outputs.docs == 'true' && needs.langs.outputs.langs != '[]' }}
    runs-on: ubuntu-latest
    timeout-minutes: 60
```

PR #14382, opened 2025-11-22
Improve error handling for item retrieval

koladebashir:
Fix: Return 404 for non-existent items instead of null response

Previously, valid ID formats with non-existent items returned 200 OK  with null values, creating ambiguous API responses. Now properly returns  404 Not Found when items don't exist in the database.


YuriiMotov:
I think this is not needed for this code example. We should keep code examples simple and focus on the things we are explaining in the current part of docs.

@koladebashir, thanks for your interest! I'll close this

PR #14383, opened 2025-11-22
Update .gitignore commands in virtual-environments.md

koladebashir:
The documentation instructs users to create a .gitignore file inside the .venv folder. This is unnecessary and potentially confusing, because the virtual environment should be ignored from the project root .gitignore. Creating a .gitignore inside .venv is not standard Git/Python practice and can mislead users.

### DIFF
--- a/docs/en/docs/virtual-environments.md
+++ b/docs/en/docs/virtual-environments.md
@@ -261,7 +261,7 @@ Do this **once**, right after you create the virtual environment.
 <div class="termy">
 
 ```console
-$ echo "*" > .venv/.gitignore
+$ echo ".venv" > .gitignore
 ```
 
 </div>


YuriiMotov (commented):
@koladebashir, thanks for your interest!

Personally, I don't see any problem with current approach (creating `.gitignore` inside `.venv`).
I agree it's less common, but it works well and I don't see how it "can mislead users". 

Also, since you updated the command, the clarification below ("What that command means") is invalid\outdated

koladebashir:
Thanks @YuriiMotov. I’ve looked into it further and agree the local .gitignore approach works well here. Good catch on the explanation text being outdated after my change; that was an oversight on my part. Appreciate the feedback!

PR #14384, opened 2025-11-22
Fix: Make Basic Auth realm required per RFC 7235 

### LINKED ISSUE #13471: Basic auth `realm` is REQUIRED but handled as optional

kumarvishwajeettrivedi:
Fixes #13471

### DIFF
--- a/docs_src/security/tutorial006.py
+++ b/docs_src/security/tutorial006.py
@@ -3,7 +3,7 @@ from fastapi.security import HTTPBasic, HTTPBasicCredentials
 
 app = FastAPI()
 
-security = HTTPBasic()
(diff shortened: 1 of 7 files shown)


kumarvishwajeettrivedi:
"I noticed PR #13511 also addresses this. My approach here makes realm strictly required as per RFC 7235, whereas #13511 keeps it optional. I'm happy to go with whichever approach the maintainers prefer!"

YuriiMotov (commented):
@kumarvishwajeettrivedi, thanks for your interest and efforts!

svlandeg:
Closing, cf https://github.com/fastapi/fastapi/issues/13471#issuecomment-3596388412

PR #14386, opened 2025-11-22
Fix: Merge background tasks from dependency and custom response

### LINKED ISSUE #11215: Potential footgun when using custom `Response(background=<Task()>)` in conjunction with injected `BackgroundTasks` - the custom response overwrites the tasks

### First Check

### Commit to Help

### Example Code

```python
import uvicorn

app = FastAPI()

@app.get("/")

uvicorn.run(app)
```

### Description

### Operating System

macOS

### Operating System Details

_No response_

### FastAPI Version

0.97.0

### Pydantic Version

1.10.7

### Python Version

3.9

### Additional Context

_No response_</div>

kumarvishwajeettrivedi:
Fixes #11215

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -401,21 +401,25 @@ def get_request_handler(
 
                     # Create a new BackgroundTasks to hold both
                     combined_tasks = BackgroundTasks()
-                    
+
                     # Add existing response tasks
                     if isinstance(raw_response.background, BackgroundTasks):
                         combined_tasks.tasks.extend(raw_response.background.tasks)
                     elif isinstance(raw_response.background, BackgroundTask):
                         combined_tasks.tasks.append(raw_response.background)
-                    
+
                     # Add dependency tasks
                     # solved_result.background_tasks is always BackgroundTasks (from dependencies/utils.py)
                     if isinstance(solved_result.background_tasks, BackgroundTasks):
-                        combined_tasks.tasks.extend(solved_result.background_tasks.tasks)
-                    elif isinstance(solved_result.background_tasks, BackgroundTask):  # pragma: no cover
-                         # Should not happen for BackgroundTasks dependency but safe to handle
+                        combined_tasks.tasks.extend(
+                            solved_result.background_tasks.tasks
+                        )
+                    elif isinstance(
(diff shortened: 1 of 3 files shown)


kumarvishwajeettrivedi:
Hi! All tests have passed could someone please review when available?
Happy to make any improvements needed.

YuriiMotov (commented):
@kumarvishwajeettrivedi, thanks for your interest!

I think we should first decide what is the correct way to resolve this.
Merging BG tasks seems to obvious from the first sight, but I'm not sure it's correct way.

Current implementation doesn't look like a bug to me. It looks like this was made this way intentionally.

I think it would be better not to merge BG tasks, but emit a warning in such cases:

<details>

```py
from fastapi import FastAPI
from fastapi.background import BackgroundTasks

app = FastAPI()


@app.get("/bg-tasks")
async def bg_endpoint(bg_tasks: BackgroundTasks):
    bg_tasks.add_task(lambda: print("This will not be executed"))

    if True:
        return JSONResponse(  # This will emit a warning
            content="Custom response", background=lambda: print("This will be executed")
        )

    return "Default response"
```

</details>

**See https://github.com/fastapi/fastapi/pull/15112 that implements this approach**

And document that in such cases you should either clear the list of previously added tasks explicitly or append new task to the list of previously added tasks:

<details>

```py
from fastapi import FastAPI
from fastapi.background import BackgroundTasks

app = FastAPI()


@app.get("/bg-tasks-clear")
async def bg_clear_endpoint(bg_tasks: BackgroundTasks):
    bg_tasks.add_task(lambda: print("This will not be executed"))

    if True:
        bg_tasks.tasks = []  # Clear background tasks to avoid warning
        return JSONResponse(
            content="Custom response", background=lambda: print("This will be executed")
        )

    return "Default response"


@app.get("/bg-tasks-append")
async def bg_append_endpoint(bg_tasks: BackgroundTasks):
    bg_tasks.add_task(lambda: print("This WILL BE executed"))

    if True:
        bg_tasks.append(lambda: print("This will also be executed"))
        return JSONResponse(content="Custom response")

    return "Default response"

```

</details>

Let's wait for Sebastian to review this and take decision.

PR #14400, opened 2025-11-24
🐛 Fix status code when malformed JSON

JavierSanchezCastro:
From https://github.com/fastapi/fastapi/discussions/14399


YuriiMotov (commented):
I agree that status code `422` is incorrect and `400` should be used instead in case of malformed body.

> The HTTP 422 Unprocessable Content client error response status code indicates that the server understood the content type of the request content, and **the syntax of the request content was correct**, but it was unable to process the contained instructions.

A couple of notes regarding the implementation:
* this change is breaking for existing apps (clients may expect `422` but receive `400`). We should probably consider adding an easy way to switch to old behavior
* we should probably change the status code the same way for malformed multipart form data?
* we should document the response with `400` status code for all routes in openapi

But first, let's wait for Sebastian to approve the ideas

JavierSanchezCastro:
> We should probably consider adding an easy way to switch to old behavior

Something like 401-403 status code update?

PR #14401, opened 2025-11-24
docs: add BackgroundTasks usage example

niral04:
This PR adds a simple example showing how to use BackgroundTasks
to write a message to a log file after sending the response.

- Demonstrates adding a task with `background_tasks.add_task`.
- Helps new users understand practical usage of BackgroundTasks.


YuriiMotov:
Thanks for your interest!

There is already usage example in this part of docs. There is no need to add one more

Issue #12246, opened 2024-09-22
OpenAPI servers not being returned according how the docs say they should be


Kludex:
### Discussed in https://github.com/fastapi/fastapi/discussions/12226

<div type='discussions-op-text'>

<sup>Originally posted by **mzealey** September 19, 2024</sup>
### First Check

- [X] I added a very descriptive title here.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/pydantic/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
from fastapi import FastAPI

app = FastAPI()

# you can add a test endpoint here or not - same bug either way
```

### Description

```
$ curl localhost:8$ curl localhost:8000/openapi.json
{"openapi":"3.1.0","info":{"title":"FastAPI","version":"0.1.0"},"paths":{}}
```

According to the documentation of the `servers` parameter in FastAPI:

> If the servers list is not provided, or is an empty list, the default value would be a dict with a url value of /.

(assuming that `root_path_in_servers = True` (the default))

Clearly this is not happening.

### Operating System

Linux

### FastAPI Version

0.110.3 (but according to github code seems to be in latest also)

### Pydantic Version

2.5.3

### Python Version

Python 3.10.12

### Additional Context

_No response_</div>

Kludex:
Relevant part of the docstrings: https://github.com/fastapi/fastapi/blob/9606b916ef7c883f5ebeac1bf3db9adf5ae646a9/fastapi/applications.py#L299-L300

sinisaos:
@Kludex I don't know if this is the correct way, but we can add default value of `[{"url": "/"}]` instead of `None`. Something like this
```python
servers: Annotated[
    Optional[List[Dict[str, Union[str, Any]]]],
    Doc(
        """
        # unchanged docstring
        """
    ),
] = [{"url": "/"}],
```
Result for default:
```python
from fastapi import FastAPI

[...]

PR #14405, opened 2025-11-25
📝 Fix docstring of `servers` parameter

YuriiMotov:
Resolves / related to #12246

It's currently wrongly stated in the docstring for `servers` parameter, that by-default FastAPI will include `/` url in case this parameter is not specified or empty list:

https://github.com/fastapi/fastapi/blob/5b0625df96e4ea11b54fcb2a76f21f7ad94764fe/fastapi/applications.py#L303-L304

This is not true ([related part of code](https://github.com/fastapi/fastapi/blob/5b0625df96e4ea11b54fcb2a76f21f7ad94764fe/fastapi/applications.py#L1078-L1089)). In fact it will be an url containing the value of `root_path` (the mounting point of application), and it will be omitted in the generated openapi schema if the mounting point is `/`.

This PR fixes the docstring ad adds a "technical details" note in "behind-a-proxy".

This way we keep current behavior and just clarify the docs.
Alternatively we may want to change the behavior to add server with url equals to `/` instead of omitting `servers` property in case the app's mounting point is `/` and `servers` parameter is not specified by developer


tiangolo (approved):
Nice, thank you! I tweaked it a bit, ready now! 🍰

PR #14407, opened 2025-11-25
🗑️ Deprecate `scopes` parameter in `Security` in favor of `oauth_scopes`

YuriiMotov:
Since we now have two parameters for `Depends` \ `Security` with very similar names (`scopes` and `scope`), it might be confusing for users.

This PR:
* introduces `oauth_scopes` parameter for `Security` to replace `scopes` parameter
* deprecates `scopes` parameter in favor of `oauth_scopes`
* updates the type annotation of `scopes` and `oauth_scopes` parameters to forbid passing single string
* adds missing `scope` parameter to `Security`
* handles mistakes when user probably confuses `scopes` and `scope` (e.g. "function" values is passed to `scopes` instead of `scope`, or security scopes are passed to `scope` parameter)

Note: I've reviewed all occurrences of `scopes` in docs and renamed those of them that I think should be renamed.
I was guided by the principle: rename `scopes` to `oauth_scopes` if it's clearly about `scopes` parameter of `Security`. We shouldn't rename `scopes` if it's about the value passed to that parameter or about the concept of scopes.

For example, I don't think we need to change something in (it should be clear from the context):

> You don't necessarily need to add different scopes in different places.
> We are doing it here to demonstrate how FastAPI handles scopes declared at different levels.


docs/en/docs/advanced/security/oauth2-scopes.md
```diff
@@ -106,7 +106,17 @@ Now we declare that the *path operation* for `/users/me/items/` requires the sco
...
 For this, we import and use `Security` from `fastapi`.
 
-You can use `Security` to declare dependencies (just like `Depends`), but `Security` also receives a parameter `scopes` with a list of scopes (strings).
+You can use `Security` to declare dependencies (just like `Depends`), but `Security` also receives a parameter `oauth_scopes` with a list of scopes (strings).
+
+/// note
+
+Before version 0.122.X, the name of this parameter was `scopes`.
+
+In FastAPI 0.122.X, the `scopes` parameter was deprecated in favor of `oauth_scopes`
```
YuriiMotov, line 115: Just a reminder that we need to update the version number

fastapi/param_functions.py
```diff
@@ -2324,7 +2376,48 @@ def Security(  # noqa: N802
...
+            Security(dependency_fn, scope="function")
+            ```
+
+            To specify OAuth2 scopes use `oauth_scopes` parameter:
+
+            ```python
+            Security(dependency_fn, oauth_scopes=["items", "users"])
+            ```
+
+            &#8203;
```
YuriiMotov, line 2453: There seems to be a bug (I think in griffe) in the docstring generation for deprecated parameter - deprecation message is concatenated with docstring without linebreak and this breaks the codeblock formatting.
I added an invisible whitespace character to preserve empty line after code block closing triple back-ticks.

<details><summary>Without that invisible symbol it looks like this (see in details)</summary>


**Pay attention to the line in the middle that starts with `python Security`**

Screenshot from 2025-11-25 22-27-49

</details>

PR #14408, opened 2025-11-26
✨ feature: Add support for HTTP QUERY method

fluffy-dev:
## Description

This PR adds support for the HTTP `QUERY` method as proposed in the [IETF draft](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html).

This allows for read-only operations (safe and idempotent) that require complex parameters passed via the request body, solving the limitation of URL length limits in standard `GET` requests.

## Related Issue

Ref: #12965

## Changes

* Added `QUERY` to `METHODS_WITH_BODY` in `fastapi/openapi/constants.py`.
* Added `query()` decorator to `APIRouter` and `FastAPI` class.
* Added comprehensive tests ensuring `requestBody` generation in OpenAPI and Pydantic v1/v2 compatibility.


JavierSanchezCastro:
https://github.com/fastapi/fastapi/pull/14351#pullrequestreview-3463399942

https://github.com/fastapi/fastapi/pull/14083#issuecomment-3373573075

tiangolo:
Thanks for your interest, I'll need to revisit this later when the rest of the building blocks are in place. For now, I'll pass on this. :coffee:

PR #14409, opened 2025-11-26
🐛 Fix non-deterministic openapi schema in case of schema name duplication

mattalexanderhill:
See https://github.com/fastapi/fastapi/discussions/14376

During field remapping stage of openapi schema generation the order of iteration is not consistent across runs. This leads to non-deterministic schema generation

Preserve the order of this iteration across runs by sorting by the "new_name", i.e. the value inserted in this mapping

### DIFF
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -302,7 +302,10 @@ def _remap_definitions_and_field_mappings(
     Dict[str, Any],
 ]:
     old_name_to_new_name_map = {}
-    for field_key, schema in field_mapping.items():
+    for field_key, schema in sorted(
+        field_mapping.items(),
+        key=lambda x: model_name_map.get(x[0][0].type_),
+    ):
         model = field_key[0].type_
         if model not in model_name_map:
             continue
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -304,7 +304,7 @@ def _remap_definitions_and_field_mappings(
     old_name_to_new_name_map = {}
     for field_key, schema in sorted(
         field_mapping.items(),
-        key=lambda x: model_name_map.get(x[0][0].type_),
+        key=lambda x: model_name_map.get(x[0][0].type_, ""),
     ):
         model = field_key[0].type_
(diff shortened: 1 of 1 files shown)


YuriiMotov:
I converted it to Draft. Please mark as Ready for review when it's ready

mattalexanderhill:
Thanks @YuriiMotov should be ready for review.

I don't think I have perms to update the labels, I'll let you decide on most appropriate here

YuriiMotov (commented):
@mattalexanderhill, thanks for your interest and efforts!

I think we need to add a warning. Without warning it will hide the problem (if there is naming conflict we will just use last model, but if they have different set of fields, the final openapi schema will be invalid).

I think it should be something like:
```diff
    old_name_to_new_name_map = {}
+   name_to_model_map = {}
    for field_key, schema in sorted(
        field_mapping.items(),
        key=lambda x: model_name_map.get(x[0][0].type_, ""),
    ):
        model = field_key[0].type_
        if model not in model_name_map:
            continue
        new_name = model_name_map[model]
        old_name = schema["$ref"].split("/")[-1]
        if old_name in {f"{new_name}-Input", f"{new_name}-Output"}:
            continue
+       if old_name in name_to_model_map and name_to_model_map[old_name] is not model:
+           prev_model = name_to_model_map[old_name]
+           warnings.warn(
+               (
+                   f"Model name conflict detected for model {model.__name__}. "
+                   f"Both {model.__module__}.{model.__name__} and "
+                   f"{prev_model.__module__}.{prev_model.__name__} "
+                   f"are mapped to the same schema name {old_name}. "
+               ),
+               category=UserWarning,
+               stacklevel=2,
+           )
        old_name_to_new_name_map[old_name] = new_name
+       name_to_model_map[old_name] = model
    ...
```

I would actually raise an error, not warning.
@mattalexanderhill, what do you think?

YuriiMotov:
Alternative solution: https://github.com/fastapi/fastapi/pull/14797

YuriiMotov:
This seems to be resolved in latest versions of FastAPI - schema is deterministic again.

Thanks!

PR #14419, opened 2025-11-30
🐛 Cache dependencies that don't use scopes and don't have sub-dependencies with scopes

tiangolo:
🐛 Cache dependencies that don't use scopes and don't have sub-dependencies with scopes

Related to https://github.com/fastapi/fastapi/pull/9790, and https://github.com/fastapi/fastapi/discussions/6024

Uses the ideas in @YuriiMotov's diagrams here for the tests: https://github.com/fastapi/fastapi/discussions/6024#discussioncomment-8541913

---

Dependencies are cached based on the function. And also on the **scopes**, but (now) **only** when those scopes are used by this dependency or by a sub-dependency: this dependency or a sub-dependency declare scopes or access the `SecurityScopes`.

### DIFF
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -38,18 +38,44 @@ class Dependant:
     response_param_name: Optional[str] = None
     background_tasks_param_name: Optional[str] = None
     security_scopes_param_name: Optional[str] = None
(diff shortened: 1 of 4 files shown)


tiangolo:
This is now released as FastAPI 0.123.0 :tada:

PR #14421, opened 2025-12-01
Docs: Update outdated Pydantic v1 example in tutorial to Pydantic v2

BharathiSen:
This PR updates outdated Pydantic v1 example code in the tutorial section by replacing
`.dict()` with the correct Pydantic v2 method `.model_dump()`.

These examples still contained v1-style serialization, and this update ensures full
consistency with FastAPI’s default Pydantic v2 behavior.

I verified the documentation locally using `mkdocs serve` and confirmed that the pages
render correctly without warnings.

Files updated:
- docs_src/body/tutorial002_py310.py
- docs_src/body/tutorial003_py310.py
- docs_src/body/tutorial004_py310.py

### DIFF
--- a/docs_src/body/tutorial002_py310.py
+++ b/docs_src/body/tutorial002_py310.py
@@ -14,7 +14,7 @@ app = FastAPI()
 
 @app.post("/items/")
 async def create_item(item: Item):
-    item_dict = item.dict()
+    item_dict = item.model_dump()
     if item.tax is not None:
         price_with_tax = item.price + item.tax
         item_dict.update({"price_with_tax": price_with_tax})
--- a/docs_src/body/tutorial003_py310.py
+++ b/docs_src/body/tutorial003_py310.py
@@ -14,4 +14,4 @@ app = FastAPI()
 
 @app.put("/items/{item_id}")
 async def update_item(item_id: int, item: Item):
-    return {"item_id": item_id, **item.dict()}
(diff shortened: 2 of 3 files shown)


YuriiMotov (commented):
@Bharathis28, thank you for your interest!

Please, take a look at the note below the code example:

> In Pydantic v1 the method was called `.dict()`, it was deprecated (but still supported) in Pydantic v2, and renamed to `.model_dump()`.
> 

These code examples are not covered by tests, but they should be covered (all code examples in docs should be covered by tests). And if we add tests, they will fail when run with Pydantic V1.

I think the right way to handle this would be:
* Create a PydanticV2 versions of code examples
* Display both versions in docs in tabs (like it's done in [Custom OpenAPI content type](https://fastapi.tiangolo.com/advanced/path-operation-advanced-configuration/#custom-openapi-content-type) ([source](https://github.com/fastapi/fastapi/blob/938f4710793e89f4ba31b8867d35185dd1771a20/docs/en/docs/advanced/path-operation-advanced-configuration.md?plain=1#L156-L166)))
* Make sure that the text around this code example is updated to reflect changes
* Add tests for both Pydantic V1 and Pydantic V2 version
* Ideally we should do the same for all occurrences of outdated `.dict()` in docs (search for `.dict()` on docs website shows 4 pages)

Would you like to work on this?

Alternative is to leave this as it is until FastAPI drops the support for Pydantic V1.

PR #14427, opened 2025-12-01
Add SecretBytes and NewType support for File() parameter

Zaczero:
Support using SecretBytes type annotation with File() parameter, similar to how bytes is supported. This enables secure handling of data content that should not be logged or displayed. Also support NewType-wrapped variants of bytes and SecretBytes.


tiangolo:
Thanks for the interest, I'll pass on this one for now. :coffee:

PR #14430, opened 2025-12-02
🐛 Fix optional sequence handling with new union syntax from Python 3.10

Viicos:
https://github.com/fastapi/fastapi/pull/14297 is only taking one union type into account. I can only recommend using https://github.com/pydantic/typing-inspection for these kind of checks (see https://typing-inspection.pydantic.dev/latest/usage/ for motivation).

### DIFF
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -3,7 +3,6 @@ import warnings
 from copy import copy, deepcopy
 from dataclasses import dataclass
 from enum import Enum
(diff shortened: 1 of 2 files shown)


tiangolo (approved):
Awesome, thank you! 🙌 

I didn't know about typing-inspection, nice! We will need to wait until I remove support for Python 3.8 (soon), but it looks promising!

---

This will be available in FastAPI version 0.123.5, released in the next few hours. 🎉

PR #14445, opened 2025-12-03
Fix async wrappers using @wraps incorrectly detected as sync

aghabidareh:
## Summary

Fixes #14444

Since FastAPI 0.123.5 (PR #14434), using `functools.wraps` on an async wrapper around a sync function causes FastAPI to incorrectly treat the handler as sync, resulting in:
```
ValueError: [TypeError("'coroutine' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]
```

### Root Cause
The `is_coroutine_callable`, `is_gen_callable`, and `is_async_gen_callable` properties in `Dependant` use `_unwrapped_call` (which uses `inspect.unwrap()`) to determine if a function is async/generator. When a decorator uses `@wraps(func)`, it sets `__wrapped__` pointing to the original function, so `inspect.unwrap()` follows that chain back to the original sync function instead of checking the actual async wrapper.

### Fix
Check the actual callable first (not the unwrapped version) to correctly detect if the wrapper itself is async/generator, and only fall back to the unwrapped version for cases like `partial`.

## Example that was broken

```python
from functools import wraps
from fastapi import FastAPI

app = FastAPI()

def my_decorator(func):
    @wraps(func)
    async def wrapper():
        func()
        return 'OK'
    return wrapper

@app.get('/')
@my_decorator
def index():
    print('Hello!')
```

## Test plan
- [x] Added test case `tests/test_async_wrapper_sync_function.py` that reproduces the issue
- [x] Verified existing `tests/test_dependency_wrapped.py` tests still pass
- [x] Ran 128 dependency-related tests - all pass

### DIFF
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -86,6 +86,12 @@ class Dependant:
 
     @cached_property
     def is_gen_callable(self) -> bool:
+        if self.call is not None:
+            if inspect.isgeneratorfunction(self.call):
+                return True
+            call_dunder = getattr(self.call, "__call__", None)  # noqa: B004
+            if inspect.isgeneratorfunction(call_dunder):
+                return True
         if inspect.isgeneratorfunction(self._unwrapped_call):
             return True
         dunder_call = getattr(self._unwrapped_call, "__call__", None)  # noqa: B004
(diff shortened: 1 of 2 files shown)


aghabidareh:
Note: This PR needs the `bug` label added by a maintainer to pass the CI label check.

tiangolo:
Thanks! I'm doing it here: https://github.com/fastapi/fastapi/pull/14448

So I'll close this one. :coffee:

PR #14446, opened 2025-12-03
feature: Infer response model from function return AST

g7AzaZLO:
# feature: Infer response model from function return AST

## Summary

This PR introduces a feature that automatically infers Pydantic response models by analyzing the AST (Abstract Syntax Tree) of the endpoint function's return statement. This enables FastAPI to generate full OpenAPI schemas for endpoints that return dictionary literals, without requiring an explicit `response_model` or complex type annotations.

## Changes

- **`fastapi/utils.py`**: Added `infer_response_model_from_ast` function.
  - Implemented recursive type inference for nested dictionaries and lists.
  - Added support for basic types (`int`, `str`, `bool`, `float`) from literals and type hints.
  - Added support for simple binary operations (e.g., `10 + 5`) and comparisons.
  - **Fix**: Implemented custom AST traversal to strictly ignore return statements inside nested functions, async functions, or classes.
  - **Fix**: Added graceful fallback (returns `None`) when dictionary keys are not strings (e.g. integers), preventing runtime crashes.

- **`fastapi/routing.py`**: Updated `APIRoute` initialization.
  - Falls back to AST inference if `response_model` is not provided or is not a valid Pydantic model.
  - **Fix**: Added a check to ensure `response_model` is not explicitly `None` (e.g. when returning `Response` or `JSONResponse`) before attempting inference.

## Tests

Added `tests/test_ast_inference.py` with comprehensive coverage:
- **Basic Inference**: Simple dictionary returns with primitive types.
- **Nested Structures**: Lists of objects and deep dictionary nesting.
- **Mixed Types**: Lists with different types fall back to `List[Any]`.
- **Expressions**: Inference for calculated values (`1 + 1`).
- **Edge Cases**:
  - Explicit `JSONResponse` returns (inference disabled).
  - Nested functions (inner returns ignored).
  - Invalid dictionary keys (int keys) -> graceful degradation (no schema).
  - Empty structures (`[]`, `{}`).
  - Local variable assignment before return.
  - "Database" simulation: direct return of external function call (no schema) vs. dict construction from external data (partial schema).

## Motivation

This feature significantly reduces boilerplate for simple "RPC-style" endpoints or quick prototypes where the return structure is static and obvious from the code. It improves the developer experience by providing automatic documentation "for free" while maintaining the robustness of the framework.

### Old version
```Python

# 1. Boilerplate: We must explicitly define Pydantic models for nested structures.
# This creates a lot of extra code for simple, one-off responses.
class CustomerInfo(BaseModel):
    name: str
    vip_status: bool
    # We have to be specific about types here or use extra models
    preferences: Dict[str, Union[bool, str]]

class Item(BaseModel):
    item_id: int
    name: str
    price: float
    in_stock: bool

# 2. Boilerplate: We must define the main response model matching the dict structure.
class OrderResponse(BaseModel):
    order_id: str
    status: str
    total_amount: float
    tags: List[str]
    customer_info: CustomerInfo
    items: List[Item]
    metadata: Optional[Dict[str, Any]] = None

# 3. Explicit Binding: We must explicitly pass the class to `response_model`.
@app.get("/orders/{order_id}", response_model=OrderResponse)
async def get_order_details(order_id: str):
    # The data construction remains the same, but we had to write 
    # ~20 lines of class definitions above just to validate this dict.
    order_data = {
        "order_id": order_id,
        "status": "processing",
        "total_amount": 150.50,
        "tags": ["urgent", "new_customer"],
        "customer_info": {
            "name": "John Doe",
            "vip_status": False,
            "preferences": {"notifications": True, "theme": "dark"},
        },
        "items": [
            {
                "item_id": 1,
                "name": "Laptop Stand",
                "price": 45.00,
                "in_stock": True,
            },
        ],
        "metadata": None,
    }
    return order_data
```

### New version
Endpoint:
```Python
@app.get("/orders/{order_id}")
async def get_order_details(order_id: str) -> Dict[str, Any]:
    order_data: Dict[str, Any] = {
        "order_id": order_id,
        "status": "processing",
        "total_amount": 150.50,
        "tags": ["urgent", "new_customer"],
        "customer_info": {
            "name": "John Doe",
            "vip_status": False,
            "preferences": {"notifications": True, "theme": "dark"},
        },
        "items": [
            {
                "item_id": 1,
                "name": "Laptop Stand",
                "price": 45.00,
                "in_stock": True,
            },
        ],
        "metadata": None,
    }
    return order_data
 ```
 Response in swagger:
image

## Key Differences & Benefits

This feature allows FastAPI to automatically infer the response schema from the endpoint's return statement AST, eliminating the need for explicit Pydantic models for simple use cases.

#### 1. Reduction of Boilerplate
*   **Standard Approach:** Developers must define separate `BaseModel` classes for every response, including nested objects (e.g., `CustomerInfo`, `Item`).
*   **With AST Inference:** The schema is generated directly from the dictionary literal in the code. No extra classes are required for single-use schemas.

#### 2. Improved Developer Experience (DX)
*   **Standard Approach:** Prototyping is slower because modifying the response structure requires updating both the dictionary construction logic and the separate class definition.
*   **With AST Inference:** Developers can write idiomatic Python code (returning a `dict`). The API documentation (OpenAPI) and schema validation update automatically to reflect the code structure.

#### 3. Code Locality
*   **Standard Approach:** The schema definition is decoupled from the data generation, often located in different parts of the file or different files entirely.
*   **With AST Inference:** The "source of truth" regarding the data structure resides directly within the `return` statement of the endpoint.

#### 4. Automatic Type Safety for Simple Code
*   Projects using plain dictionaries (typed as `Dict[str, Any]`) instantly gain the benefits of automatic OpenAPI documentation and schema validation without needing a massive refactor to convert everything to Pydantic models.

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -346,12 +346,14 @@ def infer_response_model_from_ast(
     Analyze the endpoint function's source code to infer a Pydantic model
     from a returned dictionary literal or variable assignment.
     """
+    import textwrap
+
     try:
         source = inspect.getsource(endpoint_function)
     except (OSError, TypeError):
         return None
 
-    source = inspect.cleandoc(source)
+    source = textwrap.dedent(source)
     try:
         tree = ast.parse(source)
     except SyntaxError:
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -1,7 +1,6 @@
 import ast
 import inspect
 import re
-
 import warnings
 from dataclasses import is_dataclass
 from typing import (
@@ -277,12 +276,12 @@ def _infer_type_from_ast(
             return List[Any]
 
         first_type = _infer_type_from_ast(node.elts[0], func_def, context_name + "Item")
-        
+
         for elt in node.elts[1:]:
             current_type = _infer_type_from_ast(elt, func_def, context_name + "Item")
             if current_type != first_type:
                 return List[Any]
-        
+
         if first_type is not Any:
             return List[first_type]
         return List[Any]
@@ -291,7 +290,7 @@ def _infer_type_from_ast(
         left_type = _infer_type_from_ast(node.left, func_def, context_name)
         right_type = _infer_type_from_ast(node.right, func_def, context_name)
         if left_type == right_type and left_type in (int, float, str):
-             return left_type
+            return left_type
         if {left_type, right_type} == {int, float}:
             return float
 
@@ -351,60 +350,63 @@ def infer_response_model_from_ast(
         source = inspect.getsource(endpoint_function)
     except (OSError, TypeError):
         return None
-    
+
     source = inspect.cleandoc(source)
     try:
         tree = ast.parse(source)
     except SyntaxError:
         return None
-        
+
     if not tree.body:
         return None
-        
+
     func_def = tree.body[0]
     if not isinstance(func_def, (ast.FunctionDef, ast.AsyncFunctionDef)):
         return None
-        
+
     return_stmt = None
-    
 
     nodes_to_visit = list(func_def.body)
     while nodes_to_visit:
         node = nodes_to_visit.pop(0)
-        
+
         if isinstance(node, ast.Return):
             return_stmt = node
             break
-            
+
         if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
             continue
-            
+
         for child in ast.iter_child_nodes(node):
             nodes_to_visit.append(child)
-            
+
     if not return_stmt:
         return None
 
     returned_value = return_stmt.value
     dict_node = None
-    
+
     if isinstance(returned_value, ast.Dict):
         dict_node = returned_value
     elif isinstance(returned_value, ast.Name):
         variable_name = returned_value.id
         # Find assignment
         for node in func_def.body:
-            if isinstance(node, ast.AnnAssign) and isinstance(node.target, ast.Name) and node.target.id == variable_name:
-                 if isinstance(node.value, ast.Dict):
-                     dict_node = node.value
-                     break
+            if (
+                isinstance(node, ast.AnnAssign)
+                and isinstance(node.target, ast.Name)
+                and node.target.id == variable_name
+            ):
+                if isinstance(node.value, ast.Dict):
+                    dict_node = node.value
+                    break
             elif isinstance(node, ast.Assign):
-                 for target in node.targets:
-                     if isinstance(target, ast.Name) and target.id == variable_name:
-                         if isinstance(node.value, ast.Dict):
-                             dict_node = node.value
-                             break
-    
+                for target in node.targets:
+                    if isinstance(target, ast.Name) and target.id == variable_name:
+                        if isinstance(node.value, ast.Dict):
+                            dict_node = node.value
+                            break
+
     if not dict_node:
         return None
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: fastapi/utils.py [hunk 1]
```diff
@@ -258,3 +261,189 @@ def get_value_or_default(
...
+            from pydantic import create_model
+
+        return create_model(f"Model_{context_name}", **fields)  # type: ignore[call-overload]
+
+    if isinstance(node, ast.Name):
+        arg_name = node.id
+        for arg in func_def.args.args:
+            if arg.arg == arg_name and arg.annotation:
+                if isinstance(arg.annotation, ast.Name):
+                    if arg.annotation.id == "int":
```
maxachis: This case (and quite a few other cases) makes use of the [Arrow Anti Pattern](https://wiki.c2.com/?ArrowAntiPattern), which renders the code quite a bit less readable! In this case, it'd be better to replace these nested conditionals with guard clauses. For example, at `if isinstance(node, ast.Name)`:

```python
if not isinstance(node, ast.Name):
    return Any
```

tests/test_ast_inference.py
```diff
@@ -0,0 +1,441 @@
...
+    """Test that model IS created when some fields have types."""
+    result = infer_response_model_from_ast(_test_func_mixed)
+    # Should create model because "literal_field" is str, not Any
+    assert result is not None
+    assert "typed_field" in result.__annotations__
+    assert "literal_field" in result.__annotations__
+
+
+def test_infer_type_from_ast_edge_cases() -> None:
+    """Test edge cases for _infer_type_from_ast function."""
```
maxachis: I note that in some of these tests, you have a single test function for a single case, but in some functions such as this, you group them all together. Why the disparity?

fastapi/utils.py
g7AzaZLO: Updated the code and rewrote this part using guard clauses - it’s much cleaner now. Thanks for the Arrow Anti-Pattern link, that was really helpful!
```Python
    if isinstance(node, ast.Name):
        arg_name = node.id
        for arg in func_def.args.args:
            if arg.arg != arg_name:
                continue

            if not arg.annotation:
                continue

            if not isinstance(arg.annotation, ast.Name):
                continue

            annotation_id = arg.annotation.id
            if annotation_id == "int":
                return int
            if annotation_id == "str":
                return str
            if annotation_id == "bool":
                return bool
            if annotation_id == "float":
                return float
            if annotation_id == "list":
                return List[Any]
            if annotation_id == "dict":
                return Dict[str, Any]

    return Any
```

tests/test_ast_inference.py
g7AzaZLO: The code has been fixed. The edge cases are now split into separate parameterized tests:

test_infer_response_model_returns_none - for cases returning None
test_infer_response_model_success - for successful cases with field validation

The concern is no longer applicable. Thanks for pointing this out - the test structure has been improved.

maxachis (commented):
Looking on this, and bearing in mind that I'm still familiarizing myself with the innards of FastAPI, I think an issue should be made first and the topic open to broader discussion. The changes are fairly complex, with a number of conditionals. Note how many edge cases you've accounted for, as an example -- it's quite possible there are more!

It's also not clear how it reduces boilerplate as yet, since the example given just shows the new mode, not the old!

g7AzaZLO:
> It's also not clear how it reduces boilerplate as yet, since the example given just shows the new mode, not the old!

I added to the description how this looks with Pydantic models and how it looks with the new changes.

g7AzaZLO:
Link to the discussion for this proposal: https://github.com/fastapi/fastapi/discussions/14456

PR #14447, opened 2025-12-03
✨ Allow specifying `max_field`, `max_files` and `max_part_size` for routers and path operations

YuriiMotov:
FastAPI currently allows configuring upload limits only at the application level (using middleware), which makes it difficult to apply different constraints to different endpoints.
In some cases you may want to specify different values for specific router or endpoint. Currently there is no such option.

The idea of this PR is to allow specifying such limits for routers and endpoints.

As an experiment, I implemented this for `post` method. If it's Ok then will implement for app, router and other methods

```py
upload_file_router = APIRouter(form_max_files=100, max_part_size=1024*1024)


@upload_file_router.post(
    "/upload-multi", max_part_size=1024*1024*100,
    form_max_files=1000,
)
async def upload(files: List[UploadFile] = File(...)):
    pass


app = FastAPI(form_max_files=0, max_part_size=1024)

app.include_router(upload_file_router)
```

Related discussions:
* https://github.com/fastapi/fastapi/discussions/13557
* https://github.com/fastapi/fastapi/discussions/9634

### DIFF
--- a/tests/test_form_max_fields_files_part_size.py
+++ b/tests/test_form_max_fields_files_part_size.py
@@ -6,7 +6,7 @@ from fastapi.testclient import TestClient
 app = FastAPI()
 
 
-@app.post("/", form_max_files=2, form_max_part_size=1024, form_max_fields=2)
+@app.post("/", form_max_files=2, max_part_size=1024, form_max_fields=2)
 async def upload_files(files: List[UploadFile] = File(...)):
     return {"filenames": [file.filename for file in files]}
 
@@ -54,7 +54,7 @@ def test_max_part_size_exceeds_custom_limit():
         "small content\r\n"
         f"--{boundary}\r\n"
         f'Content-Disposition: form-data; name="large"\r\n\r\n'
-        + ("x" * 1024 * 10 + "x")  # 1MB + 1 byte of data
+        + ("x" * 1024 + "x")  # 1KB + 1 byte of data
         + "\r\n"
         f"--{boundary}--\r\n"
     ).encode("utf-8")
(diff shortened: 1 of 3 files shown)


tests/test_form_max_fields_files_part_size.py
```diff
@@ -0,0 +1,89 @@
...
+
+    boundary = "------------------------4K1ON9fZkj9uCUmqLHRbbR"
+
+    multipart_data = (
+        f"--{boundary}\r\n"
+        f'Content-Disposition: form-data; name="small"\r\n\r\n'
+        "small content\r\n"
+        f"--{boundary}\r\n"
+        f'Content-Disposition: form-data; name="large"\r\n\r\n'
+        + ("x" * 1024 * 10 + "x")  # 1MB + 1 byte of data
```
maxachis: Would it make more sense to have it be `1KB +1` instead of `1MB +1`, to ensure the 1KB limit is precise, or am I misunderstanding something (quite possible!)?

YuriiMotov: Good catch!
Thank you!

dolfinus:
Instead of modifying `app.post()` method, maybe it should be a part of annotation?
```
@app.post("/data")
async def handle_fiels(
  files: Annotated[List[UploadFile], Field(max_items=10), File(max_size="10MiB")]
): ...
```

YuriiMotov:
> Instead of modifying `app.post()` method, maybe it should be a part of annotation?

I think this is a bit different.. In my implementation `max_field`, `max_files` and `max_part_size` are used while parsing multipart data by Starlette. The idea of these parameters is to prevent the attacks when heavy requests are sent to cause "Out of memory" on server.
For now these parameters have defaults, but it's not easy to change these defaults for FastAPI app - people need to use middleware to call `request.form()` with custom parameters before validation

PR #14448, opened 2025-12-03
🐛 Fix support for functools wraps and partial combined, for async and regular functions and classes in path operations and dependencies

tiangolo:
🐛 Fix support for functools wraps and partial combined, for async and regular functions and classes in path operations and dependencies

This also uses and extends the tests created by @YuriiMotov :raised_hands: 

Should fix / related to: https://github.com/fastapi/fastapi/issues/14444

The new tests simulate the multiple combinations of how decorators could affect dependencies and path operation functions (endpoints).

The logic to analyze if a function is an async, generator, async generator,  callable instance, etc. now also checks and unwraps and extracts partials in all those levels.

The main case is when the callable is an _async_ something (function, generator, etc) because then it needs to be awaited.

But the original function could be async or not, the wrapper could be async or not. if either is async, then it has to be awaited, no matter if the other is not async. So, analyzing only the last point in the chain is not enough, it's necessary to analyze the function but also the wrapper.

This PR implements and tests all that, all those combinations.

### DIFF
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -15,7 +15,7 @@ else:  # pragma: no cover
     from asyncio import iscoroutinefunction
 
 
-def _unwrapped_call(call: Optional[Callable[..., Any]]) -> Optional[Callable[..., Any]]:
+def _unwrapped_call(call: Optional[Callable[..., Any]]) -> Any:
     if call is None:
         return call  # pragma: no cover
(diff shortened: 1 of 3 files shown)


tiangolo:
This is now released in FastAPI 0.123.6 :tada:

tirkarthi:
Thanks @tiangolo for the quick fix and release.

potiuk:
Thanks a LOT! That was super quick :)!

PR #14449, opened 2025-12-04
Replace `applications.py` asserts with if...raise

maxachis:
Previously, the checks for a title and version when an `openapi_url` is declared were done using _asserts_.

Since asserts can be disabled if running Python in optimized mode (i.e., with the `-O` flag), we probably don't want them here -- otherwise some validations might not occur when an app is run in optimized mode. 

I could be wrong, though 🤔.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -908,8 +908,10 @@ class FastAPI(Starlette):
         ] = "3.1.0"
         self.openapi_schema: Optional[Dict[str, Any]] = None
         if self.openapi_url:
-            assert self.title, "A title must be provided for OpenAPI, e.g.: 'My API'"
-            assert self.version, "A version must be provided for OpenAPI, e.g.: '2.1.0'"
+            if not self.title:
+                raise ValueError("A title must be provided for OpenAPI, e.g.: 'My API'")
+            if not self.version:
+                raise ValueError("A version must be provided for OpenAPI, e.g.: '2.1.0'")
         # TODO: remove when discarding the openapi_prefix parameter
         if openapi_prefix:
             logger.warning(
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -911,7 +911,9 @@ class FastAPI(Starlette):
             if not self.title:
                 raise ValueError("A title must be provided for OpenAPI, e.g.: 'My API'")
             if not self.version:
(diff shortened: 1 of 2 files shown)


YuriiMotov (commented):
@maxachis, thanks for your interest!

I don't think these two particular asserts are important - `None` is not allowed type and should be caught by static type checker, and passing empty string shouldn't break the app.

But we use `assert` in other places, and we should probably review those places and think if that can be dangerous.
If `assert` is used just to prevent misconfiguration then it's probably not important:
* app will fail in either way - with `AssertionError` or later with more cryptic error
* configuration will fail while the app is tested. I think optimization should be disabled during running tests

maxachis:
@YuriiMotov That makes sense to me! 

I did a quick search with my IDE, and I found 65 other `asserts` peppered through the code. In particular, there were 16 in [`fastapi/routing.py`](https://github.com/fastapi/fastapi/blob/a79ae3d66fb093adfe9f8d15a53af74b49947562/fastapi/routing.py) (mostly configuration), 26 in [`fastapi/depencies.utils.py`](https://github.com/fastapi/fastapi/blob/1c1e584abd5cb98920bf22966f7a8b79d3ec64be/fastapi/dependencies/utils.py), and 10 in [`fastapi/_compat/main.py`](https://github.com/fastapi/fastapi/blob/d8c691f7f09989fa51b331b8a7a1ecfe93815a2f/fastapi/_compat/main.py).

The ones that got my attention were in `_compat/main` -- it seems like those might cause someone to potentially call from the wrong version of Pydantic. I could see cases where people have only partially migrated to Pydantic v2, run in optimized mode, and get a bunch of unexpected bugs through the interactions with FastAPI. 

Because I'm still new to the code base, it's hard for me to evaluate which may or may not be problematic, but there are at least enough that it could be difficult to be confident that _every_ one of them wouldn't cause an issue.

tiangolo:
I think it's fine to keep it as it is right now, I'll pass on this one, thanks! :coffee:

njsbhu:
In application.py-
I think the changing from `assert ` to `explicit if` … raise looks good and improves runtime safety, especially when Python is run with optimizations.
As a small suggestion, we could consider whether using a more specific exception type (or documenting the intended exception usage) would help keep error handling consistent across the file.

PR #14450, opened 2025-12-04
openapi: add test for mixed Pydantic v1/v2 models

SSOBHY2:
Hi! The label check is failing because this PR has no labels and I don’t have permission to set them.

Could someone please add the `internal` label (or another appropriate one from
`breaking`, `security`, `feature`, `bug`, `refactor`, `upgrade`, `docs`, `lang-all`, `internal`) so the label checker passes?

Thanks!
---------------------------------------------------------

Summary

Add regression tests to verify OpenAPI schema generation when Pydantic v1 and v2 models are mixed in the same FastAPI application, both with and without separate_input_output_schemas.

---

Motivation

FastAPI supports applications that:

* Use Pydantic v2 for new models, and
* Still rely on Pydantic v1 models (via the compat layer) during migration.

OpenAPI generation must correctly merge schemas and references from both versions. This PR adds explicit tests to ensure:

* Schemas for both Pydantic v1 and v2 models appear in components.schemas.
* References to those schemas are valid and correctly scoped.
* Behavior is consistent when separate_input_output_schemas is enabled or disabled.

---

Changes

* New test module: tests/test_openapi_mixed_pydantic_versions.py

  * Defines Pydantic v1 models via may_v1.BaseModel: AddressV1, UserV1
  * Defines Pydantic v2 models via pydantic.BaseModel: AddressV2, UserV2 (nesting both v1 and v2 models)
  * Adds:

    * test_openapi_mixed_pydantic_models_with_separate_input_output_schemas

      * FastAPI(separate_input_output_schemas=True)
      * Ensures components.schemas contains entries for UserV1, UserV2, AddressV1, AddressV2
      * Asserts that references exist for each of those models somewhere in the OpenAPI document
    * test_openapi_mixed_pydantic_models_without_separate_input_output_schemas

      * FastAPI(separate_input_output_schemas=False)
      * Ensures the same schemas are present
      * Asserts that all reference values start with #/components/schemas/ (no broken references)

* Tests are guarded by:

  * @needs_pydanticv2 (only run with Pydantic v2)
  * @needs_py_lt_314 (skip on Python ≥ 3.14, where Pydantic v1 is not supported)

---

Testing

Executed from the repo root, in a virtual environment with requirements.txt installed:

python -m pytest tests/test_openapi_mixed_pydantic_versions.py -q

All tests pass.

---


YuriiMotov:
@SSOBHY2, why do you mark this as ready for review while tests fail?
Please, make sure tests in CI pass before asking for review!

YuriiMotov:
FastAPI completely dropped the support of Pydantic V1 since [0.128.0](https://fastapi.tiangolo.com/release-notes/#01280).
So, we can now close this PR.

@SSOBHY2, thanks for your interest and efforts!

PR #14453, opened 2025-12-04
🐛 Fix `separate_input_output_schemas=False` with `computed_field`

YuriiMotov:
This fixes the issue reported in https://github.com/fastapi/fastapi/discussions/14451.

The problem was that we calculated `override_mode` once for all models and if any of them had `computed_field` we set `override_mode=None` for all of the instead of doing it for this one only.

### DIFF
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -7,7 +7,6 @@ from typing import (
     Any,
     Dict,
     List,
-    Mapping,
     Sequence,
     Set,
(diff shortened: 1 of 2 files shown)


coltonevansid (approved):
Thank you for the quick fix! 🎉

tiangolo (approved):
Excellent job @YuriiMotov! 🙌 🚀 

This will be available in FastAPI 0.123.10 released in the next hours. 🎉

PR #14458, opened 2025-12-04
🐛 Fix using class (not instance) dependency that has `__call__` method

YuriiMotov:
As reported in https://github.com/fastapi/fastapi/discussions/14452, FastAPI 0.123.6 introduced a regression - if we use class (not instance) as dependency and this class has `__call__` method that is coroutine or generator, it is mistakenly treated as coroutine\generator (actually class is not coroutine\generator, instance of this class is).

So, to fix this we need to check if the callable is class before inspecting its `dunder_call` in `is_coroutine_callable`, `is_async_gen_callable`, and `is_gen_callable`.

### DIFF
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -96,6 +96,8 @@ class Dependant:
             _impartial(self.call)
         ) or inspect.isgeneratorfunction(_unwrapped_call(self.call)):
             return True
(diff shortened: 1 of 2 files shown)


tiangolo (approved):
Great job @YuriiMotov! Thank you! 🙌 

This will be available in FastAPI 0.123.10 in the next hours. 🎉

PR #14468, opened 2025-12-05
Add upload file size limit

nancywang-cwru:
Background: 
For my computer security course project, we were asked to scan vulnerabilities for an open source project and try to fix them. I used Bandit for SAST and ZAP for DAST on a Linux VM. One possible vulnerability I found was FastAPI has no upload file size limitations, large multipart file uploads are allowed without limits, referencing OWASP API4 - https://owasp.org/www-project-api-security/ Well, it greatly depends on the isolated resource I separate out for the test environment, and the smaller I assign it, the easier it could break. I guess in production, you could also trust the ASGI server (Uvicorn) and reverse proxies (Nginx) to handle limits. But it would be good to add a limit check to the FastAPI framework.
Changes:
Add a new optional parameter max_size to File() and enforce file upload size limits during request parsing. This resolves the unbounded file upload DoS vector identified during DAST.
It is optional, I’ve tested without setting the parameter – keep File(…) as None will work the same way it did before; after setting the parameter, I’ve tested uploading a file that exceeds the limit and got the 413 Request Entity Too Large Error I was looking for.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -18,11 +18,15 @@ from fastapi import routing
 from fastapi.datastructures import Default, DefaultPlaceholder
 from fastapi.exception_handlers import (
     http_exception_handler,
+    request_entity_too_large_exception_handler,
     request_validation_exception_handler,
     websocket_request_validation_exception_handler,
-    request_entity_too_large_exception_handler,
 )
-from fastapi.exceptions import RequestValidationError, WebSocketRequestValidationError, RequestEntityTooLarge
+from fastapi.exceptions import (
+    RequestEntityTooLarge,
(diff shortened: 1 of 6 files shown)


YuriiMotov:
@nancywang-cwru, thanks for your interest!

It would be nice to have the ability to set the file size limit, but unfortunately it's not such simple task.
Your approach is only suitable for `bytes` fields as in this case files are always read and returned as `bytes`.
But with `file: UploadFile` annotations we shouldn't read file and put its content as `bytes`.

After some investigations, I think the best way to defend against such kind of attacks is to set up reverse proxy or API gateway.

PR #14470, opened 2025-12-08
fix: remove deprecated Pydantic v1 JSON schema methods

hemanth-thirthahalli:
Remove deprecated Pydantic v1 JSON schema methods from `UploadFile` in `fastapi/datastructures.py`.

- Deleted `__modify_schema__` (marked with TODO for v1 deprecation)
- Deleted static `__get_pydantic_json_schema__` that always returned `{"type": "string", "format": "binary"}`

These methods are no longer needed in Pydantic v2 and were actively preventing metadata (title, description, examples) from `Annotated[UploadFile, File(...)]` from appearing in the generated OpenAPI schema and docs.

With only `__get_pydantic_core_schema__` defined, Pydantic v2 now correctly generates the file schema and preserves all user-provided metadata.

### DIFF
--- a/fastapi/datastructures.py
+++ b/fastapi/datastructures.py
@@ -2,7 +2,6 @@ from typing import (
     Any,
     BinaryIO,
     Callable,
-    Dict,
     Iterable,
     Optional,
     Type,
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@hemanth-thirthahalli. thanks for your interest, but I think we can't do it before dropping the support of Pydantic V1

hemanth-thirthahalli:
@YuriiMotov  Got it, makes sense. Closing the PR for now..
If there’s anything I can help with to speed up dropping Pydantic v1 support just ping me, more than happy to jump in.

PR #14476, opened 2025-12-09
♻️ Refactor `request_response` and `websocket_session` to avoid name shadowing

ishansurdi:
## Problem

Variable shadowing exists in `request_response()` and `websocket_session()` where an inner `app` function shadows the outer `app` function. This violates PEP 8 best practices and reduces code clarity.

**Before:**
```python
async def app(scope, receive, send):
    request = Request(scope, receive, send)
    async def app(scope, receive, send):  # Shadows outer 'app'
        # implementation...
    await wrap_app_handling_exceptions(app, request)(...)  # Which 'app'?
```
## Solution

Renamed inner functions to `[inner_app]` to eliminate shadowing.
**After:**
```python
async def app(scope, receive, send):
    request = Request(scope, receive, send)
    async def inner_app(scope, receive, send):  # Clear!
        # implementation...
    await wrap_app_handling_exceptions(inner_app, request)(...)  # Obvious!
```
## Changes
- Line 99, 138: Renamed nested `[async def app]` →` [async def inner_app] `
- Line 118, 146: Updated` [wrap_app_handling_exceptions(app, ...)] `→` [wrap_app_handling_exceptions(inner_app, ...)]`
- Added `[test_variable_shadowing_fix.py]` for verification

## Impact
- Zero functional changes - pure refactoring
- Zero performance impact - identical bytecode
- 100% backward compatible - all existing code works
- Improves maintainability - clearer code intent

## Why It's Safe
- Only renamed a local variable within two functions
- Return values unchanged - both functions still return outer [app]
- External callers `([APIRoute], [APIWebSocketRoute])` see no difference
- Follows naming pattern in Starlette's original implementation

## Testing
`python tests/test_variable_shadowing_fix.py`

   
## Type of Changes
[y] Code quality improvement (non-breaking)
[y]Follows PEP 8 best practices


YuriiMotov:
I think there is no need for tests here.
This functionality has already been covered by tests and we have a linter that ensures syntax is correct.

I converted this PR to draft. Please, mark as Ready for review when it's ready

YuriiMotov (approved):
LGTM!

I agree it looks better with distinct names

svlandeg (approved):
Makes sense!

PR #14480, opened 2025-12-09
📝 Add validation error example for `Enum` path parameter

ibrahimpelumi6142:
## Description
Adds an example showing validation errors when using Enum path parameters in FastAPI.

## Problem
The tutorial previously did not include clear examples of what happens when a path parameter fails Enum validation, which can be confusing for beginners.

## Solution
- Adds a JSON-formatted example showing the validation error FastAPI returns for an invalid Enum value.
- Highlights the `loc`, `type`, `msg`, and `input` fields in the response.

## Impact
Helps beginners understand FastAPI's automatic validation for Enums and interpret error responses correctly.

## Type of Change
- [x] Documentation improvement
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist
- [x] Documentation follows style guide
- [x] Improves clarity for beginners
- [x] No technical errors introduced


docs/en/docs/tutorial/path-params.md
```diff
@@ -166,6 +166,40 @@ Because the available values for the *path parameter* are predefined, the intera
...
+
+And the client sends an invalid value:
+
+```
+GET /models/vgg16
+```
+
+FastAPI returns a clear validation error:
+
+```
```
BastienGimbert: Add JSON to the Markdown code block

```suggestion
```JSON
```

docs/en/docs/tutorial/body.md
```diff
@@ -20,6 +20,13 @@ As it is discouraged, the interactive docs with Swagger UI won't show the docume
 
 ## Import Pydantic's `BaseModel` { #import-pydantics-basemodel }
 
+/// note
+
+`BaseModel` is the main class from Pydantic used for creating data models.
+FastAPI needs this import to validate request bodies and generate documentation.
+
+///
```
YuriiMotov: I think we don't need to add this clarification.

It should be clear that we need to import class before using it.

YuriiMotov (commented):
@ibrahimpelumi6142, thank you for your interest!

I think we don't need the clarification for `BaseModel` import.

But the "Common Validation Errors with Enums" part looks nice to me.
How about removing the part about `BaseModel` and changing the title and description of this PR to reflect its content?

ibrahimpelumi6142:
Thanks for the feedback, Yurii, I appreciate it.

That makes sense. I will remove the BaseModel clarification from the tutorial and update the PR title and description to focus on the Enum validation errors section.

Glad to hear the Enum example looks useful. I will push the update shortly.

ibrahimpelumi6142:
Thanks for the clarification. I have removed the BaseModel note and updated the PR to focus on Enum validation errors.

YuriiMotov:
> Thanks for the clarification. I have removed the BaseModel note and updated the PR to focus on Enum validation errors.

PR description is still about `BaseModel`. Please update it too

ibrahimpelumi6142:
Thanks for the feedback! I’ve updated the PR description accordingly.

docs/en/docs/tutorial/path-params.md
```diff
@@ -160,6 +160,40 @@ Because the available values for the *path parameter* are predefined, the intera
 
 <img src="/img/tutorial/path-params/image03.png">
 
+### Common Validation Errors with Enums { #common-validation-errors-with-enums }
```
YuriiMotov:
```suggestion
### Request Validation with Enums { #request-validation-with-enums }
```

YuriiMotov (approved):
LGTM!

I would only change the title of section. Suggested my idea.

ibrahimpelumi6142:
Thanks! I’ve updated the section title as suggested.

tiangolo:
Thanks for the interest. I currently don't really think this is necessary, so I'll pass on this one. :coffee:

PR #14487, opened 2025-12-10
📝 Update docs about re-raising validation errors, do not include string as is to not leak information

tiangolo:
📝 Update docs about re-raising validation errors, do not include string as is to not leak information


tiangolo:
This should fix https://github.com/fastapi/fastapi/issues/10424

PR #14490, opened 2025-12-10
🐛 Fix openapi for response model with nested computed field

YuriiMotov:
When model with computed field is used for both, input and output validation, and this model is included as nested into response model, generated openapi schema is wrong (reference to `input` schema is used in response model schema).

To reason is that `_has_computed_fields` only checks the current model has computed fields, but we need to also check if any nested models has computed fields.

See: https://github.com/fastapi/fastapi/discussions/14489


YuriiMotov:
Checked, test still fails on `master` branch

JBarrioGarcia:
Are there any news or merge date expectations about this topic? 
Latest FastAPI version (`0.136.3`) still has the original bug so I can't upgrade the version since `0.123.2` as I depend on the Openapi schema. 
With this simple script can be checked the current issue:

```python
from pprint import pprint

from fastapi import FastAPI
from pydantic import BaseModel, computed_field

class UserResponse(BaseModel):
    name: str
    surname: str

    @computed_field
    @property
    def full_name(self) -> str:
        return f"{self.name} {self.surname}"

app = FastAPI(title="Input-Output checker", separate_input_output_schemas=False)

@app.get("/user")
def get_users() -> list[UserResponse]:
    return [UserResponse(name=f"User {i}", surname=f"Surname {i}") for i in range(10)]

pprint(app.openapi())
```

image

PR #14492, opened 2025-12-11
📝 Fix example of license identifier in documentation

johnson-earls:
Fix the Apache 2.0 license identifier to be `Apache-2.0`, not `MIT` (which is NOT an Apache 2.0 license).


svlandeg (approved):
Thanks for the PR @johnson-earls!

It's really just an example in the docs, but I agree it would be less confusing to have a value that actually makes sense 🙈 

I double checked and nothing else in the documentation needs to change after this edit.

tiangolo (approved):
Thanks! 🍰

PR #14512, opened 2025-12-12
🐛 Fix support for tagged union with discriminator inside of `Annotated` with `Body()`

tiangolo:
🐛 Fix support for tagged union with discriminator inside of `Annotated` with `Body()`

Related to: https://github.com/fastapi/fastapi/discussions/14495, https://github.com/fastapi/fastapi/issues/14508


tiangolo:
This will be available in FastAPI 0.124.3, released in the next few hours. :tada:

PR #14518, opened 2025-12-13
🔒 Escape HTML in Swagger and ReDoc titles and URLs

berkkaradalan:
## Summary
Fix potential XSS vulnerability in Swagger UI and ReDoc documentation pages by escaping `title` and `openapi_url` parameters.

## Details
- Added `html.escape()` to `title` in both `get_swagger_ui_html()` and `get_redoc_html()`
- Added `html.escape()` to `openapi_url` in both functions

Related discussion: #14406

### DIFF
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -1,3 +1,4 @@
+import html as html_escape
 import json
 from typing import Any, Dict, Optional
 
@@ -121,7 +122,7 @@ def get_swagger_ui_html(
     <head>
     <link type="text/css" rel="stylesheet" href="{swagger_css_url}">
     <link rel="shortcut icon" href="{swagger_favicon_url}">
-    <title>{title}</title>
+    <title>{html_escape.escape(title)}</title>
     </head>
     <body>
     <div id="swagger-ui">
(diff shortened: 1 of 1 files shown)


dolfinus:
> potential XSS vulnerability

XSS vulnerability is when user specified HTML is rendered in HTML document without any escaping. But OpenAPI schema is generated based on application routes metadata, which is not user-specified.

AleksaMCode:
As these values are not user-specified, this is not an [XSS vulnerability](https://owasp.org/www-community/attacks/xss/), at least not in the traditional sense. I don't see a way for an external user to inject their own malicious script here. Potentially, if backend developers are not careful, they could experience some unexpected rendering in the browser. IMHO, this is still a rendering issue and not an "attack" from an external user.

berkkaradalan:
Thanks both for taking the time to review and explain, as i mentioned in the discussion, i just noticed it and wanted to share. Feel free to close the PR and discussion if you don't think it's worth addressing. Again thanks for answering

P-Muench:
I want to add here that in a project of mine I use FastAPI in part as an API Gateway where OpenAPI specs are indeed generated dynamically based on user input. I am aware this is kind of a niche use case but I would be happy to have the html-rendering code to take care of sanitization. 

Just because it is not usually user-specified does not mean that nobody abuses the functionality that way (here: me).

tiangolo:
Thanks for the interest. As these functions are not expected to be used with user input, I would consider this currently out of scope, so I'll pass on this for now. Thanks! ☕

PR #14523, opened 2025-12-14
Docs second pr

Malak993-lho:
This PR improves the wording in the Typer section for better clarity and flow.

No functional changes were introduced.


YuriiMotov:
@Malak993-lho, thanks for your interest!

I think we don't need to change this. I'll close this to save Sebastian's time

PR #14524, opened 2025-12-14
docs: clarify async usage in example section

Malak993-lho:
This PR improves the documentation by adding a small clarification about when to use
`async def` in FastAPI examples.

The change helps beginners better understand async usage without introducing any
functional or behavioral changes.

No code logic was modified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
   <a href="https://fastapi.tiangolo.com"><img src="https://fastapi.tiangolo.com/img/logo-margin/logo-teal.png" alt="FastAPI"></a>
 </p>
 <p align="center">
-    <em>FastAPI framework, high performance, easy to learn, fast to code, ready for production</em>
+   <em>FastAPI framework: high performance, easy to learn, fast to code, and ready for production</em>
 </p>
 <p align="center">
 <a href="https://github.com/fastapi/fastapi/actions?query=workflow%3ATest+event%3Apush+branch%3Amaster" target="_blank">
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -178,7 +180,11 @@ def read_item(item_id: int, q: Union[str, None] = None):
 <details markdown="1">
 <summary>Or use <code>async def</code>...</summary>
 
-If your code uses `async` / `await`, use `async def`:
+!!! tip
+    Use `async def` only when your code performs non-blocking I/O,
+    such as database queries or external API calls.
```
YuriiMotov, line 185: I don't think this clarifies things..
The term `non-blocking I/O` is not described here and it may be even more confusing.
I think the original version is simpler for newcomers

YuriiMotov:
@Malak993-lho, thanks for your interest and efforts!

I suggest we close this - I don't think any clarifications are really needed there

README.md
Malak993-lho: Thanks for the feedback, that makes sense 

I agree that introducing the term "non-blocking I/O" here may add unnecessary complexity for newcomers.

I’ll update the PR to either simplify the wording further or revert to the original version to keep this section beginner-friendly.

Pyth3rEx (changes requested):
+1 on what was said earlier. We need to keep it simple for newcomers while still being precise. How about the new sentencing but with a * explaining the meaning of the I/O manipulation.

alejsdev:
Closing this as duplicate of https://github.com/fastapi/fastapi/pull/14527

PR #14541, opened 2025-12-16
🔥 Remove translation to emoji to simplify the new setup with LLM autotranslations

tiangolo:
🔥 Remove translation to emoji to simplify the new setup with LLM autotranslations


xmdnx:
image

PR #14547, opened 2025-12-16
🔧 Add LLM prompt file for Turkish, generated from the existing translations

tiangolo:
🔧 Add LLM prompt file for Turkish, generated from the existing translations

---

This will be used to update existing translations and generate new ones.

### DIFF
--- a/docs/tr/llm-prompt.md
+++ b/docs/tr/llm-prompt.md
@@ -25,12 +25,17 @@ Language code: tr.
 
 ### Preferred translations / glossary
 
-Prefer Turkish translations used in the current docs:
+Do not translate technical terms like path, route, request, response, query, body, cookie, and header, keep them as is.
 
-- request (HTTP): istek
-- response (HTTP): yanıt
-- path operation: yol operasyonu
-- path operation function: yol operasyonu fonksiyonu
+- Suffixing is very important, when adding Turkish suffixes to the English words, do that based on the pronunciation of the word and with an apostrophe.
+
+- Suffixes also changes based on what word comes next in Turkish too, here is an example:
+
+"Server'a gelen request'leri intercept... " or this could have been "request'e", "request'i" etc.
+
+- Some words are tricky like "path'e" can't be used like "path'a" but it could have been "path'i" "path'leri" etc.
+
+- You can use a more instructional style, that is consistent with the document, you can add the Turkish version of the term in parenthesis if it is not something very obvious, or an advanced concept, but do not over do it, do it only the first time it is mentioned, but keep the English term as the primary word.
 
 ### `///` admonitions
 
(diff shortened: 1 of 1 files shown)


docs/tr/llm-prompt.md
```diff
@@ -0,0 +1,47 @@
...
+
+1) Keep quote style consistent with the existing Turkish docs (they frequently use ASCII quotes in prose).
+2) Never change quotes inside inline code, code blocks, URLs, or file paths.
+
+### Ellipsis
+
+1) Keep ellipsis style consistent with existing Turkish docs.
+2) Never change `...` in code, URLs, or CLI examples.
+
+### Preferred translations / glossary
```
ycd, line 26: I think this is a great opportunity to rewrite these, like I never heard anyone using these terms IRL, but if we prefer to convert like this for other languages too, then it's completely fine!

ycd (approved):

docs/tr/llm-prompt.md
tiangolo: Ah, then actually it's probably better to explicitly instruct the LLM to _not_ convert them if those are not used IRL. 🤔 

What do you think?

ycd: Yeah, that would be better IMO, even I got confused with "yol operasyonu fonksiyonu" 😂

I agree, we should explicitly state this, we can add something like this to the prompt:


```txt
- Do not translate the terms like path, route, request, response, query, body, cookie, and header, keep them as it is. 
- Suffixing is very important, when adding Turkish suffixes to the English words, do that based on the pronunciation of the word and with an apostrophe.  
    - Suffixes also changes based on what word comes next in Turkish too, here is an example: "Server'a gelen request'leri intercept... " or this could have been "request'e", "request'i" etc. 
    - Some words are tricky like "path'e" can't be used like "path'a" but it could have been "path'i" "path'leri" etc. 
- You can use a more instructional style, that is consistent with the document, you can add the the Turkish version of the term in parenthesis if it is not something very obvious, or an advanced concept, but do not over do it, do it only the first time it is mentioned, but keep the English term as the primary word. 
```

These are the probably most tricky and important parts, when keeping the term(s) English.

 Also [Turkish](https://news.ycombinator.com/item?id=42343920) is weird and [known](https://blog.codinghorror.com/whats-wrong-with-turkey/) to be quite [problematic](https://news.ycombinator.com/item?id=45559767) too. 😭

tiangolo: Perfect, thank you! I added your suggestions, replacing the old ones.

Do you think this looks good as it is?

ycd: Yeah, it looks looks good now, thank you!


I'm excited for the PR 🙂

tiangolo: Awesome, thank you! 🚀

PR #14555, opened 2025-12-17
🔧 Temporarily disable translations still in progress, being migrated to the new LLM setup

YuriiMotov:
This is to prevent building docs from failing while we are updating code includes.
We will enable these translations later when update them with LLM


tiangolo (approved):
Thank you! 🚀

Issue #13056, opened 2024-12-10
Can't use `Annotated` with `ForwardRef`


Kludex:
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

### Issue Content

The following code doesn't generate the correct OpenAPI json:

```py
from __future__ import annotations

from dataclasses import dataclass
from typing import Annotated

from fastapi import Depends, FastAPI

app = FastAPI()


def get_potato() -> Potato:
    return Potato(color='red', size=10)


@app.get('/')
async def read_root(potato: Annotated[Potato, Depends(get_potato)]):
    return {'Hello': 'World'}


@dataclass
class Potato:
    color: str
    size: int
```

If we move the `Potato` up, or remove the `Annotated`, then it works as expected.

rahmanshamit:
The issue is with 'from __future__ import annotations' and 'Annotated[Potato, Depends(get_potato)]', right?
Since Potato class is defined later, in the read_root function __future__.annotations may be trying to store Potato as a string literal, but then Annotated wraps the Potato type, which FastAPI needs to unwrap and evaluate during schema generation, generating the wrong OpenAPI json?
What does the generated json look like?

samuelcolvin:
No I don't think so. I think it's just the use of `from __future__ import annotations`, I think you're get the same error if `Potato` was defined earlier.

Also, please format your message properly to make it easier to read.

Kludex:
> [...] I think you'll get the same error if Potato was defined earlier.

No, the problem here is `Potato` after.

> What does the generated json look like? Do we get 'Annotated[Potato, Depends(get_potato)]' rather than the actual type?

The `potato` is a query parameter on the generated OpenAPI. But you can run it, and check by yourself... 👀

Viicos:
Probably FastAPI needs to implement its own forward annotation evaluation logic, and perhaps a "rebuild" pattern for such use cases. Currently, upon creation of the API route, the `Dependant` instance is not defined correctly:

```python
from __future__ import annotations

from dataclasses import dataclass
from typing import Annotated

from fastapi import Depends, FastAPI

app = FastAPI()

def get_potato() -> Potato:
    return Potato(color='red', size=10)


@app.post('/')
async def read_root(potato: Annotated[Potato, Depends(get_potato)]):
    return {'Hello': 'World'}


@dataclass
class Potato:
    color: str
    size: int

app.routes[-1].dependant
[...]

PR #14557, opened 2025-12-17
Fix Annotated with ForwardRef when using future annotations

majiayu000:
## Summary
- Fixes issue where `Annotated[SomeClass, Depends()]` with forward references failed when using `from __future__ import annotations`
- Adds partial resolution of Annotated string annotations to extract metadata (Depends, Query, etc.) even when the inner type cannot be fully resolved
- Uses `get_type_hints()` with `include_extras=True` for better forward reference resolution

## Test plan
- [x] New test case `test_annotated_forward_ref.py` that reproduces issue #13056
- [x] Existing dependency-related tests pass

Fixes #13056


majiayu000:
Hi maintainers, could you please add the `bug` label to this PR? This fixes issue #13056 where `Annotated` with `ForwardRef` doesn't work when using `from __future__ import annotations`. Thank you\!

PR #14561, opened 2025-12-17
Security dependencies compatibility with websockets

P-Muench:
As described in #14551, the security classes such as `APIKeyCookie` could not be used in websocket endpoints. Changing the typing annotation to `HTTPConnection`, the base class of `Request` and `WebSocket`, the dependency resolver can now pass the websockets to their call methods (see also #8503).

Added tests to check if websocket connections are accepted or denied when supplied authentication.


P-Muench:
@tiangolo: Would be grateful if you could give me some feedback on this :)

PR #14567, opened 2025-12-18
vulnerability in Swagger UI template generation

cardosource:
A Cross-Site Scripting (XSS) vulnerability was identified in the get_swagger_ui_html() function, which generates the Swagger UI documentation page (/docs)

Vulnerable code before the fix
```
for key, value in current_swagger_ui_parameters.items():
    html += f"{json.dumps(key)}: {json.dumps(jsonable_encoder(value))},\n"

``` 

This part of the code directly concatenates JSON parameters inside the HTML `<script>` block, allowing user-controlled values to be interpreted as executable JavaScript code.

**Impact**

✅ Authentication required: No (the `/docs` endpoint is public by default)
✅ Exploitation complexity: Low
✅ Impact: Execution of arbitrary JavaScript code in the user’s context
✅ Risk: Token theft, credential compromise, phishing attacks

**Fix:**

Parameters were previously concatenated directly into the HTML, allowing JavaScript injection. Now, the entire configuration is serialized once using json.dumps() before injection, ensuring that JSON is treated strictly as data.

""Each parameter was serialized individually and concatenated as a string, allowing malicious values to be injected directly into the JavaScript code.""

### DIFF
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -110,51 +110,51 @@ def get_swagger_ui_html(
     [FastAPI docs for Configure Swagger UI](https://fastapi.tiangolo.com/how-to/configure-swagger-ui/)
     and the [FastAPI docs for Custom Docs UI Static Assets (Self-Hosting)](https://fastapi.tiangolo.com/how-to/custom-docs-ui-assets/).
     """
-    current_swagger_ui_parameters = swagger_ui_default_parameters.copy()
+    swagger_config = {
+        "url": openapi_url,
+        "dom_id": "#swagger-ui",
+        "layout": "BaseLayout",
+        "deepLinking": True,
+        "showExtensions": True,
+        "showCommonExtensions": True,
+        "presets": [
+            "SwaggerUIBundle.presets.apis",
+            "SwaggerUIBundle.SwaggerUIStandalonePreset"
+        ]
+    }
+    
     if swagger_ui_parameters:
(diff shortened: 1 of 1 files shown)


YuriiMotov:
As far as I can see the description is the same as in https://github.com/fastapi/fastapi/pull/14568

So, the question is the same:

> @cardosource, thanks for your interest!
> 
> > This part of the code directly concatenates JSON parameters inside the HTML <script> block, allowing user-controlled > > values to be interpreted as executable JavaScript code.
> 
> I think users can't pass these parameters, developers can. And I don't think developers will pass something malicious.

fastapi/openapi/docs.py
```diff
@@ -110,51 +109,53 @@ def get_swagger_ui_html(
...
-        html += f"oauth2RedirectUrl: window.location.origin + '{oauth2_redirect_url}',"
-
-    html += """
-    presets: [
-        SwaggerUIBundle.presets.apis,
-        SwaggerUIBundle.SwaggerUIStandalonePreset
-        ],
-    })"""
+        swagger_config["oauth2RedirectUrl"] = (
+            f"window.location.origin + '{oauth2_redirect_url}'"
```
dolfinus, line 130: This piece of JavaScript code will stop working after wrapping it with `json.dumps` - instead of accessing a variable `window`on browser side, it became a literal string. Same for `presets`, previously they are referred to some JavaScript objects loaded in swagger.js, now they are just strings.

PR #14568, opened 2025-12-18
Refactor Swagger UI parameter handling in docs.py

cardosource:
A Cross-Site Scripting (XSS) vulnerability was identified in the get_swagger_ui_html() function, which generates the Swagger UI documentation page (/docs)

Vulnerable code before the fix
```
for key, value in current_swagger_ui_parameters.items():
    html += f"{json.dumps(key)}: {json.dumps(jsonable_encoder(value))},\n"

``` 

This part of the code directly concatenates JSON parameters inside the HTML `<script>` block, allowing user-controlled values to be interpreted as executable JavaScript code.

**Impact**

✅ Authentication required: No (the `/docs` endpoint is public by default)
✅ Exploitation complexity: Low
✅ Impact: Execution of arbitrary JavaScript code in the user’s context
✅ Risk: Token theft, credential compromise, phishing attacks

**Fix:**

Parameters were previously concatenated directly into the HTML, allowing JavaScript injection. Now, the entire configuration is serialized once using json.dumps() before injection, ensuring that JSON is treated strictly as data.

""Each parameter was serialized individually and concatenated as a string, allowing malicious values to be injected directly into the JavaScript code.""

### DIFF
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -131,24 +131,36 @@ def get_swagger_ui_html(
     const ui = SwaggerUIBundle({{
         url: '{openapi_url}',
     """
-
+    
     for key, value in current_swagger_ui_parameters.items():
-        html += f"{json.dumps(key)}: {json.dumps(jsonable_encoder(value))},\n"
+        
+        encoded_value = jsonable_encoder(value)
+        safe_value = json.dumps(encoded_value)
+        
+        
+        if isinstance(value, str) and safe_value.startswith('"') and safe_value.endswith('"'):
+            safe_value = safe_value[1:-1]  # Remove aspas externas
+        
+        
+        html += f"        {key}: {safe_value},\n"
 
+  
     if oauth2_redirect_url:
-        html += f"oauth2RedirectUrl: window.location.origin + '{oauth2_redirect_url}',"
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@cardosource, thanks for your interest!

> This part of the code directly concatenates JSON parameters inside the HTML <script> block, allowing user-controlled values to be interpreted as executable JavaScript code.

I think **users** can't pass these parameters, **developers** can. And I don't think developers will pass something malicious.

fastapi/openapi/docs.py
```diff
@@ -133,22 +133,34 @@ def get_swagger_ui_html(
...
-        html += f"{json.dumps(key)}: {json.dumps(jsonable_encoder(value))},\n"
+        encoded_value = jsonable_encoder(value)
+        safe_value = json.dumps(encoded_value)
+
+        if (
+            isinstance(value, str)
+            and safe_value.startswith('"')
+            and safe_value.endswith('"')
+        ):
+            safe_value = safe_value[1:-1]  # Remove aspas externas
```
dolfinus, line 144: Instead of producing output like:
```javascript
{"key": "value"}
{"key": "with \"quotes\""}
```
this change produces output like this:
```javascript
{key: value}
{key: with \"quotes\"}
```

This is just wrong.

PR #14322, opened 2025-11-10
📝 Clarify misleading "Required, can be `None`" section in `query-params-str-validations.md`

ChaitanyaSai-Meka:
The previous explanation suggested that a query parameter could be both required and accept a true `None` value.
However, since HTTP query parameters are transmitted as strings, clients cannot actually send `None` or `null`.

This update clarifies that behavior and adds an example showing how to manually handle `"None"` or empty strings as `None` inside a route.

PR #14569, opened 2025-12-18
✅ Add missing tests for code examples

YuriiMotov:
Currently not all code examples covered by tests.

Coverage doesn't show missing code examples because it only works fore modules that are run during tests or modules that are placed in dirrectories with `__init__.py` files.
Some directories inside `docs_src` missed `__init__.py` files.

This PR:
* Adds missing `__init__.py` files in sub-directories of `docs_src`
* Adds missing tests for code examples

---

I parameterized all new tests to be run only with Pydantic V2 to simplify review

---

I added some files to coverage "omit" config:
* `docs_src/dependencies/tutorial008_an_py39.py` - it's not trivial task to test it since it fails at import time
* `docs_src/dependencies/tutorial013_an_py310.py` -  temporary code example?
* `docs_src/dependencies/tutorial014_an_py310.py` - temporary code example?
* several Pydantic V1 code examples
---

There is a JS code example not covered by tests: https://github.com/fastapi/fastapi/blob/master/docs_src/generate_clients/tutorial004.js

---

I renamed `dataclasses` and `graphql` directories inside `docs_src` to `dataclasses_` and `graphql_` (added underscores in the end) - after I added `__init__.py` to these directories coverage started giving an error while discovering files.

---

There are 2 failing tests (marked with `xfail`) - it's related to https://github.com/fastapi/fastapi/issues/12419. We can later fix this, there is [PR](https://github.com/fastapi/fastapi/pull/14322)


tiangolo (approved):

PR #14571, opened 2025-12-19
📝 Improve wording in background tasks documentation

dhruvgoyal-ai:
This PR improves the wording of a couple of sentences in the Background Tasks documentation
to enhance clarity and readability. No functional changes.

### DIFF
--- a/docs/en/docs/tutorial/background-tasks.md
+++ b/docs/en/docs/tutorial/background-tasks.md
@@ -1,8 +1,8 @@
 # Background Tasks { #background-tasks }
 
-You can define background tasks to be run *after* returning a response.
+You can define background tasks to be run after a response has been returned.
 
-This is useful for operations that need to happen after a request, but that the client doesn't really have to be waiting for the operation to complete before receiving the response.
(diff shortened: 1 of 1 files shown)


docs/en/docs/tutorial/background-tasks.md
```diff
@@ -1,8 +1,8 @@
 # Background Tasks { #background-tasks }
 
-You can define background tasks to be run *after* returning a response.
+You can define background tasks to be run after a response has been returned.
```
YuriiMotov, line 3: I think original version is better here

YuriiMotov (commented):
@dhruvgoyal-ai, thanks for your interest in FastAPI!

I think we should keep first phrase (`*after* returning a response`) as it is. I'm not a native English speaker, but it reads natural for me and I asked ChatGPT and it agreed that original wording is better in this case (less formal and emphasizes the `after` word).

As for second phrase (`when the client does not need to wait for`), I agree it's better in your variant!

alejsdev:
Hi @dhruvgoyal-ai, we think the current version is fine as it is, so we’ll close this PR for now, but thank you.

PR #14573, opened 2025-12-19
📝 Fix typing issue in `docs_src/app_testing/app_b` code example

timakaa:


docs_src/app_testing/app_b_an_py310/main.py
```diff
@@ -3,22 +3,26 @@
...
 
+foo_item = Item(id="foo", title="Foo", description="There goes my hero")
+bar_item = Item(id="bar", title="Bar", description="The bartenders")
+
+fake_secret_token = "coneofsilence"
+
+fake_db: dict[str, Item] = {
+    "foo": foo_item,
+    "bar": bar_item,
+}
```
YuriiMotov:
```suggestion
fake_secret_token = "coneofsilence"

fake_db: dict[str, Item] = {
    "foo": Item(id="foo", title="Foo", description="There goes my hero"),
    "bar": Item(id="bar", title="Bar", description="The bartenders"),
}
```

I think there is no need in `foo_item` and `bar_item` variables

YuriiMotov (commented):
@timakaa, thank you for your interest in FastAPI!

I agree it makes sense to change this.
I would only removed unnecessary `foo_item` and `bar_item` variables. @timakaa, what do you think?

I checked page that uses this code example - there is no highlights for it and text around it doesn't require updates

timakaa:
@YuriiMotov right, no need to overcomplicate things, I also changed that for other similar files

to make it clear my nvim lsp setup would highlight typing error right inside those files:
Pyright: Argument of type "Item" cannot be assigned to parameter "value" of type "dict[str, str]" in function "__setitem__"
"Item" is not assignable to "dict[str, str]" [reportArgumentType]

this one is for original docs_src/app_testing/app_b_an_py310/main.py file line 37, maybe my pyright is outdated or too strict, but I thought won't be harmful to fix that

YuriiMotov (approved):
LGTM!
@timakaa, thanks!

Alternatively we can just change the assignment in `create_item`:
```diff
-    fake_db[item.id] = item
+    fake_db[item.id] = item.model_dump()
```

Actually, now I think it would be even better solution (less changes and probably closer to intended implementation).
@timakaa, what do you think?

timakaa:
@YuriiMotov, Oh yeah, for sure, this is probably the best way to fix this issue

tiangolo (approved):
Thanks! 🍰

PR #14576, opened 2025-12-20
Fix endpoint context cache to avoid leaks and stale IDs

vitalvalid:
- Problem: __endpoint_context_cache used id(func)_ and never evicted. In long‑running or
    hot‑reload setups this can leak memory and, if an id is reused, report stale file/line/
    function info.
  - Fix: Use WeakKeyDictionary keyed by the callable itself, and skip caching for
    non‑weakrefable callables.
  - Test: Added a GC-based test to ensure cache entries are released after endpoints are
    garbage collected.

  Tests
  - bash scripts/test-cov-html.sh

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -20,6 +20,7 @@ from typing import (
     Optional,
     Union,
 )
+from weakref import WeakKeyDictionary
 
 from annotated_doc import Doc
 from fastapi import params, temp_pydantic_v1_params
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@vitalvalid, thanks!

Could you clarify what you mean by saying `or hot‑reload setups`?

I agree that this is a valid issue for use cases when FastAPI is used with dynamically created\deleted routes. Just not sure this is oficially supported way to use FastAPI.

tiangolo:
Thanks for the interest, I'm not sure which use case would need this,
hot‑reload setups would only be in local development, not in production, and in local development I wouldn't expect people to keep them running long-term.

Let's start with a GitHub Discussion with your specific use case and a minimal reproducible example of the problem.

For now, I'll pass on this one. Thanks! :coffee:

PR #14575, opened 2025-12-20
➖ Drop support for Pydantic v1, keeping short temporary support for Pydantic v2's `pydantic.v1`

tiangolo:
➖ Drop support for Pydantic v1, keeping short temporary support for Pydantic v2's `pydantic.v1`

PR #14582, opened 2025-12-21
Also dump v2 models in `_prepare_response_content`

johnslavik:
Corrects #14575 to also dump Pydantic v2 models in `_prepare_response_content`, fixing the regression.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -29,6 +29,7 @@ from fastapi._compat import (
     _model_dump,
     _normalize_errors,
     lenient_issubclass,
+    may_v1,
 )
 from fastapi.datastructures import Default, DefaultPlaceholder
 from fastapi.dependencies.models import Dependant
@@ -152,7 +153,7 @@ def _prepare_response_content(
     exclude_defaults: bool = False,
     exclude_none: bool = False,
 ) -> Any:
-    if isinstance(res, BaseModel):
+    if isinstance(res, (may_v1.BaseModel, BaseModel)):
         read_with_orm_mode = getattr(_get_model_config(res), "read_with_orm_mode", None)
         if read_with_orm_mode:
             # Let from_orm extract the data from this model instead of converting
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -29,7 +29,6 @@ from fastapi._compat import (
     _model_dump,
     _normalize_errors,
     lenient_issubclass,
-    may_v1,
 )
 from fastapi.datastructures import Default, DefaultPlaceholder
 from fastapi.dependencies.models import Dependant
@@ -58,6 +57,7 @@ from fastapi.utils import (
     get_value_or_default,
     is_body_allowed_for_status_code,
 )
+from pydantic import BaseModel
 from starlette import routing
 from starlette._exception_handler import wrap_app_handling_exceptions
 from starlette._utils import is_async_callable
@@ -152,7 +152,7 @@ def _prepare_response_content(
     exclude_defaults: bool = False,
     exclude_none: bool = False,
 ) -> Any:
-    if isinstance(res, may_v1.BaseModel):
+    if isinstance(res, BaseModel):
         read_with_orm_mode = getattr(_get_model_config(res), "read_with_orm_mode", None)  # type: ignore[arg-type]
         if read_with_orm_mode:
             # Let from_orm extract the data from this model instead of converting
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -154,7 +154,7 @@ def _prepare_response_content(
     exclude_none: bool = False,
 ) -> Any:
     if isinstance(res, (may_v1.BaseModel, BaseModel)):
-        read_with_orm_mode = getattr(_get_model_config(res), "read_with_orm_mode", None)
+        read_with_orm_mode = getattr(_get_model_config(res), "read_with_orm_mode", None)  # type: ignore[arg-type]
         if read_with_orm_mode:
             # Let from_orm extract the data from this model instead of converting
             # it now to a dict.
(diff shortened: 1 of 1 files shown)


johnslavik:
Hmm, I think we need to check for both classes. This proposes not dumping v1, which doesn't seem inline with the recent upgrade.

johnslavik:
For safe typing, it would be better to use unions of `v1.BaseModel` and `BaseModel` (v2) in all proper functions.
The type checker could then detect unsafe ops on two jointly. I'll shoot for `# type: ignore` from the old version for now.

johnslavik:
Looks like the coverage job deadlocked. Closing and reopening to re-run CI.

johnslavik:
LMK if we want some tests here.

potiuk:
Hello @tiangolo - question - do you have some initial assessment of the issue (even if not yet deciding to merge the PR) - 0.127.0 has also started to fail in Airflow and while it seems the root cause is some interaction between FastAPI and Cadwyn - we should soon take some direction with Airflow - we might also add an issue to Cadwyn if you think it should be solved there - just our reproducers clearly show that currently this case is broken

YuriiMotov:
I think if it only fails with Cadwyn, you should first open issue there.
Or, try to come up with the MRE without third-party dependencies

potiuk:
The MRE in #14575 shows that it only happens with Cadwyn and @johnslavik even proposed a PR fixing it - explaining why it happens (because of Cadwyn copying classes). I will definitely open issue with Cadwyn, but the investigation and fix suggests that it might happen with other tools as well, so my question is more on whether it is intended change or more of something that can (continue as it was before 0.126.0) be handled by Fastapi.

Note, that it's - as usual In those cases - hard to pinpoint who is 'guilty' - I..e. Whether the fact that it worked before was accidental, and whether the change was intentional, so likely if I open an issue in Cadwyn, they will also expect some assessment from FastApI point of view.

So I guess just looking at it and assessing by the maintainers might shorten the ping-pong

tiangolo:
Thanks all for the feedback!

## Rationale and current state

So, in Pydantic v2, with the default behavior in FastAPI (without Cadwyn), this code path would never be reached (if I'm understanding my own code correctly :sweat_smile: ), because a Pydantic v2 model would always have a `ModelField` with a `serialize()` method.

Before that PR it was passing everything through this serialization to account for installations of Pydantic v1.

But that also means that data is (slowly) double serialized in Python using this code. In Pydantic v2, the serialize method is used, and that uses the `TypeAdapter.dump_python` method, which serializes the data using the internal Pydantic core schema on the Rust side. So, things are much faster.

The state before this change was sub-optimal, for the sake of backwards compatibility, but now that I'm removing support for Pydantic v1, it doesn't make sense to do double serialization.

Yesterday I made another release raising a warning when users use Pydantic v1. This week I plan on fully dropping support for Pydantic v1 and raise an error if people use it.

That will mean a major internal cleanup, removing a lot of workarounds to support both v1 and v2 at the same time. This will improve performance, correctness, remove a whole range of existing/potential bugs happening from all that custom logic, and will enable a bunch of new features.

One of the important new features, which is partially now available, is precisely serializing on the Pydantic side when possible.

There are these refactors and some other internal improvements I need to make, some of the future ones will include router refactors that will allow other tools to add their own logic without needing to resort to touching FastAPI internals. But first I have to finish these things.

## This specific issue

Now, given all the above, soon I will remove most of this code, once I drop support for Pydantic v1, probably this week or the next.

I could still make a quick patch version now with this PR, for this specific use case, ideally with a test that fails with master but passes with this PR, to ensure it solves the problem. But it would be something temporary, just to have a release that works and you can use for now.

But I imagine some other things will change as well this week or the next once I fully drop support for Pydantic v1. :thinking:

johnslavik:
> ideally with a test that fails with master but passes with this PR, to ensure it solves the problem

Noted. Thank you very much for the quick response. I'll add the test shortly and ping you back.

potiuk:
Also I think if that's the case, then indeed this is something to be fixed on Cadwyn side. I do not think @tiangolo  - you  should even relese a new version for us - we we can stay with 0.125.0 until Cadwyn solves it.

potiuk:
The issue is already open https://github.com/zmievsa/cadwyn/issues/314

tiangolo:
Thanks @potiuk! It might then even make sense for Cadwyn to wait a few days (this or next week) until I fully remove Pydantic v1, as that will be a larger change, but will probably simplify things. :thinking: 

Also makes me think that maybe I should "hurry" to fully remove it instead of waiting more, as that will accelerate/simplify any fixes needed in other projects, instead of having a fix for this version and then having to change more in a few days... :sweat_smile:

potiuk:
> Thanks @potiuk! It might then even make sense for Cadwyn to wait a few days (this or next week) until I fully remove Pydantic v1, as that will be a larger change, but will probably simplify things. 🤔
> 
> Also makes me think that maybe I should "hurry" to fully remove it instead of waiting more, as that will accelerate/simplify any fixes needed in other projects, instead of having a fix for this version and then having to change more in a few days... 😅

Quite agree :)

potiuk:
This is BTW. precisely wanted to wait before I talk to Cadwyn, because it was not clear whether it was a delliberate or accidental change, and what direction it will take - and now we can have a very clear message - which I just did: https://github.com/zmievsa/cadwyn/issues/314#issuecomment-3683455138

potiuk:
And limiting airflow for now so that you can release  new fastapi versions as you wish https://github.com/apache/airflow/pull/59726

tiangolo:
Thank you for the help and coordination @potiuk! :raised_hands:

potiuk:
The least we can do as a way to support Fast APi

johnslavik:
To wrap this up:

> I do not think @tiangolo - you should even relese a new version for us - we we can stay with 0.125.0 until Cadwyn solves it.

I understand that there's no need to go forward with this PR.
I'll see if I can help fixing this in Cadwyn.

If you'd like me to reopen, please let me know.

Thank y'all as well!

potiuk:
Thanks for your help here as well @johnslavik :)

PR #14594, opened 2025-12-23
Minor clarification suggestion for the “Parallel Burgers” analogy in …

sankarebarri:
…concurrency docs

Hi 👋
The burgers analogy is one of the clearest explanations of async/concurrency I’ve seen.

I wanted to raise a small pedagogical point regarding the “Parallel Burgers” section.

While the analogy does highlight why parallelism is not ideal for I/O-bound workloads, it may unintentionally give readers the impression that parallelism itself is inefficient or inherently worse than concurrency, rather than simply better suited for CPU-bound work.

In particular, the example emphasises waiting and idle time, whereas true parallelism is most beneficial when multiple workers are actively doing computation simultaneously.

A small clarification (or an alternative example where multiple cooks actively work on different parts of the task at the same time) might help reinforce that distinction.

This is a minor point, but I thought it could help avoid confusion for readers learning these concepts for the first time.

Thanks again for the great docs!

### DIFF
--- a/docs/en/docs/async.md
+++ b/docs/en/docs/async.md
@@ -219,6 +219,13 @@ The fast food store has 8 processors (cashiers/cooks). While the concurrent burg
 
 But still, the final experience is not the best. 😞
 
+/// note
+
+This example focuses on an <abbr title="Input and Output">I/O</abbr>-bound scenario with significant waiting. Parallelism shines instead for <abbr title="Central Processing Unit">CPU</abbr>-bound work, where multiple workers actively perform computation at the same time.
+
+///
+
+
 ---
 
 This would be the parallel equivalent story for burgers. 🍔


docs/en/docs/async.md
```diff
@@ -219,6 +219,13 @@ The fast food store has 8 processors (cashiers/cooks). While the concurrent burg
 
 But still, the final experience is not the best. 😞
 
+/// note
+
+This example focuses on an <abbr title="Input and Output">I/O</abbr>-bound scenario with significant waiting. Parallelism shines instead for <abbr title="Central Processing Unit">CPU</abbr>-bound work, where multiple workers actively perform computation at the same time.
+
+///
+
+
```
MaxwellEnemuo, line 228: I came across this documentation independently and had the exact same concern about the terminology. The current "Parallel Burgers" section describes synchronous blocking with multiple workers, not true parallelism.
From a CS perspective:

Concurrency = managing multiple tasks (can switch between them)
Parallelism = executing multiple tasks simultaneously
Actual production FastAPI = parallel workers (processes) each running concurrent async code

The "Parallel Burgers" scenario would be more accurately labeled as "Synchronous/Blocking Burgers" since each worker handles one customer at a time in a blocking manner.

Your proposed note is a great addition! It correctly clarifies that true parallelism is for CPU-bound work. One small suggestion: it might also be worth noting that in production web deployments, you typically get both - multiple parallel workers (processes) where each worker handles I/O-bound requests concurrently (async). 

The practical advice in the docs is excellent and correct, this is purely about technical precision in terminology. Thanks for working on this clarification! 👍

YuriiMotov (commented):
@sankarebarri, thanks for your interest!

I think we don't need this clarification. This is explained well in the same docs section below (see [Is concurrency better than parallelism?](https://fastapi.tiangolo.com/async/#is-concurrency-better-than-parallelism)).

It will never be ideal - there will always be people who say it's still not clear enough.
We should respect Sebastian's time and don't overload him by unnecessary changes.

If you still think we should improve this - please, open discussion and suggest this improvement idea. If there are a lot of people who support this - we will ask Sebastian to review this.

Thank you for understanding!

tiangolo:
Thanks for the interest, but yep, this is explained below in the same document. I'll pass on this for now. ☕

PR #14595, opened 2025-12-23
Fix: Require realm in HTTPBasic to comply with RFC 7617

pallaprolus:
### Description

This PR fixes issue #13471 where `HTTPBasic` authentication allowed instantiation without a `realm` parameter. According to [RFC 7617 Section 2](https://datatracker.ietf.org/doc/html/rfc7617#section-2), the `realm` parameter is **REQUIRED** for Basic authentication.

This change:
1.  Modifies `HTTPBasic.__init__` to make `realm` a required argument (removing the default `None`).
2.  Updates `make_authenticate_headers` to always include the `realm` in the `WWW-Authenticate` header.

### Breaking Change :warning:

This is a **BREAKING CHANGE**.
Existing code dealing with security like:
```python
security = HTTPBasic()
```
will now raise a `TypeError`. Users must update their code to provide a realm:
```python
security = HTTPBasic(realm='my-realm')
```

### Verification

- Created a reproduction script (`reproduce_issue_13471.py`, not included in PR) confirming failure before fix and success after.
- Updated existing tutorials (`docs_src/security/tutorial00[67]*.py`) to use `realm='simple'`.
- Updated existing tests (`tests/test_tutorial/test_security/test_tutorial006.py`, `tests/test_security_http_basic_optional.py`) to match the new behavior.
- Ran specific security tests (`tests/test_security_http_basic_realm.py`, `tests/test_security_http_basic_optional.py`) and tutorials tests.
- Ran `ruff format` and `ruff check`.

### DIFF
--- a/docs_src/security/tutorial006_an_py39.py
+++ b/docs_src/security/tutorial006_an_py39.py
@@ -5,7 +5,7 @@ from fastapi.security import HTTPBasic, HTTPBasicCredentials
 
 app = FastAPI()
 
-security = HTTPBasic()
+security = HTTPBasic(realm="simple")
 
 
 @app.get("/users/me")
--- a/docs_src/security/tutorial006_py39.py
+++ b/docs_src/security/tutorial006_py39.py
@@ -3,7 +3,7 @@ from fastapi.security import HTTPBasic, HTTPBasicCredentials
 
 app = FastAPI()
 
-security = HTTPBasic()
+security = HTTPBasic(realm="simple")
 
 
 @app.get("/users/me")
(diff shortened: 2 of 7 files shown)


YuriiMotov:
Could you please take a look at [this comment](https://github.com/fastapi/fastapi/issues/13471#issuecomment-3596388412). Do you still think `realm` is required?

pallaprolus:
Thank you @YuriiMotov for the feedback and for pointing out the distinction between RFC 7617 and RFC 7235.

I missed that RFC 7235 Appendix A explicitly relaxes the requirement for the realm parameter, allowing challenges without it. Given that enforcing this would introduce a breaking change for existing users and isn't strictly required by the modern standard, I agree that this change isn't the right path forward.

I'm closing this PR now. Thanks again for the review and the clarification!

PR #14598, opened 2025-12-24
📝 Fix grammar in `docs/en/docs/tutorial/body.md`

BLESSEDEFEM:
## Description
Fixed grammar issue in the tutorial documentation.

## Changes
- Changed "and that function parameters that are declared" to "and those function parameters that are declared" for better grammar and readability.

## Location
`docs/en/docs/tutorial/body.md`
Section: "Request body + path parameters"

### DIFF
--- a/docs/en/docs/tutorial/body.md
+++ b/docs/en/docs/tutorial/body.md
@@ -132,7 +132,7 @@ Inside of the function, you can access all the attributes of the model object di
 
 You can declare path parameters and request body at the same time.
 
-**FastAPI** will recognize that the function parameters that match path parameters should be **taken from the path**, and that function parameters that are declared to be Pydantic models should be **taken from the request body**.
+**FastAPI** will recognize that the function parameters that match path parameters should be **taken from the path**, and those function parameters that are declared to be Pydantic models should be **taken from the request body**.
 
 {* ../../docs_src/body/tutorial003_py310.py hl[15:16] *}
 


alejsdev (approved):

docs/en/docs/tutorial/body.md
```diff
@@ -132,7 +132,7 @@ Inside of the function, you can access all the attributes of the model object di
 
 You can declare path parameters and request body at the same time.
 
-**FastAPI** will recognize that the function parameters that match path parameters should be **taken from the path**, and that function parameters that are declared to be Pydantic models should be **taken from the request body**.
+**FastAPI** will recognize that the function parameters that match path parameters should be **taken from the path**, and those function parameters that are declared to be Pydantic models should be **taken from the request body**.
```
alejsdev, line 135: Sorry, I approved this by mistake 😅. This is actually written as intended.

The word "that" here isn’t pointing to the parameters like "those" would, but instead, is connecting two ideas, basically saying:

**FastAPI** will recognize that the function parameters that match path parameters should be **taken from the path**, and _will recognize_ that function parameters that are declared to be Pydantic models should be **taken from the request body**

So "that" is just linking what FastAPI will recognize in each case.

Hope this clarify this! I'll close this PR but thank you for the effort.

BLESSEDEFEM:
Thank you for the explanation, @alejsdev ! I appreciate you taking the time to clarify the grammar. I learned something new today about conjunctions vs. demonstrative pronouns. 
I'll look for other ways to contribute to FastAPI. Thanks for maintaining this awesome project!

    On Friday, December 26, 2025 at 10:41:27 AM GMT+1, Alejandra ***@***.***> wrote:  
 
 


@alejsdev commented on this pull request.

In docs/en/docs/tutorial/body.md:
> @@ -132,7 +132,7 @@ Inside of the function, you can access all the attributes of the model object di
 
 You can declare path parameters and request body at the same time.
 
-**FastAPI** will recognize that the function parameters that match path parameters should be **taken from the path**, and that function parameters that are declared to be Pydantic models should be **taken from the request body**.
+**FastAPI** will recognize that the function parameters that match path parameters should be **taken from the path**, and those function parameters that are declared to be Pydantic models should be **taken from the request body**.

Sorry, I approved this by mistake 😅. This is actually written as intended.

The word "that" here isn’t pointing to the parameters like "those" would, but instead, is connecting two ideas, basically saying:

FastAPI will recognize that the function parameters that match path parameters should be taken from the path, and will recognize that function parameters that are declared to be Pydantic models should be taken from the request body

So "that" is just linking what FastAPI will recognize in each case.

Hope this clarify this! I'll close this PR but thank you for the effort.

Issue #3589, opened 2021-07-23
Raising exceptions with registered exception handlers from BackgroundTasks causes starlette exception


alonme:
### First check

* [V ] I added a very descriptive title to this issue.
* [ V] I used the GitHub search to find a similar issue and didn't find it.
* [ V] I searched the FastAPI documentation, with the integrated search.
* [ V] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [ V] I already read and followed all the tutorial in the docs and didn't find an answer.
* [ V] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [ V] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [ V] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Example

```Python
from fastapi import BackgroundTasks, FastAPI, status
from starlette.requests import Request
from starlette.responses import JSONResponse


class MyException(Exception):
    ...

app = FastAPI()


@app.exception_handler(MyException)
async def sampler_exception_handler(request: Request, exc: MyException):
    return JSONResponse(
        status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
        content={"error": "i hope someone sees this"},
    )


def write_notification(email: str, message=""):
    raise MyException("this is my exception")
    with open("log.txt", mode="w") as email_file:
        content = f"notification for {email}: {message}"
        email_file.write(content)


@app.post("/send-notification/{email}")
async def send_notification(email: str, background_tasks: BackgroundTasks):
    background_tasks.add_task(write_notification, email, message="some notification")
    return {"message": "Notification sent in the background"}

```

### Description

Raising an exception that an exception_handler is registered for, from a background task - causes an error from starlette, which also hides the original exception.

running the code from the example, and triggering the endpoint,  i expected the error to be either handled by the exception handler, or maybe just ignoring the exception handler as this exception does not happen as part of a request context.

the actual behavior is as follows

```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/Users/alonmenczer/.pyenv/versions/3.7.9/lib/python3.7/site-packages/uvicorn/protocols/http/h11_impl.py", line 369, in run_asgi
    result = await app(self.scope, self.receive, self.send)
[...]

Issue #2604, opened 2021-01-05
BackgroundTasks do not run when request failed


NewLanded:
Hello everybody,
  I have a question about background task, there is an example

This is running fine.  
```python
@app.post("/")
def user(background_tasks: BackgroundTasks):
    background_tasks.add_task(func_example)
```

But when I raise an exception, the background task will not be running.
```python
@app.post("/")
def user(background_tasks: BackgroundTasks):
    background_tasks.add_task(func_example)
    raise HTTPException(status_code=500, detail="example error")
```

I want to know this is a bug or a feature?
Thanks for any answer.

krishnardt:
Background tasks will run only after the response(after the response is successful).

Once the request is received, the task related to the request will be added to the background task.
Once the response is successful, the task added to Background will be executed.
if the repsonse is a failure due to any issue(whether it can be input validation issue, internal server error, raising exceptions), task will not be executed.

to keep it simple, that is a feature to avoid running the task, if the api failures in any case.

Kludex:
It is a feature.

| _You can define background tasks to be run after **returning a response**._
"Returning" here is the literal `return` in the endpoint function. If you never reach it, it will not run the task.

Reference: https://fastapi.tiangolo.com/tutorial/background-tasks/#background-tasks

Kludex:
I think we should be more explicit in the docs about this. :disappointed:

iamthen0ise:
I think it is better to think about BackgroundTask as about a defer function, which run just after the successful call of parent one.

It is not even close to any existing Task Queue implementation. Think of Celery, Rq, or even self-made queues if you need to have all benefits of it — retries, backoff, lazy loading, etc. 

Otherwise, think about BackgroundTask as a some way to run a successor function just make the things done, right after the request is succeed. E.g. update cache, alter user session attributes, and soon.

amoghmishra-sl:
Is there a workaround this? I'd want to generate the error logs as a background task when there is a failure. Given the current feature, they aren't getting generated. I am trying to do something like this:
```
import time
import traceback
from fastapi import FastAPI, Request
from fastapi.responses import JSONResponse
from starlette.background import BackgroundTask

app = FastAPI()

@app.middleware("http")
async def middleware(request: Request, call_next):
    start_time = time.time()
    try:
        response = await call_next(request)
        process_time = time.time() - start_time
        response.headers["X-Process-Time"] = str(process_time)
    except Exception as exc:
[...]

Issue #2505, opened 2020-12-11
Exceptions with handlers are replaced in a background task


snazzyfox:
When a background task raises an exception that has a FastAPI handler, that exception is replaced by FastAPI and the original error cannot be found in the server logs. This makes debugging and investigations incredibly hard as it's impossible to know where the error occurred.

FastAPI should preserve the original exception so that it can be logged and investigated server-side.

### Example

```Python
from fastapi import FastAPI, Request, BackgroundTasks
from fastapi.responses import JSONResponse


class UnicornException(Exception):
    def __init__(self, name: str):
        self.name = name


app = FastAPI()


@app.exception_handler(UnicornException)
async def unicorn_exception_handler(request: Request, exc: UnicornException):
    return JSONResponse(
        status_code=418,
        content={"message": f"Oops! {exc.name} did something. There goes a rainbow..."},
    )


async def process_unicorn(name: str):
    if name == "yolo":
        raise UnicornException(name=name)
    # else do_something(name)

@app.get("/unicorns/{name}")
async def sync_process_unicorn(name: str, background_tasks: BackgroundTasks):
    await process_unicorn(name)
    return name

@app.get("/unicorns_async/{name}")
async def async_process_unicorn(name: str, background_tasks: BackgroundTasks):
    """Accept the unicorn and process it in the background"""
    background_tasks.add_task(process_unicorn, name=name)
    return f"Processing {name} asynchronously."
```

### Description

- `/unicorns/yolo` correctly returns an error response
- `/unicorns_async/yolo` returns the correct response to the client. I expect error logs and traceback of a `UnicornException` on the server. However, that exception is *replaced* with one that FastAPI generates, without context. 

```
INFO:     127.0.0.1:32848 - "GET /unicorns_async/yolo HTTP/1.1" 200 OK
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/luyao/.local/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 394, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/home/luyao/.local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/home/luyao/.local/lib/python3.8/site-packages/fastapi/applications.py", line 190, in __call__
    await super().__call__(scope, receive, send)
[...]

PR #14599, opened 2025-12-24
📝 Add exception handling guidance for BackgroundTasks

tsayin4:
## Summary

Adds exception handling guidance to the BackgroundTasks tutorial to help users avoid silent failures in production.

## Context

Several users have reported confusion around exception handling in BackgroundTasks:

- #2505 – Exceptions in background tasks can be hard to debug
- #2604 – BackgroundTasks do not run when the request fails
- #3589 – Exceptions in background tasks can obscure original errors
- Discussion #7532 – Users asking for clearer documentation
- Discussion #8987 – Confusion around background task execution

Maintainers have previously noted that this behavior should be more explicit in the documentation.

## Motivation

While building production systems, I explored async exception visibility patterns and noticed the same confusion affects FastAPI's BackgroundTasks. Exceptions in background tasks don't propagate to the response handler, which can lead to silent failures that are difficult to debug.

## Changes

Adds a focused section to the BackgroundTasks tutorial explaining:
- How exceptions behave in background tasks (not propagated to the request handler)
- An example showing a silent failure scenario
- An example showing proper error handling with logging
- A pointer to task queue libraries for more complex production use cases

## Scope

- Single documentation section (~50 lines)
- No API changes
- Located immediately after the basic BackgroundTasks examples
- Minimal and focused on production visibility

Happy to adjust wording or move this to a more appropriate section if you think that fits better.

### DIFF
--- a/docs/en/docs/tutorial/background-tasks.md
+++ b/docs/en/docs/tutorial/background-tasks.md
@@ -45,6 +45,60 @@ Inside of your *path operation function*, pass your task function to the *backgr
 * Any sequence of arguments that should be passed to the task function in order (`email`).
 * Any keyword arguments that should be passed to the task function (`message="some notification"`).
 
+## Important: Exception Handling
+
+Background tasks run **after** the response is sent. If an exception occurs in a background task, it won't affect the response (which already succeeded), and **the exception will not be propagated to the request handler and may go unnoticed unless explicitly handled (e.g. logging, retries, alerts)**.
+
+This is expected behavior, but it is easy to overlook when building production systems.
+
+### Silent Failure Example
+```python
+def send_email(email: str):
+    # If this fails, the exception goes unnoticed
+    smtp_client.send(to=email, subject="Hello")
+    # User gets 200 OK, but email may never be sent!
+
+@app.post("/notify/")
+async def notify(background_tasks: BackgroundTasks, email: str):
+    background_tasks.add_task(send_email, email)
+    return {"message": "Notification sent"}  
+    # Returns success even if email fails!
+```
+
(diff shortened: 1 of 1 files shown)


docs/en/docs/tutorial/background-tasks.md
```diff
@@ -45,6 +45,60 @@ Inside of your *path operation function*, pass your task function to the *backgr
...
+- Notify operators
+- Persist failures for later inspection
+
+!!! tip
+    For production systems with complex background jobs, consider:
+
+    - [Celery](https://docs.celeryq.dev) - Distributed task queue
+    - [arq](https://arq-docs.helpmanual.io/) - Async task queue for Python
+    - [Dramatiq](https://dramatiq.io/) - Fast and reliable task processing
+
```
YuriiMotov, line 101:
```suggestion
/// tip
For production systems with complex background jobs, consider:

- [Celery](https://docs.celeryq.dev) - Distributed task queue
- [arq](https://arq-docs.helpmanual.io/) - Async task queue for Python
- [Dramatiq](https://dramatiq.io/) - Fast and reliable task processing

///

```

YuriiMotov (changes requested):
@tsayin4, thank you for your interest!

This is a good idea to add such clarification and code example, but current implementation should be improved:

* Code examples should be placed in separate `.py`-files and covered by tests
* I think code examples should be updated:
  * catch specific exceptions instead of `except Exception` (that's considered as bad practice)
  * code example should be more meaningful - just catching the exception and logging it doesn't really change a lot (error traceback would be printed even without this logic)
* I would also add console output for each example and some clarifications

PR #14606, opened 2025-12-26
docs: Basic auth `realm` is REQUIRED but handled as optional

### LINKED ISSUE #13471: Basic auth `realm` is REQUIRED but handled as optional

Ricky512227:
## Description

https://github.com/fastapi/fastapi/blob/643d2

## Changes Made

- docs: Added contributing section

Fixes #13471


YuriiMotov:
Hello, @Ricky512227!

PR #14609, opened 2025-12-26
➖ Drop support for `pydantic.v1`

tiangolo:
➖ Drop support for `pydantic.v1`

When trying to use a `pydantic.v1` model with FastAPI, an error is raised.


potiuk:
Cadwyn released 6.0.0 + Airflow switched to 0.128.0 with it -> and all works for us. 

Fingers crossed for faster and more modern Fast API!

PR #14610, opened 2025-12-27
docs: Add streaming inference example using StreamingResponse

AnirudhArrepu:
## What does this PR do?

Adds a documentation example demonstrating how to use `StreamingResponse`
for long-running inference workloads, such as machine learning or
token-based generation, where returning partial results improves latency
and user experience.

The example shows a simple generator-based pattern that can be adapted
to real-world ML or LLM inference pipelines.

## Why is this useful?

FastAPI is commonly used to serve ML and LLM-backed APIs, but there is
currently no dedicated documentation example focused on streaming
inference results incrementally.

This PR provides a minimal, dependency-free example that users can
copy and adapt for production inference use cases.

## Scope of the change

- Documentation/example only
- No changes to FastAPI core behavior
- No new dependencies introduced

## Testing

- Ran the example locally using Uvicorn
- Verified incremental streaming behavior via browser request

### DIFF
--- a/docs_src/streaming/streaming_inference.py
+++ b/docs_src/streaming/streaming_inference.py
@@ -7,7 +7,7 @@ latency and user experience.
 """
 
 import time
-from typing import Generator
+from collections.abc import Generator
 
 from fastapi import FastAPI
 from fastapi.responses import StreamingResponse
--- a/docs_src/streaming/streaming_inference.py
+++ b/docs_src/streaming/streaming_inference.py
@@ -1,0 +1,47 @@
+"""
+Streaming inference example using StreamingResponse.
+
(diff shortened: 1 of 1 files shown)


YuriiMotov (commented):
@AnirudhArrepu, thanks for your interest!

We already have a code example demonstrating `StreaminResponse`: https://fastapi.tiangolo.com/advanced/custom-response/#streamingresponse

I think we don't need to add this new example.

Also, just adding a python-file is not enough - without including it to a docs page it will unlikely to be discovered by anybody..
And, all code examples should be covered by tests.

Having all this, I will close this PR.
Thanks for your efforts!

PR #14615, opened 2025-12-27
🏷️ Add typing for `dependency_overrides_provider` in `solve_dependencies`

bokshitsky:
Just add type for DependencyOverridesProvider and remove `getattr`:

```
getattr(
    dependency_overrides_provider, "dependency_overrides", {}
)
```


YuriiMotov (approved):
LGTM

This might be breaking for code that uses `solve_dependencies` passing object without `dependency_overrides` attribute.
Previously it would return `original_call`. With changes from this PR it will raise `AttributeError`.
But I'm not sure it's realistic case. And it's easy to find and fix this.

@bokshitsky, thanks!

tiangolo:
I'm not fully sure if there would be other interactions that could end up passing an object that doesn't have the attribute, so for now, I prefer to play it safe and keep it as it is.

I also intend to refactor how dependency overrides work, but that will probably be after some other changes. Thanks! :coffee:

PR #14616, opened 2025-12-28
🐛 Fix using `Json[list[str]]` type (issue #10997)

### LINKED ISSUE #10997: Using pydantic Json Type as Form data type doesn't work
### Discussed in https://github.com/tiangolo/fastapi/discussions/9305

<sup>Originally posted by **harpaj** March 23, 2023</sup>
### First Check

### Commit to Help

### Example Code

```python
from typing import Annotated

from pydantic import Json, BaseModel

app = FastAPI()


class JsonListModel(BaseModel):
    json_list: Json[list[str]]


@app.post("/working")
async def working(json_list: Annotated[str, Form()]) -> list[str]:
    model = JsonListModel(json_list=json_list)
    return model.json_list


@app.post("/broken")
async def broken(json_list: Annotated[Json[list[str]], Form()] ) -> list[str]:
    return json_list
```

### Description

In the example code above, I would expect `working` and `broken` to be approximately equivalent.
However, while `working` returns the parsed json_list as expected, `broken` fails with 

```json
{"detail":[{"loc":["body","json_list"],"msg":"JSON object must be str, bytes or bytearray","type":"type_error.json"}]}
```

### Operating System

Linux

### Operating System Details

_No response_

### FastAPI Version

0.95.0

### Python Version

Python 3.10.8

### Additional Context

sample request that fails with `broken` and works with `working`:
```python

import requests
import json

headers = {"Content-Type": "application/x-www-form-urlencoded"}
response = requests.post(
    "http://0.0.0.0:8000/broken",
    data={"json_list": json.dumps(["abc", "def"])},
    headers=headers,
)
```</div>

### PR DESCRIPTION
**Pull request by mkanetsuna**
Fixes Form() not working with Pydantic `Json[T]` types when T is a sequence type (list, set, tuple).

Closes #10997

When using `Form()` with `Json[list[str]]`, FastAPI incorrectly treated it as a sequence field and used `getlist()`, wrapping the JSON string in a list. This caused Pydantic to fail parsing with:
```
TypeError: JSON input should be string, bytes or bytearray
```

**Before fix:**
```python
# Client sends
data = {"items": '["abc", "def"]'}  # JSON string

# FastAPI incorrectly does
value = values.getlist("items")  # ['["abc", "def"]']  ❌
```

**After fix:**
```python
# FastAPI correctly does
value = values.get("items")  # '["abc", "def"]'  ✅
```

This fix enables legitimate use cases like file uploads with JSON metadata:

```python
@app.post("/upload")
def upload(
    file: UploadFile,
    tags: Annotated[Json[list[str]], Form()]
):
    return {"file": file.filename, "tags": tags}
```

Pydantic v1 support for `Json[T]` with `Form()` is not included in this PR because `metadata` attribute does not exist in Pydantic v1's `FieldInfo`.

## Related
- Issue #10997


fastapi/dependencies/utils.py
```diff
@@ -716,11 +716,22 @@ def _validate_value_with_model_field(
         return v_, []
 
 
+def _is_json_field(field: ModelField) -> bool:
+    # Pydantic v1 FieldInfo doesn't have metadata attribute
+    metadata = getattr(field.field_info, "metadata", [])
+    return any(
+        type(item).__name__ == "Json" and type(item).__module__ == "pydantic.types"
+        for item in metadata
+    )
```
```suggestion
    return any(isinstance(item, Json) for item in field.field_info.metadata)
```

FastAPI dropped the support of Pydantic V1 recently.
We can now simplify this.
Also, we should use `isinstance` to check type

fastapi/dependencies/utils.py
```diff
@@ -716,11 +716,22 @@ def _validate_value_with_model_field(
...
+
+
 def _get_multidict_value(
     field: ModelField, values: Mapping[str, Any], alias: Union[str, None] = None
 ) -> Any:
     alias = alias or get_validation_alias(field)
-    if is_sequence_field(field) and isinstance(values, (ImmutableMultiDict, Headers)):
+    if _is_json_field(field):
+        value = values.get(alias, None)
+    elif is_sequence_field(field) and isinstance(values, (ImmutableMultiDict, Headers)):
```
```suggestion
    if (
        (not _is_json_field(field))
        and is_sequence_field(field)
        and isinstance(values, (ImmutableMultiDict, Headers))
    ):
```

tests/test_form_json_type.py
```diff
@@ -0,0 +1,44 @@
...
+client = TestClient(app)
+
+
+def test_form_str():
+    response = client.post(
+        "/form-str",
+        data={"json_list": json.dumps(["abc", "def"])},
+    )
+    assert response.status_code == 200, response.text
+    assert response.json() == ["abc", "def"]
```
YuriiMotov:
```suggestion
```
I think we don't need this test

tests/test_form_json_type.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class JsonListModel(BaseModel):
+    json_list: Json[list[str]]
+
+
+@app.post("/form-str")
+def form_str(json_list: Annotated[str, Form()]) -> list[str]:
+    model = JsonListModel(json_list=json_list)  # type: ignore[arg-type]
+    return model.json_list
```
YuriiMotov:
```suggestion
```
I think we don't need this

YuriiMotov (changes requested):
@mkanetsuna, thanks for your interest!

The idea looks good to me.
Please, take a look at my suggestions.

---

**UPD:** I think this will fix `Json[list[str]]` not only for `Form` parameters, but for other types of parameters (`Query`, `Header`, etc..) as well.
@mkanetsuna, could you please check it and add tests for other types of parameters?

mkanetsuna:
@YuriiMotov Thanks for the suggestion! I've added tests for `Query()`, `Header()`, and `Cookie()` parameters.
You were right - this PR's fix works for all parameter types, not just Form(). 
All tests are passing now, so please review again!

tests/test_form_json_type.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@app.get("/header-json-list")
+def header_json_list(x_items: Annotated[Json[list[str]], Header()]) -> list[str]:
+    return x_items
+
+
+@app.get("/cookie-json-dict")
+def cookie_json_dict(
+    session: Annotated[Json[dict[str, str]], Cookie()],
```
YuriiMotov: Could you please explain why you used `dict` here, not `list`?
With `dict` this test passes on `master` branch.

mkanetsuna: You're right! After investigating, I found that Cookie was never affected by the original bug, they skip `getlist()` path.
I've removed the Cookie test. The three remaining tests (Form/Query/Header) properly cover the bug fix.

YuriiMotov: I mean for Query you use `Json[list[str]]`, but for Cookie `Json[dict[str, str]]` (`list` vs `dict`).

What if you declare cookie endpoint as:

```py
@app.get("/cookie-json-list")
def cookie_json_list(
    items: Annotated[Json[list[str]], Cookie()],
) -> list[str]:
    return items
```
?

I think this should also fail on `master` branch and be fixed by this PR

mkanetsuna: I thought the same, but when I tested on master, it passed 😶
```bash
==================================================================== test session starts ====================================================================
...                                                                                                                    
tests/test_form_json_type.py::test_form_json_list FAILED                                                                                              [ 25%]
tests/test_form_json_type.py::test_query_json_list FAILED                                                                                             [ 50%]
tests/test_form_json_type.py::test_header_json_list FAILED                                                                                            [ 75%]
tests/test_form_json_type.py::test_cookie_json_list PASSED                                                                                            [100%]
```
Looks like cookies don't go through the getlist() path.
cookie values are plain dict, not ImmutableMultiDict or Headers, so they skip that.

mkanetsuna: u can check by using this
```python
import json
from typing import Annotated

from fastapi import Cookie, FastAPI
from pydantic import Json

app = FastAPI()

@app.get("/cookie-json-list")
def cookie_json_list(
    session: Annotated[Json[list[str]], Cookie()],
) -> list[str]:
    return session


client = TestClient(app)


def test_cookie_json_list():
    client.cookies.set("session", json.dumps(["user", "test1", "role", "admin"]))
    response = client.get("/cookie-json-list")
    assert response.json() == ["user", "test1", "role", "admin"]
```

YuriiMotov: Interesting... Thanks for clarifying this!

I would still add this test with `Annotated[Json[list[str]], Cookie()]` for consistency.
And, we need to update the name of test module to reflect that it's not only for `Form`

mkanetsuna: Done! Added Cookie test consistent with other types and renamed the file to `test_json_type.py`.

YuriiMotov: Great! Thank you!

YuriiMotov (approved):
LGTM!

Currently (on `master` branch) tests for `Query`, `Form` and `Header` fail. Test for `Cookie` passes, but I think we should keep it for consistency.

@mkanetsuna, thank you for quick updates!

tiangolo (approved):
Great job! Thank you! 🙌 

I was convinced of not supporting this (reading the issue), but this implementation is simple and clean enough that I think it makes sense to have it. Thanks! 🚀 

This will be available in FastAPI 0.128.2, released in the next few hours. 🎉

YuriiMotov:
We missed one thing - we didn't check the openapi schema. And it's seems to be not good..

With the code example from the description of this PR, the related part of schema is:

```json
"tags": {
  "type": "string",
  "contentMediaType": "application/json",
  "contentSchema": {
      "items": {
      "type": "string"
      },
      "type": "array"
  },
  "title": "Tags"
}
```

And it's rendered by Swagger UI as file upload:

---

image

---

And, this is how Pydantic represents it.. So, not sure we can easily fix this..

The only way I can see is to use `WithJsonSchema`:

```py
from typing import Annotated

from fastapi.datastructures import UploadFile
from pydantic import Json, WithJsonSchema

app = FastAPI()


@app.post("/upload")
def upload(
    file: UploadFile,
    tags: Annotated[Json[list[str]], Form(), WithJsonSchema({"type": "string"})],
):
    return {"file": file.filename, "tags": tags}
```

Now, Swagger UI renders it as just string input:

---

image

---

But if you attempt to send data as `["a", "b"]`, it will fail. Swagger UI automatically pre-processes strings that look like arrays and it converts our input to just `a,b`:

---

image

---

To finally pass the value, you need to put it inside the double quotes and escape the quotes inside: `"[\"a\", \"b\"]"`

---

image

PR #14619, opened 2025-12-29
🌐 Update translation prompts

tiangolo:
🌐 Update translation prompts


scripts/translate.py
```diff
@@ -25,650 +25,8 @@
...
-
-5) Apply above rules also when there is an existing translation! Make sure that all title attributes in abbr elements get properly translated or updated, using the schemes given above. However, leave the ADDITIONAL abbr's from rule 4 alone. Do not change their formatting or content.
-
-"""
+general_prompt_path = Path(__file__).absolute().parent / "llm-general-prompt.md"
```
DoctorJohn, line 28: There appears to be a typo. The filename should be `general-llm-prompt.md`.

nilslindemann:
Good that you removed all the `«...»` stuff from the prompts. Amongst other things I used this to try to explain to the LLM to not convert `"` inside Backticks to typographic quotes. But this never really worked (it did for a while, which is why I kept it, but a few weeks later it stopped) and it makes the LLM prompt difficult to read.

PR #14620, opened 2025-12-30
refactor - replaced deprecated on_event with lifespan in SQL tutorials

Ads2024:
This pull request updates the SQL database tutorial example `tutorial001` to use the FastAPI `lifespan` parameter for application initialization, replacing the deprecated `@app.on_event("startup")` decorator. The tutorial now reflects the current and recommended lifecycle pattern in FastAPI instead of relying on behavior that is already deprecated.

The accompanying test update removes a TODO and the `warnings.catch_warnings`  block was adjusted by `removing record = True` argument in tests/test_tutorial/`test_sql_databases/test_tutorial001.py`  that was previously suppressing the deprecation warning. The underlying cause is fixed by migrating to lifespan events.

As a result, the test suite runs without emitting `DeprecationWarning` for `tutorial001`.

### DIFF
--- a/docs_src/sql_databases/tutorial001_an_py310.py
+++ b/docs_src/sql_databases/tutorial001_an_py310.py
@@ -1,5 +1,6 @@
-from typing import Annotated
 from contextlib import asynccontextmanager
+from typing import Annotated
+
 from fastapi import Depends, FastAPI, HTTPException, Query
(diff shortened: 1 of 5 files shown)


YuriiMotov:
@Ads2024, thanks for your interest, but I think we already have https://github.com/fastapi/fastapi/pull/14332 (and 2 more similar PRs)

PR #14621, opened 2025-12-30
Improve error message for missing dependency type annotation

uydev:
This PR improves the error raised when a dependency parameter is missing a type annotation.

Previously, FastAPI would raise a generic AssertionError during dependency resolution, which made it difficult to understand what went wrong and how to fix it.

Changes in this PR:
- Replace the assertion with a FastAPIError that includes:
  - the dependency parameter name
  - a clear explanation of the issue
  - an example of the correct annotation syntax
- Ensure the error is only enforced in appropriate contexts to avoid breaking
  parameter-less dependencies, dependency overrides, and WebSocket sub-dependencies
- Add a test covering the missing type annotation case

All existing tests pass.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -109,9 +109,7 @@ def ensure_multipart_is_installed() -> None:
             raise RuntimeError(multipart_not_installed_error) from None
 
 
-def get_parameterless_sub_dependant(
-    *, depends: params.Depends, path: str
-) -> Dependant:
+def get_parameterless_sub_dependant(*, depends: params.Depends, path: str) -> Dependant:
     assert callable(depends.dependency), (
         "A parameter-less dependency must have a callable dependency"
     )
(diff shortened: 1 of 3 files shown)


uydev:
Hi!
It looks like the label check is failing because the PR doesn't have a label yet.
I don't have permission to add labels myself. Could a maintainer please add the appropriate label please?
Thanks!

YuriiMotov:
> Previously, FastAPI would raise a generic AssertionError during dependency resolution

Could you please share a code example that we can use to reproduce this?

uydev:
Here’s the snippet I use to trigger the issue:

```python

print("Running demo: expect startup to fail due to missing annotation")

app = FastAPI()


def get_db():
    return "fake-db"


@app.get("/")
def read_root(db=Depends(get_db)):
    return {"msg": "Hello World"}
```

When this module is run, FastAPI now raises the following error during dependency resolution:

```
fastapi.exceptions.FastAPIError: Dependency parameter "db" must have a type annotation. For example: `db: SomeType = Depends(...)`
```

Previously the same situation just raised a bare `AssertionError` with no guidance. With this change we now surface a clear, actionable message pointing directly at the missing annotation, which should make debugging this extremely common mistake much easier.

YuriiMotov:
> Previously the same situation just raised a bare `AssertionError` with no guidance.

The thing is that it doesn't raise `AssertionError` on `master` branch

PR #14638, opened 2026-01-02
📝 Clarify OpenAPI explanation

YASWANTH1976:
Adds a small clarification to make the OpenAPI explanation easier to understand for beginners.


tiangolo:
I don't think this is necessary, it's explained in other parts of the docs. I'll pass on this for now. ☕

PR #14641, opened 2026-01-02
🏷️ Re-export `IncEx` type from Pydantic instead of duplicating it

mvanderlee:
Loosen IncEx type declaration and ensure it stays in sync with Pydantic.
Currently it's too strict and an old copy that is out of sync. By removing the copy, we remove the possibility of it going out of sync. FastAPI only uses this for typing variables that are not used and passed directly into Pydantic.

* https://github.com/fastapi/fastapi/discussions/14640

### DIFF
--- a/fastapi/types.py
+++ b/fastapi/types.py
@@ -3,9 +3,9 @@ from enum import Enum
 from typing import Any, Callable, Optional, TypeVar, Union
 
 from pydantic import BaseModel
+from pydantic.main import IncEx as IncEx
 
 DecoratedCallable = TypeVar("DecoratedCallable", bound=Callable[..., Any])
 UnionType = getattr(types, "UnionType", Union)
 ModelNameMap = dict[Union[type[BaseModel], type[Enum]], str]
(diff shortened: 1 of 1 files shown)


YuriiMotov (approved):
@mvanderlee, thanks!

LGTM!

It was probably made this way for compatibility with Pydantic V1. But now since we dropped the support of V1, it's probably makes sense to just re-export this type from Pydantic.

FastAPI supports Pydantic 2.7+ and the declaration of this type will be different depending on Pydantic version (see [change history](https://github.com/pydantic/pydantic/blame/v2.12.5/pydantic/main.py)).

If you follow change history, there are a couple of fixes that affect this type definition.
So, at some point between Pydantic 2.7 and Pydantic 2.12 this type was incorrect.
But I think it's not a problem since all versions are compatible with the definition FastAPI currently uses (not strictly as FastAPI uses `Any` and Pydantic uses more concrete type annotations, but it shouldn't cause any issues).

tiangolo (approved):
Sounds good, thanks! 🚀

PR #14644, opened 2026-01-04
Docs: clarify response_model filtering behavior

YASWANTH1976:
Adds a short clarification explaining that response_model filtering applies only to the response and does not modify the returned object.


YuriiMotov (commented):
@YASWANTH1976, thanks for your interest!

I think this should be clear from the name of parameter and we don't need to change anything here

YASWANTH1976:
Thanks for taking a look and for the clarification.

That makes sense — I agree it can be inferred from the parameter name. 
I’ll close this PR. Thanks again for your time!

PR #14647, opened 2026-01-05
Fix HTTPBasic to always include realm in WWW-Authenticate

### LINKED ISSUE #13471: Basic auth `realm` is REQUIRED but handled as optional

WestinJiang:
Fixes fastapi/fastapi#13471

FastAPI previously returned `WWW-Authenticate: Basic` when HTTPBasic was used
without an explicit realm. Some clients expect a realm per RFC 7617 and may
fail without it.

This change ensures HTTPBasic always includes a realm (defaulting to "FastAPI")
and adds a regression test. Existing tests were updated accordingly.

Tests:
python -m pytest -q tests/test_security_http_basic_default_realm.py tests/test_security_http_basic_realm.py tests/test_security_http_basic_optional.py

### DIFF
--- a/docs_src/security/tutorial007_an_py39.py
+++ b/docs_src/security/tutorial007_an_py39.py
@@ -26,7 +26,7 @@ def get_current_username(
         raise HTTPException(
             status_code=status.HTTP_401_UNAUTHORIZED,
             detail="Incorrect username or password",
-            headers={"WWW-Authenticate": "Basic"},
+            headers=security.make_authenticate_headers(),
         )
     return credentials.username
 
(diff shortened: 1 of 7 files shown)


WestinJiang:
@rgilton
This PR is now fully green (tests + coverage).
The only remaining failure is the required `check-labels` workflow,
which I believe requires a maintainer-added label.
Could you please add one when you have a chance? Thanks!

YuriiMotov:
@WestinJiang, thanks for your interest!

Please, take a look at this comment to the issue you've mentioned: https://github.com/fastapi/fastapi/issues/13471#issuecomment-3596388412

WestinJiang:
Thanks for the clarification!

I agree that based on the updated interpretation of RFC 7235, the `realm`
parameter is no longer always required, so changing FastAPI’s default
behavior here isn’t necessary.

I’ll close this PR to avoid introducing an unnecessary behavior change.
Thanks for the detailed explanation!

PR #14649, opened 2026-01-05
Docs: clarify runtime vs OpenAPI behavior of responses model

champ-sk:


YuriiMotov (commented):
@champ-sk, thank you for interest!

I agree this might be confusing for newcomers.. (Although, this section is not for newcomers and there is a warning on the top of the page)

IMO, this line is a bit misleading:

> Those additional responses will be included in the OpenAPI schema, so they will also appear in the API docs.

The word **also** implies that it's not for schema generation only and people can infer that it's just an alternative for `response_model`.

So, it will probably be better to modify this line to make it clear that it will be used  for OpenAPI schema **only**.

How about the following version?

> Those additional responses will be included in the OpenAPI schema and will appear in the API docs.

If we correct this line, I think we will not need any additional clarifications.

@champ-sk, what do you think?

PR #14652, opened 2026-01-06
🔨 Add LLM translations tool fixer

YuriiMotov:
This PR adds script that can be used to automatically check and fix some errors in translated pages.

I added some tests to prevent breaking something in the future. They are currently placed in `scripts/tests` and are not run in CI with other tests. I don't know if we need to move them to `tests/` or keep as it is and run manually when we need to modify this script


tiangolo (approved):
Amazing job, thank you! 🙌 

Let's merge this right away and handle the rest in subsequent PRs.

PR #14654, opened 2026-01-06
Docs: clarify purpose and behavior of response_model

ynehra0777:
### Summary

Adds a clarification explaining why `response_model` is useful and how
it filters and validates response data.

### Motivation

New users often confuse request models with response models or are
unclear why extra fields are removed from responses. This PR adds a
short explanation and example to clarify the intended behavior.

### Changes

- Added a new subsection explaining the purpose of `response_model`
- Included a minimal example demonstrating field filtering

### Notes

Documentation-only change. No runtime behavior is affected.

### DIFF
--- a/docs/en/docs/tutorial/response-model.md
+++ b/docs/en/docs/tutorial/response-model.md
@@ -47,6 +47,42 @@ Notice that `response_model` is a parameter of the "decorator" method (`get`, `p
 
 `response_model` receives the same type you would declare for a Pydantic model field, so, it can be a Pydantic model, but it can also be, e.g. a `list` of Pydantic models, like `List[Item]`.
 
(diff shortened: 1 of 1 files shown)


ynehra0777:
This PR focuses on clarifying the intent and behavior of response_model
for beginners. Happy to adjust wording or placement if needed.

Thanks & Regards
Yogesh Nehra

YuriiMotov (commented):
@ynehra0777, thanks for your interest!

I think this is explained well in current version and there is no need to clarify it.
Please, read read the whole page from the start. It's said that `Response Model - Return Type` is used to validate and filter data, and also to generate output schema. And `response_model` Parameter` is just an alternative to return type annotation

PR #14655, opened 2026-01-06
Add support for Sentinel values

mvanderlee:
Fixes: https://github.com/fastapi/fastapi/discussions/14646


fastapi/dependencies/utils.py
```diff
@@ -67,6 +67,15 @@
 from starlette.websockets import WebSocket
 from typing_extensions import Literal, get_args, get_origin
 
+try:
+    # This was added in 4.14.0 (June 2, 2025) We want to be able to support older versions
+    from typing_extensions import Sentinel
```
Viicos, line 72: Note that this will need to be updated whenever sentinels gets added in the stdlib, to take both variants into account.

PR #14656, opened 2026-01-06
📝 Specify language code for code block

YuriiMotov:
We need to specify language code for this code block to be able to translate comments in translations of this page.
Otherwise [translation fixer tool](https://github.com/fastapi/fastapi/pull/14652) will replace the content of this block with content from English version.

I'll keep this PR as draft for some time to make sure we don't need more similar fixes.


tiangolo (approved):
Thanks! Let's merge this so that if we run the fixer, it will already be updated. 🤓

PR #14662, opened 2026-01-08
Require realm for HTTPBasic as per RFC 7617

Srikari12:
- Enforces that HTTPBasic requires a non-empty `realm` as per RFC 7617
- Adds/updates tests to ensure missing realm raises an error
- Keeps behavior consistent with spec and documentation

### DIFF
--- a/fastapi/security/http.py
+++ b/fastapi/security/http.py
@@ -5,7 +5,6 @@ from typing import Annotated, Optional
 from annotated_doc import Doc
 from fastapi.exceptions import HTTPException
 from fastapi.openapi.models import HTTPBase as HTTPBaseModel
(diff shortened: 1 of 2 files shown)


Srikari12:
This PR enforces RFC 7617 behavior for HTTP Basic authentication.
Suggested label: security

YuriiMotov:

Srikari12:
Thanks everyone for the discussion and for pointing me to the related comments.

After reviewing RFC 7235 (including Appendix A) and the previous discussions, I agree that realm should not be enforced as required at runtime, and that the current FastAPI behavior is correct.

Given that, I’m closing this PR to avoid conflicting behavior.

Thanks again for the review and guidance!

PR #14664, opened 2026-01-08
docs: clarify HTTP Basic realm is optional

Srikari12:
### What does this PR do?

Clarifies in the documentation that the `realm` parameter for HTTP Basic
authentication is optional, aligning with RFC 7235 Appendix A and current
FastAPI behavior.

### Why?

There has been confusion around whether `realm` is required per RFC 7617.
This update helps prevent misunderstandings without changing runtime behavior.

### Scope

- Documentation only
- No code or test changes


Srikari12:
This PR clarifies in the documentation that the HTTP Basic realm parameter is optional, aligning FastAPI’s docs with the current interpretation of RFC 7235 / RFC 7617.

No runtime behavior is changed — this update is purely documentation-focused to avoid confusion for users reading the HTTP Basic auth section.

YuriiMotov:
@Srikari12, note that you made changes in German version of that page.

Also, the new format for notes is:

```
/// note

    According to updated HTTP authentication specifications (RFC 7235),
    the `realm` parameter in HTTP Basic authentication is optional.
    FastAPI follows this behavior and does not require a `realm` to be set.

///
```

Srikari12:
> @Srikari12, note that you made changes in German version of that page.
> 
> Also, the new format for notes is:
> 
> ```
> /// note
> 
>     According to updated HTTP authentication specifications (RFC 7235),
>     the `realm` parameter in HTTP Basic authentication is optional.
>     FastAPI follows this behavior and does not require a `realm` to be set.
> 
> ///
> ```

Thanks for pointing that out! You’re right — I updated the German version by mistake.

I’ll move the change to the English documentation (`docs/en/...`) and update or revert the German file accordingly.

PR #14669, opened 2026-01-09
make type annotation for HTTPException.detail be Any

david-fong:
https://fastapi.tiangolo.com/tutorial/handling-errors/#fastapis-httpexception-vs-starlettes-httpexception

### DIFF
--- a/fastapi/exceptions.py
+++ b/fastapi/exceptions.py
@@ -41,6 +41,7 @@ class HTTPException(StarletteHTTPException):
         return {"item": items[item_id]}
     ```
     """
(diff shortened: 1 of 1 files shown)


david-fong:
Not sure what label to give this, or where to put the label.

YuriiMotov:
@david-fong, thanks for your interest and efforts!

I think we don't need to add this.
`detail` parameter of `__init__` method is already annotated as `Any` and type checkers can handle this and understand that `detail` attribute is `Any`.

So, I'll close this.
Thanks again!

PR #14670, opened 2026-01-09
Clarify "Required, can be None" section in query params docs

Anya299:
This PR clarifies how to pass None to required query parameters in FastAPI.
It adds a working example to prevent confusion. Fixes issue #12419.


YuriiMotov:
I think we already have such PR: https://github.com/fastapi/fastapi/pull/14322

Also, you made changes in  German version of page. And, all code examples should be in separate files and covered by tests.

Do you think we can close this PR?

PR #14673, opened 2026-01-09
docs: clarify dynamic usage of response status codes

YashGupta404:
This PR adds a short Notes section explaining common scenarios
where response status codes can be set dynamically.

No behavior is changed. Documentation only.


YuriiMotov:
@YashGupta404, thanks for your interest and efforts!

I think we don't need this additional clarification here - it's already explained well if you read the whole page.

I will close this to save Sebastian's time.
Thank you for understanding!

PR #14681, opened 2026-01-10
📝 Add `await` in `StreamingResponse` code example to allow cancellation

### LINKED ISSUE #14680: StreamingResponse does not work with async generators
[`StreamingResponse` docs](https://fastapi.tiangolo.com/advanced/custom-response/?h=responses#streamingresponse) state:

> Takes an async generator or a normal generator/iterator

It works as expected for normal generators, but blocks on `async` ones:

```py
from time import sleep
from fastapi import FastAPI

app = FastAPI()

async def fake_video_streamer():
    for i in range(10):
        sleep(0.1)
        yield b"some fake video bytes"

@app.get("/")
async def main():
    """
    this is as per
    https://fastapi.tiangolo.com/advanced/custom-response/?h=responses#streamingresponse
    but seems not to work (blocks for 1 sec, then returns all chunks at once):
      curl -sNo- localhost:8000
      wget -qO- localhost:8000
    """
    return StreamingResponse(fake_video_streamer())
```

however this works:

```py
def fake_video_streamer_sync():
    for i in range(10):
        sleep(0.1)
        yield b"some fake video bytes"

@app.get("/work-around")
async def works():
    """
    streams as expected (returns 1 chunk every 0.1s):
      curl -sNo- localhost:8000/work-around
      wget -qO- localhost:8000/work-around
    """
    return StreamingResponse(fake_video_streamer_sync())
```

casperdcl:
Fixes #14680

`async` generators passed to `starlette.StreamingResponse` must contain an `await` command[^starlette-1776] otherwise they block

[^starlette-1776]: https://github.com/Kludex/starlette/discussions/1776#discussioncomment-3207518


Ayu456sh:
Hi, I would like to work on this issue.
Please assign it to me.

docs_src/custom_response/tutorial007_py39.py
```diff
@@ -7,6 +9,7 @@
 async def fake_video_streamer():
     for i in range(10):
         yield b"some fake video bytes"
+        await asyncio.sleep(0.1)
```
YuriiMotov:
```suggestion
        await asyncio.sleep(0)
```

`0` is enough, no need to slow it down

YuriiMotov (commented):
@casperdcl, thanks!

I think we should add a note about this right after the code example.

Something like:

```

/// note

Note that we added `await asyncio.sleep(0)` inside the generator function.

This generator does not have any other `await` inside the loop. In asyncio, a task can only be cancelled when it reaches an `await`.

If there is no `await`, the generator can not be cancelled properly and may keep running even after cancellation is requested.

Adding `await asyncio.sleep(0)` gives the event loop a chance to handle cancellation.

///

```

What do you think?

casperdcl:
rebased & added note :)

YuriiMotov:
You shouldn't update translations. Please, only update English version

docs_src/custom_response/tutorial007_py310.py
```diff
@@ -7,6 +9,7 @@
 async def fake_video_streamer():
     for i in range(10):
         yield b"some fake video bytes"
+        await asyncio.sleep(0)  # simulate async work
```
YuriiMotov:
```suggestion
        await asyncio.sleep(0)
```

The comment is incorrect - this is not for simulation. Let's not add comment here, it's explained in the note below the code example

docs/en/docs/advanced/custom-response.md
```diff
@@ -217,7 +217,17 @@ You can also use the `status_code` parameter combined with the `response_class`
 
 Takes an async generator or a normal generator/iterator and streams the response body.
 
-{* ../../docs_src/custom_response/tutorial007_py310.py hl[2,14] *}
+{* ../../docs_src/custom_response/tutorial007_py310.py hl[4,17] *}
+
+/// note
+
+In `asyncio`, a task can only be cancelled when it reaches an `await`.
```
YuriiMotov: Why only `asyncio`? I think it will be the same for `anyio` and other async frameworks.

**UPD:** I see it was the same in my suggestion)

Let's change it to something like
```
Async task can only be cancelled when it reaches an `await`.
```

YuriiMotov (approved):
LGTM!

@casperdcl, thank you!

See also alternative wording: https://github.com/fastapi/fastapi/pull/14993

Areeb455:
Hey @casperdcl, I reviewed and tested this PR locally.

Testing Results:
- Ran the updated StreamingResponse example from docs_src.
- Verified that the stream initializes correctly and handles the generator as expected.
- Checked the doc alignment with the latest version of Starlette/FastAPI.

 The proposed fix correctly addresses the issue and makes the tutorial much more reliable for users. LGTM!

tiangolo (approved):
Thank you! 🚀 

I also had that in mind when building https://fastapi.tiangolo.com/tutorial/stream-json-lines/ and https://fastapi.tiangolo.com/advanced/stream-data/ , so that it's handled automatically with that new style. (about to be released in the next hours).

PR #14682, opened 2026-01-10
🔨 Fix retry logic for server check in `scripts/playwright`

Mohataseem89:
Problems:
break inside except stops the loop immediately after the first failure → retries are useless.
No check for response.status_code.

solution I propose:
Increased the retry attempts for server connection and added a timeout to the HTTP request.
Improved the HTTP Request Loop Logic:

### DIFF
--- a/scripts/playwright/sql_databases/image02.py
+++ b/scripts/playwright/sql_databases/image02.py
@@ -25,12 +25,16 @@ process = subprocess.Popen(
     ["fastapi", "run", "docs_src/sql_databases/tutorial002.py"],
 )
 try:
-    for _ in range(3):
+    for _ in range(10):
         try:
-            response = httpx.get("http://localhost:8000/docs")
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Good catch!

The same logic appears in other scripts in the `scripts/playwright` directory.
I think we should move this to util function and re-use in all those scripts.
@Mohataseem89, would you like to do this?

Mohataseem89:
Yes i noticed the same retry logic is duplicated in a few other scripts in the `scripts/playwright` directory. Moving it into a reusable utility function makes sense and will help keep the code consistent and easier to maintain.

I can work on extracting this into a shared util and update the related scripts accordingly

tiangolo:
This is a tiny script to generate screenshots for the docs run once and never again, there's no need to optimize it much. Thanks! :coffee:

PR #14689, opened 2026-01-11
docs: fix incorrect description of Header parameter conversion

0xFazal:
Corrected the explanation of how Header converts parameter names.

### DIFF
--- a/docs/en/docs/tutorial/header-params.md
+++ b/docs/en/docs/tutorial/header-params.md
@@ -38,7 +38,7 @@ Most of the standard headers are separated by a "hyphen" character, also known a
 
 But a variable like `user-agent` is invalid in Python.
 
-So, by default, `Header` will convert the parameter names characters from underscore (`_`) to hyphen (`-`) to extract and document the headers.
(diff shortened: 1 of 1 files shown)


alejsdev:
Hi @0xFazal, thanks for the interest. However, the doc is ok as it is and this rewording is incorrect. You write snake_case and FastAPI maps it to hyphenated HTTP headers (unless you disable that behavior with convert_underscores=False) https://fastapi.tiangolo.com/tutorial/header-params/?h=header+param#automatic-conversion
 
Given that, this will be closed. Thank you!

0xFazal:
> Hi @0xFazal, thank for the interest. However, the doc is ok as it is and this rewording is incorrect. You write snake_case and FastAPI maps it to hyphenated HTTP headers (unless you disable that behavior with convert_underscores=False) https://fastapi.tiangolo.com/tutorial/header-params/?h=header+param#automatic-conversion
>  
> Given that, this will be closed. Thank you!

Hi @alejsdev . Thanks for reviewing this. The documentation got me that the FastAPI converts hyphenated HTTP headers to underscore variables. Hence felt the need to correct the text.

PR #14696, opened 2026-01-11
🌐 Update translations for zh (update-outdated)

tiangolo:
🌐 Update translations for zh (update-outdated)

It uses the prompt file https://github.com/fastapi/fastapi/blob/master/docs/zh/llm-prompt.md.


tiangolo:
Let's re-run this with GPT-5

PR #14697, opened 2026-01-11
Fix HTTP Basic realm handling (RFC 7617)

### LINKED ISSUE #13471: Basic auth `realm` is REQUIRED but handled as optional

AnshMNSoni:
Per RFC 7617, the realm parameter is required for HTTP Basic authentication.
This PR makes realm mandatory and ensures the WWW-Authenticate header
always includes it.

Fixes #13471

### DIFF
--- a/fastapi/security/http.py
+++ b/fastapi/security/http.py
@@ -152,14 +152,7 @@ class HTTPBasic(HTTPBase):
                 """
             ),
         ] = None,
-        realm: Annotated[
-            Optional[str],
-            Doc(
-                """
-                HTTP Basic authentication realm.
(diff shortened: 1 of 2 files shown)


AnshMNSoni:
Hello maintainers, could you please add the appropriate label for this PR? Thank you.

YuriiMotov:
Please, take a look at: https://github.com/fastapi/fastapi/issues/13471#issuecomment-3596388412

svlandeg:
Closing, cf https://github.com/fastapi/fastapi/issues/13471#issuecomment-3596388412

PR #14049, opened 2025-09-06
✅ Add LLM test file

nilslindemann:
"Unit testing" and documentation of LLM instructions in the general and in language specific prompts for the translations of the documentation.

Prompt designers add tests here, other prompt designers can test for them while creating prompts.

This was previously part of #14015.

PR #14699, opened 2026-01-11
🌐 Update translations for ko (add-missing)

tiangolo:
🌐 Update translations for ko (add-missing)

It uses the prompt file https://github.com/fastapi/fastapi/blob/master/docs/ko/llm-prompt.md.


sodaMelon:
not very important but.. 

The link referenced by ```_llm-test``` (from https://github.com/fastapi/fastapi/pull/14049)
https://doublespeak.chat/#/handbook#deterministic-output
↑ doesn’t seem to be maintained or accessible anymore now. Is that correct?

hard-coders (approved):
I agree with @sodaMelon’s point that `llm-prompt.md` has room for improvement.
However, merging this PR first and addressing those issues iteratively would help deliver the Korean documentation to users sooner.

For these reasons, I’m approving this PR.

sodaMelon:
🤔
anyway i agree with @hard-coders  too 
semantically the translation reads reasonably well.

sodaMelon (approved):

tiangolo:
Awesome, thank you @hard-coders and @sodaMelon! :raised_hands:

PR #14700, opened 2026-01-11
🌐 Update translations for uk (add-missing)

tiangolo:
🌐 Update translations for uk (add-missing)

It uses the prompt file https://github.com/fastapi/fastapi/blob/master/docs/uk/llm-prompt.md.


roli2py:
I'll take it on a review

roli2py:
Figure out that LLM doesn't translate most of terms, so I started to fix this problem. PR: #14795

UPD: fix typos

wowkin2:
@roli2py not all terms should be translated because they are domain specific and commonly used as is (e.g. RAM, CLI or SDK).
Will have a look to your PR.

roli2py:
@wowkin2 okay, thanks!

tiangolo:
Let's run this again with the new LLM prompt, I'll close this one, a new one will be generated. :nerd_face:

PR #14703, opened 2026-01-11
Fix forward references inside Annotated dependencies

AnshMNSoni:
Fixes #13056

This PR fixes a bug where forward references inside typing.Annotated were not
resolved correctly for dependency injection and OpenAPI generation.

The fix resolves type hints using get_type_hints(..., include_extras=True)
before parameter analysis.

A regression test is included.


YuriiMotov:
@AnshMNSoni, thanks for your interest!

It looks like this doesn't really fix the issue - new test fails and several existing tests also started to fail..

PR #14718, opened 2026-01-13
Theme toggle functionality to Swagger UI

sukantamajhi:
Added theme toggle functionality with dark mdoe and light mode.


tiangolo:
I don't think this is necessary right now. Thanks! ☕

PR #14720, opened 2026-01-15
📝 Clarify when to use async def vs def

saimahesh8752:
This PR adds a small clarification to the documentation explaining when `async def` provides benefits and when a regular `def` function is more appropriate.

No behavior changes. Documentation-only improvement intended to reduce confusion for new users.

### DIFF
--- a/docs/en/docs/async.md
+++ b/docs/en/docs/async.md
@@ -351,7 +351,8 @@ async def read_burgers():
 
 **Clarification on `async def` usage**
 
-Using `async def` is beneficial only when the function performs asynchronous I/O operations, such as awaiting database calls, HTTP requests, or other non-blocking tasks. If a path operation function contains only blocking or CPU-bound code, defining it with `async def` does not provide performance benefits and may add unnecessary complexity. In those cases, using a regular `def` function is recommended.
+Using `async def` is beneficial primarily when the function performs I/O-bound work, such as awaiting network or disk operations (e.g. database calls, HTTP requests, or other non-blocking tasks). If a path operation function contains only blocking or CPU-bound code, defining it with `async def` does not provide performance benefits and may add unnecessary complexity. In such cases, using a regular `def` function is often simpler and clearer.
+
 
 ### More technical details { #more-technical-details }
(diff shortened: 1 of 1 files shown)


Harshpatil0508 (commented):
This clarification is helpful for newcomers.
Maybe we could also briefly mention that async def mainly helps when the function performs I/O-bound work (e.g. network or disk), otherwise def may be simpler and clearer.

saimahesh8752:
> This clarification is helpful for newcomers. Maybe we could also briefly mention that async def mainly helps when the function performs I/O-bound work (e.g. network or disk), otherwise def may be simpler and clearer.

Thanks for the suggestion!
I’ve updated the clarification to explicitly mention I/O-bound work (e.g. network or disk) and note that "def "may be simpler and clearer otherwise.

ceb10n:
Or `def` may **not** be simpler and clearer: [Be careful with non-async functions](https://github.com/kludex/fastapi-tips?tab=readme-ov-file#2-be-careful-with-non-async-functions)

As a general rule, I follow this: if you have sync blocking code, use sync routes, otherwise, always use async routes

YuriiMotov (commented):
I agree that this subject might be quite confusing for newcomers, but I'm afraid this clarification can confuse people even more..

* You can't use just `def` if you need to use async functions inside.
* I\O-bound operations can be synchronous and in this case `async def` is not a solution.

Also, it contradicts with the general recommendation to declare endpoints as `async def` unless your endpoint performs blocking I\O operations: https://fastapi.tiangolo.com/async/#path-operation-functions.

The explanation of @ceb10n looks much cleaner for me:

> if you have sync blocking code, use sync routes, otherwise, always use async routes

saimahesh8752:
Thanks for the clarification, that makes sense.
I see how this addition can conflict with the general recommendation and potentially add more confusion in this section.
I’m happy to remove the clarification and keep the guidance centralized in the main docs you referenced.

tiangolo:
Thanks for the interest, but I'll pass on this one for now. ☕

PR #14722, opened 2026-01-15
🐛 Fix copy button in custom.js

fcharrier:
I saw this bug in this discussion: https://github.com/fastapi/fastapi/discussions/14002
And I'm happy to contribute, even if it's a very small start

**Issue**: clicking the copy button was not working (error message in console)

**Fix**: the copy button is looking for a code tag, which doesn't exist, because it's replaced by a div in custom.js. I modified custom.js to put the div after the code tag, not in place

**Testing**: Tested locally on the fastAPI docs page.

### DIFF
--- a/docs/en/docs/js/custom.js
+++ b/docs/en/docs/js/custom.js
@@ -81,6 +81,8 @@ function setupTermynal() {
                     }
                 }
                 saveBuffer();
+                const inputCommands = useLines.filter(line => line.type === "input").map(line => line.value).join("\n");
+                node.textContent = inputCommands;
                 const div = document.createElement("div");
                 node.style.display = "none";
                 node.after(div);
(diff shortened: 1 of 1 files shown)


fcharrier:
Thanks @alejsdev  for your fix!

alejsdev:
Hi @fcharrier, thank you for your interest in contributing. You're right, but this fix only restored the copy button functionality, while copying the entire raw block instead of just the commands, which would be ideal.  I just pushed an extra change to extract only the input lines, get their values, and replace the hidden node's content with those commands. Now it only copies the command :smiley_cat: 

Thank you for the help! :)

fcharrier:
> Hi @fcharrier, thank you for your interest in contributing. You're right, but this fix only restored the copy button functionality, while copying the entire raw block instead of just the commands, which would be ideal. I just pushed an extra change to extract only the input lines, get their values, and replace the hidden node's content with those commands. Now it only copies the command 😺
> 
> Thank you for the help! :)

Yes, I saw your fix, and understand it. Thanks for your patience, I hope I will do it better the next time!

alejsdev (approved):

PR #14588, opened 2025-12-21
🌐 Update translations for ja (update-outdated)

tiangolo:
🌐 Update translations for ja (update-outdated)

PR #14726, opened 2026-01-17
✏️ Fix typo in `docs/ja/docs/tutorial/response-status-code.md`

NAM-MAN:


YuriiMotov:
@NAM-MAN, thank you for your interest!

We've been currently updating Japanese translations using LLM tool. There is a PR open and awaiting reviews from native speakers: https://github.com/fastapi/fastapi/pull/14588.
It would be nice if you could help us reviewing it!

As for this PR, our new translation process uses last commit dates to determine outdated pages (translation is considered as up to date if its last commit date is later than last commit date of original document).
So, fixing just one typo without syncing the whole document would break this process (this would override the commit date and we can miss this document during the next translation sync).

I'll close this PR. You can take a look at the translation of this page in PR https://github.com/fastapi/fastapi/pull/14588 to make sure that typo is fixed there.

Thank you!

PR #14727, opened 2026-01-17
Add tests for form data validation and optional fields

Vivansh27:
This PR adds unit tests for form data handling in the application:

1. `test_list_field_single_value`: Ensures that a single value in a list field is correctly wrapped as a list.
2. `test_alias_field_name_not_accepted`: Ensures that fields with unrecognized aliases return a 422 validation error.
3. `test_optional_int_empty_string`: Ensures that empty strings for optional integer fields are correctly converted to None.

These tests improve coverage for form input validation and ensure consistent API behavior.

### DIFF
--- a/tests/test_forms_single_model.py
+++ b/tests/test_forms_single_model.py
@@ -139,8 +139,8 @@ def test_extra_param_list():
         "param": "123",
         "extra_params": ["456", "789"],
     }
-   
-    
+
+
     def test_list_field_single_value():
     response = client.post(
         "/form/",
--- a/tests/test_forms_single_model.py
+++ b/tests/test_forms_single_model.py
@@ -139,3 +139,40 @@ def test_extra_param_list():
         "param": "123",
         "extra_params": ["456", "789"],
     }
+   
+    
+    def test_list_field_single_value():
+    response = client.post(
+        "/form/",
+        data={
+            "username": "Rick",
+            "lastname": "Sanchez",
+            "tags": "single",
+        },
+    )
+    assert response.status_code == 200
+    assert response.json()["tags"] == ["single"]
+
+
+    def test_alias_field_name_not_accepted():
+    response = client.post(
+        "/form/",
+        data={
+            "username": "Rick",
+            "lastname": "Sanchez",
+            "alias_with": "something",
+        },
+    )
+    assert response.status_code == 422
+
+  def test_optional_int_empty_string():
+    response = client.post(
+        "/form/",
+        data={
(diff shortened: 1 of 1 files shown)


tests/test_forms_single_model.py
```diff
@@ -139,3 +139,40 @@ def test_extra_param_list():
...
+    def test_alias_field_name_not_accepted():
+    response = client.post(
+        "/form/",
+        data={
+            "username": "Rick",
+            "lastname": "Sanchez",
+            "alias_with": "something",
+        },
+    )
+    assert response.status_code == 422
```
YuriiMotov, line 166: We already have such test: https://github.com/fastapi/fastapi/blob/0c7f2b66d769d05df42ea73f85205a3566ba127b/tests/test_request_params/test_form/test_required_str.py#L139-L156

tests/test_forms_single_model.py
```diff
@@ -139,3 +139,40 @@ def test_extra_param_list():
...
+    response = client.post(
+        "/form/",
+        data={
+            "username": "Rick",
+            "lastname": "Sanchez",
+            "age": "",
+        },
+    )
+    assert response.status_code == 200, response.text
+    assert response.json()["age"] is None
```
YuriiMotov, line 178: I think we should add such tests, but we should do it in more consistent way (different type of field, cover form declared as Pydantic model, different default values (not only `None`)).
I have a plan to add such set of tests

tests/test_forms_single_model.py
```diff
@@ -139,3 +139,40 @@ def test_extra_param_list():
...
+    response = client.post(
+        "/form/",
+        data={
+            "username": "Rick",
+            "lastname": "Sanchez",
+            "tags": "single",
+        },
+    )
+    assert response.status_code == 200
+    assert response.json()["tags"] == ["single"]
```
YuriiMotov, line 154: Something tells me that if we search in the code base we will find out that such test already exists..
I think we don't need to add this, but if we decide to add it we should do it in more consistent way (for different types (source) of parameters (e.g. Form, Query, Cookie, etc..))

YuriiMotov (commented):
@Vivansh27, thanks for your interest and efforts!

Please, take a look at my comments.

I suggest we close this PR.

I will make sure we add tests for the case with sending empty string to optional field - there is actually an issue with it and we are going to fix it. So, we will add such tests in that PR

tiangolo:
I'll pass on this for now, thanks. ☕

PR #14744, opened 2026-01-20
📝 Use `dfn` tag for definitions instead of `abbr` in docs

YuriiMotov:
In some cases the `<abbr>` tag is used to provide clarifications (or add jokes :) ) in docs. This makes it harder to instruct LLM to translate such tags correctly.

I suggest we use`<dfn>` tag for definitions: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dfn
Separating abbreviations from definitions will allow us to simplify LLM prompt and make translation results more stable.

I defined custom style in css to make it look similar to `<abbr>` tag (it still looks a bit different - `<dfn>` is _slanted_ while `<abbr>` is not, - not sure if it's important)

---

This way of using `<dfn>` doesn't actually follows the specification (according to spec, the `title` attribute [must contain the term being defined](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dfn#specifying_the_term_being_defined), not the definition), but I don't think it may cause any problems in our case.

---

Additionaly, I fixed a couple of abbr tags to follow the format that is used in other places and that is described in LLM prompt: see commit https://github.com/fastapi/fastapi/commit/b37b00ec1ce8d3f6619e4207a7f60b222ca6e65f

... and updated clarification for the word `fixed`: https://github.com/fastapi/fastapi/pull/14744/commits/60b988df414bb7fb5e54c29bea006f3ff9b5701e

### DIFF
--- a/docs/en/docs/deployment/https.md
+++ b/docs/en/docs/deployment/https.md
@@ -65,7 +65,7 @@ Here's an example of how an HTTPS API could look like, step by step, paying atte
 
 It would probably all start by you **acquiring** some **domain name**. Then, you would configure it in a DNS server (possibly your same cloud provider).
 
-You would probably get a cloud server (a virtual machine) or something similar, and it would have a <dfn title="That doesn't change">fixed</dfn> **public IP address**.
+You would probably get a cloud server (a virtual machine) or something similar, and it would have a <dfn title="Doesn't change over time. Not dynamic">fixed</dfn> **public IP address**.
 
 In the DNS server(s) you would configure a record (an "`A record`") to point **your domain** to the public **IP address of your server**.
 
--- a/docs/en/docs/deployment/https.md
+++ b/docs/en/docs/deployment/https.md
@@ -65,7 +65,7 @@ Here's an example of how an HTTPS API could look like, step by step, paying atte
 
 It would probably all start by you **acquiring** some **domain name**. Then, you would configure it in a DNS server (possibly your same cloud provider).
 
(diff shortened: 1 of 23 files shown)


YuriiMotov:
I created PR https://github.com/fastapi/fastapi/pull/14747 to update LLM-prompt (simplified instructions for `abbr` tag and added instructions for `dfn` tag).
Checked it and can confirm that it improves translations (solves problem with incorrectly translated `abbr` elements)

docs/en/docs/deployment/https.md
```diff
@@ -65,7 +65,7 @@ Here's an example of how an HTTPS API could look like, step by step, paying atte
 
 It would probably all start by you **acquiring** some **domain name**. Then, you would configure it in a DNS server (possibly your same cloud provider).
 
-You would probably get a cloud server (a virtual machine) or something similar, and it would have a <abbr title="That doesn't change">fixed</abbr> **public IP address**.
+You would probably get a cloud server (a virtual machine) or something similar, and it would have a <dfn title="Doesn't change over time. Not dynamic">fixed</dfn> **public IP address**.
```
tiangolo:
```suggestion
You would probably get a cloud server (a virtual machine) or something similar, and it would have a <dfn title="Doesn't change over time. Not dynamic.">fixed</dfn> **public IP address**.
```

tiangolo (approved):
Nice, thank you! 🚀

PR #14746, opened 2026-01-20
Use sorted() rather than list() for consistent operation_id naming

grantjenks:
The current code does list(methods)[0] where methods is a set and so picks one based on hash order (unreliable). The fix uses sorted(methods)[0] to consistently pick the first lexicographically.

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -124,7 +124,7 @@ def generate_unique_id(route: "APIRoute") -> str:
     operation_id = f"{route.name}{route.path_format}"
     operation_id = re.sub(r"\W", "_", operation_id)
     assert route.methods
-    operation_id = f"{operation_id}_{list(route.methods)[0].lower()}"
+    operation_id = f"{operation_id}_{sorted(route.methods)[0].lower()}"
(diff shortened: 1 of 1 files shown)


jogo-openai (approved):

YuriiMotov:
It's currently [recommended](https://github.com/fastapi/fastapi/pull/4851#issuecomment-2294657832) to use separate decorators for each method.

For example:

```py
@app.get("/")
@app.post("/")
def read_root():
    return {"Hello": "World"}
```

instead of

```py
@app.api_route("/", methods=["GET", "POST"])  # <- this is NOT recommended
def read_root():
    return {"Hello": "World"}
```

In case of separate decorators, `route.methods` will always contain one element and the result will be always consistent and correct.

The approach you propose doesn't fully solve the issue as it will use the same `operation_id` for both operations (but they must be unique).

There is another PR https://github.com/fastapi/fastapi/pull/10694 that attempts to solve this in a more complex way. It's in the queue to review by Sebastian.

Having this, I'l close this PR.

@grantjenks, thank you for your interest and efforts!

PR #14747, opened 2026-01-20
🔧 Update LLM-prompt for `abbr` and `dfn` tags

YuriiMotov:
Updated `scripts/general-llm-prompt.md` to simplify instructions for `abbr` tags and add instructions for `dfn` tags.

This should go after https://github.com/fastapi/fastapi/pull/14744

---

I ran translation script for `_llm-test.md` files of supported languages to sync them.
To do this I used `GPT-5` model instead of `GPT-5.2` as it gives better result (less mistakes and less diff).

I think we should consider downgrading model to `GPT-5` - I tested it on different pages (also with existing translation and without) and it gives better results most of the time


scripts/general-llm-prompt.md
```diff
@@ -421,45 +423,7 @@ Result (German):
...
-```
-<abbr title="auch bekannt als: Endpunkte, Routen">Pfad</abbr>
-<abbr title="Programm das auf Fehler im Code prüft">Linter</abbr>
-<abbr title="Konvertieren des Strings eines HTTP-Requests in Python-Daten">Parsen</abbr>
-<abbr title="vor 2023-03">0.95.0</abbr>
-<abbr title="2023-08-26">zum Zeitpunkt als das hier geschrieben wurde</abbr>
-```
-
-- If the text surrounded by the abbr element is an abbreviation and the description contains both the full phrase for that abbreviation, and other information, separated by a colon (`:`), then append a dash (`-`) and the translation of the full phrase to the original full phrase and translate the other information.
+- If the title of abbr element contains full phrase for that abbreviation, and other information, separated by a colon (`:`), then append a dash (`-`) and the translation of the full phrase to the original full phrase and translate the other information.
```
tiangolo:
```suggestion
- If the title of abbr element contains a full phrase for that abbreviation, and other information, separated by a colon (`:`), then append a dash (`-`) and the translation of the full phrase to the original full phrase and translate the other information.
```

tiangolo (approved):
Nice, thank you! 🤖

PR #14749, opened 2026-01-21
🔨 Update translation script to retry if LLM-response doesn't pass validation with Translation Fixer tool

YuriiMotov:
Sometimes LLM fails to generate translation correctly - introduces some mistakes that can be found by translation fixer tool, but can't be automatically fixed (e.g. wrong number of code blocks or code includes).

For example, in https://github.com/fastapi/fastapi/pull/14588 LLM preserved existing mistake in `virtual-environments.md` (the language of code block was specified after whitespace) and it caused pre-commit CI to fail.
Just running translation script second time on that file resolved the problem - https://github.com/fastapi/fastapi/pull/14588/commits/4eb0b839b992e8c03dd2e9e7c2eec028b3d5d5f9

This PR updates the translation script to check LLM response with translation fixer and retry if it has non-fixable mistakes

### DIFF
--- a/scripts/translate.py
+++ b/scripts/translate.py
@@ -136,7 +136,9 @@ def translate_page(
             )
             break  # Exit loop if no errors
         except ValueError as e:
-            print(f"Translation check failed on attempt {attempt_no}/{MAX_ATTEMPTS}: {e}")
+            print(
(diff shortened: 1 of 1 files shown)


tiangolo (approved):
Nice, let's try this! 🚀

PR #14750, opened 2026-01-21
docs: clarify "Required, can be None" query parameter wording

azamzar:
This PR adds a small clarification to the "Required, can be None" section.

Query parameters are always received as strings, which caused confusion for users expecting to send an actual None value. This change clarifies that a string representation must be explicitly handled in application logic.

No behavior or examples are changed.


azamzar:
Closing this PR as it duplicates #14322.
Sorry for the duplicate — it’s my first time contributing to FastAPI.
Thank you for your patience and guidance!

PR #14752, opened 2026-01-21
Fix XSS vulnerability

Tanmaykaturi:
This PR addresses a security issue identified by Semgrep. The vulnerable code used innerHTML to render user-controlled data, which could lead to XSS attacks.

**What was fixed:**

Replaced innerHTML with safe DOM manipulation (document.createElement)

Replaced direct HTML insertion with textContent for user input

Replaced attribute string concatenation with setAttribute() and className

Preserved existing behavior while removing XSS risk

### DIFF
--- a/docs/en/docs/js/termynal.js
+++ b/docs/en/docs/js/termynal.js
@@ -222,10 +222,26 @@ class Termynal {
      */
     lineDataToElements(lineData) {
         return lineData.map(line => {
-            let div = document.createElement('div');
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@Tanmaykaturi, could you please explain how you see possible scenario of using this vulnerability?

The worst think I can imagine is that someone opens a PR that injects this malicious JS code, GH actions will generate docs preview, and someone will open the docs preview link without reviewing code changes first.
But it's actually not really different from just posting a link to malicious website in the comments.

PR #14753, opened 2026-01-21
📝 Use return type annotation instead of `response_model` when possible

YuriiMotov:
Specifying the return type annotation should be a default way to specify response model.
This PR updates code examples to specify response model as return type annotation where it's possible.

I checked that surrounding text doesn't refer to `response_model` and highlights didn't get broken


tiangolo (approved):
Excellent, thank you! 🙌

PR #14754, opened 2026-01-22
style: PEP 8 fixes in openapi/utils.py

ImJuike:
**Changes:**
- Added missing trailing commas in `validation_error_response_definition`
- Fixed string concatenation in warning message

**Type:** Code style cleanup (PEP 8)

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -61,7 +61,7 @@ validation_error_response_definition = {
             "title": "Detail",
             "type": "array",
             "items": {"$ref": REF_PREFIX + "ValidationError"},
-        }
+        },
     },
 }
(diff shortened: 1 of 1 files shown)


ImJuike:
Hi! The label check is failing. Could a maintainer please add the "refactor" label? 

This PR only contains style fixes (trailing commas and string concatenation).

YuriiMotov:
We use ruff to check\fix code formatting.
If code passes ruff checks and it's readable, I think there is no need to update it.
And, if we decide to follow some additional optional rules, we should enable them in ruff's config.

In this case I think we don't need to change anything. I suggest we close this PR.

@ImJuike, thanks for your interest and efforts!

PR #14695, opened 2026-01-11
⬆️ Bump Starlette to <`0.53.0`

musicinmybrain:
In Starlette [0.51](https://github.com/Kludex/starlette/blob/0.51.0/docs/release-notes.md#0510-january-10-2026), the  warning stacklevel on `DeprecationWarning` for the wsgi module was increased. There is no impact to FastAPI; all tests still pass.

PR #14756, opened 2026-01-22
📝 Use `WSGIMiddleware` from `a2wsgi` instead of deprecated `fastapi.middleware.wsgi.WSGIMiddleware`

YuriiMotov:
`fastapi.middleware.wsgi.WSGIMiddleware` (re-exported from Starlette) is deprecated:

https://github.com/Kludex/starlette/blob/501b86185461d107869917ced367d5624f82e3fb/starlette/middleware/wsgi.py#L15-L20

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -232,7 +232,6 @@ xfail_strict = true
 junit_family = "xunit2"
 filterwarnings = [
     "error",
-    'ignore:starlette.middleware.wsgi is deprecated and will be removed in a future release\..*:DeprecationWarning:starlette',
     # see https://trio.readthedocs.io/en/stable/history.html#trio-0-22-0-2022-09-28
     "ignore:You seem to already have a custom.*:RuntimeWarning:trio",
     # TODO: remove after upgrading SQLAlchemy to a version that includes the following changes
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -178,6 +178,7 @@ tests = [
     "strawberry-graphql>=0.200.0,<1.0.0",
     "types-orjson==3.6.2",
     "types-ujson==5.10.0.20240515",
+    "a2wsgi>=1.9.0,<=2.0.0",
```
YuriiMotov, line 181: With versions prior to `1.9.0` it gives error `404 Not found`

mynewestgitaccount:
WSGIMiddleware in Starlette has been deprecated for a very long time.

The deprecation warning that is causing CI issues in #14695 from the newest version of Starlette is causing issues because Starlette [changed the stacklevel of the warning](https://github.com/Kludex/starlette/pull/3082).

I think the correct fix from the FastAPI side would be updating the existing warning filter in pyproject.toml: https://github.com/fastapi/fastapi/blob/597b435ae7f0c4694f654bce6c7c27206aa690ef/pyproject.toml#L234

YuriiMotov:
@mynewestgitaccount, thanks for pointing to that warning filter! I missed it.

I still think we should update docs to teach people use recommended `a2wsgi.WSGIMiddleware`. Starlette's version is deprecated and will be removed at some point, we need to prepare for this

tiangolo (approved):
Excellent, thank you! 🚀 🙌

PR #14758, opened 2026-01-22
📝 Update embedded code examples to Python 3.10 syntax

YuriiMotov:


svlandeg (approved):
LGTM!

ceb10n (approved):
it looks so good to get rid of Union, Optional, List, etc. 😃

ulgens:
Would the project consider using an automated solution, like blacken-docs?

YuriiMotov:
There are only a few embedded code examples in FastAPI docs. Most of code examples are in separate `.py`-files for different Python versions.
So, I'm not sure it's worth setting up the new tool and marking some code examples as exceptions (when we need to show the older syntax).

But the tool is interesting, thanks for pointing to it!

ulgens:
Thanks for the insight, that makes sense.

tiangolo (approved):
Excellent, thank you! 🙌

PR #14762, opened 2026-01-22
✨ Allow including additional scripts to Swagger UI page and adding presets

YuriiMotov:
For now there is no easy way to enable `StandaloneLayout` and dark mode that comes with it.

To do this you need to:
* include additional JS script (https://cdn.jsdelivr.net/npm/swagger-ui-dist@5/swagger-ui-standalone-preset.js)
* add `SwaggerUIStandalonePreset` to presets
* change the default `layout` swagger parameter to `StandaloneLayout` (this is possible now, but will give `No layout defined for "StandaloneLayout"` without first two steps)

With changes from this PR we will be able to enable `StandaloneLayout` the following way:

```diff
from fastapi import FastAPI
from fastapi.openapi.docs import (
    get_swagger_ui_html,
    get_swagger_ui_oauth2_redirect_html,
)

app = FastAPI(docs_url=None, redoc_url=None)


@app.get("/docs", include_in_schema=False)
async def custom_swagger_ui_html():
    return get_swagger_ui_html(
        openapi_url=app.openapi_url,
        title=app.title + " - Swagger UI",
        oauth2_redirect_url=app.swagger_ui_oauth2_redirect_url,
+       swagger_extra_js_urls=["https://cdn.jsdelivr.net/npm/swagger-ui-dist@5/swagger-ui-standalone-preset.js"],
+       swagger_extra_presets=["SwaggerUIStandalonePreset"],
+       swagger_ui_parameters={"layout": "StandaloneLayout"},
    )


@app.get(app.swagger_ui_oauth2_redirect_url, include_in_schema=False)
async def swagger_ui_redirect():
    return get_swagger_ui_oauth2_redirect_html()


@app.get("/users/{username}")
async def read_user(username: str):
    return {"message": f"Hello {username}"}

```

image

### DIFF
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -141,7 +141,6 @@ def get_swagger_ui_html(
     html += """
     presets: [
         SwaggerUIBundle.presets.apis,
-        SwaggerUIBundle.SwaggerUIStandalonePreset
         ],
     })"""
 
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -133,7 +133,9 @@ def get_swagger_ui_html(
     js_urls = [swagger_js_url]
     if swagger_extra_js_urls:
         js_urls.extend(swagger_extra_js_urls)
-    scripts_str = "\n    ".join(f'<script src="{js_url}"></script>' for js_url in js_urls)
+    scripts_str = "\n    ".join(
+        f'<script src="{js_url}"></script>' for js_url in js_urls
+    )
 
     html = f"""
     <!DOCTYPE html>
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -54,6 +54,14 @@ def get_swagger_ui_html(
             """
         ),
     ] = "https://cdn.jsdelivr.net/npm/swagger-ui-dist@5/swagger-ui-bundle.js",
+    swagger_extra_js_urls: Annotated[
+        Optional[list[str]],
+        Doc(
+            """
+            The URLs of additional JavaScript files to include.
+            """
+        ),
+    ] = None,
     swagger_css_url: Annotated[
         str,
         Doc(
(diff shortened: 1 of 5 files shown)


fastapi/openapi/docs.py
```diff
@@ -138,12 +161,18 @@ def get_swagger_ui_html(
     if oauth2_redirect_url:
         html += f"oauth2RedirectUrl: window.location.origin + '{oauth2_redirect_url}',"
 
-    html += """
+    presets = ["SwaggerUIBundle.presets.apis"]
```
sairamveereddy, line 164: This changes the default preset list from including both `SwaggerUIBundle.presets.apis` and `SwaggerUIBundle.SwaggerUIStandalonePreset` to only including `SwaggerUIBundle.presets.apis`. Even if the new `swagger_extra_presets` hook supports adding it back, dropping an existing default seems like a behavior change for users who rely on the current generated docs HTML. Would it be safer to keep `SwaggerUIBundle.SwaggerUIStandalonePreset` in the default list and append `swagger_extra_presets`, with a regression test that the existing default preset is still rendered?

sairamveereddy (commented):
I reviewed the Swagger UI preset changes and found one compatibility point that seems worth resolving before this leaves draft.

fastapi/openapi/docs.py
YuriiMotov: Thanks! I will take a look later

YuriiMotov: Ok, I looked into it and I don't remember why I removed `SwaggerUIBundle.SwaggerUIStandalonePreset` from default `presets`. But it's clear that I did it intentionally (I updated tests). I should've clarified that in the description..

Anyway, I don't think this PR has chances to be merged any time soon (we have other priorities). So, let's close this for now. We have internal task for this, and there is a link to this PR from that task. So, it will not be lost and we can continue from here than we have time for this task.

Thanks for review!

PR #14764, opened 2026-01-23
✨ Add dataclasses-based route parameters

HoverHell:
Proposal for discussion:

Adding a dataclass that represents the arguments to `api_route`
allows for custom route-building functions
without having to copy all the parameters supported in `api_route`.

For an example, see [classy_fastapi](https://gitlab.com/companionlabs-opensource/classy-fastapi/-/blob/5ae37f2395e7df5b1d3c9b570d1802f2e3fcdd1c/classy_fastapi/decorators.py#L43), or [this related code and discussion](https://stackoverflow.com/q/79873541/62821).

### DIFF
--- a/fastapi/datastructures.py
+++ b/fastapi/datastructures.py
@@ -1,3 +1,4 @@
+import dataclasses
 from collections.abc import Mapping
 from typing import (
     Annotated,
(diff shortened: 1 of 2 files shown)


HoverHell:
-> https://github.com/fastapi/fastapi/pull/14774

PR #14766, opened 2026-01-23
Fix Optional[Literal] Form field handling when value is None

developreiloyo:
### 🐛 Bug

`Optional[Literal[...]]` form fields were returning `422 Unprocessable Entity`
when the submitted value was `None`, even though the field was optional.

This behavior was inconsistent with other optional form fields such as
`Optional[str]` or `Optional[List[str]]`.

### ✅ Fix

- Properly handle `None` values for `Optional[Literal]` form fields
- Align behavior with other optional form parameters

### 🧪 Tests

- Added tests reproducing the issue
- Verified no regressions in existing request parameter tests

### DIFF
--- a/tests/test_request_params/test_form/test_optional_literal_form.py
+++ b/tests/test_request_params/test_form/test_optional_literal_form.py
@@ -1,4 +1,5 @@
 from typing import Annotated, Literal, Optional
+
 from fastapi import FastAPI, Form
 from fastapi.testclient import TestClient
 
@@ -8,9 +9,7 @@ def test_optional_literal_form_none():
     app = FastAPI()
 
     @app.post("/")
-    async def read_main(
-        attribute: Annotated[Optional[Literal["abc", "def"]], Form()]
-    ):
+    async def read_main(attribute: Annotated[Optional[Literal["abc", "def"]], Form()]):
         return {"attribute": attribute}
 
     client = TestClient(app)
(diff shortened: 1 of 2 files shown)


tests/test_request_params/test_form/test_optional_literal_form.py
```diff
@@ -0,0 +1,77 @@
...
+    @app.post("/")
+    async def read_main(attribute: Annotated[Optional[Literal["abc", "def"]], Form()]):
+        return {"attribute": attribute}
+
+    client = TestClient(app)
+
+    # FIXED: Use empty data instead of {"attribute": None}
+    response = client.post("/", data={})
+    assert response.status_code == 200
+    assert response.json() == {"attribute": None}
```
YuriiMotov, line 20: This test is incorrect..
The `attribute` parameter is **nullable**, but still **required**. So, not passing it should result in 422 error response

tests/test_request_params/test_form/test_optional_literal_form.py
```diff
@@ -0,0 +1,77 @@
...
+    async def read_main(attribute: Annotated[Optional[Literal["abc", "def"]], Form()]):
+        return {"attribute": attribute}
+
+    client = TestClient(app)
+
+    # Empty string is treated as "not provided" for optional Form fields
+    # This is consistent with FastAPI's behavior for Form data
+    response = client.post("/", data={"attribute": ""})
+    assert response.status_code == 200
+    assert response.json() == {"attribute": None}
```
YuriiMotov, line 77: This is a known issue, and it fails for `Optional[str]` as well.
Just change the type annotation to `Optional[str]` and run this test on `master` branch - you will see the same error

YuriiMotov (commented):
@developreiloyo, thanks for your interest and efforts!

There is currently issue with passing empty str to optional Form field that has non-`None` default value.
The problem is that it's not related to `Literal` - it will fail with any other type as well.

So, it should be solved in different way

tiangolo:
Thanks for the effort! We'll need to handle this in a different way. I'll pass on this for now. ☕

PR #14769, opened 2026-01-24
Docs: clarify handling of extra fields in request body

KevynValladares21:
This PR adds a short note to the Request Body tutorial clarifying that extra fields in the request body are ignored by default when using Pydantic models.

This behavior surprised me while following the tutorial, so I made it explicit to help readers understand what to expect.

### DIFF
--- a/docs/en/docs/tutorial/body.md
+++ b/docs/en/docs/tutorial/body.md
@@ -57,7 +57,7 @@ For example, this model above declares a JSON "`object`" (or Python `dict`) like
 
 /// note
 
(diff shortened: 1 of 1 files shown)


KevynValladares21:
Hi! Just checking in on this PR. It’s a small docs clarification and all checks have passed.

If you'd like any wording or placement changes, I'm happy to update it. Thanks!

YuriiMotov (commented):
@KevynValladares21, thanks for your interest!

I think we don't need this clarification.
It should be already clear that since we use Pydantic model to declare request body, people should find answers to their questions in the docs of Pydantic.
We just can't describe all Pydantic features and limitations here. We should keep docs simple

KevynValladares21:
Got it, that makes sense, thanks for the response!

I appreciate the clarification about keeping FastAPI docs focused and deferring detailed behavior to Pydantic docs. Thanks for taking a look at the PR.

PR #14771, opened 2026-01-25
Set cooldown in dependabot

p-r-a-v-i-n:
Source: https://docs.zizmor.sh/audits/#dependabot-cooldown

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,8 @@ updates:
       interval: "daily"
     commit-message:
       prefix: ⬆
+    cooldown:
+      default-days: 7
   # Python
   - package-ecosystem: "uv"
     directory: "/"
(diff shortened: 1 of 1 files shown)


ulgens:
Was there any discussion about zizmor?

svlandeg (commented):
I'm not sure I'm a fan of this `cooldown` period. Yes, new versions may have unexpected bugs, but delaying the update also delays installing the next update with the fix 🥲 

Also, it's not like `dependabot` runs daily here, so we already have an inherit cooldown period 🤷

p-r-a-v-i-n:
> I'm not sure I'm a fan of this `cooldown` period. Yes, new versions may have unexpected bugs, but delaying the update also delays installing the next update with the fix 🥲
> 
> Also, it's not like `dependabot` runs daily here, so we already have an inherit cooldown period 🤷

make sense

PR #14781, opened 2026-01-29
📝 Update `docs/en/docs/tutorial/security/oauth2-jwt.md`

zadevhub:
The word emit is difficult to follow. I had to look it up in this context. This change makes the sentence more readable and easier for the user to follow.


docs/en/docs/tutorial/security/oauth2-jwt.md
```diff
@@ -18,7 +18,7 @@ eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4
 
 It is not encrypted, so, anyone could recover the information from the contents.
 
-But it's signed. So, when you receive a token that you emitted, you can verify that you actually emitted it.
+But it's signed. So, when you receive a token that you originally issued, you can verify that it actually came from you and has not been tampered with in-transit.
```
YuriiMotov:
```suggestion
But it's signed. So, when you receive a token that you issued, you can verify that it was you who issued it.
```

I agree that `issue` is better than `emit`, but I suggest to make it shorter

tiangolo (approved):
Thanks! 🍰

PR #14786, opened 2026-01-30
🐛 Strip whitespaces from `Authorization` header credentials

WaveTheory1:
## Description

Fixes whitespace handling in authorization header parameter extraction to comply with [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1).

## Problem

The current implementation preserves leading/trailing whitespace in the extracted credential parameter. While RFC 6750 allows one or more spaces (`1*SP`) between "Bearer" and the token, the token definition (`b64token`) doesn't include whitespace characters, meaning spaces should be treated as separators only.

This causes the extracted credentials to include extraneous whitespace, which can lead to authentication failures.

## Solution

Added `.strip()` to the parameter extraction to remove leading and trailing whitespace:
```python
return scheme, param.strip()
```

### Before the fix:
```python
from typing import Annotated
from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer

app = FastAPI()
security = HTTPBearer()

@app.get("/users/me")
def read_current_user(
    credentials: Annotated[HTTPAuthorizationCredentials, Depends(security)]
):
    return {"scheme": credentials.scheme, "credentials": credentials.credentials}
```

**Test case 1: Single space (correct)**
```bash
curl -X 'GET' \
  'http://127.0.0.1:8000/users/me' \
  -H 'Authorization: Bearer xxx'
```
Response: `{"scheme":"Bearer","credentials":"xxx"}` ✓

**Test case 2: Multiple spaces**
```bash
curl -X 'GET' \
  'http://127.0.0.1:8000/users/me' \
  -H 'Authorization: Bearer   xxx'
```
Response: `{"scheme":"Bearer","credentials":"  xxx"}` ✗

**Test case 3: Trailing whitespace**
```bash
curl -X 'GET' \
  'http://127.0.0.1:8000/users/me' \
  -H 'Authorization: Bearer xxx '
```
Response: `{"scheme":"Bearer","credentials":"xxx "}` ✗

### After the fix:

All three test cases now return: `{"scheme":"Bearer","credentials":"xxx"}` ✓

## Changes

- Modified `get_authorization_scheme_param()` in `fastapi/security/utils.py` to strip whitespace from the parameter
- Added test cases to verify the fix handles multiple spaces and trailing whitespace correctly

## Checklist

- [x] Tests added to cover the bug fix
- [x] Existing tests pass


YuriiMotov (approved):
LGTM!

@WaveTheory1, thanks!

The latest RFC that defines `Authorization` header structure is RFC 9110 (see [Authorization](https://datatracker.ietf.org/doc/html/rfc9110#section-11.6.2), [credentials](https://datatracker.ietf.org/doc/html/rfc9110#section-11.4)

RFC 9110 obsoletes RFC 7235 (Authorization header is defined the same way: see [Authorization](https://datatracker.ietf.org/doc/html/rfc7235#section-4.2), [credentials]https://datatracker.ietf.org/doc/html/rfc7235#section-2.1))

RFC 7617 (HTTP Basic) doesn't specify format of Authorization header, refers to RFC 7235 (see [The 'Basic' Authentication Scheme](https://datatracker.ietf.org/doc/html/rfc7617#section-2))

RFC 7616 (Digest) [refers](https://datatracker.ietf.org/doc/html/rfc7617#section-2) to https://datatracker.ietf.org/doc/html/rfc7235#section-2.1

RFC 6750 (Bearer) defines the `Authorization` header format explicitly, and also [refers](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1) to [RFC 2617](https://datatracker.ietf.org/doc/html/rfc2617#section-2) (which is obsoleted by RFC 7235 and then by RFC 9110)

I didn't find anything about trailing whitespaces (if they are allowed by standard), but it's 100% that we shouldn't treat them as a part of credentials. So, I think it's fine to trim trailing whitespaces.

So, I think this PR is in line with all specs

WaveTheory1:
Thanks for the review @YuriiMotov! Happy to get this merged whenever you're ready 😄

tiangolo (approved):
Thanks! 🙌 

I was double checking the specs, the `token68`, the value for the token, it's defined here: https://datatracker.ietf.org/doc/html/rfc9110#section-11.2

And indeed, it says:

> The token68 syntax allows the 66 unreserved URI characters ([[URI](https://datatracker.ietf.org/doc/html/rfc3986)]), plus a few others, so that it can hold a base64, base64url (URL and filename safe alphabet), base32, or base16 (hex) encoding, with or without padding, but excluding whitespace ([[RFC4648](https://datatracker.ietf.org/doc/html/rfc4648)]).

It does say "excluding whitespace". 🤓 

This will be available in FastAPI 0.128.1, in the next few hours. 🎉

PR #14789, opened 2026-01-31
🐛 Fix TYPE_CHECKING annotations for Python 3.14 (PEP 649)

mgu:
PR #14485 fixed support for if TYPE_CHECKING, non-evaluated stringified annotations, when using `from __future__ import annotations`.

This change adds the same support when using deferred evaluation of annotations introduced in Python 3.14.

Note: I was also able to reproduce the issue by removing `from __future__ import annotations` from `tests/test_stringified_annotation_dependency.py` (using python 3.14). 

It has been [discussed here](https://github.com/fastapi/fastapi/discussions/14784)

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -204,7 +204,12 @@ def _get_signature(call: Callable[..., Any]) -> inspect.Signature:
         except NameError:
             # Handle type annotations with if TYPE_CHECKING, not used by FastAPI
             # e.g. dependency return types
-            signature = inspect.signature(call)
+            if sys.version_info >= (3, 14):
+                from annotationlib import Format
+
+                signature = inspect.signature(call, annotation_format=Format.FORWARDREF)
+            else:
+                signature = inspect.signature(call)
     else:
         signature = inspect.signature(call)
     return signature
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -6,8 +6,8 @@ needs_py39 = pytest.mark.skipif(sys.version_info < (3, 9), reason="requires pyth
 needs_py310 = pytest.mark.skipif(
     sys.version_info < (3, 10), reason="requires python3.10+"
 )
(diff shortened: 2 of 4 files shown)


tests/utils.py
```diff
@@ -6,8 +6,8 @@
 needs_py310 = pytest.mark.skipif(
     sys.version_info < (3, 10), reason="requires python3.10+"
 )
-needs_py_lt_314 = pytest.mark.skipif(
-    sys.version_info >= (3, 14), reason="requires python3.13-"
+needs_py314 = pytest.mark.skipif(
+    sys.version_info < (3, 14), reason="requires python3.14-"
```
mgu: `needs_py_lt_314` wasn't used

fastapi/dependencies/utils.py
```diff
@@ -204,7 +204,12 @@ def _get_signature(call: Callable[..., Any]) -> inspect.Signature:
         except NameError:
             # Handle type annotations with if TYPE_CHECKING, not used by FastAPI
             # e.g. dependency return types
-            signature = inspect.signature(call)
+            if sys.version_info >= (3, 14):
+                from annotationlib import Format
```
mgu, line 208: I could move this import to top level, wdyt?

YuriiMotov (approved):
LGTM!
@mgu, thank you!

Probably better title: 🐛 Fix `TYPE_CHECKING` annotations for Python 3.14 (PEP 649)

mgu:
@YuriiMotov I don't know the processes here, do I have something else to do on my side? 😊

YuriiMotov:
> @YuriiMotov I don't know the processes here, do I have something else to do on my side? 😊

Next step is review by Sebastian. We should just wait and resolve merge conflicts if any (PR will be marked with `conflicts` label if there are merge conflicts)

tiangolo (approved):
Nice, thank you! This will be available in FastAPI 0.128.1, released in the next few hours. 🚀

Issue #10787, opened 2023-12-16
Old `HTTPValidationError` and `ValidationError` OpenAPI entry


Kludex:
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

### Issue Content

Those lines need to be updated:

https://github.com/tiangolo/fastapi/blob/36c26677682c4245183912e00fc057c05cc6cf7a/fastapi/openapi/utils.py#L36-L61

They reflect how it was before 0.100.0, but with Pydantic V2, that's no longer true.

Kludex:
It would be cool to have an API to modify the `RequestValidationError` OpenAPI entry, and not only the `exception_handler`.

iudeen:
```
validation_error_definition = {
    "title": "ValidationError",
    "type": "object",
    "properties": {
        "loc": {
            "title": "Location",
            "type": "array",
            "items": {"anyOf": [{"type": "string"}, {"type": "integer"}]},
        },
        "input": {"title": "Input", "type": "any"},
        "url": {"title": "URL", "type": "string"},
        "msg": {"title": "Message", "type": "string"},
        "type": {"title": "Error Type", "type": "string"},
    },
    "required": ["loc", "msg", "type"],
}
```

This is what we need right? But this change would require changing a lot of tests.

DeepTanthetaa:
`from` fastapi import FastAPI

app = FastAPI()

# Your existing OpenAPI definitions
validation_error_definition = {...}
validation_error_response_definition = {...}

# Add the existing definitions to the app
app.openapi_schema["components"]["schemas"]["ValidationError"] = validation_error_definition
app.openapi_schema["components"]["schemas"]["HTTPValidationError"] = validation_error_response_definition

# Define the new definition for HTTPValidationError (modify as needed)
new_http_validation_error_definition = {
    "title": "ModifiedHTTPValidationError",
    "type": "object",
    "properties": {
        "detail": {
            "title": "Detail",
            "type": "array",
            "items": {"$ref": "#/components/schemas/ValidationError"},
        }
    },
}

# Update the OpenAPI entry
update_openapi_entry(app, new_http_validation_error_definition)

# Your FastAPI routes and other configurations go here

using this code stay happy 👍

YuriiMotov:
`validation_error_response_definition` seems to be correct at the moment, but `validation_error_response_definition` should be updated by adding the absent `input` and `ctx` fields.

[...]

PR #14791, opened 2026-02-01
🐛 Update `ValidationError` schema to include `input` and `ctx`

jonathan-fulton:
## Summary

Related to #10787

The OpenAPI `ValidationError` schema was outdated and didn't reflect the actual error format returned by Pydantic V2.

## Changes
Updated `validation_error_definition` in `fastapi/openapi/utils.py` to include:
- `input`: The actual input value that caused the validation error  
- `ctx`: Optional context dictionary with additional error information

## Testing
All existing tests pass with the updated schema. The test snapshots have been updated to reflect the new schema structure.

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -49,6 +49,8 @@ validation_error_definition = {
         },
         "msg": {"title": "Message", "type": "string"},
         "type": {"title": "Error Type", "type": "string"},
+        "input": {"title": "Input"},
+        "ctx": {"title": "Context", "type": "object"},
     },
     "required": ["loc", "msg", "type"],
 }
--- a/tests/test_additional_properties.py
+++ b/tests/test_additional_properties.py
@@ -89,8 +89,8 @@ def test_openapi_schema():
                         },
                         "msg": {"title": "Message", "type": "string"},
                         "type": {"title": "Error Type", "type": "string"},
-                    "input": {"title": "Input"},
-                    "ctx": {"title": "Context", "type": "object"},
+                        "input": {"title": "Input"},
+                        "ctx": {"title": "Context", "type": "object"},
                     },
                 },
                 "HTTPValidationError": {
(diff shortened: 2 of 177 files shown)


YuriiMotov:
Looks good in general, but would be nice to check and prove that client code generators interpret this schema correctly

**UPD:** I just tried generating client using `openapi-ts` (which is used in [full-stack-fastapi-template](https://github.com/fastapi/full-stack-fastapi-template)) and everything is good:

```typescript
export type ValidationError = {
    loc: Array<(string | number)>;
    msg: string;
    type: string;
    input?: unknown;
    ctx?: {
        [key: string]: unknown;
    };
};
```

jonathan-fulton:
@YuriiMotov Good point! I did some research on code generator compatibility:

**The Schema:**
- `ctx` has `type: object` ✅ - well-supported
- `input` has no explicit type - this represents "any type" which is valid JSON Schema, but some generators handle it better than others

**Testing with openapi-generator:**

I tested generating a Python client with the updated schema:
```bash
# Generated with openapi-generator-cli 7.x
openapi-generator generate -i openapi.json -g python -o ./client
```

Result: ✅ Generated successfully. The `input` field is typed as `Any` in Python, which correctly represents that it can be any JSON value (string, number, object, array, etc. - whatever the invalid input was).

**Why `input` has no type:**
Pydantic's `ValidationError` includes the actual failing input value in the `input` field, which could be any JSON type depending on what the user submitted. This matches Pydantic's behavior and the actual error responses FastAPI returns.

**Alternative if needed:**
If we want to be more explicit, we could add:
```python
"input": {"title": "Input", "anyOf": [{"type": "string"}, {"type": "number"}, {"type": "integer"}, {"type": "boolean"}, {"type": "array"}, {"type": "object"}, {"type": "null"}]}
```

But the current approach (no type = any type) is valid JSON Schema and works with the major generators I tested.

Let me know if you'd like me to test with any specific code generator!

YuriiMotov (approved):
LGTM!

@jonathan-fulton, thank you!

tiangolo (approved):
Thank you! ☕ 

This will be available in FastAPI 0.128.1, released in the next few hours.

PR #14792, opened 2026-02-01
feat: automatically support HEAD method for all GET routes

### LINKED ISSUE #1773: Automatically support HEAD method for all GET routes, as Starlette does

jonathan-fulton:
## Summary
Fixes #1773

FastAPI was returning 405 Method Not Allowed for HEAD requests to GET endpoints, contrary to HTTP specifications and Starlette's behavior. HEAD requests should work automatically for all GET endpoints.

## Changes
1. Added HEAD to the methods set when GET is present in `APIRoute.__init__`
2. Skip auto-added HEAD (when paired with GET) in OpenAPI schema generation  
3. Updated `generate_unique_id` to use deterministic method selection

## Features
- HEAD requests now work automatically for all GET endpoints
- HEAD returns same headers as GET but with empty body
- Explicit HEAD routes still work when defined before GET routes
- HEAD is not shown in OpenAPI schema (it's implicit per HTTP semantics)

## Testing
Added comprehensive tests in `tests/test_head_method.py`

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -278,6 +278,11 @@ def get_openapi_path(
     route_response_media_type: Optional[str] = current_response_class.media_type
     if route.include_in_schema:
         for method in route.methods:
+            # Skip auto-added HEAD method in OpenAPI when it's paired with GET.
+            # HEAD is automatically supported for all GET endpoints per HTTP semantics.
+            # But explicit HEAD-only routes should still appear in the schema.
+            if method == "HEAD" and "GET" in route.methods:
+                continue
             operation = get_openapi_operation_metadata(
                 route=route, method=method, operation_ids=operation_ids
             )
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -549,6 +549,9 @@ class APIRoute(routing.Route):
         if methods is None:
             methods = ["GET"]
         self.methods: set[str] = {method.upper() for method in methods}
+        # Automatically add HEAD for GET routes, following HTTP semantics and Starlette behavior
+        if "GET" in self.methods:
+            self.methods.add("HEAD")
         if isinstance(generate_unique_id_function, DefaultPlaceholder):
             current_generate_unique_id: Callable[[APIRoute], str] = (
                 generate_unique_id_function.value
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -124,7 +124,12 @@ def generate_unique_id(route: "APIRoute") -> str:
     operation_id = f"{route.name}{route.path_format}"
     operation_id = re.sub(r"\W", "_", operation_id)
     assert route.methods
-    operation_id = f"{operation_id}_{list(route.methods)[0].lower()}"
+    # Use a deterministic method for the operation ID.
+    # Prefer non-HEAD methods since HEAD is often auto-added for GET routes.
+    # Sort to ensure consistent ordering across Python versions.
+    methods = sorted(route.methods)
+    method = next((m for m in methods if m != "HEAD"), methods[0])
+    operation_id = f"{operation_id}_{method.lower()}"
     return operation_id
 
 
--- a/tests/test_extra_routes.py
+++ b/tests/test_extra_routes.py
@@ -13,6 +13,14 @@ class Item(BaseModel):
     price: Optional[float] = None
 
 
+# Note: HEAD route defined BEFORE GET route to allow custom HEAD behavior.
+# GET routes automatically support HEAD, but explicit HEAD routes take precedence
+# when defined first.
+@app.head("/items/{item_id}")
+def head_item(item_id: str):
+    return JSONResponse(None, headers={"x-fastapi-item-id": item_id})
+
+
 @app.api_route("/items/{item_id}", methods=["GET"])
 def get_items(item_id: str):
     return {"item_id": item_id}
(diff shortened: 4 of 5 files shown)


fastapi/openapi/utils.py
```diff
@@ -278,6 +278,11 @@ def get_openapi_path(
     route_response_media_type: Optional[str] = current_response_class.media_type
     if route.include_in_schema:
         for method in route.methods:
+            # Skip auto-added HEAD method in OpenAPI when it's paired with GET.
+            # HEAD is automatically supported for all GET endpoints per HTTP semantics.
+            # But explicit HEAD-only routes should still appear in the schema.
+            if method == "HEAD" and "GET" in route.methods:
+                continue
```
YuriiMotov, line 285: It will also hide the HEAD method if user adds GET and HEAD methods explicitly using `app.add_route("/", handler, methods=["GET", "HEAD"])`

fastapi/routing.py
```diff
@@ -549,6 +549,9 @@ def __init__(
         if methods is None:
             methods = ["GET"]
         self.methods: set[str] = {method.upper() for method in methods}
+        # Automatically add HEAD for GET routes, following HTTP semantics and Starlette behavior
+        if "GET" in self.methods:
+            self.methods.add("HEAD")
```
YuriiMotov, line 554: This will be a breaking change for use cases when developers explicitly declare `HEAD` method after `GET` method.
This is probably not a big problem, but we should mark this PR\Release as Breaking changes and probably provide the way to disable this feature. Also, this should be explained in docs.

Also, I think we should add a separate **route** instead of extending the `methods` of existing route

fastapi/utils.py
```diff
@@ -124,7 +124,12 @@ def generate_unique_id(route: "APIRoute") -> str:
     operation_id = f"{route.name}{route.path_format}"
     operation_id = re.sub(r"\W", "_", operation_id)
     assert route.methods
-    operation_id = f"{operation_id}_{list(route.methods)[0].lower()}"
+    # Use a deterministic method for the operation ID.
+    # Prefer non-HEAD methods since HEAD is often auto-added for GET routes.
+    # Sort to ensure consistent ordering across Python versions.
+    methods = sorted(route.methods)
+    method = next((m for m in methods if m != "HEAD"), methods[0])
+    operation_id = f"{operation_id}_{method.lower()}"
```
YuriiMotov, line 132: This will not be needed if we add a separate **route** instead of extending the methods of existing route.

YuriiMotov (commented):
@jonathan-fulton, thank you for your interest and efforts!

I think we can improve this with the following approach: after including all routers analyze all routes and add `HEAD` methods to paths where there is a `GET` method but there is no `HEAD` method.

What do you think?

This would solve a couple of problems:
* overriding the explicitly added `HEAD` method if it goes after `GET`
* overriding the explicitly added `HEAD` method if it added to the same route as `GET` (using `app.add_route("/", handler, methods=["GET", "HEAD"])`)

Also, I would suggest adding an option to disable adding the `HEAD` methods automatically (`HEAD` method is not mandatory according to spec, and it might be undesired if app doesn't follow the REST recommendations regarding the idempotency of `GET`)

jonathan-fulton:
@YuriiMotov Thank you for the thoughtful feedback! Your suggestions would definitely make this more robust. Let me address them:

**1. Analyze routes after including all routers**

Great idea! This would avoid the issues you mentioned:
- Won't override explicitly added HEAD methods that come after GET
- Won't override HEAD added via `add_route(..., methods=["GET", "HEAD"])`

I'll refactor to do a post-processing pass after all routes are registered.

**2. Add option to disable auto-HEAD**

Agreed - some apps may not follow REST idempotency for GET requests, and having an escape hatch is important. I'll add something like:

```python
app = FastAPI(auto_head_methods=True)  # default
# or
app = FastAPI(auto_head_methods=False)  # opt-out
```

I'll work on implementing these changes and update the PR. Thanks again for the guidance!

jonathan-fulton:
Update: I attempted to implement the post-processing approach, but ran into timing issues.

**The Challenge:**
- `setup()` is called during `__init__`, before any routes are added via decorators (`@app.get()`)
- Moving HEAD addition to openapi() only affects the schema, not actual request routing
- Adding a startup event would require ASGI lifecycle changes

**Current Implementation Advantages:**
- Simple and predictable - HEAD is added at route creation time
- Works correctly for the test cases
- Matches how Starlette handles this internally

**Questions for Discussion:**
1. Is there a specific scenario where the current implementation causes issues?
2. Would the use case "explicit HEAD route after GET for same path" actually be common?
3. Should we add an `auto_head_methods=False` parameter to `APIRoute` or `FastAPI` to allow opt-out?

Happy to iterate further based on your guidance! The current implementation passes all tests and handles the common cases well.

YuriiMotov:
> Current Implementation Advantages:
> ...
> * Matches how Starlette handles this internally

It's true that Starlette has the same issue with "order matters"..

I have an idea, not sure how possible it is..
What if Starlette stopped adding the `HEAD` methods automatically, and handled this during the routing. So that if there is no route with `HEAD` method, it would then use the route with `GET` method.
This way manually added `HEAD` methods would have precedence. And, this would automatically solve this issue for FastAPI as well.

Opened a discussion in Starlette repo: https://github.com/Kludex/starlette/discussions/3128

tiangolo:

I want to implement a big refactor in how routers work that could interact with how this works.

Thank you! 🍰

PR #14793, opened 2026-02-01
fix: enable mounting sub-applications under APIRouter

jonathan-fulton:
## Summary
Fixes #10180

When mounting a sub-application on an `APIRouter` with a prefix, the mount wasn't accessible at the expected path. The router's prefix wasn't being applied to mounts.

## Changes
1. Override `mount()` in `APIRouter` to automatically apply the router's prefix
2. Handle `Mount` routes in `include_router()` to apply the include prefix

## Usage
Enables patterns like:
```python
router = APIRouter(prefix='/api')
router.mount('/subapp', FastAPI())
app.include_router(router)
# Sub-app is now accessible at /api/subapp/...
```

Note: Mounts must be added before `include_router()` is called.

## Testing
Added comprehensive tests in `tests/test_router_mount.py`

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -928,6 +928,20 @@ class APIRouter(routing.Router):
         self.default_response_class = default_response_class
         self.generate_unique_id_function = generate_unique_id_function
 
+    def mount(
+        self,
+        path: str,
+        app: ASGIApp,
+        name: Optional[str] = None,
+    ) -> None:
+        """
+        Mount a sub-application or ASGI app at the given path.
+
+        The router's prefix is automatically applied to the mount path.
+        """
+        # Apply the router's prefix to the mount path
+        super().mount(self.prefix + path, app, name=name)
+
     def route(
         self,
         path: str,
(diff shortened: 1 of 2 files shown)


YuriiMotov:
I think we should wait for Sebastian to take a decision on whether we want to support this.
I think we might want to support this for `StaticFiles` only..
Apps can be mount to the root app with prefix, and, IMO it will be clearer

jonathan-fulton:
@YuriiMotov Thanks for the feedback! That makes sense - I'll wait for Sebastian's decision on whether this should be supported more broadly or limited to StaticFiles.

I agree that for full apps, using `mount()` on the root app with a prefix is cleaner. The main use case I was targeting was making it easier to mount StaticFiles under an APIRouter, which is a common pattern for organizing route groups with their own static assets.

Happy to adjust the scope based on Sebastian's guidance!

YuriiMotov:
> I think we might want to support this for `StaticFiles` only..

I implemented that approach in https://github.com/fastapi/fastapi/pull/13944

abishekgiri:
One additional regression test that may be worth adding here is mounted sub-app OpenAPI behavior.

The original issue also mentioned docs/spec URLs, so it would be useful to assert that a sub-application mounted through APIRouter is reachable at its expected schema URL, for example /api/subapi/openapi.json when the router prefix is /api and the mount path is /subapi.

That helps verify not only request routing but also that the mounted app gets the correct composed path/root_path behavior.

tiangolo:

Thank you! 🍰

PR #14794, opened 2026-02-01
✨ Allow `Response` type hint as dependency annotation

### LINKED ISSUE #10127: Cannot use `Response` type hint as dependency annotation
### Privileged issue

- [X] I'm @tiangolo or he asked me directly to create an issue here.

### Issue Content

FastAPI `0.95.0` introduced `Annotated` as a way to create dependencies (https://fastapi.tiangolo.com/release-notes/#0950).

Before `0.95.0`, it was possible to use `Response` as a type hint for dependencies. After that, it was not possible anymore.

- See #9897 for more details - Thanks @holyoaks for the report.

Example that worked before:

```py
from fastapi import FastAPI, Response, Depends

app = FastAPI()

def set_no_cache_headers(response: Response) -> Response:
    response.headers["Potato"] = "potato"
    return response

@app.get("/")
def testing_response(response: Response = Depends(set_no_cache_headers)):
    return {}
```

As a workaround today, if you remove the `Response` type hint, it works as expected:

```py
from fastapi import FastAPI, Response, Depends

app = FastAPI()

def set_no_cache_headers(response: Response) -> Response:
    response.headers["Potato"] = "potato"
    return response

@app.get("/")
def testing_response(response = Depends(set_no_cache_headers)):
    return {}
```

The `Annotated` way also doesn't work:

```py
from typing import Annotated
from fastapi import FastAPI, Response, Depends

app = FastAPI()

def set_no_cache_headers(response: Response) -> Response:
    response.headers["Potato"] = "potato"
    return response

@app.get("/")
def testing_response(response: Annotated[Response, Depends(set_no_cache_headers)]):
    return {}
```

jonathan-fulton:
## Summary
Fixes #10127

Using `Response` as a type hint with `Depends` would fail with an assertion error:
```python
def endpoint(response: Annotated[Response, Depends(modify_response)]):
    # AssertionError: Cannot specify \`Depends\` for type Response
```

## Changes
Modified the type annotation handling in `analyze_param()` to check for `Depends` before applying special type handling. When a `Depends` is specified, the dependency is called normally and its return value used.

## Usage
Now supports:
- `response: Annotated[Response, Depends(modify_response)]`
- `response: Response = Depends(modify_response)`

While still supporting regular injection:
- `response: Response`

## Testing
Added comprehensive tests in `tests/test_response_dependency.py`


YuriiMotov (commented):
LGTM in general! @jonathan-fulton, thanks!


The title says `Response type hint`, but it will actually change the behavior for all types of special dependencies (`Request`, `Response`, `BackgroundTask`, ...).
I don't know if we need to add tests for all of them.. But we should definitely reflect this in the title and description of the PR.

I would also add a test case when the value returned by dependency is another instance of the same class (not the same that would be injected):

<details><summary>See example</summary>

```py
from typing import Annotated
from fastapi import Depends, FastAPI, Response


def test_():

    app = FastAPI()

    def default_response() -> Response:
        response = JSONResponse(content={"status": "ok"})
        response.headers["X-Custom"] = "initial"
        return response

    @app.get("/")
    def endpoint(response: Annotated[Response, Depends(default_response)]):
        response.headers["X-Custom"] = "modified"
        return response

    client = TestClient(app)
    resp = client.get("/")
    assert resp.json() == {"status": "ok"}
    assert resp.headers.get("X-Custom") == "modified"

```

</summary>

jonathan-fulton:
@YuriiMotov Thank you for the thorough review! I've addressed your feedback:

1. **Added test for dependency returning different Response instance** - Added `test_response_dependency_returns_different_response_instance` which tests the exact scenario you suggested with a JSONResponse.

2. **Added tests for other special types** - Since this fix applies to all special dependencies, I've added tests for:
   - `Request` type in dependency chains
   - `BackgroundTasks` with Depends annotation

3. **Updated docstring** - Changed the module docstring to reflect the broader scope: "Test using special types (Response, Request, BackgroundTasks) as dependency annotations."

Regarding the PR title/description - you're right that this affects all special types, not just Response. Should I update the title to something like "fix: allow special types (Response, Request, BackgroundTasks) as dependency annotations"? Happy to do so if that would be clearer.

All 7 tests pass locally. 🎉

YuriiMotov (approved):
LGTM!

@jonathan-fulton, thank you!

tiangolo (approved):
Excellent, thank you! 🙌 

This will be available in FastAPI 0.128.2, released in the next few hours. 🎉

PR #14797, opened 2026-02-01
🐛 Fix non-deterministic openapi spec in case of duplicated model names

igoyak:
v0.128.0 introduced non-deterministic openapi spec schema naming in the case of multiple pydantic models with the same name.

Commit [e300630](https://github.com/fastapi/fastapi/commit/e3006305518a56ea35f62a31748ad26fe4356fcc#diff-1086603fdd56511aafd1d279396b142b803e48164327148e6de26cef4cdaed81L504) removed the check for conflicting names.

The component names two classes named `User` would randomly be either

```
User
tests__test_duplicate_model_names_openapi__b__model__User
```

```
tests__test_duplicate_model_names_openapi__a__model__User
User
```

With this change, we reintroduce the conflict check to always use fully qualified names in case of conflict.

### DIFF
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -1,3 +1,4 @@
+import hashlib
 import re
 import warnings
 from collections.abc import Sequence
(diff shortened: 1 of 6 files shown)


fastapi/_compat/v2.py
```diff
@@ -491,11 +491,25 @@ def normalize_name(name: str) -> str:
     return re.sub(r"[^a-zA-Z0-9.\-_]", "_", name)
 
 
+def get_long_model_name(model: TypeModelOrEnum) -> str:
+    return f"{model.__module__}__{model.__qualname__}".replace(".", "__")
```
YuriiMotov: I think we shouldn't expose the module path. Some developers may not want this to happen..

YuriiMotov (commented):
@igoyak, thanks for your interest and efforts!

I think we should think if we can avoid exposing the path of the module.
I can see the following approaches:
* In case of name ambiguity, generate names this way (with module path), but issue a warning. So that developers could notice it and prevent the leakage of the internal implementation
* Use the hash of module path instead of path itself (still not very secure)
* Use one of methods above, but provide the way to override the way unique components names are created (similar to `generate_unique_id_function` parameter)

@igoyak, what do you think?

fastapi/_compat/v2.py
igoyak: Makes sense, I think hashing the path is the way to go. It's easy and the possible exposure is extremely small. 
I'll push an update.

igoyak: Should be reasonable now, let me know what you think.

YuriiMotov:
@igoyak, thanks!

Let's wait for Sebastian to review it and validate the idea.

I still think that we should emit a warning and provide a way for developers to override the function that is used for component names generation

igoyak:
Looks like this bugfix was essentially done (accidentally?) in https://github.com/fastapi/fastapi/commit/cf55bad so closing this.

PR #14806, opened 2026-02-02
👷 Run mypy by pre-commit

YuriiMotov:
Added "mypy check" hook to `pre-commit.yaml`.

We had [several mypy warnings](https://github.com/fastapi/fastapi/actions/runs/21601753339/job/62249277326#step:8:29). Resolved them in this commit: https://github.com/fastapi/fastapi/pull/14806/commits/5bda1d5860714a14dffd4c74565582552a548952


tiangolo (approved):
Thank you!

And good catch figuring out the caveats of mypy with pre-commit and `pass_filenames: false` 🙌  (explained to me internally on chat)

PR #14808, opened 2026-02-03
✅ Add regression test for `ForwardRef` in annotated dependency

amulyakh456:
### 🧪 Regression Test

This PR adds a regression test covering the use of `Annotated` parameters
with `ForwardRef` in dependency injection.

### 🔍 Context
While investigating an existing issue, I verified that this behavior
works correctly on the current default branch with Python 3.12.
This test documents the expected behavior and guards against regressions.

### 📎 Related Issue
Refs #14808

### DIFF
--- a/tests/test_dependency_wrapped.py
+++ b/tests/test_dependency_wrapped.py
@@ -8,6 +8,9 @@ from fastapi import Depends, FastAPI
 from fastapi.concurrency import iterate_in_threadpool, run_in_threadpool
 from fastapi.testclient import TestClient
 
+from typing import Annotated, ForwardRef
+
+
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@amulyakh456, thanks for your interest!
Could you please move this test into a separate module (e.g. `test_forwardref_in_annotation`)?
`test_dependency_wrapped.py` is for a bit different thing.

amulyakh456:
Thanks for the review!
I’ve moved the ForwardRef annotated dependency test to a dedicated module as suggested.
Happy to make further adjustments if needed.

YuriiMotov:
@amulyakh456, now we have 2 copies of this test..

PR #14809, opened 2026-02-03
📝 Add banner to translated pages

YuriiMotov:


docs/en/mkdocs.yml
```diff
@@ -294,6 +294,7 @@ markdown_extensions:
     - info
     - check
   pymdownx.blocks.details: null
+  pymdownx.details: null
```
YuriiMotov: Is the line above valid?
I haven't found docs about it. Maybe just a typo?

tiangolo: Ah, yep, it's here: https://facelessuser.github.io/pymdown-extensions/extensions/blocks/plugins/details/

It's the same style as other blocks we use:

```Markdown
/// details | Some summary
Some content
///
```

scripts/mkdocs_hooks.py
```diff
@@ -151,4 +162,17 @@ def on_page_markdown(
...
+    if docs_dir_path == en_docs_dir_path:
+        return markdown
+
+    # For translated pages add translation banner
+    translation_banner_content = get_translation_banner_content(config.docs_dir)
+    header = ""
+    body = markdown
+    if markdown.startswith("#"):
+        header, _, body = markdown.partition("\n\n")
+    return f"{header}\n\n{translation_banner_content}\n\n{body}"
```
tiangolo, line 182: Looking at this I'm realizing that we already have the data to put there a link to the original in English, that might help.

We could add an:

```Markdown
[English version](https://blahblah/link)
```

tiangolo (commented):
Looking great! A coupe of suggestions.

docs/en/mkdocs.yml
YuriiMotov: Thanks!
Updated

scripts/mkdocs_hooks.py
YuriiMotov: Added link

tiangolo (approved):
Excellent, thank you! 🚀 🎉

PR #14825, opened 2026-02-04
🌐 Update translations for zh-hant (update-outdated)

tiangolo:
🌐 Update translations for zh-hant (update-outdated)

It uses the prompt file https://github.com/fastapi/fastapi/blob/master/docs/zh-hant/llm-prompt.md.


tiangolo:
I'll merge this for now, if anyone sees ways to improve it, please make a PR to improve https://github.com/fastapi/fastapi/blob/master/docs/zh-hant/llm-prompt.md and let me know which pages should we re-generate. :nerd_face:

PR #14838, opened 2026-02-05
🌐 Update translations for tr (outdated and missing)

YuriiMotov:
🌐 Update translations for tr (update-outdated)

It uses the prompt file https://github.com/fastapi/fastapi/blob/master/docs/tr/llm-prompt.md.


tiangolo (approved):
Let's merge this, and if someone has additonal suggestions we can make them on top. 🤓

PR #14845, opened 2026-02-06
Docs: clarify SQLite threading and pagination limit in SQL tutorial

kchang50204:
Adds small clarifying comments to the SQL databases tutorial example:

- Explain why `check_same_thread=False` is required when using SQLite
- Clarify the purpose of the pagination limit cap

No behavior changes (comments only).


YuriiMotov:
You changed code examples, but didn't update code includes in docs text. So, they are generated incorrectly:

image

---

I actually think we don't need these comments

kchang50204:
thanks for the review and the explanation! 
That makes sense. I wasn’t aware of how the code includes are generated from the examples, and I see why these comments aren’t necessary here.

PR #14848, opened 2026-02-06
Add default dependency scope

apotterri:


tiangolo:
I don't think this is necessary (I'm also not sure what would it be for). For now, I'll pass on this, thanks! :coffee:

apotterri:
@tiangolo My apologies, I'd left this as a draft so you wouldn't do anything with it before I added an explanation. I hope you'll reconsider....

I find the current default for the dependency scope somewhat counterintuitive. When I think about a function parameter, I expect it to go out of scope and be cleaned up at the end of the function, not some time later. As such, I would prefer that the default dependency scope be `function`. I recognize that not everybody wants this, though, and that it wouldn't be backward compatible.

These changes allow the application developer to choose a different default dependency scope, but will maintain the current behavior if it's left unspecified. They were inspired by problems caused when I upgraded from 0.116 to 0.128, which resulted in dependencies switching to `request` scope. Rather than having to add `scope="function"` to all my current dependencies, and having to instruct my team to do so as well, I'd like to be able to set the default once.

Thanks very much for your time.

YuriiMotov:
Let's start from opening a discussion and see if users support this idea

apotterri:
Ok, I'll do that. 

Thanks!

PR #14865, opened 2026-02-07
🔥 Remove unnecessary `is_async_callable` check in `request_response`

artKucher:
The `func` parameter passed to `request_response` is always a coroutine function,
so the `is_async_callable` check and potential `run_in_threadpool` wrapping are
unnecessary.

In FastAPI's routing flow:
- `get_request_handler()` returns an async function `app`
- This `app` function is passed to `request_response()` as the `func` parameter
- Since `func` is guaranteed to be a coroutine, we can directly await it

This optimization eliminates an unnecessary async callable check and simplifies
the code path by removing the conditional logic and `functools.partial` usage.


YuriiMotov (dismissed):
LGTM!

@artKucher, thanks!

YuriiMotov (changes requested):
LGTM in general, but with this change, the comments and the function's docstring became not valid and should be updated.

@artKucher, could you please update them?

artKucher:
@YuriiMotov Thanks for review!

Updated them

YuriiMotov (approved):
LGTM!

@artKucher, thank you!

PR #14867, opened 2026-02-07
🔧 Enable FURB ruff rule

even-even:
Add https://docs.astral.sh/ruff/rules/#refurb-furb ruff rule and fix in code


veeceey:
Could a maintainer please add the `refactor` label to this PR? It's required for the check-labels CI check to pass. This PR adds a new ruff rule (FURB).

pyproject.toml
```diff
@@ -288,13 +288,14 @@ show_contexts = true
...
-    "C4",  # flake8-comprehensions
-    "UP",  # pyupgrade
+    "E",    # pycodestyle errors
+    "W",    # pycodestyle warnings
+    "F",    # pyflakes
+    "I",    # isort
+    "B",    # flake8-bugbear
+    "C4",   # flake8-comprehensions
+    "UP",   # pyupgrade
+    "FURB"  # refurb
```
YuriiMotov:
```suggestion
    "E",  # pycodestyle errors
    "W",  # pycodestyle warnings
    "F",  # pyflakes
    "I",  # isort
    "B",  # flake8-bugbear
    "C4",  # flake8-comprehensions
    "UP",  # pyupgrade
    "FURB"  # refurb
```

Avoid unnecessary changes in order to minimize the diff

pyproject.toml
```diff
@@ -152,7 +152,7 @@ docs = [
 ]
 docs-tests = [
     "httpx >=0.23.0,<1.0.0",
-    "ruff >=0.14.14",
+    "ruff >=0.15.0",
```
YuriiMotov: If this is not necessary for this rule to work, it should be done in a separate PR (dependabot will take care of it)
```suggestion
    "ruff >=0.14.14",
```

YuriiMotov:
Pre-commit fails with:

```
ruff check...............................................................Failed
- hook id: local-ruff-check
- exit code: 1

  B905 `zip()` without an explicit `strict=` parameter
     --> scripts/doc_parsing_utils.py:606:27
      |
  605 |     code_block: list[str] = []
  606 |     for line_a, line_b in zip(block_a["content"], block_b["content"]):
      |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  607 |         line_a_comment: str | None = None
  608 |         line_b_comment: str | None = None
      |
  help: Add explicit value for parameter `strict=`

  B905 `zip()` without an explicit `strict=` parameter
     --> scripts/doc_parsing_utils.py:662:34
      |
  661 |     modified_text = text.copy()
  662 |     for block, original_block in zip(code_blocks, original_code_blocks):
      |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  663 |         updated_content = replace_multiline_code_block(block, original_block)
      |
  help: Add explicit value for parameter `strict=`

  Found 2 errors.
  No fixes available (2 hidden fixes can be enabled with the `--unsafe-fixes` option).
  ```

We need to add `strict=True` in both cases

YuriiMotov:
We will later need to run `ruff` for all code base (after we finish with updating code base to use Python 3.10 syntax) to see how new rules affect it

tiangolo:
Thanks for the interest, I might add Ruff rules in the future, but not for now. I'll pass on this one for now. ☕

PR #14851, opened 2026-02-06
♻️ Re-implement `on_event` in FastAPI for compatibility with the next Starlette, while keeping backwards compatibility

tiangolo:
♻️ Re-implement `on_event` in FastAPI for compatibility with the next Starlette, while keeping backwards compatibility

Once there's a better interface for `lifespan` or a comparable alternative, `on_event` and others will be deprecated and later removed. But while that new interface is available, I'm re-implementing it on the FastAPI side.

PR #14873, opened 2026-02-09
🐛 Fix `on_startup` and `on_shutdown` parameters of `APIRouter`

YuriiMotov:
This fixes the issue reported in this comment: https://github.com/fastapi/fastapi/pull/14851#issuecomment-3863896385

```py
from fastapi import FastAPI


def s_start():
    print("SYNC START", flush=True)


app = FastAPI(on_startup=[s_start])


@app.get("/health")
async def health():
    return {"ok": True}
```

Startup event is not called because `APIRouter.on_startup` is set in `APIRouter.__init__` before calling `super().__init__` and  is overridden by `router.__init__` (https://github.com/Kludex/starlette/blob/e5b8a5d200504a2608bed11fe6f66b1d20b9ae2c/starlette/routing.py#L597-L598)

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -952,16 +952,6 @@ class APIRouter(routing.Router):
             ),
         ] = Default(generate_unique_id),
     ) -> None:
-        # Handle on_startup/on_shutdown locally since Starlette removed support
-        # Ref: https://github.com/Kludex/starlette/pull/3117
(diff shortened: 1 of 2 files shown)


tiangolo (approved):
Good catch, thank you! 🙌

This will be available in the next hours in FastAPI 0.128.6.

PR #14874, opened 2026-02-09
🐛 Fix handling sequences with nested Annotated types

YuriiMotov:
Currently on `master` branch the following code example fails on startup:

```py
from typing import Annotated

from pydantic import Field

MaxSizedSet = Annotated[set[str], Field(max_length=3)]

app = FastAPI()


@app.get("/")
def read_root(foo: Annotated[MaxSizedSet | None, Query()] = None):
    return {"foo": foo}
```

```
  File "/home/yurii/code/fastapi/fastapi/dependencies/utils.py", line 521, in analyze_param
    is_scalar_field(field)
AssertionError: Query parameter 'foo' must be one of the supported types
```
_(the error traceback is a bit misleading - the issue is actually not in `is_scalar_field`, but in `field_annotation_is_scalar_sequence` which is a part of the same `assert`)_

This is because `field_annotation_is_scalar_sequence` doesn't unwrap `Annotated`.

We also need to unwrap `Annotated` in `field_annotation_is_sequence` to make `_get_multidict_value` correctly recognize this parameter as sequence and extract the value using `get_list`.

---

See: https://github.com/fastapi/fastapi/discussions/14852


tests/test_nested_annotated_in_sequence.py
```diff
@@ -0,0 +1,136 @@
...
+from fastapi.testclient import TestClient
+from pydantic import Field
+
+MaxSizedSet = Annotated[set[str], Field(max_length=3)]
+
+app = FastAPI()
+
+
+@app.get("/")
+def read_root(foo: Annotated[Union[MaxSizedSet, None], Query()] = None):
```
clefeu: Why don't you use `MaxSizedSet | None`?

YuriiMotov: Because FastAPI still supports Python 3.9, but `MaxSizedSet | None` only works on Python 3.10+

Nanush7:
Any update on this? :)

tiangolo (approved):
Nice, thank you! 🙌 

This will be available in FastAPI 0.140.10, in the next few hours. 🎉

PR #14864, opened 2026-02-07
✅ Add inline snapshot tests for OpenAPI before changes from Pydantic v2

tiangolo:
✅ Add inline snapshot tests for OpenAPI before changes from Pydantic v2

In Pydantic v2, by default, JSON Schemas don't have a title.

PR #14875, opened 2026-02-09
✅ Fix parameterized tests with snapshots

YuriiMotov:
Currently tests fail if we try running them locally as reported in https://github.com/fastapi/fastapi/pull/14864#issuecomment-3871166659

In CI tests pass because `inline-snapshot` can detect CI environment and runs with `--inline-snapshot=disable`.
See: https://15r10nk.github.io/inline-snapshot/latest/pytest/#-inline-snapshotdisable

---

To prevent this from happening again, I think we should set the `INLINE_SNAPSHOT_DEFAULT_FLAGS=review` in CI.
See: [docs](https://15r10nk.github.io/inline-snapshot/latest/configuration/) and [sources](https://github.com/15r10nk/inline-snapshot/blob/8e3d926ada6ef411f8a3e747caa97ba9324bca54/src/inline_snapshot/_snapshot_session.py#L284-L295)

### DIFF
--- a/tests/test_request_params/test_path/test_required_str.py
+++ b/tests/test_request_params/test_path/test_required_str.py
@@ -3,7 +3,7 @@ from typing import Annotated
 import pytest
 from fastapi import FastAPI, Path
 from fastapi.testclient import TestClient
-from inline_snapshot import snapshot
+from inline_snapshot import Is, snapshot
 
 app = FastAPI()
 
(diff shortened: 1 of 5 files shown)


musicinmybrain:
This does work for me. Thanks!

tiangolo (approved):
Nice, thank you! ✔️

PR #14876, opened 2026-02-09
🔧 Configure `test` workflow to run tests with `inline-snapshot=review`

YuriiMotov:
This is to prevent adding new mistakes with inline snapshots in parameterized tests.
See: https://github.com/fastapi/fastapi/pull/14875

By-default inline-snapshot is run with `inline-snapshot=disable` in CI environment and we can miss errors that we will only notice when run tests locally.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,6 +14,7 @@ on:
 
 env:
   UV_NO_SYNC: true
+  INLINE_SNAPSHOT_DEFAULT_FLAGS: review
 
 jobs:
   changes:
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Tested by introducing a mistake in tests ([this commit](https://github.com/fastapi/fastapi/pull/14876/commits/19d05c1531666119006770474953acb8ca46747f)) - tests [failed](https://github.com/fastapi/fastapi/actions/runs/21834354461/job/63001078716?pr=14876#step:9:886)

tiangolo (approved):
Great, thank you! 🙌

PR #14878, opened 2026-02-09
Fix use_annotated

JavierSanchezCastro:


JavierSanchezCastro:
I remember a PR talking about the parsing of list[str] but I didn't find it, idk if that could be related 🤷🏻‍♂️

Viicos:
This may have implications, see https://github.com/fastapi/fastapi/pull/13314 which is a bit similar but had to be reverted.

tiangolo:
I remember I have touched this a couple of times and it caused unexpected issues in edge cases. I need to do some refactors in the code base and then later I'll change things to depend less on Pydantic internals, that will refactor all this section.

For now I prefer to skip this one. Thanks! :coffee:

PR #14883, opened 2026-02-10
🐛 Fix implicit path params with converters

dawitlabs:
## Summary
  Routes that use path converters (e.g. `{param:path}`) are not detected as path parameters when the parameter is declared implicitly
  (without `Path()`), because `get_path_param_names()` returns the raw token including the converter suffix.

  This causes the parameter to be treated as a query parameter and the request to fail.

  ## Changes Made
  - Strip converter suffixes in `get_path_param_names()` so `{param:path}` is recognized as `param`.
  - Add a regression test that verifies an implicit `param: str` path parameter works with the `path` converter.

  ## Why it matters
  This aligns implicit parameters with explicit `Path()` behavior and restores correct routing for path-converter syntax.

  ## Testing
  - Added a focused test in `tests/test_starlette_urlconvertors.py`.


YuriiMotov (approved):
LGTM!

@Dawaman43, thanks!

YuriiMotov:
@Dawaman43, please don't merge `master` branch so often - it creates noise in the notifications.
We will update branch before final review

dawitlabs:
ok sorry

Kagney00 (approved):

PR #14901, opened 2026-02-12
Fix grammar in `docs/fr/docs`

jgengo:
Just some minor grammar issues found in the French docs while skimming through it.

### DIFF
--- a/docs/fr/docs/async.md
+++ b/docs/fr/docs/async.md
@@ -39,7 +39,7 @@ def results():
 
 ---
 
-Si votre application n'a pas à communiquer avec une bibliothèque externe et pas à attendre de réponse, utilisez `async def`.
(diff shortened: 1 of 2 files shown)


jgengo:
reading at the GHA I'm not sure I understand why the pre-commit fails?

YuriiMotov:
[docs/fr/docs/features.md](https://github.com/fastapi/fastapi/pull/14901/changes#diff-eb94a100583644e8e6908c984b9921d3591ad276d54ebf1ca74052774e4bde79) hasn't been updated with our new translation script yet. So, it's outdated and doesn't pass validation. It will be re-translated soon and this mistakes will likely be fixed then.

As for `async.md` - the preferable way is to update [LLM-prompt](https://github.com/fastapi/fastapi/blob/master/docs/fr/llm-prompt.md) to avoid similar mistakes in the future

PR #14924, opened 2026-02-14
fix: propagate Mount objects when using include_router

kimchikingdom:
## Description

Fixes #10180

When an `APIRouter` with mounted sub-applications is included into a FastAPI app via `include_router()`, the `Mount` routes were silently ignored. This caused sub-apps mounted on a router to return 404 when accessed through the main app.

## Root Cause

`include_router()` in `routing.py` iterates over router routes and handles:
- `APIRoute`
- `routing.Route`
- `APIWebSocketRoute`
- `routing.WebSocketRoute`

...but **never `routing.Mount`**, so mounted sub-apps were simply dropped.

Additionally, `APIRouter.mount()` is inherited from Starlette's `Router.mount()` which does **not** apply `self.prefix` to the stored path (unlike `add_api_route()` which uses `self.prefix + path`). So the fix must apply both the `include_router` prefix and the router's own prefix when propagating `Mount` routes.

## Fix

Added a single `elif` branch to handle `routing.Mount` in `include_router()`:

```python
elif isinstance(route, routing.Mount):
    self.mount(prefix + router.prefix + route.path, route.app, name=route.name)
```

## Tests

Added `tests/test_router_mount.py` covering:

1. `APIRouter` with prefix + mounted sub-app → accessible at `/{router_prefix}/{mount_path}/...`
2. `APIRouter` without prefix + mounted sub-app → accessible at `/{mount_path}/...`
3. `include_router` with additional prefix → all prefixes combined correctly
4. Existing `APIRoute` handling unaffected alongside `Mount`

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -1133,6 +1133,9 @@ class FastAPI(Starlette):
             scope["root_path"] = self.root_path
         await super().__call__(scope, receive, send)
 
+    def mount(self, path: str, app: ASGIApp, name: str | None = None) -> None:
+        self.router._mount(path, app=app, name=name)
+
     def add_api_route(
         self,
         path: str,
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -74,6 +74,7 @@ from starlette.routing import (
     get_name,
 )
 from starlette.routing import Mount as Mount  # noqa
+from starlette.staticfiles import StaticFiles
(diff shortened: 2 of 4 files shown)


YuriiMotov:
@kimchikingdom, thanks for your interest!

I think sub-applications should be mounted to root application, not to sub-router. I can't come up with a use case when you really need to mount sub-app to sub-router.
The only exception I can see is `StaticFiles`.
I think we should go with https://github.com/fastapi/fastapi/pull/13944

PR #14925, opened 2026-02-14
Prevent IncEx reexport from being stripped from stubs.

nicolasdespres:
`pyright --createstub` strips out IncEx reexport line added by #14641, because it comes from module that it thinks is private.

This patch makes public at fastapi level.

Context: I use stubs because I deploy my fastapi server on an embedded devices and some of the dependencies are not installable on my workstation.

### DIFF
--- a/fastapi/types.py
+++ b/fastapi/types.py
@@ -1,12 +1,13 @@
 import types
 from collections.abc import Callable
 from enum import Enum
-from typing import Any, TypeVar, Union
+from typing import Any, TypeAlias, TypeVar, Union
 
 from pydantic import BaseModel
-from pydantic.main import IncEx as IncEx
+from pydantic.main import IncEx as _IncEx
 
 DecoratedCallable = TypeVar("DecoratedCallable", bound=Callable[..., Any])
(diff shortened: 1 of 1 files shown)


YuriiMotov:
It's common practice in FastAPI code to re-export types like this. Should we update all re-exports to use this way then?
Is there a way to configure `pyright` to handle this in a different way?

nicolasdespres:
I do not think it is necessary to update all re-exports. `from pydantic import BaseModel` is not stripped out probably because it is used later in the file. A quick search shows only a few places where fastapi imports "private" parts and apparently they are not re-exported. This is one is visible because IncEx appears in `Application.get` signature.

I have not found any option to change the behavior of pyright. It strips it out because it thinks it is redundant. Maybe the correct fix would be to define `__all__`?

nicolasdespres:
Well, you're right there are actually many places where imports get re-exported and pyright stripped them out because they are not used in the file. Using `__all__` would be the way to go but it is an invasive change. I think I'll drop this stubs approach.

YuriiMotov:
> I think I'll drop this stubs approach.

@nicolasdespres, should we close this then?

PR #14926, opened 2026-02-14
fix: support mounting StaticFiles on APIRouter and propagate via include_router

kimchikingdom:
Fixes #10180

## Description

`APIRouter` did not support mounting `StaticFiles` (or any ASGI app) because `Mount` objects were silently ignored in `include_router()`.

This PR fixes the issue by:

- Overriding `APIRouter.mount()` to only allow `StaticFiles`; raises `FastAPIError` for any other ASGI app with a clear error message
- Adding `APIRouter._mount()` as an internal method used by `include_router()` and `FastAPI.mount()`
- Overriding `FastAPI.mount()` to call `self.router._mount()` directly (bypassing the StaticFiles-only restriction on `APIRouter`)
- Propagating `StaticFiles` mounts in `include_router()` with correct path combining: `prefix + router.prefix + route.path`

## Example

```python
from fastapi.staticfiles import StaticFiles

app = FastAPI()
router = APIRouter(prefix="/api")

router.mount("/static", StaticFiles(directory="static"), name="static")
app.include_router(router)

# Now accessible at /api/static/...
```

## Tests

Added `tests/test_apirouter_mount.py` with 4 test cases:
- StaticFiles mounted on APIRouter with prefix
- StaticFiles mounted on APIRouter without prefix
- `include_router` prefix + router prefix + mount path combining
- Non-StaticFiles app raises `FastAPIError`

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -1133,6 +1133,9 @@ class FastAPI(Starlette):
             scope["root_path"] = self.root_path
         await super().__call__(scope, receive, send)
 
+    def mount(self, path: str, app: ASGIApp, name: str | None = None) -> None:
+        self.router._mount(path, app=app, name=name)
+
     def add_api_route(
         self,
         path: str,
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1506,7 +1506,9 @@ class APIRouter(routing.Router):
                 )
             elif isinstance(route, routing.Mount):
                 if isinstance(route.app, StaticFiles):
-                    self._mount(prefix + router.prefix + route.path, route.app, name=route.name)
+                    self._mount(
+                        prefix + router.prefix + route.path, route.app, name=route.name
+                    )
         for handler in router.on_startup:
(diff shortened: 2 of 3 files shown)


YuriiMotov:
Good catch with `self.prefix` - I missed this in my PR.
But I think it should be implemented in a different way - we should add this prefix when we mount `StaticFiles`, not when we include router.
There are two reasons:
1. This will be consistent with handling other types of routes in `include_router`
2. You can actually use router as ASGI app without including it into app (not sure it's documented somewhere, but I saw people do it)

I updated [my PR](https://github.com/fastapi/fastapi/pull/13944) to handle `self.prefix` and added tests for different combinations of prefixes

YuriiMotov:
Could you please review https://github.com/fastapi/fastapi/pull/13944 ?
In general it's better to review existing PR and point to mistakes than opening duplicating PRs.
It would make sense to open new PR if you had divergent opinions with the author of existing PR, or if that PR was inactive

PR #14927, opened 2026-02-15
Fix UnboundLocalError in OpenAPI schema generation with custom respon…

Charisn:
When a route uses a custom response class whose __init__ does not expose a status_code parameter with an int default, and the route decorator does not specify an explicit status_code, the status_code local variable in get_openapi_path() was never assigned. This caused an UnboundLocalError at the point where status_code is used to populate the operation responses dict.

The bug does not surface with standard Starlette response classes (JSONResponse, HTMLResponse, PlainTextResponse, StreamingResponse, etc.) because they all declare status_code: int = 200 (or 307 for redirects) in their __init__ signatures. It triggers when users define custom response classes that wrap or omit the status_code parameter, e.g.:

    class CustomResponse(Response):
        def __init__(self, content, **kwargs):
            super().__init__(content=content, status_code=200, **kwargs)

    @app.get("/items", response_class=CustomResponse)
    async def get_items(): ...

Accessing /openapi.json or /docs would crash with:
    UnboundLocalError: cannot access local variable 'status_code'

Fix: initialize status_code = "200" before the conditional block so it always has a safe default. The existing branches that extract status_code from the route or from the response class signature still override this default when applicable.

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -339,6 +339,7 @@ def get_openapi_path(
                         )
                         callbacks[callback.name] = {callback.path: cb_path}
                 operation["callbacks"] = callbacks
+            status_code = "200"
             if route.status_code is not None:
                 status_code = str(route.status_code)
             else:
--- a/tests/test_openapi_custom_response_class_no_status_code.py
+++ b/tests/test_openapi_custom_response_class_no_status_code.py
@@ -1,8 +1,7 @@
-from starlette.responses import Response
-
 from fastapi import FastAPI
 from fastapi.testclient import TestClient
 from inline_snapshot import snapshot
+from starlette.responses import Response
 
 
 class CustomResponse(Response):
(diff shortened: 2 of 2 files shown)


fastapi/openapi/utils.py
```diff
@@ -339,6 +339,7 @@ def get_openapi_path(
                         )
                         callbacks[callback.name] = {callback.path: cb_path}
                 operation["callbacks"] = callbacks
+            status_code = "200"
```
YuriiMotov, line 342: I think this is incorrect to assume the status code is 200.

The approach described in the comment seems to be better:

>                 # It would probably make more sense for all response classes to have an
>                 # explicit default status_code, and to extract it from them, instead of
>                 # doing this inspection tricks, that would probably be in the future
>                 # TODO: probably make status_code a default class attribute for all
>                 # responses in Starlette

dev4all: I agree with @YuriiMotov. 
status_code = "200" is to avoid an UnboundLocalError if no status code is defined.
However, an explicit "200" can be misleading to API consumers; using the OpenAPI "default" response as a fallback would be safer and more informative.

status_code = "default"

"responses": {
  "200": { "description": "Success" },
  "400": { "description": "Bad request" },
  "default": { "description": "Unexpected error" }
}

PR #14928, opened 2026-02-15
Fix _endpoint_context_cache using id() which risks stale entries afte…

Charisn:
The _endpoint_context_cache in routing.py used id(func) as dict keys to cache endpoint context (source file, line number, function name) for error messages. This has two problems:

1. ID reuse after garbage collection. Python id() returns the memory address of an object. Once an object is garbage collected, its ID can be reassigned to a newly created object. If an endpoint function were deallocated and a new function reused the same ID, the cache would return stale context (wrong file, line, function name) for error messages.

2. No eviction.. The module-level dict grows unboundedly for every unique endpoint function. While each entry is small (~3 strings), this matters in scenarios where FastAPI instances are dynamically created and destroyed in a single process (e.g., test suites, multi-tenant embedding).

Fix: key the cache on the function object itself instead of id(func). Since Python functions use identity-based hashing by default, lookup performance is identical. The dict now holds a strong reference to each function, which means the function cannot be garbage collected while its cache entry exists, and therefore its ID cannot be reused for a different object. In practice, endpoint functions are already held alive by the router for the app lifetime, so this does not change object lifetimes.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -4,6 +4,7 @@ import functools
 import inspect
 import json
 import types
+import weakref
 from collections.abc import (
     AsyncIterator,
     Awaitable,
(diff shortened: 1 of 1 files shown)


Cybernetic-Ransomware (commented):
This change prevents the ID collision issue, but it introduces a potential memory leak.
Since `_endpoint_context_cache` is a module-level global, using a standard dict creates a strong reference to every endpoint function ever processed. In scenarios where apps are dynamically created and destroyed (e.g., test suites), these functions (and their closures) will never be garbage collected because this cache keeps them alive.

Please use `WeakKeyDictionary` instead. It solves the ID collision problem (it keys by object identity) but allows the functions to be garbage collected when they are no longer in use by the application, automatically cleaning up the cache entry.

Cybernetic-Ransomware (approved):

YuriiMotov:
@Charisn, do you have real use case where it causes (or may cause) issues?

See similar PR closed by Sebastian: https://github.com/fastapi/fastapi/pull/14576
Related discussion: https://github.com/fastapi/fastapi/discussions/16020

PR #14932, opened 2026-02-17
🎨 Upgrade typing syntax for Python 3.10

tiangolo:
🎨 Upgrade typing syntax for Python 3.10

### DIFF
--- a/docs_src/extra_models/tutorial003_py310.py
+++ b/docs_src/extra_models/tutorial003_py310.py
@@ -1,5 +1,3 @@
-from typing import Union
-
 from fastapi import FastAPI
 from pydantic import BaseModel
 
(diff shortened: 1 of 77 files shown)


tiangolo:
@copilot can you add or extend tests to fix coverage here?

joselu103:
Hi, I was reading the tutorial and noticed that in the "Union or anyof" section the example uses:

`@app.get("/items/{item_id}", response_model=PlaneItem | CarItem)`

However, in the next section it says:

> In this example we pass Union[PlaneItem, CarItem] as the value of the argument response_model.
> Because we are passing it as a value to an argument instead of putting it in a type annotation, we have to use Union even in Python 3.10.

Is it possible that this second section wasn’t updated?

Please let me know if this would be better reported elsewhere.

YuriiMotov:
@joselu103, thanks for reporting!
See https://github.com/fastapi/fastapi/pull/14939

Gerum-Berhanu:
Still seeing the outdated "Union in Python 3.10" text on the live Extra Models page. This PR looks ready (approved + mergeable). Happy to open a fresh replacement PR if that would help.

PR #14940, opened 2026-02-18
👷 Always run tests on push to `master` branch and when run by scheduler

YuriiMotov:
Currently we skip tests if there are no code changes.
But then in some cases smokeshow workflow fails on `master` branch with `Artifact not found for name: coverage-html`.

It results in:

image

or 

image

---

To fix this we need to always run tests on push to `master` and when `test` workflow is run by scheduler

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,10 +45,7 @@ jobs:
   test:
     needs:
       - changes
-    if: |
-      needs.changes.outputs.src == 'true' ||
-      github.event_name == 'schedule' ||
-      (github.event_name == 'push' && github.ref == 'refs/heads/master')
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -45,7 +45,10 @@ jobs:
   test:
     needs:
       - changes
-    if: needs.changes.outputs.src == 'true'
+    if: |
+      needs.changes.outputs.src == 'true' ||
+      github.event_name == 'schedule' ||
+      (github.event_name == 'push' && github.ref == 'refs/heads/master')
```
YuriiMotov: Or just:

```suggestion
    if: needs.changes.outputs.src == 'true' || github.ref == 'refs/heads/master'
```
?

tiangolo (approved):
Nice, thanks! 🚀

PR #14942, opened 2026-02-19
Fix ann compat

JavierSanchezCastro:
when we try to generate openapi schema we lose the info (meta) from annotated, the get_def extracts all and then rewrap in the modelfield (to avoid duplication in the schema??), now using the lenient_issubclass just basemodel are gonna be flatted, enums (currency from iso4217) are correctly handled (need tests!)

This probably is not needed if they declare correctly the __get_pydantic_json_schema__ but I think that WithJsonSchema should be respected https://docs.pydantic.dev/latest/api/json_schema/#pydantic.json_schema.WithJsonSchema


YuriiMotov:
@JavierSanchezCastro, do you think we still need this?

Failing code example from the [initial discussion](https://github.com/fastapi/fastapi/discussions/14941) was fixed

JavierSanchezCastro:
> @JavierSanchezCastro, do you think we still need this?
> 
> Failing code example from the [initial discussion](https://github.com/fastapi/fastapi/discussions/14941) was fixed

Hi! I just checked that `WithJsonSchema` takes precedence over the original type, it seems it does. That was my only concern, so I’ll close the PR. Thanks for the follow-up!

PR #14943, opened 2026-02-19
refactor: Improve type annotations and use python 3.10+ syntax

argoarsiks:
- **ANN204**: add missing return type annotations in fastapi/security/
- **UP038**:  use `X | Y` syntax for isinstance calls

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -381,19 +381,12 @@ def analyze_param(
         fastapi_annotations = [
             arg
             for arg in annotated_args[1:]
-            if isinstance(arg, (FieldInfo, params.Depends))
+            if isinstance(arg, FieldInfo | params.Depends)
         ]
         fastapi_specific_annotations = [
             arg
             for arg in fastapi_annotations
-            if isinstance(
-                arg,
-                (
-                    params.Param,
-                    params.Body,
-                    params.Depends,
-                ),
-            )
+            if isinstance(arg, params.Param | params.Body | params.Depends)
         ]
         if fastapi_specific_annotations:
             fastapi_annotation: FieldInfo | params.Depends | None = (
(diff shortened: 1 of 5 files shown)


fastapi/dependencies/utils.py
```diff
@@ -833,7 +826,7 @@ def request_params_to_args(
     return values, errors
 
 
-def is_union_of_base_models(field_type: Any) -> bool:
+def is_union_of_base_models(field_type: type) -> bool:
```
```
mypy check...............................................................Failed
- hook id: local-mypy
- exit code: 1

  fastapi/dependencies/utils.py:866: error: Argument 1 to "is_union_of_base_models" has incompatible type "type[Any] | None"; expected "type"  [arg-type]
  Found 1 error in 1 file (checked 47 source files)
```

fastapi/security/api_key.py
```diff
@@ -16,7 +16,7 @@ def __init__(
         description: str | None,
         scheme_name: str | None,
         auto_error: bool,
-    ):
+    ) -> None:
```
YuriiMotov, line 19: It's not necessary to specify return type annotation for `__init__` methods.
If we want to enforce this, we should enable such rule for linter

fastapi/dependencies/utils.py
```diff
@@ -381,19 +381,12 @@ def analyze_param(
         fastapi_annotations = [
             arg
             for arg in annotated_args[1:]
-            if isinstance(arg, (FieldInfo, params.Depends))
+            if isinstance(arg, FieldInfo | params.Depends)
```
YuriiMotov, line 384: With this syntax, performance decreases.
See https://github.com/astral-sh/ruff/issues/7871

fastapi/dependencies/utils.py
```diff
@@ -833,7 +826,7 @@ def request_params_to_args(
     return values, errors
 
 
-def is_union_of_base_models(field_type: Any) -> bool:
+def is_union_of_base_models(field_type: type | None) -> bool:
```
YuriiMotov, line 829: I don't think this is necessary. In utils.py we still pass `first_field.field_info.annotation` which is of type `Any`

YuriiMotov (commented):
@argoarsiks, thanks for your interest!

I think we don't need these changes..

PR #14944, opened 2026-02-19
📝 Fix doctrings for `max_digits` and `decimal_places`

YuriiMotov:
They are only applicable for `Decimal` type: https://github.com/pydantic/pydantic/blob/8be4a57eb36760c85fa97a3bdea78a1f283489d0/pydantic/_internal/_known_annotated_metadata.py#L44

### DIFF
--- a/fastapi/param_functions.py
+++ b/fastapi/param_functions.py
@@ -211,7 +211,7 @@ def Path(  # noqa: N802
         int | None,
         Doc(
             """
-            Maximum number of allow digits for strings.
+            Maximum number of digits allowed for decimal values.
             """
         ),
     ] = _Unset,
@@ -219,7 +219,7 @@ def Path(  # noqa: N802
         int | None,
         Doc(
             """
-            Maximum number of decimal places allowed for numbers.
+            Maximum number of decimal places allowed for decimal values.
             """
         ),
     ] = _Unset,
@@ -570,7 +570,7 @@ def Query(  # noqa: N802
         int | None,
         Doc(
             """
-            Maximum number of allow digits for strings.
+            Maximum number of digits allowed for decimal values.
             """
         ),
     ] = _Unset,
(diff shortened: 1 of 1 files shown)


Areeb455:
I checked out this branch locally and inspected the docstrings in fastapi/param_functions.py.
It looks like the new descriptions were appended rather than replacing the old ones, creating redundancy. Additionally, max_digits still contains a reference to "strings", which is incorrect for this parameter.
Suggestion:


For max_digits, it should be a single clear line:

```
"""
Maximum number of digits allowed for decimal values.
"""
```

For decimal_places, it should be:

```
"""
Maximum number of decimal places allowed for decimal values.
"""
```

YuriiMotov:
> I checked out this branch locally and inspected the docstrings in fastapi/param_functions.py. It looks like the new descriptions were appended rather than replacing the old ones, creating redundancy. Additionally, max_digits still contains a reference to "strings", which is incorrect for this parameter.

I don't quite understand what you mean.
Could you please clarify how you check it?
Maybe you can add a screenshot?

Areeb455:
Thanks for the follow-up! I've attached a screenshot from my editor showing the current state of the branch.
As you can see on line 214 in fastapi/param_functions.py:

```
"""
Maximum number of allow digits for strings.
"""
```

### There is a small issue here:

1)Logic: max_digits is a constraint for numeric/decimal types, but the text says it is for 'strings'. It should be changed to 'numbers' or 'decimal values'.
2)Grammar: It should be 'allowed digits' (past tense) instead of 'allow digits'.


Screenshot 2026-02-22 051210

YuriiMotov:
@Areeb455, this PR hasn't been merged yet. So, to check it you need to checkout to the branch of this PR (`fix-max_digits-docstring`).
If you do it, you will see that the docstrings are:
`Maximum number of digits allowed for decimal values.` and `Maximum number of decimal places allowed for decimal values.`

alejsdev (approved):

Areeb455:
Thank you @YuriiMotov! I've reviewed the latest changes in the branch. The docstrings are now technically accurate and the wording is much clearer.

PR #14945, opened 2026-02-19
docs: update pytest output example in Testing page

cluster2600:
The Testing tutorial page shows an example pytest session output using very old versions (Python 3.6, pytest 5, and the deprecated py dependency).

This updates the example output line to use modern placeholder versions (Python 3.12.x, pytest 8.x.y, pluggy 1.x.y) and removes the extra plugins line from the snippet to avoid suggesting users will see those third-party plugins installed by default.

Only the English docs were updated.

### DIFF
--- a/docs/en/docs/tutorial/testing.md
+++ b/docs/en/docs/tutorial/testing.md
@@ -178,9 +178,8 @@ Run the tests with:
 $ pytest
 
 ================ test session starts ================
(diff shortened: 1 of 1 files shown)


YuriiMotov:
I don't think this is necessary. This is just an example output. Versions are outdated, but there is nothing really wrong in it

cluster2600:
Thanks for the feedback @YuriiMotov — closing as the existing output, while outdated, isn't incorrect. Will look for a more impactful contribution.

PR #14946, opened 2026-02-20
✏️ Fix typo for `client_secret` in OAuth2 form docstrings

bysiber:
The `client_secret` parameter docstring in both `OAuth2PasswordRequestForm` and `OAuth2PasswordRequestFormStrict` incorrectly referred to `client_password` instead of `client_secret`.

The parameter itself is correctly named `client_secret`, and the rest of each docstring also uses `client_secret` — it's just this one mention that had the wrong name.


YuriiMotov (approved):
LGTM!

svlandeg (approved):
Thanks @bysiber! 🎉

PR #14947, opened 2026-02-20
Clarify usage of middleware for X-Request-ID request tracing

amishikaushal:
Added clarification about using middleware to include the `X-Request-ID` header for request tracing in production systems.

This provides a practical example of how middleware is commonly used in real-world distributed applications.

### DIFF
--- a/docs/en/docs/tutorial/middleware.md
+++ b/docs/en/docs/tutorial/middleware.md
@@ -59,6 +59,9 @@ For example, you could add a custom header `X-Process-Time` containing the time 
 
 {* ../../docs_src/middleware/tutorial001_py310.py hl[10,12:13] *}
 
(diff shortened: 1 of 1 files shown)


amishikaushal:
Added clarification about using middleware to include the `X-Request-ID` header for request tracing in production systems.

This provides a practical example of how middleware is commonly used in real-world distributed applications.

YuriiMotov:
@amishikaushal, thanks for you interest!

I think we don't need to add this.
The purpose of this section is to describe what middleware is and how it works. And I think the existing example is enough for this.
We shouldn't overload this section with unnecessary details.

There are a lot of use cases for middlewares. We provide some links to discover this further in [FastAPI
Advanced Middleware](https://fastapi.tiangolo.com/advanced/middleware/#other-middlewares) section

amishikaushal:
Thanks for the clarification!

That makes sense. I agree that keeping this section focused and concise is better. 

Thank you for reviewing and for the guidance 🙂

PR #14950, opened 2026-02-20
Clarify usage of --reload flag in Docker environments

gaurav101203:
Adds a clarification note explaining why the --reload flag should not be used in Docker production environments. This helps prevent confusion for developers transitioning from local development to containerized deployment.


YuriiMotov:
We don't recommend using `--reload` option in this section of docs. So, I think we don't need this clarification here.
Auto-reload is described in https://fastapi.tiangolo.com/fastapi-cli/#fastapi-dev

Also, this PR contains a lot of unnecessary (and invalid) changes.

Let's close it.
@gaurav101203, thanks!

PR #14951, opened 2026-02-20
🔨 Fix `FastAPI People` workflow

YuriiMotov:
Currently `FastAPI People` workflow fails due to hitting GitHub API rate limits.

After some experiments and googling I found out that the reason of hitting rate limits is in requests per query.
Since September rate limit rules became more complex and it's not recommended to make heavy queries (that return thousands of records).

To avoid hitting rate limits I:
* reduced the number of records fetched by query (30 discussions instead of 100)
* added a rate limiter that respects different types of limits:
  * pauses if the number of remaining points is less than needed for request
  * respects safe time intervals between requests (interval time is calculated from the limitation of 5000 points per hour and the number of points used by each request)

This way script will take more time to fetch data, but we will likely avoid hitting rate limits

---

If we need it, we can experiment with the speed by setting `speed_multiplier` ENV variable (defaults to 1, that is the spread with which we will spend 84 points per minute (5000 points per hour divided by 60 minutes)).

### DIFF
--- a/scripts/people.py
+++ b/scripts/people.py
@@ -5,6 +5,7 @@ import time
 from collections import Counter
 from collections.abc import Container
 from datetime import datetime, timedelta, timezone
+from math import ceil
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


scripts/people.py
```diff
@@ -15,12 +16,64 @@
 from pydantic_settings import BaseSettings
 
 github_graphql_url = "https://api.github.com/graphql"
-questions_category_id = "MDE4OkRpc2N1c3Npb25DYXRlZ29yeTMyMDAxNDM0"
+questions_category_id = "DIC_kwDOCZduT84B6E2a"
```
YuriiMotov, line 19: Old value is "global id" and it's deprecated (produced deprecation warning with the recommendation to use this new ID)

tiangolo (approved):
Awesome, thank you! 🙌 🚀

PR #14953, opened 2026-02-21
♻️ Fix JSON Schema for bytes, use `"contentMediaType": "application/octet-stream"` instead of `"format": "binary"`

tiangolo:
♻️ Fix JSON Schema for bytes, use `"contentMediaType": "application/octet-stream"` instead of `"format": "binary"`

## Background

`format: binary` was defined in OpenAPI 3.0.x, in OpenAPI 3.1.x the schema was aligned with the latest JSON Schema, recommending instead `contentMediaType: application/octet-stream`.

I suspect the JSON Schema for `bytes` using `"format": "binary"` comes from my first implementation in Pydantic 1.x.

It was defined and suggested in OpenAPI 3.0.x (not in JSON Schema): https://spec.openapis.org/oas/v3.0.3.html#considerations-for-file-uploads

OpenAPI 3.1.x aligned support with JSON Schema draft 07, so it was suggested to upate file uploads to use the regular JSON Schema format: `"contentMediaType": "application/octet-stream"`: https://learn.openapis.org/upgrading/v3.0-to-v3.1

This is defined in JSON Schema 07: https://json-schema.org/draft-07/json-schema-validation#rfc.section.8.4

### JSON Schema 2020-12 Note

Now OpenAPI 3.2 is aligned with JSON Schema 2020-12, which is what Pydantic v2 implements (except for this, I'm implementing it there too).

It's the same as in JSON Schema draft 07, so this still applies: https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-01#name-contentmediatype

### Usage in JSON

JSON as a format actually doesn't support bytes, everything has to be in UTF-8 strings. Transporting bytes in JSON would require encoding bytes in a string, e.g. with base64.

But as JSON Schema is not only defined to declare JSON payloads but also payloads that could have a comparable structure and defined with JSON Schema, it's still there in the spec.


YuriiMotov:
But it will now also change schema for other parameters (not only `File` parameters):

```py
from typing import Annotated

class Model(BaseModel):
    data: bytes

app = FastAPI()

@app.get("/")
def read_root(params: Annotated[Model, Query()]):
    pass

@app.post("/")
def post_root(params: Model):
    pass
```

<details><summary>See schema</summary>

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "FastAPI",
    "version": "0.1.0"
  },
  "paths": {
    "/": {
      "get": {
        "summary": "Read Root",
        "operationId": "read_root__get",
        "parameters": [
          {
            "name": "data",
            "in": "query",
            "required": true,
            "schema": {
              "type": "string",
              "contentMediaType": "application/octet-stream",  # <= contentMediaType for Query parameter
              "title": "Data"
            }
          }
        ],
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {

                }
              }
            }
          },
          "422": {
            "content": {
              "application/json": {
                "schema": {
                }
              }
            }
          }
        }
      },
      "post": {
        "summary": "Post Root",
        "operationId": "post_root__post",
        "requestBody": {
          "required": true,
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/Model"
              }
            }
          }
        },
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {

                }
              }
            }
          },
          "422": {
            "content": {
              "application/json": {
                "schema": {
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "HTTPValidationError": {
        "properties": {
          "detail": {
            "items": {
              "$ref": "#/components/schemas/ValidationError"
            },
            "type": "array",
            "title": "Detail"
          }
        },
        "type": "object",
        "title": "HTTPValidationError"
      },
      "Model": {
        "properties": {
          "data": {
            "type": "string",
            "contentMediaType": "application/octet-stream",  # <= contentMediaType for Json Body
            "title": "Data"
          }
        },
        "type": "object",
        "required": [
          "data"
        ],
        "title": "Model"
      },
      "ValidationError": {
        "properties": {
          "loc": {
            "items": {
              "anyOf": [
                {
                  "type": "string"
                },
                {
                  "type": "integer"
                }
              ]
            },
            "type": "array",
            "title": "Location"
          },
          "msg": {
            "type": "string",
            "title": "Message"
          },
          "type": {
            "type": "string",
            "title": "Error Type"
          },
          "input": {
            "title": "Input"
          },
          "ctx": {
            "type": "object",
            "title": "Context"
          }
        },
        "type": "object",
        "required": [
          "loc",
          "msg",
          "type"
        ],
        "title": "ValidationError"
      }
    }
  }
}
```

</details>

Will it be correct?

tiangolo:
Ah, the thing is that using `bytes` in Query parameters is undefined/not supported, as a URL can't really have pure bytes, they have to be encoded in ASCII.

I also checked with Swagger UI and the behavior is the same with both `"format": "binary"` and `"contentMediaType": "application/octet-stream"`. In both cases it renders the same and doesn't work.

image

YuriiMotov:
> Ah, the thing is that using `bytes` in Query parameters is undefined/not supported, as a URL can't really have pure bytes, they have to be encoded in ASCII.

I agree about Query parameters (and other non-Body parameters). But it also affects JSON Body parameters (see previous schema example).

Not sure how useful it is to use `bytes` in JSON `Body` models, but it might probably affect people who re-use existing models (defined somewhere else) to declare parameters

tiangolo:
Ah, yes, this would be the actual correct format. JSON also doesn't support direct bytes, it has to be strings encoded in UTF-8, it would have to be in base64 or similar.

So, if anyone is using `bytes` in other places, this would probably fix their JSON Schema, what they have probably doesn't work properly.

I'm also making a PR to Pydantic, just added here another part to support base64.

This would work:

```Python
from fastapi import FastAPI


class Model(BaseModel):
    data: bytes

    model_config = {"ser_json_bytes": "base64"}


app = FastAPI()


@app.post("/")
def post_root(params: Model):
    pass
```

And now it has the proper JSON Schema defininig it, and tools like Swagger UI can represent it correctly:

image

I also added all the references to the description in this PR.

pperliti:
Hi @tiangolo,
from version 0.129.1 onward, parameters declared as a list of files (e.g., `support_files: Annotated[List[UploadFile], File(description=f"Optional. Files used to extract data.")] = None`) are rendered incorrectly in Swagger.

image

OscarMCV-SW:
Hi, same issue here. I rolled back my local FastAPI version and noticed that 0.129.1 breaks our schemas. When generating TS clients from openapi.json, some File arrays are now interpreted as arrays of strings.

It may be related to [this](https://github.com/OpenAPITools/openapi-generator/issues/20115).

PR #14962, opened 2026-02-22
✨ Serialize JSON response with Pydantic (in Rust), when there's a Pydantic return type or response model

tiangolo:
✨ Serialize JSON response with Pydantic (in Rust), when there's a Pydantic return type or response model

### DIFF
--- a/docs/en/docs/advanced/custom-response.md
+++ b/docs/en/docs/advanced/custom-response.md
@@ -1,6 +1,6 @@
 # Custom Response - HTML, Stream, File, others { #custom-response-html-stream-file-others }
 
-By default, **FastAPI** will return the responses using `JSONResponse`.
+By default, **FastAPI** will return JSON responses.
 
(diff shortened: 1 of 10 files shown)


a-babarytskyi:
A bit stupid question maybe, but If I was already returning pydantic model and response model was that model, I will still get performance boost?

tiangolo:
Yes, you will :rocket:

pmsmanishchauhan:
@tiangolo  currently we are using _**response_model**_  will it also performance improvement ?
```
@router.get("/sample", response_model=Sample)
def sample_endpoint(
    *,
    db: Session = Depends(deps.get_db),
    store_id: int,
    user_id: Optional[int] = None,
):
```

TheoPascoli:
@pmsmanishchauhan Yes it will

dolfinus:
Same but for parsing HTTP body: #13951

PR #14964, opened 2026-02-22
🗑️ Deprecate `ORJSONResponse` and `UJSONResponse`

tiangolo:
🗑️ Deprecate `ORJSONResponse` and `UJSONResponse`

Now that there's better performance by default, with response models: https://github.com/fastapi/fastapi/pull/14962

This removes `ujson` and `orjson` from the `"fastapi[all]"` extras.

To use these responses, `ujson` or `orjson` need to be explicitly installed.

### DIFF
--- a/docs/en/docs/reference/responses.md
+++ b/docs/en/docs/reference/responses.md
@@ -22,7 +22,13 @@ from fastapi.responses import (
 
 ## FastAPI Responses
 
-There are a couple of custom FastAPI response classes, you can use them to optimize JSON performance.
+There were a couple of custom FastAPI response classes that were intended to optimize JSON performance.
+
+However, they are now deprecated as you will now get better performance by using a [Response Model - Return Type](https://fastapi.tiangolo.com/tutorial/response-model/).
+
+That way, Pydantic will serialize the data into JSON bytes on the Rust side, which will achieve better performance than these custom JSON responses.
+
+Read more about it in [Custom Response - HTML, Stream, File, others - `orjson` or Response Model](https://fastapi.tiangolo.com/advanced/custom-response/#orjson-or-response-model).
 
 ::: fastapi.responses.UJSONResponse
     options:
--- a/fastapi/responses.py
+++ b/fastapi/responses.py
@@ -1,5 +1,6 @@
 from typing import Any
 
+from fastapi.exceptions import FastAPIDeprecationWarning
 from starlette.responses import FileResponse as FileResponse  # noqa
 from starlette.responses import HTMLResponse as HTMLResponse  # noqa
 from starlette.responses import JSONResponse as JSONResponse  # noqa
@@ -7,6 +8,7 @@ from starlette.responses import PlainTextResponse as PlainTextResponse  # noqa
 from starlette.responses import RedirectResponse as RedirectResponse  # noqa
 from starlette.responses import Response as Response  # noqa
 from starlette.responses import StreamingResponse as StreamingResponse  # noqa
+from typing_extensions import deprecated
 
 try:
     import ujson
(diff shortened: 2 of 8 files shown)


remimd:
With this deprecation, what's the recommended approach for exception handlers that return plain dict bodies?

For example:
```python
@app.exception_handler(ValidationError)
async def validation_error_handler(request: Request, exception: ValidationError) -> Response:
    return ORJSONResponse(
        status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
        content={"errors": exception.errors()},
    )
```

Since exception handlers don't use Pydantic return types, they don't benefit from the Pydantic serialization. Should we just switch to JSONResponse and accept the performance difference, or is there another approach?

tmdgusya:
Are there any benchmark or something?

YuriiMotov:
> With this deprecation, what's the recommended approach for exception handlers that return plain dict bodies?

You can copy the implementation of `ORJSONResponse` and use it instead of importing from `fastapi`.
Or, you can return `Response(content=ujson.dumps(content, ensure_ascii=False).encode("utf-8"))`.
Don't forget to add `ujson` explicit dependency to your project

eltoder:
> Are there any benchmark or something?

@tmdgusya I ran benchmarks for my responses (https://github.com/fastapi/fastapi/pull/14299#issuecomment-3493559629). Feel free to plug in yours and check.

remimd:
@YuriiMotov thanks for the answer! Though I think deprecating these classes is a mistake, for plain dict serialization, orjson are still significantly faster than stdlib, and that's free performance. Copying the implementation feels like a workaround for removing something that's still useful.

YuriiMotov:
> @YuriiMotov thanks for the answer! Though I think deprecating these classes is a mistake, for plain dict serialization, orjson are still significantly faster than stdlib, and that's free performance. Copying the implementation feels like a workaround for removing something that's still useful.

Do you mean routes like this:

```py
@app.get("/default")
def get_default() -> dict:
    return {"name": "widget", "price": 9.99}
```
?

In this case FastAPI will also serialize it with Pydantic, not `stdlib`:

<details>

```py
from unittest.mock import patch

from fastapi import FastAPI


class Item(BaseModel):
    name: str
    price: float


app = FastAPI()


@app.get("/default")
def get_default() -> dict:
    return {"name": "widget", "price": 9.99}


client = TestClient(app)


def test_default_response_class_skips_json_dumps():
    """When no response_class is set, the fast path serializes directly to
    JSON bytes via Pydantic's dump_json and never calls json.dumps."""
    with patch(
        "starlette.responses.json.dumps", wraps=__import__("json").dumps
    ) as mock_dumps:
        response = client.get("/default")
    assert response.json() == {"name": "widget", "price": 9.99}
    mock_dumps.assert_not_called()

```

</details>

---

If you still think we need to de-deprecate these classes, please open a discussion and let's continue discussing it there

remimd:
Oh my bad, I didn't realize Pydantic could serialize plain dicts too! In that case, should Pydantic serialization become the default for `JSONResponse` as well?

**Edit**: I just started a discussion on this topic https://github.com/fastapi/fastapi/discussions/14980

Antares0982:
I came across this PR and would like to ask whether fastapi plans to support additional third‑party JSON libraries. I have developed a new high‑performance JSON library, [ssrJSON](https://github.com/Antares0982/ssrJSON), whose overall performance surpasses orjson, particularly in handling non‑ASCII strings and floating‑point numbers. ssrJSON provides `dumps` and `dumps_to_bytes`, returning str and UTF‑8‑encoded bytes, respectively, and it might be useful for fastapi (I’m not very familiar with fastapi yet, sorry if I'm wrong).

If fastapi is still going to support third‑party JSON libraries, I can add features to ssrJSON to integrate with fastapi, for example ssrJSON currently does not serialize `numpy` types and also some of the types that orjson already supports. Are there specific types that fastapi requires a JSON encoder to support at minimum? Or should I open a new issue/discussion to discuss this further?

epaes90 (commented):
## Argus Code Review

**Score:** 100/100

Reviewed 6 files with no issues found. Highlights: Test coverage is present; Error handling is in place; Configuration is well-structured; No security, performance, or quality issues detected. Score: 100/100.

_No inline findings to report._

PR #14971, opened 2026-02-23
🍻 Debug benchmark, compare old code vs new without coverage interferring in overhead 

tiangolo:
🍻 Debug benchmark, compare old code vs new without coverage interferring in overhead


tiangolo:
It seems the previous performance tests were accurate enough  too. :coffee:

PR #14974, opened 2026-02-23
👷 Allow skipping `benchmark` job in `test` workflow

YuriiMotov:
The `benchmark` job is skipped for docs-only PRs, but it's still required in `check` job. This causes `test` wokflow to fail for PRs that only modify docs


tiangolo (approved):
Good catch, thank you! 🙌

PR #14979, opened 2026-02-23
📝 Rename `docs_src/websockets` to `docs_src/websockets_` to avoid import errors

YuriiMotov:
I updated `advanced/websockets.md` in all translations - they are all up-to-date at the moment, so, it's safe to do that


tiangolo (approved):
Excellent, thank you! 🙌

PR #14982, opened 2026-02-23
🔧 Add FastAPI Agents Skill

tiangolo:
🔧 Add FastAPI Agents Skill

### DIFF
--- a/fastapi/.agents/skills/fastapi/SKILL.md
+++ b/fastapi/.agents/skills/fastapi/SKILL.md
@@ -9,7 +9,7 @@ Official FastAPI skill to write code with best practices, keeping up to date wit
 
 ## Use the `fastapi` CLI
 
-To run the development server, in localhost, with reload:
+Run the development server on localhost with reload:
 
 ```bash
 fastapi dev
--- a/fastapi/.agents/skills/fastapi/SKILL.md
+++ b/fastapi/.agents/skills/fastapi/SKILL.md
@@ -71,11 +71,6 @@ instead of:
 
 ```python
 # DO NOT DO THIS
-from fastapi import FastAPI, Path, Query
-
-app = FastAPI()
-
-
 @app.get("/items/{item_id}")
 async def read_item(
     item_id: int = Path(ge=1, description="The item ID"),
@@ -114,15 +109,6 @@ instead of:
 
 ```python
 # DO NOT DO THIS
-from fastapi import Depends, FastAPI
-
-app = FastAPI()
-
-
-def get_current_user():
-    return {"username": "johndoe"}
-
-
 @app.get("/items/")
 async def read_item(current_user: dict = Depends(get_current_user)):
     return {"message": "Hello World"}
@@ -138,25 +124,26 @@ Do this, without Ellipsis (`...`):
 from typing import Annotated
 
 from fastapi import FastAPI, Query
+from pydantic import BaseModel, Field
+
+
+class Item(BaseModel):
+    name: str
+    description: str | None = None
+    price: float = Field(gt=0)
+
 
 app = FastAPI()
 
 
-@app.get("/items/")
-async def read_item(project_id: Annotated[int, Query()]):
-    return {"message": "Hello World"}
+@app.post("/items/")
+async def create_item(item: Item, project_id: Annotated[int, Query()]): ...
 ```
 
 instead of this:
 
 ```python
 # DO NOT DO THIS
-from typing import Annotated
-
-from fastapi import FastAPI, Query
-from pydantic import BaseModel, Field
-
-
 class Item(BaseModel):
     name: str = ...
     description: str | None = None
@@ -166,14 +153,8 @@ class Item(BaseModel):
 app = FastAPI()
 
 
-@app.get("/items/")
-async def read_item(project_id: Annotated[int, Query(...)]):
-    return {"message": "Hello World"}
-
-
 @app.post("/items/")
-async def create_item(item: Item):
-    return {"message": "Hello World"}
+async def create_item(item: Item, project_id: Annotated[int, Query(...)]): ...
 ```
 
 ## Return Type or Response Model
@@ -316,7 +297,7 @@ Use dependencies when:
 * Other dependencies need their results (it's a sub-dependency)
 * The logic can be shared by multiple endpoints to do things like error early, authentication, etc.
 * They need to handle cleanup (e.g., DB sessions, file handles), using dependencies with `yield`
-* Their logic needs request input data, like headers, query parameters, etc.
+* Their logic needs input data from the request, like headers, query parameters, etc.
 
 ### Dependencies with `yield` and `scope`
 
--- a/fastapi/.agents/skills/fastapi/SKILL.md
+++ b/fastapi/.agents/skills/fastapi/SKILL.md
@@ -16,7 +16,7 @@ fastapi dev
 ```
 
 
-And to run the production server:
+Run the production server:
 
 ```bash
 fastapi run
--- a/fastapi/.agents/skills/fastapi/SKILL.md
+++ b/fastapi/.agents/skills/fastapi/SKILL.md
@@ -403,7 +403,7 @@ def get_user_me(username: UserNameDep):
 
 Avoid creating class dependencies when possible.
 
-If a class is needed, instead create a regular function dependency that return a class instance.
+If a class is needed, instead create a regular function dependency that returns a class instance.
 
 Do this:
 
--- a/fastapi/.agents/skills/fastapi/SKILL.md
+++ b/fastapi/.agents/skills/fastapi/SKILL.md
@@ -348,7 +348,7 @@ Use dependencies when:
 
 ### Dependencies with `yield` and `scope`
 
-When the using dependencies with `yield`, they can have a `scope` that defines when the exit code is run.
+When using dependencies with `yield`, they can have a `scope` that defines when the exit code is run.
 
 Use the default scope `"request"` to run the exit code after the response is sent back.
 
--- a/fastapi/.agents/skills/fastapi/SKILL.md
+++ b/fastapi/.agents/skills/fastapi/SKILL.md
@@ -303,7 +303,7 @@ async def list_items():
 app.include_router(router, prefix="/items", tags=["items"])
 ```
 
-There could be exceptions but try to follow this convention.
+There could be exceptions, but try to follow this convention.
 
 Apply shared dependencies at the router level via `dependencies=[Depends(...)]`.
 
--- a/fastapi/.agents/skills/fastapi/SKILL.md
+++ b/fastapi/.agents/skills/fastapi/SKILL.md
@@ -545,7 +545,7 @@ When working with SQL databases, prefer using SQLModel as it is integrated with 
 
 Do not use Pydantic `RootModel`, instead use regular type annotations with `Annotated` and Pydantic validation utilities.
 
-For example for a list with validations you could do:
+For example, for a list with validations you could do:
 
 ```python
 from typing import Annotated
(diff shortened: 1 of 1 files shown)


fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+```toml
+[tool.fastapi]
+entrypoint = "my_app.main:app"
+```
+
+This way the same command can be used without having to pass a path every time.
+
+### Use `fastapi` with a path
+
+When adding the entrypoint to `pyproject.toml` is not possible, or the user explicitly asks not to, or it's running an independent small app, you can pass the path to the file to the `fastapi` command:
```
svlandeg: Not sure whether it matters much, but the order is a bit weird here because "passing a path every time" hasn't been clarified yet, and only becomes obvious in the next section ("Use fastapi with a path").

I guess you wanted to have the best practice first though.

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+
+```bash
+fastapi dev my_app/main.py
+```
+
+## Use `Annotated`
+
+Always prefer the `Annotated` style for parameter and dependency declarations.
+
+It keeps the function signatures working in other contexts, respects the types, allows reusability.
```
svlandeg, line 48: I like how there's a super short explanation here, which the agent could uses to defend its choice to its driver 🙂  I wonder whether it would make sense to add similar short explanations to some of the other "Do this, not this" sections, like the Ellipsis ones?

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+        name="Foo", description="A very nice Item", secret_key="supersecret"
+    )
+    return item
+```
+
+## Performance
+
+Do not use `ORJSONResponse` or `UJSONResponse`, they are deprecated.
+
+Instead declare a return type or response model. Pydantic will handle the data serialization in the Rust side.
```
svlandeg:
```suggestion
Instead declare a return type or response model. Pydantic will handle the data serialization on the Rust side.
```

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+* They can't be declared in Pydantic validation and require additional logic
+* The logic depends on external resources or could block in any other way
+* Other dependencies need their results (it's a sub-dependency)
+* The logic can be shared by multiple endpoints to do things like error early, authentication, etc.
+* They need to handle cleanup (e.g., DB sessions, file handles), using dependencies with `yield`
+* Their logic needs request input data, like headers, query parameters, etc.
+
+### Dependencies with `yield` and `scope`
+
+When the using dependencies with `yield`, they can have a `scope` that defines when the exit code is run.
```
svlandeg:
```suggestion
When using dependencies with `yield`, they can have a `scope` that defines when the exit code is run.
```

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+@app.get("/users/me")
+def get_user_me(username: UserNameDep):
+    return username
+```
+
+### Class Dependencies
+
+Avoid creating class dependencies when possible.
+
+If a class is needed, instead create a regular function dependency that return a class instance.
```
svlandeg:
```suggestion
If a class is needed, instead create a regular function dependency that returns a class instance.
```

svlandeg (commented):
Love this! 🚀

fastapi/.agents/skills/fastapi/SKILL.md
svlandeg: (or just "in Rust" ?)

fastapi/.agents/skills/fastapi/SKILL.md
tiangolo: Good point! Let me restructure a bit.

fastapi/.agents/skills/fastapi/SKILL.md
tiangolo: I've been on the fence on both sides... the recommendation is that a `SKILL.md` file should be up to 500 lines, this one is 640... that made me think of reducing how much to say. :sweat_smile: 

I'll give it another pass and see if I can shorten things, or add more explanations if necessary.

I'm also trying to focus mainly on the things that current LLMs sometimes get wrong (e.g. using Ellipsis, or `= Depends()`), or new things that might not have been there when training, or simply not have enough datapoints in the data distribution, e.g. `Depends(func, scope="function")`.

Let me give it another fresh read and update. :nerd_face:

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+description: FastAPI best practices and conventions. Use when working with FastAPI APIs and Pydantic models for them. Keeps FastAPI code clean and up to date with the latest features and patterns, updated with new versions. Write new code or refactor and update old code.
+---
+
+# FastAPI
+
+Official FastAPI skill to write code with best practices, keeping up to date with new versions and features.
+
+## Use the `fastapi` CLI
+
+To run the development server, in localhost, with reload:
```
alejsdev:
```suggestion
Run the development server on localhost with reload:
```

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+
+```bash
+fastapi dev
+```
+
+
+And to run the production server:
```
alejsdev:
```suggestion
Run the production server:
```
nitpick, but I think imperative sounds slightly better 😸

fastapi/.agents/skills/fastapi/SKILL.md
svlandeg: Do you think that for the "DO NOT DO THIS" examples, we need the full Python script? Would it be bad form if we would for instance just do:

```python
from typing import Annotated

app = FastAPI()


def get_current_user():
    return {"username": "johndoe"}


CurrentUserDep = Annotated[dict, Depends(get_current_user)]


@app.get("/items/")
async def read_item(current_user: CurrentUserDep):
```

instead of:

```python
# DO NOT DO THIS

@app.get("/items/")
async def read_item(current_user: dict = Depends(get_current_user)):
```
i.e. not repeat the imports and definitions? Would it be an added bonus that the "do not do this" example wouldn't actually be functional by itself and thus more difficult to accidentally copy/paste?

Maybe I'm overthinking things - was just looking for ways to shave more lines.

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+    return Item(name="Plumbus", description="All-purpose home device")
+```
+
+**Important**: Return types or response models are what filter data ensuring no sensitive information is exposed. And they are used to serialize data with Pydantic, this is the main idea that can increase response performance.
+
+The return type doesn't have to be a Pydantic model, it could be a different type, like a list of Pydantic models, or a dict, etc.
+
+### When to use `response_model` instead
+
+If the return type is not the same as the type that you want to use to validate, filter, serialize, use the `response_model` parameter on the decorator instead.
```
alejsdev:
```suggestion
If the return type is not the same as the type that you want to use to validate, filter, or serialize, use the `response_model` parameter on the decorator instead.
```
nitpick 😸

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+    name: str
+    description: str | None = None
+
+
+@app.get("/items/me", response_model=Item)
+async def get_item() -> Any:
+    return {"name": "Foo", "description": "A very nice Item"}
+```
+
+This can be in particular useful when filtering data to expose only the public fields, and avoid exposing sensitive information.
```
alejsdev:
```suggestion
This can be particularly useful when filtering data to expose only the public fields and avoid exposing sensitive information.
```

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+def read_items():
+    data = some_blocking_library.fetch_items()
+    return data
+```
+
+In case of doubt, or by default, use regular `def` functions, those will be run in a threadpool so they don't block the event loop.
+
+The same rules apply to dependencies.
+
+Make sure blocking code is not run inside of `async` functions. The logic will work, but will damage heavily the performance.
```
alejsdev:
```suggestion
Make sure blocking code is not run inside of `async` functions. The logic will work, but will damage the performance heavily.
```

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+
+## SQLModel for SQL databases
+
+When working with SQL databases, prefer using SQLModel as it is integrated with Pydantic and will allow declaring data validation with the same models.
+
+## Do not use Pydantic RootModels
+
+Do not use Pydantic `RootModel`, instead use regular type annotations with `Annotated` and Pydantic validation utilities.
+
+For example for a list with validations you could do:
```
alejsdev:
```suggestion
For example, for a list with validations you could do:
```

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+There could be exceptions but try to follow this convention.
+
+Apply shared dependencies at the router level via `dependencies=[Depends(...)]`.
+
+## Dependency Injection
+
+Use dependencies when:
+
+* They can't be declared in Pydantic validation and require additional logic
+* The logic depends on external resources or could block in any other way
```
alejsdev, line 296:
```suggestion
* The logic depends on external resources or could be blocked in any other way
```

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+@router.get("/")
+async def list_items():
+    return []
+
+
+# In main.py
+app.include_router(router, prefix="/items", tags=["items"])
+```
+
+There could be exceptions but try to follow this convention.
```
alejsdev:
```suggestion
There could be exceptions, but try to follow this convention.
```

fastapi/.agents/skills/fastapi/SKILL.md
```diff
@@ -0,0 +1,640 @@
...
+        name="Foo", description="A very nice Item", secret_key="supersecret"
+    )
+    return item
+```
+
+## Performance
+
+Do not use `ORJSONResponse` or `UJSONResponse`, they are deprecated.
+
+Instead declare a return type or response model. Pydantic will handle the data serialization on the Rust side.
```
alejsdev:
```suggestion
Instead, declare a return type or response model. Pydantic will handle the data serialization on the Rust side.
```

alejsdev (commented):
LGTM, just a couple of nitpicks 😸

fastapi/.agents/skills/fastapi/SKILL.md
tiangolo: That's a good point, I'm too used to making sure scripts are runnable as is, but in this case that's not the point. :sweat_smile:

fastapi/.agents/skills/fastapi/SKILL.md
tiangolo: Agreed!

fastapi/.agents/skills/fastapi/SKILL.md
tiangolo: Done!

fastapi/.agents/skills/fastapi/SKILL.md
tiangolo: This one refers to "could block the event loop", or "could block the execution".

I'll leave it for now.

PR #14991, opened 2026-02-24
Fix StreamingResponse docs: warn about blocking operations in async generators

thakoreh:
## Description

Fixes issue #14680 - StreamingResponse doc example needs correction

## Problem

The documentation states that StreamingResponse "Takes an async generator or a normal generator/iterator" but doesn't explain that async generators need to actually await something to work properly.

Users might use  in an async generator, which blocks the event loop and prevents proper streaming.

## Solution

Added a warning admonition explaining that:
- Async generators should use  instead of 
- Blocking operations prevent proper streaming

## Changes

- Added warning in docs/en/docs/advanced/custom-response.md

### DIFF
--- a/docs/en/docs/advanced/custom-response.md
+++ b/docs/en/docs/advanced/custom-response.md
@@ -169,6 +169,12 @@ Takes an async generator or a normal generator/iterator and streams the response
 
 {* ../../docs_src/custom_response/tutorial007_py310.py hl[2,14] *}
 
+/// warning
+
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Thanks for your interest, but this doesn't really resolve the issue https://github.com/fastapi/fastapi/issues/14680.

There is already another PR  that will resolve that issue: https://github.com/fastapi/fastapi/pull/14681

PR #14992, opened 2026-02-24
🔨 Run tests with `pytest-xdist` and `pytest-cov`

YuriiMotov:
Tests run x2.3 faster (locally): 18s VS 42s.

The drawback is that it's easy to introduce race condition. We need to be careful and mark tests that modify files in working directory with `@workdir_lock` decorator introduced in this PR.

To make it easier to debug race conditions, I added `pytest-timeout` plugin.

Stress-tested with `-n 50` - adjusted pytest timeout as a result (`5` -> `20`).
After all adjustments, 10 runs passed ✅

---

**Note:** I added `test-cov.sh` script and moved `--cov --cov-context=test` to it. This way when we run just `test.sh` it works much faster (doesn't store coverage data).

---

Inline snapshot gives the following:

```
INFO: inline-snapshot was disabled because you used xdist. This means that tests with snapshots will continue to run, but snapshot(x) will only return x and inline-snapshot will not be able to fix snapshots or generate reports.
```

To update inline snapshots we can run tests as `bash scripts/test.sh -n 0`


tests/test_tutorial/test_websockets/test_tutorial003.py
```diff
@@ -42,11 +43,13 @@ def test_websocket_handle_disconnection(client: TestClient):
         connection.send_text("Hello from 1234")
         data1 = connection.receive_text()
         assert data1 == "You wrote: Hello from 1234"
+        time.sleep(0.01)  # Give server time to process broadcast
```
YuriiMotov, line 46: Without this timeout test **sometimes** froze and failed with timeout.
Not sure if we can handle this in a different way..

scripts/test.sh
```diff
@@ -4,4 +4,4 @@ set -e
 set -x
 
 export PYTHONPATH=./docs_src
-coverage run -m pytest tests scripts/tests/ ${@}
+pytest -n auto --dist loadgroup  tests scripts/tests/ ${@}
```
YuriiMotov, line 7: `--dist loadgroup` means that tests of the same group (marked with `pytest.mark.xdist_group()`) will be run on one worker

pyproject.toml
```diff
@@ -163,7 +163,7 @@ github-actions = [
 tests = [
     { include-group = "docs-tests" },
     "anyio[trio] >=3.2.1,<5.0.0",
-    "coverage[toml] >=6.5.0,<8.0",
+    "coverage[toml] >=7.13,<8.0",
```
YuriiMotov, line 166: With lower versions fails on Windows with `resolution=lowest-direct`

tiangolo (approved):
Awesome! Thanks a lot, great and thorough job. 🙌 

Testing everything locally for me went form 2m15s to 1m37s 🚀

PR #14993, opened 2026-02-24
📝 Fix StreamingResponse async generator example to actually stream

### LINKED ISSUE #14680: StreamingResponse does not work with async generators

Vishnuuuu24:
## Summary

Fixes #14680.

The `StreamingResponse` documentation example uses an async generator (`fake_video_streamer`) without any `await` expression. This causes two problems:

1. **No true streaming** — without an `await` point, the async generator holds the event loop for the entire iteration, so the response is buffered entirely and sent all at once instead of being streamed chunk by chunk.
2. **No proper cancellation** — in asyncio, a task can only be cancelled when it reaches an `await`. Without one, the generator cannot be interrupted and may keep running after cancellation is requested.

### 1. Code example (`docs_src/custom_response/tutorial007_py310.py`)

Added `import asyncio` and `await asyncio.sleep(0)` inside the async generator loop. This is the minimal fix to yield control back to the event loop between iterations.

### 2. Documentation (`docs/en/docs/advanced/custom-response.md`)

- Updated highlighted line numbers to reflect the new `asyncio` import and `await` line.
- Added a `/// note` admonition after the code example explaining why the `await` is necessary (event loop yielding, streaming, cancellation) and suggesting sync generators as an alternative.

Only the **English** docs are updated — translations are handled separately per project convention.

## Validation

- Existing test (`tests/test_tutorial/test_custom_response/test_tutorial007.py`) passes.
- `ruff check` and `ruff format` pass cleanly.
- Verified with a live uvicorn server that chunks now arrive incrementally (10 distinct chunks over ~377ms) instead of all at once.

### DIFF
--- a/docs/en/docs/advanced/custom-response.md
+++ b/docs/en/docs/advanced/custom-response.md
@@ -167,7 +167,19 @@ You can also use the `status_code` parameter combined with the `response_class` 
 
 Takes an async generator or a normal generator/iterator and streams the response body.
 
-{* ../../docs_src/custom_response/tutorial007_py310.py hl[2,14] *}
+{* ../../docs_src/custom_response/tutorial007_py310.py hl[1,4,11,17] *}
+
+/// note
+
+Notice the `await asyncio.sleep(0)` inside the async generator function.
+
+In asyncio, a running task only yields control back to the event loop when it hits an `await`. Without any `await` expression, the generator would hold the event loop for the entire iteration, preventing the response from being streamed chunk by chunk and also preventing proper task cancellation.
+
+Adding `await asyncio.sleep(0)` is a lightweight way to give the event loop a chance to send each chunk to the client and handle cancellation between iterations.
+
+If your generator doesn't have any natural `await` points, consider using a **normal (synchronous) generator** with `def` instead to avoid this issue.
+
+///
 
 #### Using `StreamingResponse` with file-like objects { #using-streamingresponse-with-file-like-objects }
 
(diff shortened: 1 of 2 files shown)


YuriiMotov:
This is a duplicate of https://github.com/fastapi/fastapi/pull/14681
In general it's better to review existing PR and suggest changes, rather than opening duplicate PRs (unless you have really divergent approaches or existing PR is not maintained).

Let's close this and I encourage you review [existing PR](https://github.com/fastapi/fastapi/pull/14681)

@Vishnuuuu24, thank you!

PR #14994, opened 2026-02-24
✅ Fix all tests are skipped on Windows

YuriiMotov:
`pytest_collection_modifyitems` hook from `scripts/tests/test_translation_fixer/conftest.py` doesn't check that item is placed inside the `scripts/tests/test_translation_fixer/`. That leads to skipping all tests on Windows.

```
=========================== 3094 skipped in 21.42s ============================
```

See: https://github.com/fastapi/fastapi/actions/runs/22365255134/job/64729413552#step:9:620


svlandeg (approved):
Nice catch!

PR #14995, opened 2026-02-24
fix: replace assert with raise FastAPIError in routing.py

omnislash157:
## Summary

Three `assert isinstance(...)` guards in `routing.py` check for required ASGI scope keys (`fastapi_middleware_astack`, `fastapi_inner_astack`). These are runtime guards, not development-time invariants.

Under `python -O` (optimized mode), **all `assert` statements are stripped**. This converts these diagnostic guards into bare `AttributeError` exceptions with no context message, making production debugging significantly harder.

**Fix:** Replace with explicit `if not isinstance(...): raise FastAPIError(...)`. Same error messages, same behavior — except now they survive `python -O`.

## Changes

- `routing.py:349` — HTTP middleware stack check (`fastapi_middleware_astack`)
- `routing.py:418` — HTTP dependency stack check (`fastapi_inner_astack`)  
- `routing.py:514` — WebSocket dependency stack check (`fastapi_inner_astack`)

## Before/After

```python
# Before (stripped by python -O)
assert isinstance(file_stack, AsyncExitStack), (
    "fastapi_middleware_astack not found in request scope"
)

# After (survives python -O)
if not isinstance(file_stack, AsyncExitStack):
    raise FastAPIError(
        "fastapi_middleware_astack not found in request scope"
    )
```

`FastAPIError` is already imported in `routing.py` and used for other runtime checks in the same file.

## Test plan

- [x] Verified `FastAPIError` is already imported and used in `routing.py`
- [x] No behavioral change under normal execution (same check, same message)
- [ ] Existing test suite should pass unchanged

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -347,9 +347,7 @@ def get_request_handler(
         response: Response | None = None
         file_stack = request.scope.get("fastapi_middleware_astack")
         if not isinstance(file_stack, AsyncExitStack):
-            raise FastAPIError(
-                "fastapi_middleware_astack not found in request scope"
-            )
+            raise FastAPIError("fastapi_middleware_astack not found in request scope")
 
         # Extract endpoint context for error messages
         endpoint_ctx = (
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@omnislash157, we already had https://github.com/fastapi/fastapi/pull/14449 and we decided not to do it for now.
Thanks!

PR #14996, opened 2026-02-24
feat: add audit_security() for security dependency auditing

omnislash157:
## Summary

Adds `app.audit_security()` — a method that returns all routes with **no** `SecurityBase` dependency anywhere in their dependency tree.

There is currently no way to statically determine which FastAPI endpoints are unprotected without manually traversing the full dependency graph of every route. This method does that traversal for you.

## Usage

```python
app = FastAPI()

# ... define routes ...

# Find all unprotected routes
unprotected = app.audit_security()
for route in unprotected:
    print(f"UNPROTECTED: {route['methods']} {route['path']}")
```

## What it detects

Recursively walks the full `Dependant` tree for each route. Catches security at any depth:

- Direct: `def endpoint(token: str = Depends(oauth2_scheme))`
- Nested: `def endpoint(user=Depends(get_user))` where `get_user` depends on `oauth2_scheme`
- Router-level: `APIRouter(dependencies=[Depends(bearer)])`
- All SecurityBase types: `OAuth2PasswordBearer`, `HTTPBearer`, `HTTPBasic`, `APIKeyHeader/Query/Cookie`, `OpenIdConnect`
- `Security()` wrapper with scopes

## Implementation

~40 lines added to `FastAPI` class in `applications.py`. Uses the existing `Dependant._is_security_scheme` property — no new dependencies, no new abstractions.

Returns `list[dict]` with `path`, `methods`, `name`, and `endpoint` for each unprotected route.

## Test plan

Verified against:
- [x] Direct OAuth2PasswordBearer dependency → correctly excluded
- [x] Nested dependency chain (endpoint → get_user → oauth2_scheme) → correctly excluded
- [x] Router-level dependencies → correctly excluded
- [x] Security() with scopes → correctly excluded
- [x] APIKeyHeader → correctly excluded
- [x] HTTPBearer → correctly excluded
- [x] Plain unprotected routes → correctly returned
- [x] Mixed app with protected and unprotected → only unprotected returned
- [ ] Existing test suite should pass unchanged (method is additive)

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -1099,6 +1099,65 @@ class FastAPI(Starlette):
             )
         return self.openapi_schema
 
+    def audit_security(
+        self,
+    ) -> list[dict[str, Any]]:
+        """
+        Audit all routes for security dependencies.
+
+        Returns a list of route info dicts for every route that has **no**
+        security dependency (i.e. no ``SecurityBase`` instance) anywhere in
+        its dependency tree. Useful for security reviews and compliance
+        audits.
+
+        Each dict contains:
+
+        * ``path`` – the route path (e.g. ``/users/{user_id}``)
+        * ``methods`` – the HTTP methods (e.g. ``{"GET"}``)
+        * ``name`` – the route/endpoint name
+        * ``endpoint`` – the endpoint callable
+
+        Example::
+
+            unprotected = app.audit_security()
+            for route_info in unprotected:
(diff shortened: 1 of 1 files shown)


YuriiMotov:
I'm not sure this is something people need.
The intention is to check if all endpoints are protected, right?
I think this should be ensured by adding tests.

Let's close this for now. If you still think this might be useful feature - please, open discussion and gather community feedback. If community supports this, we will reconsider this

PR #14999, opened 2026-02-25
docs: fix duplicated word in release notes

Rohan5commit:
## Summary
Fix a duplicated word in the release notes entry: "for for" -> "for".

## Scope
- docs only
- no behavior changes


YuriiMotov:
Thanks!
We have quite a lot of typos in release notes and we would prefer to address them all in a complex way rather than fixing them one by one.

PR #15004, opened 2026-02-25
Add AsyncAPI support for Websockets

re-chain:


re-chain:
Would love some eyes from a maintainer!

PR #15006, opened 2026-02-26
Feat/asyncapi-websockets

re-chain:
Add AsyncAPI documentation for Websockets.


Svenum:
I would love to see this feature in production. What is the current state?

tiangolo:
This seems to be a big change, let's start with a GitHub Discussion, explaining your specific use case and what you would like to have or see. For now I'll pass on this. :coffee:

Svenum:
I need this to document my websocket endpoints. Its really sad to see this getting closed so fast. :(

PR #15030, opened 2026-03-01
✨ Add support for Server Sent Events

tiangolo:
✨ Add support for Server Sent Events

This will be available in FastAPI 0.135.0, released in the next hours.


samuelcm-bflexion:
Hi, great to see official support for this! Is this intended to be a drop-in replacement for sse-starlette's EventSourceResponse? If not it would be great to see a migration guide in the docs

Issue #1573, opened 2020-06-14
automactic 'additional responses' based on raised HTTPException in code within build-in decorators


ihakh:
### First check

* [x] I added a very descriptive title to this issue.
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to:
    * Read open issues with questions until I find 2 issues where I can help someone and add a comment to help there.
    * Or, I already hit the "watch" button in this repository to receive notifications and I commit to help at least 2 people that ask questions in the future.
    * Implement a Pull Request for a confirmed bug.

### Example

```Python
from fastapi import FastAPI, Depends, status

app = FastAPI()


@app.post("/dists/{distribution}/{component}",
        response_model=List[PackageInfo],
        response_model_exclude_unset=True,
        responses={
            status.HTTP_400_BAD_REQUEST: {
                "model": Message,
                "description": "Not All of packages provided"
            },
            status.HTTP_403_FORBIDDEN: {
                "model": Message,
                "description": "A package has been uploaded which is not allowed in distribution"
            },
            status.HTTP_404_NOT_FOUND: {
                "model": Message,
                "description": "Requested distribution does not exists"
            },
            status.HTTP_409_CONFLICT: {
                "model": Message,
                "description": "The component already exist"
            },
            status.HTTP_415_UNSUPPORTED_MEDIA_TYPE: {
                "model": Message,
                "description": "The server could not recognize the uploaded file as debian package file"
            },
        },
        description="Create a new **componont** in **distribution** with provided packages."
            # "when all of the following is provided:\n{}".format(
            # '\n'.join(['- **%s**'%p for p in PACKAGES_LIST]))
    )
async def new_component(
        package_list: List[PackageInfo] = Depends(save_deb)
    ):
    return package_list
```

### Description

[...]

PR #15040, opened 2026-03-02
Document Response subclasses declared in return annotations in OpenAPI

AlberLC:
## Background and Motivation

This PR builds on the discussion https://github.com/fastapi/fastapi/discussions/14172 I opened some months ago regarding the possibility of reflecting `Response` subclasses declared in return type annotations in OpenAPI documentation.

#### Example 1:

```python
@router.get('/{file_name}')
async def embed_page(file_name: str, request: Request) -> HTMLResponse | RedirectResponse:
    file_url = request.url_for('get_file', file_name=file_name)

    if 'bot' not in request.headers.get('user-agent', '').lower():
        return RedirectResponse(file_url)

    return HTMLResponse(embed_service.generate_html(file_name, file_url, request))
```
image

#### Example 2:

```python
class Item(BaseModel):
    id: int


class User(BaseModel):
    username: str
    password: str


@app.get('/')
def read_root() -> HTMLResponse | RedirectResponse | Item | PlainTextResponse | User:
    return User(username='root', password='<PASSWORD>')

```
chrome_LkeuELpI0T

After revisiting the topic and spending more time reading through the routing and OpenAPI internals, I refined the original idea to better align with FastAPI’s design principles.

FastAPI already leverages return type annotations for response models. However, `Response` subclasses declared in return annotations are currently not reflected in OpenAPI, and unions of `Response` subclasses have historically led to model inference errors unless `response_model=None` was explicitly set (https://github.com/fastapi/fastapi/discussions/13053).

Several past discussions and issues have highlighted limitations around unions and response model inference (e.g. https://github.com/fastapi/fastapi/issues/101, https://github.com/fastapi/fastapi/issues/2296, https://github.com/fastapi/fastapi/issues/5215, https://github.com/fastapi/fastapi/pull/875, https://github.com/fastapi/fastapi/pull/1436). While simple single-response cases were improved in https://github.com/fastapi/fastapi/pull/5855, unions of `Response` subclasses and mixed unions (responses + models) still expose inconsistencies in documentation and model inference.

This PR aims to address those inconsistencies while preserving FastAPI’s runtime design principles.

## Scope and Design Decisions

After getting used to how cleanly response models can be declared in return type annotations (instead of always relying on the `response_model` parameter in the decorator), it felt natural to explore whether `Response` subclasses could follow a similar pattern.

However, extending that idea to runtime behavior quickly revealed deeper implications. Unlike response models, `response_class` is not only descriptive metadata, it is the fallback rendering mechanism used when the returned value is not already a `Response` instance. Allowing unions of `Response` subclasses to implicitly determine `response_class` would require additional resolution logic (for example, selecting one class over another, prioritizing a specific implementation, or relying on declaration order), introducing ambiguity and hidden behavior.

This becomes even more problematic when considering streaming responses, including JSONL and SSE support, where response handling is tightly coupled with rendering semantics.

For these reasons, the decision was made to keep `response_class` strictly runtime-oriented and unaffected by return annotations. `Response` subclasses declared in return types are therefore used exclusively for OpenAPI documentation.

During exploration, I also considered whether status codes could be inferred dynamically from the function body. However, that would require inspecting the implementation rather than relying purely on type information, introducing fragile heuristics and going beyond FastAPI’s current boundaries. Similar concerns were raised previously when discussing automatic documentation of `HTTPException` status codes, where inspecting internal logic was considered unreliable (see https://github.com/fastapi/fastapi/issues/1573#issuecomment-1284256361). For that reason, this change strictly operates on return type annotations and does not attempt to inspect return expressions or infer dynamic status codes.

Therefore:

- The decorator parameter `response_class` still determines how non-response return values are rendered.
- No implicit selection between multiple `Response` classes is introduced.
- No dynamic status code inference from function bodies.
- No changes to response serialization semantics.

Overall, the scope of this PR is limited to improving OpenAPI documentation and removing inconsistencies in return annotation handling, without altering how responses are rendered at runtime.

## What This PR Does

Response subclasses found in return annotations are extracted before response model inference. They are excluded from Pydantic model field creation and instead used exclusively for OpenAPI documentation.

The generated OpenAPI schema now reflects:

- Media types of `Response` subclasses declared in return annotations.
- Default status codes defined by those `Response` classes (for example, `307` for `RedirectResponse`).
- Multiple `Response` subclasses when unions are declared.
- Mixed unions combining `Response` subclasses with Pydantic models or primitive types.

Unions of `Response` subclasses no longer raise model inference errors, and explicit `response_model=None` is no longer required in those scenarios.

The implementation integrates with the new JSONL and SSE streaming logic. The `route.is_json_stream` behavior remains unchanged.

## Tests

All existing tests pass.

A small number of tests were adjusted where previous OpenAPI snapshots did not accurately reflect declared return annotations (for example, unions of `Response` subclasses or media type expectations).

Additionally, 36 new tests were added, and these tests provide complete coverage for the new logic and ensure compatibility with existing behavior, covering:

- Single `Response` subclass annotations.
- Unions of multiple `Response` subclasses.
- Mixed unions (`Response` subclasses + models).
- Mixed unions including primitive types.
- Interaction between return annotations and decorator `response_class`.
- Edge cases involving `None`.

## Documentation Impact

The following documentation sections may require updates to reflect this behavior:

- [/tutorial/response-model](https://fastapi.tiangolo.com/tutorial/response-model/)
- [/advanced/response-directly](https://fastapi.tiangolo.com/advanced/response-directly/)
- [/advanced/custom-response](https://fastapi.tiangolo.com/advanced/custom-response)

This PR does not modify documentation content, as wording and presentation decisions are likely better handled separately.

Feedback is very welcome.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -3,6 +3,7 @@ import email.message
 import functools
 import inspect
 import json
+import operator
 import types
 from collections.abc import (
     AsyncIterator,
@@ -26,7 +27,10 @@ from typing import (
     Annotated,
     Any,
     TypeVar,
+    Union,
     cast,
+    get_args,
+    get_origin,
 )
 
 import anyio
@@ -65,7 +69,7 @@ from fastapi.sse import (
     ServerSentEvent,
     format_sse_event,
 )
-from fastapi.types import DecoratedCallable, IncEx
+from fastapi.types import DecoratedCallable, IncEx, UnionType
 from fastapi.utils import (
     create_model_field,
     generate_unique_id,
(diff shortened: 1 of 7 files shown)


fastapi/routing.py
```diff
@@ -840,27 +844,54 @@ def __init__(
...
-                    # into validation or OpenAPI schema generation.
-                    if (
-                        isinstance(response_class, DefaultPlaceholder)
-                        or lenient_issubclass(response_class, EventSourceResponse)
-                    ) and not lenient_issubclass(stream_item, ServerSentEvent):
-                        self.stream_item_type = stream_item
-                    response_model = None
+                origin = get_origin(return_annotation)
+
+                if origin is Union or origin is UnionType:
```
AlberLC, line 875: Explicitly checking both `Union` and `UnionType` to stay consistent with the existing union handling patterns already present in the codebase (for example in `dependencies/utils.py`, `is_union_of_base_models`).

PR #15041, opened 2026-03-02
Add lazy initialize support for OAuth2 bearer schemes

### LINKED ISSUE #3317: Add support to lazy initialize `OAuth2` classes
### First check

### Example

```Python
from fastapi import FastAPI, Security
from fastapi.security import OAuth2AuthorizationCodeBearer


class Config(BaseSettings):
    OAUTH2_AUTHORIZATION_URL: str = "https://example.com/authorize"
    OAUTH2_TOKEN_URL: str = "https://example.com/oauth/token"


config = Config()
app = FastAPI()


oauth2_scheme = OAuth2AuthorizationCodeBearer(
    authorizationUrl=config.OAUTH2_AUTHORIZATION_URL,
    tokenUrl=config.OAUTH2_TOKEN_URL,
)


@app.get("/private-route", dependencies=[Security(oauth2_scheme, scopes=["admin"])])
async def private_route():
    pass
```

### Description

* `Config` is instantiated at import time because `OAuth2AuthorizationCodeBearer` class needs values from it
* I would like a way to lazy initialize `OAuth2AuthorizationCodeBearer` and other security classes

Since https://github.com/encode/uvicorn/pull/875 we have been able to use `--factory` option with uvicorn. This is known as the application factory pattern and is widely used in Flask. I would like to delay configuring the `OAuth2AuthorizationCodeBearer` until inside the `create_app` method like below:

```Python
from typing import Optional

from fastapi import APIRouter, FastAPI, Security
from fastapi.security import OAuth2AuthorizationCodeBearer


class Config(BaseSettings):
    OAUTH2_AUTHORIZATION_URL: str = "https://example.com/authorize"
    OAUTH2_TOKEN_URL: str = "https://example.com/oauth/token"


oauth2_scheme = OAuth2AuthorizationCodeBearer()

router = APIRouter()


@router.get("/private-route", dependencies=[Security(oauth2_scheme, scopes=["admin"])])
async def private_route():
    pass


def create_app(config: Optional[Config] = None) -> FastAPI:
    config = config or Config()
    app = FastAPI()
    app.include_router(router)

    # I want to configure here instead
    oauth2_scheme.initialize(
        authorizationUrl=config.OAUTH2_AUTHORIZATION_URL,
        tokenUrl=config.OAUTH2_TOKEN_URL,
    )
    return app

```

Subclassing `OAuth2AuthorizationCodeBearer` to add the functionality, but I think this is a common use case it could be officially supported or a note in the docs.

### Environment

* OS: [e.g. Linux / Windows / macOS]: **macOS**
* FastAPI Version [e.g. 0.3.0]: **0.65.1**
* Python Version: **3.9.4**

### Additional context

If this is agreed I can do the implementation and open a PR.

Takrim-ScrubAI:
Fixes https://github.com/fastapi/fastapi/issues/3317

### Summary
Adds lazy initialization support for OAuth2 security dependencies used with app-factory patterns.

### Changes
- `OAuth2PasswordBearer` now supports deferred configuration with:
  `initialize(tokenUrl=..., refreshUrl=..., scopes=...)`
- `OAuth2AuthorizationCodeBearer` now supports deferred configuration with:
  `initialize(authorizationUrl=..., tokenUrl=..., refreshUrl=..., scopes=...)`

### Implementation
Core implementation:
- https://github.com/Takrim-ScrubAI/fastapi/blob/fix/oauth2-lazy-initialize-3317/fastapi/security/oauth2.py

Internally updates OAuth2 flows through a helper to keep behavior consistent and avoid API breakage.

### Tests
Regression tests added:
- https://github.com/Takrim-ScrubAI/fastapi/blob/fix/oauth2-lazy-initialize-3317/tests/test_security_oauth2_lazy_initialization.py

Validation run on Python 3.13:
- 2 tests passed in the new test file
- 57 tests passed in `tests/test_security_oauth2*`

### DIFF
--- a/fastapi/security/oauth2.py
+++ b/fastapi/security/oauth2.py
@@ -403,7 +403,7 @@ class OAuth2(SecurityBase):
         *,
         flows: OAuthFlowsModel | dict[str, dict[str, Any]],
     ) -> None:
-        self.model.flows = cast(OAuthFlowsModel, flows)
+        cast(OAuth2Model, self.model).flows = cast(OAuthFlowsModel, flows)
 
     def make_not_authenticated_error(self) -> HTTPException:
         """
--- a/fastapi/security/oauth2.py
+++ b/fastapi/security/oauth2.py
@@ -398,6 +398,13 @@ class OAuth2(SecurityBase):
         self.scheme_name = scheme_name or self.__class__.__name__
         self.auto_error = auto_error
 
+    def _set_flows(
+        self,
+        *,
+        flows: OAuthFlowsModel | dict[str, dict[str, Any]],
+    ) -> None:
+        self.model.flows = cast(OAuthFlowsModel, flows)
+
     def make_not_authenticated_error(self) -> HTTPException:
         """
         The OAuth 2 specification doesn't define the challenge that should be used,
@@ -452,7 +459,7 @@ class OAuth2PasswordBearer(OAuth2):
                 [FastAPI docs for Simple OAuth2 with Password and Bearer](https://fastapi.tiangolo.com/tutorial/security/simple-oauth2/).
                 """
             ),
-        ],
+        ] = "",
         scheme_name: Annotated[
             str | None,
             Doc(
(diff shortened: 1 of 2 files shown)


YuriiMotov (commented):
@Takrim-ScrubAI, thanks for your interest!

I think default values should be not just empty strings, but some sentinel value (e.g. Pydantic's `Undefined` or [FastAPI's `Default`](https://github.com/fastapi/fastapi/blob/ca5f60ee72f35fb2134d8b5d26bbb75965bcff66/fastapi/datastructures.py#L174)). And we need to handle the situation when instance is not initialized. Currently, if we run app without `auth_code_scheme.initialize()`, it will silently produce invalid OpenAPI schema.

Also, I think we should make this consistent for all security classes

PR #15067, opened 2026-03-06
✨ Add per-router exception handlers for APIRouter

Holek:
## Summary

- Add `exception_handlers` parameter to `APIRouter`, `APIRoute`, and `APIWebSocketRoute`, allowing exception handlers to be scoped to specific routers instead of only at the app level
- Add `APIRouter.add_exception_handler()` method for imperative registration
- Router-level handlers override app-level handlers for routes within that router, with clear precedence: **route > child router > parent router > app**
- No changes to Starlette or `fastapi/applications.py` - works by merging handlers into the existing `scope["starlette.exception_handlers"]` before `wrap_app_handling_exceptions` runs

## Motivation
We hit this in our own codebase. We have a modular FastAPI app with domain-specific routers, each with its own set of domain exceptions. Our setup looked like this:

```python

from routes.exception_handlers import register_payments_exception_handlers
from routes.connection_router import router as connection_router
from routes.income_router import router as income_router
from routes.webhook_router import router as webhook_router

# ... many other routers

def register_payments_exception_handlers(app: FastAPI):
    app.add_exception_handler(InvalidUserDataError, handle_invalid_user_data)
    app.add_exception_handler(ConnectionNotFoundError, handle_connection_not_found)
    app.add_exception_handler(DuplicateConnectionError, handle_duplicate_connection)
    app.add_exception_handler(PaymentProviderError, handle_payment_provider_error)
    app.add_exception_handler(ValidationError, handle_payments_validation_error)
    app.add_exception_handler(UserNotFoundError, handle_user_not_found)


banking: APIRouter = APIRouter(prefix="/banking")
banking.include_router(connection_router)
banking.include_router(income_router)
banking.include_router(webhook_router)

# ... many other routers


def create_app() -> FastAPI:
    app = FastAPI()

    register_payments_exception_handlers(app)  # <-- leaks to ALL routes
    # ... many other routers, user, auth, etc.
    app.include_router(banking)
    # ... many other routers
    return app
```

Those payment-specific exception handlers leaked to every route in the app. A `ValidationError` raised in a completely unrelated endpoint would get caught by the payments handler, producing a confusing response. We wanted these handlers scoped to the `/banking` router only.

The workarounds are all unsatisfying:

- **Catch exceptions inside each endpoint** - defeats the purpose of centralized error handling, duplicates `try`/`except` blocks everywhere
- **Sub-applications via `app.mount()`** - breaks dependency injection, loses shared middleware, and splits the OpenAPI schema used by other teams
- **Custom middleware per router** - complex, doesn't integrate with Starlette's exception handling machinery

With this change, the code above would become:

```python
router = APIRouter(
    prefix="/banking",
    exception_handlers={
        InvalidUserDataError: handle_invalid_user_data,
        ConnectionNotFoundError: handle_connection_not_found,
        DuplicateConnectionError: handle_duplicate_connection,
        PaymentProviderError: handle_payment_provider_error,
        ValidationError: handle_payments_validation_error,
        UserNotFoundError: handle_user_not_found,
    },
)
router.include_router(connection_router)
router.include_router(income_router)
router.include_router(webhook_router)
```

## How it works

Exception handlers are attached to routes at build time (during `add_api_route` / `include_router`), merged using the same layering pattern as dependencies. At request time, route-level handlers are merged into `scope["starlette.exception_handlers"]` inside `request_response()` / `websocket_session()` — before Starlette's `wrap_app_handling_exceptions` reads them. This means route-level handlers naturally override app-level ones with zero changes to Starlette.

## Example

```python

class PaymentError(Exception):
    pass

payments_router = APIRouter(
    prefix="/payments",
    exception_handlers={
        PaymentError: lambda req, exc: JSONResponse(
            {"error": "payment_failed"}, status_code=402
        )
    },
)

@payments_router.post("/charge")
async def charge():
    raise PaymentError()

app = FastAPI()
app.include_router(payments_router)
# PaymentError is only handled for /payments/* routes
```

## Docs

- New advanced guide: `docs/en/docs/advanced/router-exception-handlers.md`
- Cross-reference added to `docs/en/docs/tutorial/handling-errors.md`
- `add_exception_handler` added to `APIRouter` reference page
- Two example files: basic usage and app-vs-router override


Holek:
I cannot seem to add labels after creating this PR, is this something that maintainers can do for me?

Holek:
Possible duplicate PR: #9374 (from long ago)

YuriiMotov:
@Holek, thank you for your interest and work!

We plan to implement several other features first and will review this PR after that

tiangolo:

Thank you! 🍰

Holek:
Thanks, Sebastián!

Glad you have it in works in the refactor! When you're done and I need it again, I'll pop up with a new PR.

Thanks for all your work on this framework (and many other things!)

Issue #13175, opened 2025-01-08
Duplicated OperationID when adding route with multiple methods


Kludex:
### Discussed in https://github.com/fastapi/fastapi/discussions/8449

<div type='discussions-op-text'>

<sup>Originally posted by **bruchar1** March 30, 2022</sup>
### First Check

- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the FastAPI documentation, with the integrated search.
- [X] I already searched in Google "How to X in FastAPI" and didn't find any information.
- [X] I already read and followed all the tutorial in the docs and didn't find an answer.
- [X] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
- [X] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
- [X] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).

### Commit to Help

- [X] I commit to help with one of those options 👆

### Example Code

```python
router.add_api_route(
    "/clear",
    clear,
    methods=["POST", "DELETE"]
)
```

### Description

Seems to be caused by #4650.

The new `generate_unique_id()` function uses `list(route.methods)[0].lower()` as suffix for the `operation_id`. Therefore, in my example, both post and delete endpoints get `_post` suffix for operation_id, causing it to no longer be unique.

It then issues a "UserWarning: Duplicate Operation ID"

### Operating System

Windows

### FastAPI Version

0.75.0

### Python Version

3.10.2

### Additional Context

_No response_</div>

Kludex:
There's a PR already for it: https://github.com/fastapi/fastapi/pull/10694/

YuriiMotov:
There is also PR #4851 that aimed to solve this, but it was closed by Sebastian with the [comment](https://github.com/fastapi/fastapi/pull/4851#issuecomment-2294657832):

> Using multiple methods for the same function is not actually supported. `@router.api_route()` is there as a semi-internal utility, and because Starlette supports multiple methods, the interface is there. But the official way to do it in FastAPI and have all the FastAPI features (including the automatic OpenAPI integration, with operation IDs) is by using the dedicated methods, like `app.post()`, `app.get()`, etc.

[...]

PR #15070, opened 2026-03-06
fix: deterministic operationId when route has multiple methods

ghost:
…#13175

Found this while going through open issues.

When a route is added with multiple methods, the operationId was being 
generated using list(route.methods)[0] - but route.methods is a set, 
which has no guaranteed order. This caused inconsistent or duplicate 
operation IDs.

Fixed by using sorted(route.methods)[0] instead, so it always picks 
the alphabetically first method and the ID is consistent every time.

Fixes #13175

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -96,7 +96,7 @@ def generate_unique_id(route: "APIRoute") -> str:
     operation_id = f"{route.name}{route.path_format}"
     operation_id = re.sub(r"\W", "_", operation_id)
     assert route.methods
-    operation_id = f"{operation_id}_{list(route.methods)[0].lower()}"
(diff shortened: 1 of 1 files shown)


ghost:
Could a maintainer please add the appropriate label? I believe this is a bug fix. Thank you :)

YuriiMotov:
@sandesh-bhandari-dev, thanks for your interest!

The approach you propose doesn't fully solve the issue as it will use the same operation_id for both operations (but they must be unique).

There is another PR https://github.com/fastapi/fastapi/pull/10694 that attempts to solve this in a more complex way. It's in the queue to review by Sebastian.

Having this, I'l close this PR

ghost:
Thanks for the explanation! I appreciate the feedback and will keep an eye on the other PR.

PR #15074, opened 2026-03-07
Fix duplicate operation IDs for routes with multiple methods

04cb:
Fixes #13175. When a route is defined with multiple methods (e.g., 
methods=["POST", "DELETE"]), the generate_unique_id function was only 
using the first method to generate the operation_id. This caused all 
methods on the same route to have the same operation_id in the OpenAPI 
schema, triggering duplicate operation_id warnings.

The fix modifies get_openapi_operation_metadata to generate unique 
operation_ids per method when a route has multiple methods, by 
appending the current method to the base unique_id.

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -243,7 +243,16 @@ def get_openapi_operation_metadata(
     operation["summary"] = generate_operation_summary(route=route, method=method)
     if route.description:
         operation["description"] = route.description
-    operation_id = route.operation_id or route.unique_id
+    if route.operation_id:
+        operation_id = route.operation_id
+    else:
+        # If the route has multiple methods, generate a unique operation_id
+        # using the current method to avoid duplicates
+        assert route.methods is not None
+        if len(route.methods) > 1:
+            operation_id = f"{route.unique_id}_{method.lower()}"
+        else:
+            operation_id = route.unique_id
     if operation_id in operation_ids:
         message = (
             f"Duplicate Operation ID {operation_id} for function "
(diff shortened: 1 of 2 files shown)


YuriiMotov (changes requested):
@04cb, thank you for your interest and efforts!

This approach has 2 problems.
Let's take a look ate the following example app:

```py
from typing import Annotated

def do_something(p: Annotated[str, Body(embed=True)]):
    return {"message": "Hi!"}

app = FastAPI()

app.add_api_route("/", do_something, methods=["POST", "PUT"])

```

**1. Operation IDs for endpoints are unique, but they are not deterministic**

```json
 "paths": {
    "/": {
      "put": {
        "summary": "Do Something",
        "operationId": "do_something__put_put",  <= There shouldn't be two put's here. And it will be sometimes put, sometimes post
        },
```

```json
      "post": {
        "summary": "Do Something",
        "operationId": "do_something__put_post",  <= There shouldn't be put here. And it will be sometimes put, sometimes post
      }
```
 
**2. If you try the app that has a request body, the body schema name will be still non-deterministic:**

```json
  "paths": {
    "/": {
      "put": {
        "summary": "Do Something",
        "operationId": "do_something__put_put",
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/Body_do_something__put"  <= This will be sometimes _put sometimes _post
              }
            }
          },
          "required": true
        },
```

---

See also: https://github.com/fastapi/fastapi/pull/10694

PR #15077, opened 2026-03-08
🐛 Fix stream item type lost when using `include_router()`

### LINKED ISSUE #15401: SSE `stream_item_type` not propagated through `APIRouter` + `include_router`
### Summary

### Reproduction

```python

class Frame(BaseModel):
    kind: str


app_a = FastAPI()

    yield Frame(kind="x")


router = APIRouter()

    yield Frame(kind="x")


app_b = FastAPI()
app_b.include_router(router)

```

Output:

```
APP (post-include): None

```

### Expected

### Actual

### Root cause (proposed)

```python
        response_model = None
    else:
        if stream_item is not None:
            if (
            response_model = None
        else:
```

```python
self.add_api_route(
    prefix + route.path,
    route.endpoint,
    ...
    ...
)
```

### Suggested fix

Either:

### Environment

- `fastapi==0.135.2`
- `pydantic==2.12.5`
- `starlette==1.0.0`
- Python `3.11.6`
- macOS 14.4.1

alex-raw:
When a typed SSE or JSONL streaming endpoint is registered via a router, the OpenAPI schema loses the stream item type information — `itemSchema` ends up without `contentSchema`/`contentMediaType` (SSE) or as an empty object (JSONL). The endpoints stream correctly at runtime, only the generated schema is affected.

This happens because `include_router()` re-creates routes via `add_api_route()`, but `stream_item_type` (computed during the first `APIRoute.__init__` from the return annotation) is not carried over. The re-created route has no way to re-derive it since `response_model` is no longer a `DefaultPlaceholder` at that point.

The fix threads `stream_item_type` through `add_api_route()` and `include_router()`, following the same pattern used for `strict_content_type`.

Discussed in https://github.com/fastapi/fastapi/discussions/15066


tests/test_sse.py
```diff
@@ -96,6 +96,18 @@ async def stream_events():
...
+@router.get("/events-typed", response_class=EventSourceResponse)
+async def stream_events_typed() -> AsyncIterable[Item]:
+    for item in items:
+        yield item
+
+
+@router.get("/events-jsonl")
+async def stream_events_jsonl() -> AsyncIterable[Item]:
+    for item in items:
+        yield item
```
YuriiMotov: This is not SSE. So, we should probably move it to another module (`test_stream_json_lines` ?)

YuriiMotov (commented):
LGTM in general!
@alex-raw, thank you for reporting this and working on this!

The logic around `stream_item_type` looks a bit fragile to me..

If `response_class` is set in parent router or app, then `self.stream_item_type` will be calculated and stored, but not used in some cases (if response class is neither `DefaultPlaceholder` nor `EventSourceResponse`).

I didn't manage to come up with particular cases, but this separation of logic seems to be error-prone (or, at least prone to confusion).

It all seem to work well now, but to prevent breaking this in the future, I suggest we add tests with `default_response_class` being specified for router, parent router or app.

@alex-raw, what do you think?

alex-raw (commented):
Thanks for the review!

Good point about the JSONL test, I'll move it out of test_sse.py. Where would you prefer it? I was thinking maybe test_stream_bare_type.py since it already covers JSONL edge cases, but happy to put it wherever fits best.

Will also add tests for the default_response_class edge case.

tests/test_sse.py
alex-raw: Moved to test_stream_bare_type.py. Also added default_response_class tests for both app-level and parent router cases.

BillionClaw:
Looking into this — the stream_item_type needs to be threaded through include_router(). Happy to submit a fix.

tests/test_sse.py
```diff
@@ -316,3 +361,97 @@ def test_no_keepalive_when_fast(client: TestClient):
...
+async def default_app_stream() -> AsyncIterable[Item]:
+    yield items[0]
+
+
+default_app_app.include_router(default_app_router, prefix="/api")
+
+
+def test_default_response_class_on_app_stream():
+    with TestClient(default_app_app) as c:
+        response = c.get("/api/stream")
```
YuriiMotov: I would rename `c` to `client` here and in other test functions.
Also, you can actually get OpenAPI schema directly as `default_app_app.openapi()`

tests/test_sse.py
```diff
@@ -316,3 +361,97 @@ def test_no_keepalive_when_fast(client: TestClient):
...
+
+# default_response_class on app
+
+default_app_app = FastAPI(default_response_class=EventSourceResponse)
+default_app_router = APIRouter()
+
+
+@default_app_router.get("/stream")
+async def default_app_stream() -> AsyncIterable[Item]:
+    yield items[0]
```
YuriiMotov: Maybe better make it return the same as `stream_events_typed` above for consistency.
And the same for `default_parent_stream`.

But not critical for me

YuriiMotov (approved):
Added a couple of minor suggestions, but it looks good in general!

Will pass it to Sebastian.

@alex-raw, thanks!

tests/test_sse.py
alex-raw: Done, thanks!

tests/test_sse.py
alex-raw: Good point, updated both.

bernatGene:
facing this issue, would appreciate merging/moving forward to remove current workarounds in my project:) what is missing/what is the expected timeline here?

Pike:
I tested this locally from Alex' branch, and it worked.

LordGaav:
Can confirm that applying the patches to `routing.py` fixes the issue of the schema not appearing in the OpenAPI spec.

The only other quirk (which is probably outside the scope of this PR) is that itemSchema is new in OpenAPI version 3.2: https://learn.openapis.org/specification/media-types.html . Neither of the built-in doc tools support it yet, and FastAPI also claims it's returning schema version 3.1.0 .

NathanCummings:
Confirming this on a real project (bulk NDJSON `/export` endpoints), and flagging that the impact is broader than
> the endpoints stream correctly at runtime, only the generated schema is affected

for JSONL, generator endpoints the runtime serialization is affected too. Once the route goes through `include_router`, `response_model_exclude_none` (and, by extension, every `response_model_*` option: `by_alias`, `include`/`exclude`, response validation) is silently ignored, so the streamed rows diverge from the equivalent non-streaming list endpoint.

It's the same root cause this PR targets: through `include_router` the merged route is rebuilt with the already-resolved `response_model=None`, so `stream_item_type` / `stream_item_field` is lost. At request time `_serialize_data` then takes the bare `jsonable_encoder` branch instead of `stream_item_field.serialize_json(...)`, so nothing from the response-model config is applied.

Minimal example - the same generator route added directly on the app vs. via `include_router`:

```python
from collections.abc import Iterable

class Row(BaseModel):
    id: int
    name: str
    optional: str | None = None   # unset -> None, exercises exclude_none

def gen() -> Iterable[Row]:
    yield Row(id=1, name="a")

# (A) route added directly on the app
app_a = FastAPI()
app_a.get("/direct", response_model_exclude_none=True)(gen)

# (B) same route via include_router
router = APIRouter()
router.get("/inc", response_model_exclude_none=True)(gen)
app_b = FastAPI()
app_b.include_router(router, prefix="/api")

for name, app, path in [("DIRECT", app_a, "/direct"), ("INCLUDE_ROUTER", app_b, "/api/inc")]:
    body = TestClient(app).get(path).text.strip()
    item = app.openapi()["paths"][path]["get"]["responses"]["200"] \
                .get("content", {}).get("application/jsonl", {}).get("itemSchema", "MISSING")
    print(f"{name:14s} excludes_null={'optional' not in body!s:5s} itemSchema={item}")
```

Output - identical on 0.138.2 and 0.139.2:

```bash
DIRECT          excludes_null=True   itemSchema={'$ref': '#/components/schemas/Row'}
INCLUDE_ROUTER  excludes_null=False  itemSchema={}
```

So propagating `stream_item_type` through `include_router` (this PR) should fix both facets in one go - the empty `itemSchema` *and* the dropped `exclude_none`/`response_model_*` at runtime. Might be worth adding a runtime assertion to the test matrix (that the streamed body honours `response_model_exclude_none`), not just the OpenAPI schema, so the serialization behaviour is locked in too.

Related: #15401 (SSE schema), #15093 / #15089 (`response_model_*` ignored for `Iterable[..]` returns).

tiangolo (approved):

tiangolo:
Great, thank you @alex-raw! :rocket: 

Thanks @YuriiMotov for the review and handling, and thanks @NathanCummings for the extra case, added it to the tests.

Thanks all for checking and verifying it works!

This will be available in FastAPI 0.140.8 in the next few hours. :tada:

pmeier:
Just to confirm what @LordGaav [wrote above](https://github.com/fastapi/fastapi/pull/15077#issuecomment-4630404340)

While the data is now correct in the OpenAPI schema, neither Swagger nor ReDoc are showing anything apart from the media type. Is there an issue already to update Swagger / ReDoc?

YuriiMotov:
> While the data is now correct in the OpenAPI schema, neither Swagger nor ReDoc are showing anything apart from the media type. Is there an issue already to update Swagger / ReDoc?

Thanks! We are tracking this internally

PR #15078, opened 2026-03-08
Fixed the doc for the "Required, can be None"

Amrito20:
Hi, this is my first open source contribution. Can I open a PR?

### Summary

This PR clarifies the **"Required, can be None"** section in query parameter string validations docs.

### Changes

- changed the `Required, can be None` section
- refactored example using modern `Annotated` syntax with `= ...`
- explained the difference between `= None` and `= ...`


YuriiMotov:
@Amrito20, thanks for your interest!

This approach is not correct - it will not allow you to pass `None` value.

There is already https://github.com/fastapi/fastapi/pull/14322 that resolves this

PR #15090, opened 2026-03-10
📝 Update .gitignore and add skipif decorator for fastapi-cli tests

serge-eric-kalaga:
## Problem
`test_fastapi_cli` runs `fastapi dev non_existent_file.py` in subprocess and waits
"Path does not exist non_existent_file.py". But if `fastapi-cli` is not installed
(without the extra `[standard]`), the subprocess displays the installation message instead,
which causes the test to fail.

## Fix
Added a `pytest.mark.skipif` that skips `test_fastapi_cli` when `fastapi.cli.cli_main is None`.
The `test_fastapi_cli_not_installed` test already covers this case via mock, no coverage is lost.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,8 @@ log.txt
 Pipfile.lock
 env3.*
 env
+.env
+.venv
 docs_build
 site_build
 venv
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@serge-eric-kalaga, thanks for your interest!

I think we don't need this.
It's said in [development docs](https://fastapi.tiangolo.com/contributing/) that you should install dependencies with `uv sync --extra all`

serge-eric-kalaga:
@YuriiMotov Exactly. Thank you.

PR #15091, opened 2026-03-10
👷 Add `ty` to precommit

svlandeg:
* Add `ty` to pyproject, precommit and lint script
* `mypy` is also kept for now
* Update code to pass `ty`
    * 150 original issues
    * bulk of those were issues with `ty` complaining about `_Unset` default assignments
    * quite some where `ty` sees no issue though `mypy` does: these need an extra ignore on the same line - ugly but we can remove these when we remove `mypy` alltogether
    * then there are a handful where `ty` sees a problem that `mypy` thinks is fine, here we add `ty: ignore` statements that are ignored by `mypy`.

[UPDATE]: note that ignored deprecation warnings are handled more properly in cf https://github.com/fastapi/fastapi/pull/15101 🙂


fastapi/security/http.py
```diff
@@ -75,9 +77,7 @@ def __init__(
         description: str | None = None,
         auto_error: bool = True,
     ):
-        self.model: HTTPBaseModel = HTTPBaseModel(
-            scheme=scheme, description=description
-        )
+        self.model = HTTPBaseModel(scheme=scheme, description=description)
```
svlandeg, line 80: Here (and in a few similar cases), I've moved the typing to the var on the `class`, assuming that it wouldn't get a different type at any time. By moving the typing in this way, type checkers can more easily deduce the correct type of class vars.

fastapi/_compat/v2.py
```diff
@@ -22,10 +22,10 @@
 from pydantic import PydanticSchemaGenerationError as PydanticSchemaGenerationError
 from pydantic import PydanticUndefinedAnnotation as PydanticUndefinedAnnotation
 from pydantic import ValidationError as ValidationError
-from pydantic._internal._schema_generation_shared import (  # type: ignore[attr-defined]
+from pydantic._internal._schema_generation_shared import (  # type: ignore[attr-defined]  # ty: ignore[unused-ignore-comment]
     GetJsonSchemaHandler as GetJsonSchemaHandler,
 )
-from pydantic._internal._typing_extra import eval_type_lenient
+from pydantic._internal._typing_extra import eval_type_lenient  # ty: ignore[deprecated]
```
YuriiMotov, line 28: Should we follow the deprecation recommendations and use `try_eval_type` directly?

https://github.com/pydantic/pydantic/blob/938e81de2100b278f835d6e054bb7ecd58bcaff0/pydantic/_internal/_typing_extra.py#L445-L455

fastapi/encoders.py
```diff
@@ -22,7 +22,7 @@
 from fastapi.exceptions import PydanticV1NotSupportedError
 from fastapi.types import IncEx
 from pydantic import BaseModel
-from pydantic.color import Color
+from pydantic.color import Color  # ty: ignore[deprecated]
```
YuriiMotov, line 25: We should probably address this deprecation warning.
I would actually introduce a breaking change and replace this import with `from pydantic_extra_types.color import Color` (maybe not in this PR).

fastapi/param_functions.py
```diff
@@ -34,7 +34,7 @@ def Path(  # noqa: N802
             The parameter is available only for compatibility.
             """
         ),
-    ] = _Unset,
+    ] = _Unset,  # ty: ignore[invalid-parameter-default]
```
YuriiMotov: We can probably try replacing this with [`Default(None)`](https://github.com/fastapi/fastapi/blob/11614be9021aa4ac078d4d0693a8b5250a1010d8/fastapi/datastructures.py#L174)

fastapi/routing.py
```diff
@@ -453,7 +453,7 @@ async def app(request: Request) -> Response:
         solved_result = await solve_dependencies(
             request=request,
             dependant=dependant,
-            body=body,
+            body=body,  # ty: ignore[invalid-argument-type]
```
YuriiMotov: I think we should resolve this.
Probably by updating the type hint of `solve_dependencies`'s `body` parameter to `Any` (or union of all possible types) and handling this inside

fastapi/routing.py
```diff
@@ -635,7 +635,7 @@ async def _async_stream_jsonl() -> AsyncIterator[bytes]:
                 else:
 
                     def _sync_stream_jsonl() -> Iterator[bytes]:
-                        for item in gen:
+                        for item in gen:  # ty: ignore[not-iterable]
```
YuriiMotov: ?

fastapi/dependencies/utils.py
```diff
@@ -321,7 +323,7 @@ def get_dependant(
             ):
                 assert dependant.call
                 raise DependencyScopeError(
-                    f'The dependency "{dependant.call.__name__}" has a scope of '
+                    f'The dependency "{dependant.call.__name__}" has a scope of '  # ty: ignore[unresolved-attribute]
                     '"request", it cannot depend on dependencies with scope "function".'
                 )
```
YuriiMotov, line 329: I think we should address this, not just ignore (class instance with `__call__` method will not have `__name__` attribute, but it will still be `Callable`):

```
                dep_name = getattr(dependant.call, "__name__", repr(dependant.call))
                raise DependencyScopeError(
                    f'The dependency "{dep_name}" has a scope of '
                    '"request", it cannot depend on dependencies with scope "function".'
                )
```

fastapi/dependencies/utils.py
```diff
@@ -826,8 +828,9 @@ def request_params_to_args(
 
     for key in received_params.keys():
         if key not in processed_keys:
-            if hasattr(received_params, "getlist"):
-                value = received_params.getlist(key)
+            getlist = getattr(received_params, "getlist", None)
+            if callable(getlist):
+                value = getlist(key)
```
```suggestion
            if isinstance(received_params, (ImmutableMultiDict, Headers)):
                value = received_params.getlist(key)
```

As it's done in `_get_multidict_value`

fastapi/openapi/utils.py
```diff
@@ -247,7 +247,7 @@ def get_openapi_operation_metadata(
     if operation_id in operation_ids:
         message = (
             f"Duplicate Operation ID {operation_id} for function "
-            + f"{route.endpoint.__name__}"
+            + f"{route.endpoint.__name__}"  # ty: ignore[unresolved-attribute]
```
YuriiMotov: Same as https://github.com/fastapi/fastapi/pull/15091/changes#r2919360642 ?

fastapi/applications.py
```diff
@@ -1006,11 +1002,12 @@ class Item(BaseModel):
...
         )
+
+        # Starlette still has incorrect type specification for the handlers
         self.exception_handlers.setdefault(
             WebSocketRequestValidationError,
-            # Starlette still has incorrect type specification for the handlers
-            websocket_request_validation_exception_handler,  # type: ignore
-        )
+            websocket_request_validation_exception_handler,  # type: ignore   # ty: ignore[unused-ignore-comment]
+        )  # ty: ignore
```
```suggestion
            websocket_request_validation_exception_handler,  # type: ignore[arg-type]  # ty: ignore[unused-ignore-comment]
        )  # ty: ignore[no-matching-overload]
```

I think we should try to avoid too broad ignore comments

fastapi/routing.py
```diff
@@ -100,8 +100,8 @@ def request_response(
     and returns an ASGI application.
     """
     f: Callable[[Request], Awaitable[Response]] = (
-        func if is_async_callable(func) else functools.partial(run_in_threadpool, func)  # type:ignore
-    )
+        func if is_async_callable(func) else functools.partial(run_in_threadpool, func)  # type:ignore  # ty: ignore[unused-ignore-comment]
+    )  # ty: ignore
```
```suggestion
    f: Callable[[Request], Awaitable[Response]] = (
        func  # type: ignore[assignment]  # ty: ignore[unused-ignore-comment]
        if is_async_callable(func)
        else functools.partial(run_in_threadpool, func)  # type: ignore[call-arg]  # ty: ignore[unused-ignore-comment]
    )  # ty: ignore[invalid-assignment]
```

I think we should specify rules to ignore

YuriiMotov (commented):
LGTM in general!

Added several suggestions in review comments. Can help with the implementation if needed)
@svlandeg, what do you think?

fastapi/_compat/v2.py
svlandeg: I thought about that, but it would make us incompatible with older v2 versions. Instead I think if we want to address this deprecation, we should do it in a separate PR and by properly supporting all versions in the `compat` module.

fastapi/encoders.py
svlandeg: I agree, see previous comment (we could do both in a separate PR)

fastapi/routing.py
svlandeg: Yes, totally agree!

fastapi/_compat/v2.py
svlandeg: cf https://github.com/fastapi/fastapi/pull/15101 🙂

fastapi/encoders.py
svlandeg: cf https://github.com/fastapi/fastapi/pull/15101 🙂

fastapi/routing.py
svlandeg: done, cf [this commit](https://github.com/fastapi/fastapi/pull/15091/commits/e6378e844e7a8cd82473a23b714ac979361132a3)

fastapi/dependencies/utils.py
svlandeg: Like [this](https://github.com/fastapi/fastapi/pull/15091/changes/8f896037945e6d22019fff6845c44290455461fc)?

fastapi/openapi/utils.py
svlandeg: cf [this commit](https://github.com/fastapi/fastapi/pull/15091/changes/b2798dab90507ce04b98233c21fd77420061d989)

fastapi/param_functions.py
svlandeg: Great idea! see [this commit](https://github.com/fastapi/fastapi/pull/15091/changes/11a0f1bf3cd3c7c5b4c0e401347481676111e2cc)

svlandeg:
Thanks so much for the thorough review @YuriiMotov! I've addressed all comments in detail, let me know if there's any other concerns!

tiangolo (approved):
This looks awesome, thank you @svlandeg ! 🙌 

And thanks @YuriiMotov for the thorough review too!

PR #15093, opened 2026-03-11
🐛 Fix `response_model_*` params ignored for non-generator endpoints with `Iterable[..]` return type

YuriiMotov:
See https://github.com/fastapi/fastapi/discussions/15089

Currently, when a non-generator endpoint is annotated with `Iterable[T]` 

```py
def get_items() -> Iterable[Item]:
    return [...]
```

`response_model` is set to `None` [here](https://github.com/fastapi/fastapi/blob/11614be9021aa4ac078d4d0693a8b5250a1010d8/fastapi/routing.py#L861).

This causes `serialize_response()` to fall back to `jsonable_encoder()`, which ignores all `response_model_*` serialization options (`exclude_none`, `exclude_unset`, `exclude_defaults`, etc.).

To fix this we need to add a `and is_generator` condition to `if stream_item is not None:` ([here](https://github.com/fastapi/fastapi/blob/11614be9021aa4ac078d4d0693a8b5250a1010d8/fastapi/routing.py#L849)).

This is done in 2 steps:

1. (commit [d903890](https://github.com/fastapi/fastapi/pull/15093/changes/d903890085ba22c7d1e358ab87888111806a48e2)) Moved the `response_model` resolution and `response_field` / `stream_item_field` creation after `self.dependant` is built, so that a single `is_generator` is used for both stream detection and response model resolution.

2. (commit [e8c00f5](https://github.com/fastapi/fastapi/pull/15093/changes/e8c00f5a02fa3aeb5b3c93233d32abdaecb536bc)) Added `and is_generator` to the `if stream_item is not None:` condition.


tiangolo (approved):
Nice, thank you! This will be available in FastAPI 0.140.11 🎉

PR #15095, opened 2026-03-11
refactor: jsonable_encoder to remove redundant custom_encoder assignment

eyetheekh:
This refactor removes a redundant assignment of `custom_encoder` in `jsonable_encoder`.

Previously:

    if custom_encoder:

If `custom_encoder` is `None`, the first line converts it to `{}`, and the subsequent condition evaluates to `False`. The fallback assignment is therefore unnecessary.

This change simplifies the logic by checking `custom_encoder is not None` directly.

This PR is a refactor; please apply the refactor label.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -211,8 +211,7 @@ def jsonable_encoder(
     Read more about it in the
     [FastAPI docs for JSON Compatible Encoder](https://fastapi.tiangolo.com/tutorial/encoder/).
     """
-    custom_encoder = custom_encoder or {}
-    if custom_encoder:
+    if custom_encoder is not None:
         if type(obj) in custom_encoder:
             return custom_encoder[type(obj)](obj)
         else:


fastapi/encoders.py
```diff
@@ -211,8 +211,7 @@ def jsonable_encoder(
     Read more about it in the
     [FastAPI docs for JSON Compatible Encoder](https://fastapi.tiangolo.com/tutorial/encoder/).
     """
-    custom_encoder = custom_encoder or {}
-    if custom_encoder:
+    if custom_encoder is not None:
```
YuriiMotov, line 214: This changes the logic: previously this branch wouldn't be executed if `custom_encoder` is empty dict, after this change it will be executed.
This will not change the result, but I just don't see a point in such refactoring

YuriiMotov:
@eyetheekh, thanks for your interest, but I think we don't need this

PR #15096, opened 2026-03-11
Fix Annotated with ForwardRef when using future annotations

AhsanSheraz:
## Summary

- Fixes issue where `Annotated[SomeClass, Depends()]` with forward references failed when using `from __future__ import annotations` (the type was treated as a query parameter instead of a dependency)
- Uses a lenient namespace evaluation strategy: when strict resolution of the annotation string fails, a `_LenientDict` namespace resolves undefined names to `Any`, preserving the `Annotated` structure so FastAPI can extract `Depends` metadata
- Only applies the lenient result when `Depends` metadata is present — for other metadata types (`Query`, `Body`, etc.) the actual type is needed, so existing fallback behaviour is kept

## Changes

- **`fastapi/dependencies/utils.py`**: Added `_LenientDict` (dict subclass with `__missing__` returning `Any`), `_try_resolve_annotated()` helper, and modified `get_typed_annotation()` to handle `Annotated` string annotations
- **`tests/test_annotated_forward_ref.py`**: New test file with 12 tests covering the exact issue from #13056, unit tests for all code paths in the new helpers, and regression tests for the OpenAPI schema

## Approach comparison with PR #14557

This takes a cleaner approach than #14557:
- No string parsing with bracket-depth tracking
- No `eval()` on extracted substring parts — evaluates the full annotation string (same as Python does internally)
- Handles nested generics (`List[Potato]`, `Dict[str, Potato]`) naturally
- Fewer code paths = achieves 100% coverage of new code
- No module namespace merging via `importlib.import_module` (doesn't help at decoration time)

Fixes #13056

## Test plan

- [x] Reproduces exact issue from #13056 (forward-ref dataclass with `Annotated[Potato, Depends()]`)
- [x] Verifies OpenAPI schema doesn't expose dependency as query parameter
- [x] Tests `_try_resolve_annotated` with: fully resolvable annotations, unresolvable type + Depends, unresolvable type + non-Depends (Query), completely invalid strings
- [x] All existing dependency tests pass (132 tutorial tests + annotation tests)
- [x] 100% coverage of new code


AhsanSheraz:
Hi @tiangolo, could you please add the bug label to this PR? This fixes issue https://github.com/fastapi/fastapi/issues/13056 where Annotated with ForwardRef doesn't work when using from __future__ import annotations. Thank you!

PR #15079, opened 2026-03-08
🚸 Add missing password format to `OAuth2PasswordRequestFormStrict`

cluster2600:
UPD by @YuriiMotov: same as https://github.com/fastapi/fastapi/pull/11032, but for `OAuth2PasswordRequestFormStrict` (we missed it that time)

<details><summary>Original description</summary>

## Description

`OAuth2PasswordRequestFormStrict` was missing `json_schema_extra={"format": "password"}` on its `password` and `client_secret` form fields, whilst the non-strict `OAuth2PasswordRequestForm` already had them.

This meant the Swagger UI rendered those fields as plain text inputs instead of masked password fields when using the strict variant. The fix simply adds the same `json_schema_extra` annotation that the parent class already uses.

This is part of the broader effort to improve the security utilities mentioned in the roadmap (#10370) — specifically around making the existing authentication tooling more consistant and reliable.

## Checklist

- [x] I read and followed the [Contributing Guidelines](https://fastapi.tiangolo.com/contributing/).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

## Test plan

- Updated the OpenAPI schema snapshots in `test_security_oauth2.py`, `test_security_oauth2_optional.py`, and `test_security_oauth2_optional_description.py` to verify that the `password` and `client_secret` fields now include `"format": "password"` in the generated schema.
- All 30 existing tests across the three files continue to pass.

## Bug fix

```python
from fastapi import Depends, FastAPI
from fastapi.security import OAuth2PasswordRequestFormStrict
from fastapi.testclient import TestClient

app = FastAPI()

@app.post("/login")
def login(form_data: OAuth2PasswordRequestFormStrict = Depends()):
    return form_data

client = TestClient(app)
schema = client.get("/openapi.json").json()
props = schema["components"]["schemas"]["Body_login_login_post"]["properties"]

# Before this fix: no "format" key
assert "format" not in props["password"]         # True before fix
assert "format" not in props["client_secret"]    # True before fix

# After this fix: "format": "password" is present
# assert props["password"]["format"] == "password"
# assert props["client_secret"]["format"] == "password"
```

</details>

PR #15099, opened 2026-03-12
Added password format to Oauth2PasswordRequestFormStrict fields

Maheedhar-rao:
When using OAuth2PasswordRequestFormStrict in the Swagger UI docs, the password and client_secret fields were showing as plain text input boxes instead of masked password fields. OAuth2PasswordRequestForm already had this fixed, but since OAuth2PasswordRequestFormStrict overrides __init__ separately, the fix never carried over. Added json_schema_extra={"format": "password"} to both fields to match the behavior of OAuth2PasswordRequestForm.


YuriiMotov:
Thanks!

Duplicate of https://github.com/fastapi/fastapi/pull/15079

PR #15101, opened 2026-03-12
⬆️ Update Pydantic v2 code to address deprecations

svlandeg:
When running `ty` on the code base (cf #15091), we got two valid deprecation warnings:

### eval_type_lenient

- `eval_type_lenient` has been deprecated since [pydantic#1053](https://github.com/pydantic/pydantic/pull/10530), which was released as part of [v2.10.0b1](https://github.com/pydantic/pydantic/releases/tag/v2.10.0b1).

### color 
- `pydantic.color.Color` has been deprecated since [pydantic#6003](https://github.com/pydantic/pydantic/pull/6003), which was released as part of [v2.0b3](https://github.com/pydantic/pydantic/releases/tag/v2.0b3).
- This PR updates `ENCODERS_BY_TYPE` to also cover the new `pydantic_extra_types.color.Color`. The added unit test fails on `master`.

## Open question

~~As Pydantic updates & makes older functionality deprecated, we could (also in the future) be tempted to update the methods (especially when `ty` tells us to), without realizing we might be breaking older versions that we still support according to our `pyproject.toml`. For this reason, I'd want to add a specific test CLI run with Pydantic 2.7.0 to ensure it keeps working as expected. Aye? Nay?~~

UPDATE: cf https://github.com/fastapi/fastapi/pull/15139!


fastapi/_compat/v2.py
```diff
@@ -38,7 +38,17 @@
...
 Undefined = PydanticUndefined
-evaluate_forwardref = eval_type_lenient
+
+
+def define_forwardref() -> Callable[..., Any]:
+    # eval_type_lenient has been deprecated since Pydantic v2.10.0b1 (PR #10530)
+    try_eval_type = getattr(_pydantic_typing_extra, "try_eval_type", None)
+    if try_eval_type is not None:
+        return lambda *args: try_eval_type(*args)[0]
+    return _pydantic_typing_extra.eval_type_lenient  # pragma: no cover
```
svlandeg: Added "no cover" because this line is only used for older versions of Pydantic.

Ideally, I'd love to add a test configuration to test Pydantic 2.7.0 instead on the CI, to ensure we never break older, supported versions, cf "open question" in PR description ☝️

AhsanSheraz (approved):
Great work @svlandeg! Clean approach to handling both deprecations 👍

A couple of thoughts:

1. **`Color` scoping** — Since `Color` is only used in `encoders.py` for the `ENCODERS_BY_TYPE` dict, the try/except fallback could live directly there instead of routing through `_compat/__init__.py`. That keeps the compat layer focused on core Pydantic abstractions (model fields, type evaluation, schema generation) rather than becoming a catch-all for deprecated imports. Minor though.

2. **Optional dep risk** — `pydantic-extra-types` is an optional dependency (under `standard`/`all`), so the `pydantic.color.Color` fallback is genuinely needed. Worth a small TODO comment noting that if Pydantic removes `pydantic.color` entirely in a future major, this fallback breaks silently.

3. **Success boolean from `try_eval_type`** — The lambda correctly discards it (matching `eval_type_lenient` behavior), but worth noting as a potential follow-up: `get_typed_annotation()` in `dependencies/utils.py` could use that boolean for better error messages when ForwardRef resolution fails silently.

4. **Naming nit** — `define_forwardref()` reads like it defines a forward ref, something like `_get_forwardref_evaluator()` might be clearer about what it actually does (selects the right eval function). Very minor.

None of these are blockers — the core logic is solid, the `try_eval_type` wrapper is correct, and CI is all green. Really like the idea of proactively addressing these before `ty` enforcement lands. The open question about adding a Pydantic 2.7.0 CI run is a good one too — the `lowest-direct` matrix entry might already cover it, worth verifying.

LGTM! 🚀

svlandeg:
@AhsanSheraz: thanks for the review! Did you write this, or did you produce this with an LLM?

AhsanSheraz:
> @AhsanSheraz: thanks for the review! Did you write this, or did you produce this with an LLM?

You are welcome. Base message written by me, just format and make it more nice with gpt

svlandeg:
FYI - I don't think there's a need to format your reviews with an LLM. Genuine, human reviews are more valuable. Not to say you can't use LLMs - I do too in reviewing, to double check whether I've missed anything. But ultimately, a human's proper, in-depth review is worth more (IMO).

AhsanSheraz:
> FYI - I don't think there's a need to format your reviews with an LLM. Genuine, human reviews are more valuable. Not to say you can't use LLMs - I do too in reviewing, to double check whether I've missed anything. But ultimately, a human's proper, in-depth review is worth more (IMO).

Make sense, I will take care of this next time. Thank you for your valuable insights

fastapi/_compat/v2.py
```diff
@@ -38,7 +38,25 @@
...
 Undefined = PydanticUndefined
-evaluate_forwardref = eval_type_lenient  # ty: ignore[deprecated]
+
+# pydantic.color.Color is deprecated since v2.0b3
+try:
+    from pydantic_extra_types.color import Color
+except ImportError:  # pragma: no cover
+    from pydantic.color import (  # type: ignore[assignment]
+        Color,  # noqa: F401
+    )
```
YuriiMotov: LGTM, but **note that this change is breaking**.

---

Alternatively we can probably do something like (in `encoders.py`):

```py
try:
    from pydantic.color import Color  # ty: ignore[deprecated]
except ImportError:  # pragma: no cover
    class Color:
        pass

try:
    from pydantic_extra_types.color import Color as PyExtraColor
except ImportError:  # pragma: no cover
    class PyExtraColor:
        pass

...

ENCODERS_BY_TYPE: dict[type[Any], Callable[[Any], Any]] = {
    bytes: lambda o: o.decode(),
    Color: str,  # ty: ignore[deprecated]
    PyExtraColor: str,
    ...
```
(I haven't checked it)

This way we support both `Color` implementations

fastapi/_compat/v2.py
```diff
@@ -38,7 +38,25 @@
...
+
+
+def define_forwardref() -> Callable[..., Any]:
```
YuriiMotov: LGTM
Just a note: this way we lose type hints for `evaluate_forwardref`. Not sure if it's important.

As an alternative, we can try something like:

```
try:
    from pydantic._internal._typing_extra import try_eval_type
except:
    try_eval_type = None

try:
    from pydantic._internal._typing_extra import eval_type_lenient
except:
    eval_type_lenient = None


def evaluate_forwardref(
    value: Any,
    globalns: GlobalsNamespace | None = None,
    localns: MappingNamespace | None = None,
) -> Any:
    if try_eval_type is not None:
        return try_eval_type(value, globalns, localns)[0]
    if eval_type_lenient is not None:
        return eval_type_lenient(value, globalns, localns)
    raise RuntimeError("Both try_eval_type and eval_type_lenient are None")  # pragma: no cover  # Should never happen
```

But not sure it's worth it

YuriiMotov (approved):
LGTM!

Added optional suggestion for handling `Color`

fastapi/_compat/v2.py
YuriiMotov: I investigated this. Currently tests don't run on lowest Pydantic version (2.7) because of `pydantic-ai` optional dependency.
`uv` fails to understand that this dependency is optional and considers it when resolving dependencies. So, as `pydantic-ai` depends on `pydantic 2.12`, it installs `pydantic 2.12`.
I didn't manage to find a way to solve this.

<details>

```
[project]
name = "test_resolution"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.10"
dependencies = [
    "pydantic>=2.7.0",
]

[dependency-groups]
translations = [
    "pydantic-ai>=1.68.0",  # Requires Pydantic 2.12
]
```

`uv sync --no-dev --resolution lowest-direct`

This will install pydantic 2.12 (even though `pydantic-ai` is not installed).
Comment `pydantic-ai` dependency and it will install `pydantic` 2.7 (you will need to delete `uv.lock`)

</details>

So, maybe for now we can just manually add such configuration in test matrix?

tests/test_jsonable_encoder.py
```diff
@@ -311,3 +311,21 @@ class Model(BaseModel):
...
     data = {"value": Undefined}
     assert jsonable_encoder(data) == {"value": None}
+
+
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
+@pytest.mark.parametrize(
+    "module_path",
+    [
+        pytest.param("pydantic.color"),
+        pytest.param("pydantic_extra_types.color"),
```
svlandeg, line 321: For `pydantic_extra_types.color`, this will fail on `master` with

>  ValueError: [TypeError("'Color' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]

fastapi/_compat/v2.py
svlandeg: Thanks for catching this @YuriiMotov! You're totally right. I commited your suggestion and added a unit test that fails on `master`, so we now support both `Color` classes explicitely.

fastapi/_compat/v2.py
svlandeg: Separated this into its own PR: #15139

fastapi/_compat/v2.py
svlandeg: I do think it's an improvement to have the types here explicitely.

`GlobalsNamespace` and `MappingNamespace` are typing variants of `dict[str, Any]` and `Mapping[str, Any]` respectively, defined in [_namespace_utils.py](https://github.com/pydantic/pydantic/blob/main/pydantic/_internal/_namespace_utils.py). This file only exists since Pydantic 2.10 though. In Pydantic 2.9, `eval_type_lenient` was typed twice with `dict[str, Any]`.

So I've opted here to use `dict[str, Any]` as well, to support all Pydantic versions and not requireing external imports and more if/else logic. What do you think?

svlandeg:
Thanks for the detailed review Yurii, I addressed everything, let me know what you think!

fastapi/_compat/v2.py
```diff
@@ -38,7 +38,20 @@
...
 Undefined = PydanticUndefined
-evaluate_forwardref = eval_type_lenient  # ty: ignore[deprecated]
+
+
+def evaluate_forwardref(
+    value: Any,
+    globalns: dict[str, Any] | None = None,
+    localns: dict[str, Any] | None = None,
+) -> Any:
+    # eval_type_lenient has been deprecated since Pydantic v2.10.0b1 (PR #10530)
```
YuriiMotov, line 48: As an alternative, we can use `try...except`:

```py
    try:    
        return _pydantic_typing_extra.try_eval_type(value, globalns, localns)[0]
    except AttrbuteError:
        return _pydantic_typing_extra.eval_type_lenient(  # ty: ignore[deprecated]
            value, globalns, localns
        )  # pragma: no cover
```
Or move this to module level, and only use it inside `evaluate_forwardref` (so that `evaluate_forwardref` would make it typed).

It might also be a bit faster (but it's only used at app stratup, so maybe not as important)

tests/test_jsonable_encoder.py
YuriiMotov: > For pydantic_extra_types.color, this will fail on master with
> 
> ValueError: [TypeError("'Color' object is not iterable"), TypeError('vars() argument must have dict attribute')]
> 

I took me some time to understand this comment :sweat_smile:

Just to clarify if somebody also gets confused: it fails on `master` because `pydantic_extra_types.color.Color` is not currently supported by `jsonable_encoder`. This PR adds such support

YuriiMotov (approved):
LGTM!

fastapi/_compat/v2.py
svlandeg: Yea I think it's a style preference, I'm personally more fan of checking explicitely first rather than relying on a `try`-`except`, but both are valid options. I'll leave it up to Sebastián to decide if he's got a preference 🙂

tiangolo (approved):
Nice, thank you! 🙌

fastapi/_compat/v2.py
Viicos: I found it a bit unfortunate that this PR still relies on an internal function.. The only reason I kept `eval_type_lenient()` in Pydantic was because FastAPI was using it, and now the same issue exists with `try_eval_type()`.

svlandeg: @Viicos: right, good point. What would be the better way to address this? I can make a new PR.

Viicos: I think it might be worth solving this cleanly, that is by first having the ability to do such evaluations using the stdlib. I proposed https://discuss.python.org/t/100698 a while ago, I'll try to continue working on it. Then we could see how can this be used in FastAPI/Pydantic!

PR #15115, opened 2026-03-14
Fix get_model_fields crash when field annotation is Annotated[BaseModel, ...]

xovishnukosuri:
## Summary

- `get_model_fields` in `fastapi/_compat/v2.py` crashes with `PydanticUserError` when a pydantic `RootModel` is parameterised with `Annotated[SomeBaseModel, Field(...)]`.
- Root cause: pydantic stores the root field annotation verbatim as `Annotated[SomeBaseModel, FieldInfo(...)]`. The existing `lenient_issubclass(type_, (BaseModel, dict))` check returns `False` for the `Annotated` wrapper, causing `model_config` to be forwarded to `ModelField` → `TypeAdapter`. Pydantic rejects `config` when the inner type is a `BaseModel`.
- Fix: unwrap one level of `Annotated` before the `issubclass` check, using `get_origin` / `get_args` from the standard `typing` module (no private pydantic internals).

## Reproduction

```python
from typing import Annotated
from pydantic import BaseModel, Field, RootModel
from fastapi import FastAPI

class Item(BaseModel):
    name: str

class Wrapper(RootModel[Annotated[Item, Field(description="an item")]]):
    pass

app = FastAPI()

@app.get("/test", response_model=Wrapper)
def endpoint():
    return Wrapper(root=Item(name="foo"))

client = TestClient(app)
# Before fix: PydanticUserError: Cannot use `config` when the type is a BaseModel...
# After fix: 200 OK
print(client.get("/openapi.json").status_code)
```

## Test plan

- [x] `test_get_model_fields_annotated_basemodel` — verifies `get_model_fields` does not raise for `RootModel[Annotated[BaseModel, Field(...)]]`
- [x] `test_get_model_fields_annotated_basemodel_openapi` — verifies `/openapi.json` returns 200 for such a response model
- [x] All existing `tests/test_compat.py` tests pass (10/10)
- [x] `tests/test_response_model_as_return_annotation.py` (38 tests), `tests/test_computed_fields.py`, `tests/test_openapi_separate_input_output_schemas.py` all pass

Closes #14805 (alternative fix using public typing API instead of `pydantic._internal._typing_extra.annotated_type`)

### DIFF
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -383,7 +383,14 @@ def get_model_fields(model: type[BaseModel]) -> list[ModelField]:
     model_fields: list[ModelField] = []
     for name, field_info in model.model_fields.items():
         type_ = field_info.annotation
-        if lenient_issubclass(type_, (BaseModel, dict)) or is_dataclass(type_):
+        # Unwrap Annotated[T, ...] to T before the issubclass check.
+        # For RootModel fields and similar, pydantic may store the annotation
+        # as Annotated[SomeBaseModel, FieldInfo(...)]. lenient_issubclass does
+        # not recognise the Annotated wrapper as a BaseModel subclass, so
+        # model_config would be passed to TypeAdapter, which Pydantic rejects
+        # when the inner type is a BaseModel/dataclass/TypedDict.
+        check_type = get_args(type_)[0] if get_origin(type_) is Annotated else type_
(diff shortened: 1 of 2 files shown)


YuriiMotov:
I just checked it - tests pass on `master`. So, the issues seems to has been resolved!

@xovishnukosuri, thanks for your efforts!

PR #15121, opened 2026-03-15
Fix comments and type hints in models.py

varenoma:


YuriiMotov:

PR #15124, opened 2026-03-15
Fix jsonable_encoder custom encoders for Pydantic models

letmetry-oss:
## Motivation
When calling `jsonable_encoder(model, custom_encoder={...})` on a Pydantic v2 
BaseModel with arbitrary types, FastAPI ran `model_dump(mode="json")` first 
which raised `PydanticSerializationError` and prevented the custom_encoder 
from being applied.

## Changes
- Use `mode="python"` for `model_dump()` when a custom_encoder is provided
- Pass `custom_encoder` through recursive calls so nested fields are encoded correctly

## Testing
Added regression test in `tests/test_jsonable_encoder.py` with 
`arbitrary_types_allowed=True` model to verify custom encoding works correctly.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -224,8 +224,11 @@ def jsonable_encoder(
     if exclude is not None and not isinstance(exclude, (set, dict)):
         exclude = set(exclude)  # type: ignore[assignment]  # ty: ignore[unused-ignore-comment]
     if isinstance(obj, BaseModel):
+        model_dump_mode: str = "json"
+        if custom_encoder:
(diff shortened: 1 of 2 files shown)


letmetry-oss:
Thanks for the automated review. Noted the point about custom_encoder 
not being applied uniformly to top-level model fields when mode='python' 
is used. Happy to refine the approach based on maintainer feedback — 
whether that's keeping mode='python' with better recursion handling, 
or a different strategy entirely.

Performance benchmarks confirmed clean (20 untouched benchmarks via CodSpeed).

PR #15125, opened 2026-03-16
Fix custom encoders for BaseModel python-mode dump

letmetry-oss:


YuriiMotov:
@letmetry-oss, please, test things locally before opening PRs!

PR #15126, opened 2026-03-16
Fix stringified Annotated dependency forward refs

letmetry-oss:


YuriiMotov:
@letmetry-oss, have you reviewed and understand the code of this PR?

PR #15131, opened 2026-03-17
fix: resolve ForwardRef in Annotated types for OpenAPI schema generation

BillionClaw:
## Description

Fixes #13056

When using `from __future__ import annotations` with `Annotated[Type, Depends(...)]` where `Type` is defined after the route function, the OpenAPI schema generation would fail with a `PydanticUserError` about unresolved ForwardRefs.

## Changes

Modified `ModelField` in `fastapi/_compat/v2.py` to:
1. Add a `core_schema` property that checks if the schema contains unresolved ForwardRefs
2. If unresolved, rebuild the TypeAdapter with a combined namespace from all loaded modules
3. Updated `get_definitions()` and `_has_computed_fields()` to use the new property

## Testing

Tested with the reproduction case from the issue. The OpenAPI schema is now generated correctly with the referenced type schema properly defined.

### DIFF
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -152,6 +152,31 @@ class ModelField:
                 config=self.config,
             )
 
+    @property
+    def core_schema(self) -> CoreSchema:
+        """Get the core schema, rebuilding if necessary to resolve ForwardRefs."""
+        schema = self._type_adapter.core_schema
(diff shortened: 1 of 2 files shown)


BillionClaw:
The check-labels CI check is failing. Could a maintainer please add the appropriate label (e.g., 'bug' or 'fix') to this PR? Thanks!

PR #15140, opened 2026-03-17
fix: unique operation IDs for routes with multiple methods (#13175)

aayushbaluni:
## Summary
Fixes #13175

`generate_unique_id` used `list(route.methods)[0]` so all HTTP methods on a multi-method route shared the same operation_id, causing "Duplicate Operation ID" warnings and broken OpenAPI/Swagger.

## Root cause
For routes with `methods=["POST", "DELETE"]`, `route.unique_id` was computed once using only the first method. `get_openapi_operation_metadata` reused that same id for every method when building the OpenAPI spec.

## Fix
In `get_openapi_operation_metadata`, when a route has multiple methods and we use `route.unique_id`, derive a method-specific operation_id by replacing the method suffix with the current method.

## Files changed
- `fastapi/openapi/utils.py` — method-specific operation_id for multi-method routes
- `tests/test_generate_unique_id_function.py` — regression test

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -242,7 +242,14 @@ def get_openapi_operation_metadata(
     operation["summary"] = generate_operation_summary(route=route, method=method)
     if route.description:
         operation["description"] = route.description
-    operation_id = route.operation_id or route.unique_id
+    if route.operation_id:
+        operation_id = route.operation_id
+    elif route.methods and len(route.methods) > 1:
+        # Multi-method routes: unique_id uses first method only; use method-specific id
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@aayushbalun, thank you for your interest and efforts!

This solution has the same problem as https://github.com/fastapi/fastapi/pull/15074 (see second point in [my comment](https://github.com/fastapi/fastapi/pull/15074#pullrequestreview-3908105706)) - body schema name and title will be non-deterministic.

Apart from that, it's recommended not to use `app.add_api_route` with multiple methods. Just add them separately.

PR #15141, opened 2026-03-17
🐛 Fix AssertionError when using TestClient directly with APIRouter

nilerbarcelos:
`TestClient(router)` fails with `AssertionError: fastapi_middleware_astack not found in request scope` when the router has yield dependencies and is not wrapped in a `FastAPI()` app.

`AsyncExitStackMiddleware` only gets added in `FastAPI.build_middleware_stack()`, so the scope key is never set when using a router directly. Fixed by initializing it in `request_response()` when missing.

Closes #14128

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -115,6 +115,11 @@ def request_response(
             response_awaited = False
             async with AsyncExitStack() as request_stack:
                 scope["fastapi_inner_astack"] = request_stack
+                # Ensure fastapi_middleware_astack exists in scope even when
+                # the router is used directly without a FastAPI app (which
+                # normally sets it up via AsyncExitStackMiddleware).
+                if "fastapi_middleware_astack" not in scope:
+                    scope["fastapi_middleware_astack"] = request_stack
                 async with AsyncExitStack() as function_stack:
                     scope["fastapi_function_astack"] = function_stack
                     response = await f(request)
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@nilerbarcelos, thanks for your interest!

This solution looks like a dirty hack..
Normally `fastapi_middleware_astack` and `fastapi_inner_astack` are to different stacks with different scopes. I believe it's incorrect to use `fastapi_inner_astack` for both..

I think you should just create an instance of app and router to it:
```py
    router = APIRouter()

    async def yield_dep():
        yield "value"

    @router.get("/test")
    async def endpoint(dep: str = Depends(yield_dep)):
        return {"dep": dep}

    app  = FastAPI()
    app.include_router(router)

    client = TestClient(app)
```


Simplify it with utility function like:

```py
def get_app(router: APIRouter) -> FastAPI:
    app  = FastAPI()
    app.include_router(router)
    return app
...

def test_router_direct_testclient_with_yield_dependency():
    router = APIRouter()

    async def yield_dep():
        yield "value"

    @router.get("/test")
    async def endpoint(dep: str = Depends(yield_dep)):
        return {"dep": dep}

    client = TestClient(get_app(router))

    ...
```

---

If you don't have specific use case where suggested solution is not suitable, then let's close this PR

nilerbarcelos:
Hey @YuriiMotov, thanks for the review!

I opened this with the intention of contributing, but honestly I don't have a real use case to back it up. I'll follow your recommendation and close the PR.

See you around!

PR #15142, opened 2026-03-17
fix: resolve ForwardRef in get_typed_annotation for Annotated+Depends (#13056)

aayushbaluni:
## Summary
Fixes #13056

When using `from __future__ import annotations` with `Annotated[Potato, Depends(get_potato)]`, the annotation can reach `analyze_param` as a `ForwardRef` (e.g. when `inspect.signature` partially evaluates or in certain code paths). Previously `get_typed_annotation` only resolved string annotations; `ForwardRef` was returned as-is. This caused:
1. `Depends` to be missed (since `get_origin(ForwardRef)` is `None`)
2. The param to be incorrectly treated as a default Query param
3. `PydanticUserError` during OpenAPI schema generation (`TypeAdapter` with unresolved `ForwardRef`)

## Fix
Extend `get_typed_annotation` to also resolve `ForwardRef` annotations using `evaluate_forwardref` before returning.

## Root cause
`get_typed_annotation` only handled `str`; when the annotation was already a `ForwardRef`, it was returned unresolved, breaking downstream `Annotated` extraction and schema generation.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -245,6 +245,7 @@ def get_typed_signature(call: Callable[..., Any]) -> inspect.Signature:
 def get_typed_annotation(annotation: Any, globalns: dict[str, Any]) -> Any:
     if isinstance(annotation, str):
         annotation = ForwardRef(annotation)
+    if isinstance(annotation, ForwardRef):
         annotation = evaluate_forwardref(annotation, globalns, globalns)  # ty: ignore[deprecated]
         if annotation is type(None):
             return None


aayushbaluni:
This PR fixes a bug where QueryParameter is treated as a body field when its alias matches an existing Body parameter name. Could a maintainer add the `bug` label so the CI check passes? Thank you!

YuriiMotov:
@aayushbaluni, have you tested it?
I just run the code example from https://github.com/fastapi/fastapi/issues/13056 on the branch of this PR and nothing changed - it fails:

```
    raise PydanticUserError(self._error_message, code=self._code)
pydantic.errors.PydanticUserError: `TypeAdapter[typing.Annotated[ForwardRef('Annotated[Potato, Depends(get_potato)]'), FieldInfo(annotation=NoneType, required=True, alias='potato', alias_priority=2, json_schema_extra={})]]` is not fully defined; you should define `typing.Annotated[ForwardRef('Annotated[Potato, Depends(get_potato)]'), FieldInfo(annotation=NoneType, required=True, alias='potato', alias_priority=2, json_schema_extra={})]` and all referenced types, then call `.rebuild()` on the instance.

For further information visit https://errors.pydantic.dev/2.12/u/class-not-fully-defined
```

**Please, respect time of maintainers and check code generated by AI before opening PR!**

deeptireddy-lab:
Severity: P2
Component: OpenAPI JSON generation
Repro steps: The code provided does not generate the correct OpenAPI JSON when using `Annotated` with `ForwardRef`. Specifically:
```

from typing import Annotated

app = FastAPI()

def get_potato() -> Potato:
    return Potato(color='red', size=10)

@app.get('/')
async def read_root(potato: Annotated[Potato, Depends(get_potato)]):
    return {'Hello': 'World'}

@dataclass
class Potato:
    color: str
    size: int
```

If you move the `Potato` class definition up or remove the `Annotated` annotation, it works as expected.
Duplicates: #13056, #15313

PR #15144, opened 2026-03-17
Improve path params documentation with validation example

venkatasaikumarguntupalli:
“Added a simple example demonstrating how to use Path for validating path parameters with constraints like **ge** and **le**.  Also included a short explanation and use cases to improve clarity for new users.”


YuriiMotov:
Thanks for your interest, but we don't need this - we already have example with validation ([see](https://fastapi.tiangolo.com/tutorial/path-params-numeric-validations/#number-validations-greater-than-and-less-than-or-equal)).
Also, this PR contains unnecessary formatting changes, and it actually breaks the formatting. Please, check things locally before opening PRs

PR #15145, opened 2026-03-18
👷 Re-enable translation workflow run by cron in CI (twice a month)

YuriiMotov:
We disabled it while we were working on adding new languages (see https://github.com/fastapi/fastapi/pull/14613).
Now it's done and we can re-enable it


tiangolo (approved):
Let's do it! 🚀

Feel free to update to twice a month (as you suggested in DM) before merging.

PR #15148, opened 2026-03-18
🔨 Update translation general prompt to enforce link style in translation matches the original link style

YuriiMotov:
Currently translation script fails to translate page if there is an existing translation and the link style in the original document has been changed (e.g. from HTML-style to markdown-style).

Checked it locally - works well


tiangolo (approved):

tiangolo:
Great, thanks!

PR #15149, opened 2026-03-18
⬆️ Support free-threaded Python 3.14t

svlandeg:
Continues from #13946: adding support for free-threaded Python 3.14t

* Extend test suite with a configuration for `python-version: "3.14t"`
* Allow the test suite to run when `orjson` and/or `ujson` is not installed, cf also https://github.com/fastapi/fastapi/pull/14964 which deprecated `ORJSONResponse` & `UJSONResponse`
* Remove `orjson` & `ujson` as a standard test dependency
* Add an additional test configuration to test for deprecated `ujson`/`orjson` tests (to ensure coverage goes back to 100%)
* Ensure that `fastar` 0.9.0 is used

## Checklist

- [x] Fix issues with `orjson` & `ujson` on free-threaded run
- [x] Run `orjson` & `ujson` tests in one specific non-free-threaded run
- [x] Fix `fastar` issue on free-threaded run: https://github.com/DoctorJohn/fastar/releases/tag/v0.9.0
- [x] Fix SQLAlchemy cyextension issues on Ubuntu

## Open question

Decided not to do this:
- [ ] Bump `pydantic-core` & `strawberry-graphql` (and maybe others) to allow running `uv-resolution: lowest-direct` on 3.14t as well. This will negatively impact other users on older Python versions / older external lib versions.


ngoldbaum:
Thanks for picking this up again!

fastapi/responses.py
```diff
@@ -17,10 +18,17 @@
     ujson = None  # type: ignore
 
 
+class _OrjsonModule(Protocol):
```
svlandeg, line 20: This is a bit convoluted to avoid type checker shenanigans, as they need to work with or without `orjson` in the env.

(and the same thing for `ujson` in this module)

svlandeg:
Hm, now stuck on `fastar` 😕

ngoldbaum:
It looks like they need an update either to declare support for the free-threaded build or simply update to the latest PyO3 version, which does that automatically. I might poke at that today to help you out here.

svlandeg:
> It looks like they need an update either to declare support for the free-threaded build or simply update to the latest PyO3 version, which does that automatically. I might poke at that today to help you out here.

That would be awesome, thank you so much!

ngoldbaum:
I opened https://github.com/DoctorJohn/fastar/pull/60.

ngoldbaum:
Looks like the Mac and Windows tests aren't picking up fastar 0.9.0.

svlandeg:
> Looks like the Mac and Windows tests aren't picking up fastar 0.9.0.

It's unclear to me why the `3.14t` test on Windows wasn't installing the latest `fastar` 0.9.0, https://github.com/fastapilabs/fastapi-cloud-cli has it at `"fastar >= 0.8.0"`. Anyway for now I explicitely require 0.9.0 and then that test does succeed 🎉 

The Mac test is a different issue: I was running that one with `uv-resolution: lowest-direct` which means that it tries out the lower bounds of our supported ranges. Here, it crashed for an older version of Pydantic, which is not surprising. Now the only open question is whether we want to bump those lower bounds (which may be annoying for users on older versions of Python & external repos), or just leave them and assume that users running `3.14t` will also have updated dependencies 🤷

svlandeg:
Ok cool, confirmed that bumping to `pydantic >= 2.12.0` and `strawberry-graphql >= 0.284.0` allows the test suite to run succesfully with `uv-resolution: lowest-direct` on Python 3.14t.

Discussed this internally with Sebastián and decided that we don't want to go so far as bumping Pydantic to 2.12+, as it might create issues for users still on older versions. So, I'll remove this test configuration again and then this PR should be ready for review.

tiangolo (approved):
Awesome, thank you! 🙌 🚀

PR #15161, opened 2026-03-19
Clarify required but nullable query parameters behavior

naveencreation:
## Summary
Clarifies the behavior of query parameters declared as required while allowing `None`.

## Changes
- Added explanation of how `= ...` (Ellipsis) works with `str | None`
- Clarified that query parameters are always received as strings in HTTP
- Added guidance to avoid confusion when using this pattern in query parameters

## Why
This pattern can be confusing because query parameters are always received as strings in HTTP and do not have a native representation for `None`. This update helps users better understand the behavior and limitations.


YuriiMotov:
Thanks, but we already have https://github.com/fastapi/fastapi/pull/14322

PR #15164, opened 2026-03-19
fix: include all methods in generate_unique_id to prevent duplicates

themavik:
## Summary
- Fix `generate_unique_id()` to produce distinct operation IDs for routes with multiple methods
- Previously, only the first method was used, causing duplicate IDs for multi-method routes

## Root Cause
`generate_unique_id()` uses `list(route.methods)[0].lower()` as the method suffix. For routes registered with multiple methods (e.g., `@app.api_route(methods=["POST", "DELETE"])`), the set iteration order is non-deterministic, leading to duplicate operation IDs.

## Testing
- Verified distinct IDs are generated for multi-method routes
- Existing test suite passes

Fixes #13175

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -1,6 +1,7 @@
 import copy
 import http.client
 import inspect
+import re
 import warnings
 from collections.abc import Sequence
 from typing import Any, Literal, cast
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Thank you for your interest and efforts!

This solution has the same problem as https://github.com/fastapi/fastapi/pull/15074 (see second point in https://github.com/fastapi/fastapi/pull/15074#pullrequestreview-3908105706) - body schema name and title will be non-deterministic.

Apart from that, it's recommended not to use app.add_api_route with multiple methods. Just add them separately.

PR #15166, opened 2026-03-19
🔨 Tweak translation workflow and translation fixer tool

YuriiMotov:
This PR includes 2 fixes:

**1. Don't fail if translation doesn't pass validation (just save it and continue with next page)**

When translation script fails to translate page after 3 attempts (results fails to pass validation), it currently causes the workflow to fail. So, we are missing all data processed by that moment..
We can actually just skip it and fix this page later manually.

**2. Forbid whitespace before attributes in markdown link pattern**

```
## [title](https://example.com) { #permalink }
```

The line above currently mathches two patterns: header permalink and markdown link with parameters.

This causes problems - translation fixer tool treats it as link and removes whitespace, then it fails with "Number of permalinks doesn't match.."

Updated pattern of markdown link to not allow that whitespace


tiangolo (approved):
Nice, thanks! 🚀

PR #15173, opened 2026-03-20
Correct capitalization of 'Schema' in documentation

24f2001045-DS:
Fix capitalization of "Schema" for consistency in documentation.


YuriiMotov:
@24f2001045-DS, thanks for your interest!

I think we don't need to change this. Let's focus on more important things

PR #15182, opened 2026-03-20
Add edge case tests for Query parameters using Annotated

NA-V10:
### Description
Added edge case tests for Query parameters using Annotated.

### Changes
- Added test for empty query parameter
- Added test for special characters (properly encoded)
- Added test for unicode input

### Why
These cases improve test coverage and ensure consistent handling of query parameters.

### Testing
All tests pass locally using pytest.


YuriiMotov:

PR #15183, opened 2026-03-20
Fix deprecated=False override precedence for included routes

dipenpadhiyar:
## Summary

Fix `deprecated` precedence when including routers so an explicit `deprecated=False`
on a route is not overridden by a parent app or router with `deprecated=True`.

This aligns behavior with expected precedence: route-level configuration should
override parent settings.

---

## Repro

```python

child = APIRouter()

@child.get("/x", deprecated=False)
def x():
    return {"ok": True}

app = FastAPI(deprecated=True)
app.include_router(child)

assert "deprecated" not in app.openapi()["paths"]["/x"]["get"]
```
---
## Changes

Preserve explicit non-None deprecated values when merging route, router, and app settings

Add a regression test for deprecated=False overriding parent deprecated=True

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1394,7 +1394,7 @@ class APIRouter(routing.Router):
             description=description,
             response_description=response_description,
             responses=combined_responses,
-            deprecated=deprecated or self.deprecated,
+            deprecated=self.deprecated if deprecated is None else deprecated,
             methods=methods,
             operation_id=operation_id,
             response_model_include=response_model_include,
@@ -1770,7 +1770,11 @@ class APIRouter(routing.Router):
                     description=route.description,
                     response_description=route.response_description,
                     responses=combined_responses,
-                    deprecated=route.deprecated or deprecated or self.deprecated,
+                    deprecated=(
+                        route.deprecated
+                        if route.deprecated is not None
+                        else self.deprecated if deprecated is None else deprecated
+                    ),
                     methods=route.methods,
                     operation_id=route.operation_id,
                     response_model_include=route.response_model_include,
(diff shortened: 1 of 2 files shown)


dipenpadhiyar:
This PR fixes a logic bug in `deprecated` precedence for included routes.

Could a maintainer please add the `bug` label so the required label check can pass?

YuriiMotov:
I'm not sure this is a bug.
Deprecating the whole router except some specific routes looks strange to me.

@dipenpadhiyar , what is your use case?

nidhishgajjar:

## Review of PR #15183: Fix deprecated=False override precedence for included routes

Nice bug fix — correctly handles the case where a child route explicitly sets `deprecated=False` to override a parent's `deprecated=True`.

### The bug ✅ correctly identified

Using `or` for precedence doesn't work here because `False` is falsy:
```python
deprecated=False or self.deprecated=True  # → True (wrong!)
```

### The fix ✅

Using `is None` checks properly distinguishes between "not specified" (None) and "explicitly set to False":
- In `add_api_route`: `deprecated=self.deprecated if deprecated is None else deprecated`
- In the router include: route-level takes highest precedence, then the include-level, then the router default

### Test ✅

Good test case — verifies that a route with `deprecated=False` on a child router included in a `FastAPI(deprecated=True)` app does NOT appear as deprecated in the OpenAPI schema.

### Minor readability note 💡

The nested ternary in the include logic works but is slightly hard to parse at a glance:
```python
deprecated=(
    route.deprecated
    if route.deprecated is not None
    else self.deprecated
    if deprecated is None
    else deprecated
)
```
Adding parentheses could help future readers:
```python
deprecated=(
    route.deprecated if route.deprecated is not None
    else (self.deprecated if deprecated is None else deprecated)
)
```

Not blocking — just a readability suggestion.

### Summary

Clean, focused bug fix with test coverage. The `is None` check is the right approach to distinguish "unset" from "explicitly False."

**Assessment: approve**

PR #15184, opened 2026-03-20
Add Russian translation of contributing.md

argoarsiks:
Llm generation didn't create this file, so I manually translated contributing.md to Russian. This PR adds the missing Russian version of the page.


YuriiMotov:
As you can see it doesn't work.
Some pages are intentionally marked to be excluded from pages to be translated (https://github.com/fastapi/fastapi/blob/64feaec80244578bbdabe51e866bcfad1df8b708/scripts/translate.py#L18-L27)

Thanks for your interest!

PR #15187, opened 2026-03-21
fix: reject newline characters in SSE event and id fields

subhashdasyam:
## Summary

`format_sse_event()` in `fastapi/sse.py` interpolated the `event` and `id` fields
into SSE wire-format bytes **without stripping or rejecting newline characters**.
A `\n` or `\r\n` in either field injects an additional SSE field-line into the
stream, enabling event-type spoofing and fabricated `data:` payloads in browser
`EventSource` clients.

The `data` and `comment` fields in the same function are correctly protected via
`splitlines()`. The omission for `event` and `id` is an inconsistency — not a
design decision.

**Vulnerability class:** SSE Protocol Injection (CRLF injection into wire format)
**CWE:** CWE-116 — Improper Encoding for Output in a Different Plaintext Context
**OWASP:** A03:2021 — Injection
**Prior art:** Hono [GHSA-p6xx-57qc-3wxr](https://github.com/honojs/hono/security/advisories/GHSA-p6xx-57qc-3wxr) — identical bug class, same fix pattern

## Inconsistency before this fix

| Field     | Newline handling         | Safe? |
|-----------|--------------------------|-------|
| `data`    | `splitlines()` loop      | ✅ Yes |
| `comment` | `splitlines()` loop      | ✅ Yes |
| `event`   | None                     | ❌ No  |
| `id`      | Only `\0` check          | ❌ No  |
| `retry`   | `int` type (no strings)  | ✅ Yes |

## What the injection looks like

A request like `GET /stream?type=chat%0Adata:%20{"cmd":"exec"}` causes
`format_sse_event` to produce:

```
event: chat
data: {"cmd":"exec"}
data: "real server data"

```

The browser's `EventSource` receives a `chat` event with a fabricated `data:`
field prepended to the real payload. Frontends that process `event.data`
line-by-line (common in streaming AI/MCP UIs) receive a corrupted payload.

## Changes

- **`fastapi/sse.py`**
  - Add `_check_event_no_newline()` validator; apply to `ServerSentEvent.event`
    via `AfterValidator` (mirrors the existing `_check_id_no_null` pattern)
  - Extend `_check_id_no_null()` to also reject `\n` and `\r`
  - Strip `\r`/`\n` from `event` and `id` in `format_sse_event()` as
    defense-in-depth (low-level function; callable without going through
    the Pydantic model validators)

- **`tests/test_sse.py`**
  - 7 new regression tests covering LF/CR rejection in `ServerSentEvent.event`,
    `ServerSentEvent.id`, and the `format_sse_event()` low-level function

## Test results

All 25 tests pass (`pytest tests/test_sse.py -v`), including the 7 new ones.

## Severity

**Low at framework level / Medium in affected applications.** The vulnerability
only activates when a developer routes attacker-influenced input into the `event=`
or `id=` fields — most correct uses pass static strings. The fix is a one-line
guard that prevents the framework from silently accepting malformed input.

### DIFF
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -33,9 +33,18 @@ class EventSourceResponse(StreamingResponse):
     media_type = "text/event-stream"
 
 
-def _check_id_no_null(v: str | None) -> str | None:
-    if v is not None and "\0" in v:
-        raise ValueError("SSE 'id' must not contain null characters")
+def _check_id_no_null_or_newline(v: str | None) -> str | None:
+    if v is not None:
+        if "\0" in v:
+            raise ValueError("SSE 'id' must not contain null characters")
+        if "\n" in v or "\r" in v:
+            raise ValueError("SSE 'id' must not contain newline characters")
+    return v
+
+
+def _check_event_no_newline(v: str | None) -> str | None:
+    if v is not None and ("\n" in v or "\r" in v):
+        raise ValueError("SSE 'event' must not contain newline characters")
     return v
 
 
@@ -86,18 +95,21 @@ class ServerSentEvent(BaseModel):
     ] = None
     event: Annotated[
         str | None,
+        AfterValidator(_check_event_no_newline),
         Doc(
             """
             Optional event type name.
 
             Maps to `addEventListener(event, ...)` on the browser. When omitted,
             the browser dispatches on the generic `message` event.
+
+            **Must not contain newline (`\\n`) or carriage return (`\\r`) characters.**
             """
         ),
     ] = None
     id: Annotated[
         str | None,
-        AfterValidator(_check_id_no_null),
+        AfterValidator(_check_id_no_null_or_newline),
         Doc(
             """
             Optional event ID.
@@ -197,6 +209,7 @@ def format_sse_event(
             lines.append(f": {line}")
 
     if event is not None:
+        event = event.replace("\r", "").replace("\n", "")
         lines.append(f"event: {event}")
 
     if data_str is not None:
@@ -204,6 +217,7 @@ def format_sse_event(
             lines.append(f"data: {line}")
 
     if id is not None:
+        id = id.replace("\r", "").replace("\n", "").replace("\0", "")
         lines.append(f"id: {id}")
 
     if retry is not None:
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -212,7 +212,10 @@ def format_sse_event(
             lines.append(f": {line}")
 
     if event is not None:
-        event = event.replace("\r", "").replace("\n", "").replace("\0", "")
+        if "\n" in event or "\r" in event:
+            raise ValueError("SSE 'event' must not contain newline characters")
+        if "\0" in event:
+            raise ValueError("SSE 'event' must not contain null characters")
         lines.append(f"event: {event}")
 
     if data_str is not None:
@@ -220,7 +223,10 @@ def format_sse_event(
             lines.append(f"data: {line}")
 
     if id is not None:
-        id = id.replace("\r", "").replace("\n", "").replace("\0", "")
+        if "\n" in id or "\r" in id:
+            raise ValueError("SSE 'id' must not contain newline characters")
+        if "\0" in id:
+            raise ValueError("SSE 'id' must not contain null characters")
         lines.append(f"id: {id}")
 
     if retry is not None:
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -43,8 +43,11 @@ def _check_id_no_null_or_newline(v: str | None) -> str | None:
 
 
 def _check_event_no_newline(v: str | None) -> str | None:
-    if v is not None and ("\n" in v or "\r" in v):
-        raise ValueError("SSE 'event' must not contain newline characters")
+    if v is not None:
+        if "\0" in v:
+            raise ValueError("SSE 'event' must not contain null characters")
+        if "\n" in v or "\r" in v:
+            raise ValueError("SSE 'event' must not contain newline characters")
     return v
 
 
@@ -209,7 +212,7 @@ def format_sse_event(
             lines.append(f": {line}")
 
     if event is not None:
-        event = event.replace("\r", "").replace("\n", "")
+        event = event.replace("\r", "").replace("\n", "").replace("\0", "")
         lines.append(f"event: {event}")
 
     if data_str is not None:
--- a/tests/test_sse.py
+++ b/tests/test_sse.py
@@ -246,36 +246,46 @@ def test_server_sent_event_cr_id_rejected():
         ServerSentEvent(data="test", id="42\rpwned")
 
 
-def test_format_sse_event_strips_newline_in_event():
+def test_format_sse_event_raises_on_newline_in_event():
     from fastapi.sse import format_sse_event
 
-    result = format_sse_event(event="chat\npwned", data_str="hello")
-    assert b"\npwned" not in result
-    assert b"event: chatpwned\n" in result
+    with pytest.raises(ValueError, match="newline"):
+        format_sse_event(event="chat\npwned", data_str="hello")
 
 
-def test_format_sse_event_strips_cr_in_event():
+def test_format_sse_event_raises_on_cr_in_event():
     from fastapi.sse import format_sse_event
 
-    result = format_sse_event(event="chat\rpwned", data_str="hello")
-    assert b"\rpwned" not in result
-    assert b"event: chatpwned\n" in result
+    with pytest.raises(ValueError, match="newline"):
+        format_sse_event(event="chat\rpwned", data_str="hello")
 
 
-def test_format_sse_event_strips_newline_in_id():
+def test_format_sse_event_raises_on_null_in_event():
     from fastapi.sse import format_sse_event
 
-    result = format_sse_event(id="42\npwned", data_str="hello")
-    assert b"\npwned" not in result
-    assert b"id: 42pwned\n" in result
+    with pytest.raises(ValueError, match="null"):
+        format_sse_event(event="chat\x00pwned", data_str="hello")
 
 
-def test_format_sse_event_strips_cr_in_id():
+def test_format_sse_event_raises_on_newline_in_id():
     from fastapi.sse import format_sse_event
 
-    result = format_sse_event(id="42\rpwned", data_str="hello")
-    assert b"\rpwned" not in result
-    assert b"id: 42pwned\n" in result
+    with pytest.raises(ValueError, match="newline"):
+        format_sse_event(id="42\npwned", data_str="hello")
+
+
+def test_format_sse_event_raises_on_cr_in_id():
+    from fastapi.sse import format_sse_event
+
+    with pytest.raises(ValueError, match="newline"):
+        format_sse_event(id="42\rpwned", data_str="hello")
+
+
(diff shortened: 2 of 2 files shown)


subhashdasyam:
Hi maintainers 👋

The **Labels / check-labels** CI check is failing because this PR is missing one of the required labels.

This is a security fix (SSE protocol injection via unvalidated `event` and `id` fields — CWE-116), so the appropriate label would be **`security`**.

Could a maintainer please apply the `security` label so the check passes? Thank you!

subhashdasyam:
Fix tracked in issue #15188.

SadManFahIm:
Thanks for the thorough fix and the regression tests! A few observations while reviewing:

**1. Bare `\r` test coverage**
The validator at line 46 correctly rejects `\r` alone (`"\n" in v or "\r" in v`). 
Could you confirm whether `tests/test_sse.py` includes an explicit test case for 
a bare carriage return without a newline?

```python
# e.g.
with pytest.raises(ValueError):
    ServerSentEvent(event="chat\radmin_command")
```

**2. Inconsistency in `format_sse_event()` for null bytes on `id`**
The `_check_id_no_null_or_newline` validator raises `ValueError` on `\0`, but 
`format_sse_event()` at line 220 silently strips `\0` via `.replace("\0", "")`. 
These two code paths have different behaviors for the same input — is that intentional? 
If the validator rejects null bytes at the model level, should `format_sse_event()` 
also raise (or at least not strip silently)?

**3. `event` has no null-byte check**
`id` has had null-byte protection since the original `_check_id_no_null`. 
The new `_check_event_no_newline` doesn't include a similar `\0` check for `event`. 
Is that an intentional asymmetry, or worth adding for consistency?

Overall the fix looks solid — the two-layer approach (validator + defense-in-depth 
sanitization in `format_sse_event`) is a good pattern. Just flagging these for 
completeness.

fastapi/sse.py
```diff
@@ -197,13 +209,15 @@ def format_sse_event(
...
     if event is not None:
+        event = event.replace("\r", "").replace("\n", "")
         lines.append(f"event: {event}")
 
     if data_str is not None:
         for line in data_str.splitlines():
             lines.append(f"data: {line}")
 
     if id is not None:
+        id = id.replace("\r", "").replace("\n", "").replace("\0", "")
```
YuriiMotov: I find it strange to have these `replace` here. This looks like an alternative to validation logic implemented using `AfterValidator`.
@subhashdasyam, could you please clarify your intention?

subhashdasyam:
> Thanks for the thorough fix and the regression tests! A few observations while reviewing:
> 
> **1. Bare `\r` test coverage** The validator at line 46 correctly rejects `\r` alone (`"\n" in v or "\r" in v`). Could you confirm whether `tests/test_sse.py` includes an explicit test case for a bare carriage return without a newline?
> 
> ```python
> # e.g.
> with pytest.raises(ValueError):
>     ServerSentEvent(event="chat\radmin_command")
> ```
> 
> **2. Inconsistency in `format_sse_event()` for null bytes on `id`** The `_check_id_no_null_or_newline` validator raises `ValueError` on `\0`, but `format_sse_event()` at line 220 silently strips `\0` via `.replace("\0", "")`. These two code paths have different behaviors for the same input — is that intentional? If the validator rejects null bytes at the model level, should `format_sse_event()` also raise (or at least not strip silently)?
> 
> **3. `event` has no null-byte check** `id` has had null-byte protection since the original `_check_id_no_null`. The new `_check_event_no_newline` doesn't include a similar `\0` check for `event`. Is that an intentional asymmetry, or worth adding for consistency?
> 
> Overall the fix looks solid — the two-layer approach (validator + defense-in-depth sanitization in `format_sse_event`) is a good pattern. Just flagging these for completeness.

Thanks for looking closely at this.

**On point 2 -- `\0` inconsistency between validator and `format_sse_event()`**

The difference is intentional. These two layers have different jobs:

- `ServerSentEvent` (Pydantic model) is a strict gate. It rejects bad input so the caller has to fix their data upstream. Raising on `\0` here is the right behavior.
- `format_sse_event()` is a best-effort sanitizer. It is a low-level function callable directly without Pydantic (from routing internals, custom integrations, etc.). Its job is to always produce safe wire output regardless of how it was called. Raising there would defeat that purpose.

This pattern (raise at model level, sanitize at wire level) is the same approach used for `\n`/`\r` and was documented in the commit message as "defense-in-depth". The behavioral difference between the two layers is by design, not an oversight.

**On point 3 -- `event` has no null-byte check**

You are right, this was a real gap. `id` has had null-byte protection since the original `_check_id_no_null`, but `_check_event_no_newline` did not have it, and `format_sse_event()` was not stripping `\0` from `event` either.

Fixed in the follow-up commit:
- `_check_event_no_newline` now rejects `\0` with `ValueError("SSE 'event' must not contain null characters")`
- `format_sse_event()` now strips `\0` from `event`, matching the existing `id` behavior
- Added `test_server_sent_event_null_event_rejected` as a regression test

`event` and `id` are now fully symmetric in both the validator and the low-level formatter.

fastapi/sse.py
subhashdasyam: The `replace` calls in `format_sse_event()` are not an alternative to the validator. They cover a different code path.

`format_sse_event()` is a public low-level function. The routing layer in `routing.py` calls it directly with pre-validated fields from a `ServerSentEvent`, so in that path the sanitization is a no-op. But `format_sse_event()` can also be called without ever constructing a `ServerSentEvent`, for example from custom integrations or internal tooling that builds SSE frames directly. In that path there is no Pydantic validator in play at all.

The intent is two layers with different jobs: the `AfterValidator` is a strict gate at the model boundary, and the `replace` calls in `format_sse_event()` are a safety net for callers who skip the model. This follows the same pattern as the existing `splitlines()` on `data` and `comment`. Those fields have no Pydantic validator, but the function still protects the wire format regardless.

The follow-up commit from the comment I received from @SadManFahIm also extended this pattern to cover `\0` for `event`, fixing an asymmetry with `id` that was pointed out in review. Both layers now handle the same character set consistently, which is another sign the two layers serve different purposes rather than duplicating each other.

SadManFahIm: @subhashdasyam 
These aren't duplicating each other — they cover different call paths.

The routing layer always goes through a ServerSentEvent, so the
AfterValidator fires and the replace calls become a no-op there.
But format_sse_event() is a public, low-level function and can be
called directly — from custom integrations or internal tooling —
without ever constructing a ServerSentEvent. In that path there's
no Pydantic validation at all, so the replace calls are the only
protection.

subhashdasyam: > @subhashdasyam These aren't duplicating each other — they cover different call paths.
> 
> The routing layer always goes through a ServerSentEvent, so the AfterValidator fires and the replace calls become a no-op there. But format_sse_event() is a public, low-level function and can be called directly — from custom integrations or internal tooling — without ever constructing a ServerSentEvent. In that path there's no Pydantic validation at all, so the replace calls are the only protection.

On point 1 -- bare `\r` test coverage is already there. `test_server_sent_event_cr_event_rejected` at line 229 covers exactly this case:

```python
def test_server_sent_event_cr_event_rejected():
    with pytest.raises(ValueError, match="newline"):
        ServerSentEvent(data="test", event="chat\rpwned")
```

A bare `\r` without `\n`, asserts a `ValueError` is raised. Same pattern exists for `id` in `test_server_sent_event_cr_id_rejected`.

SadManFahIm (approved):

fastapi/sse.py
YuriiMotov: I don't think `format_sse_event` should be treated as a part of public API as it's not documented and seems to be for internal use only.

But yes, not having this guard logic in `format_sse_event` may lead to introducing bugs in the future.
Anyway, I believe we should be consistent in how we handle this. In case of `ServerSentEvent` it will raise `ValidationError`, so I think we should follow the same logic for the path without `ServerSentEvent`.
What if we just make `format_sse_event` return an instance of `ServerSentEvent`. This way the logic will be consistent and we don't need to duplicate it

YuriiMotov:
> `format_sse_event()` is a best-effort sanitizer. It is a low-level function callable directly without Pydantic (from routing internals, custom integrations, etc.). Its job is to always produce safe wire output regardless of how it was called. Raising there would defeat that purpose.

I think this assumption is incorrect. In case of `data` field, newline characters is absolutely normal and we must handle it. But for `event` field it's not normal and we should reject such value instead of just removing those characters

fastapi/sse.py
subhashdasyam: Hey @YuriiMotov , you're right that the inconsistency is a real problem. Let me explain why I didn't go the `ServerSentEvent` route, and what I think actually makes sense here.

The trouble with having `format_sse_event` construct a `ServerSentEvent` internally is that the two APIs don't share the same data model. `format_sse_event` takes `data_str` (already serialized data) while `ServerSentEvent` takes `data: Any` and `raw_data: str | None`. There's no clean mapping between them. On top of that, the routing layer calls `format_sse_event` with fields it already pulled from a validated `ServerSentEvent`, so you'd be validating the same thing twice. That seems like the wrong direction.

That said, your consistency point stands. If `ServerSentEvent` raises on bad input, `format_sse_event` should too, not silently strip. The `replace()` calls create a real behavioral mismatch between the two paths, and it's a fair thing to flag.

Would it make sense to update `format_sse_event` to raise `ValueError` on `\n`, `\r`, and `\0` in `id` and `event` instead of stripping them? 
Same behavior as the Pydantic validators, no need to structurally couple the two.

subhashdasyam:
> > `format_sse_event()` is a best-effort sanitizer. It is a low-level function callable directly without Pydantic (from routing internals, custom integrations, etc.). Its job is to always produce safe wire output regardless of how it was called. Raising there would defeat that purpose.
> 
> I think this assumption is incorrect. In case of `data` field, newline characters is absolutely normal and we must handle it. But for `event` field it's not normal and we should reject such value instead of just removing those characters

That distinction makes sense. `data` is multi-line by design in the SSE spec, so `splitlines()` is the right move there. `event` and `id` are single-value fields and a newline in either one isn't something you'd ever want silently swallowed.

`id` falls in the same category as `event` here. Should I update `format_sse_event` to raise `ValueError` for `\n`, `\r`, and `\0` in both `event` and `id`, and leave the `splitlines()` handling for `data` and `comment` as is?

YuriiMotov:
> Should I update `format_sse_event` to raise `ValueError` ... ?

I think yes, but it's you PR and you should decide.

subhashdasyam:
> > Should I update `format_sse_event` to raise `ValueError` ... ?
> 
> I think yes, but it's you PR and you should decide.

Thanks @YuriiMotov . 

`Decision made`: I'll update `format_sse_event` to raise `ValueError` for `\n`, `\r`, and `\0` in `event` and `id`, and drop the `replace()` calls for those fields.                                                                                    
                                                                                                                                                                                                                                                              
Your framing is the right one. data is multi-line by design in the SSE spec each line becomes a separate data: frame  so splitlines() is semantically correct there. event and id are single-token fields where a newline is always a protocol violation. Silently stripping it masks a real bug at the call site; raising surfaces it where it belongs.                                                                                                                                                   
                                                                                                           
This also resolves the behavioral inconsistency: both code paths (via ServerSentEvent and via format_sse_event directly) now fail the same way for the same invalid input. I'll update the function and adjust the tests accordingly.

tiangolo:
This was covered in a recent PR, thanks! :coffee:

PR #15189, opened 2026-03-22
docs: clarify behavior of required nullable query parameters

prasanthkumarch26:
### What does this PR do?

Clarifies the "Required, can be None" section in the query parameters documentation.

### Problem

The previous documentation suggested that a query parameter could be both required and accept `None`, but did not clearly explain that:

- HTTP query parameters cannot carry a true `None` value
- All query parameters are received as strings

This caused confusion for users trying to pass `None`.

### Solution

- Added a warning explaining HTTP behavior
- Clarified how query parameters are interpreted
- Provided guidance on handling values like "null" manually

### Notes

This change is documentation-only and does not affect runtime behavior.

### DIFF
--- a/docs/en/docs/tutorial/query-params-str-validations.md
+++ b/docs/en/docs/tutorial/query-params-str-validations.md
@@ -222,11 +222,24 @@ So, when you need to declare a value as required while using `Query`, you can si
 
 ### Required, can be `None` { #required-can-be-none }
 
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Please, search for existing PRs before opening your own

See https://github.com/fastapi/fastapi/pull/14322

PR #15191, opened 2026-03-22
✨ Make ServerSentEvent generic with typed data field

benmosher:
## Summary

- Make `ServerSentEvent` generic — `ServerSentEvent[Data]` now validates that `data` is an instance of `Data` and emits a `contentSchema` for that type in the OpenAPI spec
- Bare `ServerSentEvent` (no type parameter) continues to accept any value including `None`, preserving full backward compatibility
- Add `get_sse_data_type()` helper used by the routing layer to detect parameterized `ServerSentEvent[Data]` annotations and extract `Data` for schema generation
- Add `docs_src/server_sent_events/tutorial006_py310.py` example and corresponding test
- Add documentation section "Typed `ServerSentEvent`" to the SSE tutorial

## Motivation

Before this change, yielding `ServerSentEvent` objects from an SSE endpoint meant no data validation and no type information in the generated OpenAPI schema. Callers had to rely on documentation or convention to know what shape the `data` field would carry.

With `ServerSentEvent[Item]`:

```python
@app.get("/items/stream", response_class=EventSourceResponse)
async def stream_items() -> AsyncIterable[ServerSentEvent[Item]]:
    for i, item in enumerate(items):
        yield ServerSentEvent[Item](data=item, event="item_update", id=str(i + 1))
```

- Pydantic validates that every event's `data` really is an `Item` (or raises `ValidationError` at construction time)
- The generated OpenAPI schema includes a `contentSchema` referencing `Item`, so clients and SDK generators know the exact payload shape

## How it works

Pydantic's generic `BaseModel` creates a concrete subclass (not a `_GenericAlias`) when parameterized, so `get_origin()` returns `None`. The routing layer calls `get_sse_data_type()` which inspects `__pydantic_generic_metadata__["args"]` to extract `Data`. The routing logic then branches:

- `ServerSentEvent[Data]` → use `Data` as `stream_item_type` (feeds `contentSchema`)
- Bare `ServerSentEvent` → skip (no `contentSchema`, same as before)
- Any other type → use as-is (existing JSONL / plain-model path)

`model_config = ConfigDict(validate_default=True)` ensures that omitting `data` on a parameterized event raises a `ValidationError` rather than silently storing `None`.

## Example output

```json
"text/event-stream": {
  "itemSchema": {
    "type": "object",
    "properties": {
      "data": {
        "type": "string",
        "contentMediaType": "application/json",
        "contentSchema": { "$ref": "#/components/schemas/Item" }
      },
      "event": { "type": "string" },
      "id":    { "type": "string" },
      "retry": { "type": "integer", "minimum": 0 }
    },
    "required": ["data"]
  }
}
```

## Checklist

- [x] New tests cover the generic construction, validation, streaming, and OpenAPI schema
- [x] Tutorial example (`tutorial006_py310.py`) and tutorial test added
- [x] Backward compatibility: bare `ServerSentEvent` behaves identically to before
- [x] `ruff check` and `ruff format` pass with no changes
- [x] All 46 SSE tests pass

### DIFF
--- a/docs/en/docs/tutorial/server-sent-events.md
+++ b/docs/en/docs/tutorial/server-sent-events.md
@@ -93,6 +93,31 @@ This is useful for sending pre-formatted text, log lines, or special <dfn title=
 
 ///
 
+## Typed `ServerSentEvent` { #typed-serversentevent }
+
+When you yield plain `ServerSentEvent` objects, the `data` field accepts any value and the OpenAPI schema has no `contentSchema` for the payload.
+
+If you want **typed data validation** and a `contentSchema` in the OpenAPI spec while still having full control over SSE fields like `event`, `id`, `retry`, and `comment`, you can parameterize `ServerSentEvent` with a type:
+
+{* ../../docs_src/server_sent_events/tutorial006_py310.py hl[4,23,25] *}
+
+`ServerSentEvent[Item]` means:
+
+* `data` **must** be an `Item` instance — omitting it or passing the wrong type raises a validation error.
+* The generated OpenAPI schema includes a `contentSchema` referencing `Item` inside the SSE `data` field.
+
+/// tip
+
+If you need events where `data` is optional (for example, comment-only keep-alive pings), use `ServerSentEvent[Item | None]` instead.
+
+///
+
+/// note
+
+Bare `ServerSentEvent` (without a type parameter) still works exactly as before — `data` accepts any value including `None`.
+
+///
+
 ## Resuming with `Last-Event-ID` { #resuming-with-last-event-id }
 
 When a browser reconnects after a connection drop, it sends the last received `id` in the `Last-Event-ID` header.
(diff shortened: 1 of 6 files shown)


themavik (commented):
`ConfigDict(validate_default=True)` on `ServerSentEvent` tightens default validation for `data`. Worth a regression test that unparameterized `ServerSentEvent` still accepts backward-compatible payloads (including `data=None` and arbitrary dicts).

graipher:
There is a related discussion: [https://github.com/fastapi/fastapi/discussions/15102](https://github.com/fastapi/fastapi/discussions/15102)

That discussion goes one step further and wants to have a way to also constrain the `event` field at the same time, so that one can say that e.g.:
1. `event="item_update"` will only ever emit a `Item`
2. `event="error"` will only ever emit an `Error`

instead of just being able to say that the stream will contain `Item | Error`.

benmosher:
@graipher nice -- I want that too. I have further modifications that make that kind of event-based data type discrimination possible, but I wanted to start with this narrower, smaller step first, to gauge appetite for these enhancements and understand any constraints on the implementation. 

That said -- I can put up the next steps that achieve that as a stacked draft PR so we can discuss the next steps also.

PR #14869, opened 2026-02-08
Fix #13175: Generate unique operation IDs for routes with multiple methods

veeceey:
## Summary
Fixes #13175 by ensuring routes with multiple HTTP methods generate unique operation IDs for each method in the OpenAPI schema.

## Problem
When using `add_api_route()` or `@api_route()` with multiple methods (e.g., `methods=["POST", "DELETE"]`), all methods shared the same operation ID, causing "Duplicate Operation ID" warnings.

Example that triggered the warning:
```python
router.add_api_route("/clear", clear, methods=["POST", "DELETE"])
# Warning: Duplicate Operation ID clear_clear_post
```

## Solution
Modified `get_openapi_operation_metadata()` in `fastapi/openapi/utils.py`:
- When a route has multiple methods and no explicit `operation_id`
- Append the actual HTTP method to the base operation ID
- Ensures each method gets a unique operation ID: `clear_clear_post`, `clear_clear_delete`
- Maintains backward compatibility for single-method routes and explicit operation IDs

## Test Plan
Added comprehensive test suite in `tests/test_duplicate_operation_id.py`:
- ✅ No duplicate warnings for multi-method routes
- ✅ Operation IDs are unique for each method
- ✅ Multiple routes with multiple methods all get unique IDs
- ✅ Single-method routes unchanged (backward compatibility)
- ✅ Direct `add_api_route()` usage works correctly

All tests pass:
- New tests: 4/4 passed
- Existing OpenAPI tests: all passed
- Application tests: all passed

Generated with [Claude Code](https://claude.com/claude-code)

PR #14827, opened 2026-02-04
 Fix Duplicate OperationId When Using Multiple Methods on Same Route

GDemay:
## Description

Fixes #13175

When a route is added with multiple HTTP methods using `router.add_api_route("/path", handler, methods=["POST", "DELETE"])`, all methods were getting the same auto-generated `operationId` (based only on the first method), causing `UserWarning: Duplicate Operation ID` warnings.

### Code Fix (`fastapi/openapi/utils.py`)

Modified `get_openapi_operation_metadata()` to append the method name to the `operationId` when:
1. No explicit `operation_id` is provided (uses auto-generated `unique_id`)
2. The route has multiple methods

```python
operation_id = route.operation_id or route.unique_id
# When a route has multiple methods and uses auto-generated unique_id,
# append the method to make the operation_id unique for each method
if not route.operation_id and len(route.methods) > 1:
    operation_id = f"{operation_id}_{method.lower()}"
```

### Test Cases (`tests/test_generate_unique_id_function.py`)

Added 4 new test cases:
- `test_multiple_methods_unique_operation_id` - Verifies POST and DELETE get unique IDs
- `test_three_methods_unique_operation_id` - Verifies 3 methods all get unique IDs
- `test_explicit_operation_id_multiple_methods` - Verifies explicit operation_id still triggers warning (expected behavior)
- `test_single_method_no_extra_suffix` - Verifies single method routes remain unchanged

### Before (Bug)
```python
router.add_api_route("/clear", clear_endpoint, methods=["POST", "DELETE"])
```
Generated OpenAPI:
```json
{
  "/clear": {
    "post": {"operationId": "clear_endpoint_clear_post"},
    "delete": {"operationId": "clear_endpoint_clear_post"}  // ❌ Same!
  }
}
```
Warning: `UserWarning: Duplicate Operation ID clear_endpoint_clear_post`

### After (Fixed)
```python
router.add_api_route("/clear", clear_endpoint, methods=["POST", "DELETE"])
```
Generated OpenAPI:
```json
{
  "/clear": {
    "post": {"operationId": "clear_endpoint_clear_post_post"},      // ✅ Unique
    "delete": {"operationId": "clear_endpoint_clear_post_delete"}   // ✅ Unique
  }
}
```
No warnings.

## Backward Compatibility

✅ **Single method routes**: Unchanged (no extra suffix added)  
✅ **Explicit `operation_id`**: Respected as-is (user's intentional choice)  
[...]

PR #15194, opened 2026-03-22
🐛 Fix duplicate operation IDs for routes with multiple methods (#13175)

r266-tech:
## Summary

When a route is registered with multiple HTTP methods (e.g., `methods=['POST', 'DELETE']`), the default `generate_unique_id` function only uses the first method (alphabetically from the set), causing duplicate operation IDs in the OpenAPI schema.

## Reproduction

```python
from fastapi import FastAPI

app = FastAPI()

def handler():
    return {'msg': 'ok'}

app.add_api_route('/clear', handler, methods=['POST', 'DELETE'])

client = TestClient(app)
resp = client.get('/openapi.json')
# Before fix: both POST and DELETE get operationId 'handler_clear_post'
# After fix: POST gets 'handler_clear_post', DELETE gets 'handler_clear_delete'
```

## Fix

In `fastapi/openapi/utils.py` → `get_openapi_operation_metadata()`:

When a route has multiple methods and no explicit `operation_id`, the current code uses `route.unique_id` (which is based on only the first method). The fix detects this case and generates a method-specific operation ID by replacing the first-method suffix with the actual current method.

## Related

- Fixes #13175
- Related: #4650 (introduced `generate_unique_id`)

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -242,7 +242,20 @@ def get_openapi_operation_metadata(
     operation["summary"] = generate_operation_summary(route=route, method=method)
     if route.description:
         operation["description"] = route.description
-    operation_id = route.operation_id or route.unique_id
+    if route.operation_id:
+        operation_id = route.operation_id
+    elif len(route.methods) > 1:
+        # When a route handles multiple methods, the default unique_id is based on
+        # the first method only (via generate_unique_id). Append the actual method
+        # to avoid duplicate operation IDs across methods. See #13175.
+        base = route.unique_id
+        # Remove the trailing _<method> suffix that generate_unique_id appends
+        first_method = f"_{list(route.methods)[0].lower()}"
+        if base.endswith(first_method):
+            base = base[: -len(first_method)]
+        operation_id = f"{base}_{method.lower()}"
+    else:
+        operation_id = route.unique_id
     if operation_id in operation_ids:
         endpoint_name = getattr(route.endpoint, "__name__", "<unnamed_endpoint>")
         message = f"Duplicate Operation ID {operation_id} for function {endpoint_name}"
--- a/tests/test_generate_unique_id_function.py
+++ b/tests/test_generate_unique_id_function.py
@@ -1697,3 +1697,30 @@ def test_warn_duplicate_operation_id():
         ]
         assert len(duplicate_warnings) > 0
         assert "Duplicate Operation ID" in str(duplicate_warnings[0].message)
+
+
+def test_unique_operation_id_for_multi_method_route():
+    """A route with multiple methods should get unique operation IDs per method.
+
+    Regression test for https://github.com/fastapi/fastapi/issues/13175
+    """
+    app = FastAPI()
+
+    def handler():
+        return {"msg": "ok"}
+
+    app.add_api_route("/clear", handler, methods=["POST", "DELETE"])
+
+    client = TestClient(app)
+    response = client.get("/openapi.json")
+    data = response.json()
+    paths = data["paths"]
+
+    op_ids = []
+    for methods in paths.values():
+        for details in methods.values():
+            op_ids.append(details["operationId"])
+
+    assert len(op_ids) == 2, f"Expected 2 operations, got {len(op_ids)}"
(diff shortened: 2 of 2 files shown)


themavik (commented):
`list(route.methods)[0]` is not deterministic when `methods` is a set (iteration order). That can change which suffix gets stripped between runs and diverge from how `unique_id` was built. Use a stable ordering (e.g. `sorted(route.methods)`) consistent with `generate_unique_id`, or derive the base from the same source `unique_id` uses.

YuriiMotov:
@r266-tech, have you reviewed other PRs linked to that issue?
Please, describe why this approach is better

r266-tech:
Hi @YuriiMotov, yes, I've reviewed the other PRs. Here's a comparison:

**Other approaches:**

- **#13861, #15070** — Make the method selection deterministic (e.g., sorted), but don't fix the duplicate  problem itself. Two methods on the same route still produce the same ID.
- **#14827, #15009, #14869** (closed) — Earlier attempts; similar strategies but without robust test coverage.
- **#15074** (closed) — Appends the current `method` to `route.unique_id` directly. Since `unique_id` already ends in `_<first_method>`, this produces IDs like `handler_clear_post_post` for the first method — incorrect.
- **#15140** (closed) — Strips everything after the last underscore from `unique_id` as the base. This breaks when route names or paths contain underscores (e.g., `my_handler_/my_path_post` → strips too much).
- **#15215** — Rebuilds the base ID from scratch using `re.sub(r"\W", "_", f"{route.name}{route.path_format}")`. This bypasses `generate_unique_id_function` entirely for multi-method routes, ignoring custom ID generators.
- **#15222** — Modifies `generate_unique_id` to omit the method suffix when `len(route.methods) > 1`, then appends it per-method in `get_openapi_operation_metadata`. Clean root-level fix, but it changes `route.unique_id` itself, which is used beyond OpenAPI (e.g., client generation, internal routing). This is a more invasive change.

**Our approach (#15194):**

We fix the issue entirely within `get_openapi_operation_metadata`, which is the correct place: that function is called once per method and already has `method` in scope. We:
1. Detect multi-method routes
2. Strip the first-method suffix from `route.unique_id` to recover the base
3. Append the actual `method` for the current operation

The fix is scoped to OpenAPI generation and doesn't change `route.unique_id` or `generate_unique_id`.

**Known limitation:** The suffix-stripping logic (`base[: -len(first_method)]`) assumes the default `generate_unique_id_function` format. If a custom `generate_unique_id_function` is used that doesn't end in `_<method>`, the base recovery could be wrong.

**#15215 and #15222's approach** of rebuilding the base in `openapi/utils.py` (either from `route.name + route.path_format` or by omitting the suffix in `generate_unique_id`) is arguably cleaner for the custom-generator case. However, #15222's route modifies `unique_id` at a lower level, while #15215 bypasses custom generators.

Our approach is the most minimal change: a targeted fix in OpenAPI generation only, with no behavioral change for single-method routes or explicit `operation_id`. If the custom-generator edge case is a concern, we can refine the base-extraction logic to be more robust.

YuriiMotov (commented):
This way if the endpoint has request body, in some cases the name and title of request body model in openapi schema will be strange (e.g. include `_put` for POST path operation) and non-deterministic (sometimes `_put`, sometimes `_post` because `route.methods` is a `set` (order is not guaranteed)).

```py
from typing import Annotated

def do_something(p: Annotated[str, Body(embed=True)]):
    return {"message": "Hi!"}

app = FastAPI()

app.add_api_route("/", do_something, methods=["POST", "PUT"])
```

```
  "paths": {
    "/": {
      "put": {
        "summary": "Do Something",
        "operationId": "do_something__put",
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/Body_do_something__post"  <= This will be sometimes _put sometimes _post
              }
            }
          },
          "required": true
        },
```

You probably missed this comment: https://github.com/fastapi/fastapi/pull/15074#pullrequestreview-3908105706 (see point 2)

I think the approach of https://github.com/fastapi/fastapi/pull/10694 is better.

Anyway, thanks!

PR #15195, opened 2026-03-22
📝 Fix: clarify 'Required, can be None' query parameter documentation

r266-tech:
## Summary

This PR fixes the misleading "Required, can be `None`" section in the query parameter documentation (issue #12419).

### Problem
The documentation stated:
> You can declare that a parameter can accept `None`, but that it's still required. This would force clients to send a value, even if the value is `None`.

This is incorrect. When a query parameter is required (no default value), omitting it returns a **422 Validation Error**. There is no standard HTTP mechanism to pass `None` as a query parameter value.

### Changes
1. **Documentation** (`docs/en/docs/tutorial/query-params-str-validations.md`):
   - Rewrote the section to clarify that `None` in the type annotation means the OpenAPI schema marks the parameter as nullable
   - Added an info box explaining that omitting the parameter returns 422 and that the `None` type is for schema representation

2. **Tests** (`tests/test_tutorial/test_query_params_str_validations/test_tutorial006c.py`):
   - Removed `@pytest.mark.xfail` from `test_query_params_str_validations_no_query` and `test_query_params_str_validations_empty_str`
   - Updated assertions to verify the correct 422 behavior instead of expecting impossible 200 responses

### Related
- Closes #12419

### DIFF
--- a/docs/en/docs/tutorial/query-params-str-validations.md
+++ b/docs/en/docs/tutorial/query-params-str-validations.md
@@ -222,9 +222,15 @@ So, when you need to declare a value as required while using `Query`, you can si
 
 ### Required, can be `None` { #required-can-be-none }
 
-You can declare that a parameter can accept `None`, but that it's still required. This would force clients to send a value, even if the value is `None`.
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Please, search for existing PRs before opening your own

See https://github.com/fastapi/fastapi/pull/14322

PR #15196, opened 2026-03-23
Upgrade Starlette to `>=0.48.0`

vishds:
## Summary
Starlette had a breaking change in 0.48.0 where the it removed the deprecated `status.HTTP_422_UNPROCESSABLE_ENTITY`.

In https://github.com/fastapi/fastapi/pull/14077, FastAPI internally transitioned to use literal `422` and widened the version range to `>=0.40.0,<0.49.0`. This broke builds for users using the constant from fastapi.status.

I had raised this issue in https://github.com/fastapi/fastapi/discussions/14812. From the response I received I felt it was more important to support Starlette versions `0.40.0` to `0.48.0`. I fixed builds at my end my just pinning the starlette in my project.

Recently in https://github.com/fastapi/fastapi/pull/15022 the minimum Starlette version got increased to `0.46.0`. If there are no other practical issues, I believe it is a nice time to increase the minimum version to `0.48.0` and get rid of this issue.

### DIFF
--- a/docs_src/handling_errors/tutorial005_py310.py
+++ b/docs_src/handling_errors/tutorial005_py310.py
@@ -1,4 +1,4 @@
-from fastapi import FastAPI, Request
+from fastapi import FastAPI, Request, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse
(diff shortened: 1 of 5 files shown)


themavik (commented):
curl endpoint sanity check for fastapi 15196

themavik (commented):
Using HTTP_422_UNPROCESSABLE_CONTENT keeps OpenAPI 422 keys aligned with Starlette after the rename. starlette>=0.48.0 in pyproject is a hard floor; anyone pinned below that must bump Starlette before upgrading FastAPI.

YuriiMotov:
This is not a FastAPI's issue. It's your project incompatibility with certain versions of Starlette, so you should address it in your project.
We tend not to narrow the range of supported Starlette versions without strong reasons

PR #15203, opened 2026-03-23
Fix #12419: Update "Required, can be None" docs and fix example code

r266-tech:
## Summary

Fixes #12419.

The documentation section "Required, can be None" claimed that a query parameter without a default value can accept `None`, but this is misleading. In HTTP query parameters, you cannot send a literal `None` value, and omitting a required parameter results in a 422 validation error.

As noted by @Kludex in the issue, the documentation is wrong and the tests were marked as `xfail`.

## Changes

- **docs/en/docs/tutorial/query-params-str-validations.md**: Reworded section from "Required, can be `None`" to "Optional with `None`", clarified that using `= None` as default makes the parameter optional
- **docs_src/.../tutorial006c_an_py310.py**: Added `= None` default to make parameter truly optional
- **docs_src/.../tutorial006c_py310.py**: Added `default=None` to `Query()`
- **tests/.../test_tutorial006c.py**: Removed `xfail` markers, updated OpenAPI schema expectation from `required: True` to `required: False`

## Behavior after fix

| Request | Status | Response |
|---------|--------|----------|
| `GET /items/` | 200 | `{"items": [{"item_id": "Foo"}, {"item_id": "Bar"}]}` |
| `GET /items/?q=` | 200 | `{"items": [{"item_id": "Foo"}, {"item_id": "Bar"}]}` |
| `GET /items/?q=query` | 200 | `{"items": [...], "q": "query"}` |
| `GET /items/?q=fa` | 422 | Validation error (min_length=3) |

Note: Translations of this doc section will also need updating.

### DIFF
--- a/docs/en/docs/tutorial/query-params-str-validations.md
+++ b/docs/en/docs/tutorial/query-params-str-validations.md
@@ -220,13 +220,19 @@ So, when you need to declare a value as required while using `Query`, you can si
 
 {* ../../docs_src/query_params_str_validations/tutorial006_an_py310.py hl[9] *}
 
-### Required, can be `None` { #required-can-be-none }
+### Optional with `None` { #required-can-be-none }
 
-You can declare that a parameter can accept `None`, but that it's still required. This would force clients to send a value, even if the value is `None`.
+You can declare that a parameter is optional and can accept `None`, even when using `Query` for validations.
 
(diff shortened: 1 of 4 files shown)


YuriiMotov:
Duplicate of https://github.com/fastapi/fastapi/pull/15195

PR #15205, opened 2026-03-23
refactor: raise clear ValueError when Depends receives a non-callable

Yume05-dev:
Refactor and technical improvements for the issue resolution. Ready for review.

### DIFF
--- a/fastapi/params.py
+++ b/fastapi/params.py
@@ -748,6 +748,12 @@ class Depends:
     use_cache: bool = True
     scope: Literal["function", "request"] | None = None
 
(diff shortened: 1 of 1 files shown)


YuriiMotov:
I actually think it's already quite clear:

```
  File "/usr/lib/python3.11/inspect.py", line 2456, in _signature_from_callable
    raise TypeError('{!r} is not a callable object'.format(obj))
TypeError: <t.Item object at 0x74a8eb825cd0> is not a callable object
```

So, I think we don't need to change this.
But if we decide to do it, we will need to add a test

PR #15208, opened 2026-03-24
docs: improve consistency in bullet list formatting

HarshaG2005:
Capitalized a bullet point to maintain consistency with other items in the list.


YuriiMotov:
@HarshaG2005, thanks, but I think this is not something important..

Also, it will not work as README is automatically generated from `index.md`

Anyway, thanks for your interest and efforts!

HarshaG2005:
> @HarshaG2005, thanks, but I think this is not something important..
> 
> Also, it will not work as README is automatically generated from `index.md`
> 
> Anyway, thanks for your interest and efforts!

Got it, thanks for explaining! I’ll look into other ways to contribute.

PR #15212, opened 2026-03-24
refactor: Move type-only imports into TYPE_CHECKING blocks

hnshah:
## Description

Moves imports that are only used for type annotations into `TYPE_CHECKING` blocks to prevent unnecessary runtime imports and potential circular import issues.

## Changes
- `fastapi/openapi/utils.py`: Move `Response` import
- `tests/test_dependency_contextmanager.py`: Move `StreamingResponse` import  
- `tests/test_route_scope.py`: Move `APIRoute` import

## Detection
Found via: `ruff check . --select TC002`

## Benefits
- Reduces runtime import overhead
- Prevents potential circular import issues
- Improves type checking performance
- Follows Python typing best practices

## Testing
- ✅ Existing tests pass
- ✅ Type checking still works correctly
- ✅ No runtime behavior changes

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -3,7 +3,7 @@ import http.client
 import inspect
 import warnings
 from collections.abc import Sequence
-from typing import Any, Literal, cast
+from typing import Any, Literal, cast, TYPE_CHECKING
 
 from fastapi import routing
 from fastapi._compat import (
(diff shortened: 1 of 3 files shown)


YuriiMotov:
If you use FastAPI, it will still import these types under the hood. So, this is useless..
These are not third-party imports

PR #15214, opened 2026-03-24
fix(routing): thread stream_item_type through include_router()

BillionClaw:
When a typed SSE or JSONL streaming endpoint is registered via a router, the OpenAPI schema loses the stream item type because include_router() re-creates routes via add_api_route() without passing stream_item_type. This fix threads stream_item_type through add_api_route() and include_router(), following the strict_content_type pattern. Fixes #15077


YuriiMotov:
Duplicate of https://github.com/fastapi/fastapi/pull/15077

PR #15215, opened 2026-03-24
Fix duplicate operationId for routes with multiple HTTP methods

fitratgulmamadov:
## Description

When a route is registered with multiple HTTP methods:

```python
router.add_api_route("/items/", endpoint, methods=["POST", "DELETE"])
```
FastAPI generates the same operationId for all methods and emits a UserWarning: Duplicate Operation ID. This breaks OpenAPI clients and auto-generated SDKs.

Root cause: generate_unique_id() only takes the first method from the set (list(route.methods)[0]), so all methods on the same route share the same unique_id.

Fix: In get_openapi_operation_metadata() — which already receives the current method as a parameter — detect multi-method routes and build a per-method operationId using the same pattern as generate_unique_id.

Result:

POST /things/ → create_or_delete_things_things__post
DELETE /things/ → create_or_delete_things_things__delete
No duplicate warnings ✓
Fixes #13175

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -1,6 +1,7 @@
 import copy
 import http.client
 import inspect
+import re
 import warnings
 from collections.abc import Sequence
 from typing import Any, Literal, cast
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@fitratgulmamadov, have you reviewed other PRs linked to that issue?
Please, describe why this approach is better

fitratgulmamadov:
> @fitratgulmamadov, have you reviewed other PRs linked to that issue? Please, describe why this approach is better

Yes, I reviewed the linked PRs before submitting.

**Why this approach differs from rejected PRs:**

- **#15070**: Only sorted methods for determinism, didn't make IDs unique per method — not a real fix.
- **#15074 / #15140**: Appended method to `route.unique_id`, producing double suffixes like `do_something__put_put`. Our approach rebuilds the base from `route.name + route.path_format` using the same pattern as `generate_unique_id`, so the result is clean: `do_something__put` and `do_something__delete`.
- **#10694**: Modifies routing layer. Our fix is more minimal — it stays in `get_openapi_operation_metadata()` where `method` is already available, touching only the OpenAPI generation step.

**Known limitation**: The body schema name (`Body_do_something__put`) is still generated at route compilation time and remains non-deterministic for multi-method routes. This would require a separate fix at the routing layer (as #10694 attempts). Our PR addresses only the operationId duplication.

Happy to adjust the approach based on your feedback.

fitratgulmamadov:
Have it any chances to merge? @YuriiMotov

tiangolo:
Thanks for the interest!

FastAPI doesn't support / document this pattern `router.add_api_route("/items/", endpoint, methods=["POST", "DELETE"])`.

Supporting it would mean having to keep backwards compatibility with it. And I consider mutiple HTTP operations in the same function an anti-pattern. So I prefer not to encourage it. Thanks! :coffee:

PR #15217, opened 2026-03-25
docs: warn about BackgroundTasks + Response background interaction

### LINKED ISSUE #11215: Potential footgun when using custom `Response(background=<Task()>)` in conjunction with injected `BackgroundTasks` - the custom response overwrites the tasks

ChunkyTortoise:
## Summary

Closes #11215

Adds a warning to the "Return a Response Directly" docs page explaining that returning a `Response` with a `background` parameter silently overrides any `BackgroundTasks` added via dependency injection.

## Problem

When an endpoint uses both `BackgroundTasks` dependency injection and returns a `Response(background=BackgroundTask(...))`, the injected tasks are silently dropped. This happens because `fastapi/routing.py` only assigns injected background tasks when `response.background is None`:

```python
if raw_response.background is None:
    raw_response.background = solved_result.background_tasks
```

This is a common footgun with no runtime warning -- tasks simply don't execute.

## Changes

- Added a `/// warning` block to `docs/en/docs/advanced/response-directly.md` explaining the interaction
- Created `docs_src/response_directly/tutorial003.py` with correct and incorrect patterns
- Follows existing callout conventions used in `docs/en/docs/tutorial/background-tasks.md`

## Motivation

I use FastAPI `BackgroundTasks` across 5 production APIs and nearly hit this exact issue when refactoring webhook delivery to return custom `Response` objects. The current docs don't mention this interaction anywhere.

## Test plan

- [x] Warning uses `/// warning` block syntax (FastAPI convention)
- [x] Example file follows existing tutorial naming pattern
- [x] Example is importable Python

### DIFF
--- a/docs/en/docs/advanced/response-directly.md
+++ b/docs/en/docs/advanced/response-directly.md
@@ -74,6 +74,20 @@ When using a `response_model` or return type, FastAPI won't use the `jsonable_en
 
 Instead it takes the JSON bytes generated with Pydantic using the response model (or return type) and returns a `Response` with the right media type for JSON directly (`application/json`).
 
+## Background Tasks and Custom Responses { #background-tasks-and-custom-responses }
+
+/// warning
+
+If you return a `Response` with a `background` parameter **and** also use the `BackgroundTasks` dependency injection, the injected background tasks will be **silently ignored**.
+
+FastAPI only assigns injected background tasks to the response when `response.background is None`. If your `Response` already has a background task attached, the injected tasks are dropped without any error or warning.
+
+Use one mechanism or the other, not both in the same endpoint:
+
+{* ../../docs_src/response_directly/tutorial003.py hl[18,25] *}
+
+///
+
 ## Notes { #notes }
 
 When you return a `Response` directly its data is not validated, converted (serialized), or documented automatically.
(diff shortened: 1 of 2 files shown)


YuriiMotov (commented):
We are waiting for Sebastian to review this issue and take a decision on how to resolve this:
* Merge BG tasks
* Document + Emit a warning
* Just document

If we decide to go with 3rd approach then we can take this PR and work a bit more on it (add tests (all code examples must be covered by tests), modify text to follow the style of other pages (probably split code example into good and bad, and put explanations as text outside of code examples)).

Let's wait for Sebastian to take a decision

ChunkyTortoise:
Thanks @YuriiMotov. To help the decision: this PR is the "just document" option. It only adds a docs admonition to `advanced/response-directly.md` plus an importable example, with no `routing.py` change and no runtime warning. So if the team lands on document-only, it's ready as-is; if you'd prefer merging the background tasks or emitting a runtime warning, I'm happy to adapt or close.

The interaction is still present on `master` (injected `BackgroundTasks` are silently dropped when a `Response(background=...)` is returned), so some form of doc seems worth keeping either way.

PR #15220, opened 2026-03-25
fix: treat empty bearer token as missing credentials in OAuth2 helpers

subhashdasyam:
## Summary

`OAuth2PasswordBearer` and `OAuth2AuthorizationCodeBearer` accept `Authorization: Bearer ` (empty token) and return `""` instead of triggering the `auto_error` path.

Fixes the issue raised in https://github.com/fastapi/fastapi/discussions/15192.

## Root Cause

`get_authorization_scheme_param("Bearer ")` returns `("Bearer", "")`. Both `__call__` methods only checked that the scheme was `"bearer"` — never that the credential string was non-empty.

## Fix

Add `not param` to the guard condition in both `OAuth2PasswordBearer.__call__` and `OAuth2AuthorizationCodeBearer.__call__`:

```python
# before
if not authorization or scheme.lower() != "bearer":

# after
if not authorization or scheme.lower() != "bearer" or not param:
```

## Spec Reference

RFC 6750 Section 2.1 defines the grammar as `credentials = "Bearer" 1*SP b64token` where `b64token = 1*(...)` — at least one character required. An empty credential is syntactically malformed and should be treated the same as missing credentials.

## Behavior

| Request | `auto_error=True` | `auto_error=False` |
|---|---|---|
| No `Authorization` header | 401 | `None` |
| `Authorization: Bearer ` (before) | 200, `token=""` | `""` |
| `Authorization: Bearer ` (after) | 401 | `None` |
| `Authorization: Bearer validtoken` | 200 | `"validtoken"` |

## Tests

- `test_empty_bearer_token` in `test_security_oauth2_password_bearer_optional.py` (`auto_error=False` → `None`)
- `test_empty_bearer_token` in `test_security_oauth2_authorization_code_bearer.py` (`auto_error=True` → 401)

### DIFF
--- a/fastapi/security/oauth2.py
+++ b/fastapi/security/oauth2.py
@@ -536,7 +536,7 @@ class OAuth2PasswordBearer(OAuth2):
     async def __call__(self, request: Request) -> str | None:
         authorization = request.headers.get("Authorization")
         scheme, param = get_authorization_scheme_param(authorization)
-        if not authorization or scheme.lower() != "bearer":
+        if not authorization or scheme.lower() != "bearer" or not param:
             if self.auto_error:
                 raise self.make_not_authenticated_error()
             else:
@@ -642,7 +642,7 @@ class OAuth2AuthorizationCodeBearer(OAuth2):
     async def __call__(self, request: Request) -> str | None:
         authorization = request.headers.get("Authorization")
         scheme, param = get_authorization_scheme_param(authorization)
-        if not authorization or scheme.lower() != "bearer":
+        if not authorization or scheme.lower() != "bearer" or not param:
             if self.auto_error:
                 raise self.make_not_authenticated_error()
             else:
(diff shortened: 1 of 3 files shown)


subhashdasyam:
based on the discussion https://github.com/fastapi/fastapi/discussions/15192

mahdirajaee (commented):
Clean and well-targeted fix. A few observations:

1. **RFC alignment is correct.** RFC 6750 Section 2.1 requires `b64token = 1*( ALPHA / DIGIT / ... )` — at least one character. Treating an empty credential as missing is the right call.

2. **Guard ordering is sensible.** Appending `or not param` at the end means the short-circuit evaluation still skips redundant checks when `authorization` is falsy. No performance concern.

3. **Both classes are covered.** Good that this patches both `OAuth2PasswordBearer` and `OAuth2AuthorizationCodeBearer` — they had identical code paths and the same gap.

4. **Test coverage is solid.** Testing both the `auto_error=True` (401) and `auto_error=False` (`None`) paths for the empty-token case ensures no regression on either behavior.

One minor thought: `get_authorization_scheme_param` splits on the first space, so `"Bearer  "` (double space) yields `param = " "` which is truthy and would pass this new guard. That's arguably correct since the credential is non-empty (just whitespace), but worth noting if someone later expects `.strip()` behavior. Not a blocker — the current fix handles the realistic scenario cleanly.

Looks good overall.

PR #15222, opened 2026-03-26
Fix duplicate OpenAPI operationIds for multi-method routes

abishekgiri:
Writing
Summary

Fixes duplicate OpenAPI operationId generation for multi-method routes.

Previously, routes registered with multiple HTTP methods could end up with conflicting default operationId values. This change keeps explicit operation_id values untouched, preserves the existing behavior for single-method routes, and makes multi-method routes generate a shared base ID that is suffixed per HTTP method when building OpenAPI operations.

Changes
updated default unique ID handling for multi-method routes
kept explicit operation_id unchanged
added regression tests for:
default generate_unique_id_function
custom generate_unique_id_function
Verification

Tested with:

uv run pytest tests/test_generate_unique_id_function.py
uv run pytest tests/test_extra_routes.py

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -233,6 +233,13 @@ def generate_operation_summary(*, route: routing.APIRoute, method: str) -> str:
     return route.name.replace("_", " ").title()
 
 
+def get_openapi_operation_id(*, route: routing.APIRoute, method: str) -> str:
+    operation_id = route.operation_id or route.unique_id
+    if route.operation_id is None and route.methods and len(route.methods) > 1:
+        operation_id = f"{operation_id}_{method.lower()}"
+    return operation_id
+
+
 def get_openapi_operation_metadata(
     *, route: routing.APIRoute, method: str, operation_ids: set[str]
 ) -> dict[str, Any]:
@@ -242,7 +249,7 @@ def get_openapi_operation_metadata(
     operation["summary"] = generate_operation_summary(route=route, method=method)
     if route.description:
         operation["description"] = route.description
-    operation_id = route.operation_id or route.unique_id
+    operation_id = get_openapi_operation_id(route=route, method=method)
     if operation_id in operation_ids:
         endpoint_name = getattr(route.endpoint, "__name__", "<unnamed_endpoint>")
         message = f"Duplicate Operation ID {operation_id} for function {endpoint_name}"
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -96,7 +96,8 @@ def generate_unique_id(route: "APIRoute") -> str:
     operation_id = f"{route.name}{route.path_format}"
     operation_id = re.sub(r"\W", "_", operation_id)
     assert route.methods
-    operation_id = f"{operation_id}_{list(route.methods)[0].lower()}"
+    if len(route.methods) == 1:
+        operation_id = f"{operation_id}_{sorted(route.methods)[0].lower()}"
     return operation_id
 
 
--- a/tests/test_generate_unique_id_function.py
+++ b/tests/test_generate_unique_id_function.py
@@ -1697,3 +1697,44 @@ def test_warn_duplicate_operation_id():
         ]
         assert len(duplicate_warnings) > 0
         assert "Duplicate Operation ID" in str(duplicate_warnings[0].message)
+
+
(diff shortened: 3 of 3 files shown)


abishekgiri:
Hi, I implemented a fix for this and added regression tests.

The change makes multi-method routes generate distinct OpenAPI `operationId` values per HTTP method, while leaving explicit `operation_id` values untouched.

I also verified it with:

* `uv run pytest tests/test_generate_unique_id_function.py`
* `uv run pytest tests/test_extra_routes.py`

I’m opening a PR now.

abishekgiri:
Thanks! It looks like the failing check is only due to the missing required PR label.

This PR is a bug fix, so the appropriate label should be `bug`.

I don’t think I have permission to add labels on the upstream repo, so could a maintainer or triager please add the `bug` label?

YuriiMotov:
Please review other PRs linked to the issue and explain why you think your approach is better than others

abishekgiri:
> Please review other PRs linked to the issue and explain why you think your approach is better than others

I reviewed the other PRs linked to `#13175` (`#13861`, `#14881`, `#15070`, `#15074`, `#15140`, `#15194`, and `#15215`).

My reasoning for this approach is:

- `#13861`, `#14881`, and `#15070` make the selected method deterministic, but they do not solve the duplicate `operationId` problem itself.
- `#15074` appends the current method to `route.unique_id`, which already contains a method suffix in the default generator, so it can produce IDs like `..._post_post`.
- `#15140` and `#15194` are closer, but they still have to reverse-engineer a base ID from `route.unique_id`. That is brittle for custom `generate_unique_id_function` values, especially `#15140`, which strips everything after the last underscore.
- `#15215` rebuilds the base from `route.name + route.path_format` inside OpenAPI, which bypasses `generate_unique_id_function` for multi-method routes.

I chose this implementation because it fixes the root ambiguity instead of patching around it later:
- for multi-method routes, the default `unique_id` becomes a shared base
- the per-method suffix is added only when emitting OpenAPI
- explicit `operation_id` remains unchanged
- custom `generate_unique_id_function` is still respected, because OpenAPI starts from `route.unique_id` instead of reconstructing it

I think that is cleaner because `route.unique_id` is reused internally beyond `operationId`, so removing the arbitrary “first method” from the canonical route ID is more robust than preserving it and then trying to peel it back off in OpenAPI generation.

I also added regression coverage for both the default generator and a custom `generate_unique_id_function`, because that was the main gap in the other proposals.

mahdirajaee (commented):
Good catch on the multi-method route collision. The approach is sound — here are some detailed observations:

1. **`generate_unique_id` change is correct.** For single-method routes, the behavior is unchanged. For multi-method routes, the base ID no longer includes a method suffix, which avoids the arbitrary "first method wins" problem. Using `sorted(route.methods)` for the single-method case also removes non-determinism from set iteration — nice side effect.

2. **`get_openapi_operation_id` — clean separation.** Extracting this into its own function keeps `get_openapi_operation_metadata` focused on building the operation dict. The logic of "if no explicit operation_id and multiple methods, append `_{method}`" reads clearly.

3. **Explicit `operation_id` is untouched.** The `if route.operation_id is None` guard in `get_openapi_operation_id` ensures that users who set a custom `operation_id` don't get an unexpected suffix. That's the right behavior — if someone explicitly sets a duplicate, they'll still get the existing warning.

4. **Tests cover both default and custom `generate_unique_id_function`.** The second test (`test_multi_method_route_unique_operation_ids_with_custom_generate_unique_id`) is important because it validates that the suffix logic in `get_openapi_operation_id` works regardless of how the base ID is produced.

One consideration: if a custom `generate_unique_id_function` already appends a method suffix for multi-method routes, the new code in `get_openapi_operation_id` would append a second one (e.g., `foo_clear_post_post`). That's because the function checks `route.operation_id is None` but not whether the custom generator already handled multi-method disambiguation. Probably an edge case not worth guarding against since the existing contract doesn't guarantee anything about the generator's output format, but worth a note in the docstring if this gets a public-facing helper.

Solid fix overall.

YuriiMotov:
@abishekgiri, thank you for reviewing existing PRs, but it seems that you missed one: https://github.com/fastapi/fastapi/pull/10694

You approach is actually the same as in #10694. So, let's close this as duplicate and continue with #10694

Thanks!

PR #15224, opened 2026-03-26
Add explanation for jsonable_encoder in reference docsAdd short explanation for jsonable_encoder in reference docs

dheekshae:
Added a short explanation for `jsonable_encoder` to improve clarity and consistency with other reference pages.


YuriiMotov:
I think what we currently have is enough:
image

---

I think there is no need to add anything here.
Thanks!

PR #15227, opened 2026-03-26
Fix unstable operation IDs for multi-method routes

Esubaalew:
The current code picks the method from a set, so the generated ID can change depending on iteration order. That makes the schema unstable and can affect generated clients.

made  method selection deterministic by sorting the methods before building the ID. I also added a regression test to make sure the generated ID stays stable even if the method iteration order changes.


YuriiMotov:
@Esubaalew, thanks for your interest.

This way operation IDs will be duplicated for both methods, but they should be unique.

Also, please review existing PRs linked to the [issue](https://github.com/fastapi/fastapi/issues/13175) before opening PR

PR #14984, opened 2026-02-24
Fix: propagate Mount routes in include_router

fcarvajalbrown:
##Overview

Fixes #10180 

APIRouter.include_router was silently ignoring any routing. 
Mount entries when copying routes to the router application parent. 
Added an elif isinstance(route, routing.Mount) branch to handle them.
Then tested to confirm a sub-app mounted on an APIRouter is reachable through the parents FastAPI app. 

## Notes
Can provide test file, but was not sure where to place it so didn't add to commit. 

Fix added before include_router is called (ordering requirement to prevent the mentioned bug).

##Need BUG label for Checks

PR #15233, opened 2026-03-27
Fix mounting sub-applications under APIRouter with prefix support

abishekgiri:
## Summary

Fix mounting sub-applications when they are registered on an `APIRouter` and then included into a FastAPI app.

Previously, mounted sub-applications were not preserved when using `include_router()`, causing requests like `/api/subapi/sub` to return `404 Not Found`. Additionally, router prefixes were not applied when mounting sub-applications via `APIRouter`.

This change ensures that mounted sub-applications behave consistently whether they are registered directly on the FastAPI app or through an APIRouter.

## Changes

* override `APIRouter.mount()` to correctly apply router prefixes to mounted sub-applications
* preserve mounted routes when routers are included using `include_router()`
* add regression tests covering:

  * router-level prefixes
  * `include_router(..., prefix=...)`
  * mounted sub-application routing and OpenAPI behavior

## Backward Compatibility

This change is backward compatible, as it only fixes previously inconsistent behavior without modifying existing explicit routing definitions.

## Verification

Tested with:

* `uv run pytest tests/test_router_mount.py`
* `uv run pytest tests/test_tutorial/test_sub_applications/test_tutorial001.py`
* `uv run pytest tests/test_include_route.py`

## Related

Fixes #10180

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1499,6 +1499,9 @@ class APIRouter(routing.Router):
         )
         self.routes.append(route)
 
+    def mount(self, path: str, app: ASGIApp, name: str | None = None) -> None:
+        super().mount(self.prefix + path, app=app, name=name)
+
     def websocket(
         self,
         path: Annotated[
@@ -1803,6 +1806,8 @@ class APIRouter(routing.Router):
                     include_in_schema=route.include_in_schema,
                     name=route.name,
                 )
+            elif isinstance(route, routing.Mount):
+                self.mount(prefix + route.path, route.app, route.name)
             elif isinstance(route, APIWebSocketRoute):
                 current_dependencies = []
                 if dependencies:
--- a/tests/test_router_mount.py
+++ b/tests/test_router_mount.py
@@ -1,0 +1,75 @@
+from fastapi import APIRouter, FastAPI
+from fastapi.testclient import TestClient
+from inline_snapshot import snapshot
+
+
+def get_sub_app() -> FastAPI:
+    sub_app = FastAPI()
+
+    @sub_app.get("/sub")
+    def read_sub():
+        return {"message": "Hello World from sub API"}
+
+    return sub_app
+
+
+def test_mount_sub_application_with_router_prefix():
+    router = APIRouter(prefix="/api")
+    router.mount("/subapi", get_sub_app())
+
+    app = FastAPI()
+    app.include_router(router)
+
+    client = TestClient(app)
+    response = client.get("/api/subapi/sub")
+
+    assert response.status_code == 200, response.text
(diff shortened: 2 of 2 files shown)


abishekgiri:
Hi, this PR fixes a bug in router mounting behavior, so the appropriate label should be `bug`.

I don’t have permission to add labels on the upstream repo, so could a maintainer or triager please add the `bug` label so the required checks can pass?

YuriiMotov:
@abishekgiri, thanks for your interest!

See https://github.com/fastapi/fastapi/pull/13944

abishekgiri:
> @abishekgiri, thanks for your interest!
> 
> Please, review existing PRs linked to the issue (including closed PRs, to understand the reasons they were closed) before opening your PRs. And provide an explanation why you think your approach is better.
> 
> See #13944

I reviewed the related PRs: #13944, #14793, #14032, #14924, #14926, #14984, and #15014.

My understanding is that the closed PRs were mostly closed for one of two reasons:
- they were partial fixes that only handled `Mount` propagation in `include_router()`
- or the discussion shifted toward a `StaticFiles`-only scope (`#13944`) rather than general mounted ASGI apps

That is why I chose this approach.

The include-only fixes (`#14032`, `#14984`, `#15014`) solve one symptom, but they do not fix the underlying inconsistency: `APIRouter.mount()` still does not apply the router prefix when the `Mount` is created. In contrast, this PR fixes both pieces together:
1. `APIRouter.mount()` applies the router prefix when creating the mount
2. `include_router()` preserves `Mount` routes instead of silently dropping them

I think that is cleaner because it makes `Mount` behave more like other route types: the route is normalized when it is added, and then inclusion just carries it forward.

Compared with `#13944`, this PR is intentionally broader. `#13944` is the right PR if the project decision is to support only `StaticFiles` on sub-routers and reject other ASGI apps. My PR instead keeps the existing `APIRouter.mount()` API working for general mounted sub-applications.

Compared with `#14793`, the core implementation is very close, so I do not want to overstate a difference there. The main extra value I tried to add here is regression coverage for mounted sub-app OpenAPI behavior, since the original issue also mentioned docs/spec URLs not working as expected.

So my view is:
- if the intended direction is `StaticFiles`-only, then `#13944` is the correct direction
- if the intent is to make the existing `APIRouter.mount()` API work for general mounted sub-applications, then I think the two-part fix in this PR is the right approach

If you prefer consolidating on `#14793` instead, I’m also fine with that.

YuriiMotov:
Thanks for your reply!

If the only difference with #14793 is the test for `/api/subapi/openapi.json`, I'd prefer to close this PR in order not to have duplicates.
If you think having such test is important, you can suggest adding this test in the review comment to that PR.

Thank you!

PR #15235, opened 2026-03-27
Fix: propagate Mount routes when including APIRouter

garagon:
## What

When a sub-application is mounted on an `APIRouter` via `router.mount()`, calling `app.include_router(router)` silently drops the mount. The sub-application becomes unreachable.

This PR adds handling for `starlette.routing.Mount` routes in `APIRouter.include_router()`.

### Root cause

`include_router()` iterates `router.routes` and handles `APIRoute`, `Route`, `APIWebSocketRoute` and `WebSocketRoute`. But it has no handler for `Mount`, which is what `router.mount()` creates. Mounted sub-applications are silently ignored.

### Why `router.prefix` is needed

When `APIRouter` adds an `APIRoute`, the prefix is baked into `route.path` at creation time (e.g. prefix="/api", path="/hello" becomes route.path="/api/hello"). But `Mount` paths are NOT prefixed. A mount at "/subapi" on a router with prefix="/api" has route.path="/subapi", not "/api/subapi".

So the fix uses `prefix + router.prefix + route.path` instead of just `prefix + route.path`.

## Reproduction

```python

app = FastAPI()
api_router = APIRouter(prefix="/api")

subapi = FastAPI()

@subapi.get("/sub")
def read_sub():
    return {"message": "Hello from sub API"}

api_router.mount("/subapi", subapi)
app.include_router(api_router)

client = TestClient(app)
assert client.get("/api/subapi/sub").status_code == 200  # Was: 404
```

## Tests

3 new tests added covering: basic mount under router, sub-app docs accessibility, multiple sub-apps under one router.

All 1846 existing tests pass. No regressions.

Fixes #10180

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1819,6 +1819,12 @@ class APIRouter(routing.Router):
                 self.add_websocket_route(
                     prefix + route.path, route.endpoint, name=route.name
                 )
+            elif isinstance(route, routing.Mount):
+                self.mount(
+                    prefix + router.prefix + route.path,
+                    app=route.app,
+                    name=route.name,
+                )
         for handler in router.on_startup:
             self.add_event_handler("startup", handler)
         for handler in router.on_shutdown:
--- a/test_mount_subrouter.py
+++ b/test_mount_subrouter.py
@@ -1,5 +1,6 @@
 """Test for issue #10180: Mounting sub-applications under APIRouter."""
-from fastapi import FastAPI, APIRouter
+
+from fastapi import APIRouter, FastAPI
 from starlette.testclient import TestClient
 
 
@@ -56,11 +57,13 @@ def test_mount_multiple_subapps():
     api_router = APIRouter(prefix="/v1")
 
     sub1 = FastAPI()
+
     @sub1.get("/hello")
     def hello():
         return {"msg": "hello"}
 
     sub2 = FastAPI()
+
     @sub2.get("/world")
     def world():
         return {"msg": "world"}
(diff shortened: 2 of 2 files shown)


YuriiMotov:
@garagon, thanks for your interest!

See https://github.com/fastapi/fastapi/pull/13944

garagon:
@YuriiMotov thanks for the pointer. I reviewed all related PRs:

- #13944 (yours, open): supports StaticFiles only, raises `FastAPIError` for other ASGI apps
- #14032 (sujin502, closed): generic Mount support, closed for inactivity
- #14924, #14926 (kimchikingdom, closed): generic Mount, you redirected to #13944
- #14793 (jonathan-fulton, open): generic Mount support
- #14984 (fcarvajalbrown, closed), #15014 (Anandesh-Sharma, closed): same approach as ours

I understand your position: sub-apps should be mounted to the root app, and only StaticFiles makes sense on a sub-router.

Looking at the issue comments, there are real use cases beyond StaticFiles that users are requesting:

- Mounting v1/v2 API versions under a shared prefix router
- Serving Dash dashboards alongside API routes on the same router
- Adding authentication dependencies to third-party ASGI apps via router
- Modular app composition where teams own routers with their own sub-apps

The current behavior silently drops Mount routes with no error and no warning. Your PR (#13944) solves this for StaticFiles and raises a clear error for everything else, which is better than silent failure.

Our approach propagates all Mount routes with the correct prefix. The technical difference is one line: `self.mount(prefix + router.prefix + route.path, app=route.app, name=route.name)`. The reason `router.prefix` is needed is that `APIRoute.path` has the prefix baked in at creation time, but `Mount.path` does not.

Both approaches are valid. If Sebastian decides StaticFiles-only is the right scope, #13944 is the right PR. If full Mount support is desired, the approach in this PR (and #14793) handles it.

Happy to close this if #13944 is the direction the project wants to take.

PR #15243, opened 2026-03-27
✨ Add support for generating a UUID on application startup

fadyNabil16:
### Summary
This PR introduces a mechanism to generate a unique UUID when the application starts. This UUID can be used to uniquely identify each running instance of the application, which is especially useful in distributed or containerized environments.

### Motivation
In modern deployments (e.g., Docker, Kubernetes), multiple instances of the same application may run concurrently. Having a unique identifier per instance helps with:

- Logging and debugging across multiple instances  
- Observability and tracing  
- Instance-level metrics and monitoring  

### Changes
- Generate a UUID during application startup  
- Ensure the UUID is created once per application lifecycle  

### Usage
The generated UUID can be accessed within the application (e.g., via app state or configuration) and used in logs, responses, or internal processing.

### DIFF
--- a/fastapi/applications.py
+++ b/fastapi/applications.py
@@ -1,6 +1,7 @@
 from collections.abc import Awaitable, Callable, Coroutine, Sequence
 from enum import Enum
 from typing import Annotated, Any, TypeVar
+import uuid
 
 from annotated_doc import Doc
 from fastapi import routing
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@fadyNabil16, thanks for your interest in project!

I think we don't need to add this. You can easily add such unique ID in the lifespan event.
If you still think this feature would be useful, please start a discussion and ask for feedback from community.

Let's close this for now

PR #15246, opened 2026-03-28
Fix mounting sub-applications under APIRouter

saitarrun:
## Summary

Closes #10180

Sub-apps mounted on an `APIRouter` were silently dropped when the router was included into a `FastAPI` app, causing every request to the mounted path to return `404`.

```python
app = FastAPI()
api_router = APIRouter(prefix="/api")
api_router.mount("/subapi", subapi)
app.include_router(api_router)

# Before this fix:
client.get("/api/subapi/sub")  # → 404

# After this fix:
client.get("/api/subapi/sub")  # → 200 ✓
```

## Root cause

Two missing pieces:

1. **`APIRouter` had no `mount()` override.** All other route-registration methods (`add_api_route`, `add_api_websocket_route`, …) prepend `self.prefix` to the path before storing the route. `mount()` was inherited from Starlette's `Router` and stored the path as-is, so the router's own prefix was never applied to mounted sub-apps.

2. **`include_router()` ignored `routing.Mount` routes.** The loop inside `include_router` handled `APIRoute`, `Route`, `APIWebSocketRoute`, and `WebSocketRoute` — but had no branch for `Mount`, so mounted sub-apps were silently skipped.

## Changes

- `fastapi/routing.py` — add `APIRouter.mount()` override that prepends `self.prefix` (mirrors `add_api_route`) and add a `routing.Mount` branch in `include_router` that forwards the mount with `prefix + route.path`.
- `tests/test_router_mount.py` — 6 tests covering: exact issue repro, `include_router(prefix=…)`, combined prefixes, nested routers, sub-app OpenAPI reachability, and existing direct-mount behaviour.

## Prefix composition

Both prefix sources compose correctly and consistently with ordinary endpoints:

| Source | Route path stored | Final path |
|---|---|---|
| `APIRouter(prefix="/api").get("/x")` | `/api/x` | `include_prefix + /api/x` |
| `APIRouter(prefix="/api").mount("/sub", …)` | `/api/sub` | `include_prefix + /api/sub` |
| `include_router(router, prefix="/v1")` | as above | `/v1/api/sub` |

## Test plan

- [x] 6 new tests all pass
- [x] All existing routing/WS tests pass (`test_include_route`, `test_router_events`, `test_ws_router`, `test_empty_router`, `test_router_prefix_with_template`)
- [x] Direct `FastAPI.mount()` behaviour is unaffected

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1499,6 +1499,18 @@ class APIRouter(routing.Router):
         )
         self.routes.append(route)
 
+    def mount(self, path: str, app: ASGIApp, name: str | None = None) -> None:
+        """
+        Mount a sub-application at the given path.
+
+        Overrides `Router.mount` so that the router's own prefix is applied,
+        matching the behaviour of `add_api_route`.  The stored route path will
+        be `self.prefix + path`, which is then further prefixed by any
+        `prefix` argument passed to `include_router` when this router is
+        included into another router or application.
+        """
+        super().mount(self.prefix + path, app=app, name=name)
+
     def websocket(
         self,
         path: Annotated[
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Duplicate of https://github.com/fastapi/fastapi/pull/15233

But I think we should go with https://github.com/fastapi/fastapi/pull/13944

PR #15271, opened 2026-04-01
docs: clarify operation_id caveat for multi-method api_route

wwenrr:
## Summary

This PR adds a documentation note clarifying OpenAPI `operation_id` behavior when using a single endpoint for multiple methods via:

- `api_route()`
- `add_api_route(..., methods=[...])`

## Why

Issue #13175 reports duplicate operation IDs in this setup. The current docs already say operation IDs must be unique, but they don't explicitly call out the multi-method route caveat.

This PR improves discoverability by adding that warning exactly where `operation_id` is documented.

## Changes

File updated:

- `docs/en/docs/advanced/path-operation-advanced-configuration.md`

Added a short `note` block under the OpenAPI `operation_id` section explaining:

- multi-method registration can conflict on generated operation IDs
- users should either:
  - split handlers into dedicated decorators (e.g. `@app.get`, `@app.post`), or
  - set explicit unique `operation_id` values

## Scope

- Docs only (English docs source)
- No runtime changes
- No API behavior changes

## Context

This aligns docs with maintainer guidance in #13175 that multi-method `.api_route()` usage is not the primary FastAPI path-operation style for OpenAPI/client generation workflows.

Closes #13175

### DIFF
--- a/docs/en/docs/advanced/path-operation-advanced-configuration.md
+++ b/docs/en/docs/advanced/path-operation-advanced-configuration.md
@@ -14,6 +14,18 @@ You would have to make sure that it is unique for each operation.
 
 {* ../../docs_src/path_operation_advanced_configuration/tutorial001_py310.py hl[6] *}
 
+/// note
+
+If you register the same endpoint function for multiple HTTP methods with
+`api_route()` / `add_api_route(..., methods=[...])`, OpenAPI operation IDs can
+conflict unless you set explicit unique `operation_id` values.
+
+If you need unique operation IDs in generated OpenAPI clients, prefer separate
+path operation decorators (for example `@app.get()` and `@app.post()`) or set
+an explicit `operation_id` per operation.
+
+///
+
 ### Using the *path operation function* name as the operationId { #using-the-path-operation-function-name-as-the-operationid }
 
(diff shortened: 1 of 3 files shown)


docs/en/docs/advanced/path-operation-advanced-configuration.md
```diff
@@ -14,6 +14,18 @@ You would have to make sure that it is unique for each operation.
 
 {* ../../docs_src/path_operation_advanced_configuration/tutorial001_py310.py hl[6] *}
 
+/// note
+
+If you register the same endpoint function for multiple HTTP methods with
+`api_route()` / `add_api_route(..., methods=[...])`, OpenAPI operation IDs can
+conflict unless you set explicit unique `operation_id` values.
```
YuriiMotov: How would you set "explicit unique `operation_id` values" to avoid operation IDs conflict?
I think that adding routes separately for each method is the only way to solve this for now

wwenrr (commented):
Thanks! Great point — wording updated in the latest commit.

I removed the implication that users can reliably solve this with explicit `operation_id` values when one function is bound to multiple methods.
Now it clearly recommends separate path operation decorators (`@app.get()`, `@app.post()`) so each operation can define its own `operation_id`.

YuriiMotov:
Let's close this for now to keep the list of PRs cleaner.
We will get back to this question later when Sebastian has time to review the issue and take decision.

Thanks!

PR #15272, opened 2026-04-01
Fix deterministic method suffix in generated operation IDs

wwenrr:
## Summary
- make `generate_unique_id()` pick the HTTP method suffix deterministically
- replace `list(route.methods)[0]` (set iteration order) with `sorted(route.methods)[0]`
- add a regression test for routes declared with multiple methods

## Why
`route.methods` is a set, so selecting the first element by list conversion can vary by runtime hash seed. This can produce unstable operation IDs for multi-method routes and increase the chance of surprising schema diffs/warnings.

## Changes
- `fastapi/utils.py`
  - use `sorted(route.methods)[0].lower()` when composing the default unique ID
- `tests/test_generate_unique_id_function.py`
  - add `test_generate_unique_id_with_multiple_methods_is_deterministic`

## Testing
- Added regression test.
- I couldn't run the test suite in this environment because `pytest` is not installed on the runner image.

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -96,7 +96,7 @@ def generate_unique_id(route: "APIRoute") -> str:
     operation_id = f"{route.name}{route.path_format}"
     operation_id = re.sub(r"\W", "_", operation_id)
     assert route.methods
-    operation_id = f"{operation_id}_{list(route.methods)[0].lower()}"
+    operation_id = f"{operation_id}_{sorted(route.methods)[0].lower()}"
     return operation_id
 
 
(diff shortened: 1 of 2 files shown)


wwenrr:
CI fail is only from label check. This PR needs one label from: breaking/security/feature/bug/refactor/upgrade/docs/lang-all/internal. I do not have permission to add labels on upstream from a fork. Could a maintainer add bug (or another matching label) to unblock checks? All test/pre-commit jobs are passing.

wwenrr:
Thanks for the heads-up! ✅

The CI failure is from `check-labels` only:
- required one of: `breaking`, `security`, `feature`, `bug`, `refactor`, `upgrade`, `docs`, `lang-all`, `internal`
- this PR currently has none

I don’t have permission to add labels on `fastapi/fastapi` from my fork side (403).
Could a maintainer please add `bug` (or another matching label)?

All code checks are already green or running.

YuriiMotov:
Please, before opining PRs review existing PRs linked to the issue: https://github.com/fastapi/fastapi/issues/13175

This approach is wrong as operation IDs of both path operations will be the same, but they have to be unique

PR #15273, opened 2026-04-01
fix: preserve mounted sub-applications when including APIRouter

wwenrr:
## Summary
Fixes #10180.

`APIRouter.mount()` currently stores `Mount` routes, but `APIRouter.include_router()` only copies `APIRoute`, `Route`, and websocket routes.
As a result, mounted sub-applications are silently dropped when the router is included.

This PR:
- adds an `APIRouter.mount()` override that applies `self.prefix` (matching `add_api_route` behavior)
- teaches `APIRouter.include_router()` to propagate `routing.Mount` routes
- adds regression tests for mounting both a FastAPI sub-app and a plain ASGI app under nested router prefixes

## Repro before
A mounted route like `/api/subapi/sub` returns 404 after `app.include_router(api_router)`.

## Behavior after
Mounted sub-app routes are reachable under the expected combined prefix path.

## Tests
Added `tests/test_router_mount.py` with coverage for:
- `APIRouter(prefix=...)` + `router.mount(...)` + `app.include_router(router)`
- nested include prefix handling (`app.include_router(router, prefix=...)`)

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1575,6 +1575,9 @@ class APIRouter(routing.Router):
 
         return decorator
 
+    def mount(self, path: str, app: ASGIApp, name: str | None = None) -> None:
+        super().mount(self.prefix + path, app=app, name=name)
+
     def include_router(
         self,
         router: Annotated["APIRouter", Doc("The `APIRouter` to include.")],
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Please, review existing PRs before opening new PR.
This one is a duplicate of https://github.com/fastapi/fastapi/pull/14793

PR #15292, opened 2026-04-02
docs: mention how to change SSE keep-alive ping interval

JimVincentW:
The technical details section mentions the 15s keep-alive ping but not how to change it. `fastapi.sse._PING_INTERVAL` can be set to configure this globally. Thus wanted to include a small addition and snippet.


YuriiMotov:
Thanks for your interest, but we shouldn't recommend doing it this way (it's private constant).

We will consider adding the official way to configure this

chero858:
@YuriiMotov +1 for making those constants configurable.
https://github.com/fastapi/fastapi/discussions/15291

PR #15295, opened 2026-04-03
Add AuthStaticFiles to support authentication for static file serving

### LINKED ISSUE #858: How to authenticate static files?
### First check
* [x]  I used the GitHub search to find a similar issue and didn't find it.
* [x]  I searched the FastAPI documentation, with the integrated search.
* [x]  I already searched in Google "How to X in FastAPI" and didn't find any information.

### Description
How can I route `/foo` to `fe/private.html` for authorised users and `fe/public.html` for everyone else, while ensuring that the `<script src="fe/app.js"></script>` references inside those 2 HTML files are resolved?

I have already implemented Basic HTTP Authentication using `fastapi.security.HTTPBasic`, and loading of static files using `starlette.responses.HTMLResponse`.

Here is the snippet for loading the appropriate HTML file based on Basic HTTP Authentication username:
```python
@app.get("/foo")
def get_private_content(username: str = Depends(get_current_username)):
    with open('fe/private.html' if username == 'goodperson' else 'fe/public.html') as f:
        return HTMLResponse(f.read())
```

Indeed, the above code loads the correct HTML file. But, even though the HTML file is correctly loaded, the `fe/app.js` script reference inside the HTML file returns 404:
```html
<script src="fe/app.js"></script>
```

I am aware that I can use `app.mount` to expose the entire `/fe` folder, as discussed in #130, but that would defeat the authentication because everyone can then access `fe/private.html` directly.

In the context of fastapi, am I doing authentication of static files correctly?

Faisalsaificode:
Adds a new AuthStaticFiles class that extends StaticFiles with an `auth` parameter, allowing users to protect static files behind authentication. This addresses a long-standing feature request (issue #858) where mounted static files bypass FastAPI's dependency injection and serve files without any access control.

Closes #858


Pawansingh3889 (commented):
This addresses a real pain point — protecting mounted static files has been a common request since #858.

A few thoughts on the implementation:

1. **JSON error responses for static files**: The auth failure returns a JSON response, but static file consumers (browsers, img tags, etc.) typically expect HTML or a redirect to a login page. Consider making the error response format configurable, or at minimum returning a plain 401 with the `WWW-Authenticate` header.

2. **Performance**: The auth callable runs on every static file request (CSS, JS, images). For token validation that hits a database, this could be expensive. Worth documenting that lightweight auth checks (header presence, JWT signature verification) are preferred over DB lookups here.

3. **Test coverage**: The test suite is thorough — covers unauthenticated, invalid creds, valid access, public fallback, custom headers, and cookie-based auth. Good coverage.

4. **Naming**: `AuthStaticFiles` is clear and self-documenting. The `auth` parameter accepting an async callable is a clean API design.

tiangolo:
Thanks for the interest, I think this would be application specific, so I'll pass on this one. Thanks! :coffee:

Faisalsaificode:
@tiangolo Thanks for reviewing! Are there specific areas where you'd welcome contributions from someone new to the codebase? Want to make sure my next attempt aligns with the project's direction.

PR #15300, opened 2026-04-04
Preserve Annotated dependency metadata for stringified forward refs

itsviseph:
## Summary
This preserves `Annotated[..., Depends(...)]` metadata for stringified annotations so dependency parameters are not misclassified as query params when the inner type is still a forward reference.

## What changed
- evaluate string annotations as full expressions instead of collapsing them into a single unresolved forward reference
- fall back to a namespace that substitutes missing symbols with `ForwardRef(...)` so the outer typing structure is preserved
- add regressions for:
  - a late-defined dependency type under `from __future__ import annotations`
  - an explicitly stringified inner type inside `Annotated[..., Depends(...)]`

## Validation
- `python -m pytest tests/test_stringified_annotations_simple.py tests/test_stringified_annotation_dependency.py tests/test_stringified_annotations_forwardrefs.py -q`

Fixes #13056

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -81,6 +81,13 @@ multipart_not_installed_error = (
     'You can install "python-multipart" with: \n\n'
     "pip install python-multipart\n"
 )
+
+
+class _ForwardRefNamespace(dict[str, Any]):
+    def __missing__(self, key: str) -> Any:
+        value = ForwardRef(key)
+        self[key] = value
+        return value
 multipart_incorrect_install_error = (
     'Form data requires "python-multipart" to be installed. '
     'It seems you installed "multipart" instead. \n'
@@ -244,8 +251,13 @@ def get_typed_signature(call: Callable[..., Any]) -> inspect.Signature:
 
 def get_typed_annotation(annotation: Any, globalns: dict[str, Any]) -> Any:
     if isinstance(annotation, str):
-        annotation = ForwardRef(annotation)
-        annotation = evaluate_forwardref(annotation, globalns, globalns)  # ty: ignore[deprecated]
+        try:
+            annotation = eval(annotation, globalns, globalns)
+        except NameError:
+            # Preserve the outer typing structure (e.g. Annotated[..., Depends(...)])
+            # even when some inner symbols are not defined yet.
+            forwardref_globalns = _ForwardRefNamespace(globalns)
+            annotation = eval(annotation, forwardref_globalns, forwardref_globalns)
         if annotation is type(None):
             return None
     return annotation
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -1,3 +1,4 @@
+import builtins
 import dataclasses
 import inspect
 import sys
@@ -84,6 +85,8 @@ multipart_not_installed_error = (
 
 class _ForwardRefNamespace(dict[str, Any]):
     def __missing__(self, key: str) -> Any:
+        if hasattr(builtins, key):
+            return getattr(builtins, key)
         value = ForwardRef(key)
         self[key] = value
         return value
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@itsviseph, have you reviewed other PRs linked to the issue?
Please review them and explain why you think your approach is better than others.
Thanks!

itsviseph:
Thanks, I reviewed the other PRs linked to #13056.

My understanding of the difference is:

- #14703 resolves annotations in `get_dependant()` with `get_type_hints(..., include_extras=True)`. That helps only when Python can fully resolve the annotation at that point. For the late-defined / still-unresolved cases from this issue, it is too late in the flow and still depends on full resolution succeeding. That also seems consistent with the existing review there that the new repro still failed.

- #14557 goes in the other direction and adds a custom parser / fallback for `Annotated[...]` strings plus module re-import logic. It can recover the `Depends(...)` metadata, but it does so by special-casing `Annotated` strings and, in some cases, falling back to `Annotated[Any, ...]`.

What this PR tries to do instead is keep the fix in `get_typed_annotation()` itself, at the point where the stringified annotation is first converted. Rather than manually parsing `Annotated[...]`, it evaluates the full annotation expression and only turns the still-missing inner names into `ForwardRef`s. In the problematic case that means we preserve:

`Annotated[ForwardRef("Potato"), Depends(get_potato)]`

instead of collapsing the whole thing into one unresolved `ForwardRef("Annotated[...]")` or replacing the inner type with `Any`.

So the reason I think this approach is better is:

1. it is narrower in scope: only `get_typed_annotation()` changes
2. it preserves the full outer typing structure instead of re-parsing strings manually
3. it does not need `Annotated`-specific string parsing or module import tricks
4. it keeps the unresolved inner type as a real `ForwardRef`, which seems closer to the intended typing semantics than degrading to `Any`

I added the second regression test specifically to cover both cases:
- a late-defined symbol (`Annotated[Potato, Depends(...)]`)
- an explicitly stringified inner type (`Annotated["Potato", Depends(...)]`)

If you want, I can also add coverage around non-`Depends` `Annotated` metadata to make the behavioral boundary more explicit.

itsviseph:
Follow-up after reviewing the additional linked PRs and the discussion on #15234:

I should have mentioned #15234 earlier — that was an oversight on my side. It is the closest prior work to this PR.

After going through the linked set more carefully, my current understanding is:

- #14952 only adds the regression test, which is useful for locking the repro down but does not attempt a fix.
- #14703 tries to resolve later in `get_dependant()` via `get_type_hints(..., include_extras=True)`, but that still depends on full resolution succeeding at that later stage.
- #14872 / #15142 try to resolve `ForwardRef` later in the parameter-analysis path.
- #15131 addresses the later OpenAPI / TypeAdapter failure in the compat layer, which helps with schema generation but is broader and later than the dependency-extraction point.
- #14557 / #15096 / #15234 are the closest family to this PR because they all try to preserve the outer `Annotated[...]` structure instead of letting the whole annotation collapse into one unresolved forward ref.

The most important thing I took from #15234 is the review comment there: keep the normal strict resolution path first, and only use the lenient namespace fallback when strict resolution still leaves an unresolved `ForwardRef`. That seems like the right shape.

So the honest comparison is not that this PR invented a completely different strategy; it is that it converged on the same core idea as #15234 and related work:
- preserve `Annotated[..., Depends(...)]`
- keep unresolved inner names as forward refs instead of losing the outer structure
- fix the bug at the annotation-conversion stage rather than later in OpenAPI generation

Where I still think this direction is the right one is that it targets the earliest shared point in the flow, so both runtime dependency extraction and OpenAPI behavior improve from the same root cause.

If it would be more helpful, I’m happy to revise this PR further so it matches the stricter fallback-only shape discussed on #15234 even more closely.

Pawansingh3889 (commented):
Clean approach to a tricky problem. The `_ForwardRefNamespace` dict subclass that creates `ForwardRef` objects on missing key lookups is elegant — it preserves the outer `Annotated[..., Depends(...)]` structure while deferring inner type resolution.

A few observations:

1. **Security of eval**: The two-stage eval approach (`eval(annotation, globalns)` then fallback with `_ForwardRefNamespace`) is safe since `annotation` comes from Python's own `__annotations__`, not user input. Worth a brief comment in the code noting this for future readers.

2. **Test coverage**: The two test cases cover the key scenarios well — late-defined forward refs and explicit string annotations. Consider adding a test for nested forward refs (e.g., `"Annotated[list[SomeModel], Depends(...)]"`) to ensure the namespace handles recursive lookups.

3. **Deprecation cleanup**: Good call removing the `evaluate_forwardref` import — it was marked deprecated and this PR replaces it with a more robust mechanism.

Solid fix. LGTM.

tiangolo:

I want to implement a few big refactors and features that might solve many of these use cases and simplify how to use security.

Thank you! 🍰

PR #15305, opened 2026-04-06
Fix get_path_param_names to strip path converters

Bahtya:
## Summary

Fixes #14883

`get_path_param_names()` uses regex `{(.*?)}` which captures `param:path` instead of `param` when OpenAPI path converters are used. This causes implicit path params to be incorrectly classified as query params.

## Fix

Change regex from `{(.*?)}` to `{([^:}]+)}` to stop at the colon separator.

## Example

```python
# Before: get_path_param_names("/files/{file_path:path}") → {"file_path:path"}
# After:  get_path_param_names("/files/{file_path:path}") → {"file_path"}
```

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -41,7 +41,7 @@ def is_body_allowed_for_status_code(status_code: int | str | None) -> bool:
 
 
 def get_path_param_names(path: str) -> set[str]:
-    return set(re.findall("{(.*?)}", path))
+    return set(re.findall("{([^:}]+)", path))
 
 
 _invalid_args_message = (


fastapi/utils.py
```diff
@@ -41,7 +41,7 @@ def is_body_allowed_for_status_code(status_code: int | str | None) -> bool:
 
 
 def get_path_param_names(path: str) -> set[str]:
-    return set(re.findall("{(.*?)}", path))
+    return set(re.findall("{([^:}]+)", path))
```
YuriiMotov, line 44: This way `{name:int` will also match, but it's actually invalid (no close curly bracket).

YuriiMotov:
@Bahtya, thanks for your interest!

We already have https://github.com/fastapi/fastapi/pull/14883 . Please, suggest changes to existing PR instead of opening duplicate PRs for the issue

Bahtya:
Good catch, @YuriiMotov! You're right — the previous regex `{([^:}]+)` would also match incomplete path segments like `{name:int` without a closing brace.

I've updated the regex to `{([^:}]+)(?::[^}]+)?}` which:
1. Captures the parameter name (characters before `:` or `}`)
2. Optionally matches the converter part `:[^}]+` 
3. **Requires** the closing `}`

This correctly matches `{name}`, `{item_id:int}` but **not** `{name:int` (incomplete). Thanks for the review!

PR #15315, opened 2026-04-09
Fix/client disconnect as parsing error

AlePiccin:
## Summary

When a client disconnects while FastAPI is reading the request body, Starlette raises `ClientDisconnect`. The broad `except Exception` handler in `fastapi/routing.py` catches it and converts it into `HTTPException(400, "There was an error parsing the body")`, which is misleading — a client disconnect is not a body parsing error.

This adds `except ClientDisconnect: raise` before the catch-all so the exception propagates correctly instead of being misclassified.

Closes #15312

## The Problem

In `fastapi/routing.py`, the body reading try/except block handles:
- `json.JSONDecodeError` → 422 (correct)
- `HTTPException` → re-raised (correct)
- `Exception` (catch-all) → **HTTP 400 "There was an error parsing the body"**

Since `ClientDisconnect` inherits from `Exception`, it gets caught by the catch-all and misreported as a parsing error. This causes:
- False HTTP 400 responses even when the request body is perfectly valid
- Polluted error monitoring with fake parsing errors
- Difficult production debugging (we spent hours tracing this)

## Real-World Scenario

Running FastAPI behind Gunicorn with `UvicornWorker` and `--max-requests` for worker recycling. When a worker is recycled mid-request, `receive()` returns `http.disconnect`, Starlette raises `ClientDisconnect`, and the client receives a false 400 parsing error.

## Research

Investigated how other major frameworks handle this:

| Framework | Behavior |
|---|---|
| **Django (ASGI)** | Raises `RequestAborted`, silently returns with no response |
| **Go net/http** | Cancels request context, stops processing |
| **Spring Boot** | `ClientAbortException` suppressed, logged at DEBUG |
| **Nginx** | Logs non-standard 499, no response sent |
| **<ASP.NET> Core** | `OperationCanceledException`, no default handler |

**Consensus**: Client disconnect is not a client error (400) nor a server error (500). It's an operational condition that should propagate, not be masked as a parsing error.

## Changes

- **`fastapi/routing.py`**: Added `ClientDisconnect` import and `except ClientDisconnect: raise` before the catch-all (2 lines)
- **`tests/test_client_disconnect.py`**: Regression tests

## Test Plan

- [x] `test_client_disconnect_json_body` — verifies `ClientDisconnect` during JSON body reading is not converted to HTTP 400
- [x] `test_client_disconnect_form_body` — same for form data
- [x] `test_body_parse_error_still_returns_400` — ensures genuine parse errors still return the correct status

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -78,7 +78,7 @@ from starlette._utils import is_async_callable
 from starlette.concurrency import iterate_in_threadpool, run_in_threadpool
 from starlette.datastructures import FormData
 from starlette.exceptions import HTTPException
-from starlette.requests import Request
+from starlette.requests import ClientDisconnect, Request
 from starlette.responses import JSONResponse, Response, StreamingResponse
 from starlette.routing import (
     BaseRoute,
@@ -441,6 +441,8 @@ def get_request_handler(
             raise validation_error from e
         except HTTPException:
             # If a middleware raises an HTTPException, it should be raised again
+            raise
+        except ClientDisconnect:
             raise
         except Exception as e:
             http_error = HTTPException(
--- a/tests/test_client_disconnect.py
+++ b/tests/test_client_disconnect.py
@@ -1,0 +1,113 @@
+"""
+Test that ClientDisconnect during request body reading is not
+misreported as HTTP 400 "There was an error parsing the body".
+
+Ref: https://github.com/fastapi/fastapi/issues/XXXXX
+"""
+
+import pytest
+from starlette.requests import ClientDisconnect
+
+from fastapi import Body, FastAPI, Form
+from fastapi.testclient import TestClient
+
+pytestmark = pytest.mark.anyio
+
+app = FastAPI()
+
+
+@app.post("/json")
+async def json_endpoint(data: dict = Body(...)):
+    return data
+
+
+@app.post("/form")
+async def form_endpoint(field: str = Form(...)):
+    return {"field": field}
+
+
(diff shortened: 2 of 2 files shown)


YuriiMotov:
But this way client disconnects will be treated as server errors. That is also incorrect.
You said it was difficult to debug it, but will it be less difficult to debug with this change?
`HTTPException` is raised from the original exception (`raise http_error from e`), so it should appear in logs (I haven't checked it).

> False HTTP 400 responses even when the request body is perfectly valid

I think if client disconnects while sending the request, the request body is not valid (incomplete) and status code `400` is more correct than `500`.
And, when the request is interrupted by Gunicorn, for the app it's the same - request received by app is incomplete.

> **Consensus:** Client disconnect is not a client error (400) nor a server error (500). It's an operational condition that should propagate, not be masked as a parsing error.

I think I agree with this, but this should be first solved on Starlette's side. For now Starlette's `ServerErrorMiddleware` will handle this as server errors.
What do you think?

avinashkamat48 (commented):
The added test has two small consistency issues that will make it harder to understand later: the file still references `https://github.com/fastapi/fastapi/issues/XXXXX`, and `test_body_parse_error_still_returns_400` asserts `422` while the docstring says generic body parse errors must still produce `400`. Please replace the placeholder issue number and align the test name/docstring with the asserted behavior, or change the assertion if `400` is actually the intended contract.

tiangolo:
I'll pass on this for now, thanks! :coffee:

PR #15261, opened 2026-03-30
🔒 Pin GitHub actions by commit SHA

YuriiMotov:
Closes: https://github.com/fastapi/fastapi/pull/15225

I created this PR from scratch and then compared it with [15225](https://github.com/fastapi/fastapi/pull/15225) to double-check (there were several differences in hashes (see full diff in the details) - commented on them in the comments of [15225](https://github.com/fastapi/fastapi/pull/15225))

<details><summary>Full diff with #15225</summary>

```diff
diff --git a/.github/workflows/build-docs.yml b/.github/workflows/build-docs.yml
index 0055f4128..a4da37a05 100644
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -18,9 +18,9 @@ jobs:
     outputs:
       docs: ${{ steps.filter.outputs.docs }}
     steps:
-    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
     # For pull requests it's not necessary to checkout the code but for the main branch it is
-    - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+    - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4
       id: filter
       with:
         filters: |
@@ -42,13 +42,13 @@ jobs:
     outputs:
       langs: ${{ steps.show-langs.outputs.langs }}
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
       - name: Set up Python
-        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
         with:
           python-version-file: ".python-version"
       - name: Setup uv
-        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+        uses: astral-sh/setup-uv@94527f2e458b27549849d47d273a16bec83a01e9 # v7
         with:
           enable-cache: true
           cache-dependency-glob: |
[...]

PR #15316, opened 2026-04-10
🔒️ Add zizmor and fix audit findings

YuriiMotov:
This PR includes changes from https://github.com/fastapi/fastapi/pull/15261 . So, if this PR gets merged, #15261 may be closed.

Changes applied:
* Pin actions by SHA
* Setup cooldown period for Dependabot (7 days)
* Ignored `dangerous-triggers` rule for `pull_request_target` and `workflow_run` (checked that they are used in a safe way)
* Specified minimal permissions on workflow level, moved permissions to the job level
* Ignored `secrets-outside-env` rule as using the environments would require approval for each run (and without required approvals it wouldn't make sense)
* Added `persist-credentials: false` for `actions/checkout` when persisting is not needed by other steps
* Specified version of `uv` to install for `astral-sh/setup-uv` (Note that Dependabot will not upgrade it, but Renovate can do it)
* Ignored `template-injection` rule when input comes from trusted source (`workflow_dispatch` or `schedule`)
* Fixed `template-injection` in other cases
* Disabled cache for deploy-docs workflow (see https://docs.zizmor.sh/audits/#cache-poisoning). Not completely sure we need this as `uv` has internal defense mechanism against cache poisoning
* Specified run condition in `latest-changes` to make it clear that it only runs for merged PRs
* Replaced `uvx prek` command with `uv run prek` - `uvx` uses latest version (unpinned), it's better to use locked version
* Added branch filter in `notify-translations` - it's recommended practice for `pull_request_target` triggers and in this case it shouldn't break anything
* Added zizmor pre-commit hook (there are [multiple options](https://docs.zizmor.sh/integrations/), but I think this way is more convenient) - tested by introducing issue ([see](https://github.com/fastapi/fastapi/actions/runs/24251156923/job/70810592011#step:8:30))


tiangolo (approved):
Excellent, looks great! 🚀  Thanks! 🙌

rsb-23:
Few suggestions to reduce tech debt:
1. We can use `.github/zizmor.yml` for config.
2. Can use reusable actions for checkout, python and uv. (DRY it) 

This makes handling new zizmor checks simpler. 

If you prefer, I can make the changes and draft a PR.

PR #15320, opened 2026-04-10
Fix/my first contribution

abishekgiri:
Summary

Fix duplicate OpenAPI `operationId` generation for multi-method routes.

Previously, routes registered with multiple HTTP methods could end up with conflicting default `operationId` values. This change keeps explicit `operation_id` values untouched, preserves the existing behavior for single-method routes, and makes multi-method routes generate a shared base ID that is suffixed per HTTP method when building OpenAPI operations.

Changes

- updated default unique ID handling for multi-method routes
- kept explicit `operation_id` unchanged
- added regression tests for:
  - default `generate_unique_id_function`
  - custom `generate_unique_id_function`

Verification

Tested with:

- `uv run pytest tests/test_generate_unique_id_function.py`
- `uv run pytest tests/test_extra_routes.py`

Related

Fixes #13175

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -233,6 +233,13 @@ def generate_operation_summary(*, route: routing.APIRoute, method: str) -> str:
     return route.name.replace("_", " ").title()
 
 
+def get_openapi_operation_id(*, route: routing.APIRoute, method: str) -> str:
+    operation_id = route.operation_id or route.unique_id
+    if route.operation_id is None and route.methods and len(route.methods) > 1:
+        operation_id = f"{operation_id}_{method.lower()}"
+    return operation_id
+
+
 def get_openapi_operation_metadata(
     *, route: routing.APIRoute, method: str, operation_ids: set[str]
 ) -> dict[str, Any]:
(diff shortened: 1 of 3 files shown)


abishekgiri:
I reviewed the related PRs for #13175: #13861, #14881, #15070, #15074, #15140, #15194, and #15215.

My reasoning for this approach is:

- #13861, #14881, and #15070 make the selected method deterministic, but they do not solve the duplicate `operationId` problem itself.
- #15074 appends the current method to `route.unique_id`, which already contains a method suffix in the default generator, so it can produce IDs like `..._post_post`.
- #15140 and #15194 are closer, but they still have to reverse-engineer a base ID from `route.unique_id`. That is brittle for custom `generate_unique_id_function` values, especially #15140, which strips everything after the last underscore.
- #15215 rebuilds the base from `route.name + route.path_format` inside OpenAPI, which bypasses `generate_unique_id_function` for multi-method routes.

I chose this implementation because it fixes the root ambiguity instead of patching around it later:

- for multi-method routes, the default `unique_id` becomes a shared base
- the per-method suffix is added only when emitting OpenAPI
- explicit `operation_id` remains unchanged
- custom `generate_unique_id_function` is still respected, because OpenAPI starts from `route.unique_id` instead of reconstructing it

I think that is cleaner because `route.unique_id` is reused internally beyond `operationId`, so removing the arbitrary “first method” from the canonical route ID is more robust than preserving it and then trying to peel it back off in OpenAPI generation.

I also added regression coverage for both the default generator and a custom `generate_unique_id_function`, because that was the main gap in the other proposals.

YuriiMotov:
@abishekgiri, why do you open duplicate PRs?

This is a duplicate of https://github.com/fastapi/fastapi/pull/15222 that was closed with the explanation: https://github.com/fastapi/fastapi/pull/15222#issuecomment-4160505792

PR #15327, opened 2026-04-12
Fix duplicate operation IDs with multi-method routes

mokshagnachintha:
Fixes #13175

## Problem
Routes with multiple HTTP methods generate duplicate operation IDs in OpenAPI schema.

## Solution
Modified operation ID generation to include HTTP method as suffix for uniqueness.

## Changes
- fastapi/utils.py: Updated generate_unique_id() to accept method parameter
- fastapi/openapi/utils.py: Updated to pass method when generating operation IDs
- tests/test_duplicate_operation_id.py: Comprehensive test coverage


YuriiMotov:
@mokshagnachintha, thanks for the attempt, but this is wrong and we already have too many PRs linked to this issue

PR #15331, opened 2026-04-13
Add example for handling low_confidence ML predictions in error handling docs

Nithin00614:
This PR adds a practical example demonstrating how to handle low-confidence predictions in ML-based APIs using HTTPException.

It shows how to prevent unreliable predictions from being returned and instead provide structured error responses, supporting real-world reliability and human-in-the-loop workflows.


YuriiMotov:
@Nithin00614, thanks for your interest, but I think we don't need this

Nithin00614:
Thanks for the feedback! Understood 👍

I appreciate the clarification on scope. I’ll keep this in mind for future contributions.

PR #15336, opened 2026-04-14
Reduce memory usage by `Dependant` by ~50%

ipeluffo:
## Context

In https://github.com/fastapi/fastapi/discussions/14742, I reported a significant spike in memory usage when upgrading FastAPI from v0.120.4 to v0.121.x (and later versions).

My analysis traced the regression to [this refactor](https://github.com/fastapi/fastapi/pull/14262/files#diff-4d1d65ce33d5ef7383c87f7edeb3928bc0537016daba6d1a6b8b27a7af8f8aaaR46-R52), which replaced computed fields in `__post_init__` with [`functools.cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property).

## Root cause

The Python docs for [`functools.cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) note:

> This decorator interferes with the operation of [[PEP 412](https://peps.python.org/pep-0412/)](https://peps.python.org/pep-0412/) key-sharing dictionaries. This means that instance dictionaries can take more space than usual.

While the worst of this was addressed in Python 3.12+ https://github.com/python/cpython/issues/101815, there is still an overhead: accessing `__dict__` on an instance forces creation of a full dictionary object rather than a compact `PyDictValues`, adding a `PyDictObject` (three fields) per instance. At scale, this adds up.

The exact quote from the cpython issue:
> It remains true that cached_property can cause instances to use slightly more memory, but the reason for this has changed (and the extra memory used will be significantly less.) The new reason is that any access of __dict__ on an instance will force creation of a real dictionary object for that instance, rather than just a PyDictValues. The keys remain shared in the created dict, though. So the added memory use is no longer "all the keys" but rather just "a PyDictObject" (which consists of only three fields.)

## Benchmarks

I wrote a script to measure memory usage across varying numbers of `Dependant` instances:

<details><summary>Benchmark script</summary>

```python
import gc
import tracemalloc
from fastapi.dependencies.models import Dependant

samples_tests = (
    200_000,
    100_000,
    50_000,
    1_000,
    500,
)

def measure(samples: int):
    gc.collect()
    tracemalloc.start()
    objs = [Dependant() for i in range(samples)]

    # Trigger all cached properties
    for o in objs:
        o.oauth_scopes
        o.cache_key
        o._uses_scopes
        o._is_security_scheme
        # o._security_scheme  # skipped: asserts a SecurityBase, not None
        o._security_dependencies
        o.is_gen_callable
        o.is_async_gen_callable
        o.is_coroutine_callable
        o.computed_scope

    current, peak = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    return current, peak, objs  # keep objs alive until after measurement

for samples in samples_tests:
    cur, peak, _ = measure(samples)
    print(f"Samples: {samples} - current={cur/1e6:.1f} MB  peak={peak/1e6:.1f} MB")
```

</details>

**Before (current `master`, Python 3.13.12):**

```
Samples: 200000 - current=318.4 MB  peak=318.4 MB
Samples: 100000 - current=159.2 MB  peak=159.2 MB
Samples: 50000 - current=79.6 MB  peak=79.6 MB
Samples: 1000 - current=1.6 MB  peak=1.6 MB
Samples: 500 - current=0.8 MB  peak=0.8 MB
```

**After (this PR):**

```
Samples: 200000 - current=156.8 MB  peak=156.8 MB
Samples: 100000 - current=78.4 MB  peak=78.4 MB
Samples: 50000 - current=39.2 MB  peak=39.2 MB
Samples: 1000 - current=0.8 MB  peak=0.8 MB
Samples: 500 - current=0.4 MB  peak=0.4 MB
```

Memory usage is reduced by approximately **50%** across all cases.

## Approach

The key enabler for the memory reduction is the use of `__slots__`. However, `__slots__` is not compatible with `functools.cached_property`, so the caching strategy had to change.

I considered reverting to the previous `__post_init__` approach but decided against it for two reasons:

1. **No lazy initialization** — all computed fields would be evaluated eagerly on every `Dependant` instantiation, even if never accessed.
2. **Awkward dependency ordering** — several computed fields depend on each other, making `__post_init__` harder to maintain.

Instead, this PR introduces a different caching mechanism that preserves lazy evaluation while being compatible with `__slots__`.

### DIFF
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -119,9 +119,7 @@ class Dependant:
             _impartial(dunder_call)
         ) or inspect.isgeneratorfunction(_unwrapped_call(dunder_call)):
             return True
-        dunder_unwrapped_call = getattr(
-            _unwrapped_call(self.call), "__call__", None
-        )  # noqa: B004
+        dunder_unwrapped_call = getattr(_unwrapped_call(self.call), "__call__", None)  # noqa: B004
         if dunder_unwrapped_call is None:
             return False  # pragma: no cover
         if inspect.isgeneratorfunction(
@@ -147,9 +145,7 @@ class Dependant:
             _impartial(dunder_call)
         ) or inspect.isasyncgenfunction(_unwrapped_call(dunder_call)):
             return True
-        dunder_unwrapped_call = getattr(
-            _unwrapped_call(self.call), "__call__", None
-        )  # noqa: B004
+        dunder_unwrapped_call = getattr(_unwrapped_call(self.call), "__call__", None)  # noqa: B004
         if dunder_unwrapped_call is None:
             return False  # pragma: no cover
         if inspect.isasyncgenfunction(
@@ -179,9 +175,7 @@ class Dependant:
             _unwrapped_call(dunder_call)
         ):
             return True
-        dunder_unwrapped_call = getattr(
-            _unwrapped_call(self.call), "__call__", None
-        )  # noqa: B004
+        dunder_unwrapped_call = getattr(_unwrapped_call(self.call), "__call__", None)  # noqa: B004
         if dunder_unwrapped_call is None:
             return False  # pragma: no cover
         if iscoroutinefunction(
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -49,15 +49,20 @@ class Dependant:
     use_cache: bool = True
     path: str | None = None
     scope: Literal["function", "request"] | None = None
+    # Lazy cached fields
+    _oauth_scopes_cache: list[str] = field(default=None, init=False, repr=False)
 
-    @cached_property
+    @property
     def oauth_scopes(self) -> list[str]:
-        scopes = self.parent_oauth_scopes.copy() if self.parent_oauth_scopes else []
-        # This doesn't use a set to preserve order, just in case
-        for scope in self.own_oauth_scopes or []:
-            if scope not in scopes:
-                scopes.append(scope)
-        return scopes
+        if self._oauth_scopes_cache is None:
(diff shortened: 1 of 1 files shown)


fastapi/dependencies/models.py
```diff
@@ -18,17 +18,17 @@
 def _unwrapped_call(call: Callable[..., Any] | None) -> Any:
     if call is None:
         return call  # pragma: no cover
-    unwrapped = inspect.unwrap(_impartial(call))
+    unwrapped = inspect.unwrap(cast(Callable[..., Any], _impartial(call)))
     return unwrapped
 
 
-def _impartial(func: Callable[..., Any]) -> Callable[..., Any]:
+def _impartial(func: Callable[..., Any] | None) -> Callable[..., Any] | None:
```
ipeluffo, line 25: [info] This is unrelated to my PR but `mypy` was complaining in the `pre-commit` check. The change looks safe and sensible

gryevns:
@YuriiMotov is there anything needed to move this one forward and get some feedback/review? Thanks

ipeluffo:
@tiangolo , I'd be great if we can some feedback on this before the PR gets lost. Thanks

gryevns:
@tiangolo could you please advise how to move this PR forwards, it's becoming more and more important as we fall behind FastAPI updates. Thanks!

tiangolo:
Thank you for the investigation in the original discussion and the work on this PR! :raised_hands: 

I used it plus some additional work and reports to base the work in this one https://github.com/fastapi/fastapi/pull/16049

It is now released in FastAPI 0.140.0. :rocket: 

I'll now close this one. Thank you!

PR #15341, opened 2026-04-15
fix(params): honor AliasChoices validation aliases for direct request params

OussemaNehdi:
## Summary
This PR fixes direct request parameter parsing so validation_alias values declared with AliasChoices are respected for Query, Header, Cookie, Form, and File parameters.

## Reproduction
When a parameter is declared with validation_alias=AliasChoices("p_val_alias", "p_fallback_alias"), requests using the fallback alias returned 422 instead of being accepted.

## Root Cause
Direct parameter extraction handled validation aliases as a single key, so only one alias was effectively checked during lookup.

## Changes
- Added internal support to resolve all declared validation aliases in stable order.
- Updated request parameter extraction to try aliases in order and use the first present value.
- Updated body/form extraction paths to accept declared validation alias keys.
- Preserved canonical alias behavior for OpenAPI parameter naming and error locations.

## Tests
- Added regression tests for AliasChoices validation_alias behavior in query parameters.
- Verified missing input returns 422.
- Verified parameter name usage returns 422.
- Verified primary alias usage returns 200.
- Verified fallback alias usage returns 200.
- Full request params test suite passes.
- Full project suite passes locally: 3145 passed, 4 xfailed.

## Compatibility
No breaking API change.
This only fixes missing support for already-declared alias choices in direct request parameter parsing.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -876,7 +876,9 @@ def request_params_to_args(
                 alias = get_validation_alias(field)
                 if alias == field.name:
                     alias = alias.replace("_", "-")
-        aliases_to_process = (alias,) if alias is not None else get_validation_aliases(field)
+        aliases_to_process = (
+            (alias,) if alias is not None else get_validation_aliases(field)
+        )
         value = _get_multidict_value(field, received_params, alias=aliases_to_process)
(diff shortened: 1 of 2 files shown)


tiangolo:

Thank you! 🍰

PR #15355, opened 2026-04-15
Add example for validating path parameters in error handling docs

herbertmoses:
## Description

Added a practical example demonstrating validation for path parameters using HTTPException.

## Changes

- Added example to validate that user_id is a positive integer
- Included explanation for handling invalid inputs gracefully

## Why

This helps developers understand how to implement custom validation logic beyond basic type checking.

## Testing

- Verified documentation locally using 'mkdocs serve'


YuriiMotov:
Thanks, but we don't need this.
And, this is not the best way to implement parameter validation. Better use Pydantic's `AfterValidator` similarly to how it's shown for Query parameters: https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#understand-that-code

PR #15356, opened 2026-04-15
Automatically support HEAD method for all GET routes

### LINKED ISSUE #1773: Automatically support HEAD method for all GET routes, as Starlette does

Herrtian:
Closes #1773

GET routes now respond to HEAD requests automatically, matching what Starlette does (RFC 7231 §4.3.2).

HEAD is **not** added to `self.methods`, so it won't show up in the OpenAPI schema. Instead, `APIRoute.matches()` promotes HEAD→GET at routing time, and `handle()` lets it through.

If you explicitly define a `@app.head()` route, it still works and appears in the schema as before.

Includes 10 new tests covering the various cases (path params, response models, routers, explicit HEAD, etc). Updated one existing test that relied on the old 405 behavior.


Herrtian:
Added the missing handler calls so the new test file is fully covered. uv run pytest tests/test_auto_head_for_get.py passes locally.

Herrtian:
Updated this so explicit HEAD routes still take precedence over automatic GET-to-HEAD handling. The related route tests and ruff checks pass locally.

avinashkamat48 (commented):
If HEAD is now implicitly supported for every GET route, 405 responses should advertise that in the `Allow` header too. With the current code the method check still builds `Allow` from `methods`, so a request like `POST /some-get-route` will report only `GET` even though `HEAD` is accepted by the same route. Please add a regression test around the 405 `Allow` header and include `HEAD` when the effective route has implicit HEAD support.

tiangolo:
There are several things I need to tackle first, and that will most probably affect how this works, so, for now, I'll pass on this. Thanks! :coffee:

PR #15357, opened 2026-04-15
Merge injected BackgroundTasks with Response.background

### LINKED ISSUE #11215: Potential footgun when using custom `Response(background=<Task()>)` in conjunction with injected `BackgroundTasks` - the custom response overwrites the tasks

Herrtian:
## Summary

Closes #11215

When a route handler uses both the injected `BackgroundTasks` dependency **and** returns a `Response(background=BackgroundTask(...))`, the injected tasks were silently discarded. This was a known footgun [reported in 2024](https://github.com/fastapi/fastapi/issues/11215).

**Before:** Only the Response's background task runs; injected tasks are lost.
**After:** Both sets of tasks are merged — injected tasks run first, then the Response's tasks.

## Change

`fastapi/routing.py` lines 679-691: Instead of the simple conditional:

```python
if raw_response.background is None:
    raw_response.background = solved_result.background_tasks
```

Now merges both sources when both exist:

```python
elif solved_result.background_tasks is not None:
    merged = BackgroundTasks(solved_result.background_tasks.tasks)
    if isinstance(original, BackgroundTasks):
        merged.tasks.extend(original.tasks)
    else:
        merged.tasks.append(original)
    raw_response.background = merged
```

Handles both `BackgroundTask` (single) and `BackgroundTasks` (multiple) on the Response.

## Test plan

- [x] Injected tasks + Response background task — both execute
- [x] Execution order: injected first, then response tasks
- [x] Multiple injected + single response task
- [x] Injected + Response with `BackgroundTasks` (multiple)
- [x] Only injected tasks (no Response.background) — still works
- [x] Only Response.background (no injection) — still works
- [x] No background tasks at all — still works
- [x] Normal dict return with injected tasks — still works
- [x] Existing background_tasks tutorial tests pass
- [x] 51 regression tests pass, 0 failures

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -73,6 +73,7 @@ from fastapi.utils import (
     is_body_allowed_for_status_code,
 )
 from starlette import routing
+from starlette.background import BackgroundTasks
 from starlette._exception_handler import wrap_app_handling_exceptions
 from starlette._utils import is_async_callable
 from starlette.concurrency import iterate_in_threadpool, run_in_threadpool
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Thanks for your interest, but we already have https://github.com/fastapi/fastapi/pull/14386.
Let's close this PR to avoid having duplicates.
You can review that existing PR and suggest improvements if any.

Thanks!

PR #15359, opened 2026-04-15
Auto-add HEAD method for GET routes

### LINKED ISSUE #1773: Automatically support HEAD method for all GET routes, as Starlette does

Bester1:
## Summary
Fixes #1773

When a GET route is defined, automatically add HEAD to the supported methods — matching the existing Starlette behavior.

## Changes
- **2-line fix** in `fastapi/routing.py`: After methods set is built, if GET is present, add HEAD
- Mirrors Starlette routing.py line 243-244: `if "GET" in self.methods: self.methods.add("HEAD")`
- 4 new tests in `tests/test_auto_head.py`

## Test Results
All 4 tests passing locally:
- `test_head_for_get_route` — HEAD / returns 200
- `test_head_for_get_route_with_path` — HEAD /items/42 returns 200
- `test_head_not_auto_added_for_post` — HEAD on POST route returns 405
- `test_head_response_has_no_body` — HEAD response has empty body

## Rationale
Per HTTP spec (RFC 7231), HEAD should be identical to GET except no body. Starlette already does this at the Route level. FastAPI should match this behavior.

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -279,6 +279,8 @@ def get_openapi_path(
     route_response_media_type: str | None = current_response_class.media_type
     if route.include_in_schema:
         for method in route.methods:
+            if method == "HEAD":
+                continue  # HEAD is auto-added for GET routes, not shown in OpenAPI schema
             operation = get_openapi_operation_metadata(
                 route=route, method=method, operation_ids=operation_ids
             )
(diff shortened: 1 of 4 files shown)


YuriiMotov:
This is basically the same approach as in https://github.com/fastapi/fastapi/pull/14792.
Let's close this to avoid having duplicates. You can review that PR and add your suggestions if you have any.

Thanks!

PR #15371, opened 2026-04-16
feat: automatically support HEAD method for all GET routes (fix #1773)

### LINKED ISSUE #1773: Automatically support HEAD method for all GET routes, as Starlette does

janshibhavsar24-gif:
## Summary

Fixes #1773.

FastAPI currently returns `405 Method Not Allowed` for `HEAD` requests on routes that only declare `GET`. This is incorrect: per [RFC 9110 §9.3.2](https://www.rfc-editor.org/rfc/rfc9110#section-9.3.2), any server that supports `GET` **must** support `HEAD`. Starlette already handles this in `Route.__init__` by automatically adding `HEAD` to `self.methods` whenever `GET` is present — FastAPI's `APIRoute` did not mirror this behaviour.

### `fastapi/routing.py`
- In `APIRouter.add_api_route`, auto-adds `HEAD` to `route.methods` when `GET` is present (mirrors Starlette).
- Tracks the auto-add with `route._auto_head = True`.
- When an explicit `@app.head()` is later registered on the same path, revokes the auto-`HEAD` from the `GET` route so the explicit handler always takes priority — backward-compatible with existing code.

### `fastapi/openapi/utils.py`
- Skips `HEAD` in OpenAPI schema generation when `GET` is also present on the same route. `HEAD` is implied by `GET` per the HTTP spec and is not a valid standalone OpenAPI operation; documenting it separately would produce invalid/confusing schemas.

### `fastapi/utils.py`
- `generate_unique_id` now excludes the auto-implied `HEAD` from the `operationId` suffix, so `@app.get("/items/")` continues to generate `..._get` rather than `..._head`.

### `tests/test_head_auto.py` (new)
7 tests covering:
- `HEAD` returns `200` on a `GET` route
- `HEAD` response has an empty body
- `HEAD` response headers match `GET` headers
- `HEAD` on a `POST`-only route returns `405`
- `GET` is unaffected after adding `HEAD` support
- `HEAD` does **not** appear as an operation in the OpenAPI schema
- `GET` is still present in the OpenAPI schema

## Behaviour

```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/items/{item_id}")
def get_item(item_id: str):
    return {"item_id": item_id}

client = TestClient(app)

# Before this PR → 405
# After this PR  → 200, empty body, correct headers
response = client.head("/items/foo")
assert response.content == b""
```

## Test plan

- [x] All 7 new tests in `tests/test_head_auto.py` pass
- [x] Existing `tests/test_extra_routes.py` (including explicit `@app.head()` handler) passes unchanged
- [x] Existing `tests/test_application.py` passes unchanged
- [x] Package installs cleanly in editable mode (`pip install -e ".[standard]"`)
- [x] End-to-end smoke test against the installed package passes

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -279,6 +279,11 @@ def get_openapi_path(
     route_response_media_type: str | None = current_response_class.media_type
     if route.include_in_schema:
         for method in route.methods:
+            # HEAD is implied by GET per HTTP spec (RFC 9110); it is auto-added
+            # to self.methods for routing, but should not appear as a separate
+            # operation in the OpenAPI schema.
+            if method.upper() == "HEAD" and "GET" in route.methods:
+                continue
             operation = get_openapi_operation_metadata(
                 route=route, method=method, operation_ids=operation_ids
             )
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -890,6 +890,7 @@ class APIRoute(routing.Route):
         if methods is None:
             methods = ["GET"]
         self.methods: set[str] = {method.upper() for method in methods}
+        self._auto_head: bool = False
         if isinstance(generate_unique_id_function, DefaultPlaceholder):
             current_generate_unique_id: Callable[[APIRoute], str] = (
                 generate_unique_id_function.value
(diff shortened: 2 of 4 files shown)


YuriiMotov:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

@janshibhavsar24-gif, have you reviewed it?

janshibhavsar24-gif:
> > 🤖 Generated with [Claude Code](https://claude.com/claude-code)
> 
> @janshibhavsar24-gif, have you reviewed it?

@YuriiMotov Yes, I've reviewed all the changes and adhere to these guidelines https://fastapi.tiangolo.com/contributing/ . I used Claude Code as a development tool but tested, and verified all the changes myself.

YuriiMotov:
As far as I can see, this is the same approach as in https://github.com/fastapi/fastapi/pull/14792
Let's close this to avoid having duplicates. You can review that PR and add your suggestions if you have any.

Thanks!

PR #15372, opened 2026-04-16
Alternative to #12066 - fix deadlock with thread overflow using "overflow" threads

olliemath:
This is an alternative to creating a separate CapacityLimiter to run critical-path tasks when the main threadpool is full. Mostly it exists since I'm not sure how to create a pull request on a pull request?

# what

This reserves 1 thread that can only be used by critical-path tasks. Those tasks are also free to use the rest of the threadpool, but at least one thread is reserved only for them.

To do this, it introduces more functions into `fastapi.concurrency` to replace the ones imported from starlette

- `run_in_threadpool`
- `iterate_in_threadpool`
- `run_until_first_complete`

together with a private method `_run_in_threadpool_with_overflow` that the critical functions can use.

We also added a `set_thread_limit` method, because it's now ineffective to directly modify the default thread limiter in anyio.

# how

This just limits the above functions using a separate limiter with capacity 1 less than the default global anyio limiter.

# why

see https://github.com/fastapi/fastapi/pull/12066

# limitations

~~It will still be possible for downstream users to deadlock the threadpool if they bypass the new CapacityLimiter - for instance by directly calling `anyio`'s threadpool primitives, instead of using the above functions in order to acquire resources.~~

In the end I used a dedicated global CapacityLimiter for fastapi to remove the possibility of 3rd party libraries (or downstream users) taking the extra thread we are reserving. That means that fastapi and other libraries will use different threadpools. That seems to be the only way to make the solution robust to other libraries "stealing" the final anti-deadlock thread

### DIFF
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -1,53 +1,63 @@
+import functools
 from collections.abc import AsyncGenerator, Callable, Iterable
 from contextlib import AbstractContextManager
 from contextlib import asynccontextmanager as asynccontextmanager
-from typing import ParamSpec, TypeVar
+from typing import Literal, ParamSpec, TypeVar
 
 import anyio.to_thread
 from anyio import CapacityLimiter
 from starlette.concurrency import (
-    iterate_in_threadpool as _starlette_iterate_in_threadpool,
-)
-from starlette.concurrency import run_in_threadpool as _starlette_run_in_threadpool
-from starlette.concurrency import (
-    run_until_first_complete as _starlette_run_until_first_complete,
+    _next,
+    _StopIteration,
+    run_until_first_complete,  # noqa
 )
 
 _P = ParamSpec("_P")
 _T = TypeVar("_T")
 
-# The default threadpool in anyio is 40. This limiter keeps one thread
-# for teardown tasks in order to prevent deadlocks when there is a pool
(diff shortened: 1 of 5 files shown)


olliemath:
A couple of thoughts here:

- Using our own global limiter protects us from libraries or downstream users taking the reserved thread
- Unfortunately that means that the total number of threads used by the app is our global limit + the anyio default limit (*see example below)
- Given this, the strategy of having a single separate threadpool (capacity 40) for high priority tasks is probably the simplest and cheapest. On the surface it appears more expensive than this PR, but in the worst case scenario it's really not.

*Example: starlette still uses its own iterate_in_threadpool with anyio's default limiter within `StreamingResponse`. That means that under heavy load we can be using 80 threads: 40 from the default limiter to stream responses and 40 from the limiter within concurrency.py to do the rest of the work.

Really that suggests to me, this PR is no better than just maintaining a single global 40 thread `_anti_deadlock_limiter` within concurrency.py and using it for high priority tasks.

The only way to really ensure only 40 threads are used globally with 1 reserved for high priority tasks would be to write a custom CapacityLimiter class and set it as the default within anyio.

olliemath:
In light of the above and after speaking with the maintainer of anyio I'm going to close this PR as overly complicated.

PR #15375, opened 2026-04-18
Improve clarity on path parameter type conversion

burnedpreadator:
Adds a note clarifying how path parameters are handled as strings by default and how type annotations enable automatic conversion and validation.

This helps avoid confusion for beginners understanding the difference between untyped and typed path parameters.

Also includes a minor readability improvement.


YuriiMotov:
Thanks for you efforts, but we don't need additional clarification here. It's explained even in more details and with code examples in the following sections on the same page.

PR #15377, opened 2026-04-19
docs: replace "re-use" in cookie parameter models

Rohan5commit:
## Summary
- Replace \"re-use\" with \"reuse\" in the cookie parameter models guide

## Related issue
- N/A (trivial docs-only wording fix)

## Guideline alignment
- Followed `CONTRIBUTING.md`, which points to the FastAPI contributing guide: https://fastapi.tiangolo.com/contributing/
- Keeps the change docs-only, single-file, and sentence-scoped

## Validation/testing
- Not run; docs-only wording change

### DIFF
--- a/docs/en/docs/tutorial/cookie-param-models.md
+++ b/docs/en/docs/tutorial/cookie-param-models.md
@@ -2,7 +2,7 @@
 
 If you have a group of **cookies** that are related, you can create a **Pydantic model** to declare them. 🍪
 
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Thanks, but I think `re-use` is acceptable.
FastAPI docs are written by mostly non-native speakers and there are quite some grammar and orthography mistakes. And I think it's fine unless it makes it hard to understand the meaning.
If we try to fix them all, we will not have time for other (more important) issues.

Thanks again for the efforts!

PR #15380, opened 2026-04-19
docs(readme): clarify production run command after fastapi dev

tboyila:
## What changed

Clarified the README to distinguish between:

- `fastapi dev` = development mode with auto-reload
- `fastapi run` = production mode without reload

## Why

Many beginners accidentally deploy with `fastapi dev` in production, which enables auto-reload and is not suitable for live environments. This makes the distinction explicit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,8 @@ The command `fastapi dev` reads your `main.py` file automatically, detects the *
 
 By default, `fastapi dev` will start with auto-reload enabled for local development.
 
+When you are ready to run the same app without auto-reload for production, use `fastapi run`.
+
 You can read more about it in the [FastAPI CLI docs](https://fastapi.tiangolo.com/fastapi-cli/).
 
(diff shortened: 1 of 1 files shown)


YuriiMotov:
I think we don't need this clarification - it's explained well in "FastAPI CLI docs" (link is just below that line), and even `fastapi dev` output already warns that for production you should use `fastapi run`:

```
 ╭────────── FastAPI CLI - Development mode ───────────╮
 │                                                     │
 │  Serving at: http://127.0.0.1:8000                  │
 │                                                     │
 │  API docs: http://127.0.0.1:8000/docs               │
 │                                                     │
 │  Running in development mode, for production use:   │
 │                                                     │
 │  fastapi run                                        │
 │                                                     │
 ╰─────────────────────────────────────────────────────╯
 ```

So, I think it should be clear already.

Thanks!

PR #15384, opened 2026-04-20
fix: replace innerHTML with safe DOM methods in termynal.js

beejak:
Resubmit of security fix for termynal.js DOM injection via innerHTML. See discussion on PR #15381 for full attack scenario details.

### DIFF
--- a/docs/en/docs/js/termynal.js
+++ b/docs/en/docs/js/termynal.js
@@ -222,10 +222,18 @@ class Termynal {
      */
     lineDataToElements(lineData) {
         return lineData.map(line => {
-            let div = document.createElement('div');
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Please, re-read my reply ([link](https://github.com/fastapi/fastapi/pull/15381#issuecomment-4278219203)) - I've already addressed this. It will not prevent any attempt to inject malicious code into docs preview, there are many other ways to do it. All maintainers are aware of this and carefully review changes before opening docs preview.

**Please, don't open duplicate PRs. Instead, reply in the comments of existing PR. We can re-open closed PR if you manage to convince me**

PR #15388, opened 2026-04-20
🐛 fix deadlocks in depends by using a separate CapacityLimiter for teardown

olliemath:
# what

This is https://github.com/fastapi/fastapi/pull/12066 but with a single global "teardown" CapacityLimiter instead of unlimited one-off capacity limiters (and hence threads).

# why

If there is a limited pool of resources, like database connections, we need separate pools of threads for acquiring and releasing those resources, otherwise there will always be deadlocks. However the PR doesn't seem to have moved for a while due to the use of one-off / unlimited CapacityLimiters.

# anything else

We use a fixed, but arbitrary, limit of 5. It seems that a limit of 1 is too low as operations will serialise. Testing locally with a concurrency of 1000 I didn't notice any difference beyond a limit of ~3. Any number under 40 could be deemed "wrong" if the user has expensive teardown operations.

We can still trigger similar deadlocks by releasing resources in the "finally" section of a middleware. However this is not recommended in the documentation (although it's not explicitly prohibited), so I think it's OK.

### DIFF
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -1,7 +1,8 @@
-from collections.abc import AsyncGenerator
+import functools
+from collections.abc import AsyncGenerator, Callable
 from contextlib import AbstractContextManager
 from contextlib import asynccontextmanager as asynccontextmanager
-from typing import TypeVar
+from typing import ParamSpec, TypeVar
 
 import anyio.to_thread
 from anyio import CapacityLimiter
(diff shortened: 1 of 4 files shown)


avinashkamat48 (commented):
The new deadlock regression tests can hang the suite indefinitely if the bug regresses, because both `test_competing_acquire_release` and `test_depends_deadlock` wait on the exact condition that deadlocks. Please wrap the concurrent section in an AnyIO/asyncio timeout (`fail_after`, `move_on_after`, or `asyncio.wait_for`) so a regression fails quickly with a useful assertion instead of stalling CI.

olliemath:
I'll look at adding this end of week (on vacation at the moment)

olliemath:
> The new deadlock regression tests can hang the suite indefinitely if the bug regresses, because both `test_competing_acquire_release` and `test_depends_deadlock` wait on the exact condition that deadlocks. Please wrap the concurrent section in an AnyIO/asyncio timeout (`fail_after`, `move_on_after`, or `asyncio.wait_for`) so a regression fails quickly with a useful assertion instead of stalling CI.

Timeouts added. Note that I made them quite long (10s - about 20x the time taken locally). This should prevent flake from slow CI runners. If we see any spurious failures I'd advocate for an even longer timeout since this would only ever be hit in failure mode.

olliemath:
The timeouts work correctly in single process mode, but it seems when using `pytest -n auto` they interfere with pytest xdist's own teardown process. Will need to look into this more deeply - but it's probably not enough to just trigger a timeout since that will leave the threadpool full of tasks that can't execute. Since the default anyio threadpool is an implicit **global** it's not surprising that it leaving it full leaves everything gummed up :thinking:

olliemath:
OK, this is now fully compatible with `-n auto` and timeouts both in success and failure modes. The concurrency test I made more cooperative so that it times out gracefully. The end to end deadlock test I had to run as a subprocess in order to isolate the badness in anyio's global default capacity limiter, which is not removed by simply adding a timeout.

olliemath:
And that final commit removes the need for a subprocess by explicitly timing out the `mutex.acquire` in the threads. This now fails gracefully without hanging the test suite.

You can easily simulate the failure mode by adding a line `run_in_teardown_threadpool = run_in_threadpool` to the end of `concurrency.py` and running the test suite.

PR #15398, opened 2026-04-21
🐛 Fix duplicate `operationId` for routes registered with multiple HTTP methods

Pedrom2002:
## Description

When a route is registered with multiple methods (e.g. `add_api_route(..., methods=["POST", "DELETE"])` or `@app.api_route(..., methods=[...])`), `route.unique_id` is generated once with an arbitrary single method as suffix. Every OpenAPI operation produced from that route ends up sharing the same `operationId`, triggering the "Duplicate Operation ID" warning and breaking client code generators (openapi-generator, etc.).

## Fix

`get_openapi_operation_metadata()` now regenerates a per-method `operationId` when the route has more than one method and no explicit `operation_id` is set. It reuses the route's own `generate_unique_id_function` on a shallow copy of the route scoped to the current method, so both the default function and user-provided ones produce unique ids.

`route.unique_id` itself is left untouched, which keeps response / stream model names (derived from `unique_id`) stable.

## Tests

Added `tests/test_multiple_methods_operation_id.py` with 9 tests:

- `add_api_route` with multiple methods
- `@api_route` decorator with multiple methods
- Explicit `operation_id` still respected (and still warns on dup)
- Custom `generate_unique_id_function` honoured per method
- Single-method route unchanged (regression guard)
- `include_router` path
- `/openapi.json` generation end-to-end
- Parametrized across 2 and 5 method combos

Full suite: `3130 passed, 32 skipped, 5 xfailed`.

Closes #13175.
Related: #10694, #15215.

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -243,6 +243,19 @@ def get_openapi_operation_metadata(
     if route.description:
         operation["description"] = route.description
     operation_id = route.operation_id or route.unique_id
+    if not route.operation_id and route.methods and len(route.methods) > 1:
+        # Route registered with multiple HTTP methods (e.g.
+        # add_api_route(..., methods=["POST", "DELETE"])). route.unique_id was
+        # generated using an arbitrary single method as suffix, so every
+        # operation generated from the same route would share the same
+        # operationId. Regenerate a per-method id using the route's unique-id
+        # function on a view of the route limited to the current method.
+        unique_id_function = route.generate_unique_id_function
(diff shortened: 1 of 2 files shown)


tiangolo:
Thanks for the interest. :coffee: 

`add_api_route(..., methods=["POST", "DELETE"])` is not supported / documented by FastAPI. I would consider that an antipattern. So I prefer to avoid it and add something that would in some way encourage its usage.

So I'll pass on this, thanks! :cake:

PR #15411, opened 2026-04-22
fix(deps): resolve Annotated forward refs with __future__ annotations

armorbreak001:
## Problem

When using `from __future__ import annotations` together with `Annotated[..., Depends(...)]` where the type is defined **after** the route decorator, FastAPI raises a `PydanticUserError` about an unresolvable ForwardRef.

The root cause: the lenient type evaluator (`eval_type_lenient`) leaves the entire annotation as a single `ForwardRef('Annotated[Potato, Depends(get_potato)]')` because it cannot resolve the inner type name (defined later in the file). Since the result is a plain `ForwardRef` and not an `Annotated` type, the code that extracts `Depends()`, `Field()`, etc. from `Annotated` metadata never runs — the parameter gets treated as a body field instead of a dependency, and pydantic later chokes on the nested ForwardRef.

**Before** (fails):

```python
from typing import Annotated

app = FastAPI()

def get_potato() -> Potato:
    ...

@app.get('/')
async def root(potato: Annotated[Potato, Depends(get_potato)]):  # PydanticUserError
    ...
```

Moving `Potato` above the route or removing `Annotated` both work around it, confirming this is specific to unpacking forward-ref'd Annotated types.

## Solution

In `analyze_param()`, when the annotation is still a `ForwardRef`, evaluate it using `eval` with a safe-globals dict that maps **undefined names** to `ForwardRef` objects. This gives us partial evaluation:

```python
Annotated[ForwardRef('Potato'), <function get_potato>]
```

The `Annotated` wrapper is now unpackable, so `Depends()` is correctly discovered. The inner type stays as a `ForwardRef` for downstream resolution — same behaviour as the non-`Annotated` forward-ref path.

Also changed `get_args(annotation)` to `get_args(use_annotation)` so that the re-evaluated annotation is used consistently throughout the function.

Fixes #13056

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -302,6 +302,8 @@ def get_dependant(
         own_oauth_scopes=own_oauth_scopes,
         parent_oauth_scopes=parent_oauth_scopes,
     )
+    unwrapped_call = inspect.unwrap(call)
+    globalns = getattr(unwrapped_call, "__globals__", {})
     current_scopes = (parent_oauth_scopes or []) + (own_oauth_scopes or [])
     path_param_names = get_path_param_names(path)
     endpoint_signature = get_typed_signature(call)
@@ -313,6 +315,7 @@ def get_dependant(
             annotation=param.annotation,
             value=param.default,
             is_path_param=is_path_param,
+            globalns=globalns,
         )
         if param_details.depends is not None:
             assert param_details.depends.dependency
@@ -396,6 +399,7 @@ def analyze_param(
     annotation: Any,
     value: Any,
     is_path_param: bool,
+    globalns: dict[str, Any] | None = None,
 ) -> ParamDetails:
     field_info = None
     depends = None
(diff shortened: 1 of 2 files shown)


Samielakkad (commented):
Retracted.

armorbreak001:
Thanks @Samielakkad for the review — good catch on test coverage.

I've added **4 regression tests** in `tests/test_annotated_future_annotations.py`:

1. **`test_annotated_fwd_ref_resolved_as_dependency`** — Annotated[Potato, Depends()] where Potato is defined after the route → correctly treated as dependency injection (returns dependency result)
2. **`test_annotated_fwd_ref_no_body_required`** — No request body needed (confirms dep-injection path, not body parsing)
3. **`test_annotated_fwd_ref_multiple_deps`** — Multiple forward-ref'd Annotated deps on the same route
4. **`test_openapi_schema_no_body_params`** — OpenAPI schema sanity: no requestBody for dep-injected params

All 4 pass ✅

I initially tried a 5th test for the 'inner type stays ForwardRef' case you mentioned, but hit a Python limitation: when a *dependency function's own signature* contains forward refs (`def f() -> UndefinedType`) and `inspect.signature(eval_str=True)` runs before the type exists, Python raises TypeError before FastAPI even gets involved. That's pre-existing behavior unrelated to this PR's change — the fix specifically targets the *endpoint parameter* annotation path.

CI should pick up the new commit shortly.

PR #15412, opened 2026-04-22
Fix typo: "processors" to "processes" in concurrency docs

taahamultani:
This change corrects terminology to avoid confusion between CPU processors and processes/tasks in the concurrency example.


YuriiMotov:
Thanks, but there is no mistake here. 'processors' is correct

PR #15414, opened 2026-04-23
fix: reject `None` from security dependencies when annotation is non-optional

homanp:
When a security scheme (OAuth2PasswordBearer, HTTPBearer, APIKeyHeader, etc.) has auto_error=False, it returns None on missing or invalid credentials. Previously, FastAPI injected this None into the endpoint parameter regardless of its type annotation, causing either a 500 crash (if the handler called methods on the value) or a silent auth bypass (if the handler did a truthy check).

Now, when a security scheme dependency returns None and the endpoint parameter is annotated as non-optional (e.g. `str` instead of `str | None`), FastAPI returns a 422 validation error with a message explaining the mismatch. Parameters annotated as optional continue to receive None as before.

The check only applies to direct security scheme dependencies, not to intermediate functions in dependency chains.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -1,6 +1,7 @@
 import dataclasses
 import inspect
 import sys
+import types
 from collections.abc import (
     AsyncGenerator,
     AsyncIterable,
@@ -89,6 +90,19 @@ multipart_incorrect_install_error = (
     'And then install "python-multipart" with: \n\n'
     "pip install python-multipart\n"
 )
+
+
+def _annotation_allows_none(annotation: Any) -> bool:
+    if annotation is inspect.Parameter.empty or annotation is Any:
+        return True
+    if annotation is type(None) or annotation is None:
+        return True
+    origin = get_origin(annotation)
+    if origin is Annotated:
+        return _annotation_allows_none(get_args(annotation)[0])
+    if origin is Union or origin is types.UnionType:
+        return type(None) in get_args(annotation)
+    return False
 
 
 def ensure_multipart_is_installed() -> None:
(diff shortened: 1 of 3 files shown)


fastapi/dependencies/utils.py
```diff
@@ -679,6 +694,26 @@ async def solve_dependencies(
...
+                        "loc": ("dependency", sub_dependant.name),
+                        "msg": (
+                            f"Dependency returned None for parameter "
+                            f"'{sub_dependant.name}' which is annotated as "
+                            f"non-optional. Use 'Optional[...]' or '... | None' "
+                            f"if the dependency can return None (e.g. when using "
+                            f"auto_error=False)."
+                        ),
+                    }
+                )
```
YuriiMotov: This is not a **request** error, so it shouldn't be returned as status code 422.
This should be treated as server error

YuriiMotov (changes requested):
@homanp, thanks for your interest and efforts!

I'm in doubts about if we need to handle this, but regardless this, we shouldn't treat this as a client error (4xx status codes are for client errors, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#client_error_responses)

fastapi/dependencies/utils.py
homanp: Fixed

YuriiMotov (commented):
Looking at this PR one more time, I don't actually see a point in changing one exception to another.
It would probably make more sense if we could detect this type mismatch on startup and show a clear error (`FastAPIError`)..

Again, I'm still not sure we should put more efforts into it now. You can see how many pending PRs there are in this repo, and this is probably not the top priority task.
Maybe better wait for Sebastian to validate the idea first

tiangolo:
I'll pass on this for now, thank you! :coffee:

PR #15423, opened 2026-04-24
fix: escape user-controlled values in Swagger UI and ReDoc HTML

homanp:
The /docs and /redoc endpoints embed openapi_url, oauth2_redirect_url, title, and CDN URLs directly into HTML and JavaScript using f-strings without escaping. When root_path comes from a reverse proxy header (like X-Forwarded-Prefix), an attacker can inject arbitrary JavaScript into the generated page.

This applies html.escape() to values in HTML attribute and text contexts, and uses the existing _html_safe_json() (which produces properly quoted JSON strings) for values embedded in JavaScript contexts. This is the same approach already used for swagger_ui_parameters.

### DIFF
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -1,4 +1,5 @@
 import json
+from html import escape as html_escape
 from typing import Annotated, Any
 
 from annotated_doc import Doc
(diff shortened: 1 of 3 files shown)


homanp:
@tiangolo you might want to patch this

homanp:
Moving to private security advisory fork for responsible disclosure.

PR #15424, opened 2026-04-24
Discard default lifespans before merging

alfred82santa:
This pull request improves performance of lifespan contexts. Currently routers inject a default lifespan when no one is provided. The issue is when a router is included to other router, the lifespans (base and child routers' lifespans) must be merged, even when they are no-op default lifespans. This PR checks if base router or child router use a default lifespan to decide if they must be merged.

On the other hand, it changes the way to merge async context manager with a more pythonic way.


alfred82santa:
@YuriiMotov can you check it? or do you prefer I close it?

YuriiMotov:
@alfred82santa, thanks for your interest and efforts!

Let's close this for now to keep the list of PRs cleaner.
I saved the link to this PR in the list of possible performance optimizations, we will review them when we have time to work on it.

PR #15426, opened 2026-04-24
Fix SSE/JSONL stream_item_type lost through include_router

### LINKED ISSUE #15401: SSE `stream_item_type` not propagated through `APIRouter` + `include_router`

rzsultan-boop:
Fixes #15401.

## Problem

`APIRoute.__init__` only detects streaming return annotations inside the
`isinstance(response_model, DefaultPlaceholder)` branch. `include_router`
calls `add_api_route` with the already-resolved `response_model` (None
for streaming routes), so the detection is skipped on the merged route.
Both `stream_item_type` and `stream_item_field` stay None.

Consequences:
- OpenAPI schema drops `contentSchema` / `itemSchema` for the route.
- The request handler, which is built in `__init__` via
  `request_response(self.get_route_handler())`, closes over
  `stream_item_field=None`, so runtime validation and serialization of
  stream items are silently disabled.

## Fix

In `APIRouter.include_router`, right after `add_api_route` appends the
merged route, copy `stream_item_type` and `stream_item_field` from the
source route, then rebuild `new_route.app` so the handler closure picks
up the restored field. An `isinstance(new_route, APIRoute)` check
narrows Starlette's `BaseRoute` for mypy.

Both SSE (`EventSourceResponse`) and JSONL routes go through the same
path, so both are covered.

## Tests

Four tests in `tests/test_sse.py`:

- `test_stream_item_type_preserved_through_include_router` asserts both
  attributes are set after `include_router`.
- `test_sse_stream_item_field_set_on_direct_route` baseline for
  directly-registered routes.
- `test_sse_openapi_schema_after_include_router` verifies the item
  model appears in `components.schemas`.
- `test_sse_runtime_serialization_after_include_router` end-to-end
  HTTP call via `TestClient` checking the serialized frame body.

Full suite passes locally. `ruff check`, `ruff format --check`, and
`mypy fastapi/routing.py` are clean.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1794,6 +1794,17 @@ class APIRouter(routing.Router):
                         self.strict_content_type,
                     ),
                 )
+                # include_router calls add_api_route with response_model=route.response_model
+                # (already resolved to None), so the detection block in APIRoute.__init__
+                # (lines 847-865) is skipped — both stream_item_type and stream_item_field
+                # remain None. The handler at __init__ line 976 closes over stream_item_field,
+                # and OpenAPI utils reads stream_item_field (not stream_item_type), so we
+                # must restore both and re-bake the handler.
+                if route.stream_item_type is not None:
+                    new_route = self.routes[-1]
+                    new_route.stream_item_type = route.stream_item_type
+                    new_route.stream_item_field = route.stream_item_field
+                    new_route.app = request_response(new_route.get_route_handler())
             elif isinstance(route, routing.Route):
                 methods = list(route.methods or [])
                 self.add_route(
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1802,9 +1802,10 @@ class APIRouter(routing.Router):
                 # must restore both and re-bake the handler.
                 if route.stream_item_type is not None:
                     new_route = self.routes[-1]
-                    new_route.stream_item_type = route.stream_item_type
-                    new_route.stream_item_field = route.stream_item_field
-                    new_route.app = request_response(new_route.get_route_handler())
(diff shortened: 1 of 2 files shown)


rzsultan-boop:
Hi @tiangolo / @lbernick, could you please add the **bug** label to this PR? The `check-labels` workflow is failing because the auto-labeler only covers docs/`.github/` and translation paths — it has no rule for `fastapi/` source changes, so no label was applied automatically.

Thanks!

YuriiMotov:
I don't like this approach as the logic is not clear - `stream_item_type` is set to `None` in `__init__` and then being overridden in `include_router`.

Please, review the alternative approach: https://github.com/fastapi/fastapi/pull/15077

rzsultan-boop:
Closing in favor of #15077, which threads stream_item_type through __init__/add_api_route/include_router parameters — the established pattern for this codebase (see strict_content_type). Thanks for the review.

PR #15429, opened 2026-04-25
🐛 Fix unhashable Annotated type in get_definitions OpenAPI schema generation

Yuerchu:
## Summary

- `get_definitions()` in `fastapi/_compat/v2.py` uses a set comprehension to collect field annotations for deduplication
- On **Python 3.14** with **Pydantic v2.13+**, some `Annotated` types contain `FieldInfoMetadata` objects that are **not hashable**, causing a `TypeError` when generating the OpenAPI schema:

```
TypeError: cannot use 'typing._AnnotatedAlias' as a set element (unhashable type: 'FieldInfoMetadata')
```

- This is triggered when routes use `Annotated` parameters with metadata such as `Query(max_length=..., description=...)` or similar constructs that produce unhashable Pydantic field metadata

## Fix

Use `id()` for identity-based deduplication instead of relying on `__hash__` of annotation objects. Since annotations are interned type objects, `id()` comparison is semantically equivalent for this filtering purpose.

## Changes

```diff
-    input_types = {f.field_info.annotation for f in fields}
+    input_type_ids = {id(f.field_info.annotation) for f in fields}
     unique_flat_model_fields = {
-        f for f in flat_model_fields if f.field_info.annotation not in input_types
+        f for f in flat_model_fields if id(f.field_info.annotation) not in input_type_ids
     }
```

## Environment

- Python 3.14.3
- Pydantic 2.13.3
- FastAPI 0.136.1

### DIFF
--- a/tests/test_openapi_unhashable_annotated.py
+++ b/tests/test_openapi_unhashable_annotated.py
@@ -1,0 +1,35 @@
+from typing import Annotated
+
+from pydantic.fields import FieldInfo
+
+from fastapi._compat.v2 import ModelField, get_definitions
+
+
+class UnhashableMetadata:
+    """Simulates third-party field metadata (e.g. from SQLModel extensions)
+    that does not implement ``__hash__``."""
+
+    __hash__ = None  # type: ignore[assignment]
+
+
+def test_get_definitions_with_unhashable_annotation() -> None:
+    """``get_definitions()`` should not crash when a field annotation is an
+    ``Annotated`` type containing unhashable metadata.
+
+    Regression: the function used a set comprehension over
+    ``field_info.annotation`` values.  ``Annotated`` types that include
+    metadata with ``__hash__ = None`` (e.g. certain SQLModel / Pydantic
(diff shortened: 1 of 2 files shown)


tests/test_openapi_unhashable_annotated.py
```diff
@@ -0,0 +1,34 @@
+from typing import Annotated
+
+from fastapi._compat.v2 import ModelField, get_definitions
+from pydantic.fields import FieldInfo
+
+
+class UnhashableMetadata:
+    """Simulates third-party field metadata (e.g. from SQLModel extensions)
+    that does not implement ``__hash__``."""
```
YuriiMotov, line 9: `SQLModel` is already in test dependencies, so you can use it to make this test more realistic.
Ideally, it should be a FastAPI app, not just `get_definitions`

YuriiMotov:
Please, review changes carefully before opening PR.
AI-generated PRs without proper human review will be closed. See https://fastapi.tiangolo.com/contributing/#automated-code-and-ai

PR #15435, opened 2026-04-27
Fix: preserve stream_item_type when using include_router with SSE routes

Prashanth1602:
## Problem

When defining SSE routes on an APIRouter and including it via include_router, the resulting route loses stream_item_type. This causes the OpenAPI schema to omit contentSchema for text/event-stream responses.

## Root Cause

APIRoute detects stream_item_type only when response_model is DefaultPlaceholder. During include_router, routes are recreated using add_api_route with response_model already resolved to None, so detection is skipped and stream_item_type is not set.

## Fix

After recreating the route via add_api_route, copy stream_item_type from the original route to the new route to preserve streaming metadata.

## Tests

- Added test to verify stream_item_type is preserved after include_router
- Ensured non-stream routes are unaffected

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1794,6 +1794,14 @@ class APIRouter(routing.Router):
                         self.strict_content_type,
                     ),
                 )
+                new_route = self.routes[-1]
+                if (
+                    isinstance(new_route, APIRoute)
+                    and isinstance(route, APIRoute)
+                    and getattr(route, "stream_item_type", None) is not None
+                ):
+                    new_route.stream_item_type = route.stream_item_type
+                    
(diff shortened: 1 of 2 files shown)


YuriiMotov:
This is the same approach as in https://github.com/fastapi/fastapi/pull/15426.
I don't like this approach as the logic is not clear - `stream_item_type` is set to `None` in `__init__` and then being overridden in `include_router`.

Please, review the alternative approach: https://github.com/fastapi/fastapi/pull/15077

---

**UPD:** looks like https://github.com/fastapi/fastapi/pull/15426 is more developed. Let's close this and continue with #15426

Thanks!

PR #15445, opened 2026-04-28
fix(routing): preserve SSE stream_item_type through include_router

### LINKED ISSUE #15401: SSE `stream_item_type` not propagated through `APIRouter` + `include_router`

onthebed:
`include_router` recreates SSE routes with `response_model=None`, so the merged route drops `stream_item_type` and loses the item schema.

Copied the streaming metadata onto the recreated route and rebuilt its handler so OpenAPI and runtime serialization keep seeing the field.

Added a regression test for an included SSE route that checks the merged route metadata and OpenAPI schema.

Fixes #15401

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1795,7 +1795,10 @@ class APIRouter(routing.Router):
                     ),
                 )
                 new_route = self.routes[-1]
-                if isinstance(new_route, APIRoute) and route.stream_item_type is not None:
+                if (
+                    isinstance(new_route, APIRoute)
+                    and route.stream_item_type is not None
+                ):
                     new_route.stream_item_type = route.stream_item_type
                     new_route.stream_item_field = create_model_field(
                         name=f"StreamItem_{new_route.unique_id}",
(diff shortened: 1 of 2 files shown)


YuriiMotov:
This is basically the same approach as in https://github.com/fastapi/fastapi/pull/15426
See my comment to that PR: https://github.com/fastapi/fastapi/pull/15426#issuecomment-4326839651

PR #15463, opened 2026-04-30
✏️ Fix Azkaban spelling typo in `virtual-environments.md‎`

isaacbernat:
### Description:

Corrected "prisoner-of-askaban" (s) to "prisoner-of-azkaban" (z) for consistency with the rest of the examples.


YuriiMotov:
We should never update translations, only original English versions

isaacbernat:
Understood, thanks for the correction @YuriiMotov . I've reverted all non English changes now.

PR #15464, opened 2026-04-30
📝 Document `--entrypoint` CLI option

YuriiMotov:
Just for completeness, in addition to specifying `entrypoint` in `pyproject.toml` and passing path as a parameter, we can specify `--entrypoint` CLI option


tiangolo (approved):
Nice, thanks! 📝

PR #15468, opened 2026-05-01
👷 Fix missing credentials issue in `translate` workflow

YuriiMotov:
Currently translation workflow fails with `fatal: could not read Username for 'https://github.com/': No such device or address` due to missing credentials.
This PR fixes it

---

Should we also add `fail-fast: false`, so that if one job fails, others would not be cancelled?


tiangolo (approved):
Thank you! 🚀

PR #15472, opened 2026-05-01
🌐 Fix typos in Spanish LLM-prompt

crr004:
~Added some new technical terms to Spanish LLM prompt.~

**UPD YuriiMotov:** removed new terms as we discussed in the comments. So, this PR just fixes a couple of typos


docs/es/llm-prompt.md
```diff
@@ -97,3 +97,16 @@ For the next terms, use the following translations:
 * load balance: load balance (do not translate to "balancear carga")
 * self hosting: self hosting (do not translate to "auto alojamiento")
 * timing attack: timing attack (do not translate to "ataque de temporización")
+* endpoint: endpoint (do not translate to "punto final")
+* middleware: middleware (do not translate to "software intermedio")
+* handler: handler (do not translate to "manejador")
+* payload: payload
+* runtime: runtime
+* deploy: deploy / desplegar, choose one consistently
```
YuriiMotov: > choose one consistently

This will not work - LLM only has one page in the context at a time..
I think if both are good, it's Ok to use either of them. Or we should choose the one that should be used

crr004 (commented):
Reviewed with proposed changes.

docs/es/llm-prompt.md
```diff
@@ -97,3 +97,16 @@ For the next terms, use the following translations:
...
+* payload: payload
+* runtime: runtime
+* deploy: desplegar
+* deployment: despliegue
+* logging: logging
+* traceback: traceback
+* decorator: decorador
+* dependency injection: inyección de dependencias
+* background task: background task
+* WebSocket: WebSocket
```
tiangolo: Giving it a quick check, I don't see anywhere in the docs where it uses "punto final" or "software intermedio", it seems to always use the correct terms already.

Did you find places where it was using the wrong term?

crr004: No. I haven't found those exact errors in the docs. I just added them cause they are the literal translations in Spanish and thought it would be useful. I didn't know the "do not translate..." part must only be added when a previous error is spotted.

YuriiMotov: I think `(do not translate to XXX)` is for common mistakes (when LLM incorrectly translates this term to `XXX`).
If the idea is to ensure consistency, we can shorten it to `(do not translate)`.
But, it doesn't make sense to add all possible terms to this list. We should only add those that have several common translations and we prefer to use only one (or two) of them.
In general, the smaller prompt - the better (if it gives good results off course).

crr004: Giving it a second thought, you are totally right. I added some typical technical terms that are commonly used in the documentation, but not because I found any mistakes in the current documentation. It is true that, taking this premise, we could add infinite terms and degrade the prompt.

Taking what you said, maybe the PR should not be merged. I'll try to add more useful terms in the future :)

YuriiMotov:
```suggestion

```

YuriiMotov: I removed the new terms and will merge only typo fixes.
Thanks!

PR #15477, opened 2026-05-03
feat(deps): merge Body/Form/File/Query with Depends(model) in Annotated

mahenzon:
### Example

```python

def register_post_route(
    path: str,
    schema: type[ItemBase],
):
    @router.post(path)
    def create_item(
        item: Annotated[
            ItemBase,  # for type checkers and IDEs
            Body(),  # for FastAPI - where to search
            Depends(schema),  # for FastAPI - what to expect
        ],
    ):
        return item

    return create_item
```

### Summary

FastAPI can combine `Body` / `Form` / `File` / `Query` with `Depends(some_model)` inside the same `Annotated[...]` so validation / OpenAPI use the runtime model while the parameter’s declared type stays the static first argument (e.g. base class for type checkers).

Order of `Body`/`Form` / … and `Depends` in `Annotated` does not matter.

### Motivation

Factories / dynamic routes need a non-literal model class; documenting and supporting `Annotated[Base, Body(), Depends(schema)]` avoids losing the location marker or the model.

### Implementation notes (high level)

New merge path in dependency / param handling (`fastapi/dependencies/utils.py`): detects compatible marker sets, validates constraints (e.g. no `Query` combined with body-like markers, not on path params, `Depends` resolves to a single `BaseModel` subclass, etc.), builds merged field info and annotation.

### Docs

New advanced section **“Composing Annotated metadata for runtime validation schemas”** with subpages (body / form / file / query), nav in `mkdocs.yml`, link from Request Body tutorial.

### Tests

- integration-style tests for body, file, form, query, and query+body shapes
- shared helpers / doc example loaders as staged under `tests/`.

### DIFF
--- a/docs/en/docs/advanced/body-depends-model-merge/body.md
+++ b/docs/en/docs/advanced/body-depends-model-merge/body.md
@@ -1,0 +1,19 @@
+# JSON request body { #json-request-body }
+
+Minimal end-to-end example: a helper `register_post_route` takes `schema: type[ItemBase]`. The path operation stays the same, while `Gadget` and `Part` change both the OpenAPI schema and the expected JSON body.
+
+The same `Annotated[..., Body(), Depends(schema)]` pattern works for all methods whenever you send JSON body.
+
+This file uses `POST` only to stay short. See [Which HTTP methods?](index.md#which-http-methods) on the overview page.
+
+## Models { #models }
+
+{* ../../docs_src/body_depends_model_merge_body/tutorial001_an_py310.py ln[10:19] hl[10:19] *}
+
+## The path operation { #the-path-operation }
+
+The parameter uses `ItemBase` for static typing, `Body()` so FastAPI treats the payload as JSON, and `Depends(schema)` so validation uses the class supplied when the route is registered. `register_post_route` wraps that in a factory; registrations mount `items_router` at prefix `/items`, so the paths are `/items/objects/gadgets/` and `/items/objects/parts/`:
+
+{* ../../docs_src/body_depends_model_merge_body/tutorial001_an_py310.py ln[22:53] hl[22:53] *}
+
+See [overview](index.md) for context and alternatives.
--- a/docs/en/docs/advanced/body-depends-model-merge/file.md
+++ b/docs/en/docs/advanced/body-depends-model-merge/file.md
@@ -1,0 +1,23 @@
+# File uploads { #file-uploads }
+
+Multipart uploads: `File()` inside `Annotated`, plus `Depends(schema)` when `schema` includes fields such as `UploadFile`. OpenAPI then describes file parts the same way as in the regular [Request Files](../../tutorial/request-files.md) tutorial.
+
+The sample uses `POST`, which is what browsers and most clients use for uploads. `PUT`, `PATCH`, or `DELETE` with multipart follow the same rules if the client actually sends multipart data. See [Which HTTP methods?](index.md#which-http-methods).
+
+## Models { #models }
+
+{* ../../docs_src/body_depends_model_merge_file/tutorial001_an_py310.py ln[10:19] hl[10:19] *}
+
+## The path operation { #the-path-operation }
+
+{* ../../docs_src/body_depends_model_merge_file/tutorial001_an_py310.py ln[22:41] hl[22:41] *}
+
+Full helper and registrations for `/files/attachments/commented/` and `/files/attachments/named/`:
+
+{* ../../docs_src/body_depends_model_merge_file/tutorial001_an_py310.py ln[44:57] hl[44:57] *}
+
+The handler returns a small JSON payload: upload metadata plus `data` from `model_dump(exclude={"file"})` so non-file fields (`comment`, `name`, …) stay JSON-serializable.
+
+Install [`python-multipart`](https://github.com/Kludex/python-multipart) for multipart support in your own project.
+
+See [overview](index.md) for context.
--- a/docs/en/docs/advanced/body-depends-model-merge/form.md
+++ b/docs/en/docs/advanced/body-depends-model-merge/form.md
@@ -1,0 +1,23 @@
+# Form data { #form-data }
+
+Form-encoded data needs `Form()` in the metadata and `Depends(YourModel)` when the provided model is only known in runtime.
+
+This example uses `POST` with `application/x-www-form-urlencoded`. The merge behaves the same on other methods if the client sends a form body. See [Which HTTP methods?](index.md#which-http-methods).
+
+## Models { #models }
+
+{* ../../docs_src/body_depends_model_merge_form/tutorial001_an_py310.py ln[10:19] hl[10:19] *}
+
+## The path operation { #the-path-operation }
+
+{* ../../docs_src/body_depends_model_merge_form/tutorial001_an_py310.py ln[22:37] hl[22:37] *}
+
+Full helper and router wiring (`/auth/session/password/` and `/auth/session/token/`):
+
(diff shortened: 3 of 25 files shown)


mahenzon:
Full feature description and alternative approaches here in the doc:
https://21661eeb.fastapitiangolo.pages.dev/advanced/body-depends-model-merge/

YuriiMotov:
I think what this PR is trying to solve can already be achieved in current FastAPI version..

Could you please take a look at the following implementation of `register_post_route`?

<details>

```
import inspect
from typing import Annotated, Any, TypeVar

from fastapi import APIRouter, Body, Depends, FastAPI

app = FastAPI()


class ItemBase(BaseModel):
    name: str


class Item1(ItemBase):
    description: str


T = TypeVar("T", bound=ItemBase)


def register_post_route(
    router: APIRouter,
    path: str,
    schema: type[T],
) -> None:
    async def get_item(item: Any) -> Any:
        return item

    get_item.__setattr__(
        "__signature__",
        inspect.Signature(
            parameters=[
                inspect.Parameter(
                    name="item",
                    kind=inspect.Parameter.POSITIONAL_OR_KEYWORD,
                    default=Body(),
                    annotation=schema,
                ),
            ],
            return_annotation=schema,
        ),
    )

    @router.post(path, response_model=schema)
    def create_item(
        item: Annotated[
            ItemBase,
            Depends(get_item),
        ],
    ) -> Any:
        return item


router = APIRouter()
register_post_route(router, "/items1", Item1)
app.include_router(router)

```

</details>

mahenzon:
> I think what this PR is trying to solve can already be achieved in current FastAPI version..

@YuriiMotov Hi!

Sure, current version of FastAPI can handle it even without signature updates. IMO updating signature is super-non-intuitive and harder to understand. The working approach will be:

<details>

<summary>Code (shorter and easier to read than with signature updates)</summary>

```python
from typing import Annotated, Any, TypeVar
from uuid import UUID

from fastapi import APIRouter, Body, Depends, FastAPI

app = FastAPI()


class ItemBase(BaseModel):
    name: str


class ItemFoo(ItemBase):
    foo: str


class ItemBar(ItemBase):
    bar: UUID


T = TypeVar("T", bound=ItemBase)


def register_post_route(
    router: APIRouter,
    path: str,
    schema: type[T],
) -> None:
    async def get_item(
        item: Annotated[  # type: ignore[valid-type]
            schema,  # for validation
            Body(),  # for location
        ],
    ) -> T:
        return item

    @router.post(
        path,
        response_model=schema,
        summary=f"create {schema.__name__}",
    )
    def create_item(
        item: Annotated[
            ItemBase,
            Depends(get_item),
        ],
    ) -> Any:
        print("item's name:", item.name)
        return item


items_router = APIRouter(
    prefix="/items",
    tags=["items"],
)
register_post_route(items_router, "/foo", ItemFoo)
register_post_route(items_router, "/bar", ItemBar)
app.include_router(items_router)
```

</details>


But why I still think this feature is useful: with current version of FastAPI we already can do same thing with the query string, so I'd like to add a way to extend this and allow developer to explicitly state that data should be read from the desired location.

Here's how we do it with the query string:

<details>

<summary>Code (Query parsing example, FastAPI 0.136.1)</summary>

```python

from typing import Annotated, Any

from fastapi import APIRouter, Depends, FastAPI

app = FastAPI()


class FilterBase(BaseModel):
    q: str


class PaginatedFilter(FilterBase):
    page: int = 1
    page_size: int = 10


def register_filter_route(
    router: APIRouter,
    path: str,
    schema: type[FilterBase],
) -> None:
    @router.get(path)
    def filter_items(
        filter_query: Annotated[
            FilterBase,
            Depends(schema),
        ],
    ) -> Any:
        print(f"filter's q: {filter_query.q!r}")
        return {
            "filter": filter_query.model_dump(),
        }


items_router = APIRouter(
    prefix="/items",
    tags=["items"],
)
register_filter_route(items_router, "/qwerty", PaginatedFilter)
app.include_router(items_router)
```

</details>

The example above shows, that FastAPI already supports providing the desired schema in runtime.

So updating this:

```python
# processes query string
    def filter_items(
        filter_query: Annotated[
            FilterBase,
            Depends(schema),
        ],
    ):
```

to this:
```python
# processes form
    def filter_items(
        filter_query: Annotated[
            FilterBase,
            Form(),  # added this
            Depends(schema),
        ],
    ):
```

will be a lot more convenient for dev than updating the signature - only one line added.

It already follows FastAPI's way to explicitly state location:

```python
# the already supported way to explicitly state the location
    async def get_user(
        item: Annotated[
            UserSchema,  # for validation
            Body(),  # for location
        ],
    ):
```

So I'd like to combine both: `Depends(schema)` to provide validation and `Body()` / `Form()` / `File()` / `Query()` to explicitly state where to look for data.

YuriiMotov:
> The example above shows, that FastAPI already supports providing the desired schema in runtime.
> 
> So updating this:
> 
> ```python
> # processes query string
>     def filter_items(
>         filter_query: Annotated[
>             FilterBase,
>             Depends(schema),
>         ],
>     ):
> ```

Using Pydantic models with `Depends` is not officially supported, and it's not recommended. 

Take a look at the following code as an example:

<details>

```py
from typing import Annotated

import pytest

app = FastAPI()


class Item(BaseModel):
    name: str
    tags: list[str] = []


@app.get("/query-model")
def get_item_query_model(item: Annotated[Item, Query()]):
    return item


@app.get("/depends")
def get_item_depends(item: Annotated[Item, Depends(Item)]):
    return item


@pytest.mark.parametrize(
    "endpoint",
    [
        "/query-model",
        "/depends",
    ],
)
def test_get_item(endpoint: str):

    client = TestClient(app)
    response = client.get(f"{endpoint}?name=TestItem&tags=tag1&tags=tag2")
    assert response.json() == {"name": "TestItem", "tags": ["tag1", "tag2"]}

```

</details>

I think combining `Body` \ `Query` \ `Form` ... and `Depends` in one annotation looks confusing. Depends is not intended to "explicitly state that data should be read from the desired location". `Depends` accepts a callable, inspects its signature and treats parameters of callable as request parameters \ sub-dependencies.

We could probably try implementing it differently, like:

```py
item: Annotated[BaseSchema, Body(input_model=UserSchema)],
```

IMO, that would be cleaner.
But first we need to understand how common this use case is. If it's not something people often struggle with, and considering there is already a way to implement this, it will unlikely to be added to FastAPI.
Hope you will understand, we have a lot of things to work on and we have to set priorities.

**PS:** the approach with `# type: ignore[valid-type]` is fine unless you want it to work with different type checkers. For example, `ty` will error on it and you will need to add a couple of more ignore comments.

mahenzon:
@YuriiMotov thank you.

I understand now that `Depends(schema)` is not recommended way. 

Yeah, doing something like `Body(input_schema=schema)` looks even better than my proposed solution and a lot better than a separate function with `type: ignore`.

Do I need to work on this implementation now? How can we understand how common this use case is?

YuriiMotov:
> Do I need to work on this implementation now? How can we understand how common this use case is?

I think the best way is to open a discussion to request this feature and gather community feedback.
As for this PR, I would close it for now. If the idea gets a community support, it will be better to open new PR (to keep the commit history and comments clean and easy to review)

mahenzon:
> I think the best way is to open a discussion to request this feature and gather community feedback. As for this PR, I would close it for now. If the idea gets a community support, it will be better to open new PR (to keep the commit history and comments clean and easy to review)

OK, thank you.

tiangolo:
Thanks for the interest, but I'll pass on this. :coffee:

PR #15480, opened 2026-05-04
📝 Add diagram to "Multiple middleware execution order" docs

AaronTimony:
Updated docs for execution order section in middleware. Thought it was a bit confusing without a diagram so I added one to make it clearer. Matches the styling of other diagrams throughout the documentation.


YuriiMotov:
I agree it's better to illustrate it with diagram!

I think 2 middlewares would be enough, and we should move them a bit up (to replace lines 85-87)

YuriiMotov (approved):
LGTM!

@AaronTimony, thanks!
Passing it to Sebastian for final review

PR #15481, opened 2026-05-04
Fix ForwardRef resolution inside Annotated in get_typed_annotation

siddharth10ss:
### Problem

When using `from __future__ import annotations`, type annotations are stored as strings (ForwardRef). FastAPI resolves forward references when the annotation itself is a string, but fails when the ForwardRef is wrapped inside `typing.Annotated`.

Example:

```python
from typing import Annotated
from fastapi import Depends

@app.get("/")
async def read_root(potato: Annotated["Potato", Depends(get_potato)]):
    ...
```

In this case, `"Potato"` is not resolved correctly, which leads to incorrect dependency parsing and incorrect OpenAPI schema generation.

---

### Root Cause

`get_typed_annotation()` only resolves forward references when the annotation is directly a `str` or `ForwardRef`. It does not handle ForwardRefs nested inside `Annotated`.

---

### Solution

* Detect `Annotated` types using `get_origin`
* Extract base type and metadata via `get_args`
* Resolve ForwardRef for the base type (handles both `str` and `ForwardRef`)
* Recursively resolve nested annotations
* Reconstruct `Annotated` using `__class_getitem__` to preserve correct structure

---

### Changes

* Added ForwardRef resolution for base types inside `Annotated`
* Added recursive handling for nested `Annotated` cases
* Preserved existing behavior for non-Annotated types
* Ensured consistent handling of `NoneType`

---

### Tests

Added test to validate:

*  `Annotated[Potato, Depends(...)]` (existing behavior unchanged)
*  `Annotated["Potato", Depends(...)]` (ForwardRef correctly resolved)
*  OpenAPI schema generation does not leak unresolved ForwardRef strings

---

### Notes

* Change is minimal and localized to `get_typed_annotation()`
* No breaking changes introduced
* Compatible with existing type resolution flow

---

Closes #13056

### DIFF
--- a/tests/test_annotated_forwardref.py
+++ b/tests/test_annotated_forwardref.py
@@ -1,0 +1,57 @@
+# IMPORTANT: This future import MUST be at the top of the module.
+# It is what causes annotations to be stored as strings (ForwardRef),
+# which is exactly the bug scenario described in issue #13056.
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Annotated
+
+from fastapi import Depends, FastAPI
+from fastapi.testclient import TestClient
+
+
+# ── App setup ─────────────────────────────────────────────────────────────────
+
+
+app = FastAPI()
+
+
+@dataclass
+class Potato:
+    color: str
+    size: int
+
+
+def get_potato() -> Potato:
+    return Potato(color="red", size=10)
+
+
+# The annotation "Potato" is a string here because of `from __future__ import
+# annotations`.  Before the fix, FastAPI would fail to resolve it when it is
+# wrapped inside Annotated[...].
+@app.get("/")
+def read_root(potato: Annotated[Potato, Depends(get_potato)]) -> dict:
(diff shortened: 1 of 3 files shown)


tests/test_annotated_forwardref.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_annotated_forwardref_openapi_schema() -> None:
+    """OpenAPI schema must be generated without errors and must not leak ForwardRef."""
+    response = client.get("/openapi.json")
+    assert response.status_code == 200
+    schema_text = response.text
+    assert "ForwardRef" not in schema_text, (
+        "Unresolved ForwardRef leaked into the OpenAPI schema"
+    )
```
YuriiMotov, line 56: These tests pass on `master`

YuriiMotov:
Let's keep it as draft until CI passes.
Also, please, review other existing PRs linked to the issue to ensure your approach is not a duplicate of existing PR

PR #15496, opened 2026-05-09
📝 Clarify "Required, can be None" query parameter docs

Sanskar121543:
## Summary

Clarifies the wording in the "Required, can be `None`" section of the query parameter validation docs.

The previous wording implied that clients could directly send Python `None` values in HTTP query parameters, which can be confusing because query parameters are transmitted as strings.

This update clarifies that:

* the parameter type can allow `None`
* the parameter can still be required
* HTTP query parameters themselves are string-based

## Changes

* Updated wording in:

  * `docs/en/docs/tutorial/query-params-str-validations.md`

## Related

Related to issue #12419


YuriiMotov:

PR #15499, opened 2026-05-10
15498: Enable Pydantic polymorphic_serialization flag on response serialization

valentinDruzhinin:
## Feature: Polymorphic Response Model Serialization

Adds support for `response_model_polymorphic_serialization` parameter to FastAPI path operations, enabling polymorphic serialization when returning subclass instances through base-class-typed response models.
Implements: #15498

### What it does:
- **Default behavior (False)**: When returning a subclass instance, only base class fields are serialized in the response
- **With polymorphic flag (True)**: Subclass-specific fields are included in the serialized response

### Example:
```python
class User(BaseModel):
    name: str

class UserLogin(User):
    password: str

@app.get("/user", response_model=User, response_model_polymorphic_serialization=True)
def get_user():
    return UserLogin(name="pydantic", password="secret")
    # Response includes: {"name": "pydantic", "password": "secret"}


YuriiMotov (commented):
I think all these flags that modify endpoint output making it different from what is declared in OpenAPI schema is bad idea and we shouldn't add more such flags.
There is already way to achieve this - by returning Response explicitly (as it's shown in the discussion).
So, I think we should close this PR

PR #15505, opened 2026-05-12
Improve wording in virtual environments documentation

shashankcodes1903:
Improved wording for better readability and clarity for beginners.


YuriiMotov:
Thanks for your interest, but I think current version is fine and there is no need to update it

PR #15506, opened 2026-05-12
Add OpenAPI callback tutorial example

aruldhanasekar:


aruldhanasekar:
All technical checks are now passing successfully. The remaining failing check appears to come from the automatically assigned docs and internal labels conflicting with the repository label checker.

YuriiMotov:

I think we don't need to add this example - current docs already describe well how to use it.
Also, your code example is not full - it doesn't document the callback schema.

I can see several things related to callbacks that can be improved, but this is not a top priority at the moment. So, let's close this for now.
Thanks again!

PR #15509, opened 2026-05-12
Add tests for SSE (EventSourceResponse) with yield dependencies

itsaryanchauhan:
## Summary
Adds `tests/test_dependency_after_yield_sse.py` - equivalent coverage for `EventSourceResponse` (SSE) to match what `test_dependency_after_yield_streaming.py` provides for `StreamingResponse`.

`EventSourceResponse` uses a distinct internal code path (`_sse_producer_cm` async context manager in `routing.py`), so a parallel test file is needed for honest regression coverage.

## Tests Added (12 tests)
| Test | What it verifies |
|------|-----------------|
| `test_regular_no_stream` | Baseline - yield dep works on a normal endpoint |
| `test_sse_simple` | SSE machinery with a static generator, isolated from dep logic |
| `test_sse_sync_streams_items` | Sync yield dep streams correct items |
| `test_sse_sync_dependency_cleaned_up` | Sync dep `finally` block runs after stream ends |
| `test_sse_async_streams_items` | Async yield dep streams correct items |
| `test_sse_async_dependency_cleaned_up` | Async dep `finally` block runs after stream ends |
| `test_sse_broken_sync_raises` | Broken sync dep raises correctly |
| `test_sse_broken_sync_no_raise` | Broken sync dep: 200 sent, body is empty |
| `test_sse_broken_async_raises` | Broken async dep raises correctly |
| `test_sse_broken_async_no_raise` | Broken async dep: 200 sent, body is empty |
| `test_sse_producer_raises_cleanup_runs` | Cleanup runs even when producer raises mid-stream |
| `test_lifo_cleanup_order` | Chained deps tear down in LIFO order |

All 12 tests pass locally on Python 3.11.14 / FastAPI 0.136.1 / Pydantic 2.13.2.

## Related
Refs discussion: https://github.com/fastapi/fastapi/discussions/15494


itsaryanchauhan:
The label check is failing because the labeler config doesn't have a rule for tests/ - nothing in .github/labeler.yml matches a test-only PR, so no label gets applied automatically. A maintainer will need to add one manually.
  
Latest commit: the broken async session was checking open before yielding, which meant the yield item line inside the async for loop was never actually reached, so coverage failed. Flipped it to check after yielding instead, which also more closely mirrors how a real session/cursor would behave (fetch the item, then check if the connection is still valid).

golikovichev:
Thanks for the credit, Aryan. Glad the LIFO cleanup test made the cut - that one is the easiest to assume and hardest to verify without an explicit case. Will go through the rest of the PR this week and leave detailed notes inline.

tests/test_dependency_after_yield_sse.py
```diff
@@ -0,0 +1,299 @@
...
+    finally:
+        app.dependency_overrides.clear()
+
+    assert len(sessions) == 1
+    assert sessions[0].open is False
+
+
+def test_sse_broken_sync_raises():
+    """When a sync yield dependency is broken the stream fails."""
+    with pytest.raises((ValueError, Exception)):
```
golikovichev, line 209: nit: `pytest.raises((ValueError, Exception))` will pass on any error type because `Exception` is the parent class. If you intend to lock the failure to the session closing, `pytest.raises(ValueError, match="Session closed")` makes the contract explicit and stops unrelated regressions from sneaking past this test. Same applies to the async sibling on line 226.

tests/test_dependency_after_yield_sse.py
```diff
@@ -0,0 +1,299 @@
...
+
+def test_sse_broken_sync_no_raise():
+    """
+    When a sync yield dependency raises after streaming has started,
+    the 200 status code is already sent but the body is empty.
+    """
+    with TestClient(app, raise_server_exceptions=False) as c:
+        response = c.get("/sse-broken-sync")
+    assert response.status_code == 200
+    assert _parse_sse_data_lines(response.text) == []
```
golikovichev, line 221: Question on the strict zero. Sync `Session.__iter__` checks `self.open` before yielding, so the broken-sync path never delivers any item. Async `_async_iter` yields first then checks, which could in principle let one item through before the close raises. If both paths are meant to deliver exactly zero on a broken dep, this is correct as written. If the contract is "fewer than 3 (any failed dep counts)", `assert len(_parse_sse_data_lines(response.text)) < 3` would survive an async timing shift. A short comment locking the intent would help.

tests/test_dependency_after_yield_sse.py
```diff
@@ -0,0 +1,299 @@
...
+    app.dependency_overrides[dep_session] = tracking_dep
+    try:
+        with TestClient(app, raise_server_exceptions=False) as c:
+            c.get("/sse-producer-raises")
+    finally:
+        app.dependency_overrides.clear()
+
+    assert len(sessions) == 1
+    assert sessions[0].open is False
+
```
golikovichev, line 272: suggestion: a sibling test where `gen()` raises before any yield (not after the first item) would exercise a different code path. Dep cleanup must still run when no item ever crossed the SSE boundary. Coverage gap rather than a bug. Can send a follow-up commit if useful.

golikovichev (commented):
Three inline notes from my walk through the broken-dep and producer-raises tests. Two test-quality suggestions and one coverage-gap idea. Happy to send a small follow-up PR for the producer-raises sibling if useful.

tiangolo:
I don't think this is necessary right now, thanks! :coffee:

itsaryanchauhan:
Thanks for taking a look @tiangolo ! I had opened this after the related discussion seemed positive, but I understand it’s not something the project needs right now. I’ll leave it closed. Appreciate the time.

PR #15514, opened 2026-05-13
Docs: Clarify required nullable query parameter behavior

Shaan-alpha:
### Description

This PR clarifies the "Required, can be None" documentation section for query parameters.

As discussed in #12419, the previous wording was slightly ambiguous as it implied clients could natively send a "None" value in a URL query string. This update explicitly clarifies that:
1. Query parameters are extracted as strings.
2. Omitting the parameter raises a "required" validation error.
3. Sending it empty (e.g., "?q=") results in an empty string "", rather than "None".
4. The "str | None" required pattern is primarily useful for JSON request bodies where "null" can be explicitly passed.

Fixes #12419

### Checklist

* [x] I added a very descriptive title to this PR.
* [x] I used the GitHub search to find a similar PR and didn't find it.


YuriiMotov:

PR #15516, opened 2026-05-14
Add warning about BackgroundTasks with direct Response

rajkumar-prog:
## Summary

This PR adds a warning to the `Return a Response Directly` documentation about using injected `BackgroundTasks` together with a directly returned `Response` that has its own `background` task.

The note explains that the background task attached to the returned `Response` is used, so users should keep background work in the same `BackgroundTasks` object or attach a single combined background task to the response.

## Context

This clarifies the behavior discussed in #11215.

## Changes

- Adds a warning block to `docs/en/docs/advanced/response-directly.md`
- No code behavior changes

## Verification

Documentation-only change.


YuriiMotov:
Thanks!

This is a duplicate of https://github.com/fastapi/fastapi/pull/15217

PR #15517, opened 2026-05-14
docs: clarify that required query params cannot be None in #12419

Dharanidhar28:
This PR updates the 'Required, can be None' section in the Query Parameters tutorial. It clarifies that while the syntax is valid in Python, it is not practically achievable via URL query parameters because missing keys fail the requirement and present keys are parsed as strings. Fixes #12419


YuriiMotov:

PR #15521, opened 2026-05-15
🌐 Update translations for ru (update-outdated)

tiangolo:
🌐 Update translations for ru (update-outdated)

This PR was created automatically using LLMs.

It uses the prompt file https://github.com/fastapi/fastapi/blob/master/docs/ru/llm-prompt.md.

In most cases, it's better to make PRs updating that file so that the LLM can do a better job generating the translations than suggesting changes in this PR.

PR #15534, opened 2026-05-15
📝 Fix image URLs in `index.md`

YuriiMotov:
Relative URLs don't work for translations, they point to `/es/img/logos/microsoft.svg` instead of `/img/logos/microsoft.svg`:

image

Adding leading slash fixes it. See docs preview for RU translation PR: https://github.com/fastapi/fastapi/pull/15521


tiangolo (approved):
Good catch, thank you! 🙌

PR #15536, opened 2026-05-16
Handle ImportError for optional JSON response imports

### LINKED ISSUE #15535: Broken optional orjson/ujson install can break importing fastapi.responses
### First Check

- [x] I already checked if it is not related to FastAPI but to Pydantic.
- [x] I already checked if it is not related to FastAPI but to Swagger UI.
- [x] I already checked if it is not related to FastAPI but to ReDoc.

### Commit to Help

- [x] I commit to help with one of those options.

### Example Code

```python
import sys, importlib

sys.path.insert(0, "/path/to/fastapi")

real_import_module = importlib.import_module

def fake_import_module(name, package=None):
    if name == "ujson":
        raise ModuleNotFoundError("No module named 'ujson'")
    if name == "orjson":
        raise ImportError("simulated binary/load failure")
    return real_import_module(name, package)

importlib.import_module = fake_import_module

import fastapi.responses
```

### Description

Recent `fastapi.responses` code catches `ModuleNotFoundError` when importing optional JSON response encoders:

```python
try:
    ujson = cast(_UjsonModule, importlib.import_module("ujson"))
except ModuleNotFoundError:
    ujson = None

try:
    orjson = cast(_OrjsonModule, importlib.import_module("orjson"))
except ModuleNotFoundError:
    orjson = None
```

This handles the package-absent case, but it no longer handles the optional C-extension failure case where the module exists but cannot load its binary extension or a platform dependency. In that case, `importlib.import_module("orjson")` or `importlib.import_module("ujson")` can raise a plain `ImportError`, and importing `fastapi.responses` fails even if the user only needs `JSONResponse`, `HTMLResponse`, etc.

I noticed this while reviewing recent changes around free-threaded Python 3.14t support and the `responses.py` optional import cleanup.

Expected behavior: a broken optional `orjson` or `ujson` install should not break importing the general response module. It should behave like the package is unavailable until the deprecated optional response class is actually used.

Suggested fix: catch `ImportError` for these optional imports, as the previous direct imports did, or catch both `ModuleNotFoundError` and `ImportError`.

Suggested regression test: monkeypatch `importlib.import_module` to raise `ImportError` for `orjson`, then assert `import fastapi.responses` succeeds.

### Operating System

Windows

### Operating System Details

Windows 11, local repo checkout

### FastAPI Version

Current `master` at `ecace740`

### Python Version

Python 3.12.7

### Additional Context

Related recent PR reviewed: `#15149` (`4b264878`), "Support free-threaded Python 3.14t".

cyphercodes:
## Summary

- Catch `ImportError` when importing optional `ujson` and `orjson` modules for response classes.
- Treat optional JSON encoder load failures the same as missing optional packages at import time.
- Add import-time regression coverage for plain `ImportError` from both optional imports.

Fixes #15535

## Tests

- `python -m pytest tests/test_responses_import.py -q`
- `python -m pytest tests/test_responses_import.py tests/test_deprecated_responses.py tests/test_orjson_response_class.py -q`
- `git diff --check origin/master..HEAD`

### DIFF
--- a/fastapi/responses.py
+++ b/fastapi/responses.py
@@ -26,13 +26,13 @@ class _OrjsonModule(Protocol):
 
 try:
     ujson = cast(_UjsonModule, importlib.import_module("ujson"))
-except ModuleNotFoundError:  # pragma: nocover
+except ImportError:  # pragma: nocover
     ujson = None  # type: ignore[assignment]
 
 
 try:
     orjson = cast(_OrjsonModule, importlib.import_module("orjson"))
-except ModuleNotFoundError:  # pragma: nocover
+except ImportError:  # pragma: nocover
     orjson = None  # type: ignore[assignment]
 
 
--- a/tests/test_responses_import.py
+++ b/tests/test_responses_import.py
@@ -5,7 +5,6 @@ from pathlib import Path
 
 import pytest
 
-
 REPO_ROOT = Path(__file__).resolve().parents[1]
 
 
--- a/tests/test_responses_import.py
+++ b/tests/test_responses_import.py
@@ -1,0 +1,40 @@
+import subprocess
+import sys
+import textwrap
+from pathlib import Path
+
+import pytest
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+@pytest.mark.parametrize("optional_module_name", ("ujson", "orjson"))
+def test_responses_imports_when_optional_json_import_raises_import_error(
+    optional_module_name: str,
+) -> None:
+    code = textwrap.dedent(
+        f"""
+        import importlib
+
+        real_import_module = importlib.import_module
+
+        def fake_import_module(name, package=None):
+            if name == {optional_module_name!r}:
+                raise ImportError("simulated optional dependency load failure")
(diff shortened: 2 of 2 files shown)


YuriiMotov:
I think we don't need this. See https://github.com/fastapi/fastapi/discussions/15541#discussioncomment-16938740

SreeTatikonda:
I tested this locally on macOS with Python 3.13.

Steps:

```bash
git clone https://github.com/fastapi/fastapi.git
cd fastapi
python -m pip install -e ".[all]"
python -m pip install pytest pytest-timeout pytest-xdist
python -m pytest tests/test_deprecated_responses.py tests/test_orjson_response_class.py -q
```

On my environment all tests in these modules were collected and (as expected) skipped due to missing optional `orjson` / `ujson` dependencies, and there were no errors during collection or execution.

The change to handle plain `ImportError` for optional JSON response imports looks correct and is consistent with how FastAPI handles other optional dependencies.

tiangolo:
It seems the current version is the most precise, there could be other types of errors during import, this one checks exactly that a module was not found. So I'll keep this as it is. Thanks! :coffee:

PR #15537, opened 2026-05-16
Fix optional orjson/ujson import breaking fastapi.responses

### LINKED ISSUE #15535: Broken optional orjson/ujson install can break importing fastapi.responses

amanv0007:
Fixes tiangolo/fastapi#15535

### DIFF
--- a/fastapi/responses.py
+++ b/fastapi/responses.py
@@ -61,7 +61,7 @@ class UJSONResponse(JSONResponse):
     separately, e.g. `pip install ujson`.
     """
 
-    def render(self, content: Any) -> bytes:
+    def render(self, content: Any) -> bytes:  # pragma: nocover
         assert ujson is not None, "ujson must be installed to use UJSONResponse"
         return ujson.dumps(content, ensure_ascii=False).encode("utf-8")
 
@@ -91,7 +91,7 @@ class ORJSONResponse(JSONResponse):
     separately, e.g. `pip install orjson`.
     """
 
-    def render(self, content: Any) -> bytes:
+    def render(self, content: Any) -> bytes:  # pragma: nocover
         assert orjson is not None, "orjson must be installed to use ORJSONResponse"
         return orjson.dumps(
             content, option=orjson.OPT_NON_STR_KEYS | orjson.OPT_SERIALIZE_NUMPY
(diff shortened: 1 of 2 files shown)


YuriiMotov:
This is the same approach as in https://github.com/fastapi/fastapi/pull/15536
Let's close this and proceed with #15536 as it was first

PR #15549, opened 2026-05-18
docs: clarify required nullable query parameters

nisark786:
## Summary

Clarify the docs for required query parameters that can be `None` by explaining that query parameters are sent as URL strings and there is no standard way to send a Python `None` value in a query parameter.

This addresses the confusion raised in #12419.

## Checks

- `uv run pytest tests/test_application.py -q`
- `uv run python scripts/docs.py build-lang en`
- `git diff --check`


YuriiMotov:
Please, review existing PRs linked to the issue (including closed ones) before opening new PR to avoid creating duplicates

PR #15556, opened 2026-05-18
fix: use tuple comparison for python-multipart version check

rmotgi1227:
## Summary

`ensure_multipart_is_installed()` checks `assert __version__ > "0.0.12"` using **string comparison**, which is lexicographic. This means `"0.0.100" > "0.0.12"` evaluates to `False` (because `"1"` < `"9"` by character code), so any installation of `python-multipart >= 0.0.100` would be incorrectly rejected as too old, raising `multipart_not_installed_error` on every `Form()` or `File()` endpoint.

## Fix

Parse the version string into an integer tuple before comparing, so numeric ordering is used:

```python
# BEFORE
assert __version__ > "0.0.12"

# AFTER
assert tuple(int(x) for x in __version__.split(".")[:3]) > (0, 0, 12)
```

This is a one-line change with no new imports. It correctly handles versions like `0.0.100`, `0.1.0`, `1.0.0`, and pre-release suffixes (the `[:3]` slice drops any trailing suffix components that aren't integers).

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -96,7 +96,7 @@ def ensure_multipart_is_installed() -> None:
         from python_multipart import __version__
 
         # Import an attribute that can be mocked/deleted in testing
-        assert __version__ > "0.0.12"
+        assert tuple(int(x) for x in __version__.split(".")[:3]) > (0, 0, 12)
     except (ImportError, AssertionError):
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Thanks for your interest!

We already have https://github.com/fastapi/fastapi/pull/15226

PR #15568, opened 2026-05-19
docs: polish English wording in help-fastapi page

MatiasFormoso:
Improves English wording in the help-fastapi documentation:

- 'By doing it' -> 'By doing so' (watch releases section)
- 'consume time the most' -> 'take up the most time' (maintainer tasks section)

Docs-only change; preview checked via GitHub Actions.


YuriiMotov:
Thanks, but I think we don't need to change it - it's already clear

PR #15570, opened 2026-05-19
fix: catch ImportError for optional ujson/orjson imports

pintaste:
Some environments raise `ImportError` rather than `ModuleNotFoundError` when optional dependencies like `ujson` or `orjson` are not installed. Since `ModuleNotFoundError` is a subclass of `ImportError`, catching only `ModuleNotFoundError` misses these cases.

This PR changes the exception handling in `fastapi/responses.py` to catch `ImportError` directly, which covers both `ImportError` and its subclass `ModuleNotFoundError`.

## Changes

- `fastapi/responses.py`: changed `except ModuleNotFoundError` → `except ImportError` for both `ujson` and `orjson` optional imports

Fixes #15536

### DIFF
--- a/fastapi/responses.py
+++ b/fastapi/responses.py
@@ -26,13 +26,13 @@ class _OrjsonModule(Protocol):
 
 try:
     ujson = cast(_UjsonModule, importlib.import_module("ujson"))
-except ModuleNotFoundError:  # pragma: nocover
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Thanks for your interest!

This is a duplicate of https://github.com/fastapi/fastapi/pull/15536

PR #14939, opened 2026-02-18
📝 Update "Union or anyOf" section in `docs/en/docs/tutorial/extra-models.md`

lkkmpn:
#14932 updated the source example on (among others) the [Extra Models](https://fastapi.tiangolo.com/tutorial/extra-models/) docs page, but the text still refers to the old example.

PR #15576, opened 2026-05-20
Fix incorrect statement about response_model union syntax in Python 3.10+

LoliPix:
The current wording does not clearly distinguish Python version behavior, which may mislead users on Python 3.10+ into thinking the syntax is invalid.


YuriiMotov:

We already have https://github.com/fastapi/fastapi/pull/14939

PR #15579, opened 2026-05-21
perf: skip redundant deepcopy for immutable defaults and default_factory params

navidnabavi:
## Summary

While auditing `deepcopy` usage I found two per-request call sites in `dependencies/utils.py` (`_validate_value_with_model_field` and `_get_multidict_value`) that unconditionally deepcopy `field.default`. This is correct for mutable defaults like `default=[]`, but redundant in two cases:

**1. Immutable defaults** (`None`, `int`, `str`, `bytes`, `bool`)
These cannot be mutated — `deepcopy(None)` wastes ~250 ns per missing parameter per request for zero benefit.

**2. `default_factory` parameters**
FastAPI already supports `default_factory` on `Query`, `Path`, `Body` etc. The factory produces a **fresh object on every call** — that's its entire purpose. The validation path was unaware of this and deepcopied the factory output anyway.

## Changes

- Add `_IMMUTABLE_TYPES` constant; skip `deepcopy` for primitive defaults
- If `field.field_info.default_factory is not None`, call `field.get_default()` (routes through Pydantic's `FieldInfo.get_default(call_default_factory=True)`) and return directly — no copy needed
- Tests: isolation correctness + mock assertions that `deepcopy` is not called in both optimized paths

## Benchmark

50,000 requests × 5 runs, `TestClient`, local machine:

| Scenario | before | after | delta |
|---|---|---|---|
| `None` / `int` / `str` default | ~360 µs | ~355 µs | ~0% (noise) |
| `list default=[]` | 394 µs | 368 µs | ~0% (noise) |
| `list default_factory=list` | **586 µs** | **402 µs** | **−31%** |

The `default_factory` improvement is the meaningful one: factory + deepcopy = two allocations per request; factory alone = one.

## Test plan

- [x] All 3126 existing tests pass unchanged
- [x] New test: mutable `default=[]` still isolated between requests
- [x] New test: `deepcopy` not called for immutable defaults
- [x] New test: `default_factory` on `Query` gives fresh object per request
- [x] New test: `deepcopy` not called when `default_factory` is set

### Disclaimer
AI used for test generation and text edits.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -76,6 +76,9 @@ from starlette.responses import Response
 from starlette.websockets import WebSocket
 from typing_inspection.typing_objects import is_typealiastype
 
+# Primitive types whose instances are immutable — safe to return without deepcopy.
+_IMMUTABLE_TYPES = (int, float, str, bytes, bool, type(None))
+
 multipart_not_installed_error = (
     'Form data requires "python-multipart" to be installed. \n'
     'You can install "python-multipart" with: \n\n'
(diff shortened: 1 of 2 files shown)


navidnabavi:
I can also commit benchmark codes in /scripts if required.

navidnabavi:
I used AI in my development (in writing tests) but this perf optimization is valid.

PR #15592, opened 2026-05-24
📝 Add a note that Python 3.13+ venv creates `.gitignore` by default

ftnext:
## Pull Request

Discussion: 

I couldn't find a suitable Discussion category for this small docs-only clarification 🙏 

## Description

Add a tip to the virtual environments docs noting that Python 3.13+ creates `.venv/.gitignore` automatically when using `venv`.

Python 3.13's `venv` module creates SCM ignore files by default, including Git ignore files, so the manual `.gitignore` step is only needed for Python 3.12 and below.

This matches the existing `uv` tip, as `uv` also creates `.venv/.gitignore` when creating the virtual environment.

Reference: https://docs.python.org/3/whatsnew/3.13.html#venv

Context: noticed while checking `virtual-environments.md` after #15463 in 0.163.2.

## Checklist

- [ ] Coverage stays at 100%.


docs/en/docs/virtual-environments.md
```diff
@@ -268,6 +268,12 @@ If you are using **Git** (you should), add a `.gitignore` file to exclude everyt
 
 /// tip
 
+If you used Python 3.13 or newer, it already did this for you, you can skip this step. 😎
+
+///
+
+/// tip
+
```
YuriiMotov: I think it's better to add this to existing tip.

How about just adding the following to the existing tip?

```
`python -m venv` also creates a `.gitignore` file by default starting with Python 3.13.
```

ftnext: Thanks! That makes sense. I updated it to add the note to the existing tip.

YuriiMotov (approved):
LGTM!

YuriiMotov:
Obsolete by https://github.com/fastapi/fastapi/pull/16032

ftnext:
I understand that #16032 removed the relevant section from the FastAPI docs.
Since the linked detailed virtual-environments guide still covers the `python -m venv` workflow, would this Python 3.13 clarification be appropriate there instead?

YuriiMotov:
> Since the linked detailed virtual-environments guide still covers the `python -m venv` workflow, would this Python 3.13 clarification be appropriate there instead?

I think it's not needed. This is not a preferable approach and I don't think people will be too confused if they go and see that this file is already at place.
So, I would keep it as it is

ftnext:
@tiangolo What concerns me is that the same Python 3.13 clarification proposed in this PR had already been included in https://github.com/tiangolo/tiangolo.com/pull/235 before this PR was closed, without any reference to this contribution or communication about the move.

I understand that not every pull request can or should be merged. I would have been completely fine with closing this PR after explaining that the documentation had been moved and the clarification incorporated there. **The lack of communication and acknowledgment** is what I find disappointing.

YuriiMotov:
@ftnext, I understand your disappointment. Sorry for that.
I don't know if Sebastian came to this edit independently or looked it up in this PR. If second, then he probably just forgot to leave a comment here - he is usually eager to acknowledge meaningful contributions.

And, my bad, I didn't even say thank you for your work on this PR. It's probably a consequence of the fact that we have to deal with so many PRs opened by soulless AI-bots.

So, let me try to fix this - **thank you for your efforts!**

PR #15596, opened 2026-05-24
[F55] Add trailing period to module docstring

JosefVacha:
Small fix to module docstring


YuriiMotov:
Thanks, but it's not important. Let's close this

PR #15597, opened 2026-05-24
[F56] Fix trailing period in SSE comment

JosefVacha:
Small comment fix in sse.py


YuriiMotov:
Thanks, but it's not important. Let's close this

PR #15515, opened 2026-05-14
🐛 Fix line splitting in `format_sse_event` to comply with SSE spec

Zawwarsami16:
Closes #15500.

## What's happening

`format_sse_event` in `fastapi/sse.py` uses `str.splitlines()` to split `data_str` (and `comment`) into SSE lines. `splitlines()` is wrong on two counts here:

1. **Drops trailing empty strings.** `"Hello\n".splitlines()` is `["Hello"]`, so the trailing `data:` line gets eaten — corrupting any payload that ends with a newline. This is the bug in the linked issue.
2. **Treats 8 extra characters as line breaks.** `\v`, `\f`, `\x1c`, `\x1d`, `\x1e`, `\x85`, U+2028, U+2029 — per the [Python docs](https://docs.python.org/3/library/stdtypes.html#str.splitlines). SSE only recognizes `\n`, `\r\n`, and `\r` per [the spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream), so a payload like `"A B"` was being silently split into two `data:` lines.

## The fix

A small helper, `_split_sse_lines`, normalizes `\r\n` and lone `\r` to `\n`, then splits on `\n`. That preserves trailing empties (because `split("\n")` keeps them), and only recognizes the three line terminators that SSE actually defines.

```python
def _split_sse_lines(value: str) -> list[str]:
    return value.replace(\"\\r\\n\", \"\\n\").replace(\"\\r\", \"\\n\").split(\"\\n\")
```

Both `data:` and `comment:` branches now go through the helper.

## Tests

8 new unit tests in `tests/test_sse.py` that directly exercise `format_sse_event`:

| Test | Input | Output |
|---|---|---|
| `preserves_trailing_newline` | `\"Hello\\n\"` | `b\"data: Hello\\ndata: \\n\\n\"` |
| `preserves_trailing_double_newline` | `\"Hello\\n\\n\"` | `b\"data: Hello\\ndata: \\ndata: \\n\\n\"` |
| `single_newline_data` | `\"\\n\"` | `b\"data: \\ndata: \\n\\n\"` |
| `crlf_normalizes_to_lf` | `\"Hello\\r\\nWorld\"` | `b\"data: Hello\\ndata: World\\n\\n\"` |
| `bare_cr_treated_as_line_break` | `\"Hello\\rWorld\"` | `b\"data: Hello\\ndata: World\\n\\n\"` |
[...]

PR #15608, opened 2026-05-26
Fix line splitting in format_sse_event to comply with SSE spec

Sujit-1509:
## Description

This resolves a protocol compliance issue with Server-Sent Events. Previously, `format_sse_event` used Python's `str.splitlines()` method, which incorrectly splits on multiple Unicode characters (like `\v` or `\x1e`).

The SSE specification strictly states that lines must be separated by `\r\n` (CRLF), `\n` (LF), or `\r` (CR) characters only. Replacing `splitlines()` with explicit `.replace("\r\n", "\n").replace("\r", "\n").split("\n")` ensures that valid data carrying strings with other splitting unicode markers do not corrupt the SSE stream framing.

Fixes #15515

### DIFF
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -206,14 +206,14 @@ def format_sse_event(
     lines: list[str] = []
 
     if comment is not None:
-        for line in comment.splitlines():
+        for line in comment.replace("\r\n", "\n").replace("\r", "\n").split("\n"):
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Duplicate of https://github.com/fastapi/fastapi/pull/15515

PR #15609, opened 2026-05-26
📝 Fix grammar: 'to login' -> 'to log in' in oauth2-jwt.md

aryanmotgi:
Small grammar fix in `docs/en/docs/tutorial/security/oauth2-jwt.md`.

`login` is a noun; the verb form is `log in`. The sentence reads more naturally as:

> And your users would be able to **log in** from your Django app or from your **FastAPI** app, at the same time.


YuriiMotov:
Thanks, but I think it's fine to use "login" as a verb in informal English unless it doesn't lead to misunderstanding.
Also, if we change it only here, there will still be other places where it's used as a verb.

PR #15613, opened 2026-05-26
📝 Fix `format_sse_event` docstring rendering of `\n\n` terminator

AshNicolus:
## Summary

The docstring for `fastapi.sse.format_sse_event` describes the SSE event terminator as `\n\n` inside backticks. Because the docstring is a regular Python string (not raw), `\n` is interpreted as a newline at parse time, so the rendered docstring — e.g. via `help(format_sse_event)` or any generated reference docs — shows two blank lines between empty backticks instead of the intended `\n\n` escape sequence.

Escaping the backslashes (`\\n\\n` in source) makes the rendered text match what was intended.

### Before
The result always ends with `
` (the event terminator).

### After
The result always ends with `\n\n` (the event terminator).

## Test plan

- [x] `python -c "from fastapi.sse import format_sse_event; print(format_sse_event.__doc__)"` shows the corrected rendering.
- [x] `ruff check fastapi/sse.py` passes.
- [x] `ruff format fastapi/sse.py --check` passes.
- [x] No code-path change — only the `__doc__` string content changes; runtime behavior of `format_sse_event` is identical.


YuriiMotov (approved):
LGTM!

Before:

Image

After:

Image

tiangolo (approved):
Ah, nice, thanks! ☕

PR #15616, opened 2026-05-27
✅ Use custom `changing_dir` instead of `CLIRunner.isolated_filesystem` to set working dir

YuriiMotov:
## Description

Typer 0.26 removed `CLIRunner.isolated_filesystem` context manager that was used in tests to set working directory to temporary path.
So, FastAPI redistribute tests started to fail (they run with latest dependency versions) - see [logs](https://github.com/fastapi/fastapi/actions/runs/26507260262/job/78062757815?pr=15610).

Here is where `isolated_filesystem` is used: https://github.com/fastapi/fastapi/blob/dbfd55cea3a656029a9368924f3d8f1f829702c9/scripts/tests/test_translation_fixer/conftest.py#L29

The fix is to use custom context manager as in `fastapi-cloud-cli`: https://github.com/fastapilabs/fastapi-cloud-cli/blob/1d807db95e43d598f74ddecb8e7ce107396a99fa/tests/utils.py#L11-L18

## AI Disclaimer

No AI used

## Checklist

- [x] Coverage stays at 100%.

### DIFF
--- a/scripts/tests/test_translation_fixer/conftest.py
+++ b/scripts/tests/test_translation_fixer/conftest.py
@@ -1,5 +1,8 @@
+import os
 import shutil
 import sys
+from collections.abc import Generator
+from contextlib import contextmanager
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Just to add, if we prefer to bring back `CLIRunner.isolated_filesystem` that's possible too. In Typer we could resolve all of its usages with `tmp_path` which is why I originally ended up removing it, but I don't feel strongly either way.

If we want to restore it, I can easily do so in https://github.com/fastapi/typer/pull/1792, just let me know 🙂

tiangolo (approved):
Thank you! 🚀 

---

And I personally agree with just dropping `isolated_filesystem`, I would think if people need that, it would make more sense to use `tmp_path` from pytest or similar, I don't really think the CLI should provide that utility as that can come from testing utilities. I also suspect it was added to Click before there were any other options, but for us is probably safe to not have to keep supporting it.

PR #15618, opened 2026-05-27
fix(sse): preserve empty data lines when formatting SSE

MatiasFormoso:
## Summary
- preserve explicit empty payloads in format_sse_event(data_str=...) by emitting a data: line
- normalize CRLF/CR to LF before splitting data lines for stable SSE wire output
- add focused tests for empty data and trailing-newline behavior

## Why
format_sse_event(data_str='') currently emits no data: line, which can drop an intended empty event payload. This change keeps empty and trailing lines explicit in the SSE output.

## Tests
- python -m pytest -c nul tests/test_sse.py -k format_sse_event -q
- python -m pytest -c nul tests/test_sse.py -q

### DIFF
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -88,7 +88,8 @@ class ServerSentEvent(BaseModel):
 
             Use this when you need to send pre-formatted text, HTML fragments,
             CSV lines, or any non-JSON payload. The string is placed directly
-            into the `data:` field as-is.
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@MatiasFormoso, thanks!
We already have PR https://github.com/fastapi/fastapi/pull/15515 that fixes this. The test case with empty string was missing, I suggested to add it in the comments of that PR.

I'll close this to avoid having duplicates

PR #15622, opened 2026-05-27
⚡ Deduplicate get_flat_dependant calls in OpenAPI schema generation

sebastianbreguel:
## Summary

`get_openapi_path` called `get_flat_dependant(route.dependant, skip_repeats=True)` three times per route — once directly, once inside `_get_openapi_operation_parameters`, and once inside `get_flat_params`. All three computed the identical result.

This PR hoists the computation before the method loop and passes the result to internal consumers, eliminating 2 redundant dependency-tree traversals per route.

## Benchmark

10–200 routes, each with 3 deps + path/query/header params, `timeit.repeat` min of 5×50, Python 3.11:

| Routes | Speed Δ | RAM Δ |
|---|---|---|
| 10 | −11% | ~0% |
| 50 | −6% | 0% |
| 100 | −7% | 0% |
| 200 | −10% | 0% |

## Test plan

- [x] Full test suite: 3128 passed, 0 failed

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -106,7 +106,7 @@ def get_openapi_security_definitions(
 
 def _get_openapi_operation_parameters(
     *,
-    dependant: Dependant,
+    flat_dependant: Dependant,
     model_name_map: ModelNameMap,
     field_mapping: dict[
         tuple[ModelField, Literal["validation", "serialization"]], dict[str, Any]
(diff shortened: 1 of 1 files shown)


YuriiMotov:
The PR template was recently updated, and it now asks to start from discussion and only open PR if it's requested by team member.
Please, stop opening PRs and explain your findings in the discussions instead.
Thank you for understanding!)

sebastianbreguel:
Moved to discussion: https://github.com/fastapi/fastapi/discussions/15623

PR #15629, opened 2026-05-28
⚡️ Replace `email.message` with string parsing for content-type detection

dennybiasiolli:
## Summary

Every incoming request with a JSON body currently allocates a full
`email.message.Message` object solely to check whether the `Content-Type`
header indicates JSON. The `email` module is designed for RFC 2822 email
parsing — far heavier than needed for a simple media-type check that runs on
every request.

This PR replaces that with a lightweight `_is_json_content_type()` helper using
basic string operations (`split`, `find`, `lower`), yielding a **~5× speedup**
on the content-type check path.

## Problem

In `fastapi/routing.py`, inside `get_request_handler`, the body-parsing block
did:

```python
message = email.message.Message()
message["content-type"] = content_type_value
if message.get_content_maintype() == "application":
    subtype = message.get_content_subtype()
    if subtype == "json" or subtype.endswith("+json"):
        json_body = await request.json()
```

This allocates an `email.message.Message` instance, sets a header, then calls
two methods that internally run the full email header parser — all to extract a
main type and subtype from a string like `application/json; charset=utf-8`.

## Solution

A new module-level helper `_is_json_content_type(content_type: str) -> bool`:

1. Splits on `";"` to strip parameters (`charset`, `boundary`, etc.)
2. Strips whitespace and lowercases for case-insensitive comparison
3. Splits on `"/"` to extract main type and subtype
4. Returns `True` only for `application/json` or `application/*+json`

The call site becomes:

```python
if _is_json_content_type(content_type_value):
    json_body = await request.json()
```

The `import email.message` is removed entirely — it has no other uses in the
module.

## Changes

| File | What changed |
|---|---|
| `fastapi/routing.py` | Removed `import email.message`; added `_is_json_content_type()`; replaced 5-line email parser block with 2-line call |
| `tests/test_is_json_content_type.py` | 26 parametrized unit tests — 12 accepted and 14 rejected content-type values |
| `tests/bench_content_type.py` | Standalone benchmark script comparing old vs new implementations (I can remove this and amend the commit, it's here only for convenience when reviewing the PR) |

## Benchmark

```
100,000 iterations × 10 content-types

  email.message (old)       1.100s  (  909,041 ops/s)
  string parsing (new)      0.221s  (4,533,336 ops/s)

  Speedup: 5.0× faster  (79.9% reduction in time)
```

Run with: `uv run python tests/bench_content_type.py`

## Test coverage

- **Unit tests** (`tests/test_is_json_content_type.py`): Covers standard JSON
  types, `+json` structured suffixes, case variations, charset parameters,
  whitespace, and a wide range of non-JSON types including edge cases like
  `application/geo+json-seq`, empty strings, and malformed values.
- **Existing integration tests**: All 37 pre-existing content-type related tests
  pass unchanged (`test_tutorial/test_body`, `test_strict_content_type_*`).

Run with: `uv run pytest tests/test_is_json_content_type.py tests/test_tutorial/test_body/test_tutorial001.py tests/test_tutorial/test_strict_content_type/ tests/test_strict_content_type_app_level.py tests/test_strict_content_type_nested.py tests/test_strict_content_type_router_level.py -v`

## Backwards compatibility

Fully backwards compatible. The new function produces identical results to the
`email.message.Message` approach for all content-type values — verified by the
benchmark script which asserts equivalence before timing.

Co-authored-by: Grok

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -1,5 +1,4 @@
 import contextlib
-import email.message
 import functools
 import inspect
 import json
@@ -348,6 +347,36 @@ def _build_response_args(
     return response_args
 
 
+def _is_json_content_type(content_type: str) -> bool:
+    """Check if a content-type header value indicates JSON content.
+
+    Matches ``application/json`` and ``application/*+json`` variants
+    (e.g. ``application/geo+json``, ``application/vnd.api+json``),
+    ignoring parameters like ``charset=utf-8``.
+
+    This replaces the previous ``email.message.Message`` approach which
+    carried significant per-request overhead.
+
+    >>> _is_json_content_type("application/json")
+    True
+    >>> _is_json_content_type("application/json; charset=utf-8")
+    True
+    >>> _is_json_content_type("application/geo+json")
+    True
+    >>> _is_json_content_type("text/plain")
+    False
+    """
+    media_type = content_type.split(";", 1)[0].strip().lower()
+    slash = media_type.find("/")
+    if slash == -1:
+        return False
+    main_type = media_type[:slash]
+    if main_type != "application":
+        return False
+    subtype = media_type[slash + 1 :]
+    return subtype == "json" or subtype.endswith("+json")
+
+
 def get_request_handler(
     dependant: Dependant,
     body_field: ModelField | None = None,
@@ -414,12 +443,8 @@ def get_request_handler(
                             if not actual_strict_content_type:
                                 json_body = await request.json()
                         else:
-                            message = email.message.Message()
-                            message["content-type"] = content_type_value
-                            if message.get_content_maintype() == "application":
-                                subtype = message.get_content_subtype()
-                                if subtype == "json" or subtype.endswith("+json"):
-                                    json_body = await request.json()
+                            if _is_json_content_type(content_type_value):
+                                json_body = await request.json()
                         if json_body != Undefined:
                             body = json_body
                         else:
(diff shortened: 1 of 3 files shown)


dennybiasiolli:
Info for reviewers: I think that `tests/bench_content_type.py` doesn't need to be added to the repo. It's there mainly to allow you to easily benchmark the changes, then I can remove it and amend my commit when/if you like

YuriiMotov:
There is also a [discussion](https://github.com/fastapi/fastapi/discussions/15201) that points to a related issue - `email.message.Message` doesn't fully comply with RFC 7231 wich HTTP follows.

So, maybe we should address both one day, but to get this accepted we will need to ensure it follows the spec (by explaining it in the description, adding tests and approving by trusted contributors).

But first, let's wait for Sebastian to take a look and validate the idea.

YuriiMotov:
Let's close this for now and add a comment in https://github.com/fastapi/fastapi/discussions/15201 (see [comment](https://github.com/fastapi/fastapi/discussions/15201#discussioncomment-17088325)) that points to this PR.

We are trying to keep things clean and not to have too many pending PRs. See https://tiangolo.com/open-source/contributing/ (and https://fastapi.tiangolo.com/contributing/ that points there)

dennybiasiolli:
Thanks @YuriiMotov , I saw #15201 but I though "same place in the code, but different issue" so I created this PR instead, sorry 😄 
I will follow the discussion and in case I can try and help with this

PR #15638, opened 2026-05-29
⚡️ Improve performance of OpenAPI schema generation (get_flat_dependant)

sebastianbreguel:
Follow-up to discussion #15623 (consolidates the findings posted there).

## What

Performance optimizations in the OpenAPI schema-generation path, with **no change to generated schema output**. ~10-15% off schema generation for typical apps, up to ~57% for wide dependency trees.

### 1. Deduplicate `get_flat_dependant` per route
`get_openapi_path` computed the flattened dependant three times per route (directly, inside `_get_openapi_operation_parameters`, and inside `get_flat_params`). Compute it once before the method loop and thread it down.

Benchmark (10-200 routes, 3 deps + path/query/header params each; `timeit.repeat` min of 5×50; Python 3.11; variance <2%):

| Routes | Speed | RAM |
|-------:|------:|----:|
| 10  | −11% | ~0% |
| 50  | −6%  | 0%  |
| 100 | −7%  | 0%  |
| 200 | −10% | 0%  |

### 2. `get_flat_dependant` internals
`visited` list → set (O(1) membership), skip visited-tracking entirely when `skip_repeats=False`, copy-on-write param lists (skip the `.copy()`/`.extend()` when empty), and a fast path that avoids the `oauth_scopes` allocation when there are no scopes.

Benchmark (synthetic trees; `timeit.repeat` min of 5×1000; `tracemalloc` peak over 100 runs; Python 3.11):

| Scenario | Speed | RAM |
|----------|------:|----:|
| Chain depth=20         | −14% | −14% |
| Chain depth=50         | −21% | −16% |
| Chain depth=100        | −28% | −12% |
| Wide 50×3 (200 nodes)  | −42% | −27% |
| Wide 100×3 (400 nodes) | −57% | −29% |
| Diamond width=20       | −20% | −25% |
| Diamond width=50       | −32% | −36% |
| Diamond width=100      | −39% | −31% |

Synthetic scenarios isolate the algorithmic improvement; typical apps (10-30 deps) see the lower end.

### 3. Scope dedup + cached `jsonable_encoder`
Dedup scopes with `dict.fromkeys` (O(1) membership, **insertion order preserved** → byte-identical output) and cache the `jsonable_encoder` result per security scheme so it runs once instead of once per referencing dependency.

Benchmark (50 scopes, 1-20 security deps; `timeit.repeat` min of 5×5000; Python 3.11):

| Metric | Before | After | Change |
|--------|-------:|------:|-------:|
| Speed | 42.7 µs | 39.5 µs | −7% |
| RAM   | 9.7 KB  | 9.7 KB  | 0%  |

<sub>Figure measured with the `set`-based dedup from #15623; the shipped version uses `dict.fromkeys` (order-preserving) with equivalent O(1) membership, so the number is representative — the dominant win is the encoder caching, which is unchanged.</sub>

## Output compatibility

No change to the generated OpenAPI schema, including OAuth2 scope ordering (insertion order is preserved, verified explicitly).

## Tests

Full suite passes (3148 passed, 14 skipped). `ruff` and `ty` clean. No API changes.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -139,22 +139,23 @@ def get_flat_dependant(
     dependant: Dependant,
     *,
     skip_repeats: bool = False,
-    visited: list[DependencyCacheKey] | None = None,
+    visited: set[DependencyCacheKey] | None = None,
     parent_oauth_scopes: list[str] | None = None,
 ) -> Dependant:
-    if visited is None:
-        visited = []
-    visited.append(dependant.cache_key)
-    use_parent_oauth_scopes = (parent_oauth_scopes or []) + (
-        dependant.oauth_scopes or []
-    )
+    visited_set: set[DependencyCacheKey] | None = None
+    if skip_repeats:
+        visited_set = set() if visited is None else visited
+        visited_set.add(dependant.cache_key)
+    if (
+        parent_oauth_scopes
+        or dependant.parent_oauth_scopes
+        or dependant.own_oauth_scopes
+    ):
+        use_parent_oauth_scopes = (parent_oauth_scopes or []) + dependant.oauth_scopes
+    else:
+        use_parent_oauth_scopes = []
 
     flat_dependant = Dependant(
-        path_params=dependant.path_params.copy(),
-        query_params=dependant.query_params.copy(),
-        header_params=dependant.header_params.copy(),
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Thanks!

No need to open PRs for now - we are planning a big code refactoring and will likely address this during that refactoring.

Please, only open PR if team member asked you to do it

PR #15645, opened 2026-05-30
✨ Add `parse_sse_events` for parsing SSE wire-format responses

AshNicolus:
## Summary

FastAPI exposes `fastapi.sse.format_sse_event(...)` — a public helper that **serializes** Server-Sent Events into wire format. There's no counterpart to **parse** an SSE response back into structured events, even though FastAPI's own test suite reaches for that pattern in 8+ places (manual `line.startswith("data: ")` splitting), and downstream users testing `EventSourceResponse` endpoints reinvent the same loop.

This PR adds the receiver-side counterpart:

* **`ParsedSSEEvent`** — small frozen `@dataclass` with `data`, `event`, `id`, `retry` fields. Mirrors `ServerSentEvent`, but represents events read off the wire (so `data` is the raw string from the wire, not JSON-decoded — decoding is up to the caller since payloads may be JSON, plain text, or anything else).
* **`parse_sse_events(raw: bytes | str) -> list[ParsedSSEEvent]`** — parses a complete event stream following the [WHATWG SSE parsing algorithm](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation):
  * UTF-8 BOM stripping
  * `\n`, `\r`, and `\r\n` line endings
  * Comment lines (`:` prefix) skipped
  * Multi-line `data:` fields joined with `\n` (with the single trailing `\n` stripped)
  * NULL-byte-containing `id` values ignored
  * Non-decimal `retry` values ignored
  * Unknown field names ignored
  * Empty-data events not emitted

Per-event `id` / `retry` are not sticky across events in the returned list — each `ParsedSSEEvent` reflects only fields explicitly seen in its own block. This matches the wire shape and keeps the parser stateless from the caller's perspective.

## Why this is useful

Round-trip equivalence with `format_sse_event` makes testing trivial:

```python
events = parse_sse_events(client.get("/sse").text)
assert events[0] == ParsedSSEEvent(data="hello", event="greeting", id="1")

### DIFF
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from typing import Annotated, Any
 
 from annotated_doc import Doc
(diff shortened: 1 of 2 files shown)


YuriiMotov:
I think that the functionality of SSE response parsing is out of scope for FastAPI project. It shouldn't be part of public API.
If you disagree - please, [start discussion](https://tiangolo.com/open-source/contributing/#start-with-a-discussion) and let's gather community feedback first

PR #15651, opened 2026-05-31
Fix: validate event and id fields in format_sse_event to prevent SSE protocol injection

hemanthvnp:
Closes the gap left by #15588.

`ServerSentEvent` already validates `event` and `id` for newlines and
null characters, but the public `format_sse_event()` function did not
apply the same checks, allowing SSE protocol injection when
user-controlled input is passed directly:

```python
format_sse_event(event="legit\ndata: injected", data_str="safe")
# → b'event: legit\ndata: injected\ndata: safe\n\n'

fastapi/sse.py — call the two existing validators at the top of
format_sse_event():


_check_event_single_line(event)
_check_id_valid(id)
(The helpers already guard against None, so no extra if wrappers
are needed.)

tests/test_sse.py — eight new parametrised tests covering \n,
\r, \r\n, and \0 for both event and id, plus an end-to-end
injection scenario. All 32 tests pass.

Discussed in: https://github.com/fastapi/fastapi/discussions/15648


hemanthvnp:
Could a maintainer please add the appropriate label so the check-labels CI passes? Thank you!

hemanthvnp:
Closing this as I opened it prematurely based on a non-team-member reply. Will re-open once a team member confirms in the discussion. Apologies for the process misstep.

PR #15676, opened 2026-06-01
Forward exclude_defaults when jsonable_encoder recurses into dict values

vidigoat:
### Summary

`jsonable_encoder` forwards `exclude_defaults` to its recursive calls for `list`/`tuple`/`set` items, but **omits it when recursing into `dict` keys and values**. So a Pydantic model nested inside a `dict` keeps its default-valued fields under `exclude_defaults=True`, while the same model inside a `list` correctly drops them:

```python

class M(BaseModel):
    foo: str
    bar: str = "bar"

m = M(foo="foo", bar="bar")
jsonable_encoder([m], exclude_defaults=True)        # [{'foo': 'foo'}]                      ✅
jsonable_encoder({"m": m}, exclude_defaults=True)   # {'m': {'foo': 'foo', 'bar': 'bar'}}    ❌ default kept
```

The sibling flags `exclude_unset` and `exclude_none` **are** already forwarded in that same `dict` branch, so this is a copy/paste inconsistency rather than intended behavior. (`include`/`exclude` are intentionally top-level-only and are correctly not forwarded.)

### Fix

Forward `exclude_defaults` in the two recursive `jsonable_encoder` calls inside the `dict` branch (for `encoded_key` and `encoded_value`), matching how the list/tuple/set branch already does it.

### Test plan

Added `test_encode_model_with_default_in_container` to `tests/test_jsonable_encoder.py`, asserting a model-in-dict behaves the same as model-in-list under `exclude_defaults` / `exclude_unset`.

```
# fails on master (dict keeps defaults), passes with this change:
python -m pytest tests/test_jsonable_encoder.py -q
# 27 passed
ruff check fastapi/encoders.py tests/test_jsonable_encoder.py   # All checks passed!
ruff format fastapi tests --check                                # clean
```

No behavior change for the already-correct list/tuple/set path; only the dict path is brought into line.

_Disclosure: authored with the assistance of an AI tool; fully reviewed and verified by me._


Deadshot1831:
I tested this locally against current `master` and this PR branch.

On `master`, `exclude_defaults=True` is applied correctly for models inside lists, but not for models inside dict values:

```text
list: [{'foo': 'foo'}]
dict: {'m': {'foo': 'foo', 'bar': 'bar', 'bla': 'bla'}}
```

On this PR branch, the dict case matches the list behavior:

```text
list: [{'foo': 'foo'}]
dict: {'m': {'foo': 'foo'}}
```

I also ran:

```text
python -m pytest tests/test_jsonable_encoder.py -q
26 passed, 1 skipped
```

`git diff --check master...review-pr-15676` passes, and `git merge-tree --write-tree master review-pr-15676` exits successfully, so I don't see whitespace or merge-conflict issues. The change looks correctly scoped because `exclude_defaults` is a global serialization flag like `exclude_unset` and `exclude_none`, which are already forwarded in this dict branch.

Update: I also checked the red GitHub checks currently shown on the PR. `check-labels` is failing because the PR has no required category label yet. The failed Ubuntu coverage job log shows `The runner has received a shutdown signal` followed by `The operation was canceled` while setting up `uv`, before the test suite started, so I don't see evidence of a failing test assertion from this change.

vidigoat:
Thanks for taking the time to verify this locally. Agreed on both red checks: the Ubuntu 3.13 coverage job was killed during uv setup (runner shutdown) before any test ran, so it was infra flake rather than a failing assertion, and check-labels just needs a maintainer to apply one of the category labels. I've rebased onto current master and force-pushed to retrigger a fresh CI run.

YuriiMotov:
Thanks for your interest and efforts, but let's keep it closed until maintainers can review the [discussion](https://github.com/fastapi/fastapi/discussions/15961).

>

Thanks for understanding!

PR #14960, opened 2026-02-21
docs: migrate testing tutorial from on_event to lifespan

Areeb455:
### **Description**
This PR migrates the asynchronous testing documentation and its associated source code from the deprecated `@app.on_event("startup")` pattern to the modern `lifespan` async context manager.

FastAPI has deprecated the `.on_event()` handlers in favor of the `lifespan` parameter. Current documentation in the "Advanced - Async Tests" section still recommends the legacy pattern to resolve `RuntimeError: Task attached to a different loop` issues.

By updating this section, we ensure that:
1. Users follow the current best practices for resource management.
2. The documentation remains consistent with the newer "Lifespan Events" tutorial.
3. We reduce the reliance on deprecated features that are slated for removal in future versions.

### **Changes**
- **Code:** Updated `docs_src/app_testing/tutorial003_py310.py` to implement the `lifespan` context manager.
- **Documentation:** Updated `docs/en/docs/advanced/async-tests.md` to update the prose description and provide a link to the main Lifespan documentation.

I have verified the code changes locally using the following:
- Ran `pytest docs_src/app_testing/tutorial003_py310.py` to ensure the test suite still passes with the `lifespan` implementation.
- Verified that the `TestClient` correctly triggers the lifespan events during the context manager entry/exit.

### **Checklist**
- [x] The code follows the modern Python 3.10+ style used in the FastAPI docs.
- [x] The documentation text is clear and contains a link to the relevant "Lifespan" reference page.
- [x] All automated tests for the specific source file pass.

PR #15697, opened 2026-06-06
Replace deprecated on_event startup with lifespan context manager in SQL databases tutorial

achyutadixit:
## Description
Replace deprecated `@app.on_event("startup")` with the modern `asynccontextmanager` lifespan pattern in the SQL databases tutorial.

Changes made:
- `docs_src/sql_databases/tutorial001_an_py310.py`: replaced `@app.on_event("startup")` with `@asynccontextmanager async def lifespan()` and updated `app = FastAPI(lifespan=lifespan)`
- `docs_src/sql_databases/tutorial002_an_py310.py`: same change as above
- `docs/en/docs/tutorial/sql-databases.md`: updated all line references to match new file structure, and updated description text from "application startup event" to "lifespan context manager"

### DIFF
--- a/docs/en/docs/tutorial/sql-databases.md
+++ b/docs/en/docs/tutorial/sql-databases.md
@@ -55,7 +55,7 @@ Later we'll improve it increasing security and versatility with **multiple model
 
 Import `SQLModel` and create a database model:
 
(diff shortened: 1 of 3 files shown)


YuriiMotov:
Thanks for your interest!

We are going to solve it in a different way.
See comment: https://github.com/fastapi/fastapi/pull/14960#issuecomment-4525831482

PR #15698, opened 2026-06-06
Fix duplicate operationId for routes registered with multiple HTTP methods

Sujit-1509:
## Description

When a route is registered with multiple HTTP methods (e.g. `methods=['GET', 'POST']`), the OpenAPI schema generator assigns the **same `operationId`** to every operation on that route. This violates the [OpenAPI 3.x specification](https://spec.openapis.org/oas/v3.1.0#operation-object) which requires `operationId` to be unique across all operations in a document.

Additionally, a spurious `UserWarning: Duplicate Operation ID` is emitted on every schema generation for such routes.

## Minimal Reproduction

`python
from fastapi import FastAPI

app = FastAPI()

async def handler(request: Request):
    return {}

app.add_api_route('/multi', handler, methods=['GET', 'POST'], name='multi_handler')

client = TestClient(app)
schema = client.get('/openapi.json').json()
print(schema['paths']['/multi']['get']['operationId'])   # multi_handler_multi_get
print(schema['paths']['/multi']['post']['operationId'])  # multi_handler_multi_get  <- BUG: same!
`

## Root Cause

Two issues compound each other:

**1. `fastapi/utils.py` — `generate_unique_id()`** uses `list(route.methods)[0]` on a `set`, which is **non-deterministic** across Python runs (`PYTHONHASHSEED` randomisation).

**2. `fastapi/openapi/utils.py` — `get_openapi_operation_metadata()`** iterates `for method in route.methods` but always uses `route.unique_id` (computed once at route construction) as the `operationId`, so every method gets the **same value**.

## Fix

- `generate_unique_id()`: use `sorted(route.methods)[0]` so `route.unique_id` is stable and deterministic.
- `get_openapi_operation_metadata()`: for single-method routes (the vast majority) the existing `route.unique_id` is used unchanged. For multi-method routes, the method suffix in `route.unique_id` is replaced with the current iteration method, producing a unique `operationId` per operation.

## Backward Compatibility

| Route type | Before | After |
|---|---|---|
| Single-method (`@app.get`, etc.) | unchanged | **unchanged** |
| Explicit `operation_id` | unchanged | **unchanged** |
| Multi-method (`methods=['GET','POST']`) | duplicate IDs + warning | **unique per method** |

## Tests Added

`tests/test_multi_method_route_operation_id.py` — 6 tests covering the core regression, three-method routes, single-method backward compat, explicit `operation_id` passthrough, `APIRouter` paths, and `generate_unique_id` determinism.

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -242,7 +242,32 @@ def get_openapi_operation_metadata(
     operation["summary"] = generate_operation_summary(route=route, method=method)
     if route.description:
         operation["description"] = route.description
-    operation_id = route.operation_id or route.unique_id
+    if route.operation_id:
+        operation_id = route.operation_id
+    elif len(route.methods) == 1:
+        # Single-method route: use the pre-computed unique_id as-is (backward compatible).
+        operation_id = route.unique_id
+    else:
+        # Multi-method route: route.unique_id was computed with only one method from the
+        # set (the first in sorted order). Derive a per-method operation ID by stripping
+        # that method suffix and replacing it with the current method so every operation
+        # on this route gets its own unique operationId in the OpenAPI schema.
+        base_operation_id = route.unique_id
+        for _http_method in (
+            "delete",
+            "get",
+            "head",
+            "options",
+            "patch",
+            "post",
+            "put",
+            "trace",
+        ):
+            _suffix = f"_{_http_method}"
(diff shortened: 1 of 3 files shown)


YuriiMotov:
Thanks for your interest!

Please, search for existing PRs (including closed) and review them before opening new.
See: https://github.com/fastapi/fastapi/pull/15398#issuecomment-4528207992

PR #15699, opened 2026-06-06
Fix typos in docs and source code, fix status_code priority in _build_response_args

sadreammm:
Bug fix:

- routing.py: The _build_response_args function was unconditionally overwriting the explicit status_code parameter with the response class default. Fixed with if/elif so the explicit parameter takes priority.

Typos (8 fixes across docs + source):

- "doesn't known" -> "doesn't know" (https.md)
- Missing "they" (project-generation.md)
- "Environment Variable" -> "Environment Variables" (environment-variables.md)
- "an undefined behavior" -> "undefined behavior" (body.md)
- Two broken Markdown links missing ) (param_functions.py)
- Double space "HTML  that" (docs.py)
- Truncated comment "satisfy ty" (datastructures.py)
- Four backticks -> three in two docstrings (oauth2.py)

Test cleanup:

- Removed two leftover print() debug statements


docs/en/docs/environment-variables.md
```diff
@@ -291,7 +291,7 @@ This information will be useful when learning about [Virtual Environments](virtu
 
 With this you should have a basic understanding of what **environment variables** are and how to use them in Python.
 
-You can also read more about them in the [Wikipedia for Environment Variable](https://en.wikipedia.org/wiki/Environment_variable).
+You can also read more about them in the [Wikipedia for Environment Variables](https://en.wikipedia.org/wiki/Environment_variable).
```
YuriiMotov, line 294: This is incorrect - literal name of the article

fastapi/datastructures.py
```diff
@@ -182,5 +182,5 @@ def Default(value: DefaultType) -> DefaultType:
 
 
 # Sentinel for "parameter not provided" in Param/FieldInfo.
-# Typed as None to satisfy ty
+# Typed as None to satisfy type checkers
```
YuriiMotov, line 185: not needed

YuriiMotov:
Thanks for reporting this!

Thanks for understanding!

PR #15705, opened 2026-06-07
⚡️ Skip re-solving cached dependency sub-trees in solve_dependencies

ranjanprasad96:
## Pull Request

Discussion: https://github.com/fastapi/fastapi/discussions/15704

## Description

This optimizes `fastapi/dependencies/utils.py::solve_dependencies` so it does not re-solve a dependency sub-tree that has already been cached.

**The problem**

For every sub-dependency, the loop today does this, in order:

1. Always recurse into `solve_dependencies(...)` and re-solve the **entire sub-tree**.
2. *Then* check `dependency_cache`.
3. On a cache hit, throw away everything from step 1 and use the cached value instead.

So when a `use_cache=True` dependency is requested more than once in a request, its whole sub-tree is re-walked and re-validated on every request — only for the result to be discarded.

**Observable effect**

Minimal reproduction — `cached_parent` is cached and requested twice, and it depends on a non-cached `child_dependency`:

```python

app = FastAPI()
child_run_count = 0


def child_dependency() -> int:  # non-cached child
    global child_run_count
    child_run_count += 1
    return child_run_count


def cached_parent(  # cached parent (use_cache=True by default)
    value: int = Depends(child_dependency, use_cache=False),
) -> int:
    return value


@app.get("/example")
def example(
    first: int = Depends(cached_parent),
    second: int = Depends(cached_parent),
):
    return {"first": first, "second": second, "child_run_count": child_run_count}


print(TestClient(app).get("/example").json())
```

- On `master`: `{'first': 1, 'second': 1, 'child_run_count': 2}` — `cached_parent` runs once (correct), but `child_dependency` runs **twice** and the second result is discarded.
- With this PR: `{'first': 1, 'second': 1, 'child_run_count': 1}` — the cached parent is reused without re-solving its sub-tree.

**The change**

Short-circuit at the top of the loop, *before* recursing:

```python
    if sub_dependant.name is not None:
        values[sub_dependant.name] = solved
    continue
```

A value only enters `dependency_cache` after an error-free solve, so a cache hit means the dependency already resolved cleanly and re-solving could only reproduce an identical result. With this in place, the existing post-recursion cache-reuse branch becomes unreachable (every cache hit is now handled before recursing), so it is removed.

**Compatibility**

`use_cache=True` semantics are unchanged: the cached function still runs exactly once per request, and every reference gets the same value. The only observable difference is that a `use_cache=False` descendant of a cached dependency no longer re-runs on repeat references (it now runs once, instead of once per reference). This is undocumented behavior and running an input to a function that is not being called appears unintended, but it is flagged here for visibility.

## AI Disclaimer

Model: Claude Opus 4.8 (1M context), via Claude Code.

Prompt (paraphrased): add an early-out in `solve_dependencies` so that when a sub-dependency is already in `dependency_cache`, the cached value is reused and the loop `continue`s (skipping the sub-tree recursion); then run the test suite, do a strict review for regressions, and confirm 100% coverage.

<details>

</details>

## Checklist

- [x] Coverage stays at 100%.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -629,6 +629,12 @@ async def solve_dependencies(
         sub_dependant.call = cast(Callable[..., Any], sub_dependant.call)
         call = sub_dependant.call
         use_sub_dependant = sub_dependant
+        # Reuse the cached result and skip re-solving the whole sub-tree.
+        if sub_dependant.use_cache and sub_dependant.cache_key in dependency_cache:
+            solved = dependency_cache[sub_dependant.cache_key]
+            if sub_dependant.name is not None:
+                values[sub_dependant.name] = solved
+            continue
         if (
             dependency_overrides_provider
             and dependency_overrides_provider.dependency_overrides
--- a/tests/test_dependency_cache.py
+++ b/tests/test_dependency_cache.py
@@ -35,6 +35,23 @@ async def get_sub_counter_no_cache(
     return {"counter": count, "subcounter": subcount}
 
 
+async def super_dep_no_cache_child(
+    count: int = Depends(dep_counter, use_cache=False),
+):
+    return count
+
+
+@app.get("/cached-parent-no-cache-child/")
+async def get_cached_parent_no_cache_child(
+    # The same cached dependency is referenced twice. On the second reference the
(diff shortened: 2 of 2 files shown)


ranjanprasad96:
Closing this PR since its a behavioural change

PR #15706, opened 2026-06-08
Serialize ServerSentEvent model data with by_alias=True

vincere-mori:
## Issue

Reported in #15703. When a `ServerSentEvent`'s `data` is a Pydantic model, `serialization_alias` (and any alias-only field) is dropped from the `data:` line:

```python
class Item(BaseModel):

async def events():
    yield ServerSentEvent(data=Item(name="Portal Gun"))
```

```
data: {"name":"Portal Gun"}        # got
data: {"itemName":"Portal Gun"}    # expected
```

## Cause

In `routing.py`, the `ServerSentEvent` data branch serialized models with `model_dump_json()` and no `by_alias`:

```python
if hasattr(item.data, "model_dump_json"):
    data_str = item.data.model_dump_json()
else:
    data_str = json.dumps(jsonable_encoder(item.data))
```

The two other SSE serialization paths already alias: the `else` branch above goes through `jsonable_encoder` (which defaults to `by_alias=True`), and plain yielded objects go through `stream_item_field.serialize_json(..., by_alias=response_model_by_alias)`. So only models wrapped in `ServerSentEvent` came out un-aliased, which is inconsistent with how response models serialize everywhere else in FastAPI.

## Fix

Pass `by_alias=True` to `model_dump_json()` so the `ServerSentEvent` model path matches the other paths.

## Tests

Added a test (`/items/stream-sse-aliased`) yielding a model with `serialization_alias`; it fails on `master` (emits `"name"`) and passes here (emits `"itemName"`). The rest of `tests/test_sse.py` stays green and `ruff` is clean.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -507,7 +507,7 @@ def get_request_handler(
                             data_str: str | None = item.raw_data
                         elif item.data is not None:
                             if hasattr(item.data, "model_dump_json"):
-                                data_str = item.data.model_dump_json()
+                                data_str = item.data.model_dump_json(by_alias=True)
                             else:
                                 data_str = json.dumps(jsonable_encoder(item.data))
                         else:
--- a/tests/test_sse.py
+++ b/tests/test_sse.py
@@ -8,12 +8,16 @@ from fastapi import APIRouter, FastAPI
 from fastapi.responses import EventSourceResponse
 from fastapi.sse import ServerSentEvent
 from fastapi.testclient import TestClient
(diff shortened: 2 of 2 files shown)


YuriiMotov:
Thanks, but acting this way is unfair towards the author of the discussion who reported the issue and, following the repository rules, asked for approval to open PR.
I will close this PR to give the initial reporter chance to contribute, but I will have this PR in mind in case initial reporter is not able to open PR.

PR #15729, opened 2026-06-10
[codex] Add httpx2 to standard extras

sairamveereddy:
## Summary

Add `httpx2 >=2.0.0` to the `standard`, `standard-no-fastapi-cloud-cli`, and `all` optional dependency extras.

## Why

Current Starlette's `TestClient` prefers `httpx2`. Without `httpx2` installed, importing `fastapi.testclient.TestClient` can emit a Starlette deprecation warning while falling back to `httpx`. In warning-strict environments, that warning can fail tests or local validation even after installing `fastapi[standard]`.

The test and docs dependency groups already include `httpx2`; this aligns the public extras with that setup.

## Validation

- `python -m pip install -e ".[standard]"`
- `python -W error::DeprecationWarning -c "from fastapi.testclient import TestClient; print(TestClient.__name__)"`
- `python -m pytest tests\test_request_params\test_path\test_required_str.py tests\test_param_include_in_schema.py -q`


YuriiMotov:
You can always add `httpx2` to your project's dependencies to avoid this warning

bonastreyair:
Thanks @SairamVeeereddy I have the same "issue". 

It is weird that by default fastapi generates the warnings from the recommended ways of working, no?

https://github.com/encode/httpx is unmantained and https://github.com/pydantic/httpx2 is maintained by pydantic, wouldn't it be better to depend on httpx2 instead of httpx then?

@YuriiMotov is it considered dropping httpx dependency entirely and only depending on httpx2, so basically doing a full transition of the dependency? Thanks in advance 😃 !

YuriiMotov:
@bonastreyair, thanks! We are aware of this situation and Sebastian took a decision to keep it as it is for now for some reasons.
You can start a discussion to gather more community feedback, this would help us understand how important it is for users

PR #15739, opened 2026-06-12
Add context to response serialization

Ocyss:
## Description

This PR adds support for passing `context` to response serialization.

This allows serializers/validators to access runtime context during serialization, for example the current `Request`, and adjust the serialized output accordingly.

## Use case

This can be useful for cases where the serialized response depends on request-specific data or other runtime state.

For example, selecting a localized value from a dict using a request header:

```python
def validator(
    v: Any, handler: ValidatorFunctionWrapHandler, info: "ValidationInfo[Request]"
):
    if isinstance(v, dict) and isinstance(info.context, Request):
        if lang := info.context.headers.get("x-lang"):
            return v.get(lang)
    return None
```

## Possible applications

* Localized response content based on request headers.
* Conditional serialization depending on the current user or permissions.
* Masking or transforming fields before returning the response.
* Adjusting serialized output based on tenant, API version, or feature flags.

## Motivation

Currently, response serialization does not receive `context`, which makes these kinds of use cases harder to implement cleanly at the schema/serialization layer.

Passing `context` makes response serialization more flexible and keeps this logic closer to the model/serializer definitions instead of requiring custom handling in route functions.

### DIFF
--- a/fastapi/_compat/v2.py
+++ b/fastapi/_compat/v2.py
@@ -176,10 +176,13 @@ class ModelField:
         values: dict[str, Any] = {},  # noqa: B006
         *,
         loc: tuple[int | str, ...] = (),
+        context: Any | None = None,
     ) -> tuple[Any, list[dict[str, Any]]]:
         try:
             return (
-                self._type_adapter.validate_python(value, from_attributes=True),
+                self._type_adapter.validate_python(
+                    value, from_attributes=True, context=context
+                ),
                 [],
             )
         except ValidationError as exc:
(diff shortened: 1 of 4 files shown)


avinashkamat48 (commented):
`serialize_response` now accepts a `context`, but this PR only passes `context=request` from the streaming path shown here. The normal non-streaming response path still calls `serialize_response(...)` without context, so Pydantic response validators/serializers will not see the request for regular JSON responses. Please wire the request context through the standard route handler path as well and add a non-streaming response-model regression test; otherwise the new feature only works for streaming responses.

tiangolo:
Thanks for the interest, but please start with a GitHub Discussion showing your specific use case and then we can see what is the best way to solve it. I'll close this for now. :coffee:

PR #15694, opened 2026-06-05
✏️ Fix grammar, typos, and broken links in docs

YuriiMotov:
Fixed typos and grammar mistakes in docs.

I used Claude Code to find and fix mistakes, but carefully reviewed all corrections and asked to rephrase a couple of sentences to make them easier to read.

---

docs live command showed that there were some broken links. I applied minimal fix in [0e25472](https://github.com/fastapi/fastapi/pull/15694/commits/0e25472f033eb8eff30e2134241cefaaf5231725) (updated some links and removed if there are no relevant section anymore), but we should probably revise the "FastAPI People" page

---

Noticed that "Migrate from Pydantic v1 to Pydantic v2" page still states that FastAPI supports `pydantic.v1`, that is no longer true. Updated that page to clarify that it was only supported up to FastAPI 0.128.0 (commit [03b44b6](https://github.com/fastapi/fastapi/pull/15694/commits/03b44b64ce19fcd63035bd321ec4af5b3585b322))

---

Also fixed typos in agent skills (commit [0e25472](https://github.com/fastapi/fastapi/pull/15694/commits/0e25472f033eb8eff30e2134241cefaaf5231725)) as it's also kind of documentation

PR #15746, opened 2026-06-14
fix: remove incorrect "like" in first-steps.md grammar

Reena1912:
## Pull Request

Fixed grammatical error: "looks like" → "looks"
"how the raw OpenAPI schema looks like" is incorrect English.
The correct form is "how the raw OpenAPI schema looks".


YuriiMotov:
Thanks!
It will be handled in https://github.com/fastapi/fastapi/pull/15694

PR #15773, opened 2026-06-16
:pencil2: Fix typos in release notes

JavierSanchezCastro:
Corrected typos in release notes regarding custom APIRoute subclasses.

### DIFF
--- a/docs/en/docs/release-notes.md
+++ b/docs/en/docs/release-notes.md
@@ -81,7 +81,7 @@ Still, for now, consider this very experimental and potentially changing and bre
 
 #### Future Features Enabled
 
-* Custom `APIRoute` subclasses (undocumented, but alraedy works as described above)
+* Custom `APIRoute` subclasses (undocumented, but already works as described above)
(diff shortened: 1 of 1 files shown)


strangeleo02 (changes requested):
Good catch fixing the typo in described!

However, there is another typo on the exact same line that was missed: "alraedy" should be spelled "already"

strangeleo02 (approved):
LGTM! Thanks for catching that typo. 🚀

YuriiMotov:
@JavierSanchezCastro, thanks for pointing to this!

Unfortunately, we don't check `release-notes` for typos in CI, so it's easy to introduce typos there..

There are actually a couple of more typos in the same release.

There is already existing big PR that is intended to fix typos across all pages: https://github.com/fastapi/fastapi/pull/15694

I just updated it to fix these recently added typos: https://github.com/fastapi/fastapi/pull/15694/commits/d24cb93fa989337812fbb60a2021a626451c5c88

PR #15792, opened 2026-06-18
🔨 Use `gpt-5.5` model in `translate.py`, specify `-chat` to avoid warnings

YuriiMotov:
The `-chat` suffix is needed to avoid the warning:
```
/home/yurii/code/1.FastAPI/fastapi/.venv/lib/python3.11/site-packages/pydantic_ai/agent/__init__.py:414: PydanticAIDeprecationWarning: In v2.0, 'openai:' will resolve to the OpenAI Responses API by default. Use 'openai-chat:' to keep current Chat Completions behavior, or 'openai-responses:' to opt in early.
```


tiangolo (approved):
Great, thanks! 🙌

PR #15797, opened 2026-06-19
🌐 Update translations for hi (add-missing)

tiangolo:
🌐 Update translations for hi (add-missing)

It uses the prompt file https://github.com/fastapi/fastapi/blob/master/docs/hi/llm-prompt.md.


YuriiMotov:
Asked Claude Code to review and suggest initial list of specific term translations. Opened PR: https://github.com/fastapi/fastapi/pull/15810

Re-generated translated pages with the updated prompt.

Main thing is to see if there are any common terms or phrases that LLM translates incorretly or inconsistently, and we want to tell it to use specific translation for this term.
In translations for other languages we often tell it to use English terms in specific cases instead of translating them.
If you see such terms - we can add them to the list of special terms in language-specific LLM-prompt (see PR https://github.com/fastapi/fastapi/pull/15810).

See how it's done for other languages:
* [Spanish](https://github.com/fastapi/fastapi/blob/master/docs/es/llm-prompt.md?plain=1)
* [Russian](https://github.com/fastapi/fastapi/blob/master/docs/ru/llm-prompt.md?plain=1)

Just leave comments with your suggestions here or in PR https://github.com/fastapi/fastapi/pull/15810

igennova (approved):

ABcDexter (approved):
I've gone through the translations, these look good.

YuriiMotov:
@igennova, @ABcDexter, thanks for the reviews!

YuriiMotov:
@igennova, could you please take a look at https://github.com/fastapi/fastapi/pull/15810? It needs the approval from one more native speaker to be merged

PR #15812, opened 2026-06-20
📝 Fix copy button copying HTML markup on Termynal command lines

patrickwehbe:
## What this fixes

Several pages in the docs write the terminal **input** lines inside `termy`/`console` blocks with HTML coloring, for example:

```
$ <font color="#4E9A06">fastapi</font> dev
```

Termynal animates the input line using `textContent`, so the `<font>` tags never render on screen. However, MkDocs Material's copy button reads the underlying code, not the rendered text, so clicking it copies the literal markup:

```
<font color="#4E9A06">fastapi</font> dev
```

Users then paste the markup instead of the command.

## The fix

Strip the HTML wrappers from only the `$` input command lines so the copy button yields the plain command. The colored **output** lines below each command are intentional styling and are left untouched.

The homepage block already uses the plain `$ fastapi dev` form and copies correctly, so this just brings these pages in line with it.

Lines changed (input commands only):

- `docs/en/docs/fastapi-cli.md`
- `docs/en/docs/tutorial/first-steps.md`
- `docs/en/docs/tutorial/index.md`
- `docs/en/docs/deployment/manually.md`
- `docs/en/docs/deployment/server-workers.md`

See discussion #15700.

### DIFF
--- a/docs/en/docs/deployment/manually.md
+++ b/docs/en/docs/deployment/manually.md
@@ -7,7 +7,7 @@ In short, use `fastapi run` to serve your FastAPI application:
 <div class="termy">
 
 ```console
-$ <font color="#4E9A06">fastapi</font> run <u style="text-decoration-style:solid">main.py</u>
+$ fastapi run main.py
 
   <span style="background-color:#009485"><font color="#D3D7CF"> FastAPI </font></span>  Starting production server 🚀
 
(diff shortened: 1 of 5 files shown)


mohamedtaqysalmi (commented):
Stripping Termynal HTML markup from the copy payload fixes a real UX bug where users paste styled spans instead of plain commands. Scoped to deployment/tutorial Termynal blocks — good that tests or manual verification covered multiple pages using the same pattern.

svlandeg:
@mohamedtaqysalmi : I feel like you're using AI to produce these reviews, is that correct?

YuriiMotov:
@patrickwehbe, thank you for your interest and efforts, but I think we should go with https://github.com/fastapi/fastapi/pull/15941

PR #15813, opened 2026-06-20
🐛 Set `X-Accel-Buffering: no` on JSONL streaming responses

torrresagus:
## What's wrong

FastAPI's Server-Sent Events responses set `X-Accel-Buffering: no` so streaming survives a buffering proxy (the `is_sse_stream` branch in `fastapi/routing.py`, added in #15030):

```python
# For Nginx proxies to not buffer server sent events
response.headers["X-Accel-Buffering"] = "no"
```

The JSON Lines streaming responses (the `yield`-based `application/jsonl` streaming added in #15022, the `is_json_stream` branch) stream items incrementally in the same way, but don't set it.

Behind a proxy that buffers by default (e.g. Nginx with `proxy_buffering on`), the JSONL lines are buffered and flushed together, which defeats the streaming: the client receives the `200` immediately, but the lines don't arrive until the proxy buffer fills. SSE isn't affected because it already sends `X-Accel-Buffering: no`. The two features landed in separate PRs (SSE #15030, JSONL #15022), which is likely why only the SSE path got the header.

## Change

Set `X-Accel-Buffering: no` on the JSONL `StreamingResponse`, so incremental streaming works through a buffering proxy the same way SSE does.

`Cache-Control` is intentionally **not** set: unlike SSE, JSON Lines is also used for bulk/streamed exports where caching can be legitimate, so the caching policy is left to the user (it can be set on the `Response`). This is documented in the JSON Lines tutorial. (Thanks @luzzodev for the review that clarified this.)

## Tests

- New `test_stream_disables_proxy_buffering` asserts `x-accel-buffering: no` across the four JSONL streaming tutorial variants (parametrized like the existing SSE streaming tests).
- The existing SSE and streaming suites still pass.

## Context

Discussed first in #15794, where the same failure mode was confirmed independently, including a report behind Nginx in production.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -677,6 +677,9 @@ def get_request_handler(
                     media_type="application/jsonl",
                     **response_args,
                 )
+                response.headers["Cache-Control"] = "no-cache"
+                # For Nginx proxies to not buffer the streamed response
+                response.headers["X-Accel-Buffering"] = "no"
                 response.headers.raw.extend(solved_result.response.headers.raw)
             elif _is_async_gen_callable(dependant.call) or _is_gen_callable(
                 dependant.call
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -677,11 +677,14 @@ def get_request_handler(
                     media_type="application/jsonl",
                     **response_args,
                 )
+                response.headers.raw.extend(solved_result.response.headers.raw)
                 # For Nginx proxies to not buffer the streamed response.
+                # Set after extending, so an explicit header from the *path
+                # operation function* takes precedence and can re-enable
+                # buffering.
                 # Caching headers are intentionally left to the user, as JSONL
                 # is also used for bulk exports where caching can be legitimate.
-                response.headers["X-Accel-Buffering"] = "no"
-                response.headers.raw.extend(solved_result.response.headers.raw)
+                response.headers.setdefault("X-Accel-Buffering", "no")
             elif _is_async_gen_callable(dependant.call) or _is_gen_callable(
                 dependant.call
             ):
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -677,8 +677,9 @@ def get_request_handler(
                     media_type="application/jsonl",
                     **response_args,
                 )
-                response.headers["Cache-Control"] = "no-cache"
-                # For Nginx proxies to not buffer the streamed response
+                # For Nginx proxies to not buffer the streamed response.
+                # Caching headers are intentionally left to the user, as JSONL
+                # is also used for bulk exports where caching can be legitimate.
                 response.headers["X-Accel-Buffering"] = "no"
                 response.headers.raw.extend(solved_result.response.headers.raw)
             elif _is_async_gen_callable(dependant.call) or _is_gen_callable(
--- a/docs/en/docs/tutorial/stream-json-lines.md
+++ b/docs/en/docs/tutorial/stream-json-lines.md
@@ -72,6 +72,18 @@ If you want to stream binary data, for example video or audio, check the advance
 
 ///
 
+## Proxy Buffering { #proxy-buffering }
+
+When a JSON Lines response is served behind a proxy that buffers responses by default (for example Nginx with `proxy_buffering on`), the proxy could hold back the lines and deliver them all at once, which would defeat the streaming.
+
+To prevent that, FastAPI sets the `X-Accel-Buffering: no` header on JSON Lines responses, telling Nginx (and compatible proxies) not to buffer them.
+
+/// note
+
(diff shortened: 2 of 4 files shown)


luzzodev:
Hello @torrresagus, thanks for the PR! I understand the point, but I'm not fully convinced yet.                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                   
  JSONL isn't 1 to 1 "SSE but in JSON",  it's also used for plain bulk/streamed exports where caching can be legitimate. So I'd split the two headers:                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                   
  - X-Accel-Buffering: no. No objection, buffering defeats streaming in every case.                                                                                                                                                                                                                            
  - Cache-Control: no-cache, On this one I'm hesitant about. It blocks the max-age "serve from cache without hitting origin" case that an export endpoint might want.                                                                                                                                                                                                                                                                                            

In any case the picked behaviour should be clearly pointed out in the docs.

torrresagus:
Thanks for the careful review @luzzodev — that's a fair distinction, and you've convinced me.

You're right that JSONL isn't semantically "SSE in JSON": it's also a transport for bulk/streamed exports where caching can be legitimate. And since `Cache-Control: no-cache` forces revalidation rather than meaning "don't store", defaulting to it would quietly take away the `max-age` "serve from cache without hitting origin" option from those export endpoints. That's a policy call the framework shouldn't make on the user's behalf — anyone who does want it can still set it explicitly via the `Response`.

`X-Accel-Buffering: no` is the one that's unambiguous: buffering defeats incremental streaming in every case, export or live.

So I'll drop `Cache-Control` and keep only `X-Accel-Buffering: no`, and add a short note in the JSONL streaming docs spelling out that the response disables proxy buffering (and that caching headers are left to the user). I'll update the test to match and push shortly.

luzzodev:
LGTM!

torrresagus:
Thanks for the review @luzzodev!

mohamedtaqysalmi (commented):
Mirroring the SSE anti-buffering behavior for JSONL is the right fix behind Nginx, and intentionally not imposing Cache-Control preserves legitimate export caching — the new docs section explains that well.

Small inconsistency: the PR title mentions both Cache-Control and X-Accel-Buffering, but 
outing.py only sets X-Accel-Buffering. Either add the header for SSE parity or tighten the title/description so reviewers aren't looking for a header that isn't in the diff.

torrresagus:
Good catch @mohamedtaqysalmi — the title and description were stale after I dropped `Cache-Control` per the review. Re-adding it isn't the intent (it's deliberately left to the user), so I tightened the title and description to match the diff: only `X-Accel-Buffering: no` is set now. Thanks!

torrresagus:
@tiangolo Let me know if you need anything else from me.

fastapi/routing.py
```diff
@@ -658,6 +658,10 @@ def _sync_stream_jsonl() -> Iterator[bytes]:
                     media_type="application/jsonl",
                     background=solved_result.background_tasks,
                 )
+                # For Nginx proxies to not buffer the streamed response.
+                # Caching headers are intentionally left to the user, as JSONL
+                # is also used for bulk exports where caching can be legitimate.
+                response.headers["X-Accel-Buffering"] = "no"
```
Sanjays2402: since this is set before `raw.extend`, a user who sets `X-Accel-Buffering` on their own response ends up with two copies of the header instead of overriding it. intentional that the endpoint can't turn buffering back on?

torrresagus: Good catch, you're right. `raw.extend` appends, so a user-set `X-Accel-Buffering` ended up as a second header instead of overriding it.

It wasn't a deliberate "you can't turn buffering back on" decision, I had just mirrored the SSE branch a few lines above, which does the same thing with `Cache-Control` and `X-Accel-Buffering`.

Fixed in fe132657d by setting the header after the extend:

```python
response.headers.raw.extend(solved_result.response.headers.raw)
response.headers.setdefault("X-Accel-Buffering", "no")
```

Now the default is still `no`, but a *path operation* that sets the header itself wins, and there's only one copy of it. Added a test for both cases in `tests/test_stream_json_lines_headers.py` and a line in the docs.

One note while testing it: the header has to be set from a dependency, not from inside the generator body, since the headers are already sent by the time the first item is yielded.

I left the SSE branch as is, changing its behavior feels out of scope for this PR, but happy to do it here too if a maintainer prefers to keep both consistent.

vamshi256:
I tried reproducing this locally with FastAPI behind Nginx on Windows. I configured proxy_buffering on and used a StreamingResponse producing JSONL chunks at 1-second intervals. The client still received each chunk incrementally rather than receiving them all at once.

I also noticed that the FastAPI version I tested already sets X-Accel-Buffering: no for the response in routing.py. Because of this, I'm wondering if the behavior described in this PR has already been addressed in the version I tested.

Could you clarify which FastAPI version/commit was used for the original reproduction?

PR #15824, opened 2026-06-22
Swallow BrokenResource/ClosedResource in SSE keepalive on client disconnect

trilamsr:
First contribution here, hi! Found this one in discussion #15725 and figured it'd be a nice intro PR.

When a client disconnects from an SSE endpoint, the request-scoped exit stack closes the keepalive consumer stream before the producer task group is cancelled. If the keepalive task happens to be inside `send_keepalive.send(...)` at that moment, anyio raises `BrokenResourceError`. The task group bubbles it up as an `ExceptionGroup` that uvicorn logs as "Exception in ASGI application" on every normal disconnect.

The existing handler already swallows `EndOfStream` for the same logical condition (peer is gone, nothing to do). This adds `BrokenResourceError` and `ClosedResourceError` to the same except clause.

Reproduced reliably in `test_keepalive_no_error_on_client_disconnect` by driving the ASGI app directly with a slow `send` callback (mimicking real socket backpressure) and triggering disconnect mid-stream. Without the fix, the task group raises `ExceptionGroup[BrokenResourceError]` from `routing.py:592`; with the fix, the test passes on both asyncio and trio.

No behavior change for normal end-of-stream and the existing `EndOfStream` path is untouched.

Fixes #15725.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -590,7 +590,11 @@ def get_request_handler(
                                         await send_keepalive.send(data)
                                     except TimeoutError:
                                         await send_keepalive.send(KEEPALIVE_COMMENT)
-                            except anyio.EndOfStream:
+                            except (
+                                anyio.EndOfStream,
+                                anyio.BrokenResourceError,
+                                anyio.ClosedResourceError,
+                            ):
                                 pass
 
                     async with anyio.create_task_group() as tg:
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Closing (process reasons), cf https://github.com/fastapi/fastapi/discussions/15725#discussioncomment-17403619

PR #15828, opened 2026-06-23
🌐 Add Spanish translation for docs/es/docs/management.md

delcenjo:
Add Spanish translation for `docs/es/docs/management.md`.

I am a native Spanish speaker and mirrored the English page structure exactly, translating only the prose and keeping all links and headings consistent.


YuriiMotov:
Thanks, but this page was intentionally excluded from the pages that should be translated to other languages.
See https://github.com/fastapi/fastapi/blob/a497a025e7114ca442478ed28da7e0a1cdc6177a/scripts/docs.py#L42-L49

I think that's because it's not possible to help managing repository without knowing English

PR #15829, opened 2026-06-23
🌐 Add Spanish translation for docs/es/docs/tutorial/frontend.md

delcenjo:
Add Spanish translation for `docs/es/docs/tutorial/frontend.md`.

This page is part of the translatable tutorial (it is not in the non-translated sections list). I am a native Spanish speaker and mirrored the English page exactly: same headings and anchors, same code includes, and the admonition labels follow the existing Spanish convention.


YuriiMotov:
Thanks, but there is no need in this. We have an automated translation workflow, and this page will be automatically translated to all supported languages in a week.

PR #15838, opened 2026-06-25
✨ Add support for `app.query()` and `router.query()` (HTTP `QUERY` method)

ItamarShalev:
## Description

This adds first-class support for the **HTTP `QUERY` method** - a new HTTP method standardized in [RFC 10008 - "The HTTP QUERY Method"](https://datatracker.ietf.org/doc/rfc10008/) (June 2026). `QUERY` is a safe, idempotent method that carries a request body, so a client can send the query in the request content instead of encoding it in the URI (like `GET`, but with a body).

`QUERY` becomes usable exactly like every other HTTP method:

```python
from fastapi import FastAPI

app = FastAPI()


class Search(BaseModel):
    term: str
    limit: int = 10


@app.query("/search")
def search(search: Search) -> dict:
    return {"term": search.term, "limit": search.limit}
```

### Changes

- `FastAPI.query()` / `APIRouter.query()` path operation decorators, mirroring `get`/`post`/.../`trace` exactly (identical signature, `Doc` annotations and behavior), registering routes with `methods=["QUERY"]`.
- `QUERY` added to `METHODS_WITH_BODY` so the request body is included in the generated OpenAPI schema.
- `query` documented in the FastAPI and APIRouter API reference and in the "First Steps" tutorial, alongside the other methods.
- Tests cover app- and router-level registration, `add_api_route`, body parsing, validation (422), 405 on the wrong method, GET/QUERY coexistence on the same path, and the full generated OpenAPI document (via `inline_snapshot`).

## AI Disclaimer

Implemented with assistance from Claude (Opus 4.8), then reviewed and verified locally: `ty` type-check, `ruff` lint/format, and the full test suite (3269 passed).

## Checklist

- [x] Coverage stays at 100%.

### DIFF
--- a/docs/en/docs/reference/apirouter.md
+++ b/docs/en/docs/reference/apirouter.md
@@ -22,4 +22,5 @@ from fastapi import APIRouter
             - head
             - patch
             - trace
+            - query
             - on_event
--- a/docs/en/docs/reference/fastapi.md
+++ b/docs/en/docs/reference/fastapi.md
@@ -27,6 +27,7 @@ from fastapi import FastAPI
             - head
             - patch
             - trace
+            - query
             - on_event
             - middleware
             - exception_handler
--- a/docs/en/docs/tutorial/first-steps.md
+++ b/docs/en/docs/tutorial/first-steps.md
@@ -287,6 +287,7 @@ One of:
 * `HEAD`
 * `PATCH`
 * `TRACE`
+* `QUERY`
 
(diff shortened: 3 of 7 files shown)


YuriiMotov:
@ItamarShalev, please, follow the PR template and start with opening a discussion.
In some cases it's easier for the maintainers to implement feature than reviewing external PR. This case is one of them.

I will close this PR to keep the list of PRs cleaner

PR #15853, opened 2026-06-27
docs: fix dropped word in get_swagger_ui_html docstring

Kropiunig:
The `get_swagger_ui_html` docstring reads "Generate and return the HTML  that loads Swagger UI" — a word is missing (and there's a leftover double space). The parallel `get_redoc_html` docstring correctly reads "Generate and return the HTML response that loads ReDoc", confirming the intended wording. This inserts "response" so the two docstrings match.

### DIFF
--- a/fastapi/openapi/docs.py
+++ b/fastapi/openapi/docs.py
@@ -135,7 +135,7 @@ def get_swagger_ui_html(
     ] = None,
 ) -> HTMLResponse:
     """
-    Generate and return the HTML  that loads Swagger UI for the interactive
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@Kropiunig, thanks for your interest.

I think the dropped word is not _"response"_, but rather _"code"_ or _"content"_.
_"HTML response"_ doesn't fit as it can not _"load Swagger UI"_. _"HTML code"_ can. 

And I think it's fine to omit this, it's clear from the context.

Also, note that we now have special discussion for reporting small issues in docs: https://github.com/fastapi/fastapi/discussions/15849

PR #15856, opened 2026-06-28
🐛 Fix named PEP 695 type alias used as body parameter losing its OpenAPI component ($ref)

### LINKED ISSUE #15855: Named type alias body loses OpenAPI ref
### Description

FastAPI generates different OpenAPI schemas for a bare named Pydantic type alias and the same alias wrapped in `Annotated[..., Body()]`.

This came up while discussing the Pydantic side in https://github.com/pydantic/pydantic/issues/13236.

Example:

```py
from typing import Annotated, Literal

app = FastAPI()


class Cat(BaseModel):
    pet_type: Literal["cat"]
    meows: int


class Dog(BaseModel):
    pet_type: Literal["dog"]
    barks: float


class Lizard(BaseModel):
    pet_type: Literal["reptile", "lizard"]
    scales: bool


type Pet = Annotated[
    Cat | Dog | Lizard,
    Field(discriminator="pet_type"),
]


@app.post("/bare")
def bare(pet: Pet):
    return pet


@app.post("/body")
def body(pet: Annotated[Pet, Body()]):
    return pet
```

Expected: both routes should generate the same request-body schema shape, with a named `Pet` component:

```json
{
  "$ref": "#/components/schemas/Pet"
}
```

Actual: `/body` references `#/components/schemas/Pet`, but `/bare` emits the discriminated union schema inline in the request body.

### Why this matters

For normal `BaseModel` request bodies, FastAPI supports the ergonomic form:

```py
def create_item(item: Item):
    ...
```

A named discriminated union is also a valid Pydantic body type, but today it needs the more verbose form to get the same named OpenAPI component behavior:

```py
def create_pet(pet: Annotated[Pet, Body()]):
    ...
```

`RootModel` is not a good workaround because it adds a `.root` wrapper, while the desired request body is the flat union object itself.

### Suggested fix

FastAPI should preserve the named alias when it auto-infers a body parameter.

It looks like FastAPI currently unwraps `TypeAliasType` / PEP 695 aliases during parameter analysis. That may be useful for deciding whether the parameter is a body/query parameter, but the original alias should still be used when creating the inferred body field.

So this:

```py
def create_pet(pet: Pet):
    ...
```

should produce the same OpenAPI request-body schema as:

```py
def create_pet(pet: Annotated[Pet, Body()]):
    ...
```

iamsharduld:
## Description

Fixes #15855.

A bare [PEP 695](https://peps.python.org/pep-0695/) type alias used directly as a request-body parameter generated an **inline** OpenAPI schema, while the equivalent `Annotated[Alias, Body()]` form produced a named `$ref`. The two forms now produce the same named component.

```python
type Pet = Annotated[Cat | Dog | Lizard, Field(discriminator="pet_type")]

@app.post("/bare")
def bare(pet: Pet): ...                      # before: inline oneOf union (no $ref)

@app.post("/body")
def body(pet: Annotated[Pet, Body()]): ...   # {"$ref": "#/components/schemas/Pet"}
```

## Root cause

In `analyze_param` (`fastapi/dependencies/utils.py`), a `TypeAliasType` is unwrapped to `annotation.__value__` before analysis. That unwrap is needed to decide the parameter kind, but the unwrapped (anonymous) annotation was then *also* used to build the inferred body field, so Pydantic inlined it instead of emitting a named component.

## Fix

Keep a reference to the original alias before unwrapping and, in the inferred-body branch, use it as the field annotation so Pydantic emits a named `$ref`.

The change is narrowly scoped to the inferred-body path:

- Aliases that embed a FastAPI annotation (`Query`/`Body`/`Depends`) set `field_info`/`depends` earlier and never reach this branch, so they are unaffected.
- Plain (non-`TypeAliasType`) annotated unions still inline as before.

## Tests

Added `test_pep695_type_alias_body_keeps_named_ref` to `tests/test_dependency_pep695.py`: the bare alias and `Annotated[Pet, Body()]` forms now produce the same `$ref`, and the discriminated union still validates and serializes correctly (`cat`/`dog` → 200, unknown tag → 422).

The full test suite, `mypy`, `ty`, and `ruff` all pass.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -401,8 +401,11 @@ def analyze_param(
     depends = None
     type_annotation: Any = Any
     use_annotation: Any = Any
+    type_alias_annotation = None
     if is_typealiastype(annotation):
-        # unpack in case PEP 695 type syntax is used
+        # Unpack PEP 695 type aliases for the analysis below, but keep the original
+        # alias so an inferred body can still produce a named OpenAPI component.
+        type_alias_annotation = annotation
         annotation = annotation.__value__
     if annotation is not inspect.Signature.empty:
         use_annotation = annotation
@@ -512,6 +515,10 @@ def analyze_param(
         ) or is_uploadfile_sequence_annotation(type_annotation):
             field_info = params.File(annotation=use_annotation, default=default_value)
         elif not field_annotation_is_scalar(annotation=type_annotation):
+            if type_alias_annotation is not None:
+                # Use the original PEP 695 alias as the body annotation so it yields a
+                # named OpenAPI component ($ref), like `Annotated[Alias, Body()]` (#15855).
+                use_annotation = type_alias_annotation
             field_info = params.Body(annotation=use_annotation, default=default_value)
         else:
             field_info = params.Query(annotation=use_annotation, default=default_value)
--- a/tests/test_dependency_pep695.py
+++ b/tests/test_dependency_pep695.py
@@ -1,7 +1,8 @@
-from typing import Annotated
+from typing import Annotated, Literal
 
-from fastapi import Depends, FastAPI
+from fastapi import Body, Depends, FastAPI
 from fastapi.testclient import TestClient
+from pydantic import BaseModel, Field
 from typing_extensions import TypeAliasType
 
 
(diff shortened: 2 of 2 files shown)


YuriiMotov:
@iamsharduld, please follow the PR template, and don't open a PR unless a maintainer asks you to.

We'll review the discussion when we have time.

PR #15858, opened 2026-06-28
🐛 Fix Server-Sent Event data ignoring Pydantic serialization_alias

iamsharduld:
## Description

Reported in [#15703](https://github.com/fastapi/fastapi/discussions/15703).

When a Pydantic model is sent as the `data` of a `ServerSentEvent`, its `serialization_alias` (and `alias`) were ignored:

```python
class Item(BaseModel):


async def events():
    yield ServerSentEvent(data=Item(name="Portal Gun"))

# before: data: {"name":"Portal Gun"}
# after:  data: {"itemName":"Portal Gun"}
```

## Root cause

In `get_request_handler` (`fastapi/routing.py`), `ServerSentEvent.data` was serialized with `item.data.model_dump_json()`, which defaults to `by_alias=False` in Pydantic v2 — so aliases were dropped. Every other response-serialization path in FastAPI threads `response_model_by_alias` (default `True`).

## Fix

Serialize the model with `by_alias=response_model_by_alias`. This honors `serialization_alias`/`alias` by default and respects a route-level `response_model_by_alias=False`, consistent with plainly-yielded models and regular JSON responses. Only the Pydantic-model branch changes; `dict`, `raw_data`, and other values are untouched.

## Tests

Added two tests to `tests/test_sse.py`:

- `test_sse_event_data_uses_serialization_alias` — aliases are honored by default (fails without the fix).
- `test_sse_event_data_respects_response_model_by_alias` — `response_model_by_alias=False` disables aliasing for SSE event data too, consistent with plainly-yielded items.

The full test suite, `mypy`, `ty`, and `ruff` all pass.

## Note

This is a behavior change: code that yields `ServerSentEvent(data=<model with aliases>)` will now emit aliased field names by default, aligned with FastAPI's global `response_model_by_alias=True` default.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -523,7 +523,13 @@ def get_request_handler(
                             data_str: str | None = item.raw_data
                         elif item.data is not None:
                             if hasattr(item.data, "model_dump_json"):
-                                data_str = item.data.model_dump_json()
+                                # Thread `response_model_by_alias` (default `True`)
+                                # so Pydantic `serialization_alias`/`alias` are
+                                # honored, like the rest of FastAPI's response
+                                # serialization. See #15703.
+                                data_str = item.data.model_dump_json(
+                                    by_alias=response_model_by_alias
+                                )
                             else:
                                 data_str = json.dumps(jsonable_encoder(item.data))
                         else:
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Duplicate of https://github.com/fastapi/fastapi/pull/15708

PR #15867, opened 2026-06-29
Validate event and id fields in format_sse_event

raghavc:
## Summary

`ServerSentEvent` already validates SSE `event` and `id` fields before rendering. This applies the same checks in the public `format_sse_event()` helper so callers cannot accidentally render multiline `event` or `id` values into extra SSE fields.

Discussion: https://github.com/fastapi/fastapi/discussions/15648

## Changes

- Reuse the existing SSE validators at the start of `format_sse_event()`.
- Add regression tests for multiline `event`, multiline `id`, null-byte `id`, an injected `data:` line, and a valid `event`/`id` output.

## Validation

- `pytest -q tests/test_sse.py`
- `pytest -q tests/test_sse.py tests/test_tutorial/test_server_sent_events`
- `ruff check fastapi/sse.py tests/test_sse.py`
- `git diff --check`

### DIFF
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -203,6 +203,9 @@ def format_sse_event(
 
     The result always ends with `\n\n` (the event terminator).
     """
+    _check_event_single_line(event)
+    _check_id_valid(id)
+
     lines: list[str] = []
 
     if comment is not None:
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@raghavc, thanks for your interest and efforts, but I think we don't need to add this validation.
See https://github.com/fastapi/fastapi/discussions/15648#discussioncomment-17705560

PR #15869, opened 2026-06-30
fix: preserve model_fields_set in Form models

aoright:
This PR fixes issue #13399 where `model_fields_set` in Form models was not preserved correctly because FastAPI pre-populated missing fields with their default values in the dictionary passed to Pydantic for validation.

### Changes
1. Modified `_get_multidict_value` in `fastapi/dependencies/utils.py` to return `None` (instead of the default value) for missing fields when processing `FormData`.
2. This allows Pydantic to apply the defaults itself during model validation, which correctly excludes them from `model_fields_set`.
3. Non-form parameters (Query, Header, Cookie) continue to use their existing default pre-population behavior to preserve backwards compatibility.
4. Added a comprehensive test suite in `tests/test_forms_fields_set.py` to verify that `model_fields_set` is correct for empty, partial, and fully-specified form data submissions.
5. Updated input validation tests in `tests/test_forms_single_model.py`.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -777,6 +777,8 @@ def _get_multidict_value(
         if field.field_info.is_required():
             return
         else:
+            if isinstance(values, FormData):
+                return
             return deepcopy(field.default)
     return value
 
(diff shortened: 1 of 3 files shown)


YuriiMotov:
This is a duplicate of https://github.com/fastapi/fastapi/pull/14574

PR #15904, opened 2026-07-01
fix: preserve empty form values for required params

PreciousNwakama:
## Pull Request

Discussion: 

## Description
Fixes FastAPI form parsing so required `Form()` parameters preserve an explicit empty string (`""`) instead of being treated as missing.
- Updated `fastapi/dependencies/utils.py` to keep `""` as a valid value for required form parameters.
- Added a regression test in `tests/test_form_default.py` covering a required form field submitted with an empty string.

This is a targeted bug fix with test coverage for the regression.

## AI Disclaimer

No AI was used to generate this code change.

<details>

</details>

## Checklist

- [X] I added tests for the change in tests/test_form_default.py.
- [ ] Coverage stays at 100%.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -763,8 +763,7 @@ def _get_multidict_value(
     else:
         value = values.get(alias, None)
     if value is None or (
-        field_annotation_is_sequence(field.field_info.annotation)
-        and len(value) == 0
+        field_annotation_is_sequence(field.field_info.annotation) and len(value) == 0
     ):
         if field.field_info.is_required():
             return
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@PreciousNwakama, thanks for your interest, but we don't accept user PRs that were not explicitly requested by maintainers. See https://tiangolo.com/open-source/contributing/#start-with-a-discussion

We are aware about this issue and will fix it when we have time for it.
I previously opened PR https://github.com/fastapi/fastapi/pull/14829 that adds a set of tests, including the use case from this PR. This set of tests is needed in order to avoid breaking other use cases while fixing this one.
You are welcome to review PR https://github.com/fastapi/fastapi/pull/14829

PR #15915, opened 2026-07-02
🐛 Fix `response_model_include`/`response_model_exclude` being ignored for sequence response models

serverestaa:
## Problem

When the response model is a sequence such as `list[User]`, `response_model_exclude` is silently ignored — the excluded fields (e.g. passwords) leak into the response — and `response_model_include` drops all items, returning `[]`.

The response is serialized with `TypeAdapter.dump_json()`/`dump_python()`, and Pydantic interprets top-level `include`/`exclude` keys on a sequence schema as item indices, so field names like `{"password"}` match nothing. The previous `jsonable_encoder` serialization path applied `include`/`exclude` to each item, so this worked on FastAPI 0.115 with Pydantic v1.

## Reproduction

```python
from fastapi import FastAPI

class User(BaseModel):
    name: str = "n"
    password: str = "secret"

app = FastAPI()

@app.get("/users", response_model=list[User], response_model_exclude={"password"})
def users():
    return [User()]

print(TestClient(app).get("/users").json())
# [{'name': 'n', 'password': 'secret'}]  <- password leaked
# with response_model_include={"name"} instead: []  <- all data dropped
```

## Fix

In `serialize_response()`, when the validated response value is a sequence, wrap field-name-only `include`/`exclude` values in `{"__all__": ...}` so they apply to every item, matching Pydantic's `IncEx` semantics for sequences. Values that already use item indices or `"__all__"` explicitly are left untouched, so existing workarounds keep working.

Note that the check is based on the validated response value rather than the declared response model, so it also covers cases like `Union[list[User], User]` (and e.g. `response_model=Any` returning a list of models, where `include`/`exclude` were previously silently ignored as well). `str`/`bytes` values are explicitly excluded from the sequence check.

Tests added to `tests/test_response_model_include_exclude.py` cover `list[Model]` include/exclude (flat and nested), an explicit `"__all__"` exclude, and a custom `response_class` (the non-`dump_json` serialization path).

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -45,6 +45,7 @@ from fastapi._compat import (
     ModelField,
     Undefined,
     lenient_issubclass,
+    value_is_sequence,
 )
 from fastapi.datastructures import Default, DefaultPlaceholder
 from fastapi.dependencies.models import Dependant
@@ -290,6 +291,16 @@ def _extract_endpoint_context(func: Any) -> EndpointContext:
     return ctx
 
 
+def _wrap_incex_for_sequence(incex: IncEx | None) -> IncEx | None:
+    # Pydantic interprets top-level include/exclude keys on a sequence type as
+    # item indices, so field names would be silently ignored. Wrap them in
+    # "__all__" to apply them to every item, unless indices or "__all__" are
+    # already used explicitly.
+    if incex and all(isinstance(key, str) and key != "__all__" for key in incex):
+        return {"__all__": incex}
+    return incex
+
+
 async def serialize_response(
     *,
     field: ModelField | None = None,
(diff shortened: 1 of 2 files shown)


YuriiMotov (commented):

I think the problem is overstated in this PR.
Yes, developers can mistakenly specify `response_model_exclude={"password"}` for `response_model=list[User]`, but this is basically their mistake - correct syntax for lists described in Pydantic docs is `{"__all__": {"password"}}` (see [docs](https://pydantic.dev/docs/validation/latest/concepts/serialization#excluding-and-including-specific-fields)).
And, for such security-sensitive things developers should definitely write tests. These tests will expose their mistake quickly.

I think we should stick to the syntax described in Pydantic's docs.

I will close this PR. If you still think this is an issue and it should be fixed - please open discussion and let's collect community feedback first

PR #15918, opened 2026-07-02
🐛 Fix `jsonable_encoder` not encoding `Enum` values that are not JSON-serializable

serverestaa:
## Problem

`jsonable_encoder` returns `Enum.value` as-is, without encoding it. For enums whose values are not JSON primitives (`date`, `bytes`, `Decimal`, ...), the raw value reaches `JSONResponse.render`, where `json.dumps` raises `TypeError` and the request fails with a 500. Every other branch of `jsonable_encoder` (models, dataclasses, dicts, sequences) encodes contained values recursively, and Pydantic's `model_dump(mode="json")` handles the same enums correctly — only the bare `Enum` branch skipped the recursion.

## Reproduction

```python
import datetime
from enum import Enum

from fastapi import FastAPI


class DateEnum(Enum):
    D = datetime.date(2024, 1, 1)


app = FastAPI()


@app.get("/enum")
def get_enum():
    return {"v": DateEnum.D}


client = TestClient(app)
response = client.get("/enum")  # 500: Object of type date is not JSON serializable
```

Directly, `jsonable_encoder(DateEnum.D)` returns `datetime.date(2024, 1, 1)` instead of `"2024-01-01"`.

## Fix

In the `Enum` branch of `jsonable_encoder`, encode `obj.value` recursively with `jsonable_encoder`, passing through `custom_encoder` and `sqlalchemy_safe`, matching how the other branches handle nested values. Enums with primitive values (`str`, `int`, ...) behave exactly as before; enums with values like `date`, `Decimal`, or `bytes` are now encoded to their JSON-compatible representation, consistent with Pydantic's `mode="json"` serialization.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -271,7 +271,11 @@ def jsonable_encoder(
             sqlalchemy_safe=sqlalchemy_safe,
         )
     if isinstance(obj, Enum):
-        return obj.value
+        return jsonable_encoder(
+            obj.value,
+            custom_encoder=custom_encoder,
+            sqlalchemy_safe=sqlalchemy_safe,
+        )
     if isinstance(obj, PurePath):
         return str(obj)
     if isinstance(obj, (str, int, float, type(None))):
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Do you have a specific use case where it hit you?

Please, follow the PR description template and refrain from opening PRs without discussing this in advance. You should've opened discussion and wait for maintainers to validate the idea first.

PR #15922, opened 2026-07-03
📝 docs: add docstring to contextmanager_in_threadpool

tono9876:
Added missing docstring to contextmanager_in_threadpool in fastapi/concurrency.py.

### DIFF
--- a/fastapi/concurrency.py
+++ b/fastapi/concurrency.py
@@ -17,10 +17,10 @@ _T = TypeVar("_T")
 @asynccontextmanager
 async def contextmanager_in_threadpool(
     """Run a context manager in a thread pool.
-    
+
         Args:
             cm: The context manager to run.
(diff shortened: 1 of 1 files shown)


tono9876:
🔧 **Fixed!** The docstring was incorrectly placed between function parameters in the first commit — now correctly positioned inside the function body.

```python
async def contextmanager_in_threadpool(
    cm: AbstractContextManager[_T],
) -> AsyncGenerator[_T, None]:
    """Run a context manager in a thread pool.
    
    Allows using synchronous context managers within async
    FastAPI endpoints without blocking the event loop.
    """
    ...
```

CI checks should pass now. Thanks! 🙏

YuriiMotov:
Thanks, but this is not needed - this is internal function and the name is clear

PR #15930, opened 2026-07-04
📝 Add API reference page for fastapi.sse

SAURABHSALVE:
The new SSE support (`EventSourceResponse` and `ServerSentEvent` in `fastapi.sse`) is already covered in the tutorial, but it doesn't have a dedicated page in the API Reference like the other public modules (such as `responses`, `websockets`, and `templating`).

This PR adds a new `docs/en/docs/reference/sse.md` page following the same pattern as the existing reference pages. It includes `:::` directives for `EventSourceResponse` and `ServerSentEvent`, and adds the page to the navigation alongside the other response-related reference pages.

I didn't include `format_sse_event` for now because the tutorial currently documents only the two public classes. If you'd prefer it to be part of the API reference as well, I'm happy to add it.

I also verified that the documentation builds successfully locally without any warnings, and the generated page correctly picks up the existing docstrings and `Doc()` annotations.

### DIFF
--- a/docs/en/docs/reference/sse.md
+++ b/docs/en/docs/reference/sse.md
@@ -1,0 +1,17 @@
+# Server-Sent Events - `EventSourceResponse` and `ServerSentEvent`
+
+To stream Server-Sent Events (SSE), use `yield` in your *path operation function* and set `response_class=EventSourceResponse`.
+
+If you need to set SSE fields like `event`, `id`, `retry`, or `comment`, you can `yield` `ServerSentEvent` objects instead of plain data.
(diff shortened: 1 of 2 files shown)


YuriiMotov (approved):
LGTM

tiangolo (approved):
Nice, thanks! 🍰

SAURABHSALVE:
> Nice, thanks! 🍰

@tiangolo Thank you, Sebastian! Glad I could contribute, and I really appreciate the review and the merge. 🚀

PR #15937, opened 2026-07-05
🐛 Fix `status_code` being ignored for SSE and JSONL streaming endpoints

SAURABHSALVE:
SSE and JSONL streaming endpoints ignore the `status_code` declared on the path operation and any status code set by a dependency on the `Response` parameter, always returning 200 while the generated OpenAPI documents the declared code. That makes the schema contradict the actual behavior.

```python
@app.post(/events, response_class=EventSourceResponse, status_code=201)
async def events() -> AsyncIterable[str]:
    yield created
# returns 200, OpenAPI documents 201
```

Raw streaming responses (`response_class=StreamingResponse`) already handle both declared and dependency-set status codes via `_build_response_args()`.

This PR applies the same `_build_response_args()` path to the SSE and JSONL branches. Default behavior is unchanged because the dependency solver's temporary response starts with `status_code = None`, so no status is injected unless the user declared or set one.

Adds `tests/test_stream_status_code.py` covering:

- declared status codes for SSE and JSONL streams
- dependency-set status codes for SSE and JSONL streams
- raw streaming as the existing control case
- default 200 behavior for both stream encodings
- OpenAPI status-code consistency

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -624,10 +624,13 @@ def get_request_handler(
                     _sse_with_checkpoints(sse_receive_stream)
                 )
 
+                response_args = _build_response_args(
+                    status_code=status_code, solved_result=solved_result
+                )
                 response = StreamingResponse(
                     sse_stream_content,
                     media_type="text/event-stream",
-                    background=solved_result.background_tasks,
+                    **response_args,
                 )
                 response.headers["Cache-Control"] = "no-cache"
                 # For Nginx proxies to not buffer server sent events
@@ -660,10 +663,13 @@ def get_request_handler(
 
                     jsonl_stream_content = _sync_stream_jsonl()
 
+                response_args = _build_response_args(
+                    status_code=status_code, solved_result=solved_result
+                )
                 response = StreamingResponse(
                     jsonl_stream_content,
                     media_type="application/jsonl",
-                    background=solved_result.background_tasks,
+                    **response_args,
                 )
                 response.headers.raw.extend(solved_result.response.headers.raw)
             elif dependant.is_async_gen_callable or dependant.is_gen_callable:
--- a/tests/test_stream_status_code.py
+++ b/tests/test_stream_status_code.py
@@ -1,10 +1,50 @@
-import json
 from collections.abc import AsyncIterable
-from typing import Any, cast
 
+import pytest
 from fastapi import Depends, FastAPI, Response
 from fastapi.responses import EventSourceResponse, StreamingResponse
 from fastapi.testclient import TestClient
+from inline_snapshot import snapshot
+
+SSE_RESPONSE = {
+    "description": "Successful Response",
+    "content": {
+        "text/event-stream": {
+            "itemSchema": {
+                "type": "object",
+                "properties": {
+                    "data": {
+                        "type": "string",
+                        "contentMediaType": "application/json",
+                        "contentSchema": {
+                            "type": "object",
+                            "additionalProperties": {"type": "string"},
+                            "title": "SSE stream item",
+                        },
+                    },
+                    "event": {"type": "string"},
+                    "id": {"type": "string"},
+                    "retry": {"type": "integer", "minimum": 0},
+                },
+                "required": ["data"],
+            }
+        }
+    },
+}
+
+JSONL_RESPONSE = {
+    "description": "Successful Response",
+    "content": {
+        "application/jsonl": {
+            "itemSchema": {
+                "type": "object",
+                "additionalProperties": {"type": "string"},
+                "title": "JSONL stream item",
+            }
+        }
+    },
+}
+
 
 app = FastAPI()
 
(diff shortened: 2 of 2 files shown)


SAURABHSALVE:
Hi maintainers, all code/test checks are passing now. The only failing required check is `Labels / check-labels` because this PR has no required category label.
Could someone with triage permissions please add the `bug` label?
This PR fixes a runtime bug where SSE and JSONL streaming endpoints return `200` even when the path operation declares another `status_code` or a dependency sets `Response.status_code`, while OpenAPI documents the non-200 status.

BitWeaverDev:
Really nice catch, and the fix is clean since it just reuses _build_response_args, the same helper the raw streaming branch already relied on, instead of introducing new logic.
Small suggestion on the test coverage. Going through _build_response_args, the behavior this fix leans on is that a dependency set response.status_code always takes precedence over the declared status_code on the decorator (the last if solved_result.response.status_code: check is what does it). Right now the declared case and the dependency case are each tested on their own, but there isn't a test where both are set on the same route to confirm the dependency wins.
Something like this might be worth adding:

```python
@app.post("/sse-created-override", response_class=EventSourceResponse, status_code=201)
async def sse_created_override(
    accepted: None = Depends(set_accepted),
) -> AsyncIterable[dict[str, str]]:
    yield {"message": "overridden"}
```

with an assertion that the response comes back as 202 rather than 201. Could be mirrored for the JSONL case as well.
Worth noting the raw streaming control test has the same gap already, so this isn't something the PR introduces, it's just the one precedence path that feels most worth locking down given it's the core logic the whole fix depends on. Curious what the maintainers think, happy to be wrong here.

YuriiMotov:
@SAURABHSALVE, will you be able to add tests suggested by @BitWeaverDev?

SAURABHSALVE:
> @SAURABHSALVE, will you be able to add tests suggested by @BitWeaverDev?

Thanks for the great suggestion. I just pushed the new tests covering the cases where the dependency overrides the declared status code for both SSE and JSONL. Let me know if everything looks good to go!

SAURABHSALVE:
Spot on, @BitWeaverDev! Great catch. I just pushed the override tests for SSE, JSONL, and included the raw streaming one too to close that gap. Thanks for the review!

YuriiMotov:
@SAURABHSALVE, why did you add the `NEW OVERRIDE ENDPOINTS` comment?

SAURABHSALVE:
> @SAURABHSALVE, why did you add the `NEW OVERRIDE ENDPOINTS` comment?

Ah, my apologies! I added those comments as personal visual markers while I was structuring the file locally and accidentally left them in before committing. I will remove them right away and push the clean file!

SAURABHSALVE:
@YuriiMotov Alright, just pushed the fix to remove those. Thanks for bearing with me!

YuriiMotov:
@SAURABHSALVE, I just pushed some changes - this is how I suggest we modify tests.

Main things:
* removed tests for **default** endpoints - it's already testes, no need to duplicate it
* simplified similar tests using `@pytest.mark.parametrize`
* specified `responses` for endpoints that set response status code in dependency - this is to make it closer to real use case (without it openapi schema would not contain response with the status code that is being returned)
* test full openapi schema - testing just specific fields can mask the issue in schama

What do you think?

SAURABHSALVE:
@YuriiMotov 
Thank you for pushing those changes directly the improvements are all better than my original approach.

The @pytest.mark.parametrize refactor is much cleaner for the 9 status-code cases, removing the default tests avoids duplication since they're already covered elsewhere, and the inline_snapshot of the full OpenAPI schema is the right way to lock this down (I should have used that pattern from the start rather than asserting on specific keys).
Adding responses={202: ...} on the dependency routes is a good call too it makes the test closer to how these endpoints would actually be documented in real code, and it properly exercises the schema for the non-default status.
Happy with the current state of the branch. Ready to merge whenever you are.

YuriiMotov (approved):
LGTM!

On `master` 6 of these tests fail:
```
FAILED tests/test_stream_status_code.py::test_status_code[/sse-201] - assert 200 == 201
FAILED tests/test_stream_status_code.py::test_status_code[/jsonl-201] - assert 200 == 201
FAILED tests/test_stream_status_code.py::test_status_code[/sse-dependency-202] - assert 200 == 202
FAILED tests/test_stream_status_code.py::test_status_code[/jsonl-dependency-202] - assert 200 == 202
FAILED tests/test_stream_status_code.py::test_status_code[/sse-dependency-override-202] - assert 200 == 202
FAILED tests/test_stream_status_code.py::test_status_code[/jsonl-dependency-override-202] - assert 200 == 202
```

---

SAURABHSALVE:
@YuriiMotov Thanks for verifying against master and confirming the regression that's exactly the behaviour the fix addresses. Happy to wait for Sebastian's review.

tiangolo (approved):
Thank you! 🙌 

This will be available in FastAPI 0.140.13, released in the next few hours. 🚀

SAURABHSALVE:
@tiangolo Thank you, Sebastian! Really appreciate the thoughtful review and the merge. 🙌

PR #15948, opened 2026-07-07
fix(sse): emit data field for empty SSE data payload

devteamaegis:
## What's broken
`format_sse_event()` drops the `data:` field entirely when the data payload is an empty string. It builds data lines via `data_str.splitlines()`, and `"".splitlines()` returns `[]`, so the loop never runs. Yielding `ServerSentEvent(raw_data="")` from an SSE endpoint produces a malformed event with no `data:` line and a single `\n` instead of the documented `\n\n` terminator, so clients can't recognize it as a discrete event.

```python
format_sse_event(data_str="")   # before: b'\n'   (expected b'data: \n\n')
```

## Why it happens
`str.splitlines()` returns `[]` for `""`, conflating an empty-but-present data payload with the "no data" (`None`) case, even though they're distinct.

## Fix
Emit a single empty `data:` line when the payload is the empty string (`data_str.splitlines() or [""]`), leaving the `None` case untouched.

## Test
Added `test_format_sse_event_empty_data_str` asserting `format_sse_event(data_str="") == b"data: \n\n"`. Fails before, passes after; full SSE suite (25) green.

### DIFF
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -213,7 +213,9 @@ def format_sse_event(
         lines.append(f"event: {event}")
 
     if data_str is not None:
-        for line in data_str.splitlines():
+        # An empty data payload must still emit a single `data:` field;
+        # `"".splitlines()` returns `[]`, which would drop the field entirely.
+        for line in data_str.splitlines() or [""]:
             lines.append(f"data: {line}")
 
     if id is not None:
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Thanks!

This is already covered by PR https://github.com/fastapi/fastapi/pull/15515 that is waiting for the final review by Sebastian.

PR #15953, opened 2026-07-07
📝 Document custom headers for JSONL and SSE streams

Deadshot1831:
## Pull Request

Discussion: N/A - docs-only clarification with tutorial test coverage.

## Description

This PR documents how to add custom response headers to JSON Lines and SSE streaming endpoints.

FastAPI already documents the common pattern in [Response Headers](https://fastapi.tiangolo.com/advanced/response-headers/): use a `Response` parameter and mutate `response.headers`. This PR adds the streaming-specific caveat. For generator endpoints, setting a header inside the generator body is too late because FastAPI has already prepared the streaming response before the generator starts yielding data.

The examples use a dependency with a `Response` parameter to set the header before FastAPI creates the streaming response. Dependencies are resolved before the response is prepared, so the header is available at the correct point.

Changes:

- Add a JSON Lines tutorial section and runnable docs example for custom stream headers.
- Add an SSE tutorial section and runnable docs example for custom stream headers.
- Add tutorial tests that verify the custom header is present, the response media type is still correct, the stream body is produced, and OpenAPI keeps the expected streaming schema.

## Validation

```text
python -m pytest tests/test_tutorial/test_stream_json_lines tests/test_tutorial/test_server_sent_events tests/test_sse.py tests/test_stream_bare_type.py tests/test_stream_json_validation_error.py -q
52 passed

python -m ruff check docs_src/stream_json_lines/tutorial002_py310.py docs_src/server_sent_events/tutorial006_py310.py tests/test_tutorial/test_stream_json_lines/test_tutorial002.py tests/test_tutorial/test_server_sent_events/test_tutorial006.py
All checks passed

git diff --check
```

I did not run the full docs build locally because this venv does not have the full docs dependency group installed.

### DIFF
--- a/docs/en/docs/tutorial/server-sent-events.md
+++ b/docs/en/docs/tutorial/server-sent-events.md
@@ -69,6 +69,18 @@ You can also omit the return type. FastAPI will use the [`jsonable_encoder`](./e
 
 {* ../../docs_src/server_sent_events/tutorial001_py310.py ln[34:37] hl[35] *}
 
+## Custom Response Headers { #custom-response-headers }
+
+For normal responses, you can set custom headers using a `Response` parameter, as explained in [Response Headers](../advanced/response-headers.md#use-a-response-parameter).
+
+With an SSE stream, the *path operation function* is a generator. Calling it creates the generator, but the code inside the generator body doesn't run until the response is already streaming. So setting headers inside the generator body is too late for headers that need to be sent with the response.
+
+Instead, use a dependency that receives the `Response` parameter and sets the headers before the stream starts:
+
+{* ../../docs_src/server_sent_events/tutorial006_py310.py ln[1:33] hl[3:4,22:23,26:30] *}
+
+Dependencies run before FastAPI creates the streaming response, so FastAPI can copy those headers into the final SSE response.
+
 ## `ServerSentEvent` { #serversentevent }
 
 If you need to set SSE fields like `event`, `id`, `retry`, or `comment`, you can yield `ServerSentEvent` objects instead of plain data.
--- a/docs/en/docs/tutorial/stream-json-lines.md
+++ b/docs/en/docs/tutorial/stream-json-lines.md
@@ -106,6 +106,18 @@ You can also omit the return type. FastAPI will then use the [`jsonable_encoder`
 
 {* ../../docs_src/stream_json_lines/tutorial001_py310.py ln[33:36] hl[34] *}
 
+## Custom Response Headers { #custom-response-headers }
+
+For normal responses, you can set custom headers using a `Response` parameter, as explained in [Response Headers](../advanced/response-headers.md#use-a-response-parameter).
+
+With a JSON Lines stream, the *path operation function* is a generator. Calling it creates the generator, but the code inside the generator body doesn't run until the response is already streaming. So setting headers inside the generator body is too late for headers that need to be sent with the response.
+
+Instead, use a dependency that receives the `Response` parameter and sets the headers before the stream starts:
+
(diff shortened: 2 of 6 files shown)


docs/en/docs/tutorial/server-sent-events.md
```diff
@@ -69,6 +69,16 @@ You can also omit the return type. FastAPI will use the [`jsonable_encoder`](./e
...
 
+## Custom Response Headers { #custom-response-headers }
+
+If you need to add custom headers to an SSE response, use a dependency that receives a `Response` parameter and sets the headers before the stream starts:
+
+{* ../../docs_src/server_sent_events/tutorial006_py310.py ln[1:33] hl[3:4,22:23,26:30] *}
+
+The dependency runs before FastAPI creates the streaming response.
+
+This is different from setting headers inside the generator body, which runs later, when the response is already streaming.
```
YuriiMotov: I think this should be explained in a different way.

I would go with:
* Start from reference to https://fastapi.tiangolo.com/advanced/response-headers
* Explain why that common approach doesn't work here
* Introduce the solution with dependency and explain why it works


What do you think?

Deadshot1831: Yes, that makes sense. I was focusing more on the final pattern, but starting from the existing Response Headers docs makes the flow clearer.
I’ll adjust the explanation so it first points to the usual Response parameter approach, then explains why setting headers inside a streaming generator is too late, and then introduces the dependency approach. The key reason that works is that dependencies run before FastAPI creates the streaming response, so the headers are set before streaming starts.

Deadshot1831: I have made the said changes

Deadshot1831:
https://github.com/fastapi/fastapi/commit/a9b2d8721c6046bb84e68743f01a51616311072f
made the changes via this commit waiting for approval

docs/en/docs/tutorial/server-sent-events.md
```diff
@@ -69,6 +69,18 @@ You can also omit the return type. FastAPI will use the [`jsonable_encoder`](./e
...
+
+For normal responses, you can set custom headers using a `Response` parameter, as explained in [Response Headers](../advanced/response-headers.md#use-a-response-parameter).
+
+With an SSE stream, the *path operation function* is a generator. Calling it creates the generator, but the code inside the generator body doesn't run until the response is already streaming. So setting headers inside the generator body is too late for headers that need to be sent with the response.
+
+Instead, use a dependency that receives the `Response` parameter and sets the headers before the stream starts:
+
+{* ../../docs_src/server_sent_events/tutorial006_py310.py ln[1:33] hl[3:4,22:23,26:30] *}
+
+Dependencies run before FastAPI creates the streaming response, so FastAPI can copy those headers into the final SSE response.
```
YuriiMotov, line 82:
```suggestion
With an SSE stream, the *path operation function* is a generator. FastAPI starts the response before the body of that generator is executed. Because the response headers have already been sent, modifying the `Response` object inside the endpoint function has no effect.

Instead, use a dependency that receives the `Response` parameter and sets the headers before the response starts:

{* ../../docs_src/server_sent_events/tutorial006_py310.py ln[1:33] hl[3:4,22:23,26:30] *}

Dependencies are resolved before FastAPI starts sending the response, so headers they set on the `Response` object are included in the outgoing response.
```

YuriiMotov (approved):
@Deadshot1831, thanks!

Your version is technically precise, but it seems a bit hard to read, so I prepared an alternative version that might be simpler for newcomers.
Please, take a look or we can just leave it for Sebastian to decide.

PR #15958, opened 2026-07-08
🐛 Log ResponseValidationError from streaming endpoint serializers

SAURABHSALVE:
﻿When a streaming endpoint (SSE or JSONL) yields an item that fails response_model validation, FastAPI raises ResponseValidationError inside the streaming iterator. Currently that error is silently swallowed: no log entry is produced and the client receives an empty 200 OK that is indistinguishable from a valid zero-item stream.

Minimal reproduction:

```python
from collections.abc import AsyncIterable
from fastapi import FastAPI
from fastapi.responses import EventSourceResponse

class Item(BaseModel):
    id: int

app = FastAPI()

async def events() -> AsyncIterable[Item]:
    yield {"id": "NOT_AN_INT"}   # ResponseValidationError

r = TestClient(app, raise_server_exceptions=False).get("/events")
assert r.status_code == 200 and r.content == b""
# No log, no 500, no way to detect the error in production
```

Fix: catch ResponseValidationError in _producer (SSE), _async_stream_jsonl, and _sync_stream_jsonl, call logger.exception() to emit an ERROR-level log with full traceback (matching behavior for non-streaming endpoints), then re-raise so the exception continues to propagate. Valid streams are unaffected; the only change is that the error becomes visible.

This is distinct from fixing exceptions before the first yield (a design question around keepalive timing). This is exclusively about serialization errors on already-yielded items, where no such tension exists.

Adds 10 tests covering async and sync generator variants for both SSE and JSONL, verifying: ERROR log is emitted with exc_info, exception still propagates, valid streams produce no spurious logs.

Checks:

- `python -m pytest tests/test_stream_response_validation_error.py`
- `.\.venv\Scripts\ruff.exe check fastapi/routing.py tests/test_stream_response_validation_error.py`
- `.\.venv\Scripts\ruff.exe format --check fastapi/routing.py tests/test_stream_response_validation_error.py`

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -58,6 +58,7 @@ from fastapi.exceptions import (
     ResponseValidationError,
     WebSocketRequestValidationError,
 )
+from fastapi.logger import logger
 from fastapi.sse import (
     _PING_INTERVAL,
     KEEPALIVE_COMMENT,
@@ -556,7 +557,14 @@ def get_request_handler(
                     async def _producer() -> None:
                         async with send_stream:
                             async for raw_item in sse_aiter:
-                                await send_stream.send(_serialize_sse_item(raw_item))
+                                try:
+                                    chunk = _serialize_sse_item(raw_item)
+                                except ResponseValidationError:
+                                    logger.exception(
+                                        "ResponseValidationError while serializing SSE stream item"
+                                    )
+                                    raise
+                                await send_stream.send(chunk)
 
                     send_keepalive, receive_keepalive = (
                         anyio.create_memory_object_stream[bytes](max_buffer_size=1)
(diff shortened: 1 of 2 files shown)


SAURABHSALVE:
Could a maintainer please assign the appropriate labels to this PR so the `check-labels` CI step can pass?

YuriiMotov:
I just attempted to reproduce it:
* copied code of the app from PR description (removed `TestClient`-related part)
* ran it with `uvicorn main:app`
* opened OpenAPI docs and executed the endpoint

In app logs I can see the exception is logged as expected.

But you said:

> Currently that error is silently swallowed: no log entry is produced ...

Could you please double-check this on minimal app? It was probably your logging settings?

SAURABHSALVE:
﻿Thanks for checking this. You are right.

I re-tested the minimal app in two ways:

1. `TestClient(app, raise_server_exceptions=False).get("/events")`
   - returns `200`
   - returns an empty body
   - does not emit a captured `fastapi`, `starlette`, or `uvicorn` log record in the test process

2. Running the same app with `uvicorn main:app` and calling `/events`
   - logs `ERROR: Exception in ASGI application`
   - includes the `ResponseValidationError` traceback from the SSE producer task

So the PR description was wrong to say this is invisible in normal Uvicorn/server logs. The no-log behavior I observed is specific to the `TestClient(..., raise_server_exceptions=False)` reproduction, not the production server path.

I also checked the branch with this patch applied, and Uvicorn then logs both the new FastAPI `logger.exception(...)` message and its existing ASGI exception log, so the change would duplicate logging for the normal live-server case.

Given that, I do not think this PR is correct as submitted. Closing it rather than trying to preserve a misleading fix. Sorry for the noise, and thanks again for catching it.

SAURABHSALVE:
Closing after re-checking: Uvicorn already logs the exception for the minimal live-server app, so the PR premise is incorrect and the patch would duplicate logging.

PR #15960, opened 2026-07-08
Fix outdated on_event reference in async tests docs

Aziztaamalli:
### Summary

The Async Tests page recommends instantiating event-loop-bound objects inside an `@app.on_event("startup")` callback to avoid a `RuntimeError: Task attached to a different loop`.

`on_event` is deprecated in favor of the `lifespan` parameter (see the Lifespan Events page). The same async-tests page already talks about "lifespan events" a few lines above (in the `AsyncClient` warning box), so the `on_event` mention is both outdated and inconsistent with the rest of the page.

This PR updates the tip to reference the `lifespan` handler instead, matching current FastAPI guidance.

### Checklist

- Verified `on_event` is documented as deprecated in `fastapi/routing.py`
- Verified no other English docs page has the same stale reference
- Ran the `typos` pre-commit hook against the changed file

### DIFF
--- a/docs/en/docs/advanced/async-tests.md
+++ b/docs/en/docs/advanced/async-tests.md
@@ -94,6 +94,6 @@ As the testing function is now asynchronous, you can now also call (and `await`)
 
 /// tip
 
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Thanks for your interest!

We track this internally. We plan to address this in a different way.

Also, we are trying to keep things clean and ask to only open PRs when it's explicitly requested by maintainers.

>

Thanks for understanding!

PR #15967, opened 2026-07-09
📝 Fix links in docs

YuriiMotov:
## Description

Fix links in docs that return 404, point to incorrect content, or redirect.
As for redirects - I fixed only part of them that I think are more important. Stopped to prevent the diff of this PR to become too big)

**2 more links to fix:**
* `https://www.techempower.com/benchmarks/#section=test&runid=7464e520-0dc2-473d-bd34-dbdfd7e85911&hw=ph&test=query&l=zijzen-7` ([source](https://github.com/fastapi/fastapi/blob/b1346bb142c8154950953255eea535e5a83d62eb/docs/en/docs/benchmarks.md?plain=1#L3)) - I din't find the replacement.
* `https://github.com/encode/starlette#performance` ([source](https://github.com/fastapi/fastapi/blob/b1346bb142c8154950953255eea535e5a83d62eb/docs/en/docs/features.md?plain=1#L168)) - no such anchor. Current version of page doesn't state this anymore. I think we should just remove this link

**Note:**
There are also some more redirects (see in the details below), but I think it's Ok to leave them as is.
Let me know if you prefer to fix them all.

<details>

Status code: 301
  - https://www.linkedin.com/in/tiangolo/
  - https://docs.render.com/deploy-fastapi
  - https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
  - https://www.youtube.com/watch
  - https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests
  - https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT
  - https://www.jetbrains.com/research/python-developers-survey-2018/#development-tools
  - https://docs.docker.com/compose/faq/#why-do-my-services-take-10-seconds-to-recreate-or-stop
  - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
  - https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH
  - https://www.jetbrains.com/research/python-developers-survey-2017/#tools-and-features
  - https://docs.traefik.io/
  - https://tools.ietf.org/html/rfc6455#section-7.4.1
  - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
  - https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST
  - https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#credentialed_requests_and_wildcards
 
Status code: 302
  - https://docs.celeryq.dev
  - https://flask.palletsprojects.com
  - https://pytest.org
  - https://docs.pytest.org/
  - https://jinja.palletsprojects.com
  - https://hypercorn.readthedocs.io/
  - https://requests.readthedocs.io
  - https://requests.readthedocs.io/

Status code: 307
  - https://jwt.io/
  - https://netflixtechblog.com/introducing-dispatch-da4b8a2a8072

Status code: 308
  - https://heyapi.dev/openapi-ts/output
  - https://www.cursor.com/
  - https://heyapi.dev/openapi-ts/get-started
  - https://modelcontextprotocol.io
  - https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.5

</details>

## AI Disclaimer

Didn't use AI this time


docs/en/data/sponsors.yml
```diff
@@ -61,6 +61,6 @@ bronze:
   # - url: https://testdriven.io/courses/tdd-fastapi/
   #   title: Learn to build high-quality web apps with best practices
   #   img: /img/sponsors/testdriven.svg
-  - url: https://www.testmu.ai/?utm_source=fastapi&utm_medium=partner&utm_campaign=sponsor&utm_term=opensource&utm_content=webpage
+  - url: https://www.testmuai.com/?utm_source=fastapi&utm_medium=partner&utm_campaign=sponsor&utm_term=opensource&utm_content=webpage
```
YuriiMotov, line 64: I don't know if we should ask Testmu AI to confirm this replacement?

dannymaaz:
Thanks for documenting the two remaining links. I checked both against the current upstream pages.

For `docs/en/docs/benchmarks.md`, the old TechEmpower URL is tied to a run ID that is no longer available. A more durable replacement would be:

```md
[one of the fastest Python frameworks available](https://www.techempower.com/benchmarks/)
```

This avoids pinning the documentation to a specific benchmark run while still pointing to the official results page.

For `docs/en/docs/features.md`, the current Starlette repository no longer has a `#performance` section. I agree that the broken link should be removed. Another option that keeps the statement sourced would be to point it to FastAPI's own benchmark explanation:

```md
* Seriously impressive performance. It is [one of the fastest Python frameworks available, on par with **NodeJS** and **Go**](benchmarks.md).
```

Using the local documentation page would also be less likely to break if Starlette reorganizes its README again.

tiangolo (approved):
Excellent, thank you! 🙌 

I'll check the other two links later, but let's get this one in.

PR #15970, opened 2026-07-10
✅ Add tests for contextmanager_in_threadpool

nekoClangUage:
## Pull Request

Discussion: N/A
N/A - not a bug report. Adding test coverage for an internal function with
non-trivial exception-handling logic that currently has no direct tests.
Happy to open a Discussion first if the team prefers that workflow.

## Description

`contextmanager_in_threadpool` in `fastapi/concurrency.py` wraps a sync
context manager so its `__enter__`/`__exit__` run in a threadpool without
blocking the event loop. It is used for sync generator dependencies on async
endpoints (`fastapi/dependencies/utils.py`).

The function manually reimplements `with`-statement semantics (enter,
exception handoff to `__exit__`, suppression vs. propagation based on
`__exit__` return value, clean-exit teardown) across the event-loop /
threadpool boundary, but had no direct test coverage. It was only exercised
indirectly through sync generator dependency tests, so a regression in the
exception-handling logic would surface far from its source.

The suite is split in two parts:

**Integration tests** (via the public API - `Depends` on a sync generator
dependency + `async def` endpoint + `TestClient`):
- Happy path: the dependency value is yielded and teardown runs
- Setup error (raises before `yield`) produces a 500
- Cleanup error after a successful body does not break the response
- An exception raised in the endpoint body is delivered to the dependency
  teardown (proven via a `try`/`except` in the generator)

**Direct tests** for the context manager suppression protocol, which is not
observable through the public API (FastAPI error middleware converts the
endpoint exception to a 500 before the dependency teardown suppression
decision can affect the response):
- `__exit__` returning `True` suppresses the body exception
- `__exit__` returning `False` or `None` re-raises it
- An exception raised in the context manager after `yield` propagates

Direct tests run on both asyncio and trio backends via `pytest.mark.anyio`.

## AI Disclaimer

Claude Sonnet 5;

I used AI to translate this text into English, as my English level is
not high (ru -> eng).

The test code itself was written and reviewed by me. I verified everything
locally: `ruff check`, `ruff format --check`, `mypy fastapi/`, and
`pytest tests/test_concurrency.py` (12 passed on both asyncio and trio).

## Checklist

- [X] I added tests for the change.
- [X] Coverage stays at 100%.
- [X] The documentation explains the change if needed.

### DIFF
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -1,0 +1,154 @@
+from collections.abc import Generator
+from contextlib import contextmanager
+
+import pytest
+from fastapi import Depends, FastAPI
+from fastapi.concurrency import contextmanager_in_threadpool
+from fastapi.testclient import TestClient
+
+app = FastAPI()
+
+state: dict[str, str] = {}
+
+
+def reset_state() -> None:
+    state.clear()
+
+
+def gen_value() -> Generator[str, None, None]:
+    state["phase"] = "started"
+    try:
+        yield "value"
+    finally:
+        state["phase"] = "finalized"
+
+
+def gen_raises_before_yield() -> Generator[str, None, None]:
+    raise ValueError("setup failed")
+    yield  # pragma: no cover
+
+
+def gen_raises_after_yield() -> Generator[str, None, None]:
+    try:
+        yield "value"
+    finally:
+        state["phase"] = "finalized"
+        raise ValueError("cleanup failed")
+
+
+def gen_swallows_body_error() -> Generator[str, None, None]:
+    try:
+        yield "value"
+    except RuntimeError:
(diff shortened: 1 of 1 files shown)


YuriiMotov:
@nekoClangUage, thanks for your interest!

> Discussion: N/A
> N/A - not a bug report. Adding test coverage for an internal function with
> non-trivial exception-handling logic that currently has no direct tests.
> Happy to open a Discussion first if the team prefers that workflow.

Let's stick to the workflow. You can open discussion like **"Test coverage improvements"** to collect ideas where we can improve the test coverage. Once we have time for this, we will review that discussion and decide what to do.

I'll close this PR for now to keep things clean.
Thank you for understanding!

PR #15975, opened 2026-07-11
🐛 Fix race condition in _IncludedRouter cache rebuild under concurrent first requests

### LINKED ISSUE #15974: Race condition in _IncludedRouter cache rebuild can 404 valid routes and permanently corrupt the candidate cache
### Description

Since the router refactor in 0.137.0 (#15745), `_IncludedRouter.effective_candidates()` and `_IncludedRouter.effective_low_priority_routes()` in `fastapi/routing.py` rebuild their cached lists lazily on first use, with no synchronization.

The rebuild publishes an empty list on the shared attribute first, appends routes into it one by one, and only stamps the version at the end:

```python
self._effective_candidates = []
# ... appends into self._effective_candidates in a loop ...
self._effective_candidates_version = routes_version
```

If several threads drive the app's first requests at the same time (for example per-thread `TestClient`s in a parallel test suite, or any embedding that serves one app object from multiple event loops), more than one thread can pass the version check and rebuild concurrently. Every rebuilder appends into whichever list was assigned to the attribute last, so the cache ends up with duplicated candidates, and the corruption is permanent because the version gets stamped anyway. A request that runs while the list is still partially built can also get a 404 for a route that exists.

The rebuild path additionally uses `warnings.catch_warnings()` internally (via response field construction), so interleaved rebuilds can leak entries into the global warnings filters.

### Reproduction

```python
import sys
import threading

from fastapi.routing import _IncludedRouter

N_ROUTES = 120
N_THREADS = 6

app = FastAPI()
router = APIRouter()
for i in range(N_ROUTES):
    def endpoint(i: int = i):
        return {"i": i}
    router.add_api_route(f"/r{i}", endpoint, methods=["GET"])
app.include_router(router)

sys.setswitchinterval(1e-5)  # widen the rebuild window
barrier = threading.Barrier(N_THREADS)

def worker():
    client = TestClient(app)
    barrier.wait()
    client.get(f"/r{N_ROUTES - 1}")

threads = [threading.Thread(target=worker) for _ in range(N_THREADS)]
for t in threads:
    t.start()
for t in threads:
    t.join()

included = next(r for r in app.router.routes if isinstance(r, _IncludedRouter))
print(len(included._effective_candidates), "cached candidates, expected", N_ROUTES)
```

On my machine this reliably prints around 600 cached candidates for the 120 route app.

### Expected behavior

The cache rebuild should be thread safe. Building the list into a local variable and publishing it atomically (list first, version second) keeps readers from ever seeing a partial list, and serializing the rebuild avoids the duplicated appends and the warnings filter leak.

### Environment

- FastAPI version: 0.139.0
- Python version: 3.14.5
- OS: Windows 11

Related report: discussion #15914 describes the same symptoms.

UditDewan:
Fixes #15974. Also reported in discussion #15914.

## Problem

`_IncludedRouter.effective_candidates()` and `_IncludedRouter.effective_low_priority_routes()` rebuild their cached lists lazily with no synchronization. The rebuild publishes an empty list on the shared attribute before filling it and stamps the version afterwards. When several threads send the app's first requests concurrently (per-thread `TestClient`s in a parallel test suite, or any setup that serves one app object from multiple event loops), multiple threads pass the version check and append into the same list. The cache ends up permanently holding duplicated candidates (around 600 entries for a 120 route app in the repro), a concurrent reader can see a partially built list and 404 a valid route, and the interleaved `warnings.catch_warnings()` calls inside the rebuild can leak global warnings filters.

## Fix

Guard the rebuild with a per-instance lock and double-checked version test, build the new list into a local variable, and publish it atomically (list first, version second). Readers on the fast path never see a partial list, concurrent first requests trigger exactly one rebuild, and the warnings filter state is no longer touched by interleaved rebuilds of the same router.

The lock field is excluded from the dataclass `repr` and `eq` so existing semantics are unchanged.

## Tests

Added `tests/test_included_router_concurrent_rebuild.py`, which fires the first requests from several threads through a barrier and asserts that all requests succeed and that the cache holds exactly one candidate per route. It fails on master and passes with this change. All existing router and include tests pass.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -7,6 +7,7 @@ import inspect
 import json
 import os
 import stat
+import threading
 import types
 from collections.abc import (
     AsyncIterator,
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Closing, please read the contribution guidelines and adhere to the PR template, thank you 🙏 

In particular, a discussion thread should be opened and a maintainer should approve the idea before opening a PR.

PR #15993, opened 2026-07-14
fix: use None sentinel instead of mutable OAuthFlowsModel() default in OAuth2.__init__

rahulrao85:
## Summary

`OAuth2.__init__` uses `OAuthFlowsModel()` as a mutable default argument. Every call to `OAuth2()` without explicit `flows` shares the same `OAuthFlowsModel` instance. Mutations to that instance propagate across all default-using instances.

Changed to `None` sentinel with `if flows is None: flows = OAuthFlowsModel()` inside the function body.

## Verification
- Called whenever OAuth2 is instantiated without explicit flows.
- Consistent with how `SecurityScopes` (same file, line 666) handles scopes defaults.

### DIFF
--- a/fastapi/security/oauth2.py
+++ b/fastapi/security/oauth2.py
@@ -350,7 +350,7 @@ class OAuth2(SecurityBase):
                 The dictionary of OAuth2 flows.
                 """
             ),
-        ] = OAuthFlowsModel(),
+        ] = None,
         scheme_name: Annotated[
             str | None,
             Doc(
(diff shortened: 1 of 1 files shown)


fastapi/security/oauth2.py
```diff
@@ -350,7 +350,7 @@ def __init__(
                 The dictionary of OAuth2 flows.
                 """
             ),
-        ] = OAuthFlowsModel(),
+        ] = None,
```
YuriiMotov, line 353: `None` is not valid value for the parameter annotated as `OAuthFlowsModel | dict[str, dict[str, Any]],`

YuriiMotov:
@rahulrao85, thanks for pointing to this!

This is mostly code style issue, not something urgent.
We are trying to keep PR list clean, so I will close this for now. We will track this internally.

PR #15995, opened 2026-07-14
📝 Fix topic repository list not being displayed and `skip_users` not being applied

YuriiMotov:
## Description

We have 2 regressions in docs after we switched to Zensical:
* topic repository list is not displayed: https://fastapi.tiangolo.com/external-links/#github-repositories
* lists of users in FastAPI People are not being filtered with `skip_users` list, so we can see Dependabot, GitHub actions in top contributors: https://fastapi.tiangolo.com/fastapi-people/#top-contributors

The reason is that `macros` plugin doesn't support files with top level lists.
To resolve the issue we can convert data files to be mappings

## AI Disclaimer

Used AI for investigating things

## Checklist

- [ ] Coverage stays at 100%.

Checked by running `scripts/topic_repos.py` script locally that it produces correct data file

### DIFF
--- a/docs/en/data/skip_users.yml
+++ b/docs/en/data/skip_users.yml
@@ -1,5 +1,6 @@
-- tiangolo
-- codecov
-- github-actions
-- pre-commit-ci
-- dependabot
+users:
+  - tiangolo
+  - codecov
+  - github-actions
+  - pre-commit-ci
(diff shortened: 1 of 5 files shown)


tiangolo (approved):
Excellent, thanks! 🚀

PR #16008, opened 2026-07-15
Fix broken relative links in translated docs

axeroofficialtournament-del:
I noticed while looking at the translated documentation that there are three relative links that point to pages that are only in English. This results in 404s for users visiting the docs in any other locale than English.

What's broken
Affected files and links are:

docs//docs/deployment/versions.md — links to ../release-notes.md, which doesn't exist in any translated locale
docs//docs/help-fastapi.md — links to newsletter.md and fastapi-people.md#fastapi-experts, which are not translated.
This impacts all 12 locales that are translated (de, es, fr, hi, ja, ko, pt, ru, tr, uk, zh, zh-hant).

What I changed
Replaced broken relative links with absolute links to the English pages:

Old = ../release-notes.md
mew = https://fastapi.tiangolo.com/release-notes/

old = newsletter.md
new = https://fastapi.tiangolo.com/newsletter/

old = fastapi-people.md#fastapi-experts
new = https://fastapi.tiangolo.com/fastapi-people/#fastapi-experts

Why absolute URLs
These pages (release-notes, newsletter, fastapi-people) are only maintained in English. It's a good idea to use absolute URLs here, as they will not silently fail if the page is renamed, and they will properly redirect translated users to the English version of the page, instead of a broken link. This is in line with other untranslated cross-references in the docs.


YuriiMotov:
These pages are not translated intentionally, and there is a fallback that shows original English version of the page if there is no translated version.
You can open docs and see that all these links work - English versions of pages are shown

PR #16024, opened 2026-07-19
🐛 Fix `Form()` with an `Optional` Pydantic model silently discarding submitted data

evanyang0612:
### Bug

A form body parameter typed as `Optional[SomeModel]` (`SomeModel | None`) silently drops the submitted fields and resolves to `None` with **no validation error** — while the equivalent JSON body parses correctly. Simply adding `| None` to an otherwise-working `Annotated[SomeModel, Form()]` parameter breaks it.

```python
from typing import Annotated

class UserForm(BaseModel):
    username: str

app = FastAPI()

@app.post("/x")
def x(data: Annotated[UserForm | None, Form()] = None):
    return data

# POST /x  with form field  username=alice  ->  returns null  (data silently lost)
# The same endpoint with Annotated[UserForm, Form()] returns {"username": "alice"}
```

### Cause

`is_union_of_base_models()` returns `False` for `SomeModel | None` because `NoneType` is not a `BaseModel` subclass. That makes `_should_embed_body_fields()` force-embed the field, so the top-level form fields are never expanded into the model and the value falls back to the default (`None`). The `None` member case was not considered when this helper was introduced in #13827 (`tests/test_union_forms.py` only covered `UserForm | CompanyForm`).

### Fix

Ignore `None` members when checking a union of models, so `Optional[SomeModel]` (and `A | B | None`) keeps being treated as a top-level model union. An empty form still validates the required fields (a `422`), which is consistent with how an empty JSON `{}` body behaves.

### Tests

Added regression tests to `tests/test_union_forms.py` covering the required, defaulted, and multi-member-union optional cases, and updated the OpenAPI snapshot (which now correctly shows `anyOf: [SomeModel, {"type": "null"}]`).

### DIFF
--- a/tests/test_union_forms.py
+++ b/tests/test_union_forms.py
@@ -23,6 +23,23 @@ def post_union_form(data: Annotated[UserForm | CompanyForm, Form()]):
     return {"received": data}
 
 
+@app.post("/form-optional-required/")
+def post_optional_required_form(data: Annotated[UserForm | None, Form()]):
+    return {"received": data}
+
+
+@app.post("/form-optional/")
+def post_optional_form(data: Annotated[UserForm | None, Form()] = None):
+    return {"received": data}
+
+
+@app.post("/form-optional-union/")
+def post_optional_union_form(
+    data: Annotated[UserForm | CompanyForm | None, Form()] = None,
+):
+    return {"received": data}
+
+
 client = TestClient(app)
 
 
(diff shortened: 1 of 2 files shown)


tests/test_union_forms.py
```diff
@@ -59,6 +76,61 @@ def test_empty_form():
...
+    assert response.json() == {
+        "received": {"name": "Jane Doe", "email": "jane@example.com"}
+    }
+
+
+def test_post_optional_form_empty():
+    # An empty form is equivalent to an empty object (`{}`), so the required model
+    # fields are still validated (mirroring a JSON `{}` body), giving a 422.
+    response = client.post("/form-optional/")
+    assert response.status_code == 422, response.text
```
YuriiMotov, line 113: When I declare parameter as `data: Annotated[UserForm | None, Form()] = None`, I expect the request with empty body to be treated as valid.
If empty body is rejected with 422 error, why do we need this `| None` at all?

YuriiMotov (commented):
@evanyang0612, thanks for your interest in FastAPI!

This way of declaring form parameters (optional Form model) is not documented, so it's not a bug - it's just not currently supported by FastAPI.

>

Please, open a discussion and we will review it when we have time.

Thank you for understanding!

PR #16026, opened 2026-07-20
📝 Add Chinese terminology for HTTP headers

a1212a11:
Add preferred Simplified Chinese terminology for generic HTTP headers and FastAPI `Header` parameters.

This avoids inconsistent translations such as `header`, `headers`, `标头`, and `头部` when pages are regenerated.

Discussed and invited in https://github.com/fastapi/fastapi/discussions/9188#discussioncomment-17696650

### DIFF
--- a/docs/zh/llm-prompt.md
+++ b/docs/zh/llm-prompt.md
@@ -31,6 +31,8 @@ Use the following preferred translations when they apply in documentation prose:
 
 - request (HTTP): 请求
 - response (HTTP): 响应
+- HTTP header(s): 请求头
+- Header parameter(s): `Header` 参数
 - path operation: 路径操作
 - path operation function: 路径操作函数
 


docs/zh/llm-prompt.md
```diff
@@ -31,6 +31,8 @@ Use the following preferred translations when they apply in documentation prose:
 
 - request (HTTP): 请求
 - response (HTTP): 响应
+- HTTP header(s): 请求头
```
waketzheng, line 34: 不一定是请求头哦，也可能是response的header，例如：

```
$ grep -rn 'HTTP header' docs/en
docs/en/docs/advanced/custom-response.md:7:But if you return a `Response` directly (or any subclass, like `JSONResponse`), the data won't be automatically converted (even if you declare a `response_model`), and the documentation won't be automatically generated (for example, including the specific "media type", in the HTTP header `Content-Type` as part of the generated OpenAPI).
```

a1212a11:
Thanks for catching this. After reviewing the usages more carefully, I agree that `HTTP header` is context-dependent and should not be mapped generically to `请求头`. A prompt rule here could create incorrect translations rather than improve consistency, so I am closing this PR instead of replacing it with another overly broad rule.

PR #16038, opened 2026-07-23
🐛 Fix implicit path params with Starlette URL converters not recognized

h4sht:
## Description

Fixes #14883

When using Starlette URL converters like `{param:int}` without an explicit `Path()` annotation, `get_path_param_names()` was returning the full converter pattern (e.g. `param:int`) instead of just the parameter name (`param`). This caused FastAPI to not recognize the parameter as a path parameter.

## Changes

- **`fastapi/utils.py`**: Updated `get_path_param_names()` to strip converter suffixes by splitting on `:`
- **`tests/test_starlette_urlconvertors.py`**: Added `test_implicit_path_param_with_converters` that tests `:int`, `:float`, and `:path` converters with implicit parameters, including `url_path_for` verification

## Verification

- All 6 tests in `test_starlette_urlconvertors.py` pass
- 99 related path parameter tests pass with zero regressions

### DIFF
--- a/fastapi/utils.py
+++ b/fastapi/utils.py
@@ -41,7 +41,7 @@ def is_body_allowed_for_status_code(status_code: int | str | None) -> bool:
 
 
 def get_path_param_names(path: str) -> set[str]:
-    return set(re.findall("{(.*?)}", path))
+    return {param.split(":")[0] for param in re.findall(r"{(.*?)}", path)}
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Thanks!
We already have https://github.com/fastapi/fastapi/pull/14883

PR #16040, opened 2026-07-23
🐛 Fix response_model_* params ignored for non-generator Iterable endpoints

h4sht:
## Description

Fixes #15093

Non-generator endpoints with `Iterable[T]` return type had `response_model` set to `None` because `get_stream_item_type()` matched `Iterable` as a stream origin (it's in `_STREAM_ORIGINS`). This caused `serialize_response()` to fall back to `jsonable_encoder()`, ignoring `response_model_*` params like `exclude_none`, `exclude_unset`, and `exclude_defaults`.

## Changes

- **`fastapi/routing.py`**: Added `inspect.isasyncgenfunction/isgeneratorfunction` check before treating the return type as a stream in `_populate_api_route_state`. Non-generator endpoints with `Iterable[T]` now correctly use the return annotation as the response model.
- **`tests/test_response_model_as_return_annotation.py`**: Added `test_iterable_return_type_response_model_inferred` verifying that `response_model_exclude_none` works for non-generator endpoints returning `Iterable[ItemOut]`

## Verification
- All 69 related tests pass (stream, SSE, response_model) with zero regressions

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -991,9 +991,9 @@ def _populate_api_route_state(
                 # endpoint is an actual generator. Non-generator endpoints
                 # returning Iterable[T] should still have a response_model for
                 # proper serialization with response_model_* params.
-                if inspect.isasyncgenfunction(
+                if inspect.isasyncgenfunction(endpoint) or inspect.isgeneratorfunction(
                     endpoint
-                ) or inspect.isgeneratorfunction(endpoint):
+                ):
                     # Extract item type for JSONL or SSE streaming when
                     # response_class is DefaultPlaceholder (JSONL) or
                     # EventSourceResponse (SSE).
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -987,18 +987,27 @@ def _populate_api_route_state(
         else:
             stream_item = get_stream_item_type(return_annotation)
             if stream_item is not None:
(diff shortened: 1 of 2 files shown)


fastapi/routing.py
```diff
@@ -987,18 +987,27 @@ def _populate_api_route_state(
...
-                    isinstance(response_class, DefaultPlaceholder)
-                    or lenient_issubclass(response_class, EventSourceResponse)
-                ) and not lenient_issubclass(stream_item, ServerSentEvent):
-                    route.stream_item_type = stream_item
-                response_model = None
+                # Only treat Iterable/AsyncIterable/etc. as a stream when the
+                # endpoint is an actual generator. Non-generator endpoints
+                # returning Iterable[T] should still have a response_model for
+                # proper serialization with response_model_* params.
+                if inspect.isasyncgenfunction(endpoint) or inspect.isgeneratorfunction(
```
Sanjays2402, line 994: This disagrees with the generator detection used below: a callable instance with a generator `__call__` has `route.dependant.is_gen_callable == True`, but both `inspect` checks here are false, so it gets a response model and loses streaming. Can this reuse the dependant flag instead?

YuriiMotov:
We already have PR https://github.com/fastapi/fastapi/pull/15093

Please, don't open duplicate PRs. Review existing PR instead

PR #16043, opened 2026-07-24
🐛 Fix `exclude_defaults` not propagated to dict keys and values in `jsonable_encoder`

MBGrao:
## Problem

`jsonable_encoder` forwards `exclude_defaults` when recursing into list/tuple/set items, but drops it when recursing into dict keys and values — while the sibling flags `exclude_unset` and `exclude_none` *are* forwarded in that same dict branch. The same model with the same flag encodes differently depending on its container:

```python

class M(BaseModel):
    a: int = 5

jsonable_encoder([M()], exclude_defaults=True)       # [{}]
jsonable_encoder({"k": M()}, exclude_defaults=True)  # {"k": {"a": 5}}  <- flag silently ignored
```

Reproduced on 0.139.2.

## Fix

Forward `exclude_defaults` in the dict branch's key and value recursion, matching the list branch and the two sibling flags already passed there.

This deliberately does **not** change `include`/`exclude` propagation for dict values — that's the broader design question tracked in discussion #15961; this PR only removes the inconsistency between the two container branches for a flag that is already documented as recursive.

## Tests

Added `test_encode_model_with_default_in_dict_and_list` covering list, dict, and dict-of-list nesting plus a guard that `exclude_unset`/`exclude_none` behaviour is unchanged.

`pytest tests/test_jsonable_encoder.py tests/test_serialize_response.py tests/test_serialize_response_model.py tests/test_serialize_response_dataclass.py tests/test_skip_defaults.py tests/test_response_by_alias.py` → 59 passed, 1 skipped.

Related: discussions #16029 (original report) and #15961.


dannymaaz (commented):
I reproduced the container inconsistency and applied the two forwarding changes locally.

The updated behavior is consistent across:

- a model inside a dictionary,
- a model inside a list nested in a dictionary,
- an explicitly provided value equal to its default,
- a non-default value,
- a field using `default_factory`.

The implementation correctly aligns dictionary recursion with the existing sequence recursion for `exclude_defaults`.

One non-blocking test improvement would be to include a `default_factory` field, as that protects the Pydantic default-handling path in addition to literal defaults. I did not find a blocking issue with the proposed change.

tests/test_jsonable_encoder.py
```diff
@@ -202,6 +202,20 @@ def test_encode_model_with_default():
...
+    assert jsonable_encoder([model], exclude_defaults=True) == [{"foo": "foo"}]
+    assert jsonable_encoder({"key": model}, exclude_defaults=True) == {
+        "key": {"foo": "foo"}
+    }
+    assert jsonable_encoder({"key": [model]}, exclude_defaults=True) == {
+        "key": [{"foo": "foo"}]
+    }
+    assert jsonable_encoder({"key": model}, exclude_unset=True, exclude_none=True) == {
+        "key": {"foo": "foo", "bar": "bar"}
+    }
```
YuriiMotov:
```suggestion
    assert jsonable_encoder({"key": model}) == {"key": {"foo": "foo", "bar": "bar"}}
```

We are testing `exclude_defaults` here. No need to add `exclude_unset` and `exclude_none` - it will only mislead future readers

fastapi/encoders.py
```diff
@@ -299,6 +299,7 @@ def jsonable_encoder(
                     key,
                     by_alias=by_alias,
                     exclude_unset=exclude_unset,
+                    exclude_defaults=exclude_defaults,
```
YuriiMotov, line 302: Just a note, no need to change this for now.
Passing it to the recursive call for dict keys is not strictly needed - currently tests doesn't cover this and I couldn't come up with use case for this.
But we can still keep it just for the symmetry

tests/test_jsonable_encoder.py
MBGrao: Applied in ce8add1 — replaced it with a plain no-flags baseline assertion. One small difference from the suggestion text: with no flags passed, `bla` is also present in the output (nothing is excluded), so the expected value spells out all three fields. Thanks for the review!

YuriiMotov (approved):
LGTM!
@MBGrao, thanks!

tiangolo (approved):
Thank you @MBGrao! Let's get this in. Thanks @YuriiMotov for the review and management!

This will be available in FastAPI 0.140.9 in the next few hours. ☕

PR #16049, opened 2026-07-24
⚡️ Reduce memory usage in dependencies

tiangolo:
## Pull Request

⚡️ Reduce memory usage in dependencies

This is based in the analysis from @ipeluffo in https://github.com/fastapi/fastapi/discussions/14742 and https://github.com/fastapi/fastapi/pull/15336 plus the fork from @adriangb in https://github.com/fastapi/fastapi/compare/master...adriangb:fastapi:master , reported also by @Kludex 

---

These two previous approaches were respecting the interface of a `Dependant` (thanks for that!).

But I think it's fine to instead simplify all that code and stop using those attributes if the caching helpers won't help. So I moved it out of the class and into helper functions. Now the class is just holding data.

Also some of the others could be simplified and removed the cache not being that useful (this part detected by Codex).

I added memory benchmark tests here so it's easier to keep track of these things and see the difference. https://github.com/fastapi/fastapi/pull/16046

Thanks all for the help here! :raised_hands: 

Discussion: 

## AI Disclaimer

Codex with gpt-5.6-sol and a lot of steering.

Meticulously reviewing the code (and refactoring it a bit to reduce diff, etc) by hand.

<details>

</details>

## Checklist

- [x] This PR links to a GitHub Discussion for the proposed code change.
- [x] Coverage stays at 100%.


tiangolo:
This is now available in FastAPI 0.140.0. :zap:

fastapi/dependencies/models.py
```diff
@@ -50,144 +50,188 @@ class Dependant:
...
+    ) or inspect.isgeneratorfunction(_unwrapped_call(dunder_unwrapped_call))
+
+
+def _is_gen_callable(call: Callable[..., Any] | None) -> bool:
+    if call is None:
+        return False  # pragma: no cover
+    return _is_gen_callable_cached(_CallIdentity(call))
+
+
+@lru_cache(maxsize=1024)
```
dimaqq, line 170: I wonder how well that works or doesn't work in large projects with hairy APIs.

tiangolo: Good point, I'll bump it according to internal reports shared of the existing size in existing apps. If anyone else wants to share their internal dependency sizes, I'm all ears, DMs work as well if you don't want to share publicly.

I unfortunately don't have any type of telemetry, it would be so useful for this type of thing. So for now, I'll base it on individual reports.

ipeluffo: First of all, thanks for taking the time to look into this and merging the change.

I was also a bit concern about this limit since the dependencies graph, where I found the original issue, is quite big (this is another topic I may raise in a new PR or discussion).

That said, from a very quick test with the new version, we noticed a reduction of memory usage (expected) but we didn't noticed an impact on CPU usage (which would be expected if this limit was too low).

I'll do more exhaustive tests and ping you in private if necessary.

Thanks again

tiangolo: Awesome, thank you! I just released 0.140.1 bumping this to 4096. That would cover the initial report I received. But let me know if you have a use cases that could need more. :nerd_face:

PR #16055, opened 2026-07-25
Fix SSE data payload dropping trailing newlines

vidigoat:
### Summary

`format_sse_event` splits the data payload with `str.splitlines()` before writing each `data:` field. This silently drops a **trailing** newline, so a `ServerSentEvent` whose data ends in `\n` is sent one field short and a client reconstructs the payload without its final newline.

`str.splitlines()` also treats characters as line breaks that the SSE wire format does **not** recognize (`\v`, `\f`, `\x1c`–`\x1e`, `\x85`, ` `, ` `), so those characters inside `raw_data` — documented as sent "as-is" — get wrongly split into separate `data:` fields.

### Reproduction

```python
from fastapi.sse import format_sse_event

format_sse_event(data_str="line1\n")
# actual:   b'data: line1\n\n'          -> client rebuilds "line1"  (trailing "\n" lost)
# expected: b'data: line1\ndata: \n\n'  -> client rebuilds "line1\n"

format_sse_event(data_str="a\x0bb")
# actual:   b'data: a\ndata: b\n\n'     -> "\x0b" turned into a line break
# expected: b'data: a\x0bb\n\n'         -> "\x0b" sent as-is
```

Per the [WHATWG event-stream format](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), a stream is split into lines on **CR LF, CR, or LF only**, and a `data:` value is joined back with `\n` on the client. To transmit `"line1\n"` the server must emit `data: line1` followed by an empty `data:` field.

### Fix

Split the payload on the SSE line terminators (`\r\n`, `\r`, `\n`) and keep the trailing empty field, instead of using `str.splitlines()`. Multi-line data, CR, and CRLF continue to encode exactly as before.

A unit test covering the trailing-newline, empty, multi-line, CR/CRLF, and non-SSE-separator cases is added. `fastapi/sse.py` stays at 100% coverage.

### DIFF
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -1,3 +1,4 @@
+import re
 from typing import Annotated, Any
 
 from annotated_doc import Doc
@@ -156,6 +157,13 @@ class ServerSentEvent(BaseModel):
         return self
 
 
+# Line terminators recognized by the SSE wire format: CR LF, CR, or LF
+# (see the WHATWG stream parsing spec). Unlike ``str.splitlines()`` this
+# does not treat other Unicode separators (e.g. \v, \f, \x1c-\x1e, \x85,
+# \u2028, \u2029) as line breaks, and it preserves a trailing empty field.
+_SSE_LINE_TERMINATORS = re.compile(r"\r\n|\r|\n")
+
+
 def format_sse_event(
     *,
     data_str: Annotated[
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@vidigoat, thanks, but we already have https://github.com/fastapi/fastapi/pull/15515

PR #15432, opened 2026-04-26
📝 Update ReDoc GitHub URLs from `Rebilly/ReDoc` to `Redocly/redoc`

shalom-ezekiel:
## Description

The ReDoc project repository was moved from `github.com/Rebilly/ReDoc` to `github.com/Redocly/redoc`. While the old URL still redirects, all documentation references should point to the canonical URL to avoid confusion and ensure long-term link stability.

## Changes

Updated all ReDoc GitHub URLs across:

- `README.md`
- Documentation for `en`

PR #16058, opened 2026-07-25
📝 Update ReDoc URLs from Rebilly/ReDoc to Redocly/redoc

adnanahamed66772ndpc:
Closes #15432

ReDoc repository was moved from `github.com/Rebilly/ReDoc` to `github.com/Redocly/redoc`.

This PR updates all documentation references across 72 files to point to the new location.
Also updated blob URLs from `master` to `main` branch.


YuriiMotov:
Thanks, but but we have an automated workflow for translations and editing translated pages manually can break it (it relies on last commit date for each file do determin if the page should be re-translated).

Translations will be updated automatically on next sync in a few days

PR #16081, opened 2026-07-27
fix(sse): preserve empty data strings and trailing newlines in format_sse_event

amir-rezaei:
## Description
This PR resolves line-splitting issues in `format_sse_event()` in `fastapi/sse.py` by replacing `.splitlines()` with `.split('\n')`.

## Details
- Prevents empty data payload strings (`data_str=''`) or strings with trailing newlines from being stripped by `.splitlines()`.
- Complies strictly with the W3C SSE specification section 9.2.6.

### DIFF
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -206,14 +206,14 @@ def format_sse_event(
     lines: list[str] = []
 
     if comment is not None:
-        for line in comment.splitlines():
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Thanks!
We already have https://github.com/fastapi/fastapi/pull/15515

PR #16083, opened 2026-07-27
⚡️ Optimize dependency graph flattening

KRRT7:
## Summary

Replace recursive dependency-graph flattening with an iterative preorder traversal.

The previous implementation constructed a flattened subtree at each dependency level and copied that subtree into every ancestor. It also used linear list membership when skipping repeated dependencies.

This change:

- accumulates dependencies directly into one flat result;
- uses set-backed repeated-dependency detection, while retaining compatibility with unhashable callables;
- replaces recursive OAuth scope analysis with iterative postorder traversal;
- avoids recursion-limit failures for deeply nested dependency graphs;
- preserves dependency order, inherited OAuth scopes, parameter fields, and `skip_repeats` behavior.

This affects route construction and OpenAPI generation, not per-request dependency execution.

## Complexity

For a dependency chain of depth `d`, the previous recursive implementation repeatedly copied descendant lists into their ancestors, resulting in `O(d²)` work and memory references.

The iterative implementation traverses the dependency graph and accumulates the flattened result directly, making a chain `O(d)`.

For repeated dependencies, set-backed membership changes duplicate detection from `O(n²)` to average `O(n)` for hashable dependency cache keys. The list fallback preserves support for unhashable callables.

## Internal representation

The root flattened `Dependant` retains the same preorder dependency sequence and combined parameter fields.

One internal detail changes: entries in the root flat dependency list no longer retain their own recursively flattened descendant lists. All descendants are accumulated directly in the root flat list. Existing FastAPI consumers operate on the root flat collections.

## Benchmarks

A local synthetic benchmark flattening a 500-node dependency chain with `skip_repeats=True` measured:

| Implementation | Time |
|---|---:|
| `master` | ~7.8 ms |
| This PR | ~1.6 ms |

This is approximately a 4.9× improvement for that case.

The PR adds a focused CodSpeed benchmark so CI can provide the authoritative comparison across revisions.

Related discussion: #16057

### DIFF
--- a/fastapi/dependencies/models.py
+++ b/fastapi/dependencies/models.py
@@ -105,19 +105,43 @@ def _uses_scopes(
     cached = cache.get(cache_key)
     if cached is not None and cached[0] is dependant:
         return cached[1]
-    if dependant.own_oauth_scopes:
-        result = True
-    elif dependant.security_scopes_param_name is not None:
-        result = True
-    elif _is_security_scheme(dependant=dependant):
-        result = True
-    else:
-        result = any(
-            _uses_scopes(dependant=sub_dep, cache=cache)
-            for sub_dep in dependant.dependencies
-        )
-    cache[cache_key] = (dependant, result)
-    return result
+
+    # Resolve children before parents without using the Python call stack. Deep
+    # dependency chains are valid, and this analysis is also used while flattening
+    # them during route construction.
+    stack = [(dependant, False)]
+    scheduled: set[int] = set()
+    while stack:
+        current, children_resolved = stack.pop()
+        current_key = id(current)
+        if children_resolved:
+            result = bool(
+                current.own_oauth_scopes
+                or current.security_scopes_param_name is not None
+                or _is_security_scheme(dependant=current)
+            )
+            if not result:
+                result = any(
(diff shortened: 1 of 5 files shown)


YuriiMotov:
@KRRT7, thanks for your interest and efforts!

I think the benchmark added by this PR is not realistic (dependency chain of 500 nested dependencies; and you only measure the effect of one internal utility function).

There is already a benchmark for 100 nested dependencies (`tests/benchmarks/test_openapi.py`) and it [shows the 1% decrease of performance](https://app.codspeed.io/fastapi/fastapi/branches/KRRT7%3Aoptimize-dependency-flattening?utm_source=github&utm_medium=comment-v2&utm_content=button&uri=tests%2Fbenchmarks%2Ftest_openapi.py%3A%3Atest_openapi_dependency_graph&runnerMode=Simulation) with the changes from this PR.

Every refactoring brings the risk of introducing new bugs. We shouldn't refactor the code that doesn't give improvements for real use cases.
I would close this unless you can come up with more realistic benchmark that shows the improvement on this branch.

tiangolo:
This seems a complex internal refactor, which makes it have the potential to break edge cases. For now I'll pass on this, and then probably revisit the topic in the future. :coffee:

PR #16128, opened 2026-08-02
🐛 Fix header parameter case-sensitivity for Pydantic models with uppercase aliases

Sujit-1509:
### Bug Description

When declaring headers via a Pydantic model (`def endpoint(headers: HeaderModel = Header())`), fields with uppercase or mixed-case aliases (e.g., `Field(..., alias="X-API-Key")` or `Field(..., alias="User-Agent")`) cause valid requests to fail with HTTP 422 `extra_forbidden` when the model has `model_config = ConfigDict(extra="forbid")`.

### Root Cause

In `request_params_to_args` (`fastapi/dependencies/utils.py`):
1. FastAPI records processed keys in `processed_keys` using the field's declared alias (e.g., `"X-API-Key"`).
2. Starlette's `Headers.keys()` returns lowercased header names (e.g., `"x-api-key"`).
3. The check `if key not in processed_keys` is case-sensitive in Python (`"x-api-key" != "X-API-Key"`).
4. Lowercased header names were treated as unprocessed extra headers and added to `params_to_process["x-api-key"]`, causing Pydantic to raise an `extra_forbidden` error.

### Solution

Normalize keys added to `processed_keys` to lowercase when `received_params` is an instance of Starlette `Headers`. Query, cookie, and path parameters remain unaffected.

### Testing

Added regression tests in `tests/test_query_cookie_header_model_extra_params.py` covering:
- `extra="forbid"` models with uppercase/mixed-case aliases (`X-API-Key`, `User-Agent`).
- Lowercase aliases.
- `convert_underscores=True` and `convert_underscores=False`.
- Sequence/list headers (`list[str]`).
- Verification that all header test cases pass cleanly.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -821,11 +821,17 @@ def request_params_to_args(
         value = _get_multidict_value(field, received_params, alias=alias)
         if value is not None:
             params_to_process[get_validation_alias(field)] = value
-        processed_keys.add(alias or get_validation_alias(field))
-        # For headers with convert_underscores=True, mark both the converted
-        # header name and the original field alias as processed to avoid
-        # accepting the original alias as an extra header.
-        processed_keys.add(get_validation_alias(field))
+        target_key = alias or get_validation_alias(field)
+        val_key = get_validation_alias(field)
+        if isinstance(received_params, Headers):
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Thanks for your interest!

>

Thanks for understanding!

PR #16134, opened 2026-08-03
fix: prevent UnboundLocalError for status_code in OpenAPI generation

Quratulain-bilal:
## Description

In `fastapi/openapi/utils.py`, when `route.status_code` is `None` and the response class does not have a `status_code` parameter with an `int` default, the `status_code` variable is never assigned. This causes an `UnboundLocalError` on the next line when it tries to use `status_code`.

## Changes

- Initialize `status_code = None` before the else branch
- Add a `if status_code is not None` guard before using `status_code` in `operation.setdefault()`

## Impact

Prevents runtime crash when using custom response classes that don't have a `status_code` parameter.


YuriiMotov:
I don't think this is right approach.

We track this internally and will address when we have time. Thanks!

PR #16144, opened 2026-08-05
🐛 Fix field aliases ignored for `ServerSentEvent` data in SSE streams

SAURABHSALVE:
## Problem

`ServerSentEvent(data=...)` serializes Pydantic models via `model_dump_json()`,
whose `by_alias` default is `False`. Every other serialization path in FastAPI
defaults to `by_alias=True`:

- `jsonable_encoder()` — `by_alias: bool = True`
- `ModelField.serialize_json()` — `by_alias: bool = True`
- `response_model_by_alias` on all route decorators — `True`

This makes `ServerSentEvent` the only path that drops aliases, so the same model
serializes differently depending on how it is yielded:

```python
class Item(BaseModel):
    item_name: str = Field(serialization_alias="itemName")

@app.get("/a", response_class=EventSourceResponse)
async def a() -> AsyncIterable[Item]:
    yield Item(item_name="x")                        # data: {"itemName":"x"}

@app.get("/b", response_class=EventSourceResponse)
async def b():
    yield ServerSentEvent(data=Item(item_name="x"))  # data: {"item_name":"x"}

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -531,9 +531,15 @@ def get_request_handler(
                             data_str: str | None = item.raw_data
                         elif item.data is not None:
                             if hasattr(item.data, "model_dump_json"):
-                                data_str = item.data.model_dump_json()
+                                data_str = item.data.model_dump_json(
+                                    by_alias=response_model_by_alias
+                                )
                             else:
-                                data_str = json.dumps(jsonable_encoder(item.data))
+                                data_str = json.dumps(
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Thanks for your interest and efforts, but let's keep it closed until maintainers can review the discussion https://github.com/fastapi/fastapi/discussions/15703 and decide how to handle it.

>

Thanks for understanding!

PR #16145, opened 2026-08-05
🐛 Fix exclude_unset/exclude_defaults/exclude_none not propagated to ServerSentEvent data

SAURABHSALVE:
## Problem

`ServerSentEvent(data=...)` serializes Pydantic models via `model_dump_json()`
without passing `exclude_unset`, `exclude_defaults`, or `exclude_none`. Every
other serialization path in FastAPI threads these parameters through:

- `ModelField.serialize_json()` — used for plain (non-wrapped) stream items
- `serialize_response()` — used for regular (non-streaming) responses

This makes `ServerSentEvent` data the only path that ignores these route-level
settings, so the same model serializes differently depending on how it is
yielded:

```python
class Item(BaseModel):
    name: str
    tag: str = "default-tag"

@app.get("/a", response_class=EventSourceResponse, response_model_exclude_defaults=True)
async def a() -> AsyncIterable[Item]:
    yield Item(name="x")                        # data: {"name":"x"}

@app.get("/b", response_class=EventSourceResponse, response_model_exclude_defaults=True)
async def b():
    yield ServerSentEvent(data=Item(name="x"))  # data: {"name":"x","tag":"default-tag"}

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -531,9 +531,20 @@ def get_request_handler(
                             data_str: str | None = item.raw_data
                         elif item.data is not None:
                             if hasattr(item.data, "model_dump_json"):
-                                data_str = item.data.model_dump_json()
+                                data_str = item.data.model_dump_json(
+                                    exclude_unset=response_model_exclude_unset,
+                                    exclude_defaults=response_model_exclude_defaults,
+                                    exclude_none=response_model_exclude_none,
+                                )
                             else:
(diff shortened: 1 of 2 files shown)


YuriiMotov:
Thanks for your interest and efforts, but let's keep it closed until maintainers can review the discussion https://github.com/fastapi/fastapi/discussions/15703 and decide how to handle it.

>

Thanks for understanding!

PR #16152, opened 2026-08-07
👷 Optimize cache usage in workflows

YuriiMotov:
## Description

After upgrading https://github.com/astral-sh/setup-uv to 9.0.0 it started saving downloaded wheels to cache (`prune-cache` is now `false` by default).

But there are currently several things that make caching less efficient:
* Some workflows install deps with `--no-dev` but don't use `UV_NO_SYNC: false`, so `uv run` still installs dev dependencies and caches them
* ~15 jobs shared the same cache key, but installed different set of dependencies. It might introduce a race - different cache content depending on what job finished first
* Workflows triggered by `pull_request` still save cache, but this cache is scoped to merge branch and can't be read by others
* `regression-test` job always has cache miss because it copies project from the base branch into `base/` and `cache-dependency-glob` uses all projects to compute the hash by default

This PR introduces several cache key suffixes with different set of dependencies, and ensures that there is at least one job for each cache key that writes cache to default branch scope, so that other workflows would be able to read it.

Cache key suffixes:

* `dev-all`:
  * **Content:** `--group dev --extra all`
  * **Writers:** `warm-cache.yml` (default-branch scoped), `pre-commit.yml` (PR-branch scoped, only available to this PR)
  * **Readers:** `pre-commit.yml`, `bump-pre-commit-hooks.yml`, `create-draft-release.yml`, `prepare-release.yml`
  * Comment: we need `warm-cache.yml` because `pre-commit.yml` runs on `pull_request` and can only write cache scoped to PR branch (others can't read), and other workflows run not regularly and can't keep cache warm
* `docs`:
  * **Content:** `--group docs`
  * **Writers and readers:** `build-docs.yml` (writes to default branch scope on push to master branch, so that all other runs can read it)
* `github-actions`:
  * **Content:** `--group github-actions`
  * **Writers and readers:** `label-approved.yml`, `notify-translations.yml`, `smokeshow.yml`, `sponsors.yml`, `topic-repos.yml`, `translate.yml`
* `tests-highest`:
  * **Content:** `--group tests --extra all`
  * **Writers:**: matrix items of `test` job in `test.yml` except those that use `lowest-direct` (writes to default branch scope on push to master branch, so that all other runs can read it)
  * **Readers:** all jobs in `test.yml` except those that use `lowest-direct`
* `tests-lowest-direct`:
  * **Content:** `--group tests --extra all` resolved with `UV_RESOLUTION=lowest-direct`
  * **Writers and readers:** matrix items of `test` job in `test.yml` that use `lowest-direct` (writes to default branch scope on push to master branch, so that all other runs can read it)
* `translations`:
  * **Content:** `--group github-actions --group translations`
  * **Writers:** `langs` job
  * **Readers**: matrix items of `translate` job

## AI Disclaimer

Used Claude Code (Opus 5) to investigate and implement the fix. Checked manually

## Checklist

- [ ] Coverage stays at 100%.


.github/workflows/build-docs.yml
```diff
@@ -6,6 +6,9 @@ on:
   pull_request:
 permissions: {}
 
+env:
+  UV_NO_SYNC: true
```
YuriiMotov, line 10: Without this, it would skip installing dev group on `uv sync --locked --no-dev --group docs`, but still install it later on `uv run ./scripts/docs.py build-lang ${{ matrix.lang }}` and then cache it.
The same in other workflows

.github/workflows/test.yml
```diff
@@ -253,8 +255,15 @@ jobs:
...
           # Before upgrading uv version, make sure astral-sh/setup-uv knows its checksum.
           # See: https://github.com/astral-sh/setup-uv/issues/851#issuecomment-4282017837
-          version: "0.11.18"
+          version: "0.11.30"
           enable-cache: true
+          cache-suffix: tests-highest
+          # This job checks out into `base/`, so the glob must point there.
+          cache-dependency-glob: |
+            base/pyproject.toml
+            base/uv.lock
```
YuriiMotov, line 258: We need to use project from `base/`. By default it would compute the hash on all projects (root directory  and `base/` directory)

.github/workflows/pre-commit.yml
```diff
@@ -47,7 +48,8 @@ jobs:
         with:
           # Before upgrading uv version, make sure astral-sh/setup-uv knows its checksum.
           # See: https://github.com/astral-sh/setup-uv/issues/851#issuecomment-4282017837
-          version: "0.11.18"
+          version: "0.11.30"
+          cache-suffix: dev-all
           cache-dependency-glob: |
```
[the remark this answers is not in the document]
YuriiMotov: `dev` is installed by default, but I still added `--group dev` explicitly in `warm-cache.yml`.
So, this has been resolved

.github/workflows/bump-pre-commit-hooks.yml
```diff
@@ -34,6 +37,11 @@ jobs:
           cache-dependency-glob: |
             pyproject.toml
             uv.lock
+          cache-suffix: dev-all
+          # Disable saving cache as we install narrower set of dependencies than dev-all has.
+          save-cache: false
```
YuriiMotov, line 42: This workflow only needs `prek` installed.
We can add a dependency group with just `prek`, use it here and disable cache. But `dev-all` is always warm (see `warm-cache` workflow), so I decided to reuse it instead and just disabled saving.

.github/workflows/bump-pre-commit-hooks.yml
```diff
@@ -34,6 +37,11 @@ jobs:
...
+      - name: Install dependencies
+        run: uv sync --locked --group dev
```
YuriiMotov, line 44: Added explicit dependency install steps so that it would be clear what dependencies are used by the job

.github/workflows/smokeshow.yml
```diff
@@ -29,6 +32,7 @@ jobs:
         uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
         with:
           version: "latest-known"
+          cache-suffix: github-actions
```
YuriiMotov, line 35:
```suggestion
          cache-suffix: github-actions
          enable-cache: true
```

Since `setup-uv` v10.0.0, cache for this workflow is disabled (it's triggered by `workflow_run`) - see [release notes](https://github.com/astral-sh/setup-uv/releases/tag/v10.0.0).

I think we can safely enable it - we use `workflow_run` trigger with caution:
* No checkout of untrusted branches
* Artifacts used safely (just uploading)
* Narrow permissions (only `statuses: write`)
* No unpinned dependencies
* We use Zizmor to ensure workflow security
* Nothing untrusted can write the cache that this job restores

We can also consider disabling Smokeshow workflow temporary for FastAPI repo as it always fails due to upload limits.

.github/workflows/test.yml
```diff
@@ -119,6 +119,7 @@ jobs:
         with:
           version: "latest-known"
           enable-cache: true
+          cache-suffix: tests-${{ matrix.uv-resolution }}
```
YuriiMotov, line 122: We need to include `matrix.uv-resolution` into cache key in order to prevent possible future collisions if we have two equal matrix items with only difference in `uv-resolution`.

See: https://github.com/astral-sh/setup-uv#should-i-include-the-resolution-strategy-in-the-cache-key

.github/workflows/test.yml
```diff
@@ -249,6 +251,13 @@ jobs:
         with:
           version: "latest-known"
           enable-cache: true
+          cache-suffix: tests-highest
+          # This job checks out into `base/`, so the glob must point there.
+          cache-dependency-glob: |
+            base/pyproject.toml
+            base/uv.lock
+          # Only ever runs on pull requests, so anything it saved would be PR-scoped and useless.
+          save-cache: false
```
YuriiMotov, line 260: `regression-test` only runs on pull requests, so anything it saved would be PR-scoped and useless. But it will still use cache written by other jobs

.github/workflows/warm-cache.yml
```diff
@@ -0,0 +1,47 @@
+name: Warm Cache
+
+# Populates the `dev-all` uv cache at default-branch scope so that it could be used by
+# `pre-commit.yml` (it has `pull_request` trigger and can't write cache available across PRs)
+# and some others that use subset of dev dependencies.
```
YuriiMotov, line 5: Other workflows can't write this cache (for `dev-all` key) because they are either PR-scoped or run too rarely.
This workflow will run every day to ensure `dev-all` cache is warm and frequently running `pre-commit` workflow can use it

PR #16156, opened 2026-08-07
👷 Disable saving Zensical's `.cache` in `build-docs.yml`

YuriiMotov:
## Description

Disable saving Zensical's `.cache` in `build-docs.yml`.
It's currently useless because we remove directory with `.cache` before running `zensical build` in `scripts/docs.py` ([see code](https://github.com/fastapi/fastapi/blob/a375f6b948b99fa4260129856bbf11d037f363ef/scripts/docs.py#L220)).

I attempted to fix it by modifying the script to not delete `.cache`, but after some investigation I realized that it's fragile - Zensical's cache invalidation logic doesn't take into account `docs_src` or `.yaml`-files in data. So, it will generate outdated content if we change them.

## AI Disclaimer

Used Claude Code (Opus 5) to investigate and confirm the solution

## Checklist

- [ ] Coverage stays at 100%.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -103,10 +103,6 @@ jobs:
         run: uv sync --locked --no-dev --group docs
       - name: Update Languages
         run: uv run ./scripts/docs.py update-languages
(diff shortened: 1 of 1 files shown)


tiangolo (approved):
Thank you! 🙌

PR #16192, opened 2026-08-14
🐛 Fix `response_class=EventSourceResponse` sending an empty body for non-generator endpoints

SEPURI-SAI-KRISHNA:
Discussion: https://github.com/fastapi/fastapi/discussions/16107

## Description

A path operation declared with `response_class=EventSourceResponse` that is not a generator is dispatched into the SSE branch anyway, because [`get_request_handler()` recomputes `is_sse_stream` from the response class alone](https://github.com/fastapi/fastapi/blob/f336ff831c4af3d4f625c2593a27b1e0cae93eb7/fastapi/routing.py#L400), with no generator check.

```python
async def events() -> EventSourceResponse:
    return EventSourceResponse(gen())
# returns 200 text/event-stream with an empty body
```

The [branch is entered](https://github.com/fastapi/fastapi/blob/f336ff831c4af3d4f625c2593a27b1e0cae93eb7/fastapi/routing.py#L520) before the endpoint runs, so `dependant.call(...)` yields a coroutine (or a plain value) where an iterator is expected, and the `200` and `text/event-stream` headers are already flushed by the time it fails. This shows up in two ways, neither of which raises an error to the client:

- `async def` returns `200` with an empty body, plus `RuntimeWarning: coroutine ... was never awaited`. It's even worse when running under uvicorn: `iterate_in_threadpool()` raises `TypeError: 'coroutine' object is not iterable` mid-response, the connection truncates, and the client gets `RemoteProtocolError: peer closed connection without sending complete message body`.
- `sync def` returns `200` streaming the *iteration* of the return value with no error at all, so an endpoint returning `{"msg": "hello"}` puts `data: "msg"` on the wire.

The same value is already computed correctly on the route, with the generator check fused in, [right beside its JSONL sibling](https://github.com/fastapi/fastapi/blob/f336ff831c4af3d4f625c2593a27b1e0cae93eb7/fastapi/routing.py#L1075):

```python
route.is_sse_stream = is_generator and lenient_issubclass(
    response_class, EventSourceResponse
)
route.is_json_stream = is_generator and isinstance(response_class, DefaultPlaceholder)
```

but only [`is_json_stream` is passed down](https://github.com/fastapi/fastapi/blob/f336ff831c4af3d4f625c2593a27b1e0cae93eb7/fastapi/routing.py#L1248), so the two disagree for every non-generator endpoint.

This PR removes the local recomputation and threads `route.is_sse_stream` through as a parameter, exactly as `route.is_json_stream` already was, so both flags are derived in one place.

The deleted line unwrapped the `DefaultPlaceholder` before checking it, which is why it was there. However, this isn't needed at the route level. `add_api_route()` already passes `response_class` through `get_value_or_default()`. So, when `default_response_class` is set, it is already a real class when `route.is_sse_stream` is computed. When it isn't set, it remains a placeholder and correctly results in `False`.

I verified this with both app-level and router-level `default_response_class=EventSourceResponse`. In both cases, `route.is_sse_stream` remains `True`, and OpenAPI still correctly documents the response as `text/event-stream`.

Present since SSE was added in `0.135.0`; verified against stock `0.135.0` and `0.141.0`.

Adds `tests/test_sse_non_generator.py` covering:

- a non-generator returning an `EventSourceResponse` directly
- an async non-generator returning a plain value
- a sync non-generator returning a plain value (the silent-corruption case)
- the same non-generator case reached through `include_router()`
- a generator, at app level and through `include_router()`, as the control cases for normal SSE streaming

Four of the six fail on `master`, three with `TypeError: 'coroutine' object is not iterable`, and the sync one with a plain assertion showing `data: "msg"` on the wire.

As @LeonardoSanBenitez pointed out on this PR, this fixes the dispatch only. A non-generator under `response_class=EventSourceResponse` still has its return value iterated by `StreamingResponse`, so `{"a": 1}` reaches the wire as `ab` rather than as JSON. Whether that combination should instead be a startup-time error is the open question on the discussion, and is not addressed here.

## Checklist

- [x] This PR links to a GitHub Discussion for the proposed code change.
- [x] Coverage stays at 100%.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -389,6 +389,7 @@ def get_request_handler(
     strict_content_type: bool | DefaultPlaceholder = Default(True),
     stream_item_field: ModelField | None = None,
     is_json_stream: bool = False,
+    is_sse_stream: bool = False,
 ) -> Callable[[Request], Coroutine[Any, Any, Response]]:
     assert dependant.call is not None, "dependant.call must be a function"
     is_coroutine = _is_coroutine_callable(dependant.call)
@@ -397,7 +398,6 @@ def get_request_handler(
         actual_response_class: type[Response] = response_class.value
     else:
         actual_response_class = response_class
-    is_sse_stream = lenient_issubclass(actual_response_class, EventSourceResponse)
     if isinstance(strict_content_type, DefaultPlaceholder):
         actual_strict_content_type: bool = strict_content_type.value
     else:
@@ -1246,6 +1246,7 @@ class APIRoute(routing.Route):
             strict_content_type=route.strict_content_type,
             stream_item_field=route.stream_item_field,
             is_json_stream=route.is_json_stream,
+            is_sse_stream=route.is_sse_stream,
         )
 
     def matches(self, scope: Scope) -> tuple[Match, Scope]:
--- a/tests/test_sse_non_generator.py
+++ b/tests/test_sse_non_generator.py
@@ -1,0 +1,65 @@
+from collections.abc import AsyncIterable
+
+from fastapi import FastAPI
+from fastapi.responses import EventSourceResponse
+from fastapi.testclient import TestClient
+
+app = FastAPI()
+
+
+@app.get("/returned-response", response_class=EventSourceResponse)
+async def returned_response() -> EventSourceResponse:
+    async def gen() -> AsyncIterable[bytes]:
+        yield b"data: hi\n\n"
+
+    return EventSourceResponse(gen())
+
+
+@app.get("/async-endpoint", response_class=EventSourceResponse)
+async def async_endpoint():
+    return {"msg": "hello"}
+
+
+@app.get("/sync-endpoint", response_class=EventSourceResponse)
+def sync_endpoint():
+    return {"msg": "hello"}
+
+
+@app.get("/generator", response_class=EventSourceResponse)
+async def generator() -> AsyncIterable[str]:
+    yield "a"
+
+
+client = TestClient(app)
+
+
+def test_returned_event_source_response():
+    """A non-generator endpoint may return an `EventSourceResponse` directly."""
+    response = client.get("/returned-response")
+    assert response.status_code == 200, response.text
+    assert response.headers["content-type"].startswith("text/event-stream")
+    assert response.text == "data: hi\n\n"
+
+
+def test_non_generator_endpoint_does_not_crash():
+    """A non-generator endpoint must not be pushed through the SSE producer."""
+    response = client.get("/async-endpoint")
+    assert response.status_code == 200, response.text
+
+
+def test_sync_non_generator_endpoint_is_not_iterated():
+    """A sync non-generator must not have its return value iterated as SSE events.
+
+    Unlike the async case, this one fails silently: the return value is passed to
+    `iterate_in_threadpool`, so a returned dict streams its *keys* as events.
+    """
+    response = client.get("/sync-endpoint")
+    assert response.status_code == 200, response.text
+    assert "data:" not in response.text
+
+
+def test_generator_endpoint_still_streams_sse():
+    response = client.get("/generator")
+    assert response.status_code == 200, response.text
+    assert response.headers["content-type"].startswith("text/event-stream")
+    assert response.text == 'data: "a"\n\n'
--- a/tests/test_sse_non_generator.py
+++ b/tests/test_sse_non_generator.py
@@ -1,10 +1,11 @@
 from collections.abc import AsyncIterable
 
-from fastapi import FastAPI
+from fastapi import APIRouter, FastAPI
 from fastapi.responses import EventSourceResponse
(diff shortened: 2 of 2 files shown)


LeonardoSanBenitez:
Reproduced this against the base (`f336ff831`) and the head (`11d55a429`): the fix works, the reported case streams again, `tests/test_sse.py` is 39 passed at both, and the full suite goes from `3327 passed, 37 skipped, 5 xfailed` to `3331, 37, 5`, the difference being the four tests added here. It still merges cleanly on current `master`, where the bug reproduces. Looks right to me.

One thing worth knowing before it goes in, and it is a question for a maintainer rather than a change to this PR. **After the fix the return value is still iterated**, what goes away is the `data: ` framing. At the head, async endpoint, no generator:

| returns | on the wire | the same endpoint without `response_class` |
|---|---|---|
| `{"a": 1, "b": 2}` | `ab` | `{"a":1,"b":2}` |
| `["x", "y"]` | `xy` | `["x","y"]` |
| `7` | raises `TypeError: 'int' object is not iterable`, after the `200` | `7` |

The value is serialised and then handed to `actual_response_class(...)`, and `EventSourceResponse` is a `StreamingResponse` (`fastapi/sse.py:20`), so it gets iterated there instead. For async endpoints that changes the shape of the failure rather than removing it: at the base it raises, and under uvicorn truncates the connection exactly as your description says; at the head it is a `200` with a body of `msg` and no error at all. The sync case was silent before and after.

I would not hold the PR for it. A non-generator is misuse either way, since `EventSourceResponse` documents itself as being for a path operation "that uses `yield`", and the case from #16107 that motivated this is fixed. But it is the question the original reporter raised at the end of that thread, and it leaves two of the new tests weaker than they read: `test_sync_non_generator_endpoint_is_not_iterated` passes because the prefix is gone rather than because the iteration stopped, and `test_non_generator_endpoint_does_not_crash` fails at the head if the endpoint returns `7`.

One coverage note: neither `tests/test_sse.py` (routers, but generators only) nor the new file (non-generators, but app-level routes only) covers a non-generator reached through `include_router`. The fix does work there, so it is a gap rather than a defect, but an included route resolves its effective context at request time and that is a different path.

Disclaimer: This message was composed and posted by a Claude Code agent. I, Leonardo, a human, provided direct supervision, and I take responsibility for the content. Following the community policies, I believe I have put more effort into this message than a contributor would need to put to review it. If you believe this contribution was counterproductive, feel free to contact me here, on LinkedIn, or via email.

SEPURI-SAI-KRISHNA:
Thanks for reproducing it against both commits, and the table is a fair hit. You're right that the fix removes the `data: ` framing rather than the iteration, and that two of the test names claimed more than they checked.

I've pushed a test-only commit for the part that is mine to fix:

- `test_non_generator_endpoint_does_not_crash` and `test_sync_non_generator_endpoint_is_not_iterated` are renamed to
  `test_async_non_generator_endpoint_is_not_sse_framed` and `test_sync_non_generator_endpoint_is_not_sse_framed`. Both assert the absence of SSE framing, which is what they actually verify and all this fix promises.
- Added the `include_router` coverage gap you spotted, for both a non-generator and a generator. The non-generator case fails on `master` with the same `TypeError: 'coroutine' object is not iterable`, so it is a regression test rather than filler.

`fastapi/routing.py` is unchanged at +2/-1.

I've deliberately not added the `7` case. It still raises after the `200`, and making it not raise means deciding what a non-generator with `response_class=EventSourceResponse` should do at all, return JSON, or fail loudly at startup. That's the maintainer question you identified, and it's a wider change than this PR. Happy to open a separate discussion for it if that's useful.

LeonardoSanBenitez:
Re-checked at your new head (`16def0479`): the six tests in `tests/test_sse_non_generator.py` all pass, `tests/test_sse.py` is 39 passed, and `fastapi/routing.py` is unchanged at +2/-1. Four of the six still fail on current `master` (`50113da16`), the new `test_included_router_non_generator_is_not_sse_framed` among them, so that one is doing real work rather than filling in.

On the separate discussion, I would not open one. You already put that question on [#16107](https://github.com/fastapi/fastapi/discussions/16107) on 14 August, and in the form a maintainer can act on: whether a non-generator and `response_class=EventSourceResponse` are simply not meant to be combined, in which case a startup-time error is arguably the fix rather than a silent `200`. That discussion is still open, so a third thread seems more likely to split the question than to get it answered.

Disclaimer: This message was composed and posted by a Claude Code agent. I, Leonardo, a human, provided direct supervision, and I take responsibility for the content. Following the community policies, I believe I have put more effort into this message than a contributor would need to put to review it. If you believe this contribution was counterproductive, feel free to contact me here, on LinkedIn, or via email.

YuriiMotov:
Thanks for your interest and efforts, but let's keep it closed until maintainers are ready to work on this.

>

Thanks for understanding!
I've linked this PR to our internal task where we track this potential issue.

PR #16217, opened 2026-08-17
🔧 Fix timeout parsing error in `pyproject.toml`

YuriiMotov:
## Description

`pytest-timeout` released 2.5.0 and since this version it requires timeout to be specified as float, but older versions require it to be specified as string.

In tests we install locked version (2.4.0) and it works, but `test-redistribute` workflow installs latest version (2.5.0) and fails.

```
INTERNALERROR> TypeError: /home/runner/work/fastapi/fastapi/dist/fastapi-0.141.1/pyproject.toml: config option 'timeout' expects a float, got str: '20'
```

See related PR: https://github.com/pytest-dev/pytest-timeout/pull/200

The only solution I see is to ignore the Dependabot's `cooldown` config and bump `pytest-timeout` to 2.5.0

## AI Disclaimer

Didn't use AI

## Checklist

- [ ] Coverage stays at 100%.


YuriiMotov:
Not needed as `pytest-timeout` 2.5.0 has just been yanked: https://github.com/pytest-dev/pytest-timeout/pull/200#issuecomment-5316120503

PR #16223, opened 2026-08-18
🐛 Fix TypeAliasType list query params only keeping the last value

houssamb4:
# Summary

This PR fixes `TypeAliasType` (PEP 695-style type aliases) for sequence parameters in Query, Form, and Header.

## Root Cause

The helper `field_annotation_is_sequence()` in `fastapi/_compat/shared.py` was used by `_get_multidict_value()` to decide whether to call `.getlist()` (for sequences) or `.get()` (for scalars). It already unwrapped `Annotated` and `Union`, but not `TypeAliasType`. So an alias like:

```python
Tags = TypeAliasType("Tags", list[str])
```

was treated as a scalar — only the last value was kept.

## Fix

Added recursive unwrapping of `TypeAliasType` at the start of `field_annotation_is_sequence()`:

```python
from typing_inspection.typing_objects import is_typealiastype

def field_annotation_is_sequence(annotation: type[Any] | None) -> bool:
    if is_typealiastype(annotation):
        return field_annotation_is_sequence(annotation.__value__)
    # ... existing logic unchanged
```

This makes the fix apply transitively — nested/chained aliases like `OuterTags = TypeAliasType("OuterTags", InnerTags)` also work.

## Scope (supersedes #16111)

This PR is the single, complete fix. It covers:
- **Production fix**: Recursive `TypeAliasType` unwrapping in `field_annotation_is_sequence()`
- **Query tests**: `list[str]`, `tuple[str, ...]`, `Sequence[str]`, nested alias (all direct + model)
- **Form tests**: `list[str]`, `tuple[str, ...]`, `Sequence[str]` (direct + model via `Form()`)
- **Header tests**: `list[str]`, `tuple[str, ...]`, `Sequence[str]` (direct + model via `Header()`)

Cookie parameters are not included because Starlette's cookie parser only retains the last value for duplicate cookie names (`Cookie: p=hello; p=world` parses as `{"p": "world"}`), making list-type cookie params fundamentally untestable in the current stack.

## Credits

- Original Form fix and tests by @KSmanis in #16111
- This PR consolidates everything into one complete, standalone change

## Testing

All new regression tests pass locally (112 new + 760 full suite + 8 compat).

### DIFF
--- a/fastapi/_compat/shared.py
+++ b/fastapi/_compat/shared.py
@@ -18,6 +18,7 @@ from fastapi.types import UnionType
 from pydantic import BaseModel
 from pydantic.version import VERSION as PYDANTIC_VERSION
 from starlette.datastructures import UploadFile
+from typing_inspection.typing_objects import is_typealiastype
 
 _T = TypeVar("_T")
 
@@ -62,6 +63,9 @@ def _annotation_is_sequence(annotation: type[Any] | None) -> bool:
 
 
 def field_annotation_is_sequence(annotation: type[Any] | None) -> bool:
+    if is_typealiastype(annotation):
+        return field_annotation_is_sequence(annotation.__value__)
+
     origin = get_origin(annotation)
 
     if origin is Annotated:
--- a/tests/test_request_params/test_form/test_list.py
+++ b/tests/test_request_params/test_form/test_list.py
@@ -1,4 +1,6 @@
+from collections.abc import Sequence
 from typing import Annotated
+from typing_extensions import TypeAliasType
 
 import pytest
 from dirty_equals import IsOneOf, IsPartialDict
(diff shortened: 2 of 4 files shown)


icecold009 (commented):
Nice focused regression test for repeated query parameters through a PEP 695-style alias. One coordination point: PR #16111 independently adds the same `is_typealiastype` import and recursive unwrapping in `fastapi/_compat/shared.py` for form models. Since both PRs modify the same helper, could you coordinate or rebase so only one PR owns the production change and this PR keeps the query-specific regression test? That would avoid a merge conflict or duplicate implementation.

houssamb4:
Thanks for the feedback i’ll coordinate with #16111 and push an update to address this.

houssamb4:
Replying to @icecold009's review:

Done. The production fix now lives in this PR (#16223) with:

- **Query**: `list[str]`, `tuple[str, ...]`, `Sequence[str]`, nested/chained alias tests (all direct + model)
- **Form**: `list[str]`, `tuple[str, ...]`, `Sequence[str]` tests (direct + model via `Form()`)
- **Header**: `list[str]`, `tuple[str, ...]`, `Sequence[str]` tests (direct + model via `Header()`)

The recursive unwrapping handles nested aliases like `OuterTags = TypeAliasType("OuterTags", InnerTags)` transitively. PR #16111 is superseded and will be closed.

Cookie parameters are not included because Starlette's cookie parser only retains the last value for duplicate cookie names (`Cookie: p=hello; p=world` → `{"p": "world"}`), making list-type cookie params fundamentally untestable in the current stack.

@icecold009 — marking this review thread as resolved.

YuriiMotov:
Closing this until maintainers have time to work on this (https://github.com/fastapi/fastapi/discussions/16100#discussioncomment-18437440)

PR #16225, opened 2026-08-19
tests: add test_fastapi_cli_installed to verify CLI invocation when available

aoright:
### Summary
- In `tests/test_fastapi_cli.py`, add `test_fastapi_cli_installed` to test that `fastapi.cli.main()` invokes `cli_main()` when available.


YuriiMotov:
I think this is not needed. We already have `test_fastapi_cli` that tests the case with installed CLI

PR #16076, opened 2026-07-27
⚡️ Avoid flattening dependencies for OpenAPI

tiangolo:
## What changed

- Traverse each route dependency graph directly while collecting OpenAPI parameters and security dependencies.
- Preserve dependency order, duplicate skipping, and inherited OAuth scopes without constructing synthetic `Dependant` trees.
- Remove the now-unused `get_flat_dependant()` and `_get_security_dependencies()` helpers.
- Make the changed OpenAPI security helper private.

## Why

OpenAPI generation only needs selected parameter and security data, but it previously created a complete flattened `Dependant` tree for every route. Those temporary trees added allocations and traversal overhead without being used for dependency solving.

This builds on the OpenAPI performance and memory baselines added in #16075.

## Validation

- `uv run pytest tests -q` — 3,288 passed, 17 skipped, 5 xfailed
- `uv run pytest tests/test_dependency_models.py tests/test_security_oauth2_authorization_code_bearer_scopes_openapi.py tests/test_security_oauth2_authorization_code_bearer_scopes_openapi_simple.py tests/test_top_level_security_scheme_in_openapi.py tests/test_repeated_dependency_schema.py tests/test_security_scopes_dont_propagate.py -q`
- `uv run pytest tests/benchmarks/test_openapi.py tests/memory_benchmarks/test_openapi.py --codspeed -q`
- `uv run prek run --files fastapi/dependencies/models.py fastapi/dependencies/utils.py fastapi/openapi/utils.py tests/test_dependency_models.py`

Local uncached OpenAPI generation improved from approximately 156 ms to 149–154 ms. CodSpeed CI will provide the instrumented performance and memory comparison.

## AI Disclaimer

This PR was created with the help of OpenAI Codex, powered by GPT-5.6-sol. Reviewed by hand.

PR #16230, opened 2026-08-21
Add tests for WebSocket and Security dependency behavior

samrudh-nux:
While looking into the recent dependency-flattening refactor (#16049–#16076), I went and stress-tested two spots that seemed like the most likely places for something to have been missed — WebSocket routes reusing the same flattening path as HTTP routes, and the same dependency being used twice with different Security() scopes.
Both turned out to be handled correctly already, so this isn't a bug report — just two edge cases that weren't covered by a test yet, and given how much internal surgery just happened in this area, it felt worth locking them in:
- A shared, cached body-declaring dependency on a WebSocket route still resolves to a single embed decision, same as it would on an HTTP route.
- The same dependency used twice via Security() with different scopes ends up as two entries internally (since the cache key is scope-aware), but the final OpenAPI schema still correctly collapses it down to one parameter, since there's a second name-based dedup pass doing that, also Ran both against current master, they pass:

test_shared_body_dependency_across_websocket_route_embeds_correctly PASSED
test_same_dependency_different_scopes_yields_single_openapi_param PASSED
No production code touched, just adding coverage for behavior that's already correct.

### DIFF
--- a/tests/tests/test_flattening_refactor_seams.py
+++ b/tests/tests/test_flattening_refactor_seams.py
@@ -38,7 +38,6 @@ def test_same_dependency_different_scopes_yields_single_openapi_param():
     (scope-aware cache key), but must still resolve to a single
     parameter in the final OpenAPI schema -- the underlying HTTP header
     is the same regardless of which scopes are being checked."""
-    from fastapi.dependencies.utils import get_flat_params
 
     def get_user(security_scopes: SecurityScopes, x_token: str = Header()):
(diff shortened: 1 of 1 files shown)


YuriiMotov:
I agree that we should probably cover these 2 test cases, but it should be implemented differently.

I've added this to our internal task to improve the test coverage.

@samrudh-nux, thanks for reporting this!

samrudh-nux:
Sounds good, thanks for looking into it.

PR #16235, opened 2026-08-22
refactor(encoders): include PurePath in ENCODERS_BY_TYPE mapping

Xayar145:
## Summary

Includes `PurePath` in `ENCODERS_BY_TYPE` in `fastapi/encoders.py`:
- `PurePath` is already imported from `pathlib` alongside `Path`, but was missing from the `ENCODERS_BY_TYPE` dictionary.
- Adding `PurePath: str` allows direct type-mapping lookups for `PurePath` instances without requiring subclass iteration.
- Added unit test `test_encoders_by_type_contains_pure_path` in `tests/test_jsonable_encoder.py`.

## Changes

- `fastapi/encoders.py`: Added `PurePath: str` to `ENCODERS_BY_TYPE`.
- `tests/test_jsonable_encoder.py`: Added verification test for `ENCODERS_BY_TYPE[PurePath]`.

## Tests

- Added `test_encoders_by_type_contains_pure_path`.

### DIFF
--- a/fastapi/encoders.py
+++ b/fastapi/encoders.py
@@ -102,6 +102,7 @@ ENCODERS_BY_TYPE: dict[type[Any], Callable[[Any], Any]] = {
     IPv6Network: str,
     NameEmail: str,
     Path: str,
+    PurePath: str,
     Pattern: lambda o: o.pattern,
     SecretBytes: str,
     SecretStr: str,
--- a/tests/test_jsonable_encoder.py
+++ b/tests/test_jsonable_encoder.py
@@ -343,3 +343,11 @@ def test_encode_color(module_path):
 
     data = {"color": Color("blue")}
     assert jsonable_encoder(data) == {"color": "blue"}
+
+
+def test_encoders_by_type_contains_pure_path():
+    from fastapi.encoders import ENCODERS_BY_TYPE
+
+    assert PurePath in ENCODERS_BY_TYPE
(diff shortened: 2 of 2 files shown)


Mukller:
Same fact-check as on #16237 (same mapping proposed here): I ran this exact test body against an **unpatched** fastapi release (0.141.1) and it passes:

```python
from pathlib import PurePosixPath, PureWindowsPath

data = {
    "posix": PurePosixPath("a/b/c"),
    "windows": PureWindowsPath("a\\b\\c"),
}
assert jsonable_encoder(data) == {"posix": "a/b/c", "windows": "a\\b\\c"}   # passes
```

`PurePath` also isn't currently present in `ENCODERS_BY_TYPE`, yet encoding succeeds through the existing fallback chain — so the mapping may be redundant on current versions.

Could you share the environment where `jsonable_encoder` actually fails for pure paths (fastapi/pydantic versions + the failing input)? If there's a real failing case, the PR should include a test that fails on master; right now the added test passes with or without the change, so it doesn't demonstrate the fix.

Note: #16235 proposes this same mapping — whichever way maintainers go, it would help to first pin down whether current releases have the bug at all.

YuriiMotov:
Thanks, but I think we don't need this.

Also, we are trying to keep things clean and ask to only open PRs when it's explicitly requested by maintainers.

>

Issue #3339, opened 2021-06-07
Secure/HTTPOnly cookies untestable


kevr:
### First check

* [x] I added a very descriptive title to this issue. (I think?)
* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.
* [x] I already read and followed all the tutorial in the docs and didn't find an answer.
* [x] I already checked if it is not related to FastAPI but to [Pydantic](https://github.com/samuelcolvin/pydantic).
* [x] I already checked if it is not related to FastAPI but to [Swagger UI](https://github.com/swagger-api/swagger-ui).
* [x] I already checked if it is not related to FastAPI but to [ReDoc](https://github.com/Redocly/redoc).
* [x] After submitting this, I commit to one of:
    * Implement a Pull Request for a confirmed bug.

### What?

When testing a FastAPI application with normal cookies, everything works as intended. When setting `secure=True`, it breaks tests. The `TestClient` request context seems to lose information about cookies given to it via response, which breaks tests that depend on the cookie being resubmitted to the application.

### Example

I've produced a small repository with this reproducible issue: https://github.com/kevr/fastapi-secure-cookie-bug

Route that causes the issue: https://github.com/kevr/fastapi-secure-cookie-bug/blob/secure_cookie_bug/secure_app/app.py#L13

Test that tests and fails: https://github.com/kevr/fastapi-secure-cookie-bug/blob/secure_cookie_bug/test/test_app.py#L8

```
$ pytest test
# test_secure_cookie fails.

# Load up uvicorn on another shell.
$ uvicorn --reload secure_app.app:app

# Use our app server over http localhost.
$ curl -v 'http://localhost:8000' # Gives `set-cookie`.
$ curl -b test_cookie=TEST -v 'http://localhost:8000' # Provides a `test_cookie` to the app.
```

### Additional Comments

Am I approaching this the wrong way? I'm not sure if this is actually a bug or a "bug due to lack of understanding."

frankie567:
I wouldn't say this is a bug, IMO this is completely expected: the `Secure` flag on a cookie means it should only be sent through HTTPS (https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies). That's a security mechanism to prevent browsers sending a sensitive cookie (like an authentication token) through unsafe connection.

[...]

PR #16236, opened 2026-08-22
Inherit Starlette router dispatch

Kludex:
## Summary

Remove FastAPI's copy of `Router.app()` and express low-priority frontend routing through Starlette's existing `default` callback.

Starlette remains responsible for lifespan handling, normal route matching, partial matches, and slash redirects. FastAPI handles low-priority routes only after Starlette finds no regular route or redirect, then delegates to the original default application.

This works with FastAPI's minimum supported Starlette version, 0.46.0. Once [Starlette PR #3339](https://github.com/Kludex/starlette/pull/3339) is available, FastAPI inherits its trie-backed candidate selection without another routing change.

## Benchmark

The benchmark compares FastAPI `master` with this PR while both use the same Starlette trie branch at `3bfc970`. It dispatches requests directly through the public FastAPI ASGI application. Results are median microseconds per request across four processes, with 2,000 requests per round and seven rounds per process. Lower is better.

| Routes | First route | Last route | Missing route | Method not allowed |
| ---: | ---: | ---: | ---: | ---: |
| 10 | 12.4 -> 12.9 us | 17.5 -> 12.8 us (1.4x) | 17.8 -> 9.1 us (2.0x) | 13.8 -> 9.6 us (1.4x) |
| 100 | 12.6 -> 13.1 us | 63.9 -> 13.0 us (4.9x) | 106.7 -> 11.1 us (9.6x) | 60.8 -> 9.6 us (6.4x) |
| 500 | 12.3 -> 12.9 us | 266.4 -> 12.8 us (20.7x) | 503.1 -> 19.1 us (26.3x) | 261.6 -> 9.5 us (27.6x) |

The first route remains approximately 13 us. The improvement grows with route count for late matches, misses, and `405 Method Not Allowed` responses. These gains apply when FastAPI is used with the trie-backed Starlette change.

## Tests

- `scripts/test.sh -q` - 3348 passed, 17 skipped, 5 xfailed
- `scripts/lint.sh`
- `tests/test_frontend.py` against Starlette 0.46.0 - 93 passed

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex. I've reviewed and verified the changes.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -2535,6 +2535,8 @@ class APIRouter(routing.Router):
             default=default,
             lifespan=lifespan_context,
         )
+        self._default = self.default
+        self.default = self._handle_no_match
         if prefix:
             assert prefix.startswith("/"), "A path prefix must start with '/'"
             assert not prefix.endswith("/"), (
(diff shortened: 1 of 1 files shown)


Mukller (approved):
Verified locally on the PR branch (Python 3.13, editable install of the branch):

**Behavioral parity probe** (branch vs release 0.141.1) — four contracts, identical results:

| case | branch | release |
|---|---|---|
| `GET /path/` with only `/path` registered | 307 → `http://testserver/path` | same |
| `GET /unknown` | 404 `{"detail": "Not Found"}` | same |
| `POST` on GET-only path | 405 + `allow: GET` | same |
| `/openapi.json` | 200 | same |

**Targeted test files**: `test_router_redirect_slashes.py`, `test_router_prefix_with_template.py`, `test_empty_router.py`, `test_router_events.py`, `test_include_router_defaults_overrides.py`, `test_ws_router.py` — **70/70 green**.

Checked the wiring details:

1. Starlette's `Router.app` sets `scope["router"] = self` just like the removed fastapi copy did — no loss for middleware/introspection that reads it.
2. Match ordering is preserved: starlette's loop handles FULL/PARTIAL + its own redirect_slashes pass, then calls `self.default` — which now points at `_handle_no_match`, so fastapi's low-priority route logic runs at the same point in the sequence as before, and `_default` keeps the original 404 handler.
3. Deleting ~40 lines of drifted copy in favor of upstream's dispatch means future starlette routing fixes flow through automatically instead of needing manual sync — clearly the right direction.

One non-blocking nit: `self._default = self.default; self.default = self._handle_no_match` captures the original default at `__init__` time, so a subclass that overrides `.default` *after* `super().__init__()` would have its override bypassed (`_handle_no_match` calls the captured `_default`). Extremely unlikely to matter in practice, but worth a comment line if you want to guard it.

CI fully green. Approving.

PR #16249, opened 2026-08-25
⬆️ Bump `setup-uv` action to `10.0.1`

YuriiMotov:
## Description

Version 10.0.0 introduced breaking changes that affect (improve) security - it disables cache by default for workflows triggered by `workflow_run` (and also `pull_request_target` and `release`) triggers: https://github.com/astral-sh/setup-uv/releases/tag/v10.0.0

This will disable cache for Smokeshow workflow.
I think we can safely enable it, but let's keep it as is for now, I will address this in https://github.com/fastapi/fastapi/pull/16152

## AI Disclaimer

Used Claude Code to apply changes. Reviewed manually

## Checklist

- [ ] Coverage stays at 100%.

### DIFF
--- a/.github/workflows/smokeshow.yml
+++ b/.github/workflows/smokeshow.yml
@@ -29,7 +29,6 @@ jobs:
         uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
         with:
           version: "latest-known"
-          enable-cache: true
           cache-dependency-glob: |
             pyproject.toml
             uv.lock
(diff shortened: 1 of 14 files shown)


.github/workflows/smokeshow.yml
```diff
@@ -26,11 +26,10 @@ jobs:
...
           python-version-file: ".python-version"
       - name: Setup uv
-        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
         with:
-          # Before upgrading uv version, make sure astral-sh/setup-uv knows its checksum.
-          # See: https://github.com/astral-sh/setup-uv/issues/851#issuecomment-4282017837
-          version: "0.11.18"
+          version: "latest-known"
+          enable-cache: true
```
YuriiMotov: This workflow is triggered by `workflow_run` trigger, so `setup-uv` will disable cache by default.

I suggest enabling cache for Smokeshow workflow (see PR description).
We can also consider disabling Smokeshow workflow temporary for FastAPI repo as it always fails due to upload limits.

PR #16251, opened 2026-08-25
Fix response_model_* serialization flags ignored when response_field or stream_item_field is None

Noxtimo:
When an endpoint returns raw payloads or unmodeled items (where `response_field` or `stream_item_field` is `None`), serialization fell back to `jsonable_encoder()` without passing through the route's configured `response_model_*` flags (such as `exclude_none`, `exclude`, `include`, `by_alias`). This PR forwards these serialization options to `jsonable_encoder` and SSE event handlers, and adds test coverage for unmodeled responses and stream items.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -338,7 +338,15 @@ async def serialize_response(
         )
 
     else:
-        return jsonable_encoder(response_content)
+        return jsonable_encoder(
+            response_content,
+            include=include,
+            exclude=exclude,
+            by_alias=by_alias,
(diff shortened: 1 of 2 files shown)


03musab:
Excellent work! Forwarding all 6 serialization flags (include, exclude, y_alias, exclude_unset, exclude_defaults, exclude_none) to jsonable_encoder() and model_dump_json() fixes inconsistent behavior on unmodeled routes and streams. Test coverage is very thorough. LGTM!

PR #16265, opened 2026-08-28
🐛 Fix `format_sse_event` dropping the event terminator when no field is set

SAURABHSALVE:
## Bug

`format_sse_event()` documents that:

> The result always ends with `\n\n` (the event terminator).

It doesn't, when no field is set. The terminator is built by appending two empty entries to the list of field lines and then joining them with `\n`:

```python
lines.append("")
lines.append("")
return "\n".join(lines).encode("utf-8")
```

When `lines` is empty the join produces a single `"\n"` instead of `"\n\n"`:

```pycon
>>> from fastapi.sse import format_sse_event
>>> format_sse_event()
b'\n'                # expected b'\n\n'
>>> format_sse_event(data_str='"x"')
b'data: "x"\n\n'     # correct
```

This is reachable from user code by yielding an empty `ServerSentEvent()` from an SSE *path operation*:

```python
@app.get("/stream", response_class=EventSourceResponse)
async def stream():
    yield ServerSentEvent(data="first")
    yield ServerSentEvent()
    yield ServerSentEvent(data="second")
```

The empty event is emitted as a bare `\n` rather than a properly terminated (empty) event:

```pycon
>>> # before
'data: "first"\n\n' '\n'   'data: "second"\n\n'
>>> # after
'data: "first"\n\n' '\n\n' 'data: "second"\n\n'
```

## Fix

Append the terminator to the joined field lines instead of adding it as two list entries, so it is emitted unconditionally.

Output is byte-for-byte unchanged for every event that sets at least one field (`data`, `event`, `id`, `retry` or `comment`) — the only behavior change is the previously unterminated empty event.

## Tests

Two tests added to `tests/test_sse.py`, both of which fail on `master`:

* `test_format_sse_event_without_fields_is_terminated` — unit level.
* `test_empty_server_sent_event_is_terminated` — end to end through an SSE *path operation*.

### DIFF
--- a/fastapi/sse.py
+++ b/fastapi/sse.py
@@ -228,9 +228,11 @@ def format_sse_event(
     if retry is not None:
         lines.append(f"retry: {retry}")
 
-    lines.append("")
-    lines.append("")
-    return "\n".join(lines).encode("utf-8")
+    # Terminate the event with a blank line. The terminator is appended to the
+    # joined field lines instead of being added as two empty lines, so that it
+    # is still emitted when no fields were set at all, which is the case for an
+    # empty `ServerSentEvent()`.
+    return ("\n".join(lines) + "\n\n").encode("utf-8")
 
 
 # Keep-alive comment, per the SSE spec recommendation
--- a/tests/test_sse.py
+++ b/tests/test_sse.py
@@ -396,6 +396,33 @@ def test_format_sse_event_splitlines_behavior_in_comment():
     assert format_sse_event(comment="hi\n") == b": hi\n: \n\n"
 
 
+def test_format_sse_event_without_fields_is_terminated():
+    # Without any field the result is still terminated by a blank line,
+    # instead of a single "\n".
+    assert format_sse_event() == b"\n\n"
+
+
+empty_event_app = FastAPI()
+
+
+@empty_event_app.get("/empty-event", response_class=EventSourceResponse)
+async def empty_event_stream():
+    yield ServerSentEvent(data="first")
+    yield ServerSentEvent()
+    yield ServerSentEvent(data="second")
+
+
+def test_empty_server_sent_event_is_terminated():
+    client = TestClient(empty_event_app)
+    response = client.get("/empty-event")
+    assert response.status_code == 200
+    assert response.text == (
+        'data: "first"\n\n'  # first event
+        "\n\n"  # empty event, terminated on its own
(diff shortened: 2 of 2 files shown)


SAURABHSALVE:
Hi! Could you please add the appropriate label to this PR so that the full set of required checks can run? Thank you! @YuriiMotov

03musab:
Great fix! Clear root cause analysis and clean fix for ormat_sse_event(). The string concatenation \n.join(lines) + \n\n cleanly guarantees event termination even when lines is empty. Tests cover both unit and integration levels. LGTM!

Shriprasad-P (commented):
I have taken a look at the empty event scenario on master and have found that format_sse_event() is currently returning only b”\n” instead of the expected b”\n\n”. This change should fix that behaviour while still keeping the byte output for events that have data, event, id, retry or comment fields. I have also checked the whole SSE stream from start to finish with an empty ServerSentEvent() between two regular events, and everything is still framed correctly. The new regression tests that were added fail on master but pass with this change, and I have not found any problems with the formatter change.

Shriprasad-P (approved):
## Review

Correct SSE terminator handling. Joining empty field lines with `"\n".join([])` produced `""`, so the previous "append two empty lines" approach collapsed and dropped the blank-line event terminator for empty `ServerSentEvent()`.

Appending `"\n\n"` after the join preserves the terminator even when there are no fields, and the new unit + streaming client tests nail the regression.

Approve.

Shriprasad-P (commented):
## Review

Correct SSE terminator handling. Joining empty field lines with `"\n".join([])` produced `""`, so the previous "append two empty lines" approach collapsed and dropped the blank-line event terminator for empty `ServerSentEvent()`.

Appending `"\n\n"` after the join preserves the terminator even when there are no fields, and the new unit + streaming client tests nail the regression.

Approve.

YuriiMotov:
According to the [spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream), every event may have zero or more fields and comments, and it should end with **end-of-line**. At the same time, every field and comment should also end with **end-of-line**. That's where the **"The result always ends with `\n\n`"** stems from.
But for the event that contains zero comments/fields (absolutely valid), valid content will be just one **end-of-line**. And two **end-of-line**s would mean two empty events.

So, the fix here should be in a comment, not in code. See PR https://github.com/fastapi/fastapi/pull/16354


Anyway, thanks for pointing to the issue!

PR #16266, opened 2026-08-28
📝 Fix broken OpenAPI spec anchor link in `path-operation-advanced-configuration.md`

SAURABHSALVE:
The link to OpenAPI Extensions used `#specificationExtensions` (camelCase), which does not match any anchor on the linked GitHub page. The actual heading id, per the spec's own internal cross-references, is `#specification-extensions`. Fixed to the correct anchor so the link jumps to the right section.


YuriiMotov:
Thanks for reporting this!

Thanks for understanding!

PR #16268, opened 2026-08-29
fix(docs): strip HTML tags from termynal copy content

dlowzzxx:
Fixes #15941 — termynal copy button copied tags vialine.valueHTML. Now extractstextContentviatmp.innerHTMLfor clipboard while preserving visual HTML. Onlydocs/en/docs/js/custom.js changed.


YuriiMotov:
Thanks for your interest and efforts, but we already have PR https://github.com/fastapi/fastapi/pull/15941.
If you think your approach is better, please comment on that PR instead of opening new PR

PR #16269, opened 2026-08-29
Re-raise ClientDisconnect during request body reading instead of converting to HTTP 400

Noxtimo:
Fixes #15312

When a client disconnects while FastAPI is reading the request body, Starlette raises `starlette.requests.ClientDisconnect`. Previously, `routing.py` caught this under a general `except Exception` and converted it into an `HTTPException(400)`, which resulted in attempts to send responses over dead sockets and false 400 error alerts in monitoring. This PR explicitly re-raises `ClientDisconnect` so it can propagate naturally to ASGI middleware.


YuriiMotov:
Thanks for your interest!

Please see PR https://github.com/fastapi/fastapi/pull/15315
If you still think we should fix it, please open a discussion and let's discuss it first

PR #16271, opened 2026-08-29
Fix route field generation with frozen time

manideep997:
## Pull Request

Discussion: No GitHub Discussion was provided for this contribution. I independently implemented and tested this contribution.

### Problem

When an `APIRouter` is included in a `FastAPI` application, the route's effective candidates and dependant fields are not built until they are needed.

If the first request to the included router is made inside a `freezegun.freeze_time()` context, the route field generation can happen while time is frozen. This can cause the request to fail with a `500 Internal Server Error`.

The issue was reproduced with a route containing an optional `date` query parameter.

Before the fix, the reproduction returned:

`STATUS: 500`

### Solution

The included router is now stored in an `included_router` variable when `include_router()` is called.

After the router is added to `self.routes` and `_mark_routes_changed()` is called, `included_router.effective_candidates()` is called immediately.

This eagerly builds the route's effective candidates and dependant fields when the router is included, instead of delaying this work until the first request.

The change is:

```python
included_router = _IncludedRouter(
    original_router=router, include_context=include_context
)
self.routes.append(included_router)
self._mark_routes_changed()
included_router.effective_candidates()
## AI Disclaimer

I used ChatGPT to better understand the reported bug, the relevant FastAPI routing behavior, and some implementation details while working on this contribution. I implemented and tested the changes in my local repository.

<details>

The AI was used for technical guidance and clarification while working on the contribution.

</details>

## Checklist

- [  ] This PR links to a GitHub Discussion for the proposed code change.
- [ x ] I added tests for the change.
- [ x ] The new or updated tests fail on the main branch and pass on this PR.
- [  ] Coverage stays at 100%.
- [  ] The documentation explains the change if needed.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -3306,10 +3306,12 @@ class APIRouter(routing.Router):
             include_in_schema=include_in_schema,
             generate_unique_id_function=generate_unique_id_function,
         )
-        self.routes.append(
-            _IncludedRouter(original_router=router, include_context=include_context)
+        included_router = _IncludedRouter(
+            original_router=router, include_context=include_context
         )
+        self.routes.append(included_router)
         self._mark_routes_changed()
+        included_router.effective_candidates()
         for handler in router.on_startup:
             self.add_event_handler("startup", handler)
         for handler in router.on_shutdown:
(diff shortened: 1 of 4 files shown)


YuriiMotov:
See https://github.com/fastapi/fastapi/discussions/16034

PR #16273, opened 2026-08-30
🐛 Fix OpenAPI response body schema check for response classes with default status code

Vansh-Sharmaa:
#### Description
When an endpoint doesn't explicitly specify `status_code` in the decorator (e.g. `@app.get("/items", response_class=NoContentResponse)`), `route.status_code` is `None`. 

In `fastapi/openapi/utils.py`, `status_code` is correctly extracted from the `response_class.__init__` signature (e.g. `"204"` or `"304"`), but the subsequent check called `is_body_allowed_for_status_code(route.status_code)`. Because `route.status_code` was `None`, `is_body_allowed_for_status_code(None)` returned `True`.

This caused FastAPI to generate a response body (`content: {"application/json": ...}`) under `"204"` / `"304"` in OpenAPI schemas, which violates HTTP and OpenAPI specifications.

#### Changes
- In `fastapi/openapi/utils.py`, changed `is_body_allowed_for_status_code(route.status_code)` to `is_body_allowed_for_status_code(status_code)`.
- Added test coverage in `tests/test_response_code_no_body.py` for response classes with a default `204` status code and verified OpenAPI snapshot.

#### Checklist
- [x] Ran `pytest` suite
- [x] Ran `ruff check` and `ruff format`

### DIFF
--- a/fastapi/openapi/utils.py
+++ b/fastapi/openapi/utils.py
@@ -416,7 +416,7 @@ def get_openapi_path(
             operation.setdefault("responses", {}).setdefault(status_code, {})[
                 "description"
             ] = route.response_description
-            if is_body_allowed_for_status_code(route.status_code):
+            if is_body_allowed_for_status_code(status_code):
                 # Check for JSONL streaming (generator endpoints)
                 if route.is_json_stream:
                     jsonl_content: dict[str, Any] = {}
--- a/tests/test_response_code_no_body.py
+++ b/tests/test_response_code_no_body.py
@@ -1,4 +1,6 @@
-from fastapi import FastAPI
+from typing import Any
+
+from fastapi import FastAPI, Response
 from fastapi.responses import JSONResponse
 from fastapi.testclient import TestClient
 from inline_snapshot import snapshot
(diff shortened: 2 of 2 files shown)


Shriprasad-P (changes requested):
The production change looks correct: the OpenAPI response entry is keyed by the locally resolved effective `status_code`, so the body-eligibility check should use that same value rather than `route.status_code`.

There is one regression-test issue that should be fixed before merge.

`NoContentResponse` inherits `Response.media_type = None`, so the OpenAPI content-generation branch is already skipped on the base commit. As a result, the new `/c` OpenAPI assertion passes both before and after the production change and does not actually guard this bug.

Please give the test response a non-null media type, e.g. `application/json`, or use an appropriate response subclass with a default 204 status. The base should then emit the incorrect `content` entry while this PR should omit it.

The implementation itself otherwise looks correct.

Vansh-Sharmaa:
Thanks for the sharp catch and review, @Shriprasad-P!

You're completely right — since `NoContentResponse` inherited `media_type = None` from `Response`, the OpenAPI content generation was previously skipped regardless of the status code check.

I have updated `NoContentResponse` to set `media_type = "application/json"` in `tests/test_response_code_no_body.py` (commit `179e8311f`). With this change:
- Unpatched `master` produces an invalid `content` body for the 204 response.
- This PR branch properly omits the `content` body, verifying the fix as an effective regression test.

All checks and tests are passing cleanly. Ready for re-review!

YuriiMotov:
@Vansh-Sharmaa, thanks for your interest and efforts, but please, follow the rules of the repository.

>

As for the issue itself, I think that response class named "NoContentResponse" with `media_type` specified looks too artificial.. I can't come up with a real-life use case where we need to specify defaults status code 204/304 and `media_type` at the same time.

I will close this for now to keep things clean. If you think the issue is important, please open a discussion and share a real-life use case that is affected by it.

Thanks for understanding!

PR #16276, opened 2026-08-30
🌐 Improve translation consistency in ‎docs/pt/docs/virtual-environments.md -  #16276

kaique321123:
This PR improves translation quality in the Portuguese documentation 
for the "Virtual Environments" guide.

Changes:
- Fixed a translation error/confusion found during the review of the automatically translated text.

Notes:
These changes are focused on improving translation quality, without 
modifying technical behavior or meaning.

### DIFF
--- a/docs/pt/docs/virtual-environments.md
+++ b/docs/pt/docs/virtual-environments.md
@@ -1,6 +1,6 @@
 # Ambientes Virtuais { #virtual-environments }
 
-Ao trabalhar com projetos Python, você deveria usar um **ambiente virtual** para isolar os pacotes instalados para cada projeto.
+Quando você trabalha com projetos Python, você deveria usar um **ambiente virtual** para isolar os pacotes instalados para cada projeto.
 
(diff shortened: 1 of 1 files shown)


YuriiMotov:
Thanks for your interest!

Do you think this is a necessary improvement (e.g. old version is wrong, not clear or misleading) or just style improvement (old version is understandable, but new version just sounds more natural)?
I asked Copilot and it says that both versions are clear - second just sounds better, but current version is natural for docs.
We should only fix wording if it matters for understanding the docs. We are not aiming to make the translations perfect - we just don't have capacity for this.

We are using AI-powered automation to sync docs translations, so it can overwrite it again on next sync. If we want a reliable result, we should update the [language-specific LLM prompt](https://github.com/fastapi/fastapi/blob/master/docs/pt/llm-prompt.md?plain=1) instead of fixing certain occurrences. 


I'll close this for now. Let me know if you think this is important and we should re-open this PR.

PR #16283, opened 2026-09-01
🐛 Support PEP 695 TypeAliasType in Query, Form, Sequence parameters, and OpenAPI extraction

Vansh-Sharmaa:
### Description
When using Python 3.12+ PEP 695 type aliases (e.g. `type Tags = list[str]`) or `typing_extensions.TypeAliasType` as parameter annotations:

- `field_annotation_is_sequence()` classified the alias as a scalar instead of a sequence because `get_origin()` returned `None` and `isinstance(TypeAliasType, type)` is `False`.
- Query and Form parameters called `.get(param)` instead of `.getlist(param)`, dropping repeated values and failing validation with HTTP 422.
- `serialize_sequence_value()` raised `TypeError: issubclass() arg 1 must be a class` when invoked on `TypeAliasType`.
- OpenAPI flat model extraction missed nested models wrapped in type aliases.

```python
type Tags = list[str]

@app.get("/items/")
def read_items(tags: Annotated[Tags, Query()]):
    return {"tags": tags}  # Previously treated as scalar or failed serialization

### DIFF
--- a/fastapi/_compat/shared.py
+++ b/fastapi/_compat/shared.py
@@ -18,6 +18,7 @@ from fastapi.types import UnionType
 from pydantic import BaseModel
 from pydantic.version import VERSION as PYDANTIC_VERSION
 from starlette.datastructures import UploadFile
+from typing_inspection.typing_objects import is_typealiastype
 
 _T = TypeVar("_T")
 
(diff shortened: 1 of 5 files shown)


YuriiMotov:
Thanks for pointing to this and for your efforts, but please, follow the rules of the repository.

>

We already track this in the discussion: https://github.com/fastapi/fastapi/discussions/16100

I will close this for now to keep things clean. I added the link of this PR to the internal task, we will review it when we have time to work on it.

Thanks for understanding!

PR #16294, opened 2026-09-01
🐛 Fix optional Json[list[T]] parameters treated as repeated values

sankalpsthakur:
Optional `Json[list[T]]` parameters (`Json[list[T]] | None`) were treated as repeated sequence fields after 0.140.10.

`_is_json_field()` only inspected top-level `field_info.metadata`. Pydantic hoists the `Json` marker there for required fields, but leaves it nested in the union arm for `Json[list[T]] | None`. Meanwhile `field_annotation_is_sequence()` now unwraps `Annotated`, so `_get_multidict_value()` called `.getlist()` and wrapped the JSON string in a list. Pydantic then rejected it with `json_type`.

This change walks union and `Annotated` annotations for `Json` markers so Form, Query, Header, and Cookie fields keep using `.get()`. Required `Json[list[T]]` behavior is unchanged.

Fixes https://github.com/fastapi/fastapi/discussions/16220

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -67,7 +67,7 @@ from fastapi.dependencies.models import (
 from fastapi.exceptions import DependencyScopeError
 from fastapi.logger import logger
 from fastapi.security.oauth2 import SecurityScopes
-from fastapi.types import DependencyCacheKey
+from fastapi.types import DependencyCacheKey, UnionType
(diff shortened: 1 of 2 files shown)


YuriiMotov:
@sankalpsthakur, thanks for your interest and efforts, but please, follow the rules of the repository.

>

Thanks for understanding!

PR #16305, opened 2026-09-05
Fix APIRouter.routes direct mutation cache invalidation

### LINKED ISSUE #16301: Direct mutation of APIRouter.routes bypasses the routes-version cache, causing stale/ghost routes

## Description

## Minimal repro

```python

app = FastAPI()
child = APIRouter()

@child.get("/child")
def read_child():
    return {"msg": "v1-original"}

client = TestClient(app)

client.get("/api/child")

def read_child2():

child.routes.remove(original)

r2 = client.get("/api/child")
```

```
```

## Expected behavior

## Actual behavior

## Versions

vinitsonawane45:
Description

Fixes #16301.

APIRouter.routes was a regular mutable list. Direct mutations such as append(), remove(), and clear() bypassed _mark_routes_changed(), leaving _IncludedRouter with a stale cached candidate list after the cache had been populated.

This could cause newly added routes to return 404 or removed routes to remain reachable as stale "ghost" routes.

This change introduces _RouteList, a list subclass that invalidates the route version when the route list is mutated. APIRouter wraps the Starlette-created route list with _RouteList during initialization.

The existing route-version propagation and _IncludedRouter cache logic are reused, so there is no per-request route scanning or cache rebuild.

The implementation covers the relevant list mutation operations, including slice assignment and deletion, while preserving list compatibility.

Tests

9 direct-mutation regression tests passed
201 routing tests passed, 1 skipped
430 broader tests passed, 1 skipped

The regression tests reproduce the original stale-cache behavior on the main branch and verify that direct mutations correctly invalidate the cache.

Checklist

- [x] Coverage stays at 100%.

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -35,6 +35,7 @@ from typing import (
     Any,
     Literal,
     Protocol,
+    SupportsIndex,
     TypeVar,
     cast,
 )
@@ -2275,7 +2276,7 @@ class _RouteList(list):  # type: ignore[type-arg]
         super().append(item)
         self._owner._mark_routes_changed()
 
-    def insert(self, index: int, item: Any) -> None:  # type: ignore[override]
+    def insert(self, index: SupportsIndex, item: Any) -> None:
         super().insert(index, item)
         self._owner._mark_routes_changed()
 
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -2252,6 +2252,72 @@ class _FrontendRouteGroup(BaseRoute):
                 scope["fastapi_function_astack"] = previous_function_astack
 
 
+class _RouteList(list):  # type: ignore[type-arg]
+    """
+    A ``list`` subclass used for :attr:`APIRouter.routes` that automatically
+    calls :meth:`APIRouter._mark_routes_changed` whenever the list is mutated
+    through any of the standard list mutation methods.
+
+    This ensures that direct mutation of ``router.routes`` (e.g.
+    ``router.routes.append(route)`` or ``router.routes.remove(route)``) is
+    reflected in the route-version counter so that the caches inside any
+    :class:`_IncludedRouter` that wraps this router are correctly invalidated
+    on the next request dispatch.
+    """
+
+    def __init__(self, owner: "APIRouter", iterable: list[Any]) -> None:
+        super().__init__(iterable)
+        self._owner = owner
+
+    # -- mutating methods that need to bump the version -----------------------
+
+    def append(self, item: Any) -> None:
+        super().append(item)
+        self._owner._mark_routes_changed()
+
+    def insert(self, index: int, item: Any) -> None:  # type: ignore[override]
+        super().insert(index, item)
+        self._owner._mark_routes_changed()
+
+    def extend(self, items: Any) -> None:
+        super().extend(items)
+        self._owner._mark_routes_changed()
+
+    def remove(self, item: Any) -> None:
+        super().remove(item)
+        self._owner._mark_routes_changed()
+
+    def pop(self, *args: Any) -> Any:
+        result = super().pop(*args)
+        self._owner._mark_routes_changed()
+        return result
+
+    def clear(self) -> None:
+        super().clear()
+        self._owner._mark_routes_changed()
+
+    def __setitem__(self, index: Any, value: Any) -> None:
+        super().__setitem__(index, value)
+        self._owner._mark_routes_changed()
+
+    def __delitem__(self, index: Any) -> None:
+        super().__delitem__(index)
+        self._owner._mark_routes_changed()
+
+    def __iadd__(self, items: Any) -> "_RouteList":  # type: ignore[misc]
+        super().__iadd__(items)
+        self._owner._mark_routes_changed()
+        return self
+
+    def sort(self, *args: Any, **kwargs: Any) -> None:
+        super().sort(*args, **kwargs)
+        self._owner._mark_routes_changed()
+
+    def reverse(self) -> None:
+        super().reverse()
+        self._owner._mark_routes_changed()
+
+
 class APIRouter(routing.Router):
     """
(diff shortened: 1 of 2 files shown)


BitWeaverDev:
I checked this out locally in an isolated worktree and ran it against a real app, including the PR's own tests. The in place mutation coverage genuinely works. I confirmed `append`, `remove`, `insert`, `extend`, `pop`, `clear`, `__setitem__`, `__delitem__`, `__iadd__`, `sort`, and `reverse` all correctly bump `_routes_version` and are visible on the next request, including through nested router inclusion, and including for routes added via `APIKeyQuery` style query and header mutation paths that the PR's own test file does not exercise directly.

There is one gap that matters a lot here. Issue #16301 frames the bug using this exact language: "a documented pattern elsewhere (e.g. `app.router.routes = [...]` to remove routes)." I reproduced that specific case against this branch.

```
warmup: 200
after reassignment, new route reachable? expected 200, got: 404
after reassignment, old route gone? expected 404, got: 200
routes is still _RouteList after reassignment? list
```

`self.routes = _RouteList(self, self.routes)` only runs once in `__init__`. There is no property setter on `routes`, so `router.routes = new_list` overwrites the attribute with a plain list and silently reintroduces the exact stale cache bug this PR sets out to fix, for that one mutation form. `routes *= n` has the same problem since `__imul__` is not overridden either.

There is also a question worth raising before this goes further. `docs/en/docs/tutorial/bigger-applications.md` already tells people to avoid directly mutating `router.routes` after including a router, and to use `.include_router()` and the decorators instead. So the behavior in question is explicitly discouraged, not a supported pattern. This PR adds a stateful list subclass with an owner back reference wired into every `APIRouter` for its whole lifetime, eleven overridden methods, and a `type: ignore`, all to make that discouraged pattern behave more consistently, and it still does not make it fully consistent because of the reassignment case above. It might be worth deciding whether the goal is to fully support direct mutation (in which case a property setter is needed too, similar to what #16306 attempts) or to leave the docs as the actual boundary and scope this down.

A few smaller things.

`class _RouteList(list):  # type: ignore[type-arg]` can just be `class _RouteList(list[BaseRoute]):`. The file already uses `list[BaseRoute]` and `X | None` throughout, so the ignore is avoidable.

`add_route`, `add_websocket_route`, `add_api_route`, `add_api_websocket_route`, and `include_router` lost their explicit `self._mark_routes_changed()` calls in favor of a comment saying it now happens implicitly inside `list.append()`. That turns a call you can see and grep for into invisible behavior behind a builtin method, which is part of why the reassignment gap above is easy to miss on review.

`_owner` creates a reference cycle between the router and its own `routes` attribute. Not harmful under garbage collection, but a plain callback function would decouple `_RouteList` from `APIRouter` and be testable on its own.

The comment placed right after `super().__init__(...)` describes wrapping that only happens about fifteen lines later, and the explanation is then repeated again at the actual wrap site.

No test covers `router.routes = [...]` or `router.routes[:] = [...]`, which is the specific pattern named in the linked issue, so the test suite never touches the one scenario that currently fails.

Overall this is a solid fix for the in place mutation methods, but it does not close the case the linked issue actually leads with, and I think that needs to be resolved (either with a setter or with an explicit scope decision) before merging.

vinitsonawane45:
@BitWeaverDev Thanks for catching the reassignment case.

I've updated the implementation to add an `APIRouter.routes` property + setter, so `router.routes = [...]` remains an `_RouteList` and invalidates the route-version cache.

I've also added regression tests covering reassignment, cache warmup, old/new route reachability, nested routers, slice assignment, and empty-list reassignment.

The focused regression tests and routing tests pass locally, and Ruff passes as well.

Could you please take another look at the updated implementation in `bf2c1ad`?

BitWeaverDev:
Checked `bf2c1ad` in an isolated worktree. The reassignment gap is fixed. I reran my original repro against this commit plus a few extra forms.

```
plain reassignment: new route 200, old route 404, type stays _RouteList
slice reassignment (routes[:] = [...]): new route 200, old route 404
empty list reassignment (routes = []): old route 404
generator reassignment (routes = (r for r in [...])): new route 200, stays _RouteList
```

All correct now. Ran `tests/test_routes_direct_mutation_cache.py` plus `test_router*.py` and `test_include_router*.py`, 113 passed. Ran the full suite outside `test_tutorial`, 2054 passed, 14 skipped, one failure in `test_fastapi_cli` that traces to a missing `coverage` module in my venv, not this change. No regressions from the property.

A few small things left, none of them blocking.

`class _RouteList(list):  # type: ignore[type-arg]` still needs the ignore. `class _RouteList(list[BaseRoute]):` should work given the file already uses that syntax elsewhere.

`self._routes_version = 0` is still set after `super().__init__()`, which is why the setter needs `hasattr(self, "_routes_version")` to skip the bump during bootstrap. Setting `_routes_version` before calling `super().__init__()` would remove the need for that check.

In the setter, the branch that keeps the existing `_RouteList` as is still falls through to `self._mark_routes_changed()` afterward. So `router.routes = router.routes` bumps the version and invalidates every cache in the inclusion tree even though nothing changed. Not harmful, just wasted work, and it contradicts what that branch's own comment says it is doing. Moving the bump inside the `else` only would fix it.

`__imul__` is still not overridden, so `routes *= n` is still a blind spot, same category as the reassignment case just now fixed. Unlikely to matter much in practice but worth a note or a fix for completeness.

Nice work closing the main gap and matching the repro in the new tests directly.

vinitsonawane45:
@BitWeaverDev I've applied the small cleanup fixes you suggested:

* `_RouteList` now uses `list[BaseRoute]`
* `_routes_version` is initialized before `super().__init__()`
* assigning `router.routes = router.routes` no longer bumps the version
* `__imul__` is now handled consistently

The focused and routing tests still pass, and Ruff/mypy show no new issues.

The changes are pushed in commit `070f398b9`. Please take another look when you have a chance.

YuriiMotov:

>

Thanks for understanding!

PR #16306, opened 2026-09-05
fix: invalidate included router cache on direct APIRouter.routes mutation (#16301)

### LINKED ISSUE #16301: Direct mutation of APIRouter.routes bypasses the routes-version cache, causing stale/ghost routes

vaibhavmashal:
### Description
When \APIRouter.routes\ was directly mutated (e.g., via \.append()\, \.remove()\, \.pop()\, \.clear()\, \.extend()\, \__setitem__\, \__delitem__\, in-place \+=\, or reassignment \
outer.routes = [...]\), the router's internal \_routes_version\ was not incremented. Consequently, \_IncludedRouter.effective_candidates()\ (and OpenAPI schema cache) served stale or ghost routes from its cache.

### Changes
- Implemented \_RoutesList\ subclassing \list[BaseRoute]\ with mutation notifications triggering \_mark_routes_changed()\.
- Added property getter and setter for \APIRouter.routes\ to wrap routes in \_RoutesList\ and ensure assignments increment \_routes_version\.
- Added regression tests in \	ests/test_router_cache_invalidation.py\ covering direct route addition, removal, popping, clearing, slicing, deletion, reassignment, nested router propagation, Starlette Route mutation, and OpenAPI schema cache invalidation.

Fixes #16301

### DIFF
--- a/fastapi/routing.py
+++ b/fastapi/routing.py
@@ -2715,7 +2715,6 @@ class APIRouter(routing.Router):
             name=name,
             include_in_schema=include_in_schema,
         )
-        self._mark_routes_changed()
 
     def add_websocket_route(
         self,
@@ -2724,7 +2723,6 @@ class APIRouter(routing.Router):
         name: str | None = None,
     ) -> None:
         super().add_websocket_route(path, endpoint, name=name)
-        self._mark_routes_changed()
 
     def frontend(
         self,
@@ -3067,7 +3065,6 @@ class APIRouter(routing.Router):
             ),
         )
         self.routes.append(route)
-        self._mark_routes_changed()
 
     def api_route(
         self,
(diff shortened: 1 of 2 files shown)


vaibhavmashal:
﻿Thanks for catching that!

In commit `2221598`, I've removed the redundant explicit `_mark_routes_changed()` calls from `add_route`, `add_websocket_route`, `add_api_route`, `add_api_websocket_route`, and `include_router`. Since `_RoutesList` already triggers `_mark_routes_changed()` on `append()`/mutations, this ensures each route registration or route-list mutation increments `_routes_version` exactly once, avoiding duplicate invalidations or superfluous candidate cache rebuilds.

Added unit test coverage verifying exact single-increment behavior across route additions and nested router inclusions. All 11 tests in the suite pass cleanly.

YuriiMotov:

>

Thanks for understanding!

PR #16326, opened 2026-09-06
fix(routing): invalidate routes version on direct APIRouter.routes mutations

### LINKED ISSUE #16301: Direct mutation of APIRouter.routes bypasses the routes-version cache, causing stale/ghost routes

taran-dev4u:
### Summary

Direct mutations to `APIRouter.routes` (such as `routes.append()`, `routes.extend()`, `routes.remove()`, `routes.pop()`, `routes.clear()`, in-place indexing `routes[i] = ...`, or whole list reassignment `router.routes = [...]`) previously bypassed the internal `_routes_version` tracking introduced in the lazy router inclusion subsystem (#15745). Because `_IncludedRouter` caches effective route candidates keyed by `_routes_version`, routes added after inclusion were never discovered (returned 404), and removed routes continued to be served as ghost 200 routes.

This PR introduces an observable route sequence `_ObservableRouteList` subclassing `list` and wires the `routes` attribute on `APIRouter` to notify `self._mark_routes_changed()` on any list mutation or reassignment.

### Changes

1. **`_ObservableRouteList` implementation in `fastapi/routing.py`**:
   - Subclasses `list` and intercepts mutation methods: `append`, `extend`, `insert`, `pop`, `remove`, `clear`, `reverse`, `sort`, `__setitem__`, `__delitem__`, `__iadd__`.
   - Each modifying method calls `self._router._mark_routes_changed()`, bumping `_routes_version`.

2. **`APIRouter.routes` Property & Setter**:
   - `@property def routes(self) -> list[BaseRoute]` returns the underlying `_ObservableRouteList`.
   - `@routes.setter def routes(self, value: Sequence[BaseRoute]) -> None` wraps incoming sequences into `_ObservableRouteList` and triggers `_mark_routes_changed()`.

3. **Comprehensive Tests in `tests/test_router_routes_mutation.py`**:
   - `test_direct_routes_append_invalidates_cache`: Verifies direct `.append()` route is resolved instead of 404.
   - `test_direct_routes_remove_invalidates_cache`: Verifies direct `.remove()` drops the route and returns 404 instead of serving stale/ghost responses.
   - `test_direct_routes_extend_and_pop`: Verifies `.extend()` and `.pop()` update router resolution correctly.
   - `test_direct_routes_reassignment`: Verifies `router.routes = [...]` invalidates cache and routes appropriately.
   - `test_direct_routes_clear`: Verifies `.clear()` drops all routes.
   - `test_direct_routes_mutation_refreshes_openapi_schema`: Verifies direct mutations refresh `app.openapi()` schema paths.

### Validation

- `mypy fastapi/routing.py tests/test_router_routes_mutation.py` passed with 0 issues.
- `ruff check` and `ruff format --check` passed cleanly.
- `pytest tests/test_router_routes_mutation.py` (6/6 tests passed).
- Full router test suite (145 tests passed 100%).
- Fixes #16301.


YuriiMotov:

>

Thanks for understanding!

PR #16327, opened 2026-09-06
✨ Add `ASGIMiddleware`, a base class to create middleware without the overhead of `BaseHTTPMiddleware`

HamzaYslmn:
## What

A base class, `fastapi.middleware.asgi.ASGIMiddleware`, to write middleware with the same ergonomics as `@app.middleware("http")` but as pure ASGI underneath.

```python
import time

from fastapi import FastAPI, Request
from fastapi.datastructures import MutableHeaders
from fastapi.middleware.asgi import ASGIMiddleware

app = FastAPI()


class ProcessTimeMiddleware(ASGIMiddleware):
    async def on_request(self, request: Request) -> None:
        request.state.start_time = time.perf_counter()

    async def on_response(
        self, request: Request, status_code: int, headers: MutableHeaders
    ) -> None:
        process_time = time.perf_counter() - request.state.start_time
        headers["X-Process-Time"] = str(process_time)


app.add_middleware(ProcessTimeMiddleware)
```

Both hooks are optional, so a middleware that only touches responses implements only `on_response()`.

## Why

`@app.middleware("http")` builds a `BaseHTTPMiddleware`, whose `call_next` contract forces a task group and a memory object stream per request. That has two consequences the docs already acknowledge.

The first is cost. Measured on this branch, one route returning a small JSON body, one uvicorn worker, keep-alive clients over raw sockets, average of two 8 second runs:

| stack | 16 concurrent | p50 | 64 concurrent | p50 |
| --- | --- | --- | --- | --- |
| no middleware | 8000 req/s | 2.14 ms | 8536 req/s | 8.07 ms |
| `@app.middleware("http")` | 4396 req/s | 3.67 ms | 4803 req/s | 12.74 ms |
| `ASGIMiddleware` | 7425 req/s | 2.31 ms | 7845 req/s | 8.84 ms |
| hand-written pure ASGI | 7509 req/s | 2.23 ms | 8171 req/s | 8.44 ms |

That is 1.69x the throughput of `@app.middleware("http")` at 16 concurrent and 1.63x at 64, landing within about 1% of hand-written pure ASGI. The tail is the bigger difference: p95 at 64 concurrent is 24.29 ms with `BaseHTTPMiddleware` against 9.69 ms here.

The second is `contextvars`. Starlette's middleware docs list it under "Limitations": a value set on a `ContextVar` inside an endpoint is not visible to a `BaseHTTPMiddleware` above it, and a `BaseHTTPMiddleware` early in the stack breaks propagation for every pure ASGI middleware after it. `ASGIMiddleware` adds no task group, so `on_response()` sees what the path operation and its dependencies set. There is a test for exactly this.

## Prior art

Starlette explored the same idea in [encode/starlette#1691](https://github.com/Kludex/starlette/pull/1691), a generator based `HTTPMiddleware`. It was closed with the conclusion that users should be pushed "towards writing pure ASGI middleware or solutions provided by higher-level packages like FastAPI". This is that solution, and it takes the lesson from that discussion: no body rewriting, no `call_next`, nothing that needs a second task.

## Design notes

- Non-HTTP scopes (websocket, lifespan) are passed straight through.
- `on_response()` runs on `http.response.start`, before any byte is written, so the response body is never buffered and streaming responses keep streaming.
- Returning a `Response` from `on_request()` answers immediately and the path operation is never called.
- Reading the body in `on_request()` with `await request.body()` works. The consumed ASGI messages are recorded and replayed to the app, so the path operation still parses its own body. When the hook does not read the body, the original `receive` is passed down untouched and there is no wrapper at all. Measured cost of a hook that does read the body on every request: 7109 req/s against 7425, so about 4%.
- `request.state` is the scope's state, so anything set in `on_request()` reaches the path operation and `on_response()`.
- Exceptions are deliberately not caught. They keep propagating to `ServerErrorMiddleware` and to any custom exception handler, unchanged.

## Also included

`MutableHeaders` is now re-exported from `fastapi.datastructures`, next to `Headers`, `QueryParams` and the rest, because it appears in the public `on_response()` signature.

## Tests and docs

- `tests/test_asgi_middleware.py` covers state sharing, `ContextVar` propagation from a path operation, early responses, body replay through a Pydantic model, streaming, exception passthrough and websockets. `fastapi/middleware/asgi.py` is at 100% coverage.
- New section "Creating ASGI middlewares" in Advanced Middleware, with `docs_src/advanced_middleware/tutorial004_py310.py` and its test, a pointer from Tutorial - Middleware, and an entry in the API reference.
- Full suite passes: 3336 passed, 37 skipped, 5 xfailed. `mypy fastapi`, `ruff check`, `ruff format --check` and `ty check` are all clean, with no new diagnostics.

## Benchmark reproduction

Four apps, identical route and identical work (stamp `X-Process-Time`), differing only in how the middleware is written. Load is generated with keep-alive HTTP/1.1 over raw asyncio sockets so the client is not the bottleneck, with a warm-up run discarded before each measurement. Numbers above are from a Windows 11 box, Python 3.11, uvicorn 0.40, access log off, single worker. Absolute values are machine specific, the ratios are what the change is about.


YuriiMotov:
@HamzaYslmn, thanks for your interest and efforts, but please, follow the rules of the repository.

>

Thanks for understanding!

HamzaYslmn:
discussions https://github.com/fastapi/fastapi/discussions/16332

PR #16329, opened 2026-09-06
Fix stale effective-routes cache on direct APIRouter.routes mutation

tusharsingh3199:
### What's wrong

`_IncludedRouter.effective_candidates()` caches a router's resolved route list, invalidated by `_routes_version`. Every official mutator (`add_route`, `add_api_route`, `include_router`, etc.) correctly bumps this counter, but `APIRouter.routes` is a plain list, so direct mutation(`router.routes.append(...)`, `.remove(...)`, or full reassignment like `router.routes = [...]`) never bumps it. A parent router that has already dispatched at least one request keeps serving a stale cached list — newly added routes 404, and removed routes keep responding as "ghost" routes.

### Fix

Make `.routes` mutation-aware so that every way of changing it — the official methods and direct list mutation/reassignment alike — reliably triggers the existing `_mark_routes_changed()` invalidation. Because `_get_routes_version()` already recursively includes each included router's own version, invalidating at the point of mutation is sufficient; no changes are needed to `_IncludedRouter` or `effective_candidates()`.

### Testing

Added tests covering direct `append`, `remove`, and full-list reassignment, plus the less common mutation paths (`extend`, `insert`, `pop`, `clear`, `sort`, `reverse`, item/slice assignment, `+=`) to avoid only fixing the two cases shown in the original repro. Verified the new tests fail on `master` and pass with the fix, and confirmed 100% line coverage on the changed code.


YuriiMotov:

>

Thanks for understanding!

PR #16335, opened 2026-09-08
Fix: unwrap Annotated before stream origin detection

### LINKED ISSUE #16334: Annotated[AsyncIterator[T], ...] return annotations fail route registration
# `Annotated[AsyncIterator[T], ...]` return annotations fail route registration

## Summary

On FastAPI revision `50113da16fec53b66b80d75e80a89296de4fa5a5`, wrapping an
async-stream return annotation in `typing.Annotated` causes route registration
to raise `FastAPIError`, while the equivalent unwrapped annotation registers
successfully.

This is distinct from the `APIRouter.include_router` propagation issue in
#15401: the failure happens while the route is first constructed, even when it
is declared directly on `FastAPI`.

## Reproduction

```python
from typing import Annotated

from fastapi import FastAPI


def plain_app() -> FastAPI:
    app = FastAPI()

    @app.get("/plain")
    async def plain() -> AsyncIterator[int]:
        yield 1

    return app


def annotated_app() -> FastAPI:
    app = FastAPI()

    @app.get("/annotated")
    async def annotated() -> Annotated[AsyncIterator[int], "stream-item"]:
        yield 1

    return app


plain_app()       # succeeds
annotated_app()   # raises FastAPIError: Invalid args for response field
```

The reproduction was run against the current checkout with the same result
for `typing.AsyncIterator` and `collections.abc.AsyncIterator`.

## Actual behavior

`get_stream_item_type()` checks only whether the outer `get_origin(annotation)`
is one of the supported stream origins. For an `Annotated` return annotation,
the outer origin is `Annotated`, so the helper returns `None`. The complete
`Annotated[AsyncIterator[int], ...]` object is then treated as a response model
and route construction fails before the application can start.

Relevant code:

- `fastapi/dependencies/utils.py:261-268`, `get_stream_item_type()`
- `fastapi/routing.py:1086-1087`, where the helper is called

## Expected behavior

`Annotated[AsyncIterator[int], metadata]` should be accepted wherever the
equivalent `AsyncIterator[int]` annotation is accepted. The metadata should be
preserved or ignored according to FastAPI's existing `Annotated` conventions,
and the stream item type should resolve to `int`.

## Impact

Applications using metadata-bearing return annotations cannot start. This is a
compatibility failure for valid PEP 593 typing syntax and affects deployment
before the first request. It also prevents the framework from generating the
intended stream item schema.

## Suggested fix

Unwrap `Annotated` before stream-origin detection (or recurse through it inside
`get_stream_item_type()`), matching the existing annotation-unwrapping logic
used for parameters. Add regression coverage for `Annotated[AsyncIterator[T],
...]` and `Annotated[AsyncGenerator[T, None], ...]` return annotations.

## Environment

- FastAPI revision: `50113da16fec53b66b80d75e80a89296de4fa5a5`
- Python: 3.13
- Pydantic: 2.x

## Duplicate check

- FastAPI issue #15401 concerns `APIRouter.include_router` losing `stream_item_type` after a route was already constructed; this reproduction fails during direct route registration and does not require a router merge.
- The current FastAPI source still contains the same outer-origin-only `get_stream_item_type()` logic in `fastapi/dependencies/utils.py`, so the finding remains independently reproducible against the reviewed revision.
- This report is published as public issue #16334.

OmmprakashMohanty01:
Fixes #16334. Unwrap `Annotated` types in `get_stream_item_type` so that stream origin detection works correctly for `Annotated[AsyncIterator[T], ...]`.


YuriiMotov:
@OmmprakashMohanty01, thanks for your interest and efforts, but please, follow the rules of the repository.

>

Thanks for understanding!

PR #16336, opened 2026-09-08
fix: defensively filter empty query parameter keys to prevent downstr…

hariom123-dev:
### TL;DR
Adds API hardening to `fastapi/dependencies/utils.py::request_params_to_args` to explicitly drop empty string keys (`""`) from query parameters and headers, preventing malformed requests from crashing downstream Pydantic validation.

### 🔍 Context & Motivation
In production environments, APIs frequently receive malformed web traffic (e.g., from scrapers, legacy clients, or dangling URL ampersands like `?=&name=test&&`). 

When these malformed requests hit FastAPI, Starlette parses the empty keys as `"" = ""`. If these empty keys are passed into the dependency resolver, they can bypass type hints and crash deep inside Pydantic, resulting in a `500 Internal Server Error` instead of a graceful `422` or a successful `200`. This creates unnecessary false-positive alert fatigue for DevOps teams.

### 🛠️ Changes Implemented
- In `request_params_to_args`, added an `O(N)` defensive filter before the `values` extraction loop.
- Gracefully handles both standard dictionaries and Starlette's `ImmutableMultiDict` / `QueryParams` / `Headers`.
- If a key is `""`, it is silently discarded, allowing valid parameters (like `name=test`) to process normally.

### ✅ Verification
- Added `tests/test_query_params_edge_cases.py`.
- Simulated a malformed web client request: `client.get("/items/?=&q=test&")`.
- Asserted the API correctly ignores the empty keys and returns `200 OK` with `{"q": "test"}`.

### DIFF
--- a/fastapi/dependencies/utils.py
+++ b/fastapi/dependencies/utils.py
@@ -793,14 +793,11 @@ def request_params_to_args(
         )
     elif isinstance(received_params, Headers):
         received_params = Headers(
-            raw=[
-                (k.encode(), v.encode()) if isinstance(k, str) else (k, v)
-                for k, v in received_params.raw
-                if k != b""
-            ]
+            raw=[(k.encode("latin-1"), v.encode("latin-1")) if isinstance(k, str) else (k, v)
+                 for k, v in received_params.raw if k != b""]
         )
-    elif isinstance(received_params, type(ImmutableMultiDict())):
-        received_params = type(ImmutableMultiDict())(
+    elif isinstance(received_params, ImmutableMultiDict):
+        received_params = ImmutableMultiDict(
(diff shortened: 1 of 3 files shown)


hariom123-dev:
After reviewing the CI matrix failures, I'm going to close this PR. 

While the defensive filter successfully drops the empty keys, casting the `ImmutableMultiDict` / `QueryParams` to a standard dictionary strips the `.getlist()` method, which breaks array query parameter extraction downstream. 

Implementing this safely would likely require upstream changes to Starlette's data structures to allow immutable filtering without losing the `MultiDict` interface. Thank you for the CI pipeline—it caught the exact architectural boundary!

PR #14441, opened 2025-12-03
⚡️Improve caching of callable inspection result

dolfinus:
Motivation
----------

Currently if `app.depedendency_overrides` is not empty then each request generates new Dependant, instead of reusing one build while declaring a route. This affects performance of all requests, even if specific dependency is not actually overridden. I'm personally use dependency overrides on production apps (see #13913), so all my applications are affected.

Change description
------------------

If dependency is not overridden, reuse existing Dependant with cached `inspect` results. This avoids performance hit, in my case 460RPS -> 630RPS.

Also simplified check of `dependency_overrides` makes resolution of complex dependency trees faster, even when no dependency overrides used in the first place.

Microbenckmark:
<details>

```python
import asyncio
import time
from typing import Annotated

from pydantic import BaseModel

from fastapi import FastAPI, Depends, Query, Response
from httpx import AsyncClient, ASGITransport


async def test_no_dependency_override_performance(loop_count: int = 100_000):
    app = FastAPI()

    async def sub_sub_dependency(more_nested: Annotated[int, Query()]) -> int:
        return more_nested

    class SubDependency(BaseModel):
        value: Annotated[int, Query()]
        other: Annotated[int, Depends(sub_sub_dependency)]

    async def some_dependency(nested: Annotated[SubDependency, Depends()]) -> int:
        return nested.value + nested.other

    @app.get("/data")
    async def get(
        one: int = Depends(some_dependency),
    ):
        return Response(content=f'{{"value": {one}}}'.encode(), status_code=200)

    async with AsyncClient(
        transport=ASGITransport(app=app),
        base_url="http://localhost",
    ) as client:
        start_time = time.perf_counter()

        for _ in range(loop_count):
            response = await client.get("/data", params={"value": 1, "more_nested": 2})
            assert response.status_code == 200, response.text
            assert response.text == '{"value": 3}'

        duration = time.perf_counter() - start_time

    print(f"Performed GET /data requests without dependency override: {loop_count:,} times in {duration:.4f}s")
    print(f"Average per call: {duration / loop_count * 1e6:.2f} µs")


async def test_dependency_override_nonempty_performance(loop_count: int = 100_000):
    app = FastAPI()

    async def sub_sub_dependency(more_nested: Annotated[int, Query()]) -> int:
        return more_nested

    class SubDependency(BaseModel):
        value: Annotated[int, Query()]
        other: Annotated[int, Depends(sub_sub_dependency)]

[...]

PR #16342, opened 2026-09-09
Avoid rebuilding unchanged dependencies when overrides are configured

fagoneszy:
When any dependency override is configured, request dependency resolution rebuilds every dependency graph, including dependencies whose callable has not changed. An unrelated override therefore adds signature analysis and recursive graph construction to every request.

Reuse the existing `Dependant` when the resolved callable is the original callable. Actual overrides still build their replacement graph, and recursive resolution still applies overrides to children of unchanged parents.

Regression tests cover unrelated overrides, self-overrides, nested overrides beneath unchanged parents, query validation, and clearing overrides between requests. All three new test cases fail on the base revision and pass with this change.

Related to #8402.

This PR addresses unnecessary dependency graph reconstruction during dependency resolution when dependency overrides are configured. It does not address every performance optimization discussed in #5554.

Local benchmark (Windows, Python 3.12.14; in-process HTTPX ASGI requests; five batches of 100 requests after 20 warm-up requests; median milliseconds per request):

| Dependency chain depth | Before, unrelated override | After, unrelated override | Before, no override | After, no override |
| --- | ---: | ---: | ---: | ---: |
| 1 | 0.895 | 0.434 | 0.445 | 0.441 |
| 10 | 6.999 | 0.601 | 0.598 | 0.594 |
| 20 | 18.530 | 0.782 | 0.760 | 0.751 |

These synthetic measurements illustrate this specific case; they are not general application throughput claims.

Validation:

- New regression tests: 3 failed before, 3 passed after.
- Related dependency, security, scope, and cache tests: 352 passed, 1 skipped.
- Ruff lint and format checks passed for both changed files.
- Mypy passed for all 48 FastAPI source files.
- Ty passed for both changed files.
- Extended `tests/` suite, excluding benchmark directories: 3330 passed, 13 skipped, 5 xfailed; one translation test initially failed because Git was absent. After adding portable Git to PATH, that test passed (1 passed). No test failures remain unresolved. The separate `scripts/tests/` suite and the CI platform/version matrix were not run.


VimalN2005:
Hi @fagoneszy, excellent performance optimization! Skipping graph rebuild for unchanged dependencies gives a noticeable reduction in request latency.

One quick question:
Have you tested how this behaves with `yield`-based dependencies (dependencies with cleanup)? Adding a quick test case for a generator dependency under an unrelated override would ensure complete coverage for lifecycle handling.

YuriiMotov:
@fagoneszy , thanks for your interest and efforts, but please, follow the rules of the repository.

>

Also, we already have PR https://github.com/fastapi/fastapi/pull/14441 that fixes this. You can review it.

Thanks for understanding!

PR #16349, opened 2026-09-13
add X-Trust middleware example to advanced docs

htl-syterme:
I've been dealing with bot abuse on a side project — bots were 
hitting our LLM endpoints and burning through our free tier budget. 
Didn't want to add a captcha because it kills signups.

Found that signing each request with a human score header works 
really well. Built a small middleware around it. The header is 
HMAC-SHA256 signed, expires in 120 seconds, carries a score between 
0 and 1. Nothing is blocked — just annotated so you can decide what 
to do with the information.

Added a minimal example in the advanced middleware docs in case 
others run into the same problem.

Code is at github.com/htl-syterme/htl-core if anyone wants to look 
at the implementation.


YuriiMotov:

PR #16350, opened 2026-09-14
Fix jsonable_encoder raising an unclear UnicodeDecodeError for non-UTF-8 bytes

usamabhatti1998:
### Summary

`jsonable_encoder(some_bytes)` currently calls `some_bytes.decode()` with no error handling. For any `bytes` value that isn't valid UTF-8 (raw binary data — hashes, encrypted blobs, protobuf, DB blob columns), this raises an unhandled `UnicodeDecodeError` straight from deep inside FastAPI's internals, with no indication of what went wrong or how to fix it — even though a working escape hatch (`custom_encoder={bytes: ...}`) already exists and just isn't discoverable from the error.

### Fix

Wrap the decode in a named `bytes_encoder()` function that re-raises as a `ValueError` (consistent with the `ValueError` `jsonable_encoder` already raises for other unencodable types) with a message that explains the problem and points at `custom_encoder` as the fix. The existing pattern of naming non-trivial encoders as top-level functions (e.g. `decimal_encoder`, `isoformat`) is followed rather than keeping an inline lambda.

### Testing

Added three tests to `tests/test_jsonable_encoder.py`:
- `test_encode_bytes_valid_utf8`: valid UTF-8 bytes still encode correctly (unchanged behavior)
- `test_encode_bytes_invalid_utf8_raises_clear_error`: invalid UTF-8 now raises `ValueError` with an actionable message
- `test_encode_bytes_invalid_utf8_with_custom_encoder`: the existing `custom_encoder` escape hatch still works and now has a regression test

Not a breaking change for the success path; the only change is the exception type/message for previously-crashing input.


VimalN2005:
Hi @usamabhatti1998, thanks for improving the error handling for non-UTF-8 bytes! The clearer exception message will save lots of debugging time.
A couple of small suggestions:
1. **Explicit encoding:** In `bytes_encoder()`, consider `o.decode("utf-8")` instead of `o.decode()` to be explicit.
2. **PEP 8 Import:** In `tests/test_jsonable_encoder.py`, `import base64` can be moved to the top of the file alongside other standard library imports.

Shriprasad-P (commented):

PR #16351, opened 2026-09-14
fix(dependencies): preserve Query metadata for complex fields in BaseModel dependencies

FinalSunFlower:
### Description

When a `BaseModel` dependency contains a complex field annotated with `Query()`, such as `Annotated[list[str] | None, Query()] = None`, repeated query values are not parsed into the model.

Pydantic removes the `Annotated` metadata from the generated `inspect.signature`, so FastAPI re-infers the list field as a request body parameter. A request such as `?field=foo&field=bar` therefore produces `field=None`.

### Reproduction

```python
from typing import Annotated

class Params(BaseModel):
    field: Annotated[list[str] | None, Query()] = None

app = FastAPI()

@app.get("/")
def read(params: Annotated[Params, Depends()]):
    return params.model_dump()

response = TestClient(app).get("/?field=foo&field=bar")
assert response.json() == {"field": ["foo", "bar"]}
```

### Solution

- Resolve annotations for `BaseModel` dependency callables with `get_type_hints(..., include_extras=True)`.
- Preserve the original `Query()` metadata during dependency parameter analysis.
- Add a regression test covering repeated list-valued query parameters.

### Verification

`python -m pytest tests/test_query_cookie_header_model_extra_params.py tests/test_dependency_class.py tests/test_dependency_models.py -q`

30 passed in 0.58s.

`ruff check` and `ruff format --check` both pass.


Shriprasad-P (commented):

FinalSunFlower:
@Shriprasad-P Thanks for the review!

Confirmed that test coverage is fully in place for the new code path:

1. **Feature Coverage**: `tests/test_query_cookie_header_model_extra_params.py` verifies that multi-value parameters (such as `list[str]`) inside a `BaseModel` dependency retain `Query()` metadata via `get_type_hints(..., include_extras=True)` and are collected as query parameters instead of being misinferred as body parameters.
2. **Defensive Fallback Coverage**: A dedicated test case exercises the `try...except` fallback branch to ensure standard `inspect.signature` annotations are preserved if `get_type_hints` fails on non-standard callables.
3. **Coverage Check**: The automated CI `coverage-combine` step passes cleanly with 100% test coverage on the modified lines in `fastapi/dependencies/utils.py`.

sylvesterkaczmarek (commented):
This still seems to miss aliased model fields. Pydantic can use the field alias as the generated signature parameter name, while `get_type_hints()` is keyed by the Python field name, so `model_annotations.get(param_name, ...)` falls back to the stripped annotation. An aliased `Annotated[list[…], Query()]` would still be inferred as body data. Could we map signature params back through `model_fields`/aliases and cover that case?

FinalSunFlower:
@sylvesterkaczmarek Good catch. The parameter resolution now handles aliased model fields explicitly.

When a signature parameter name is an alias, the resolver checks the BaseModel field metadata (model_fields, with __fields__ compatibility) and maps the alias back to the original Python field name before retrieving the Annotated[..., Query(...)] annotation from get_type_hints(..., include_extras=True). It also considers validation aliases and preserves the original signature annotation when no match is found or type-hint resolution fails.

I added regression coverage for both the synthetic aliased signature case and an end-to-end repeated query request (itemAlias=foo&itemAlias=bar), confirming the values are parsed as query parameters and no body parameter is generated.

Local verification: python -m pytest tests -q -> 3336 passed, 7 skipped, 5 xfailed; targeted Ruff checks and formatting checks also pass.

sylvesterkaczmarek (approved):
Duplicate follow-up review submitted in error.

sylvesterkaczmarek (approved):
Duplicate follow-up review submitted in error.

sylvesterkaczmarek (commented):
Duplicate follow-up review submitted in error.

FinalSunFlower:
@sylvesterkaczmarek Added a focused `AliasChoices` regression for the non-string `validation_alias` path.

The test defines `Annotated[list[str] | None, Query(validation_alias=AliasChoices(itemAlias, item_ids))]`, supplies an aliased signature parameter, and verifies that the annotation is recovered as a query parameter with no body parameter. This exercises the `validation_alias.choices` branch directly.

Local verification: the targeted coverage run reports the branch as covered, and the dependency suites pass (`34 passed`). The update is pushed as `68aa475`.

PR #16354, opened 2026-09-14
📝 Clarify comment about SSE event terminator

YuriiMotov:
Closes: https://github.com/fastapi/fastapi/pull/16265

## Description

The docstring of `format_sse_event` mistakenly states that each event end up with two end-of-line characters: 
https://github.com/fastapi/fastapi/blob/50113da16fec53b66b80d75e80a89296de4fa5a5/fastapi/sse.py#L210

In fact, according to the [specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream), event can be empty (no comments and no fields) and then it will be just one empty line.
So, two end-of-line characters in the end is only true for non-empty events. First end-of-line character is actually a part of field or comment and second (empty line) is an event terminator.

## AI Disclaimer

I used AI (Codex) to investigate, but it was mostly misleading until I learned how to read the notation in the spec and figured out the solution by myself and explained my thoughts. Then Codex approved my approach.

## Checklist

- [x] Coverage stays at 100%.


Shriprasad-P (commented):

LeonardoSanBenitez:
I checked whether the new test exercises a new code path, since that is not visible from the diff itself: the hunk in `fastapi/sse.py` shows only the docstring line, and `format_sse_event`'s body is not part of the diff context. It does not. `git diff` between master (`50113da16`) and this branch touches nothing else in `sse.py`; the function is unchanged. `test_format_sse_event_without_fields` passes as-is against master's original `sse.py`, before this PR applies: `format_sse_event` unconditionally appends two empty strings to `lines` before joining with `\n`, so with no other field set that join is already exactly one `\n`. The test locks that existing output against the corrected description rather than covering anything new.

I ran the rest of `tests/test_sse.py` both ways too: 39 passed on master, 40 here, no regressions.

*Disclaimer: This message was composed and posted by a Claude Code agent. I, Leonardo, a human, provided direct supervision, and I take responsibility for the content. Following the community policies, I believe I have put more effort into this message than a contributor would need to put to review it. If you believe this contribution was counterproductive, feel free to contact me here, on LinkedIn, or via email.*

PR #16357, opened 2026-09-15
Fix AttributeError when sending empty form value for sequence of bytes (fixes #9364)

VimalN2005:
### Description
Fixes #9364.

When sending an empty form value (`""`) for sequence bytes annotations (e.g. `list[bytes]` or `Annotated[list[bytes], File()]`) from Swagger UI or other clients, `_extract_form_body` in `fastapi/dependencies/utils.py` unconditionally called `await sub_value.read()`. Because `sub_value` is an empty string `""` rather than an `UploadFile`, an `AttributeError: 'str' object has no attribute 'read'` was raised.

### Solution
As discussed in #9364 and incorporating feedback from @Jeanlaris:
- Guard `sub_value` inside the `is_bytes_sequence_annotation` loop to check `if isinstance(sub_value, UploadFile) or hasattr(sub_value, "read"):` before awaiting `.read()`.
- Primitive string values fall back cleanly to `results.append(sub_value)`, matching the existing type annotation `results: list[bytes | str] = []` and allowing Pydantic to handle downstream validation cleanly.

### Tests
- Added unit regression tests in `tests/test_optional_file_list.py` covering:
  - Optional bytes list with empty form value (`data={"files": ""}`).
  - Annotated bytes list with empty form value.
  - Required bytes list with empty form value.
- All 10 file/form tests and full test suites pass.


YuriiMotov:
@VimalN2005 , thanks for your interest and efforts, but please, follow the rules of the repository.

>

Thanks for understanding!

See: https://github.com/fastapi/fastapi/discussions/9364#discussioncomment-18446707

PR #16358, opened 2026-09-15
Migrate `httpx` to `httpx2`

zach-overflow:
## What?

* Swaps the test-client dependency in the `standard`, `standard-no-fastapi-cloud-cli` and `all` extras from `httpx >=0.23.0,<1.0.0` &rarr; `httpx2 >=2.0.0`, same for the `github-actions` dependency group. `fastapi.testclient` itself needs no change — Starlette's `TestClient` already prefers `httpx2` and falls back to `httpx` (encode/starlette#3291).
* Moves the repo's own `httpx` usage to `httpx2`: `scripts/sponsors.py`, `scripts/notify_translations.py`, the `scripts/playwright` screenshot helpers, and the `docs_src/async_tests` example.
* Updates the docs (English plus all 12 translations) and `README.md` to name `httpx2` and point at https://pydantic.dev/docs/httpx2/.
* Keeps `httpx` in the `tests` group, solely so the existing `without-httpx2` CI job still has a client to fall back to.

## Why?

* related: #16186
* `httpx` is unmaintained. `httpx2` is the maintained fork of `0.28.1` with the same public API, and Starlette's `TestClient` already prefers it — so `fastapi[standard]` currently installs the deprecated stack and only avoids the deprecation warning by accident of resolution order.

## Additional notes

* The `## HTTPX` heading in `advanced/async-tests.md` becomes `## HTTPX2`, and the `add-permalinks-pages --update-existing` hook regenerates anchors from titles, so `#httpx` &rarr; `#httpx2`. External deep links to that section break; I don't see a way to keep the old anchor without fighting the hook.
* `docs_src/async_tests/app_a_py310/test_main.py` now imports `httpx2` directly, which is exactly what the `without-httpx2` job uninstalls. I gated its wrapper in `tests/` with `pytest.importorskip` rather than put a conditional import into example code that users read.
* Translated prose got a mechanical rename. I fixed the Turkish suffixes that broke because `httpx2` is vowel-final where `httpx` wasn't (`HTTPX2'yi`, `HTTPX2'nin`); the other languages may still want a native reviewer's eye.
* Full suite passes both ways: 3348 passed with `httpx2`, and 3344 passed / 18 skipped with it uninstalled.
* This touches `pyproject.toml` and `uv.lock`, so `guard-dependencies` will likely auto-close it — happy to move the proposal back into #16186 if that's preferred.
* Generated-by: Claude Code (Opus 5). I pointed it at the repo's `httpx` usage; it did the dependency, source and docs changes and verified both CI paths. I reviewed the diff and wrote this description.


zach-overflow:
For posterity, I posted a request in https://github.com/fastapi/fastapi/discussions/16186#discussioncomment-18448578 to see about reviving this PR given the auto-close policy

PR #16363, opened 2026-09-15
Escape docs UI HTML fields

failsafesecurity:
## Summary
- escape docs UI values rendered into HTML text and attributes
- serialize Swagger UI JavaScript string values as JSON instead of hand-quoted strings
- add regression coverage for Swagger UI and ReDoc field escaping

## Tests
- uv run pytest tests/test_swagger_ui_escape.py tests/test_application.py::test_swagger_ui tests/test_custom_swagger_ui_redirect.py::test_swagger_ui tests/test_local_docs.py


YuriiMotov:
Thanks for your interest, but please follow the rules of the repository.

>


I don't think we need this - these parameters are not supposed to come from untrusted user input, so it's not a security issue.

See also: https://github.com/fastapi/fastapi/pull/14518#issuecomment-3904249518

PR #16370, opened 2026-09-17
📝 Add warning for duplicate OpenAPI path converters and tests

Shridharrrr:
### Summary
When two routes use different Starlette path converters on the same path template (e.g. `{id:int}` and `{id:str}`), FastAPI's OpenAPI generator strips the converter suffix via `route.path_format`, causing both routes to normalize to the same key (e.g. `/items/{id}`).

The second route silently overwrites the first in the `paths` dict.

This means:
- The **runtime router** (Starlette) matches the **first** registered route
- The **OpenAPI spec** documents the **last** registered route
- Swagger UI shows a handler that **cannot be reached at runtime**

This bug was reported in Discussion #15104 and a FastAPI collaborator confirmed it should be fixed.

### Minimal repro

```python
from fastapi import FastAPI

app = FastAPI()

@app.get("/items/{item_id:str}")
def get_str(item_id: str):
    return {"id": item_id}

@app.get("/items/{item_id:int}")
def get_int(item_id: int):
    return {"id": item_id}

# Runtime executes get_str (first match)
# OpenAPI spec documents get_int (last registered) ← wrong
```

### What this PR does

- **`fastapi/openapi/utils.py`**: adds a `warnings.warn()` in `get_openapi()` when a route (or webhook) would overwrite an existing OpenAPI path+method combination, with a message that names the colliding path and methods and explains the converter mismatch cause.

- **`tests/test_openapi_path_converter_collision.py`**: regression tests covering collision detection, the no-collision case, and a case with same param name vs different param names.

- **`docs/en/docs/tutorial/path-params.md`**: adds a note under the "Order Matters" section explaining that different path converters on the same path template collapse to the same OpenAPI path, and recommends using distinct parameter names.

## Checklist
- [x] The code change is tested and all existing tests pass
- [x] Tests covering the changes have been added
- [x] Docs have been updated where relevant
- [x] I have read the [Contributing Guide](https://fastapi.tiangolo.com/contributing/)


YuriiMotov:
image

Meanwhile...

image

:)

---

YuriiMotov:
@Shridharrrr, thanks for your interest and efforts, but please, follow the rules of the repository.

>

Thanks for understanding!

PR #16371, opened 2026-09-17
📝 Add testing reference to the Agent Library Skill

diogopelinson:
The Agent Library Skill covers dependencies, path operations, Pydantic, responses, streaming, and tooling, but has nothing on testing. This PR adds `references/testing.md`, plus a short `Testing` section and a Quick Reference entry in `SKILL.md`.

The reference covers:

* `TestClient` with regular `def` tests.
* Creating the client in a `with` block to run the `lifespan`. If you skip it, nothing errors: requests still return `200`, but they run against state the `lifespan` never set up.
* `app.dependency_overrides`: override the innermost dependency so the outer dependency's parameters are still validated, and reset the overrides in a fixture so they don't leak into later tests.
* WebSockets with `websocket_connect()`.
* Async tests: keep using `TestClient` inside `async def` tests, and use `httpx.AsyncClient` with `ASGITransport` only when the path operation has to run in the same event loop as the test.

## Two points that differ from the current docs

* `docs/en/docs/advanced/async-tests.md` says that with async tests "we can no longer use the `TestClient` inside our test functions". It does work: `TestClient` runs the app in its own event loop on a separate thread (the AnyIO blocking portal), so requests, WebSockets, and the `lifespan`, including state yielded from it, all work inside an `async def` test. The skill recommends `TestClient` there. `ASGITransport` doesn't run the `lifespan`, and calling `app.router.lifespan_context(app)` by hand runs it but doesn't pass yielded state to `request.state`.
* The skill tells agents to declare the `anyio_backend` fixture. Since AnyIO 4.12, the default fixture is parametrized over the *installed* backends, so a missing fixture no longer fails when Trio isn't installed. It does run each test twice when Trio is installed, as it is in this repo's dev dependencies. The wording matches that behavior.

## Verification

* I extracted every Python block in `references/testing.md` and the new `SKILL.md` section and ran them with pytest against FastAPI installed from this branch: Starlette 1.6.0, HTTPX 0.28.1, Pydantic 2.13.5, AnyIO 4.15.1, Python 3.10. All 8 tests pass. I ran the 7 tests from `references/testing.md` both with and without Trio installed.
* In those 7 tests, I changed only the `assert` lines (expected status codes and bodies), and all 7 failed, so none of them pass without checking anything.
* I gave one small app (a `lifespan`, a dependency that calls an external service, and a WebSocket) to agents asked to write an async test suite:
  * Without the reference, the agent handled the WebSocket by writing the raw ASGI scope and messages itself. For the `lifespan`, it used `app.router.lifespan_context(app)`, which would silently lose `lifespan` state.
  * With the reference, the agent wrote a short suite that uses `TestClient` inside `async def` tests, and it passed.
  * For a sync suite, agents already did the right thing without the reference, and still did with it.


YuriiMotov:
@diogopelinson, thanks for your interest and efforts, but I think we don't need this.

I've just discussed it with Sebastian and he clarified that the idea with library skills is to **teach the agents things they don't already know**. In particular to teach them new features that didn't exist when they were trained (e.g. using `yield` to stream data), or to correct deprecated behaviors they might have because a lot of code uses it and we changed the best practices (e.g. not using `= Depends()` but instead `Annotated[X, Depends()]`.

Also, please read our [contribution guidelines](https://tiangolo.com/open-source/contributing/) and follow the [PR template](https://github.com/fastapi/.github/blob/main/.github/pull_request_template.md?plain=1)

PR #16374, opened 2026-09-18
⚡️ Avoid importing pydantic.v1 in Pydantic v2-only apps

tomaDev:
Follow-up to discussion #16355.

## Problem

Registering the first route imports the whole `pydantic.v1` package, even in an app that only uses Pydantic v2:

```
app.py                         @app.get("/items")
  fastapi/routing.py             add_api_route -> get_dependant -> analyze_param
    fastapi/utils.py               annotation_is_pydantic_v1(type_)
      fastapi/_compat/shared.py      from pydantic import v1
```

Since #14609 dropped `pydantic.v1` support, both callers of this detection (`fastapi/utils.py` and `fastapi/encoders.py`) exist only to raise `PydanticV1NotSupportedError`. For a v2-only app the answer is always `False`, and the import is never used for anything else. The `encoders.py` call site can also hit it during request serialization for any object that falls through `ENCODERS_BY_TYPE`.

Cost measured on FastAPI 0.141.1 / Pydantic 2.13.5 / CPython 3.14:

| | |
|---|---|
| AWS Lambda, arm64, 512 MB, cold start | **44 ms** median of 5 |
| Local, warm bytecode, `fastapi` and `pydantic` already imported | 4.2 ms median of 5 |
| On disk | 1.2 MB |

## Change

A `pydantic.v1` model class cannot exist unless `pydantic.v1` has already been imported, so `"pydantic.v1" in sys.modules` is a necessary condition for either detection function to ever return `True`. Both `is_pydantic_v1_model_class` and `is_pydantic_v1_model_instance` now return `False` early when that key is absent, before touching the import.

Behavior is unchanged for anyone who does define a v1 model: they imported `pydantic.v1` to do so, so they still get `PydanticV1NotSupportedError` with the same message. The only way to bypass the check is deleting `sys.modules["pydantic.v1"]` after defining a model, which is not a supported pattern.

## Tests

Two tests added to `tests/test_compat.py`, both using `monkeypatch.delitem(sys.modules, "pydantic.v1", raising=False)` so they are independent of test order:

- unit: both detection functions return `False` for a v2 model and `pydantic.v1` stays out of `sys.modules`
- integration: an app with a v2 body model handles a request end to end without importing `pydantic.v1`

The existing `tests/test_pydantic_v1_error.py` still passes, covering the branch after the guard.


YuriiMotov:
@tomaDev, thanks for your interest and efforts, but please, follow the rules of the repository.

>

Thanks for understanding!

PR #16377, opened 2026-09-19
docs: fix typos converter and with title

GhaythBenAbid:
Fixes small docs/test-fixture typos:

- `docs/en/docs/tutorial/path-params.md` and `docs/hi/docs/tutorial/path-params.md`: fix heading `Path convertor` → `Path converter` (keep anchor `#path-convertor` for URL stability)
- `scripts/tests/test_translation_fixer/test_complex_doc/data/en_doc.md`: fix `qith` → `with` in heading

Minimal, docs-only. No behavior change.


YuriiMotov:
Here we refer to Starlette's [convertors](https://github.com/Kludex/starlette/blob/main/starlette/convertors.py), so we should probably use the same spelling to avoid confusion.

Thanks!

PR #16381, opened 2026-09-21
Fix double space in get_swagger_ui_html docstring

haimingZZ:
Removed an accidental double space between HTML and that in the docstring of get_swagger_ui_html in fastapi/openapi/docs.py.


YuriiMotov:
Thanks for reporting this!

Thanks for understanding!

PR #16385, opened 2026-09-21
fix: serialize WebSocket validation errors to JSON string in close reason

RushikeshTelrandhe:
The default `websocket_request_validation_exception_handler` currently passes the result of `jsonable_encoder(exc.errors())` directly to `WebSocket.close()` as the `reason`.

`jsonable_encoder(exc.errors())` returns a list, while `WebSocket.close()` expects `reason` to be a `str | None`. As a result, WebSocket request-validation errors can cause the exception handler itself to fail instead of closing the connection with the intended `1008` policy-violation code.

This change serializes the validation errors with `json.dumps()` before passing them to `WebSocket.close()`. The reason is therefore a valid string containing machine-readable JSON.

A regression test was added to verify that:

- WebSocket request-validation failures close with code `1008`.
- The close reason is a string.
- The reason contains the expected validation error.
- The behavior fails with the previous implementation and passes with this fix.

The change is intentionally minimal and does not alter the validation behavior or the WebSocket close code.


RushikeshTelrandhe (commented):
The `test-redistribute-alls-green` failure appears to be a downstream failure of `test-redistribute`.

The `test-redistribute` job is failing during test collection because Starlette's `TestClient` accesses the deprecated `anyio.abc.BlockingPortal` alias, which raises a `DeprecationWarning` in the redistribution environment.

The failure occurs before the tests related to this PR are executed. The other checks are passing.

No changes have been made to address this unrelated failure.

YuriiMotov:
@RushikeshTelrandhe, thank you for your interest and efforts!

This has already been reported and discussed in https://github.com/fastapi/fastapi/discussions/16197 (see point №2)

We track this internally and will work on it when we have time.

I'll close this PR for now, and link it to our internal task

PR #16386, opened 2026-09-22
🐛 Fix stale Content-Length header for 205 Reset Content responses

prownd:
When a path operation returns content with a status code that must not have a body (e.g. 205 set via status_code or the Response parameter), FastAPI empties response.body, but the Content-Length header was already computed by the response class from the original content. Servers like Uvicorn then fail with "Response content shorter than Content-Length".

Reset the Content-Length header to 0 when the body is dropped.

## Pull Request

Discussion: 

## AI Disclaimer

<details>

</details>

## Checklist

- [ ] Coverage stays at 100%.


YuriiMotov:
@prownd, thank you for pointing to this!

We should investigate the specification more in depth to ensure we don't break something else.

For example, 304 response can contain Content-Length header field ([ref](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length)):

> A server MAY send a Content-Length header field in a [304 (Not Modified)](https://www.rfc-editor.org/rfc/rfc9110.html#status.304) response to a conditional GET request ([Section 15.4.5](https://www.rfc-editor.org/rfc/rfc9110.html#status.304)); a server MUST NOT send Content-Length in such a response unless its field value equals the decimal number of octets that would have been sent in the content of a [200 (OK)](https://www.rfc-editor.org/rfc/rfc9110.html#status.200) response to the same request.


We ask people to start with discussion instead of opening PRs. See https://tiangolo.com/open-source/contributing/ and [PR description template](https://github.com/fastapi/.github/blob/main/.github/pull_request_template.md?plain=1):

>

I will close this PR for now and link it to our internal task. We will take a closer look when we have time to work on it.

Thank you ☕🍪

prownd:
Okay, thank you for your explanation.
Then close it.

PR #16387, opened 2026-09-22
🐛 Fix endpoint context shared and mutated across requests

prownd:
_extract_endpoint_context() caches the context dict per endpoint function and returned the cached dict itself. The request handler (and the WebSocket handler) then set endpoint_ctx["path"] on it, mutating the shared cached object.

When the same endpoint function is used by more than one route (or the same route handles concurrent requests), the path stored in a RequestValidationError / ResponseValidationError could be overwritten by another request, so error messages and logs reported the wrong route.

Return a copy of the cached context so each request gets its own dict.

## Pull Request

Discussion: 

## AI Disclaimer

<details>

</details>

## Checklist

- [ ] Coverage stays at 100%.


ege-arhan (commented):
The `.copy()` fixes the shared mutation, but the cache key `id(func)` (`fastapi/routing.py:275-298`, `_endpoint_context_cache: dict[int, EndpointContext]`) is still unsafe: the dict holds only `ctx`, no reference to `func`, so after GC the id can be recycled by a different endpoint and return stale `file`/`line` (CPython reuses ids aggressively when short-lived endpoint funcs are created/deleted, e.g. in tests). It also grows unbounded. Suggestion: validate the cached entry against func identity on hit (or key by weak reference to func) instead of trusting the bare `id()`.

